PR #14, opened 2021-05-07
Update .licenserc.yaml

kezhenxu94:
Fix `licensers.yaml`, some indentations are not correct, and the spdx-id is not needed if you have customized license header.


fulmicoton:
Thank you

PR #18, opened 2021-05-10
Adding the storage implementation

fulmicoton:
Cleaned up and moved the storage abstraction to quickwit.


mosuka (approved):
LGTM.

storage/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+        Ok(())
+    }
+
+    async fn test_write_get_all(storage: &mut dyn Storage) -> anyhow::Result<()> {
+        let test_path = Path::new("write_and_read_all");
+        storage
+            .put(test_path, PutPayload::from(b"abcdef".to_vec()))
+            .await?;
+        let payload = storage.get_all(test_path).await?;
+        assert_eq!(&payload[..], &b"abcdef"[..]);
```
evanxg852000, line 92: Maybe we can stick with one form in regard to the previous test `&b"abcdef"` or `b"abcdef".as_ref()`

storage/src/error.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+/// Generic Result type for storage operations.
+pub type StorageResult<T> = Result<T, StorageError>;
+
+impl StorageError {
+    /// Returns the corresponding `StorageErrorKind` for this error.
+    pub fn kind(&self) -> StorageErrorKind {
+        self.kind
+    }
```
evanxg852000, line 111: What do you think about mergimg into the above `impl StorageError`

storage/src/error.rs
```diff
@@ -0,0 +1,121 @@
...
+/// Storage error kind.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum StorageErrorKind {
+    /// The target index does not exists
+    DoesNotExist,
+    /// The request credentials do not allow for this operation.
+    Unauthorized,
+    /// A third-party service forbids this operation.
+    Service,
+    ///
```
evanxg852000, line 35: being nitty about this empty comment. `InternalError` is self describing thought

storage/src/object_storage/file_slice_stream.rs
```diff
@@ -0,0 +1,118 @@
...
+    inner: ReaderStream<R>,
+    remaining: u64,
+}
+
+impl<R> FileSliceStream<R>
+where
+    R: AsyncRead + AsyncSeek + Unpin,
+{
+    pub async fn try_new(mut reader: R, range: Range<u64>) -> io::Result<Self> {
+        assert!(range.end >= range.start);
```
evanxg852000, line 47: Do we want to allow requesting a fetch of 0 bytes? `range.end >= range.start`
Could possibly be useful for checking object reachability but is this really the intent?
we could also early return if this the case avoiding seeks

storage/src/object_storage/policy.rs
```diff
@@ -0,0 +1,86 @@
...
+            self.max_num_parts > 0,
+            "Misconfiguration: max_num_parts == 0 makes no sense."
+        );
+        if len < self.multipart_threshold_num_bytes || self.max_num_parts == 1 {
+            return len;
+        }
+        let max_num_parts = self.max_num_parts as u64;
+        // complete part is the smallest integer such that
+        // <max_num_parts> * <min_part_len> >= len.
+        let min_part_len = 1u64 + (len - 1u64) / max_num_parts;
```
evanxg852000, line 64: Is there anything sneaky being avoided here with the  +/- 1u64?

storage/src/storage.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+impl<'a> From<&'a [u8]> for PutPayload {
+    fn from(payload_bytes: &[u8]) -> Self {
+        From::from(payload_bytes.to_vec())
+    }
+}
+
+/// Storage meant to receive and server quickwit's split.
```
evanxg852000, line 71:
```suggestion
/// Storage meant to receive and serve quickwit's split.
```

storage/src/storage.rs
```diff
@@ -0,0 +1,96 @@
...
+///
+/// Object storage are the primary target implementation of this trait,
+/// and its interface is meant to allow for multipart download/upload.
+#[async_trait]
+pub trait Storage: Send + Sync + fmt::Debug + 'static {
+    /// Saves a file into the storage.
+    async fn put(&self, path: &Path, payload: PutPayload) -> StorageResult<()>;
+
+    /// Downloads an entire file and write it in a from the storage into a local file.
```
evanxg852000, line 80:
```suggestion
    /// Downloads an entire file and writes it into a local file.
```

storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -0,0 +1,578 @@
...
+
+    async fn create_multipart_requests(
+        &self,
+        payload: PutPayload,
+        len: u64,
+        part_len: u64,
+    ) -> io::Result<Vec<Part>> {
+        assert!(len > 0);
+        let chunks = split_range_into_chunks(len, part_len);
+        // Note that it should really the first chunk, but who knows... and it is very cheap to compute this anyway.
```
evanxg852000, line 262:
```suggestion
        // Note that it should really be the first chunk, but who knows... and it is very cheap to compute this anyway.
```

storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -0,0 +1,578 @@
...
+        let largest_chunk_num_bytes = chunks
+            .iter()
+            .map(|chunk| chunk.end - chunk.start)
+            .max()
+            .expect("The policy should never emit no chunks.");
```
evanxg852000, line 267:
```suggestion
            .expect("The policy should never emit an empty list of chunk.");
```

storage/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+    async fn test_write_and_get_slice(storage: &mut dyn Storage) -> anyhow::Result<()> {
+        let test_path = Path::new("write_and_read_slice");
+        storage
+            .put(
+                test_path,
+                PutPayload::from(b"abcdefghiklmnopqrstuvxyz".to_vec()),
+            )
+            .await?;
+        let payload = storage.get_slice(test_path, 3..6).await?;
+        assert_eq!(&payload[..], b"def".as_ref());
```
evanxg852000, line 82: I think `as_ref` could be omitted
```suggestion
        assert_eq!(&payload[..], b"def");
```

storage/src/storage.rs
fulmicoton: thx

storage/src/error.rs
fulmicoton: Good point! I wonder if there is a lint for that?

storage/src/object_storage/file_slice_stream.rs
fulmicoton: Good point! Let's return an Error if range.end < range.start. It is much better

Fetch 0 bytes is supported but no need to optimize it.

storage/src/object_storage/policy.rs
fulmicoton: Yeah we don't want a normal division. We want the definition above the formula... I don't know a simpler 
way to write it.

Issue #12, opened 2021-04-20
Implement the meta store interface and its impl for its file-backed implementation


guilload:
# Purpose and Challenge

Quickwit needs a way to keep track of the list of splits available for a given index. Currently the metastore's implementation is backed by PostGreSQL.

However, deploying PostgreSQL might be a cumbersome requirement for the user who just wants an index for a single large static dataset in their datalake, or who wants to test out whether Quickwit's search is working as advertised before throwing themselves into deploying multiple services.

For this reason, there will be different implementation of the datastore, the most constrained one, being stored directly on S3.

# API Definition

## Functional need

- Indexer need to publish the splits they have finished building.
- Searcher need to list the splits in an index, possibly pruning them based on some metadata.
- Mergers need to publish the outcome of a merge, (generally the atomic publication of a split and the deletion of some other).
- Retention manager need to be able to remove split based on their timestamps (and possibly some metadata?).

Note that some of these operation imply some interaction with the storage. This is not the role of the meta store, and the correct implementation of its client application is assumed. The meta store semantics however, need to make this coordination possible... (See section below)

## File management need, and the split lifecycle

Quickwit needs a way to ensure that we can cleanup unused files, and this process needs to be resilient to any fail-stop failures.

Because our metastore are typically more constrained than a filesystem (e.g.: object storage), we cannot rely in a write ahead log. We do not want to rely on a strategy that would require to list the files in the store as well, because:
- we might end up removing some user file non related with quickwit.
- listing files is notoriously slow and clumsy on object storage.

Instead, we rely atomically shifting the status of splits.

A split goes into the following lifecycle

```
Building
   -stage->
Staged
   -publish->
Published
   -markAsDelete->
MarkAsDeleted
   -delete->
Deleted
```

Building splits and deleted splits are not states that are registered in the metastore, and should have no durable left-over anywhere.

We maintain the following invariant:
If a split has a files in the storage, it MUST be registered in the meta store, and its state can be either
- Staged:  the splits is almost ready. Some of its files may have been uploaded in the storage.
- Published: the splits is ready and published.
- ScheduledForDelete: the split is scheduled to be deleted.

**Before** creating any file, we need to stage the split. If there is a failure, upon recovery we schedule for delete all of the staged splits.

**A client may not necessarily remove file from storage right after marking it as deleted.** A CLI client may delete files right away, but a more serious deployment should probably only delete these files after a period of time to give some time to running search queries to finish.

# API Definition

## Stage splits (W)

[...]

PR #25, opened 2021-05-13
Add the meta store trait

mosuka:
### Context / purpose
Add the meta store trait. (#12)

### Description
Create a meta store trait to manage index splits.

### How was this PR tested?

### Checklist
- [ ] tested locally
- [ ] added unit tests
- [ ] included documentation


meta-store/src/object_storage/s3_compatible.rs
```diff
@@ -0,0 +1,80 @@
...
+use async_trait::async_trait;
+use rusoto_core::Region;
+
+use quickwit_storage::S3CompatibleObjectStorage;
+
+use crate::meta_store::{IndexId, MetaStore, MetaStoreResult, SplitId, SplitManifest, State};
+
+/// S3 Compatible object storage implementation.
+#[derive(Debug)]
+pub struct S3CompatibleMetaStore {
```
fulmicoton, line 34: This is too specific.

We do not want a S3CompatibleMetaStore... Instead we want a `SingleFileMetaStore` that wraps a `Box<dyn Storage>`.

The benefit is very important:
- other projects will be able to use the ram storage for unit tests.
- we will eventually be able to handle other object storages (non S3 compatible) in the future
- we will be able to use the local filesystem
- we can rely on the storage uri resolver in the CLI.

meta-store/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+edition = "2018"
+
+[dependencies]
+anyhow = '1.0'
+async-trait = '0.1'
+thiserror = '1.0'
+
+quickwit-storage = { version = "0.1.0", path = "../storage" }
+
+[dependencies.rusoto_core]
```
fulmicoton, line 14: there should not be any dependency to rusoto.

This crate should only rely on the Storage abstraction instead of interfacing directly to S3.

meta-store/src/object_storage/s3_compatible.rs
mosuka: @fulmicoton 
Changed it to a single file meta store.

meta-store/Cargo.toml
mosuka: @fulmicoton 
Removed dependency to rusoto.

meta-store/src/meta_store.rs
```diff
@@ -0,0 +1,163 @@
...
+use std::io;
+use std::ops::Range;
+
+use async_trait::async_trait;
+use thiserror::Error;
+
+pub type IndexId = String;
+pub type SplitId = String;
+
+pub enum State {
```
evanxg852000, line 35: @fulmicoton @mosuka @guilload  can we be more specific as In `SplitState` ?

mosuka: @evanxg852000 @fulmicoton @guilload 
Yes, I think `SplitState` is better than `State`. Shall I rename it?

meta-store/src/meta_store.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    // Number of merge this segment has been subjected to during its lifetime.
+    generation: usize,
+}
+
+/// A split manifest carries all meta data about a split
+/// and its files.pub
+pub struct SplitManifest {
+    // the index it belongs to or is destined to belong to.
+    id: IndexId,
```
evanxg852000, line 68: @mosuka  @fulmicoton @guilload  we need to agree on this. `id` might not be what we want I think. we can rename it to `index` and move `split_uri` from `SplitMetaData` to `SplitManifest` as the `Id`.
The Idea is when holding a SplitManifest, I don't have to go down to get the id or uri in order to resolve the storage

meta-store/src/meta_store.rs
fulmicoton: I agree for SplitState

mosuka:
@guilload @fulmicoton @evanxg852000 
I'll separate the PR of adding meta store implementation from the PR of implementation for its file-backed implementation.
Sorry for the inconvenience.
The PR here will be for `MetaStore` trait.

meta-store/src/meta_store.rs
mosuka: @evanxg852000 @guilload @fulmicoton 
Structures have been updated.
Referring to the JSON in the comments below, I have added a new IndexMetaData to include the index information.
https://github.com/quickwit-inc/quickwit/issues/2#issuecomment-839954684
Please give me feedback.

meta-store/src/meta_store.rs
```diff
@@ -0,0 +1,193 @@
...
+    // the splits is ready and published.
+    Published,
+    // the split is scheduled to be deleted.
+    ScheduledForDeleted,
+}
+
+/// A split manifest carries all meta data about a split
+/// and its files.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct SplitManifest {
```
evanxg852000, line 83: When holding a SplitManifest, i have no way of knowing the SplitId. I think we need to have the id here. Also do we want  to separate the slit_uri from the split_id. we could just strop the uri and have a method that extract the splt_id from the split_uri @fulmicoton @guilload

meta-store/src/meta_store.rs
```diff
@@ -0,0 +1,193 @@
...
+use async_trait::async_trait;
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+// pub type IndexId = String;
+
+pub type SplitId = String;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct IndexMetaData {
```
evanxg852000, line 37: @guilload I think this is where we need to have the other members
```rust
 pub timestamp_field: Option<String>,
 pub doc_builder_info: DocBuilderInfo,
```

guilload (approved):

PR #26, opened 2021-05-14
Implement CLI  subcommand stubs

guilload:
### Description
- generate CLI from YAML file
- set up CLI for the `new`, `index`, `search` and `delete` commands
- pass CLI args to placeholder functions

### How was this PR tested?
In debug mode, I ran all the commands adding arguments one by one. For instance:
```bash
RUST_LOG=debug cargo run -- index --index-path ~/foo --input-path ~/bar --temp-dir ~/tmp --num-threads 4 --heap-size 4gib --overwrite
May 14 12:23:50.563 DEBUG quickwit_cli: indexing index_path="/home/guilload/foo" input_path="/home/guilload/bar" temp_dir="/home/guilload/tmp" num_threads=4 heap_size=4294967296 overwrite=true  
```

### Checklist
- [x] tested locally
- [x] added unit tests; do we want some, seems overkill to me for setting up the CLI
- [x] included documentation

### DIFF
--- a/quickwit-cli/src/cli.yaml
+++ b/quickwit-cli/src/cli.yaml
@@ -1,0 +1,107 @@
+name: Quickwit CLI
+about: Index and search structured or unstructured data from the command line
+author: Quickwit, Inc. <hello@quickwit.com>
+
+subcommands:
+    - new:
+        about: Creates an index
+        args:
+            - index-path:
+                help: Location of the target index
+                long: index-path
+                value_name: INDEX PATH
+                required: true
+            - timestamp-field:
+                help: Creates a time-series index with the specified timestamp field
+                long: timestamp-field
+                value_name: FIELD NAME
+            - no-timestamp-field:
+                help: Creates a non-time-series index
+                long: no-timestamp-field
+            - overwrite:
+                help: Overwrites pre-existing index
+                long: overwrite
+        groups:
+            - index-type:
+                required: true
+                args:
+                    - timestamp-field
+                    - no-timestamp-field
+    - index:
+        about: Indexes a dataset
+        args:
+            - index-path:
+                help: Location of the target index
+                long: index-path
+                value_name: INDEX PATH
+                required: true
+            - input-path:
+                help: Location of the source dataset
+                long: input-path
+                value_name: INPUT PATH
+            - temp-dir:
+                help: Creates intermediate files in this directory
+                long: temp-dir
+                value_name: TEMP DIR
+                default_value: /tmp
+            - num-threads:
+                help: Number of threads allocated to the process
+                long: num-threads
+                value_name: NUM THREADS
+                default_value: '2'
+            - heap-size:
+                help: Amount of memory allocated to the process
+                long: heap-size
+                value_name: HEAP SIZE
+                default_value: 2G
+            - overwrite:
+                help: Overwrites pre-existing index
+                long: overwrite
+    - search:
+        about: Searches an index
+        args:
+            - index-path:
+                help: Location of the target index
+                long: index-path
+                value_name: INDEX PATH
+                required: true
+            - query:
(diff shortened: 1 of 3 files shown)


fulmicoton (changes requested):
The current approach makes unit testing and evolution difficult.

I think it is worth having a struct for each command.

IndexCliOption, SearchCliOption, NewCliOption...

And a enum 

enum CliOptions {
   Index(IndexCliOption),
   Search(SearchCliOption),
   //...
}

And have an isolated function dedicate to parse cli args into a CliOptions.

It would allow us to:
- unit test this part
- avoid edit all command call sites everytime we add an option.
- the number of options will likely inflate. Having many positional arguments that change a lot is error prone. A struct would disambiguate this.

quickwit-cli/src/cli.yaml
```diff
@@ -0,0 +1,107 @@
+name: Quickwit CLI
+about: Index and search structured or unstructured data from the command line
+author: Quickwit, Inc. <hello@quickwit.com>
+
+subcommands:
+    - new:
+        about: Creates an index
+        args:
+            - index-path:
```
fulmicoton, line 9: index-uri?

quickwit-cli/src/cli.yaml
```diff
@@ -0,0 +1,107 @@
...
+                long: temp-dir
+                value_name: TEMP DIR
+                default_value: /tmp
+            - num-threads:
+                help: Number of threads allocated to the process
+                long: num-threads
+                value_name: NUM THREADS
+                default_value: '2'
+            - heap-size:
+                help: Amount of memory allocated to the process
```
fulmicoton, line 53: "This memory budget is split between the indexing threads."

quickwit-cli/src/cli.yaml
```diff
@@ -0,0 +1,107 @@
...
+                long: heap-size
+                value_name: HEAP SIZE
+                default_value: 2G
+            - overwrite:
+                help: Overwrites pre-existing index
+                long: overwrite
+    - search:
+        about: Searches an index
+        args:
+            - index-path:
```
fulmicoton, line 63: index-uri?

quickwit-cli/src/cli.yaml
```diff
@@ -0,0 +1,107 @@
...
+                help: Location of the target index
+                long: index-path
+                value_name: INDEX PATH
+                required: true
+            - query:
+                help: Query expressed in Tantivy syntax
```
fulmicoton, line 69: // TODO link to documentation?

quickwit-cli/src/cli.yaml
```diff
@@ -0,0 +1,107 @@
...
+                long: start-offset
+                value_name: OFFSET
+                default_value: '0'
+            - search-fields:
+                help: Searches only in those fields
+                long: search-fields
+                value_name: FIELD NAME
+                multiple: true
+            - start-timestamp:
+                help: Filters out documents before that timestamp (time-series indexes only)
```
fulmicoton, line 89: indexes with timestamp only

quickwit-cli/src/cli.yaml
```diff
@@ -0,0 +1,107 @@
...
+                long: start-timestamp
+                value_name: TIMESTAMP
+            - end-timestamp:
+                help: Filters out documents after that timestamp (time-series indexes only)
```
fulmicoton, line 93: indexes with timestamp only

quickwit-cli/src/cli.yaml
guilload: In find that "uri" is not very user-friendly so I'd rather use "path" externally and "uri" internally. WDYT?

fulmicoton (approved):

PR #29, opened 2021-05-17
Basic File Storage Impl

evanxg852000:
### Context / purpose
The quickwit-cli needs to work on both s3 and file system storage. this includes a File Storage implementation in the storage crate. https://github.com/quickwit-inc/quickwit/issues/20

### Description
Added an implementation of the File Storage 

### How was this PR tested?
Unit test have been written, thought not much to be done since the abstraction was already well in place. 

### Checklist
- [x] tested locally
- [x] added unit tests
- [x] included documentation
- [x] proof read

### DIFF
--- a/storage/src/error.rs
+++ b/storage/src/error.rs
@@ -111,9 +111,9 @@ impl StorageError {
 
 impl From<io::Error> for StorageError {
     fn from(err: io::Error) -> StorageError {
-        StorageError {
-            kind: StorageErrorKind::Io,
-            source: anyhow::Error::from(err),
+        match err.kind() {
+            io::ErrorKind::NotFound => StorageErrorKind::DoesNotExist.with_error(err),
+            _ => StorageErrorKind::Io.with_error(err),
         }
     }
 }
--- a/storage/src/file_storage.rs
+++ b/storage/src/file_storage.rs
@@ -1,0 +1,201 @@
+/*
+    Quickwit
+    Copyright (C) 2021 Quickwit Inc.
+
+    Quickwit is offered under the AGPL v3.0 and as commercial software.
+    For commercial licensing, contact us at hello@quickwit.io.
+
+    AGPL:
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as
+    published by the Free Software Foundation, either version 3 of the
+    License, or (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+use crate::{PutPayload, Storage, StorageErrorKind, StorageFactory, StorageResult};
+use async_trait::async_trait;
+use bytes::BytesMut;
+use std::fmt;
+use std::io::SeekFrom;
+use std::ops::Range;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+use tokio::fs;
+use tokio::io::{AsyncReadExt, AsyncSeekExt};
+
+/// File system compatible storage implementation.
+#[derive(Clone)]
+pub struct FileStorage {
+    root: PathBuf,
+}
+
+impl fmt::Debug for FileStorage {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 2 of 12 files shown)


storage/src/file_storage.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    fn uri(&self) -> String {
+        format!("file://{}", self.root.to_string_lossy())
+    }
+}
+
+/// A File storage resolver
+#[derive(Debug, Clone)]
+pub struct FileStorageFactory {
+    protocol: String,
```
fulmicoton, line 120: we don't need a protocol member.

storage/src/file_storage.rs
```diff
@@ -0,0 +1,200 @@
...
+    fn default() -> Self {
+        FileStorageFactory {
+            protocol: "file".to_string(),
+        }
+    }
+}
+
+impl StorageFactory for FileStorageFactory {
+    fn protocol(&self) -> String {
+        self.protocol.clone()
```
fulmicoton, line 133: "file".to_string()

storage/src/file_storage.rs
```diff
@@ -0,0 +1,200 @@
...
+}
+
+impl FileStorage {
+    /// Creates a file storage instance given a root path.
+    pub fn new(root: PathBuf) -> Self {
+        Self { root }
+    }
+
+    /// Creates a file storage instance given a uri
+    pub fn from_uri(uri: &str) -> StorageResult<FileStorage> {
```
fulmicoton, line 52: Im ok with that...
I think I would have prefered something that takes a filepath however, and extract the filepath from the uri in the Factory instead.

fulmicoton: Ah you also have the FileStorage::new, so no need to change anything I think.

storage/src/file_storage.rs
```diff
@@ -0,0 +1,200 @@
...
+            .unwrap();
+        assert_eq!(err.kind(), StorageErrorKind::DoesNotExist);
+
+        let err = file_storage_factory.resolve("test://").err().unwrap();
+        assert_eq!(err.kind(), StorageErrorKind::DoesNotExist);
+        Ok(())
+    }
+
+    // #[tokio::test]
+    // async fn test_file_storage_builder() -> anyhow::Result<()> {
```
fulmicoton: no need for a builder.

storage/src/storage_resolver.rs
```diff
@@ -48,6 +48,7 @@ impl Default for StorageUriResolver {
             per_protocol_resolver: Default::default(),
         };
         resolver.register(RamStorageFactory::default());
+        resolver.register(FileStorageFactory::default());
```
fulmicoton, line 51: sweet you even thought about adding the FileStorageFactory !

storage/src/file_storage.rs
evanxg852000: I was still cleaning it. so I just pushed my latest proofread

guilload (dismissed):

guilload:
LGTM.

I suggest we improve the naming with something like `LocalFileStorage,` which is more explicit IMO. I think we also return `StorageErrorKind::DoesNotExist` inappropriately in some cases. We probably need more error types (`InvalidInput`?). That's not something specific to your PR, though. We can fix that later.

evanxg852000:
Thanks @guilload for the feedback. I will discuss this @fulmicoton.

storage/src/local_file_storage.rs
```diff
@@ -0,0 +1,191 @@
...
+    async fn copy_to_file(&self, path: &Path, output_path: &Path) -> StorageResult<()> {
+        let full_path = self.root.join(path);
+        fs::copy(full_path, output_path).await?;
+        Ok(())
+    }
+
+    async fn get_slice(&self, path: &Path, range: Range<usize>) -> StorageResult<Vec<u8>> {
+        let full_path = self.root.join(path);
+        let mut file = fs::File::open(full_path).await.map_err(|_| {
+            StorageErrorKind::DoesNotExist.with_error(anyhow::anyhow!(
```
fulmicoton, line 89: can you implement a From<io::Error> for storage error that has specific logic for the DoesNotExist case based on the io::ErrorKind.

storage/src/error.rs
```diff
@@ -111,9 +111,9 @@ impl StorageError {
 
 impl From<io::Error> for StorageError {
     fn from(err: io::Error) -> StorageError {
-        StorageError {
-            kind: StorageErrorKind::Io,
-            source: anyhow::Error::from(err),
+        match err.kind() {
+            io::ErrorKind::NotFound => StorageErrorKind::DoesNotExist.with_error(err),
```
fulmicoton, line 115: yes

fulmicoton (approved):

PR #30, opened 2021-05-17
Add `exists` method to `Storage` trait

guilload:
### Context / purpose
Backporting this patch from the Common Crawl repo. I fixed a few things while I was at it (doc, missing retries).

### Description
The implementation is trivial for ram storage. For S3, we issue a head object request. A `404` or `NoSuchKey` response means that the object does not exist.

It is used in CommonCrawl to check for the presence of a success file.

### How was this PR tested?
The patch was tested while I was building the Common Crawl index. That's how I found out about the Rusoto bug.

### Checklist
- [x] tested locally
- [x] added unit tests
- [x] included documentation

### DIFF
--- a/storage/src/object_storage/error.rs
+++ b/storage/src/object_storage/error.rs
@@ -27,13 +27,14 @@ use std::io;
 use rusoto_core::RusotoError;
 use rusoto_s3::{
     AbortMultipartUploadError, CompleteMultipartUploadError, CreateMultipartUploadError,
-    DeleteObjectError, GetObjectError, PutObjectError, UploadPartError,
+    DeleteObjectError, GetObjectError, HeadObjectError, PutObjectError, UploadPartError,
 };
 
 use crate::retry::IsRetryable;
 use crate::{StorageError, StorageErrorKind};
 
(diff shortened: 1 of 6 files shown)


storage/src/object_storage/error.rs
```diff
@@ -27,13 +27,14 @@ use std::io;
...
     AbortMultipartUploadError, CompleteMultipartUploadError, CreateMultipartUploadError,
-    DeleteObjectError, GetObjectError, PutObjectError, UploadPartError,
+    DeleteObjectError, GetObjectError, HeadObjectError, PutObjectError, UploadPartError,
 };
 
 use crate::retry::IsRetryable;
 use crate::{StorageError, StorageErrorKind};
 
-pub struct RusotoErrorWrapper<T: StdError>(RusotoError<T>);
+pub struct RusotoErrorWrapper<T: StdError>(pub RusotoError<T>);
```
guilload, line 36: `.source()` returns a `dyn Error` which prevented me for inspecting the inner `RustoError`.

fulmicoton (approved):

PR #31, opened 2021-05-19
Index command [WIP]

evanxg852000:
### Context / purpose
A small PR proposing to move all subcommand in their own module

### Description
- I have moved the command parsing in `cli_command` module
- the index command is now inside the `indexing` module
- Included tokio and default help when  no command is passed

### How was this PR tested?

### Checklist
- [x] tested locally
- [ ] added unit tests
- [ ] included documentation


quickwit-cli/src/indexing/split_metadata.rs
```diff
@@ -0,0 +1,69 @@
...
+use serde::{Deserialize, Serialize};
+use std::ops::Range;
+use uuid::Uuid;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct SplitMetaData {
+    // The split id
+    pub split_id: Uuid,
+
+    // Split uri. In spirit, this uri should be self sufficient
```
fulmicoton, line 32: we need to change this comment. the `split_uri` can be relative to the index_uri.

fulmicoton, line 32: we need to change this comment. the `split_uri` can be relative to the index_uri.

quickwit-cli/src/indexing/byte_units.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+use std::convert::TryFrom;
+
+/// Base 2 bytes.
+#[derive(Clone, Debug)]
+pub enum ByteUnits {
```
fulmicoton, line 27: `impl Serialize` and have it output Bytes.

quickwit-cli/src/indexing/byte_units.rs
```diff
@@ -0,0 +1,128 @@
...
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+use std::convert::TryFrom;
+
+/// Base 2 bytes.
+#[derive(Clone, Debug)]
```
fulmicoton, line 26: We can make it `Copy`.

quickwit-cli/src/indexing/byte_units.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// Implementation of String to ByteUnits coversion
+///
+/// The following regex defines the specification:
+/// ^[0-9]+_?(b|bytes|bytes|kb|kib|mb|mib|gb|gib)?$
+/// Examples:
+/// 2, 2b, 2byte, 2_bytes => ByteUnits::Bytes(2)
+/// 2kb, 2kib => ByteUnits::KiB(2)
+/// 2mb, 2mib => ByteUnits::MiB(2)
+/// 2gb, 2gib => ByteUnits::GiB(2)
```
fulmicoton, line 53: Can we replace all of the b by B in the docstring?

I know the lowercase normalization makes it so that both are correct, but kb usually   means kilobits.
We might want to avoid confusing the user.

Ideally we should explicitly explain that in the comment.

"This function does not interpret lowercase b as meaning 'bit'. It interprets 'b' regardless of its case, as Byte."

quickwit-cli/src/indexing/byte_units.rs
```diff
@@ -0,0 +1,128 @@
...
+            ByteUnits::try_from("2mib").unwrap().bytes(),
+            2 * 1024 * 1024
+        );
+        assert_eq!(
+            ByteUnits::try_from("2_mib").unwrap().bytes(),
+            2 * 1024 * 1024
+        );
+
+        assert_eq!(
+            ByteUnits::try_from("7gb").unwrap().bytes(),
```
fulmicoton, line 112: use the casing in all of these tests GB etc.

quickwit-cli/src/indexing/document_indexer.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    // Finalise last split if it has docs
+    if current_split.metadata.num_records > 0 {
+        let split = std::mem::replace(
+            &mut current_split,
+            Split::create(&args, schema.clone()).await?,
+        );
+
+        let finalise_split_task = finalize_split(split, statistic_sender.clone()).await?;
```
fulmicoton, line 90: s/finalise/finalize/g ... Make sure it is replaced everywhere.

quickwit-cli/src/indexing/document_indexer.rs
```diff
@@ -0,0 +1,130 @@
...
+        split_merge_tasks.push(finalise_split_task);
```
fulmicoton, line 91: Can you add a TODO.  we have no push back mechanism at this point...
Meaning if the finalize task cannot keep up with ingestion, we might end up with a lot of those running.

quickwit-cli/src/indexing/document_indexer.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    // for all split finilisation tasks
```
fulmicoton, line 94: the comment is not helping :)

quickwit-cli/src/indexing/document_indexer.rs
```diff
@@ -0,0 +1,130 @@
...
+    let merge_results = futures::future::try_join_all(split_merge_tasks).await?;
+    //TODO let (splits, errors) = merge_results.iter().partition(|item| item.is_ok());
+
+    let splits = merge_results
+        .into_iter()
+        .filter(|result| result.is_ok())
```
fulmicoton, line 100: you can use `filter_map` or `flat_map` instead of the filter / map for that.
That will avoid the unwrap.

At this point I have no clue what an error means however.
What kind of error are we trying to dismiss here?

quickwit-cli/src/indexing/document_indexer.rs
```diff
@@ -0,0 +1,130 @@
...
+use tokio::task::JoinHandle;
+
+use crate::indexing::statistics::StatisticEvent;
+
+use crate::indexing::split::Split;
+
+pub async fn index_documents(
+    args: &IndexDataArgs,
+    mut document_receiver: Receiver<String>,
+    statistic_sender: Sender<StatisticEvent>,
```
fulmicoton, line 36: I like that you thought of a way to collect statistics!
I wonder if a channel is the right way to do this, but we can keep it that way for the moment.

quickwit-cli/src/indexing/document_indexer.rs
```diff
@@ -0,0 +1,130 @@
...
+        };
+
+        current_split.add_document(doc)?;
+        if current_split.has_enough_docs() {
+            let split = std::mem::replace(
+                &mut current_split,
+                Split::create(&args, schema.clone()).await?,
+            );
+
+            let finalise_split_task = finalize_split(split, statistic_sender.clone()).await?;
```
fulmicoton, line 78: The intermediate variable is very helpful for readability! Good job!

quickwit-cli/src/indexing/document_indexer.rs
```diff
@@ -0,0 +1,130 @@
...
+    statistic_sender
+        .send(StatisticEvent::SplitCreated {
+            id: split.id.to_string(),
+            num_docs: split.metadata.num_records,
+            size_in_bytes: split.metadata.size_in_bytes,
+            num_failed_to_parse_docs: split.metadata.num_parsing_errors,
+        })
+        .await?;
+
+    Ok(task::spawn(async move {
```
fulmicoton, line 125: commit and merge are CPU intensive operation. They should not happen in a regular tokio task. 
Checkout `task::spawn_block` 
https://docs.rs/tokio/1.6.0/tokio/index.html#cpu-bound-tasks-and-blocking-code

(You can ignore the story about rayon. We will reduce the maximum number of threads in the blocking thread pool of tokio)

fulmicoton (changes requested):
Can you split your PR in 2, the (hotdirectory stuff in a PR) and then the index command on another PR (on top of the first one).

I also added a bunch of comments. Please have a look at them.

quickwit-cli/src/indexing/byte_units.rs
fulmicoton: After that you can use it directly in the Split structs.

hot-directory/src/caching_directory.rs
```diff
@@ -0,0 +1,336 @@
...
+use std::path::{Path, PathBuf};
+use std::sync::{Arc, Mutex};
+use tantivy::directory::error::{DeleteError, OpenReadError, OpenWriteError};
+use tantivy::directory::{FileHandle, OwnedBytes, WatchHandle, WritePtr};
+use tantivy::{AsyncIoResult, Directory, HasLen};
+use tracing::{info, warn};
+
+#[derive(Hash, Debug, Clone, PartialEq, Eq)]
+struct SliceAddress {
+    path: PathBuf,
```
fulmicoton, line 37: General comment about `hot-directory`.

We can merge it now, but I suspect it does not meet code quality. I'll do a pass over it.

quickwit-cli/src/indexing/document_indexer.rs
evanxg852000: The reason I use channel is that the spec requested to display the throughput as the work is in progress. so this kind of is like an event-based statistic collection.

quickwit-cli/src/indexing/byte_units.rs
evanxg852000: I just found out this was a waste of energy. it exists in this crate `byte-unit`

fulmicoton: it looks perfect indeed

quickwit-cli/src/indexing/document_indexer.rs
evanxg852000: The reason I did not use the spawn_blocking like we did in the barrel repo is that `tantivy::merge_segments`  is not behind a future interface. so we already know this is going to starve the runtime. 
Though I agree this should be the same for `index_writter.commit()` since it is not future-based API. 
But tantivy users should be able to use `index_writer.merge().await` without worrying about this. because this merge interface is already behind a future interface. 
on another note, I thought commit does not do much work. I probably need to dig deeper into that path.

At this stage, it's the commit portion that should go into `tokio::spwn_blocking` i think. and `async index_writer.merge()` should probably make use of `spawn_blocking` internally

quickwit-cli/src/indexing/document_indexer.rs
fulmicoton: Having statistics is a good idea. Using is not necessarily the best way.

quickwit-cli/src/indexing/document_indexer.rs
fulmicoton: index_writer.merge() runs the computation on a different thread. The future it returns is a oneshot...

So yes you are right, having it withing a spawned task is ok.

Commit is not cheap at all... It needs to be behind a spawn_blocking.

quickwit-cli/src/indexing/index.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    let mut statistic_collector = StatisticsCollector::new();
+    let statistic_sender = statistic_collector.start_collection();
+    let (document_sender, document_receiver) = channel::<String>(DOCUMENT_CHANNEL_SIZE);
+    let retrieve_fut = retrieve_documents(args.input_uri.clone(), document_sender);
+    let index_fut = index_documents(&args, document_receiver, statistic_sender.clone());
+    let (_, splits) = try_join!(retrieve_fut, index_fut)?;
+
+    // TODO: this is a little bit subobtimal, we might want to deploy progressively the splits once they are ready
```
fulmicoton, line 50: this is massively suboptimal :).

quickwit-cli/src/indexing/document_indexer.rs
evanxg852000: Just saying that the reason I had doubt was that the merge was already returning a future and the commit was not. so this kind of made me doubt since I thought commit was cheap and therefore was left as it is. I have the commit now wrapped in `spawn_blocking`

fulmicoton: @evanxg852000 if you have a doubt, please investigate the code or ask. 
Committing can have dire consequences, and you cannot rely on code review to catch all of these.

If you do not want to interrupt your flow with that kind of investgation, just drop a TODO that your can revisit after your PR took form.

quickwit-cli/src/indexing/index.rs
fulmicoton: (We can solve it in a different PR however can you open an issue ?)

evanxg852000: Yes, this is something I wanted us to discuss after my prototype but it looks like you guys already saw.   The solution is to start uploading as soon as it's ready but keeping a bound on the number of splits that can be published concurrently.

PR #32, opened 2021-05-19
Add single file meta store implementation

mosuka:
### Context / purpose
Add a single file meta store implementation.

### Description
Simply serialize all of the splits associated with a single index into one large JSON file.
https://github.com/quickwit-inc/quickwit/issues/12

### How was this PR tested?

### Checklist
- [x] tested locally
- [x] added unit tests
- [x] included documentation

### DIFF
--- a/meta-store/src/meta_store.rs
+++ b/meta-store/src/meta_store.rs
@@ -20,6 +20,8 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+mod single_file;
+
 use std::collections::HashMap;
 use std::fmt::{Debug, Display};
 use std::io;
@@ -171,7 +173,6 @@ impl From<io::Error> for MetaStoreError {
     }
 }
 
-#[allow(dead_code)]
 pub type MetaStoreResult<T> = Result<T, MetaStoreError>;
 
 #[async_trait]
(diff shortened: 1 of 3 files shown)


meta-store/src/meta_store.rs
```diff
@@ -61,19 +67,22 @@ pub struct SplitMetaData {
     generation: usize,
 }
 
+/// A split manifest entry.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ManifestEntry {
+    /// The file name.
     file_name: String,
+    /// File siize in bytes.
```
fulmicoton, line 75: typo: size...

Also, the checksum is missing.

But I think we don't want the manifest entry in the meta store anymore so it does not matter I guess.

meta-store/src/meta_store/single_file.rs
```diff
@@ -0,0 +1,1027 @@
...
+                // this split will be ignored.
+                continue;
+            }
+
+            // filter by time range
+            match time_range {
+                Some(ref filter_time_range) => {
+                    match split_manifest.metadata.time_range {
+                        Some(ref split_time_range) => {
+                            if split_time_range.contains(&filter_time_range.start)
```
guilload, line 243: We want to return [2,3,4]. I believe this predicate is going to miss split 3. (hope this renders well)
 ```
           <----------->
 <--1--><--2--><--3--><--4-->
 ````

guilload (changes requested):
The readability of this PR can be greatly improved using `Options` and `Result` combinators such as `Ok_or*` or `map_err`.

PR #33, opened 2021-05-19
Attempt to stitch everything together

guilload:
Attempt to stitch everything together... for lack of a better title.

Most importantly, this PR tries to **simplify and fill the gaps** in the `Metastore` trait to implement the commands we need for the CLI:
- store `split_id` instead of `split_uri`, making a split's URI the concatenation of index URI + split ID; moving and copying split becomes easier
- remove the `SplitManifest`, we don't want to generate large metastore files; the list of files belonging to a split can still be generated from the segment IDs found in the `meta.json` file, everything else is stored in `SplitMetadata`
- rely exclusively on `index_uri` in the `Metastore` APIs **for now**, we will revisit them later when we introduce a SQL implementation
- add `MetastoreUriResolver`, the `MetastoreUriResolverFactory` is still missing though

Other changes:
- rename `MetaStore` and `MetaData` to  `Metastore` and `Metadata`, the latter occurrences are more common
- rename `meta-store` and `storage` folders to `quickwit-metastore` and `quickwit-storage`, having consistent folder and crate names makes writing infrastructure code easier (Dockerfile, Makefile)
- commit basic index API in `quickwit-core`
- add `quickwit-doc-mapping` crate with dummy code, @fmassot

@evanxg852000 @fulmicoton @mosuka PTAL


quickwit-cli/src/main.rs
```diff
@@ -152,17 +155,24 @@ impl CliCommand {
     }
 }
 
-fn create_index(args: CreateIndexArgs) -> anyhow::Result<()> {
+async fn create_index(args: CreateIndexArgs) -> anyhow::Result<()> {
```
fulmicoton, line 158: I'd rather have `create_index_cli` / `create_index`.

quickwit-core/src/index.rs
```diff
@@ -0,0 +1,83 @@
...
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+use quickwit_doc_mapping::DocMappingMetadata;
+use quickwit_metastore::{MetastoreUriResolver, SplitState};
+
+type IndexUri = String;
+
+// anyhow errors are fine for now but we'll want to move to a proper error type eventually.
+pub async fn create_index(
```
fulmicoton, line 29: I think we can put those in different files eventually. No need to do that in that PR though.

quickwit-core/src/lib.rs
```diff
@@ -20,4 +20,4 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-mod meta_store;
```
fulmicoton, line 23: sometimes git diff algorithm is really lame :D

quickwit-doc-mapping/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+pub enum DocMappingType {
+    Dynamic,
+    Opentelemetry,
```
fulmicoton, line 25: OpenTelemetry

quickwit-doc-mapping/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+    Static,
+}
+
+pub struct DocMappingMetadata {
```
fulmicoton, line 29: I think it could be 

```rust
enum DocMapping {
   Dynamic,  
   OpenTelemetry,
   Static(StaticDocMapping),
}
```
and we would just tweak serde to serialize it the way you are suggesting with your DocMappingMetadata structure.
(I forgot if this is the default or not)

quickwit-metastore/src/metastore.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct MetaDataSet {
+    index: IndexMetadata,
+    splits: HashMap<SplitId, SplitMetadata>,
+}
+
+#[async_trait]
+pub trait Metastore: Send + Sync + 'static {
+    /// Creates and index.
```
fulmicoton, line 86: typo: Creates an index.

quickwit-metastore/src/metastore_resolver.rs
```diff
@@ -0,0 +1,82 @@
...
+                uri
+            ))
+        })?;
+        let resolver = self
+            .per_protocol_resolver
+            .get(protocol)
+            .ok_or_else(|| MetastoreResolverError::ProtocolUnsupported(protocol.to_string()))?;
+        let metastore = resolver
+            .resolve(uri)
+            .map_err(MetastoreResolverError::FailedToOpenMetastore)?;
```
fulmicoton, line 79: We need to stitch the StorageResolver wrapping the SingleFileMetaStore as a fallback solution I think.

quickwit-metastore/src/metastore_resolver.rs
```diff
@@ -0,0 +1,82 @@
...
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use crate::{Metastore, MetastoreResolverError};
+
+/// A metastore factory builds a [`Metastore`] object from an URI.
+#[cfg_attr(any(test, feature = "testsuite"), mockall::automock)]
```
fulmicoton, line 29: It might be worth highlighting the intent in the comment here, or put some examples.

At the moment, `SingleFileMetaStore` is the only implementation available, so a metastore uri is simply the index storage uri. (BTW do we add metastore.json to that?)
In the future, we will handle different backend (e.g. postgresql) that will also be identified by an URI.

fulmicoton (changes requested):
Please fix the typo... For the other comments, some of them are arguable so let me think what you think.

fulmicoton (approved):

PR #34, opened 2021-05-20
Add single file metastore implementation

mosuka:
### Context / purpose
Add a single file meta store implementation.

### Description
Simply serialize all of the splits associated with a single index into one large JSON file.
#12

### How was this PR tested?

### Checklist
- [x] tested locally
- [x] added unit tests
- [x] included documentation

### DIFF
--- a/quickwit-metastore/src/metastore.rs
+++ b/quickwit-metastore/src/metastore.rs
@@ -39,6 +39,8 @@ pub type IndexUri = String;
 /// A split ID.
 pub type SplitId = String;
 
+pub static FILE_FORMAT_VERSION: &str = "0";
+
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct IndexMetadata {
     version: String,
--- a/quickwit-metastore/src/metastore.rs
+++ b/quickwit-metastore/src/metastore.rs
@@ -20,6 +20,8 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+pub mod single_file_metastore;
+
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::ops::Range;
@@ -27,8 +29,9 @@ use std::ops::Range;
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 
-use crate::error::MetastoreResult;
 use quickwit_doc_mapping::DocMapping;
+
+use crate::MetastoreResult;
 
 /// An index URI, such as `file:///var/lib/quickwit/indexes/nginx` or `s3://my-bucket/indexes/nginx`.
 pub type IndexUri = String;
@@ -75,8 +78,8 @@ pub enum SplitState {
     ScheduledForDeletion,
 }
 
-#[derive(Debug, Serialize, Deserialize)]
-pub struct MetaDataSet {
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MetadataSet {
     index: IndexMetadata,
     splits: HashMap<SplitId, SplitMetadata>,
 }
--- a/quickwit-metastore/src/metastore/single_file_metastore.rs
+++ b/quickwit-metastore/src/metastore/single_file_metastore.rs
@@ -62,12 +62,10 @@ impl Metastore for SingleFileMetastore {
 
         // Check for the existence of index.
         // Load metadata set If it has already been stored in the storage, if not, create a new one.
-
         let exists = self.storage.exists(&path).await.map_err(|e| {
             MetastoreErrorKind::InternalError
                 .with_error(anyhow::anyhow!("Failed to read storage: {:?}", e))
         })?;
-
         let metadata_set = if exists {
             // Get metadata set from storage.
             let contents = self.storage.get_all(&path).await.map_err(|e| {
--- a/quickwit-metastore/src/metastore/single_file_metastore.rs
+++ b/quickwit-metastore/src/metastore/single_file_metastore.rs
@@ -19,9 +19,11 @@
 use std::collections::HashMap;
 use std::ops::Range;
 use std::path::Path;
-use std::sync::{Arc, RwLock};
+use std::sync::Arc;
 
 use async_trait::async_trait;
+use tokio::sync::RwLock;
+
 use quickwit_doc_mapping::DocMapping;
 use quickwit_storage::{PutPayload, Storage};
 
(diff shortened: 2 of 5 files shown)


quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -0,0 +1,1352 @@
...
+
+use crate::{
+    IndexMetadata, IndexUri, MetadataSet, Metastore, MetastoreErrorKind, MetastoreResult, SplitId,
+    SplitMetadata, SplitState, FILE_FORMAT_VERSION,
+};
+
+/// Single file meta store implementation.
+pub struct SingleFileMetastore {
+    storage: Arc<dyn Storage>,
+    data: Arc<RwLock<HashMap<IndexUri, MetadataSet>>>,
```
fulmicoton, line 36: let's use a tokio RwLock in async functions. 

They are not blocking...

quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -0,0 +1,1352 @@
...
+        {
+            let mut data = self.data.write().unwrap();
+
+            // Check for the existence of index.
+            if !data.contains_key(&index_uri) {
+                return Err(MetastoreErrorKind::IndexDoesNotExist
+                    .with_error(anyhow::anyhow!("Index does not exist: {:?}", &index_uri)));
+            }
+
+            let metadata_set = data.get_mut(&index_uri).unwrap();
```
fulmicoton, line 147: Whenever possible avoid unwrap...

It makes it much easier to make it clear that panicking is not possible.

Here is an idiom that allows you to avoid the Option unwrap.
```rust
let metadata_set = data.get_mut(&index_uri)
    .ok_or_else(|| ...)?;
```

Note your code will actually end up shorter.

If you are not fond of those functional API, it is possible also to do...

```
let metadata_set = 
      if let Some(metadata_set) = ... {
         metadata_set     
     } else {
         return Err(...);
      }
```

quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -0,0 +1,1352 @@
...
+                    .with_error(anyhow::anyhow!("Split already exists: {:?}", &split_id)));
+            }
+
+            // Update split metadata.
+            metadata_set.splits.insert(split_id.clone(), split_metadata);
+
+            // Serialize metadata set.
+            contents = match serde_json::to_vec(&metadata_set) {
+                Ok(c) => c,
+                Err(e) => {
```
fulmicoton, line 161: Alternatively 

```
contents = serde_json::to_vec(&metadata_set)
    .map_err(|e| {
        // ...
    })?;

```

quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -0,0 +1,1352 @@
...
+        _doc_mapping: DocMapping,
+    ) -> MetastoreResult<()> {
+        // Serialized metadata set.
+        let contents: Vec<u8>;
+
+        {
+            let mut data = self.data.write().unwrap();
+
+            // Check for the existence of index.
+            if data.contains_key(&index_uri) {
```
fulmicoton, line 64: There is a recurring pattern here.

Two options to make the code simpler.

# Option 1

Introduce a private fn in the metastore...
```
fn mutate_meta_store<F: FnOnce(Entry<_,_,_>) -> Result<_, _>>
```

# Option 2 (my personal favorite)

Introduce a get and a put method that individually take the lock.
The lock will be taken twice (get as a read, and put as a write, and get would have to clone the metadata).
(Note the clone part might not be great for performance if we have a millions of splits.)

What I like about this solution is that the code is trivial and cannot panic while the lock is being held.

quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -0,0 +1,1352 @@
...
+        // Check for the existence of index.
+        if !data.contains_key(&index_uri) {
+            return Err(MetastoreErrorKind::IndexDoesNotExist
+                .with_error(anyhow::anyhow!("Index does not exist: {:?}", &index_uri)));
+        }
+
+        let metadata_set = data.get_mut(&index_uri).unwrap();
+
+        // filter by split state.
+        let it = metadata_set
```
fulmicoton, line 261: `it` is not a good variable name as it does not help the reader.

`split_with_meta_matching_state_it` would be better.

quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -0,0 +1,1352 @@
...
+                            continue;
+                        }
+                    }
+                }
+                None => {
+                    // if `time_range` is omitted, the metadata is not filtered.
+                    splits.push(split_metadata.clone());
+                }
+            }
+        }
```
fulmicoton, line 295: Can you find a way to make this easier to read?

fulmicoton (changes requested):
Please have a look at the comments.
Generally speaking this does the job, but the code quality could be improved.

quickwit-metastore/src/metastore/single_file_metastore.rs
mosuka: Changed to use tokio RwLock.

quickwit-metastore/src/metastore/single_file_metastore.rs
mosuka: Also fixed a similar code.

quickwit-metastore/src/metastore/single_file_metastore.rs
mosuka: Also fixed a similar code.

quickwit-metastore/src/metastore/single_file_metastore.rs
mosuka: Changed.

quickwit-metastore/src/metastore.rs
```diff
@@ -20,22 +20,27 @@
...
 
+use crate::MetastoreResult;
+
 /// An index URI, such as `file:///var/lib/quickwit/indexes/nginx` or `s3://my-bucket/indexes/nginx`.
 pub type IndexUri = String;
 
 /// A split ID.
 pub type SplitId = String;
 
+pub static FILE_FORMAT_VERSION: &str = "0";
```
fulmicoton, line 42: I think this is not used anywhere.

quickwit-metastore/src/metastore/single_file_metastore.rs
fulmicoton: It might not be as easy as I suggested...

fulmicoton (approved):

PR #38, opened 2021-05-24
Fixes testsuite feature for the storage crate.

fulmicoton:
Also, this adds a mock interface for the storage trait.


mosuka (approved):
LGTM.

PR #40, opened 2021-05-24
Reorganize CLI command handlers into modules

evanxg852000:
### Context / purpose
Since the `quickwit-core` contains the core function, we need a way to pass the CLI arguments (CreateIndexArgs, DeleteIndexArgs, IndexDataArgs, SearchIndexArgs) to those core function.

### Description
I have moved the CLI argument structs to quickwit-core and reorganized the functions into their own module. Index_data is in a folder module because it will need more separation. other modules can do the same if need be.

### How was this PR tested?
- Compiled and ran the test
- Executed few commands

### Checklist
- [x] tested locally

### DIFF
--- a/quickwit-cli/src/main.rs
+++ b/quickwit-cli/src/main.rs
@@ -24,40 +24,9 @@ use anyhow::bail;
 use byte_unit::Byte;
 use clap::{load_yaml, value_t, App, AppSettings, ArgMatches};
 use std::path::PathBuf;
-use tracing::debug;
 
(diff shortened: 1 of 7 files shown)


fmassot:
@evanxg852000 when I see that
`use quickwit_core::{create_index_cli, delete_index_cli, index_data_cli, search_index_cli};`

I'm thinking we are doing the wrong thing. Let's focus on the index PR and we will handle that later.

evanxg852000:
ok

PR #41, opened 2021-05-24
Fix metastore bugs

mosuka:
### Context / purpose
Fix metastore bugs.

### Description
Fixed a bug that the internal state of the metastore is updated even when writing to the storage fails.
In addition, when an existing index is specified in the argument of the `create_index` function of `Metastore` trait, the index has opened, but the function name does not match the behavior, so the `create_index` function and the `open_index function` are now prepared separately.

### How was this PR tested?
Ran `cargo test test_single_file_metastore`

### Checklist
- [x] tested locally
- [x] added unit tests
- [x] included documentation

### DIFF
--- a/quickwit-metastore/src/metastore/single_file_metastore.rs
+++ b/quickwit-metastore/src/metastore/single_file_metastore.rs
@@ -63,44 +63,6 @@ impl SingleFileMetastore {
             data: Arc::new(RwLock::new(HashMap::new())),
         })
     }
-
-    async fn split_exists(&self, index_uri: IndexUri, split_id: SplitId) -> MetastoreResult<bool> {
-        let data = self.data.read().await;
-
-        // Check for the existence of index.
-        let metadata_set = data.get(&index_uri).ok_or_else(|| {
-            MetastoreErrorKind::IndexIsNotOpen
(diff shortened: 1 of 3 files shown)


quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -105,23 +126,80 @@ impl Metastore for SingleFileMetastore {
...
+        let mut data = self.data.write().await;
+        data.insert(index_uri.clone(), metadata_set.clone());
+
         Ok(())
     }
 
-    async fn delete_index(&self, index_uri: IndexUri) -> MetastoreResult<()> {
+    async fn open_index(&self, index_uri: IndexUri) -> MetastoreResult<()> {
+        // Check for the existence of index.
+        let exists = self.index_exists(index_uri.clone()).await.map_err(|e| {
```
fulmicoton, line 138: Why do we check for existence?
Each call is an extra 70ms of latency.

We can rely on get_all directly.

mosuka: You're absolutely right. I'll fix it.

mosuka: Deleted a redundant existence check at `open_index`.

PR #46, opened 2021-05-25
Added lint forbidding missing docs in quickwit-core

fulmicoton:
@evanxg852000 Can you add the docs for the different methods? Some of the specs should be on notion.
@guilload can you review them?

### DIFF
--- a/quickwit-cli/src/main.rs
+++ b/quickwit-cli/src/main.rs
@@ -23,7 +23,6 @@
 use anyhow::{bail, Context};
 use byte_unit::Byte;
 use clap::{load_yaml, value_t, App, AppSettings, ArgMatches};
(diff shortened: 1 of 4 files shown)


evanxg852000:
I have added the docs, but we probably need to agree if this is the way forward. I felt like the function in core are not the same as the once in `quickwit-cli`. The description in [Notion](https://www.notion.so/Quickwit-CLI-2763f23b46ff4532abd4a16293e9e05b) is tailored to what the functions in `quickwit-cli` (create_index_cli, search_index_cli, etc...) do. I have use some but not the full spec for functions in `quickwit-core`. Also, I found that we still have some interfaces not matching the description which is ok since this is a work-in-progress.
@fulmicoton  Please let me know your take.

PR #51, opened 2021-05-25
Implements Index feature [WIP]

evanxg852000:
### Context / purpose
Implements the document indexing feature in quickwit-cli

### How was this PR tested?
for now by running the command 
`cargo run -- index --index-uri file://./mydata --input-path ./logs.txt`
the following will read document from `./log.txt` and create the splits in `./mydata`

As note by @fulmicoton [bug](https://github.com/quickwit-inc/quickwit/issues/59).
The LocalFileStorage needs to exactly behaves like an Object Storage with file separator. 
This currently limits this implementation to working properly with only LocalFileStorage till that bug is fixed. 

### Checklist
- [x] added unit tests
- [x] tested locally

### DIFF
--- a/quickwit-core/src/indexing/document_retriever.rs
+++ b/quickwit-core/src/indexing/document_retriever.rs
@@ -20,34 +20,91 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+use anyhow::Context;
+use async_trait::async_trait;
 use std::path::PathBuf;
-use tokio::fs::File;
 use tokio::io::AsyncBufReadExt;
 use tokio::io::{self, BufReader};
-use tokio::sync::mpsc::Sender;
+use tokio::{
+    fs::File,
+    io::{Lines, Stdin},
+};
+
+#[cfg(test)]
+use std::io::Cursor;
+
+/// A trait for retrieving input content online at a time
+#[async_trait]
+pub trait DocumentRetriever {
+    /// Reads one line from underlying input buffer.
+    async fn next_document(&mut self) -> anyhow::Result<Option<String>>;
+}
 
 /// Retrieves new-line delimited json documents from StdIn or from a local File
-pub async fn retrieve_documents(
-    input_uri: Option<PathBuf>,
-    document_sender: Sender<String>,
-) -> anyhow::Result<()> {
-    match input_uri {
-        Some(path_buf) => {
-            let file = File::open(path_buf).await?;
-            let reader = BufReader::new(file);
-            let mut lines = reader.lines();
-            while let Some(line) = lines.next_line().await? {
-                document_sender.send(line).await?;
+pub enum FileOrStdInDocumentRetriever {
+    StdIn(Lines<BufReader<Stdin>>),
+    File(Lines<BufReader<File>>),
+}
+
+impl FileOrStdInDocumentRetriever {
+    /// creates an instance of [`DocumentRetriever`]
+    pub async fn create(input_uri: &Option<PathBuf>) -> anyhow::Result<Self> {
+        match input_uri {
+            Some(path_buf) => {
+                let file = File::open(path_buf).await?;
+                let reader = BufReader::new(file);
+                Ok(Self::File(reader.lines()))
+            }
+            None => {
+                let file = io::stdin();
+                let reader = BufReader::new(file);
+                Ok(Self::StdIn(reader.lines()))
             }
         }
-        None => {
-            let file = io::stdin();
-            let reader = BufReader::new(file);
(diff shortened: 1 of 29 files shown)


quickwit-core/src/indexing/index.rs
```diff
@@ -0,0 +1,94 @@
...
+        ),
+        finalize_split(split_receiver, statistic_sender.clone()),
+    )?;
+
+    statistic_collector.lock().await.display_report();
+    Ok(())
+}
+
+/// Clears the index by un-publishing all published splits.
+/// TODO: is this ok? should we hard delete using the storage?
```
evanxg852000, line 79: @fmassot @fulmicoton @guilload when `--overwrite` is specified do we want to soft delete or hard delete?

fmassot: If you delete the metastore splits without deleting files, then how will you cleanup your s3 files? I would not want to keep files. 
So first, for reset index command, I see 3 steps:
- mark all splits as `ScheduledForDeletion` (not sure if this state was implemented for that). From now, the search server will stop querying these splits
- delete s3 files and on success delete the metastore split, this would avoid some ghost files staying on s3.

The way you coded reset_index makes me think of another command like "unpublish splits" where you want to update the metastore splits to say that "ok now don't use that splits". Let's see that later because I don't see clearly use cases.

fmassot (commented):
3 remarks:
- delete_index command only delete index on the metastore but it's probably not finished, so put a TODO there
- how would you integrate aws region to be able to upload splits on s3? Unfortunately region is not in the uri
- I would create metastore and storage: first you see immediately config issues, then you can handle exotic client parameters at start and avoid polluting elsewhere, and more generally that's what you do when you have a server because clients are shared.

quickwit-core/src/indexing/index.rs
evanxg852000: @fmassot The reason for this was that I had a chat with @fulmicoton  at the start on the metastore that we probably did not want to delete the stuff immediately. My conclusion from that discussion was that we probably will have a scheduled task to delete these files. Currently, this just do the first step you mention I think. I believe unpublish means `ScheduledForDeletion`

quickwit-core/src/indexing/document_retriever.rs
```diff
@@ -0,0 +1,110 @@
...
+
+/// A trait for retrieving input content online at a time
+#[async_trait]
+pub trait DocumentRetriever {
+    /// Reads one line from underlying input buffer.
+    async fn next_document(&mut self) -> anyhow::Result<Option<String>>;
+}
+
+/// Retrieves new-line delimited json documents from StdIn or from a local File
+pub enum FileOrStdInDocumentRetriever {
```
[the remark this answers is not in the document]
evanxg852000: I think a rename possibly makes sence.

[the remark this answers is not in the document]
fulmicoton: Sounds good.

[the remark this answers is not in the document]
fulmicoton: Actually Lines<BufReader<Box<dyn AsyncIoRead>>> or something like that would work too...

The BufReader should mitigate the impact of the dynamic dispatch.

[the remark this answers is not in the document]
guilload: yes, if we're Ok with the dynamic dispatch, we could get rid of the `DocumentRetriever` altogether and write something like this directly in the `index` method:
```rust
    let async_read: Box<dyn AsyncRead + Unpin> = match index_path {
        Some(path) => Box::new(File::open(path).await?),
        _ => Box::new(tokio::io::stdin()),
    };

    let mut lines = BufReader::new(async_read).lines();

    while let Some(line) = lines.next_line().await? {
   }
   ```
   
 I guess this is ok for now.

File: quickwit-core/src/indexing/index.rs [hunk 3]
```diff
@@ -0,0 +1,123 @@
...
+    try_join!(
+        index_documents(
+            &params,
+            metastore,
+            storage_resolver,
+            document_retriever,
+            split_sender,
+            statistic_sender.clone()
+        ),
+        finalize_split(split_receiver, statistic_sender.clone()),
```
guilload, line 82: Since publishing is part of finalizing, splits will be published as they are produced. Ideally, we'd like to publish all the splits at once when indexing is done (see this [comment](https://github.com/quickwit-inc/quickwit/issues/3#issuecomment-843839740)).

We can address that in a follow-up PR.

evanxg852000: ok

PR #54, opened 2021-05-26
If the index is not loaded into the cache, load it from storage.

mosuka:
### Context / purpose
If the index is not loaded into the cache, load it from storage in a lazy manner.

### Description
https://github.com/quickwit-inc/quickwit/issues/50
The open_index function of the `Metastore` trait will be moved to the private function of the `SingleFileMetastore`.
Change functions that operate splits, such as `list_splits`, `stage_split`, `publish_split`, `mark_split_as_deleted` and `delete_split` to lazily load the index's metadata set into the cache if the index is not open.

### How was this PR tested?
Ran `cargo test -- --nocapture test_single_file_metastore`

### Checklist
- [x] tested locally
- [x] added unit tests
- [x] included documentation

### DIFF
--- a/quickwit-metastore/src/metastore/single_file_metastore.rs
+++ b/quickwit-metastore/src/metastore/single_file_metastore.rs
@@ -53,7 +53,7 @@ fn is_disjoint(left: &Range<u64>, right: &Range<u64>) -> bool {
 /// Single file meta store implementation.
 pub struct SingleFileMetastore {
     storage: Arc<dyn Storage>,
-    data: Arc<RwLock<HashMap<IndexUri, MetadataSet>>>,
+    cache: Arc<RwLock<HashMap<IndexUri, MetadataSet>>>,
 }
 
 #[allow(dead_code)]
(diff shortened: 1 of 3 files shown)


fulmicoton (changes requested):
This is no addressing the issue.

When we call list and the index is not loaded, we need to call open_index in a lazy manner.

We need to remove the calls to open_index (and the tests should still pass, as the private method open_index would be called instead of returning an error.)

The `IndexIsNotOpen` error should be removed too.

mosuka:
I update the PR. Please take another look.

quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -53,7 +53,7 @@ fn is_disjoint(left: &Range<u64>, right: &Range<u64>) -> bool {
 /// Single file meta store implementation.
 pub struct SingleFileMetastore {
     storage: Arc<dyn Storage>,
-    data: Arc<RwLock<HashMap<IndexUri, MetadataSet>>>,
+    cache: Arc<RwLock<HashMap<IndexUri, MetadataSet>>>,
```
fulmicoton, line 56: good idea.

fulmicoton (approved):

PR #60, opened 2021-05-27
Improve documentation of the MetaStore trait

mosuka:
### Context / purpose
Improve documentation of the MetaStore trait

### Description
Added a more detailed explanation of the Metastore trait's function.

https://github.com/quickwit-inc/quickwit/issues/58

### How was this PR tested?

### Checklist
- [X] included documentation


quickwit-metastore/src/metastore.rs
```diff
@@ -108,22 +108,35 @@ pub struct MetadataSet {
 #[async_trait]
 pub trait Metastore: Send + Sync + 'static {
     /// Creates an index.
+    /// This API creates index metadata set in the metastore.
+    /// If the index metadata set already exists, it will be loaded.
```
fulmicoton, line 112: This is not what happens if index metadata already exists.

fulmicoton (changes requested):
The first comment is crucial and wrong so I stopped reading. Please careful review your own code, and resend for review.

mosuka:
Sorry for the inconvenience.
Corrected the wrong document and the typos. Please take another look.

quickwit-metastore/src/metastore.rs
```diff
@@ -108,22 +108,37 @@ pub struct MetadataSet {
...
     /// Publishes a split.
+    /// This API only updates the state of the split from Staged to Published.
+    /// At this point, the split files are assumed to have already been uploaded.
+    /// If the split is already published, this API call returns a success.
+    /// An error will occur if you specify an index or split that does not exist in the storage.
     async fn publish_split(&self, index_id: &str, split_id: &str) -> MetastoreResult<()>;
 
     /// Lists the splits.
+    /// Returns a list of splits that intersect the given time_range and split_state.
+    /// Regardless of the time range filter, if a split has no timestamp it is always returned.
```
fulmicoton, line 140: The current implementation in the single file store is different.

We need to change the doc or change the code and the unit tests.

fulmicoton (changes requested):
Could you also write something in the MetaStore documentation itself about the life cycle of a split.

Explaining that before uploading any file, one needs to stage the split. 
Explain that it makes it possible to identify the split files that should be garbage collected.
etc.

quickwit-metastore/src/metastore.rs
mosuka: The documentation is wrong and will be corrected.

mosuka: I misread the specification, so I will fix the code and add some tests.

mosuka: Fixed a bug in the time range specification and added the unit tests.

mosuka:
Thanks for reviewing. I've added detailed documentation.

fulmicoton (approved):

PR #61, opened 2021-05-27
Fixed LocalFileStorage bug

evanxg852000:
### Context / purpose
[Issues-59](https://github.com/quickwit-inc/quickwit/issues/59)

### Description
Implemented fix for LocalFileStorage to transparently create & remove intermediate directories

### Checklist
- [x] Clarify the semantics in the Storage trait.
- [x] Fix the LocalFileStorage implementation
- [x] Add tests

### DIFF
--- a/quickwit-storage/src/lib.rs
+++ b/quickwit-storage/src/lib.rs
@@ -138,6 +138,24 @@ pub(crate) mod tests {
         Ok(())
     }
 
+    async fn test_write_and_delete_with_dir_separator(
+        storage: &mut dyn Storage,
+    ) -> anyhow::Result<()> {
+        let test_path = Path::new("foo/bar/write_and_delete_with_separator");
+        let payload_bytes = b"abcdefghijklmnopqrstuvwxyz".as_ref();
+        storage
+            .put(test_path, PutPayload::from(payload_bytes))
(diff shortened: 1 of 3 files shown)


quickwit-storage/src/lib.rs
```diff
@@ -138,6 +138,24 @@ pub(crate) mod tests {
...
     }
 
+    async fn test_write_and_delete_with_dir_separator(
+        storage: &mut dyn Storage,
+    ) -> anyhow::Result<()> {
+        let test_path = Path::new("foo/bar/write_and_delete_with_separator");
+        let payload_bytes = b"abcdefghijklmnopqrstuvwxyz".as_ref();
+        storage
+            .put(test_path, PutPayload::from(payload_bytes))
+            .await?;
```
fulmicoton, line 148: Can you add a test here as well?
```
assert!(matches!(            storage.exists(Path::new("foo/bar")).await,            Ok(false)        ));
```

Should return false even before deleting.

fmassot:
I always have a bad feeling when we are making some path join root + path... we will look at that here https://github.com/quickwit-inc/quickwit/issues/53
After adding the test from @fulmicoton I think we can merge.

PR #69, opened 2021-05-28
Index wikipedia [NOT FOR MERGE]

evanxg852000:
### Context / purpose
This is a test drive branch to try out the indexing command against the Wikipedia English corpus. 
We should revisit this and try out to see how are we doing for indexing a real dataset.

### How to run it.
`cargo run --release  -- index --index-uri file://./data --input-path ./wiki-articles.json`
`cargo run  -- index --index-uri s3://quickwit-dev/evance/test --input-path ./wiki-articles.json` [subject to implementing](https://github.com/quickwit-inc/quickwit/issues/70)

### How we currently do on my box?
We use 2 threads for indexing with 1GB allocated to each.
Spec:
Lenovo Thinkpad:
Intel(R) Xeon(R) W-10855M CPU @ 2.80GHz
RAM: 32GB

Statistics
Num documents: 5032105
Num parse errors: 0
Num splits: 3
Total size: 8886 MB
Index size: 6386 MB
Indexing throughput: 35.98MB/s
Elapsed time: 4.12min

### Checklist
- [ ] Still need to harmonize the storage and run this against S3

### DIFF
--- a/quickwit-core/src/indexing/document_indexer.rs
+++ b/quickwit-core/src/indexing/document_indexer.rs
@@ -22,6 +22,8 @@
 
 use quickwit_metastore::Metastore;
 use quickwit_storage::StorageUriResolver;
+use tantivy::schema::TextFieldIndexing;
+use tantivy::schema::TextOptions;
 use std::sync::Arc;
 use tantivy::{schema::Schema, Document};
 use tokio::sync::mpsc::Sender;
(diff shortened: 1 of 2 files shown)


fmassot:
@evanxg852000 I prefer to do rebase from main when possible instead of using merge

evanxg852000:
I will do a rebase next time. But they all achieve the same thing. Just that a rebase without conflict does not creates a commit. but if it has conflict you end up with the same merge commit like here. By the way the CI failed because of s3 localstack so I just re-run it.

fmassot:
we can close now that we can use wikipedia doc mapper properly

PR #74, opened 2021-05-31
Fixes & statistics on Indexing

evanxg852000:
### Context / purpose

### Description
- Added changes to display statistics
- Added temporary changes for `MetastoreUriResolver` can resolve multiple protocols
- changed split publishing to happen at the end but still need to be atomic [see](https://github.com/quickwit-inc/quickwit/issues/71)
- added command report and how a user can search his index

### How was this PR tested?
I Ran the command `quickwit-cli index --index-uri 's3://quickwit-dev/evance/van2' --input-path ./wiki-articles-10000.json`

### Checklist
- [x] tested locally

### DIFF
--- a/quickwit-core/Cargo.toml
+++ b/quickwit-core/Cargo.toml
@@ -10,7 +10,6 @@ anyhow = "1.0"
 async-trait = "0.1"
 quickwit-doc-mapping = { version = "0.1.0", path = "../quickwit-doc-mapping" }
 quickwit-metastore = { version = "0.1.0", path = "../quickwit-metastore" }
-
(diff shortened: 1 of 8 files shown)


fmassot:
great @evanxg852000 
would be better if we can split this PR in several parts: I don't want to merge the wikipedia index schema into the main branch for example.
At least I see 3 different PR:
- MetastoreUriResolver to resolve multiple protocols
- changed split publishing to happen at the end but still need to be atomic see
 - other like changes to display statistics

evanxg852000:
Ok

Issue #28, opened 2021-05-17
Specify and implement `DocMapping`


guilload:
# Why we need a doc mapping
## What is a "doc mapping"?

Let's quote [Elasticsearch doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html): 

> Mapping is the process of defining how a document, and the fields it contains, are stored and indexed.

## Tantivy provides only a schema

As mentioned [here](https://github.com/tantivy-search/tantivy/issues/464), tantivy provides already a schema which defines the set of fields for a given index and how it is stored, the user has yet to define how he will convert input document (in json typically) to data that will be in the index. 

## Do we need more that a schema?

Tantivy already provides some useful information in the schema about how data needs to be stored (see indexing options). But we may need something more to handle other cases:
- define what metadata we may store: typically we want to be able to store or not the entire document in a _source field like ES does it
- more complex fields needs a mapping: for example, multifields (see tantivy [issue](https://github.com/tantivy-search/tantivy/issues/464)), one day we will have geo fields :)
- handle [coerce](https://www.elastic.co/guide/en/elasticsearch/reference/current/coerce.html#coerce)  
- define a facet for a given field
- define how to control dynamic field mapping (to avoid mapping explosion for example)

# Where to implement it?

I would do it in tantivy but we can start a "dumb" implementation for quickwit's need.

# Specification

The mapping is a json (or yaml?) file and sufficient to create the index schema.
To start with, we can just use tantivy schema as it will do the job.

## Format

```json
{
  "store_source": true,
  "ignore_unknown_fields": true,
  "default_search_fields": [
    "body", "title"
  ],
  "properties": [
    {
      "source_name": "timestamp",
      "target_name": "timestamp2",
      "type": "u64",
      "indexed": true,
      "stored": true
    },
    {
      "source_name": "severity_text",
      "type": "text",
      "record": "basic",
      "tokenizer": "raw",
      "stored": true
    },
    {
      "source_name": "resources.label",
      "type": "my_custom_text"
    },
    {
      "source_name": "meta",
      "type": "array<organization>"
    },
    {
      "source_name": ""
    }
  ],
  "field_types": [
    {
      "name": "my_custom_text",
      "type": "text",
      "analyzer": "my_analyzer"
    },
    {
      "name": "organization",
      "type": "object",
      "properties": [
        {
          "source_name": "name",
          "type": "text"
        },
        {
          "source_name": "created_at",
          "type": "u64"
        }
[...]

PR #75, opened 2021-05-31
Add doc mapper basic implementations

fmassot:
This PR aims at solving issue #28
It's definitely not the implementation that we target but there is the trait and basic implementations that we can use to index wikipedia.
This will be followed by another PR for the full implementations of default doc mapper and all flatten doc mapper.


evanxg852000 (commented):
Looks good to me so far.

fulmicoton (approved):

PR #77, opened 2021-05-31
Displaying statistics for Index command

evanxg852000:
### Context / purpose

### Description
- Added changes to display statistics
- Added command report and how a user can search his index

### How was this PR tested?
I Ran the command quickwit-cli index --index-uri 's3://quickwit-dev/evance/van2' --input-path ./wiki-articles-10000.json

### Checklist
- [x] tested locally


quickwit-core/src/indexing/statistics.rs
```diff
@@ -151,14 +159,47 @@ impl StatisticsCollector {
...
+        if elapsed_secs >= 60 {
+            println!(
+                "Indexded {} documents in {:.2$}min",
+                self.num_docs,
+                elapsed_secs.max(1) as f64 / 60f64,
+                2
+            );
+        } else {
+            println!(
+                "Indexded {} documents in {}s",
```
fulmicoton, line 181: typo

quickwit-core/src/indexing/statistics.rs
```diff
@@ -151,14 +159,47 @@ impl StatisticsCollector {
...
 
     /// Display a one-shot report.
     pub fn display_report(&self) {
-        //TODO: better display stats
-        println!("Statistics: {:?}", self);
+        let elapsed_secs = self.start_time.elapsed().as_secs();
+        println!();
+        if elapsed_secs >= 60 {
+            println!(
+                "Indexded {} documents in {:.2$}min",
```
fulmicoton, line 174: typo

quickwit-core/src/indexing/statistics.rs
```diff
@@ -102,8 +112,6 @@ impl StatisticsCollector {
         let (event_sender, mut event_receiver) = channel(1000);
         task::spawn(async move {
             while let Some(event) = event_receiver.recv().await {
-                //TODO: discuss the grannular level of changes we want to display
-                // live in the terminal
                 let mut statistics = moved_statistics_collector.lock().await;
```
fulmicoton, line 115: there are only counters in there.
Should we remove use atomics and remove the mutex?

Or even better...
Remove the channel and use methods to increment instead of emitting events?

quickwit-core/src/indexing/statistics.rs
```diff
@@ -91,6 +99,8 @@ impl StatisticsCollector {
             num_published_splits: 0,
             total_bytes_processed: 0,
             total_size_splits: 0,
+            start_time: Instant::now(),
+            stdout: stdout(),
```
fulmicoton, line 103: I'd rather have, an Arc<StatisticsCollector> that contains the model of the data.

It contains start_time, and all of the counters as Atomics. We can use super relaxed ordering (I guess... let's just make we don't divide anything by 0 as a result!?).

We could then have a task dedicating to display these stats. That task would be spawned from the cli function.

evanxg852000: I thought of atomic counters but i was thinking keeping this way is simpler. I only used channel and events here because of the requirement of displaying a kind of live progress on statistics. It was a way to keep all this in on place.

evanxg852000:
@fulmicoton  I have updated the branch. 
I recall you mentioned we could use a task at the top level for cli commands to display info in the console. But I just stuck with the task at the indexing. you can best design that abstraction if we need it.
I used `tokio::sync:watch` to listen to task completion since I am using a `loop` for displaying the statistic. It required a king of single-producer-multi-consumer channel.

quickwit-core/src/indexing/document_indexer.rs
```diff
@@ -28,9 +28,9 @@ use tokio::sync::mpsc::Sender;
 
 use crate::indexing::document_retriever::DocumentRetriever;
 use crate::indexing::split::Split;
-use crate::indexing::statistics::StatisticEvent;
 
 use super::IndexDataParams;
+use super::INDEXING_STATISTICS;
```
fulmicoton, line 33: Can we avoid the global here? It does not seem necessary and it prevents index_documents from being usable in a different context.

quickwit-core/src/indexing/index.rs
```diff
@@ -89,12 +90,12 @@ pub async fn index_data(
...
             split_sender,
-            statistic_sender.clone(),
         ),
-        finalize_split(split_receiver, statistic_sender.clone()),
+        finalize_split(split_receiver, task_completed_sender),
+        start_statistics_reporting(task_completed_receiver),
     )?;
 
-    statistic_collector.lock().await.display_report();
+    println!("You can now query your index with `quickwit search --index-path {} --query \"barack obama\"`" , params.index_uri.display());
```
fulmicoton, line 98: can we move that in the call in the `index_data_cli` function?

quickwit-core/src/indexing/index.rs
```diff
@@ -79,8 +80,8 @@ pub async fn index_data(
     }
 
     let document_retriever = Box::new(DocumentSource::create(&params.input_uri).await?);
-    let (statistic_collector, statistic_sender) = StatisticsCollector::start_collection();
     let (split_sender, split_receiver) = channel::<Split>(SPLIT_CHANNEL_SIZE);
+    let (task_completed_sender, task_completed_receiver) = watch::channel::<bool>(false);
```
fulmicoton, line 84: can we move that in the call in the index_data_cli function?

fulmicoton: The rationale is that this function could then be used by a different client that the quickwit-cli

quickwit-core/src/indexing/index.rs
```diff
@@ -89,12 +90,12 @@ pub async fn index_data(
             storage_resolver,
             document_retriever,
             split_sender,
-            statistic_sender.clone(),
         ),
-        finalize_split(split_receiver, statistic_sender.clone()),
+        finalize_split(split_receiver, task_completed_sender),
+        start_statistics_reporting(task_completed_receiver),
```
fulmicoton, line 95: Can we do the reporting in the cli only?

fulmicoton (changes requested):
- remove the global. replace it by an Arc<IndexingStatistics>
- put the reporting task in the CLI.
- remove the watch channel from the parameter from core. You can either use a watch channel in the CLI side to keep the API in core clean.

evanxg852000:
@fulmicoton from your comments, this is what we will end up with in `index_data_cli`?
``` rust
   let (task_completed_sender, task_completed_receiver) = watch::channel::<bool>(false);
    let indexing_future = task::spawn(async move {
        index_data(metastore_uri, index_id, params).await?;
        task_completed_sender.send(true)?;
    });
    let reporting_future = start_statistics_reporting(task_completed_receiver);

    try_join!(indexing_future, reporting_future)?;
```
Regarding the `Arc<Statistics>`,  should we create it here and pass it down to the `index_data` ? 
Also, are you planning of making it fully optional `Option<Arc<Statistics>>` ?

fulmicoton:
@evanxg852000 Exactly!

fulmicoton:
> Also, are you planning of making it fully optional Option<Arc<Statistics>>

Let's force the user to submit one.
 In the future, we can try to make the interface a tad user friendly... Probably by introducing a listener interface.

fulmicoton (approved):

PR #78, opened 2021-05-31
Support multiple protocol on MetastoreUriResolver

evanxg852000:
### Context / purpose

### Description
Added temporary changes to MetastoreUriResolver so we can resolve multiple protocols

### DIFF
--- a/quickwit-metastore/src/metastore/single_file_metastore.rs
+++ b/quickwit-metastore/src/metastore/single_file_metastore.rs
@@ -377,8 +377,8 @@ impl Default for SingleFileMetastoreFactory {
 
 #[async_trait]
 impl MetastoreFactory for SingleFileMetastoreFactory {
-    fn protocols(&self) -> Vec<String> {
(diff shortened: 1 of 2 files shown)


fmassot:
LGFM. I think such a fix is ok right now and we can see later if this is sufficient to handle other backends.

fulmicoton (changes requested):
I am not fond of changing the Metastore interface for this.

What about the following solution :\
- remove the SingleFileMetastoreFactory
- use SingleFileMetaStore as the fallback in MetastoreUriResolver.

In other words, fn resolve would look something like...
```
if let Some(resolver) = self
            .per_protocol_resolver
            .get(protocol) {
  return resolver.resolve(...)
}
let storage = self.storage_uri_resolver.resolve(uri)?;
let single_file_metastore = SingleFileMetastore::new(storage);
Ok(single_file_metastore)
```

fmassot:
excellent idea

PR #86, opened 2021-06-01
Remove DocMapperType using TypeTag

evanxg852000:
### Context / purpose
The DocMapperType was only used for serialization. It needs to be replaced with DocMapper.
[See Issue 85](https://github.com/quickwit-inc/quickwit/issues/85)

### Description
Added changes to use typetag crate for serializing the DocMapper Trait

### How was this PR tested?
- Run the create index command
- Verify the created index metadata jso file.

### Checklist
- [x] tested locally
- [x] updated unit tests
- [x] included documentation


quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -424,7 +424,7 @@ mod tests {
             let index_metadata = IndexMetadata {
                 index_id: index_id.to_string(),
                 index_uri: "ram://indexes/my-index".to_string(),
-                doc_mapper_type: DocMapperType::AllFlatten,
+                doc_mapper: build_doc_mapper("all_flatten", None).unwrap(),
```
fulmicoton, line 427: why not `Box::new(AllFlatten::default())`

quickwit-doc-mapping/src/mapper.rs
```diff
@@ -43,95 +44,88 @@ use tantivy::{
...
-    }
-}
+// TODO: this is a placeholder, to be removed when it will be implementend in the search-api crate
+pub struct SearchRequest {}
 
 /// Build a doc mapper given the doc mapper type.
-pub fn build_doc_mapper(mapper_type: DocMapperType) -> anyhow::Result<Box<dyn DocMapper>> {
-    match mapper_type {
-        DocMapperType::Default(config) => {
+pub fn build_doc_mapper(
```
fulmicoton, line 63: I don't think we need this function at all, do we?

evanxg852000: I think @fmassot  is in a better position to answer this. I believe there is going to be validation on the config.

quickwit-metastore/src/metastore/single_file_metastore.rs
evanxg852000: So my thinking was that since `build_doc_mapper` is exported then it should be the way to construct Doc_Mapper from outside. instead of exposing the concrete types. The mapper modules were declared as private as `mod all_flatten_mapper;`

quickwit-cli/src/main.rs
```diff
@@ -34,7 +33,7 @@ use quickwit_core::indexing::{index_data, IndexDataParams};
 
 struct CreateIndexArgs {
     index_uri: String,
-    doc_mapper_type: DocMapperType,
+    doc_mapper_type: String,
```
fulmicoton, line 36: I think that could be Box<dyn DocMapper>

quickwit-metastore/src/metastore/single_file_metastore.rs
fulmicoton: We can make the concrete DocMapper struct public. The module can stay private, but you can do a public reexport.

```
mod all_flatten_mapper;

pub use self::all_flatten_mapper::AllFlattenMapper;
```

quickwit-cli/src/main.rs
evanxg852000: Not sure I understand?

quickwit-doc-mapping/src/mapper.rs
fmassot: let's remove it as currently we only need something to build the doc mapper from the type in the cli. So we can put that only in the cli.

quickwit-doc-mapping/src/mapper.rs
```diff
@@ -43,95 +44,88 @@ use tantivy::{
 /// - a way to build a tantivy::Query from a SearchRequest
 /// - a way to build a tantivy:Schema
 ///
-pub trait DocMapper: Send + Sync + 'static {
+#[typetag::serde(tag = "type", content = "attributes")]
```
fmassot, line 47: attributes is useless.

quickwit-cli/src/main.rs
fulmicoton: It could have been Box<dyn DocMapper> or Arc<dyn DocMapper> instead of string.

quickwit-doc-mapping/src/mapper.rs
fulmicoton: Why do we need that in the CLI?

quickwit-cli/src/main.rs
fmassot: To build the doc mapper we would need the json file though.

evanxg852000: So I still beleive  this might not be the right approach. this is the usage of create_index from a client perspective.
```rust
let doc_mapper = build_doc_mapper(&args.doc_mapper_type, mapper_config)?;
    let index_metadata = IndexMetadata {
        index_id: index_id.to_string(),
        index_uri: args.index_uri.to_string(),
        doc_mapper,
    };
    create_index(metastore_uri, index_metadata).await?;
```
as you can see what you are suggesting, we put it `IndexMetadata`. so a client to use the `create_index` from core needs to construct metastore_uri & index_metadata. I think we should not move the construction of the mapper in the cli parsing.

evanxg852000: @fulmicoton  ignore this I have a way forward now with @fmassot .

fmassot:
sounds really good!

PR #88, opened 2021-06-02
Atomically publish splits

evanxg852000:
### Context / purpose
We need a way in the Metastore to have an all-or-nothing behavior during split publishing. 
[See Issue 71](https://github.com/quickwit-inc/quickwit/issues/71)

### Description
- Changed metastore publish split interface to accepts multiple split at once.
- Updated indexing command to use this new interface 

### How was this PR tested?
Run the create & index command on a local wikipedia corpus.
`cargo run -- new --index-uri file://./data --no-timestamp-field --doc-mapper-type wikipedia`
`cargo run -- index --index-uri file://./data --input-path ./wiki-articles.json`

### Checklist
- [x] tested locally
- [x] added unit tests

### DIFF
--- a/quickwit-core/src/indexing/split.rs
+++ b/quickwit-core/src/indexing/split.rs
@@ -287,10 +287,6 @@ mod tests {
                 Ok(())
             },
         );
-        mock_metastore
-            .expect_publish_splits()
-            .times(0)
-            .returning(|_uri, _id| Ok(()));
 
         let metastore = Arc::new(mock_metastore);
         let split_result = Split::create(
--- a/quickwit-core/src/indexing/split.rs
+++ b/quickwit-core/src/indexing/split.rs
@@ -177,13 +177,13 @@ impl Split {
         Ok(manifest)
     }
 
-    /// Publish the split in the metastore.
-    pub async fn publish(&self) -> anyhow::Result<()> {
-        self.metastore
-            .publish_split(&self.index_uri, &self.id.to_string())
-            .await?;
-        Ok(())
-    }
+    // /// Publish the split in the metastore.
+    // pub async fn publish(&self) -> anyhow::Result<()> {
+    //     self.metastore
+    //         .publish_split(&self.index_uri, &self.id.to_string())
+    //         .await?;
+    //     Ok(())
+    // }
 }
 
 async fn put_to_storage(storage: &dyn Storage, split: &Split) -> anyhow::Result<Manifest> {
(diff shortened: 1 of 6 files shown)


quickwit-core/src/indexing/split.rs
```diff
@@ -296,8 +288,8 @@ mod tests {
             },
         );
         mock_metastore
```
fulmicoton, line 290: you can remove these 4 lines entirely if you expect it 0 times. If it is called the unit test will panic.

quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -239,31 +239,44 @@ impl Metastore for SingleFileMetastore {
...
+                }
+                _ => {
+                    return Err(MetastoreErrorKind::SplitIsNotStaged
+                        .with_error(anyhow::anyhow!("Split ID is not staged: {:?}", split_id)));
+                }
             }
         }
 
+        // Update the splits state to `Published`.
+        updatable_split_ids.into_iter().for_each(|split_id| {
```
fulmicoton, line 274: a for-loop is actually more readable in that case.

```
for split_id in updatable_split_ids {
   // ...
}
```

quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -239,31 +239,44 @@ impl Metastore for SingleFileMetastore {
...
 
-        // Check for the existence of split.
-        let split_metadata = metadata_set.splits.get_mut(split_id).ok_or_else(|| {
-            MetastoreErrorKind::SplitDoesNotExist
-                .with_error(anyhow::anyhow!("Split does not exist: {:?}", split_id))
-        })?;
+        let mut updatable_split_ids = vec![];
+        for split_id in split_ids {
+            // Check for the existence of split.
+            let split_metadata = metadata_set.splits.get(split_id).ok_or_else(|| {
```
fulmicoton, line 252: I love that you that you made sure that not mutate the split_metadatas if we are to emit an error.

That being said, the metadata_set you are editing right now is a copy of the actual metadata_set.
We "put" it at the end of the function, so you could have mutated it within the match case if you wanted.

fulmicoton (changes requested):
Please at least remove the useless code in the unit test (times(0)) and consider simplifying the code in publish_splits.

quickwit-metastore/src/metastore/single_file_metastore.rs
evanxg852000: I knew it was a copy from `get_index` but I was being paranoid here.

PR #94, opened 2021-06-07
Remove dangling split folder

evanxg852000:
### Context / purpose
A useless split folder is created in the indexing target folder after the index command.
[See Issue 92](https://github.com/quickwit-inc/quickwit/issues/92)

### Description
- Removed the temperary fix we previously had in `SingleFileStorageFactory::from_uri()` 
- simplified last split sending code

### How was this PR tested?
Changed the number of docs per split and indexed wikipedia.json on file:// and s3 

### Checklist
- [x] tested locally

### DIFF
--- a/quickwit-cli/src/main.rs
+++ b/quickwit-cli/src/main.rs
@@ -112,7 +112,7 @@ impl CliCommand {
             .map(|field| field.to_string());
         let overwrite = matches.is_present("overwrite");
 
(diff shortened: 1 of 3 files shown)


fmassot:
Can you check that when you run create / index / delete, there's nothing no dangling folder ? It seems that deleting does not work

fmassot:
ok don't take into account my comment, I will merge this PR right now and you will open a new PR with all the small fixes concerning the CLI.

PR #101, opened 2021-06-08
Avoid merging segments if there is only one #99

### LINKED ISSUE #99: Avoid merging if there is only one segment

fulmicoton:
Closes #99

### DIFF
--- a/quickwit-core/src/indexing/split.rs
+++ b/quickwit-core/src/indexing/split.rs
@@ -32,6 +32,8 @@ use std::sync::Arc;
 use std::time::Instant;
 use std::{path::PathBuf, usize};
 use tantivy::Directory;
+use tantivy::SegmentId;
+use tantivy::SegmentMeta;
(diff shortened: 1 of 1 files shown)


quickwit-core/src/indexing/split.rs
```diff
@@ -153,14 +168,23 @@ impl Split {
     }
 
     /// Merge all segments of the split into one.
-    pub async fn merge_all_segments(&mut self) -> anyhow::Result<tantivy::SegmentMeta> {
-        let segment_ids = self.index.searchable_segment_ids()?;
+    ///
+    /// If there is only one segment and it has no delete, avoid doing anything them.
+    pub async fn merge_all_segments(&mut self) -> anyhow::Result<()> {
```
evanxg852000, line 173: Absolutely makes sense to not return the `SegmentMeta` since we are not currently using it. :+1:

evanxg852000 (approved):

PR #102, opened 2021-06-08
Delete Index Command

evanxg852000:
### Context / purpose
Need to implement the delete command. [See Issue 6](https://github.com/quickwit-inc/quickwit/issues/6)

### Description
- Implemented the delete index command to support `dry_run` flag
- Changed the `metastore::mark_split_as_deleted` to `metastore::mark_splits_as_deleted` to emulate atomic operation.
- Changed the `metastore::delete_split` to `metastore::delete_splits` o emulate atomic operation.
- Make use of the delete feature in `reset_index` helper function

### How was this PR tested?
```bash
cargo run -- new --index-uri s3://quickwit-dev/evance/data --no-timestamp-field --doc-mapper-type wikipedia  
cargo run -- index --index-uri s3://quickwit-dev/evance/data --input-path wiki-articles-10000.json
cargo run -- index --index-uri s3://quickwit-dev/evance/data --input-path wiki-articles-10000.json  --overwrite 
cargo run -- delete --index-uri s3://quickwit-dev/evance/data --dry-run
cargo run -- delete --index-uri s3://quickwit-dev/evance/data 

cargo run -- new --index-uri file://quickwit-dev/evance/data --no-timestamp-field --doc-mapper-type wikipedia  
cargo run -- index --index-uri file://quickwit-dev/evance/data --input-path wiki-articles-10000.json 
cargo run -- index --index-uri file://quickwit-dev/evance/data --input-path wiki-articles-10000.json  --overwrite
cargo run -- delete --index-uri file://quickwit-dev/evance/data --dry-run
cargo run -- delete --index-uri file://quickwit-dev/evance/data
```

### Checklist
- [x] tested locally
- [x] added unit tests


quickwit-core/src/index.rs
```diff
@@ -65,28 +69,114 @@ pub async fn search_index(metastore_uri: &str, index_id: &str) -> anyhow::Result
...
+    let split_ids = active_splits
+        .iter()
+        .map(|split_meta| split_meta.split_id.as_str())
+        .collect::<Vec<_>>();
+    metastore
+        .mark_splits_as_deleted(index_id, split_ids)
+        .await?;
+
+    let files = garbage_collect(metastore_uri, index_id, storage_resolver).await?;
     metastore.delete_index(index_id).await?;
```
fulmicoton, line 106: We still have a race condition here. If someone adds a new split right before we delete the index, we will end up with a orphan split in the storage.

fulmicoton: Can you add a TODO here to fix that later?
It is not a major bug in the sense, user are a bit at fault here.

fulmicoton (approved):
LGTM can  you just add a comment about the race condition?

quickwit-core/src/index.rs
evanxg852000: Do you mean when we a user runs the `index` command on another terminal while the `delete` was being performed?

fmassot: yes exactly

fmassot: or another user

PR #109, opened 2021-06-09
Restrictions on Index Id

evanxg852000:
### Context / purpose
We need to define and enforce the `index_id` format. [See Issue 56](https://github.com/quickwit-inc/quickwit/issues/56)

### Description
Added regex validation when the index_id is extracted

### Checklist
- [x] tested locally
- [x] updated test


quickwit-cli/src/main.rs
```diff
@@ -227,6 +227,11 @@ fn extract_metastore_uri_and_index_id_from_index_uri(
 ) -> anyhow::Result<(&str, &str)> {
     let parts: Vec<&str> = index_uri.rsplitn(2, '/').collect();
     if parts.len() == 2 {
+        let index_id_pattern = regex::Regex::new(r"^[a-zA-Z][a-zA-Z0-9_\-]*$").unwrap();
```
fulmicoton, line 230: can we avoid compiling the regex at every call? (see once_cell)

fulmicoton, line 230: can we avoid compiling the regex at every call? (see once_cell)

quickwit-cli/src/main.rs
```diff
@@ -227,6 +227,11 @@ fn extract_metastore_uri_and_index_id_from_index_uri(
...
+        if !index_id_pattern.is_match(parts[0]) {
+            let error_message = format!("Invalid index_id `{}`. Only alpha-numeric, `-` and `_` characters allowed. Cannot start with `-`, `_` or digit.", parts[0]);
+            anyhow::bail!(error_message);
```
fulmicoton, line 233: `bail!` does the formatting for you.

fulmicoton (changes requested):

fulmicoton (approved):

quickwit-cli/src/main.rs
```diff
@@ -708,11 +715,19 @@ mod tests {
...
     #[test]
     fn test_extract_metastore_uri_and_index_id_from_index_uri() -> anyhow::Result<()> {
-        let index_uri = "file:///indexes/wikipedia";
+        let index_uri = "file:///indexes/wikipedia-xd_1";
         let (metastore_uri, index_id) =
             extract_metastore_uri_and_index_id_from_index_uri(index_uri)?;
         assert_eq!("file:///indexes", metastore_uri);
-        assert_eq!("wikipedia", index_id);
+        assert_eq!("wikipedia-xd_1", index_id);
+        extract_metastore_uri_and_index_id_from_index_uri("file:///indexes/_wikipedia")
```
fulmicoton, line 723: nitpick: assert!(result.is_err()); instead of unwrap_err

quickwit-cli/src/main.rs
```diff
@@ -226,8 +228,13 @@ impl CliCommand {
 fn extract_metastore_uri_and_index_id_from_index_uri(
     index_uri: &str,
 ) -> anyhow::Result<(&str, &str)> {
+    static INDEX_ID_PATTERN: Lazy<Regex> =
+        Lazy::new(|| Regex::new(r"^[a-zA-Z][a-zA-Z0-9_\-]*$").unwrap());
     let parts: Vec<&str> = index_uri.rsplitn(2, '/').collect();
     if parts.len() == 2 {
```
fulmicoton, line 234: nitpick: we prefer early return at quickwit (discussed a long time ago in the chat.)
https://medium.com/swlh/return-early-pattern-3d18a41bba8#:~:text=Return%20early%20is%20the%20way,when%20conditions%20are%20not%20met.

PR #122, opened 2021-06-14
Handling docs with timestamp 

evanxg852000:
### Context / purpose
We need to handle documents with timestamp during indexing.
[See Issue 108](https://github.com/quickwit-inc/quickwit/issues/108)

### Description
- Implemented time range calculation during indexing when `timestamp-field` is specified.
- Make sure during indexing we use schema that is provided by `DocMapper::schema` interface.

### How was this PR tested?
- Run the indexing command 

### Checklist
- [x] tested locally
- [x] added unit tests

### DIFF
--- a/quickwit-core/src/indexing/split.rs
+++ b/quickwit-core/src/indexing/split.rs
@@ -141,13 +141,13 @@ impl Split {
         let mut computed_time_range = None;
         if let Some(timestamp_field) = self.timestamp_field {
             let mut split_time_range = self.metadata.time_range.clone().unwrap_or(Range {
-                start: u64::MAX,
-                end: u64::MIN,
+                start: i64::MAX,
+                end: i64::MIN,
             });
 
             if let Some(timestamp) = doc
                 .get_first(timestamp_field)
-                .and_then(|field_value| field_value.u64_value())
+                .and_then(|field_value| field_value.i64_value())
             {
                 if timestamp < split_time_range.start {
                     split_time_range.start = timestamp;
@@ -461,7 +461,7 @@ mod tests {
 
         let mut schema_builder = Schema::builder();
         schema_builder.add_text_field("name", STRING);
-        let timestamp = schema_builder.add_u64_field("timestamp", FAST);
+        let timestamp = schema_builder.add_i64_field("timestamp", FAST);
         let schema = schema_builder.build();
 
         let mut mock_metastore = MockMetastore::default();
--- a/quickwit-core/src/indexing/split.rs
+++ b/quickwit-core/src/indexing/split.rs
@@ -116,8 +116,13 @@ impl Split {
             index.writer_with_num_threads(params.num_threads, params.heap_size as usize)?;
         index_writer.set_merge_policy(Box::new(NoMergePolicy));
         let index_uri = metastore.index_metadata(&params.index_id).await?.index_uri;
-        let metadata = SplitMetadata::new(id.to_string());
-
+        let mut metadata = SplitMetadata::new(id.to_string());
+        if timestamp_field.is_some() {
+            metadata.time_range = Some(Range {
+                start: i64::MAX,
+                end: i64::MIN,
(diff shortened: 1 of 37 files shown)


quickwit-core/src/indexing/split.rs
```diff
@@ -127,17 +132,39 @@ impl Split {
...
+            timestamp_field,
         })
     }
 
     /// Add document to the index split.
     pub fn add_document(&mut self, doc: Document) -> anyhow::Result<()> {
-        //TODO: handle time range when docMapper is available
+        let mut computed_time_range = None;
+        if let Some(timestamp_field) = self.timestamp_field {
+            let mut split_time_range = self.metadata.time_range.clone().unwrap_or(Range {
```
fmassot, line 143: I would avoid the clone here

quickwit-core/src/indexing/split.rs
```diff
@@ -127,17 +132,39 @@ impl Split {
...
+                start: u64::MAX,
+                end: u64::MIN,
+            });
+
+            if let Some(timestamp) = doc
```
fmassot, line 148: sounds like we are doing extra stuff here just because we have a tantivy document that is not aware of the timestamp field.
Moreover here we need to know which type we have and... actually it's not a u64. Can you change that ?

Can you also add a todo here so that we find another way to do that later ? I have some idea but not sure if they are good. we need to see that with @fulmicoton

fmassot: basically I would create a struct QuickwitDocument that will be aware of timestamp stuff. This may be overkill though...

evanxg852000: I was already changing the u64 to i64 because of the merge. 
As to regard to `time_range.clone()`, my reasoning is that since we are using Range<i64> that is cheap to clone.

quickwit-core/src/indexing/split.rs
evanxg852000: Ok, My reasoning is that since we are using Range<i64> that is cheap to clone.

quickwit-core/src/indexing/split.rs
fmassot: cool :)

quickwit-core/src/indexing/split.rs
```diff
@@ -127,17 +131,43 @@ impl Split {
...
+                }
+            }
+            computed_time_range = Some(split_time_range);
+        }
+
         self.index_writer
             .as_ref()
             .ok_or_else(|| anyhow::anyhow!("Missing index writer."))?
             .add_document(doc);
         self.metadata.num_records += 1;
```
fmassot, line 169: I would only do the assignment in the if condition. 
Here we are saying that if one document has no timestamp information, we delete the information for the split

evanxg852000: This is fine now and as discussed, I now have a function to perform the time_range computation.

fulmicoton (changes requested):
See comments

quickwit-core/src/indexing/split.rs
```diff
@@ -127,20 +132,53 @@ impl Split {
...
+        self.metadata.time_range = computed_time_range;
         Ok(())
     }
 
+    /// Compute the split new time range based on incomming document.
+    fn compute_time_range(&self, doc: &Document) -> Option<Range<i64>> {
```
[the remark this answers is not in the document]
evanxg852000: So computing before has the advantage of not cloning the doc on line 145.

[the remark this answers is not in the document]
fmassot: cloning the doc seems expensive for just updating some metadata, I think we can find another way to do that.

quickwit-core/src/indexing/split.rs
fmassot: yes no over engineering

PR #125, opened 2021-06-15
Time Range filter in QuicwitCollector

evanxg852000:
### Context / purpose
Need to prune docs that don't fall in time range before they get to Collector. 
[See Issue](https://github.com/quickwit-inc/quickwit/issues/123)

**Note: Let's hold this till (https://github.com/tantivy-search/tantivy/issues/1084) in merged into our internal tantivy as to avoid cascading todos.*

### DIFF
--- a/quickwit-search/src/collector.rs
+++ b/quickwit-search/src/collector.rs
@@ -31,6 +31,7 @@ use tantivy::collector::Collector;
 use tantivy::collector::SegmentCollector;
 use tantivy::fastfield::DynamicFastFieldReader;
 use tantivy::fastfield::FastFieldReader;
+use tantivy::schema::Field;
 use tantivy::DocId;
 use tantivy::Score;
 use tantivy::SegmentOrdinal;
@@ -212,6 +213,9 @@ pub struct QuickwitCollector {
     pub start_offset: usize,
     pub max_hits: usize,
     pub sort_by: SortBy,
+    pub timestamp_field: Option<Field>,
+    pub start_timestamp: Option<i64>,
+    pub end_timestamp: Option<i64>,
 }
 
 impl QuickwitCollector {
@@ -313,6 +317,10 @@ pub fn make_collector(
         start_offset: search_request.start_offset as usize,
         max_hits: search_request.max_hits as usize,
         sort_by: doc_mapper.default_sort_by(),
+        //handle error when PR#122 is merged
+        timestamp_field: doc_mapper.timestamp_field(),
+        start_timestamp: search_request.start_timestamp,
+        end_timestamp: search_request.end_timestamp,
     }
 }
 
--- a/quickwit-search/src/collector.rs
+++ b/quickwit-search/src/collector.rs
@@ -336,13 +336,12 @@ pub fn make_collector(
     doc_mapper: &dyn DocMapper,
     search_request: &SearchRequest,
 ) -> anyhow::Result<QuickwitCollector> {
-    let timestamp_field_opt = doc_mapper.timestamp_field()?;
     Ok(QuickwitCollector {
         split_id: String::new(),
         start_offset: search_request.start_offset as usize,
         max_hits: search_request.max_hits as usize,
         sort_by: doc_mapper.default_sort_by(),
-        timestamp_field_opt,
+        timestamp_field_opt: doc_mapper.timestamp_field(),
         start_timestamp_opt: search_request.start_timestamp,
         end_timestamp_opt: search_request.end_timestamp,
     })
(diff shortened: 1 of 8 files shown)


fmassot:
This does the job but it looks like more than a temporary solution.
But I would need some insights from @fulmicoton  about how we will handle query with ranges. 
If we look at this interesting [example](https://github.com/tantivy-search/tantivy/blob/dff0ffd38adc0861fc2b0463feed969a4ce18ee2/examples/custom_collector.rs) in tantivy we can see how we will handle filtering for fields which have a fast field. For fields that are only indexed, I guess the job will be done in the searcher. 
@evanxg852000, you can have a look at that to see if your first solution is sufficient or not.

fmassot:
@evanxg852000 can you make a rebase ?

fmassot:
Ok I'm not very fond of the implementation because it's very specific. 

Let's try something more generic:
```
pub struct QuickwitSegmentCollector {
    total_num_hits: u64,
    split_id: String,
    sort_by: SortingFieldComputer,
    predicate: Predicate,
    hits: BinaryHeap<PartialHitHeapItem>,
    max_hits: usize,
    segment_ord: u32,
}
```

where `Predicate` is a predicate which will be instantiated in QuickwitCollector::for_segment. A kind of resolve will be made to build either the case where if we have a timestamp we will have a predicate based on it and if we have no timestamp then it will always return true.

Now we can collect like this:
```
if self.predicate.call(doc_id) {
    self.total_num_hits += 1;
    self.collect_top_k(doc_id);
}
```

evanxg852000:
I think the TimestampFilter is in the same spirit just with the difference that we have it as `Option`, I could have it always available and more the logic you are describing inside.

quickwit-search/src/filters.rs
```diff
@@ -0,0 +1,85 @@
...
+        if expected_type != field_schema_type {
+            return Err(TantivyError::SchemaError(format!(
+                "Field {:?} is of type {:?}!={:?}",
+                field_entry.name(),
+                expected_type,
+                field_schema_type
+            )));
+        }
+
+        let timestamp_field_reader = segment_reader.fast_fields().u64(field)?;
```
fulmicoton, line 66: why not i64?

quickwit-search/src/filters.rs
```diff
@@ -0,0 +1,85 @@
...
+
+        Ok(TimestampFilter {
+            field,
+            time_range: (lower_bound, upper_bound),
+            timestamp_field_reader,
+        })
+    }
+
+    pub fn is_within_range(&self, doc_id: DocId) -> bool {
+        let timestamp_value = self.timestamp_field_reader.get(doc_id) as i64;
```
fulmicoton, line 82: u64. Casting i64 as u64 is almost never corrct in tantivy context.

quickwit-search/src/collector.rs
```diff
@@ -307,13 +335,16 @@ fn top_k_partial_hits(mut partial_hits: Vec<PartialHit>, num_hits: usize) -> Vec
 pub fn make_collector(
     doc_mapper: &dyn DocMapper,
     search_request: &SearchRequest,
-) -> QuickwitCollector {
-    QuickwitCollector {
+) -> anyhow::Result<QuickwitCollector> {
```
fulmicoton, line 338: why the anyhow::Result?

quickwit-search/src/collector.rs
```diff
@@ -176,6 +179,14 @@ impl SegmentCollector for QuickwitSegmentCollector {
     type Fruit = LeafSearchResult;
 
     fn collect(&mut self, doc_id: DocId, _score: Score) {
+        if let Some(ref timestamp_filter) = self.timestamp_filter_opt {
```
fulmicoton, line 182: The code redundancy is quite WTF.
```
fn collect(doc_id: DocId, score: Score) {
   if !self.accept_document(doc_id) {
       return doc_id;
   }
   self.total_num_hits += 1;                
   self.collect_top_k(doc_id);
}
```

fulmicoton (changes requested):
See comments inlined.

fulmicoton:
@evanxg852000 following on @fmassot , I suggest you do something halfway, keep the timestamp_filter_opt (which has the merit of being super explicit for the reader), and isolate its use in a private function that you could call 

```
fn accept_doc(&self, doc: DocId)
```

or even 

```
fn is_within_timerange(&self, doc: DocId) -> bool {
}
```

if you prefer being explicit (until we add other kind of filters)

quickwit-search/src/filters.rs
evanxg852000: There is a todo a bit up for the issues in Tnativy. It was merged so I was thinking we can chat how to bring that change in our internal tantivy since there have been other changes.

quickwit-search/src/collector.rs
evanxg852000: Ah change of docMapper API.

fulmicoton (approved):

PR #135, opened 2021-06-17
Add rest server

fulmicoton:


quickwit-serve/src/rest.rs
```diff
@@ -0,0 +1,143 @@
...
+use warp::Filter;
+use warp::Rejection;
+use warp::Reply;
+
+use crate::ApiError;
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct HitJson {
+    timestamp: i64,
```
fmassot, line 39: what is timestamp in HitJson?

quickwit-cli/src/cli.yaml
```diff
@@ -90,6 +90,29 @@ subcommands:
...
+                long: host
+                default_value: '0.0.0.0'
+                value_name: HOSTNAME
+            - port:
+                help: Port the REST API server should bind to.
+                long: port
+                default_value: '8080'
+                value_name: PORT
+            - peer_seeds:
+                help: List of peer socket address (e.g. 192.1.1.3:12001) to connect to to form a cluster.
```
evanxg852000, line 112: typo `to to from a cluster`

quickwit-cli/src/cli.yaml
```diff
@@ -90,6 +90,29 @@ subcommands:
                 help: Filters out documents after that timestamp (time-series indexes only)
                 long: end-timestamp
                 value_name: TIMESTAMP
+    - serve:
+        about: Serve a REST API to search an index
+        args:
+            - metastore-uri:
```
evanxg852000, line 96: Is this change only for `serve` or have we agreed to change the external view of users from `index-uri` to `metastore-uri` ?

quickwit-cli/src/main.rs
```diff
@@ -32,12 +32,17 @@ use quickwit_metastore::IndexMetadata;
 use quickwit_metastore::MetastoreUriResolver;
 use quickwit_proto::SearchRequest;
 use quickwit_search::single_node_search;
+use quickwit_serve::serve_cli;
```
evanxg852000, line 35: Maybe we can refactor by following what we have been doing. create `serve_cli` in the `quickwit-cli` crate to allow it to display user messages like.:
```bash 
Server running on http://localhost:8000  
You can search with `curl http://localhost:8000/api/v1/search?query=ada`
```

quickwit-serve/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+use quickwit_metastore::MetastoreUriResolver;
+use quickwit_search::SearchServiceImpl;
+use quickwit_storage::StorageUriResolver;
+use tracing::debug;
+mod error;
+mod rest;
+
+pub use crate::error::ApiError;
+
+pub async fn serve_cli(args: ServeArgs) -> anyhow::Result<()> {
```
evanxg852000, line 36: Maybe we can refactor by following what we have been doing. create `serve_cli` in the `quickwit-cli` crate to allow it to display user messages like.:
```bash 
Server running on http://localhost:8000  
You can search with `curl http://localhost:8000/api/v1/search?query=ada`
```
Also the server start even with a wrong metastore-uri.
we could handle this in smaller tasks though.

quickwit-serve/src/rest.rs
```diff
@@ -0,0 +1,296 @@
...
+fn make_reply<T: serde::Serialize>(result: Result<T, ApiError>) -> reply::WithStatus<reply::Json> {
+    match result {
+        Ok(success) => reply::with_status(reply::json(&success), StatusCode::OK),
+        Err(error) => reply::with_status(reply::json(&error), error.http_status_code()),
+    }
+}
+
+/// This function returns a formated error based on the given rejection reason.
+async fn recover_fn(rejection: Rejection) -> Result<impl Reply, Rejection> {
+    match rejection.find::<serde_qs::Error>() {
```
fmassot, line 153: Rejection can have other causes that serde_qs error, I would add a todo here to handle all cases. Cf [example](https://github.com/seanmonstar/warp/blob/master/examples/rejections.rs#L84-L90)

I tried to make a "HEAD" request in the test and I receive a 404 response but normally I should have had a 405 method not allowed.

quickwit-cli/src/cli.yaml
fulmicoton: That's a good question... One benefit of the metastore like this is that we allow for more than one index.

I guess we could also offer more than one index-uri?
Let me try it.

fmassot (approved):

evanxg852000 (approved):

quickwit-cli/src/cli.yaml
fmassot: As for me, this is the way to go because we want to serve several indexes. 
But the user starts by giving an index-uri... and we don't explain to him that we deduce from it a metastore uri. And maybe it would be simpler to say to the user: "give a us a metastore uri and a index name"....

quickwit-cli/src/main.rs
```diff
@@ -23,7 +23,7 @@
 use anyhow::{bail, Context};
 use byte_unit::Byte;
 use clap::{load_yaml, value_t, App, AppSettings, ArgMatches};
-use once_cell::sync::Lazy;
+use quickwit_common::extract_metastore_uri_and_index_id_from_index_uri;
```
evanxg852000, line 26: Nice move :+1:

evanxg852000:
It works fine :+1:

PR #137, opened 2021-06-18
Warmup Fast Fields

evanxg852000:
### Context / purpose
Add implementation of warm_up for fast fields. 
[See Issue](https://github.com/quickwit-inc/quickwit/issues/129)

### Description
Implemented based on what is in `barrel` repo. we still need to discuss how we gather all the fast fields base on the request.

### DIFF
--- a/quickwit-doc-mapping/src/mapper.rs
+++ b/quickwit-doc-mapping/src/mapper.rs
@@ -48,7 +48,7 @@ pub enum SortBy {
     /// Sort by a specific field.
     SortByFastField {
         /// Field to sort by.
-        field_name: String,
+        field: Field,
         /// Order to sort by. A usual top-K search implies a Descending order.
         order: SortOrder,
     },
--- a/quickwit-doc-mapping/src/mapper.rs
+++ b/quickwit-doc-mapping/src/mapper.rs
@@ -48,7 +48,7 @@ pub enum SortBy {
     /// Sort by a specific field.
     SortByFastField {
         /// Field to sort by.
-        field: Field,
+        field_name: String,
         /// Order to sort by. A usual top-K search implies a Descending order.
         order: SortOrder,
     },
--- a/quickwit-search/src/collector.rs
+++ b/quickwit-search/src/collector.rs
@@ -79,16 +79,12 @@ fn resolve_sort_by(
     segment_reader: &SegmentReader,
 ) -> tantivy::Result<SortingFieldComputer> {
     match sort_by {
-        SortBy::SortByFastField { field_name, order } => {
-            if let Some(field) = segment_reader.schema().get_field(&field_name) {
-                let fast_field_reader = segment_reader.fast_fields().u64_lenient(field)?;
-                Ok(SortingFieldComputer::SortByFastField {
-                    fast_field_reader,
-                    order: *order,
-                })
-            } else {
-                Ok(SortingFieldComputer::SortByDocId)
-            }
+        SortBy::SortByFastField { field, order } => {
+            let fast_field_reader = segment_reader.fast_fields().u64_lenient(*field)?;
+            Ok(SortingFieldComputer::SortByFastField {
+                fast_field_reader,
+                order: *order,
+            })
         }
         SortBy::DocId => Ok(SortingFieldComputer::SortByDocId),
     }
@@ -244,8 +240,9 @@ impl QuickwitCollector {
             fast_fields.push(*timestamp_field);
         }
 
-        //TODO: add sort fast field if `self.sort_by`  is of variant `SortBy::SortByFastField`.
-        //      This needs the doc_mapper but let's clarify the api first.
+        if let SortBy::SortByFastField { field, .. } = self.sort_by {
+            fast_fields.push(field);
+        }
 
         //TODO: find a way to add remaining fastfields from user search request.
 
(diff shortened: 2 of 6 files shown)


evanxg852000:
Yes, that's exactly what I was also thinking too.

quickwit-search/src/collector.rs
```diff
@@ -237,6 +233,21 @@ impl QuickwitCollector {
...
+        let mut fast_fields = vec![];
+        if let Some(ref timestamp_field) = self.timestamp_field_opt {
+            fast_fields.push(*timestamp_field);
+        }
+
+        if let SortBy::SortByFastField { field, .. } = self.sort_by {
+            fast_fields.push(field);
+        }
+
+        //TODO: find a way to add remaining fastfields from user search request.
```
fmassot, line 247: I think you can remove the comment, see the general comment.

fmassot (commented):
My understanding of fast field was incomplete, I hope it's no more the case as I wrote these words.
First all tantivy queries use the inverted indexes, no fast field here. 
Second, it is possible to use a fast field to filer out some documents, we are doing is for the timestamp filtering because we have the start and end parameters in the request. 
**But** when the user writes a range query on another field like "price>1000", it will create a RangeQuery and tantivy uses the inverted index and the term dictionnary to build the list of matching doc ids. So in this case, we would need to load many terms!!!!! And as far as I understand the code, we are not currently warming the terms in the range query, I guess @fulmicoton did not do that because it's expensive and not very smart.
 
So now I'm wondering how we will be efficient on range queries. My intuition tells me that we need to build a compact index for that like the bkd tree.

quickwit-doc-mapping/src/mapper.rs
```diff
@@ -46,7 +46,7 @@ pub enum SortBy {
     /// Sort by a specific field.
     SortByFastField {
         /// Field to sort by.
-        field_name: String,
+        field: Field,
```
fulmicoton, line 49: we want to keep it a String, because the schema might change.

quickwit-search/src/leaf.rs
```diff
@@ -68,15 +68,23 @@ async fn warmup(
...
-    _quickwit_collector: &QuickwitCollector,
+async fn warm_up_fastfields(
+    searcher: &Searcher,
+    quickwit_collector: &QuickwitCollector,
 ) -> anyhow::Result<()> {
-    // TODO implement me
+    let mut warm_up_futures = Vec::new();
+    for field in quickwit_collector.fast_fields() {
+        for segment_reader in searcher.segment_readers() {
+            let fast_field_slice = segment_reader.fast_fields().fast_field_data(field, 0)?;
```
fulmicoton, line 82: we also need to warm up the timestamp fielld if there is filtering involved

quickwit-doc-mapping/src/mapper.rs
fmassot: I'm not sure about that. 
I think that the unique identifier of a field would be its id and that's why I like to have a field, because it's just the id.

Another rationale I had on that topic is : the doc mapper is responsible for defining the schema, it's the doc mapper that knows the up-to-date schema. That's why we can safely do the name resolving when building the doc mapper.

It may not be true on an old schema : for example if the schema has changed in the past and that the user rename the field, we will not be able to do the resolving on the old segment schema because the name was different in the past.

quickwit-search/src/leaf.rs
evanxg852000: Not sure i understand this, but the `timestamp_field` is already part of the fast fields list as long as there is one.

fmassot: can you also check that the field is fast just after making the resolving ?

quickwit-search/src/collector.rs
```diff
@@ -334,6 +335,24 @@ fn top_k_partial_hits(mut partial_hits: Vec<PartialHit>, num_hits: usize) -> Vec
...
+/// Extracts all fast field names.
+fn extract_fast_field_names(doc_mapper: &dyn DocMapper) -> Vec<String> {
+    let mut fast_fields = vec![];
+    let timestamp_field_name_opt = doc_mapper.timestamp_field_name();
+    if let Some(timestamp_field) = timestamp_field_name_opt.as_ref() {
+        fast_fields.push(timestamp_field.clone());
+    }
+
+    if let SortBy::SortByFastField { field_name, .. } = doc_mapper.default_sort_by() {
+        if Some(&field_name) == timestamp_field_name_opt.as_ref() {
```
fmassot, line 347: I think this line is wrong

fmassot (changes requested):
can you add a test ?

evanxg852000:
I can add test, but since our Trait is complex, I don't think `mockall` can help. I have a custome `DocMapper` impl and this makes me think of moving `extract_fast_field_names`. If this is ok I can push.

quickwit-search/src/collector.rs
evanxg852000: Are you saying we should not compare field names?

fmassot (approved):

fmassot:
LGTM.
Note for @fulmicoton, an important issue remains, we have no test on timestamp filtering, this will be addressed in another PR by @evanxg852000

PR #138, opened 2021-06-18
Integration tests

evanxg852000:
### Context / purpose
Think of basic integration test & implement [See Issue](https://github.com/quickwit-inc/quickwit/issues/134)

### Description
- Implemented basic integration tests for the commands `new` , `index`, `delete`.
- Enable storage test for localstack in CI with `-- --ignored` cargo test flag

### Checklist
- [x] add piped input test
- [x] refactor
- [x] add localstack s3 test: currently, we cannot interact directly with s3 or storage; making it hard to create region with custom endpoint. Discuss ways to override rusoto region or endpoint to make using easy to use localstack. we could also provide a `s3+localstack://` URL scheme.


quickwit-cli/tests/cli.rs
```diff
@@ -0,0 +1,243 @@
...
+    Ok(())
+}
+
+#[test]
+fn test_cmd_delete() -> Result<()> {
+    let test_env = create_test_env(false)?;
+    create_wikipedia_index(&test_env);
+
+    // Delete empty index.
+    make_command(format!("delete --index-uri {} --dry-run", test_env.uri).as_str())
```
fmassot, line 181: why using dry-run ?

fmassot (commented):
One general remark : I really like the way tests are done in ripgrep.
Something like this would be very cool
`
cmd = TestCommand("delete", test_env_struct);
cmd.arg("-Eutf-16le").arg("Шерлок Холмс");
assert_something(cmd.output())
`
https://github.com/BurntSushi/ripgrep/blob/master/tests/feature.rs#L5-L12
and https://github.com/BurntSushi/ripgrep/blob/master/tests/util.rs

What do you think?

You don't have to rewrite everything but just having some way to have more control on arguments would be nice.

quickwit-cli/tests/cli.rs
evanxg852000: I am asserting on the output of the execution to see if the correct message is shown to the user. since the actual deletion is tested later at `delete_index(&test_env);`

evanxg852000:
I can switch to adding the arg like you are suggesting, but I felt it's more natural fo anyone reading this how to use the commands by seeing something like this `format!("delete --index-uri {} --dry-run", test_env.uri)`. if we are all ok that adding the args one at a time is a better approach then I can change to that.

quickwit-cli/tests/cli.rs
fmassot: indeed, so the name `delete_index` is confusing as it's also doing some assertions. 

Can you rename it and do the same for all functions that do some assertions (renaming like delete_index_and_check_success would be sufficient) ? Another possibility is to do the assertions in the test directly.

fmassot:
> I can switch to adding the arg like you are suggesting, but I felt it's more natural fo anyone reading this how to use the commands by seeing something like this `format!("delete --index-uri {} --dry-run", test_env.uri)`

It's more food for thoughts here, we can keep right now the format function. But in the end you want something that wraps the command to better control args and env settings and in the same time be as readable as the "format" command. Don't you think the tests in ripgrep achieves that ? :)

evanxg852000:
So `ripgrep` is just using the standard `Command` along with the directory, and probably this is my issue because I am implicitly doing this work.
```rust
  for arg in argument.split_whitespace() {
      cmd.arg(arg);
  }
```
We have use cases that are sequential, that is we need an index and data to be indexed before searching. so even if we wrap our commands we still need these sequences to happen for each test.

quickwit-cli/tests/cli.rs
fulmicoton: renaming is good but `delete_index_and_check_success` is not helping.

`delete_index_dry_run` would be nice.

quickwit-cli/tests/cli.rs
```diff
@@ -0,0 +1,243 @@
...
+        format!(
+            "new --index-uri {} --doc-mapper-type wikipedia",
+            test_env.uri
+        )
+        .as_str(),
+    )
+    .assert()
+    .success();
+    let metadata_file_exist = object_storage.exists(Path::new("quickwit.json")).await?;
+    assert_eq!(metadata_file_exist, true);
```
fulmicoton, line 220: nitpick: assert! would have been ok here.

quickwit-cli/tests/helpers.rs
```diff
@@ -0,0 +1,122 @@
...
+    cmd
+}
+
+/// A struct to hold few info about the test environement.
+pub struct TestEnv {
+    pub dir: TempDir,
+    pub path: PathBuf,
+    pub uri: String,
+    pub config_path: PathBuf,
+    pub log_data_path: PathBuf,
```
fulmicoton, line 94: nitpick: i prefer a name like log_input_filepath to make it unambiguous that it is the file that contains the input stuff as opposed to the directory that will receive the log index data.

quickwit-cli/tests/helpers.rs
```diff
@@ -0,0 +1,122 @@
...
+    let mut cmd = Command::cargo_bin("quickwit-cli").unwrap();
+    for arg in argument.split_whitespace() {
+        cmd.arg(arg);
+    }
+    cmd
+}
+
+/// A struct to hold few info about the test environement.
+pub struct TestEnv {
+    pub dir: TempDir,
```
fulmicoton, line 90: horrible name. what dir is this?

quickwit-cli/tests/helpers.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub path: PathBuf,
```
fulmicoton, line 91: horrible name. what path is this? is it a file or a dir?

quickwit-cli/tests/helpers.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub uri: String,
```
fulmicoton, line 92: Same as above. Horrible name.

quickwit-cli/tests/helpers.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub config_path: PathBuf,
+    pub log_data_path: PathBuf,
+    pub wiki_data_path: PathBuf,
+}
+
+/// Creates all necessary artifacts in a test environement.
+pub fn create_test_env(is_s3: bool) -> anyhow::Result<TestEnv> {
```
fulmicoton, line 99: An enum would be great. 

```
enum TestStorage {
    S3ViaLocalStack(PathBuf),
    LocalFileSystem
}

impl TestStorage {
   fn uri(&self) -> String {
     //...
   }
}
```

Note that the point is NOT to reorganize the code but to make the bool meaning clear in the call sites.

quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -132,8 +132,8 @@ pub fn parse_split_uri(split_uri: &str) -> Option<(String, PathBuf)> {
     static SPLIT_URI_PTN: OnceCell<Regex> = OnceCell::new();
     SPLIT_URI_PTN
         .get_or_init(|| {
-            // s3://bucket/path/to/split
-            Regex::new(r"s3://(?P<bucket>[^/]+)/(?P<path>.*)").unwrap()
+            // s3://bucket/path/to/split or s3+localstack://bucket/path/to/split
```
fulmicoton, line 135: great comment!

quickwit-storage/src/object_storage/s3_compatible_storage_uri_resolver.rs
```diff
@@ -29,16 +29,13 @@ use crate::{S3CompatibleObjectStorage, StorageFactory};
 /// S3 Object storage Uri Resolver
 pub struct S3CompatibleObjectStorageFactory {
     region: Region,
-    protocol: String,
+    protocol: &'static str,
```
fulmicoton, line 32: yes!

quickwit-cli/tests/helpers.rs
```diff
@@ -0,0 +1,122 @@
...
+
+const WIKI_JSON_DOCS: &str = r#"{"body": "foo", "title": "shimray", "url": "https://wiki.com?id=10"}
+{"body": "bar", "title": "shimray", "url": "https://wiki.com?id=12"}
+{"body": "baz", "title": "preshow", "url": "https://wiki.com?id=11"}
+{"body": "buz", "title": "frederick", "url": "https://wiki.com?id=48"}
+{"body": "biz", "title": "modern", "url": "https://wiki.com?id=13"}
+"#;
+
+/// Creates a file with a content
+pub fn create_file(path: &Path, data: &str) {
```
fulmicoton, line 66: nitpick: `https://doc.rust-lang.org/std/fs/fn.write.html` does this.

quickwit-cli/tests/helpers.rs
fulmicoton, line 94: I am not entirely confident in the following comment...

Wouldn't having only one member called `input_docs_filepath` in place of `log_data_path` and `wiki_data_path` worked just as well?

fulmicoton (changes requested):
Cool work! Much better than what I had in mind.

I added a bunch of comments on how to improve TestEnv.
As always, feel free to ignore the nitpicks.

The naming of the members of TestEnv was a bit sloppy for the reader to understand what this is about. Of course in unit tests it is ok to be a little bit more sloppy, but since it is the test framework, I think it is important to be a little bit clearer to make sure everyone uses your framework as you intended it.

quickwit-cli/tests/helpers.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Path of the temporary directory of the test.
+    pub local_directory_path: PathBuf,
+    /// The index uri.
+    pub index_uri: String,
+    /// Resource files needed for the test.
+    pub resource_files: HashMap<&'static str, PathBuf>,
+}
+
+pub enum TestStorageType {
+    S3ViaLocalStaorage(PathBuf),
```
fmassot, line 86: typo `S3ViaLocalStorage` is better

.github/workflows/ci.yml
```diff
@@ -65,4 +65,4 @@ jobs:
         env:
           AWS_ACCESS_KEY_ID: dummy-access-key
           AWS_SECRET_ACCESS_KEY: dummy-secret-key
-        run: cargo test --all
+        run: cargo test --all-features
```
fmassot, line 68: great!

quickwit-cli/tests/cli.rs
```diff
@@ -0,0 +1,311 @@
...
+
+#[tokio::test]
+#[cfg_attr(not(feature = "ci-test"), ignore)]
+async fn test_all_with_s3_localstack() -> Result<()> {
+    let s3_path = PathBuf::from("quickwit-integration-tests/indices/data1");
+    let test_env = create_test_env(TestStorageType::S3ViaLocalStaorage(s3_path))?;
+    let object_storage =
+        S3CompatibleObjectStorage::from_uri(localstack_region(), &test_env.index_uri)?;
+
+    println!("EVAN");
```
fmassot, line 283: remove `println!`

fmassot (changes requested):

fmassot (approved):

PR #151, opened 2021-06-23
Filtering Optimization

evanxg852000:
### Context / purpose
Optimize filtering if the predicate is always matched [see issue](https://github.com/quickwit-inc/quickwit/issues/139)

### Description
- implemented optimization
- added test for filtering

### Checklist
- [x] tested locally
- [x] added unit tests

### DIFF
--- a/quickwit-doc-mapping/src/default_doc_mapper/default_mapper.rs
+++ b/quickwit-doc-mapping/src/default_doc_mapper/default_mapper.rs
@@ -257,7 +257,7 @@ impl DocMapper for DefaultDocMapper {
         //TODO: This is just a placeholder implementation
         // allowing us to test few things up front.
         let schema = self.schema();
-        let default_fields = vec![schema.get_field("body").unwrap()];
+        let default_fields: Vec<Field> = self.default_search_fields.values().cloned().collect();
         let query_parser = tantivy::query::QueryParser::new(
             schema,
             default_fields,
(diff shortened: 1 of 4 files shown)


quickwit-doc-mapping/src/default_doc_mapper/default_mapper.rs
```diff
@@ -253,13 +253,27 @@ impl DocMapper for DefaultDocMapper {
         Ok(document)
     }
 
-    fn query(&self, _request: &SearchRequest) -> Result<Box<dyn Query>, QueryParserError> {
-        todo!()
+    fn query(&self, request: &SearchRequest) -> Result<Box<dyn Query>, QueryParserError> {
+        //TODO: This is just a placeholder implementation
+        // allowing us to test few things up front.
+        let schema = self.schema();
+        let default_fields = vec![schema.get_field("body").unwrap()];
```
fmassot, line 260: you can use the default_fields from the doc mapper directly! it's a hash map with keys as field name and values as field.

fulmicoton: (I think we should ditch it though to be coherent with the previous discussions... Only manipulating "field names" and no Field outside of the scale of a spit has the merit of allowing schema change. It does not matter too too much however.

fmassot: yes you're right, let's keep the field names only, so a Vec<String> is good enough here.

evanxg852000: Thanks @fmassot  :+1: this has the merit of making my test fully independent in the meantime.

fmassot (approved):
Ok for me. I opened an issue to change the doc mapper behavior relative to field resolution : https://github.com/quickwit-inc/quickwit/issues/152

fmassot (approved):

fmassot (approved):

PR #153, opened 2021-06-24
Remove field resolution from DocMapper

evanxg852000:
### Context / purpose
Remove field resolution from DocMapper [See Issue](https://github.com/quickwit-inc/quickwit/issues/152)

### Description
Currently, we are removing just the reference to `Field` on the interface. We still have internal field resolution in order to validate. 

### Checklist
- [ ] tested locally
- [ ] updated unit tests


quickwit-doc-mapping/src/default_doc_mapper/mod.rs
```diff
@@ -51,7 +51,7 @@ pub fn is_valid_field_mapping_name(field_mapping_name: &str) -> bool {
 const TANTIVY_DOT_SYMBOL: &str = "__dot__";
 
 /// Given a field name which can contains some dots, return the schema field
-fn resolve_field_name(schema: &Schema, field_name: &str) -> Option<Field> {
+pub fn resolve_field_name(schema: &Schema, field_name: &str) -> Option<Field> {
```
fmassot, line 54: I don't like to see this function public, this is specific to the default doc mapper and should stay private

quickwit-doc-mapping/src/mapper.rs
```diff
@@ -79,12 +80,12 @@ pub trait DocMapper: Send + Sync + Debug + DynClone + 'static {
     }
     /// Returns the timestamp field.
     fn timestamp_field(&self) -> Option<Field> {
-        None
+        self.timestamp_field_name()
```
fmassot, line 83: don't use resolve_field_name here

fulmicoton (approved):

PR #171, opened 2021-07-02
LRU cache for the storage Issue/167

fulmicoton:


quickwit-storage/src/cache/storage_with_cache.rs
```diff
@@ -0,0 +1,111 @@
...
+
+#[async_trait]
+impl Storage for StorageWithCache {
+    async fn put(&self, path: &Path, payload: PutPayload) -> StorageResult<()> {
+        // We actually do not cache anything on put as it is meaningless in Quickwit.
+        self.storage.put(path, payload).await
+    }
+
+    async fn copy_to_file(&self, path: &Path, output_path: &Path) -> StorageResult<()> {
+        if let Some(bytes) = self.cache.get_all(path).await {
```
evanxg852000, line 43: I guess this implies that the cache if it contains an entire local file, then it's always in sync with the disk version?

quickwit-storage/src/cache/in_ram_slice_cache.rs
```diff
@@ -0,0 +1,213 @@
...
+/// A simple in-resident memory slice cache.
+pub struct SliceCache {
+    inner: Mutex<NeedMutSliceCache>,
+}
+
+impl SliceCache {
+    /// Creates an slice cache with the given capacity.
+    ///
+    /// The SliceCache is guaranteed to not store more than `capcity_in_bytes`.
+    /// If None, SliceCache acts as if the capacity is unlimited.
```
evanxg852000, line 103: nitpick:
I think this doc comment should go on `NeedMutSliceCache`.  we already have `with_capacity_in_bytes` and `with_infinite_capacity` methods that explain.  My point is because `capcity_in_bytes` here is not an Option

evanxg852000:
Really well written code :clap:

quickwit-storage/src/cache/storage_with_cache.rs
fulmicoton: Very good point. I changed the implementation to panic on write operations.

quickwit-storage/src/cache/in_ram_slice_cache.rs
fulmicoton: yes you are right

PR #181, opened 2021-07-05
Automate Binary Build

evanxg852000:
### Context / purpose
We need to automate building Quickwit binaries build. [See](https://github.com/quickwit-inc/quickwit/issues/168)

### Description
using CI and `quickwit-inc/upload-to-github-release@v1` to upload the binaries.

### How was this PR tested?
Testing the resulting binaries on platforms I have: macos-10.15, ubuntu-20.04

### Checklist
- [x] add step to build for other architectures
- [x] tested locally on each platform (macos-10.15, ubuntu-20.04)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,7 +4,7 @@ on:
   push:
     tags:
       - '*'
-
+      
 jobs:
   publish:
     name: Publish for ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


evanxg852000:
@Paul Masurel I have been having issues with `OpenSSL` in the cross-compilation steps. 
I have  added this under in quickwit telemetry and just want your feedback on it 
`openssl = { version = '0.10', features = ["vendored"] }` https://docs.rs/openssl/0.10.35/openssl

fmassot (approved):

fmassot:
I noticed that the name of the binaries contain the word `unknown`, why is that ? I saw that vector has the same `unknown` so this is ok I guess but I would be interested by the reason :)

PR #182, opened 2021-07-05
add coverage gh action

PSeitz:


PSeitz:
Some needs to enable the project in codecov.io

PR #188, opened 2021-07-06
removing deps to openssl.

fulmicoton:


evanxg852000:
Cool :+1:  this solves the issue. I was thinking this should not matter on `quickwit-cli` since it was a dev dep.

Issue #53, opened 2021-05-26
Investigate security considerations on the local file storage


fulmicoton:
The URI resolution scheme for the file system can have some important risk implication.
We also want to handle file:///var/local and the incorrect but common file://var/local format. 

We need to make sure someone cannot access files that are not under the root directory using the "../" schemefor instnace.

Generally speaking: we should put some thoughts in it. 

----

UPDATE:

We want to enforce the fact that path accessed by the FileStorage are child path of the FileStorage root.

fulmicoton:
(As far as I could tell #61 is not solving this bug.)

fmassot:
> We also want to handle file:///var/local and the incorrect but common file://var/local format.
`file://var/local` currently works but `var/local` is interpreted as a variable path. 
I think we can consider that `file://var/local` be the same `file:///var/local'

Concerning security issues on path, I think we can simply ban the use of `..`. We accept only absolute path without parent directory so that when we manipulate path inside quickwit there is no suprise.

PR #193, opened 2021-07-06
Prevents interpreting a local file storage uri as relative.

### LINKED ISSUE #126: Need to be stricter on index-uri ?
This is clearly not a bug but I expect some users will do the same error as me.

I expected the index-uri to have an absolute uri but if you have only two slashes it will be relative. 

So having `file://folder1/folder2/folder3/folder4/wikipedia` will create 5 folders because I forget the slash.

I suggest to support only absolute path.

fmassot:
Closes #126


quickwit-storage/src/local_file_storage.rs
```diff
@@ -46,18 +46,31 @@ impl fmt::Debug for LocalFileStorage {
 }
 
 impl LocalFileStorage {
-    /// Creates a file storage instance given a root path.
-    pub fn new(root: PathBuf) -> Self {
-        Self { root }
-    }
-
     /// Creates a file storage instance given a uri
+    /// Both scheme `file:///{path}` and `file://{path}` are accepted.
```
fulmicoton, line 50: It does not solve #53 contrary to what the comment says.
(it does solve #208 and #126 however)

fulmicoton (approved):
Ok for me. But I changed the description.
It does not fix the attack surface I had in mind.
Also we need to fix the unit test.

PR #196, opened 2021-07-06
move code to constructor

PSeitz:
move code to CreateIndexArgs constructor
add lib test
add serial for test dependency
add prepare_test script to create and clean s3 test data bucket

### DIFF
--- a/quickwit-cli/tests/cli.rs
+++ b/quickwit-cli/tests/cli.rs
@@ -28,8 +28,10 @@ use crate::helpers::{create_test_env, make_command, spawn_command};
 use anyhow::Result;
 use helpers::{TestEnv, TestStorageType};
 use predicates::prelude::*;
+use quickwit_cli::{create_index_cli, CreateIndexArgs};
 use quickwit_storage::{localstack_region, S3CompatibleObjectStorage, Storage};
 use serde_json::{Number, Value};
+use serial_test::serial;
 use std::{
     io::Read,
     path::{Path, PathBuf},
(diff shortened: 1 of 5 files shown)


quickwit-cli/tests/prepare_tests.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+awslocal s3 mb s3://quickwit-integration-tests && awslocal s3 rm --recursive s3://quickwit-integration-tests
```
fmassot, line 2: I think rm is not necessary as we don't use localstack persistence

PSeitz: if something goes wrong it's necessary

fmassot: and this works only if localstack is started

quickwit-cli/tests/cli.rs
```diff
@@ -272,9 +274,12 @@ async fn test_cmd_dry_run_delete_on_s3_localstack() -> Result<()> {
 }
 
 #[tokio::test]
+#[serial]
```
fmassot, line 277: why do you need serial ?

quickwit-cli/tests/cli.rs
```diff
@@ -330,11 +335,91 @@ async fn test_all_with_s3_localstack() -> Result<()> {
...
+    let metadata_file_exist = object_storage.exists(Path::new("quickwit.json")).await?;
+    assert_eq!(metadata_file_exist, false);
+
+    Ok(())
+}
+
+#[tokio::test]
+#[serial]
+#[cfg_attr(not(feature = "ci-test"), ignore)]
+async fn test_all_with_s3_localstack_api() -> Result<()> {
```
fmassot, line 364: what is the difference between `test_all_with_s3_localstack_api` and `test_all_with_s3_localstack_cli` ?

fmassot: can you be more explicit about the difference between the two tests ?

quickwit-cli/tests/prepare_tests.sh
fmassot: ok, locally it is necessary as it's annoying to stop and restart localstack

quickwit-cli/tests/cli.rs
PSeitz: the tests have a dependency currently since they both start a webserver

quickwit-cli/tests/cli.rs
PSeitz: The api test is not completely switching to the api test for all commands. the idea is that one tests the cli and one tests the api via internal structs.

quickwit-cli/tests/prepare_tests.sh
PSeitz: in the localstack api I see only a start command, no stop command

quickwit-cli/tests/cli.rs
fmassot: indeed!

quickwit-cli/tests/cli.rs
fmassot: I saw that afterwards, not sure if it's worth doing it as cli tests are quite heavy. That's ok for now but we need to keep in mind that to avoid inflating too much cli tests.

PR #198, opened 2021-07-06
add equality comparator

PSeitz:
use CreateIndex constructor and make fields private
replace matches! with assert_eq
add test comment


fmassot (approved):
LGTM

PR #216, opened 2021-07-07
add hdfs_doc_mapper.json

PSeitz:


fulmicoton:
I recommend renaming `hdfslogsdataset_doc_mapper.json` or something like that to make sure people don't think it is a conf to use hdfs as a storage.

examples/doc_mappers/hdfs_doc_mapper.json
```diff
@@ -0,0 +1,40 @@
+{
+    "store_source": true,
```
fulmicoton, line 2: I think `"store_source": false` might be better here. (The json in the _source field can be confusing)

PSeitz:
> I recommend renaming `hdfslogs_doc_mapper.json` or something like that to make sure people don't think it is a conf to use hdfs as a storage.

ah yeah true

examples/doc_mappers/hdfs_doc_mapper.json
PSeitz: done

PR #221, opened 2021-07-08
A faster CI

fulmicoton:
This accelerates our CI process by reallying on an external pre-built docker container on amazon ECR.
The recipe used to build this container is available on github.
https://github.com/quickwit-inc/quickwit-builder

Right now its merit is to cache cargo-tarpaulin and cargo-check.

This PR also reorganizes CI steps and removes "ignore" steps from the normal CI process.
Code coverage is the only time we run these tests and it is triggered on merge into main.


.github/workflows/coverage.yml
```diff
@@ -17,13 +15,13 @@ jobs:
...
           components: rustfmt # fmt used by the protobuf code generation
-      - name: Create bucket in localstack
-        run: docker-compose -f localstack.yaml exec -T localstack awslocal s3 mb s3://quickwit-integration-tests
       - uses: actions-rs/install@v0.1
         name: install cargo tarpaulin
         with:
           crate: cargo-tarpaulin
           use-tool-cache: true
+      - name: Create bucket in localstack
+        run: docker-compose -f localstack.yaml exec -T localstack awslocal s3 mb s3://quickwit-integration-tests
```
PSeitz, line 24: this references localstack.yml, but it was deleted

PSeitz:
would be good to add the dockerfile also under .github/workflows

Issue #238, opened 2021-07-09
When doing a distributed search with multiple nodes, each leaf node will search all the splits.


mosuka:
**Describe the bug**
When doing a distributed search with multiple nodes, each leaf node will search all the splits.

**Expected behavior**
Make leaf_search search for the splits assigned by root_search.

PR #239, opened 2021-07-09
Fixed a bug in distributed search.

mosuka:
### Context / purpose
Fixed a bug in distributed search.

### Description
See #238 

### Checklist
- [x] tested locally
- [x] added unit tests
- [x] included documentation


quickwit-search/src/root.rs
```diff
@@ -103,6 +104,7 @@ pub async fn root_search(
             split_uris: jobs.iter().map(|job| job.split.clone()).collect(),
```
fulmicoton, line 104: so these are split_id, not split_uris right?

We should probably rename?

fulmicoton (changes requested):
I think split_uris should be renamed split_ids

Can you open a ticket to add unit tests?
It is quite difficult, but we should be able to test root and leaf independently via mockall. (I did so in barrel)

quickwit-search/src/root.rs
mosuka: Yes, I’ll rename to split_ids.

mosuka:
I created an issue for adding a unit test for distributed search.
https://github.com/quickwit-inc/quickwit/issues/240

fulmicoton (approved):

mosuka:
@fulmicoton 
Thanks.

Issue #269, opened 2021-07-11
Search server return 200 but log an ERROR


fmassot:
Request executed in browser on hdfs log dataset

```
http://quickwit-test-392975419.us-east-1.elb.amazonaws.com/api/v1/hdfs_logs/search?query=%22resource.18%22
```

or the curl command:
```bash
curl -v 'http://your-host/api/v1/hdfs_logs/search?query=%22datanode+12%22'  -H "Content-Type: application/json"
```

both gives:

```
Jul 11 11:10:51.577 ERROR quickwit_search::root: err=Failed to search a leaf node due to Status { code: InvalidArgument, message: "Invalid query: QueryParserError(The field '\"body\"' does not have positions indexed)", metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Sun, 11 Jul 2021 11:10:51 GMT"} } }
```

The error is expected but the server return a 200 with this result:

```json
{
  "numHits": 0,
  "hits": [],
  "numMicrosecs": 2944
}
```

PR #275, opened 2021-07-12
Deserialize error coming from grpc and fail if any split fails

fmassot:
This is linked to issue #269 which we cannot ignore for the release.

SearchError are deserialized to make sure that the server answers correctly to the client. For example, a error in the query parsing should return a 400.


fmassot:
closing as this is now done in #279

PR #279, opened 2021-07-12
Improves Error handling and different bugs.

### LINKED ISSUE #269: Search server return 200 but log an ERROR

fulmicoton:
Closes #269.

SearchError are JSON serialized automatically by the grpc adapter.
Upon one leaf failure the whole request now fails (before it was just ignore the single leaf error).
We do not hold the search pool lock forever in the root and in the assign_client method.

### DIFF
--- a/quickwit-search/src/root.rs
+++ b/quickwit-search/src/root.rs
@@ -22,7 +22,7 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use anyhow::Context;
+use tantivy::TantivyError;
 use tokio::task::JoinHandle;
 use tracing::*;
 
@@ -35,10 +35,12 @@ use tantivy::collector::Collector;
 use tokio::task::spawn_blocking;
 
 use crate::client_pool::Job;
+use crate::error::parse_grpc_error;
 use crate::list_relevant_splits;
 use crate::make_collector;
 use crate::ClientPool;
 use crate::SearchClientPool;
+use crate::SearchError;
 
 // Measure the cost associated to searching in a given split metadata.
 fn compute_split_cost(_split_metadata: &SplitMetadata) -> u32 {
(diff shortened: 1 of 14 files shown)


quickwit-search/src/root.rs
```diff
@@ -80,44 +82,33 @@ pub async fn root_search(
...
     debug!(assigned_leaf_search_jobs=?assigned_leaf_search_jobs, "Assigned leaf search jobs.");
 
-    let clients = client_pool.clients.read().await;
-
     // Create search request with start offset is 0 that used by leaf node search.
     let mut search_request_with_offset_0 = search_request.clone();
     search_request_with_offset_0.start_offset = 0;
     search_request_with_offset_0.max_hits += search_request.start_offset;
 
     // Perform the query phese.
```
fmassot, line 90: typo `phase`

fmassot (approved):
Great job as always fulmicoton! 

I think we can merge it but I have two things in mind:
- the tests are useless to check if the serialization/deserialization of errors is working well as we use the mocking service
- on the handling errors, it's worth noting that for some errors like invalid query, this is not useful to retry.

PR #280, opened 2021-07-12
Error message in Installer

evanxg852000:
### Context / purpose
Closes https://github.com/quickwit-inc/quickwit/issues/276

### Description
Just now falling back to GLIBC gnu when musl is not detected. 

### How was this PR tested?
Testing on the different OS to make sure nothing was broken
OS openSUSE-Leap-15 (works)
OS Fedora-Workstation-Live-x86_64-29-1.2 (works)
OS CentOS-7-x86_64 (works) - musl version detected & installed
OS manjaro-xfce-17.1.7-stable-x86_64 (works)

### Checklist
- [x] tested locally
- [ ] added unit tests
- [ ] included documentation

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -249,18 +249,18 @@ get_gnu_musl_glibc() {
   local _ldd_version
   local _glibc_version
   _ldd_version=$(ldd --version)
-  if [[ $_ldd_version =~ "GNU" ]]; then
+  if ldd --version 2>&1 | grep -Eq 'GNU'; then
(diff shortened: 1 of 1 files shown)


fmassot:
There is already a fallback on unknow gnu. The message error is here to tell that we don't find the arch from ldd. But this is ok.
Can we just avoid the 
```
sh: 252: [[: not found
sh: 259: [[: not found 
```
and print that we fallback to unknowngnu ?

PR #289, opened 2021-07-13
Fix patch

evanxg852000:
### Context / purpose
Patch for https://github.com/quickwit-inc/quickwit/issues/282

### Description
make sure we dont delete beyond root of the storage


examples/index_configs/wikipedia_index_config.json
```diff
@@ -13,7 +13,7 @@
         {
             "name": "url",
             "type": "text",
-            "indexed": "false",
```
fulmicoton, line 16: good catch

PR #290, opened 2021-07-13
Added some cargo meta information for publication

fulmicoton:


evanxg852000 (approved):
Shouldn't we delete the `quickwit-proto/README.md` then ?

Looks good and links are working

PR #299, opened 2021-07-14
Support for comment in Index Config

evanxg852000:
### Context / purpose
Closes https://github.com/quickwit-inc/quickwit/issues/298

### Description
added `json-comment-rs` to support comment in index config json format

### How was this PR tested?
run the commands and added updated to our integration tests

### Checklist
- [x] tested locally
- [ ] added unit tests
- [ ] included documentation


fmassot:
@evanxg852000 did you check the code of json_comments crate? I'm ok with PR if the code is "good" in this crate, can you have a look?

evanxg852000:
Yes, I read the code and it looked fine to me with the tests. There is also this `hjson-rust` crate based on serde, but it seems abandoned since the dependencies are old and not being updated.

fmassot:
Indeed, the code is clean and well commented.

PR #302, opened 2021-07-14
Throughput based on interval

evanxg852000:
### Context / purpose
Closes https://github.com/quickwit-inc/quickwit/issues/284

### Description
Implemented throughput calculation based on interval

### How was this PR tested?
run it and see behavior

### Checklist
- [x] tested locally
- [x] run it on larger dataset

### DIFF
--- a/quickwit-cli/src/lib.rs
+++ b/quickwit-cli/src/lib.rs
@@ -428,15 +428,13 @@ impl ThroughputCalculator {
         self.processed_bytes_values.pop_front();
         let current_instant = Instant::now();
         let (first_instant, first_processed_bytes) = *self.processed_bytes_values.front().unwrap();
-        let elapsed_time = (current_instant - first_instant).as_secs();
+        let elapsed_time = (current_instant - first_instant).as_millis() as f64 / 1_000f64;
         self.processed_bytes_values
             .push_back((current_instant, current_processed_bytes));
 
-        //TODO: what should we do when all docs are processed and `current_processed_bytes`
-        // is no longer increasing?
         (current_processed_bytes - first_processed_bytes) as f64
             / 1_000_000f64
-            / elapsed_time.max(1) as f64
+            / elapsed_time.max(1f64) as f64
     }
 }
 
--- a/quickwit-cli/src/lib.rs
+++ b/quickwit-cli/src/lib.rs
@@ -37,6 +37,7 @@ use quickwit_search::single_node_search;
 use quickwit_search::SearchResultJson;
 use quickwit_storage::StorageUriResolver;
 use quickwit_telemetry::payload::TelemetryEvent;
+use std::collections::VecDeque;
 use std::env;
 use std::io;
 use std::io::Stdout;
@@ -57,8 +58,8 @@ use tracing::debug;
 
 use quickwit_core::{create_index, delete_index, index_data, IndexDataParams, IndexingStatistics};
 
-/// Throughput calculation window size in seconds.
-const THROUGHPUT_WINDOW_SIZE: u64 = 5;
+/// Throughput calculation window size.
+const THROUGHPUT_WINDOW_SIZE: usize = 5;
 
 #[derive(Debug)]
 pub struct CreateIndexArgs {
@@ -310,8 +311,7 @@ pub async fn start_statistics_reporting(
         let mut stdout_handle = stdout();
         let start_time = Instant::now();
         let is_tty = atty::is(atty::Stream::Stdout);
-        let mut throughput_calculator =
-            ThroughputCalculator::new(start_time, THROUGHPUT_WINDOW_SIZE);
+        let mut throughput_calculator = ThroughputCalculator::new(start_time);
         loop {
             // Try to receive with a timeout of 1 second.
             // 1 second is also the frequency at which we update statistic in the console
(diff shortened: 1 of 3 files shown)


quickwit-cli/src/lib.rs
```diff
@@ -391,6 +405,53 @@ fn display_statistics(
     Ok(())
 }
 
+/// ThroughputCalculator is used to calculate throughput on time interval.
+struct ThroughputCalculator {
+    /// clock for tracking the time
+    clock: Instant,
+    /// How long it takes before calculating the throughput.
+    window_size: u64,
```
fulmicoton, line 413: You could make that kind of stuff a const and not have to check for it being 0.
Less code and less trouble.

In practise it will never be configured.

quickwit-cli/src/lib.rs
```diff
@@ -391,6 +405,53 @@ fn display_statistics(
...
+
+        // Only recompute if we have reached the end of a window
+        // and the number of processed byte has changed.
+        if elapsed_secs % self.window_size == 0
+            && self.processed_bytes_from_last_window != processed_bytes
+        {
+            let window_processed_bytes = processed_bytes - self.processed_bytes_from_last_window;
+            self.throughput =
+                window_processed_bytes as f64 / 1_000_000f64 / self.window_size as f64;
+            self.processed_bytes_from_last_window = processed_bytes;
```
fulmicoton, line 443: There is a complicated big here.

If for one reason or another elapsed_secs skips 8s. For instance... It was 7s30 and the next tick it is 9s due to some lag in the terminal output.
Another possibility is we changed the refresh interval in the terminal.

What will happen?
Or maybe 
What happens then?

quickwit-cli/src/lib.rs
```diff
@@ -391,6 +405,53 @@ fn display_statistics(
...
+            clock,
+            window_size,
+            processed_bytes_from_last_window: 0,
+            throughput: 0f64,
+        }
+    }
+
+    /// Calculates the throughput.
+    pub fn calculate(&mut self, processed_bytes: usize) -> f64 {
+        let elapsed_secs = self.clock.elapsed().as_secs();
```
fulmicoton, line 433: I was hoping for rolling windows but I guess this is ok.

evanxg852000: What do you mean by rolling windows?

fulmicoton: I think sliding window is a more common term? (or maybe they don't mean exactly the same I am unsure.)

Anyway, right now your solution updates the value at a period that is equal to the window size. That's not great for the user.
After all, the actually value is defined for every single line of logs.

```
(number of bytes(t_n)  -  number of bytes(t_n-8) / (instant(t) - instant(t_n-8)))
```

quickwit-cli/src/lib.rs
```diff
@@ -393,6 +407,39 @@ fn display_statistics(
...
+    /// Calculates the throughput.
+    pub fn calculate(&mut self, current_processed_bytes: usize) -> f64 {
+        self.processed_bytes_values.pop_front();
+        let current_instant = Instant::now();
+        let (first_instant, first_processed_bytes) = *self.processed_bytes_values.front().unwrap();
+        let elapsed_time = (current_instant - first_instant).as_secs();
+        self.processed_bytes_values
+            .push_back((current_instant, current_processed_bytes));
+
+        //TODO: what should we do when all docs are processed and `current_processed_bytes`
```
evanxg852000, line 435: @guilload  I need your opinion here. 
So the `processed_bytes` is the number or byte ingested from documents when indexing. From the point the last document is indexed to the time the last split is uploaded, the `processed_bytes` no longer increases making the throughput drop gradually to zero. 
is this drop ok to show to the user or could we just display that last valid throughput value for the rest of the time?

quickwit-cli/src/lib.rs
```diff
@@ -393,6 +407,39 @@ fn display_statistics(
...
+            processed_bytes_values,
+        }
+    }
+
+    /// Calculates the throughput.
+    pub fn calculate(&mut self, current_processed_bytes: usize) -> f64 {
+        self.processed_bytes_values.pop_front();
+        let current_instant = Instant::now();
+        let (first_instant, first_processed_bytes) = *self.processed_bytes_values.front().unwrap();
+        let elapsed_time = (current_instant - first_instant).as_secs();
```
fulmicoton, line 431: The result here is equivalent to rounding to the closest second.
The time interval thing is not necessarily super accurate as it depends on the OS and tokio's scheduler, so it is better to work with a better precision.
`as_millis()` + cast as float and and divide the result by 1_000 should do the trick.

quickwit-cli/src/lib.rs
fulmicoton: This is a good thing. That's what the  user needs to see.

guilload: Yes, that's fine if we display that last valid throughput value after the last document has been ingested. However, to Paul's point, we need to display something else to inform the user that the program is actually doing something and not hanging.

fulmicoton (approved):

quickwit-cli/src/lib.rs
evanxg852000: Trying to come up with another metric, I have added the elapsed_time for now since that one changes. 
```bash
quickwit index --index-uri 'file://home/evance/quickwit/data' --input-path ./zero/wiki-articles-10000.json
Num docs:    9970 Parse errs:     0 Staged splits:   0 Input size:    11MB Thrghput: 11.18MB/s Time: 00:00:01
Num docs:   10000 Parse errs:     0 Staged splits:   0 Input size:    11MB Thrghput:  5.62MB/s Time: 00:00:02
Num docs:   10000 Parse errs:     0 Staged splits:   0 Input size:    11MB Thrghput:  3.74MB/s Time: 00:00:03
Num docs:   10000 Parse errs:     0 Staged splits:   0 Input size:    11MB Thrghput:  2.81MB/s Time: 00:00:04
Num docs:   10000 Parse errs:     0 Staged splits:   0 Input size:    11MB Thrghput:  0.02MB/s Time: 00:00:05
Num docs:   10000 Parse errs:     0 Staged splits:   1 Input size:    11MB Thrghput:  0.00MB/s Time: 00:00:05

Indexed 10000 documents in 5s
You can now query your index with `quickwit search --index-uri file://home/evance/quickwit/data --query "barack obama"`
```

fulmicoton (approved):

PR #309, opened 2021-07-15
Add gc (garbage collect) command

evanxg852000:
### Context / purpose
Closes https://github.com/quickwit-inc/quickwit/issues/207

### Description
Add `gc` command that removes old staged and deleted split files from the index.

### How was this PR tested?
- Changing manually some split state from the metastore file and running the `gc` command.

### Checklist
- [x] tested locally
- [x] updated tests
- [ ] included documentation (todo in https://github.com/quickwit-inc/quickwit/issues/312)

### DIFF
--- a/quickwit-cli/src/lib.rs
+++ b/quickwit-cli/src/lib.rs
@@ -54,7 +54,9 @@ use tokio::time::timeout;
 use tokio::try_join;
 use tracing::debug;
 
-use quickwit_core::{create_index, delete_index, index_data, IndexDataParams, IndexingStatistics};
+use quickwit_core::{
+    clean_index, create_index, delete_index, index_data, IndexDataParams, IndexingStatistics,
+};
 
 #[derive(Debug)]
 pub struct CreateIndexArgs {
@@ -115,6 +117,11 @@ pub struct SearchIndexArgs {
 pub struct DeleteIndexArgs {
     pub index_uri: String,
     pub dry_run: bool,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct CleanIndexArgs {
+    pub index_uri: String,
 }
 
 pub async fn create_index_cli(args: CreateIndexArgs) -> anyhow::Result<()> {
@@ -285,13 +292,38 @@ pub async fn delete_index_cli(args: DeleteIndexArgs) -> anyhow::Result<()> {
             "The following files will be removed along with the index at `{}`",
             args.index_uri
         );
-        for file in affected_files {
-            println!(" - {}", file.display());
+        for file_entry in affected_files {
+            println!(" - {}", file_entry.file_name);
         }
         return Ok(());
     }
 
     println!("Index successfully deleted at `{}`", args.index_uri);
+    Ok(())
+}
+
+pub async fn clean_index_cli(args: CleanIndexArgs) -> anyhow::Result<()> {
+    debug!(
+        index_uri = %args.index_uri,
+        "clean-index"
+    );
+    quickwit_telemetry::send_telemetry_event(TelemetryEvent::Clean).await;
+
+    let (metastore_uri, index_id) =
+        extract_metastore_uri_and_index_id_from_index_uri(&args.index_uri)?;
+    let affected_files = clean_index(metastore_uri, index_id).await?;
+
+    let deleted_bytes: u64 = affected_files
+        .iter()
+        .map(|entry| entry.file_size_in_bytes)
+        .sum();
+
+    if deleted_bytes > 0 {
+        println!("{}MB of storage reclaimed.", deleted_bytes / 1_000_000);
+    } else {
+        println!("No dangling files to reclaime.");
+    }
+    println!("Index successfully cleaned at `{}`", args.index_uri);
     Ok(())
 }
 
--- a/quickwit-cli/src/lib.rs
+++ b/quickwit-cli/src/lib.rs
@@ -55,7 +55,8 @@ use tokio::try_join;
 use tracing::debug;
 
 use quickwit_core::{
-    clean_index, create_index, delete_index, index_data, IndexDataParams, IndexingStatistics,
+    create_index, delete_index, garbage_collect_index, index_data, IndexDataParams,
+    IndexingStatistics,
 };
 
 #[derive(Debug)]
@@ -120,7 +121,7 @@ pub struct DeleteIndexArgs {
 }
 
 #[derive(Debug, PartialEq, Eq)]
-pub struct CleanIndexArgs {
+pub struct GarbageCollectIndexArgs {
     pub index_uri: String,
 }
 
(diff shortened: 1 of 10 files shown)


quickwit-cli/src/lib.rs
```diff
@@ -293,6 +301,37 @@ pub async fn delete_index_cli(args: DeleteIndexArgs) -> anyhow::Result<()> {
...
+    let (metastore_uri, index_id) =
+        extract_metastore_uri_and_index_id_from_index_uri(&args.index_uri)?;
+    let affected_files = garbage_collect_index(metastore_uri, index_id).await?;
+
+    let deleted_bytes: u64 = affected_files
+        .iter()
+        .map(|entry| entry.file_size_in_bytes)
+        .sum();
+
+    if deleted_bytes > 0 {
```
fulmicoton, line 320: Can we do the test on affect_files being empty or not? (I'm not sure if files with 0 byte are allowed or not on S3 but it seems more "logical")

quickwit-cli/src/lib.rs
```diff
@@ -293,6 +301,37 @@ pub async fn delete_index_cli(args: DeleteIndexArgs) -> anyhow::Result<()> {
...
+pub async fn garbage_collect_index_cli(args: GarbageCollectIndexArgs) -> anyhow::Result<()> {
+    debug!(
+        index_uri = %args.index_uri,
+        "garbage-collect-index"
+    );
+    quickwit_telemetry::send_telemetry_event(TelemetryEvent::GarbageCollect).await;
+
+    let (metastore_uri, index_id) =
+        extract_metastore_uri_and_index_id_from_index_uri(&args.index_uri)?;
+    let affected_files = garbage_collect_index(metastore_uri, index_id).await?;
```
fulmicoton, line 313: Should we rename `deleted_files` or is it more subtle than this?

quickwit-core/src/index.rs
```diff
@@ -99,11 +98,42 @@ pub async fn delete_index(
...
+    index_id: &str,
+) -> anyhow::Result<Vec<FileEntry>> {
+    let metastore = MetastoreUriResolver::default()
+        .resolve(&metastore_uri)
+        .await?;
+    let storage_resolver = StorageUriResolver::default();
+
+    // schedule all staged splits for delete
+    let staged_splits = metastore
+        .list_splits(index_id, SplitState::Staged, None)
```
fulmicoton, line 125: I wonder if we should garbage collect "staged + older than 1h", just to be sure we are not removing staged splits that are about to be uploaded...

Right now our metastore is not concurrent, but it will eventually be.

quickwit-core/src/index.rs
```diff
@@ -99,11 +98,42 @@ pub async fn delete_index(
...
+        .await?;
+    let split_ids = staged_splits
+        .iter()
+        .map(|split_meta| split_meta.split_id.as_str())
+        .collect::<Vec<_>>();
+    metastore
```
fulmicoton, line 131: Good job nailing the weird logic.

quickwit-core/src/index.rs
```diff
@@ -134,23 +164,29 @@ pub async fn garbage_collect(
...
 
-    let mut files = vec![];
+    let mut file_entries = vec![];
     while let Some(delete_result) = delete_stream.next().await {
         let deleted_files = delete_result.map_err(|error| {
             warn!("Some split files were not deleted.");
             error
         })?;
-        files.extend(deleted_files);
+        file_entries.extend(deleted_files);
```
fulmicoton, line 173: good idea

quickwit-core/src/index.rs
```diff
@@ -99,11 +98,42 @@ pub async fn delete_index(
...
+        .collect::<Vec<_>>();
+    metastore
+        .mark_splits_as_deleted(index_id, split_ids)
+        .await?;
+
+    let file_entries = garbage_remove(metastore.as_ref(), index_id, storage_resolver).await?;
+    Ok(file_entries)
 }
 
 /// Removes all danglings files from an index specified at `index_uri`.
```
fulmicoton, line 139: Garbage_remove and garbage_collect do very different things, yet the name nor the comments are helping understand the difference.

fulmicoton (changes requested):
See comments inline.

quickwit-cli/tests/cli.rs
```diff
@@ -217,6 +217,13 @@ fn test_cmd_delete() -> Result<()> {
         test_env.resource_files["logs"].as_path(),
     );
 
+    make_command(format!("gc --index-uri {}", test_env.index_uri).as_str())
```
fulmicoton, line 220: Can we add a test case to check one case where files are indeed deleted?

quickwit-cli/src/lib.rs
evanxg852000: No, it's not especially since we don't have `dry-run` like `delete` command

quickwit-core/src/index.rs
evanxg852000: Indeed you are right, I think we need to have a concept of `updated_at` timestamp on the split to check when was the status updated and ignore the new ones

guilload:
S3 does allow objects with size 0.

Should all destructive commands (delete, gc, ...) have a `dry-run` option?

evanxg852000:
I thought garbage collection is not really a destructive command per see. Since it's supposed to clean the index

quickwit-core/src/index.rs
fulmicoton: that sounds good yes. Can you open an issue and take over that?

evanxg852000: I was planning to handle it here, but yes separating always gives more context to the ticket. I will add test and clean do we can close this. I already added TODO

fulmicoton:
@evanxg852000 Your point is good. A dry-run can still be a nice to have for a scared user or for debugging but yes, it is not as critical as for delete. On the same spirit, I think we do not show the list file that are being GCed? It could be nice.

quickwit-core/src/indexing/split.rs
```diff
@@ -329,6 +330,10 @@ async fn put_split_files_to_storage(
     Ok(manifest)
 }
 
+/// [`FileEntry`] is an alias of [`ManifestEntry`] for
+/// holding the full path & size of a file.
+pub type FileEntry = ManifestEntry;
```
fmassot, line 335: Why not renaming `ManifestEntry` with `FileEntry` which is definitely better?

fmassot (approved):
Nice work Evance. I add small comments that are not a priority, I think the renaming would be good but this can be done later.

quickwit-cli/tests/cli.rs
```diff
@@ -225,6 +234,61 @@ fn test_cmd_delete() -> Result<()> {
     Ok(())
 }
 
+#[tokio::test]
+async fn test_cmd_garbae_collect() -> Result<()> {
```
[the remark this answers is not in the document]
evanxg852000: Good catch

quickwit-cli/tests/cli.rs
```diff
@@ -225,6 +234,61 @@ fn test_cmd_delete() -> Result<()> {
...
+        .mark_splits_as_deleted(index_id, split_ids)
+        .await?;
+    make_command(format!("gc --index-uri {} --dry-run", test_env.index_uri).as_str())
+        .assert()
+        .success()
+        .stdout(predicate::str::contains(
+            "The following files will be garbage collected.",
+        ))
+        .stdout(predicate::str::contains("/hotcache"))
+        .stdout(predicate::str::contains("/.manifest"));
```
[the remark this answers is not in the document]
evanxg852000: This is just checking if the `dry-run` listing contains few files we know exist in the split.
we could construct the list of split files since we know the split_id but I was keeping it simple.
The split deletion is & checking happens in the next lines. I will add a check to make sure the split folder is actually deleted.

quickwit-core/src/indexing/split.rs
evanxg852000: It's not the same purpose, though the type structure is the same. 
- ManifestEntry is used in manifest and hold a relative path
- FileEntry is used for the user interface and holds the full path
I could be wrong but that was my thought process.

fmassot: Make sense but it's hard to guess this difference when looking at ManifestEntry struct. For me, it means that we can just rename it.
Generally speaking, I don't really like aliasing because it can make me believe I'm manipulating different things but it's not the case.

fmassot (approved):

PR #315, opened 2021-07-16
Docker image build & push

evanxg852000:
### Context / purpose
Closes https://github.com/quickwit-inc/quickwit/issues/270

### Description
Add workflow to build & push docker image to docker hub

### How was this PR tested?
- Make and publish a release
- Pull and run the image

### Checklist
- [x] tested locally
- [ ] added unit tests
- [x] included documentation

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,14 +1,14 @@
 FROM alpine:3.14
 LABEL maintainer="Quickwit, Inc. <hello@quickwit.io>"
 
-# Install tool
+# Install tools
 RUN apk add --no-cache curl
 
-# Create working directory
-RUN mkdir -p /usr/quickwit
+# Install Quickwit
+RUN curl -L https://install.quickwit.io | sh
+RUN mv ./quickwit /usr/local/bin/quickwit
+
 WORKDIR /usr/quickwit
 EXPOSE 8080
 
-RUN curl -L https://install.quickwit.io | sh
-
-ENTRYPOINT [ "ash" ]
+ENTRYPOINT [ "quickwit" ]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,14 @@
+FROM alpine:3.14
+LABEL maintainer="Quickwit, Inc. <hello@quickwit.io>"
+
+# Install tool
+RUN apk add --no-cache curl
+
+# Create working directory
+RUN mkdir -p /usr/quickwit
+WORKDIR /usr/quickwit
+EXPOSE 8080
+
+RUN curl -L https://install.quickwit.io | sh
+
+ENTRYPOINT [ "ash" ]
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -38,8 +38,17 @@ cargo install quickwit-cli
 
 ## Use the docker image
 
-Coming soon.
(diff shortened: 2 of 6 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+# Install tool
+RUN apk add --no-cache curl
+
+# Create working directory
+RUN mkdir -p /usr/quickwit
+WORKDIR /usr/quickwit
+EXPOSE 8080
+
+RUN curl -L https://install.quickwit.io | sh
```
PSeitz, line 12: adding quickwit to the path, so the user can just type `quickwit`

docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM alpine:3.14
+LABEL maintainer="Quickwit, Inc. <hello@quickwit.io>"
+
+# Install tool
+RUN apk add --no-cache curl
+
+# Create working directory
+RUN mkdir -p /usr/quickwit
```
guilload, line 8: `WORKDIR` will create the directory for you.

guilload: I believe the Linux convention for installing third-party software is to use the `/opt` folder. However, in this case, since we only copy a binary, let's copy the file directly in `/usr/local/bin`.

docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+WORKDIR /usr/quickwit
+EXPOSE 8080
+
+RUN curl -L https://install.quickwit.io | sh
+
+ENTRYPOINT [ "ash" ]
```
guilload, line 14: Let's use `quickwit` as the entry point.

docs/getting-started/installation.md
```diff
@@ -38,8 +38,17 @@ cargo install quickwit-cli
 
 ## Use the docker image
 
-Coming soon.
+If you use docker, this might be one of the quickest way to get going. 
+The following command will pull the image from [dockerhub](https://hub.docker.com/r/quickwitinc/quickwit)
```
guilload, line 42: I own the `quickwit` namespace on DockerHub. I just gave @fulmicoton access to the account.

docs/getting-started/installation.md
```diff
@@ -38,8 +38,17 @@ cargo install quickwit-cli
...
+and gets you right in the running container ready to execute Quickwit commands.
 
+```bash
+docker run -it quickwitinc/quickwit
```
guilload, line 46: This example needs to be improved so that users know how to mount a local directory into the image and can start indexing and querying indexes.

guilload (changes requested):

docs/getting-started/installation.md
evanxg852000: Ok cool, then we can just update the needed credentials on github environment so that we can push the image to that channel.

docs/getting-started/installation.md
evanxg852000: Yes definitely if we are using quickwit as entrypoint

evanxg852000:
@PSeitz @fmassot @guilload  I have applied the requested changes with the docs

guilload:
Thanks for adding the extensive documentation. You can go ahead and merge this PR. Then, I'll take of migrating the image to the `quickwit` Docker Hub account.

guilload (approved):

PR #317, opened 2021-07-17
Some fix of comments

djKooks:
### Context / purpose
Fix some typo inside comments/docs

### Description
Hello,
This PR includes several fix of comments/docs I found while reviewing the code.
Please kindly check 🙏

### DIFF
--- a/quickwit-core/src/indexing/split.rs
+++ b/quickwit-core/src/indexing/split.rs
@@ -88,7 +88,7 @@ impl fmt::Debug for Split {
     }
 }
 
-/// returns true iff merge is required to reach a state where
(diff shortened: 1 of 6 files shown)


quickwit-core/src/indexing/split.rs
```diff
@@ -88,7 +88,7 @@ impl fmt::Debug for Split {
     }
 }
 
-/// returns true iff merge is required to reach a state where
+/// returns true if merge is required to reach a state where
```
fulmicoton, line 91: Iff is short for "if and only if".

fulmicoton:
Thank you @djKooks!

PR #323, opened 2021-07-20
New search service client and unit test for root_search

### LINKED ISSUE #295: Add unit tests for the root logic
We don't have any unit test for the root logic.
In barrel, it was done by injecting mock search-service in the pool.

We could do something similar.
It might require some juggling with the SearchServiceClient type

mosuka:
### Context / purpose
Closes #295 

### Description
The external world has been modified so that only `quickwit_search::SearchServiceClient` and `quickwit_search::SearchService` need to be manipulated.
- `WrappedSearchServiceClient` will be changed to `SearchServiceClient`. It embeds a `SocketAddr`, as was done for the `WrappedSearchServiceClient` in the previous implementation.
- It allows for calling directly a SearchService trait implementation, which is useful both for mocking but also can be used to bypass the transport when root node equals leaf node.
- It does the job of wrapping the `tonic::Request` and parsing the `SearchError` automatically, simplifying the root search code and the mocking.
- Add unit tests for the `root_search` logic.

### How was this PR tested?
Run the unit tests.

```
$ cargo test -- --nocapture test_root_search_single_split
$ cargo test -- --nocapture test_root_search_multiple_splits
```

### Checklist
- [x] tested locally
- [x] added unit tests
- [x] included documentation


quickwit-search/src/client_pool/search_client_pool.rs
```diff
@@ -32,26 +32,59 @@ use tracing::*;
 
 use quickwit_cluster::cluster::Cluster;
 
-use crate::client::{create_search_service_client, WrappedSearchServiceClient};
+use crate::client::create_search_service_client;
 use crate::client_pool::{ClientPool, Job};
 use crate::rendezvous_hasher::{sort_by_rendez_vous_hash, Node};
 use crate::swim_addr_to_grpc_addr;
+use crate::SearchServiceClient;
+// use crate::SearchService;
```
fulmicoton, line 40: remove commented out line.

quickwit-search/src/root.rs
```diff
@@ -232,9 +229,9 @@ pub async fn root_search(
                 debug!(leaf_search_result=?leaf_search_result, "Leaf search result.");
                 leaf_search_results.push(leaf_search_result)
             }
-            Err(grpc_error) => {
-                let leaf_search_error = parse_grpc_error(&grpc_error.status);
-                error!(error=?grpc_error, "Leaf request failed");
+            Err(node_search_error) => {
+                let leaf_search_error = node_search_error.search_error.clone();
```
fulmicoton, line 233: is cloning actually required?

quickwit-search/src/root.rs
```diff
@@ -269,12 +266,8 @@ pub async fn root_search(
...
                 let handle = tokio::spawn(async move {
-                    match search_client_clone
-                        .client()
-                        .fetch_docs(fetch_docs_request)
-                        .await
-                    {
-                        Ok(resp) => Ok(resp.into_inner()),
+                    match search_client_clone.fetch_docs(fetch_docs_request).await {
+                        Ok(resp) => Ok(resp),
                         Err(err) => Err(anyhow::anyhow!("Failed to fetch docs due to {:?}", err)),
```
fulmicoton, line 271: Why do we convert the error here? Wouldn't it be better to not convert the error and change the error type on line 258?

quickwit-cluster/src/cluster.rs
```diff
@@ -88,8 +91,24 @@ pub struct Member {
     pub is_self: bool,
 }
 
+/// Cluster meant to manage Quickwit's cluster formation and its events.
+#[async_trait]
+pub trait Cluster: Send + Sync + 'static {
```
fulmicoton, line 96: What is the point of this cluster trait (as opposed to just having a `EpidemicCluster` implementing these public methods?

fulmicoton (changes requested):
Minor changes requested

quickwit-cluster/src/cluster.rs
```diff
@@ -88,8 +91,24 @@ pub struct Member {
...
+    /// Return cluster members.
+    fn members(&self) -> Vec<Member>;
+
+    /// Return cluster member change watcher.
+    fn member_change_watcher(&self) -> WatchStream<Vec<Member>>;
+
+    /// Specify the address of a running node and join the cluster to which the node belongs.
+    fn add_peer_node(&self, peer_addr: SocketAddr);
+
+    /// Leave the cluster it is joining in.
```
fulmicoton, line 106: I don't think this is correct english.
`Leaves the cluster. `

quickwit-search/src/root.rs
mosuka: This cloning was not required.

quickwit-search/src/root.rs
mosuka: Yes, indeed. I will change the error type.

quickwit-cluster/src/cluster.rs
mosuka: At the moment it only supports cluster formation using Artillery, but I thought it would be useful in the future to have a trait that would make it easy to use different cluster formations.

PSeitz: This may be controversial, but I usually don't add abstractions if not necessary, since working with abstractions lowers readability imo.

quickwit-cluster/src/cluster.rs
```diff
@@ -98,9 +117,12 @@ pub struct Cluster {
 
     /// A receiver(channel) for exchanging members in a cluster.
     members: watch::Receiver<Vec<Member>>,
+
+    /// A stop flag of cluster monitoring task.
```
PSeitz, line 121: some info when and why this is used would be good

quickwit-cluster/src/cluster.rs
```diff
@@ -307,6 +351,96 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_cluster_multiple_node() -> anyhow::Result<()> {
+        let tmp_dir = tempfile::tempdir()?;
+
+        let host_key_path1 = tmp_dir.path().join("host_key1");
+        let host_key1 = read_host_key(host_key_path1.as_path())?;
+        let tmp_port1 = available_port()?;
+        let addr_str1 = format!("127.0.0.1:{}", tmp_port1);
+        let remote_host1: SocketAddr = addr_str1.as_str().parse()?;
+        let cluster1 = EpidemicCluster::new(host_key1, remote_host1)?;
```
PSeitz, line 390: this should be moved to a helper function

quickwit-search/src/client.rs
```diff
@@ -19,39 +19,132 @@
...
 
-use quickwit_proto::search_service_client::SearchServiceClient;
+use crate::error::parse_grpc_error;
+use crate::SearchError;
+use crate::SearchService;
 
-#[derive(Debug, Clone)]
-pub struct WrappedSearchServiceClient {
-    client: SearchServiceClient<Channel>,
+/// Impl is an enumlation that meant to manage Quickwit's search service client types.
```
PSeitz, line 35: enumlation = enumeration?

quickwit-search/src/client.rs
```diff
@@ -19,39 +19,132 @@
...
+#[derive(Clone)]
+enum Impl {
```
PSeitz, line 37: Impl is a too generic name

quickwit-search/src/client.rs
```diff
@@ -19,39 +19,132 @@
...
+    Local(Arc<dyn SearchService>),
+    Grpc(quickwit_proto::search_service_client::SearchServiceClient<Channel>),
```
PSeitz, line 39: Why don't we always use grpc, even on self?

quickwit-cluster/src/cluster.rs
mosuka: @fulmicoton @PSeitz Thanks. I'll revert it.

quickwit-search/src/client_pool/search_client_pool.rs
```diff
@@ -32,26 +32,58 @@ use tracing::*;
...
+        for member in cluster.members() {
+            let grpc_addr = swim_addr_to_grpc_addr(member.listen_addr);
+            match create_search_service_client(grpc_addr).await {
+                Ok(client) => {
+                    debug!(grpc_addr=?grpc_addr, "Add a new client to connect to the members of the cluster.");
+                    clients.insert(grpc_addr, client);
+                }
+                Err(err) => {
+                    error!(grpc_addr=?grpc_addr, err=?err, "Failed to create search client.")
+                }
```
PSeitz, line 84: This may not be the best example, but we should think about feature flags, where to add them and how to control them.

E.g. here I could image a setting that optionally ignores errors
(should setting up a cluster fail, because the client creation for one member failed?)

quickwit-search/src/client_pool/search_client_pool.rs
```diff
@@ -177,6 +208,185 @@ impl ClientPool for SearchClientPool {
...
+    #[tokio::test]
+    async fn test_search_client_pool_single_node() -> anyhow::Result<()> {
+        let tmp_dir = tempfile::tempdir()?;
+
+        let host_key_path = tmp_dir.path().join("host_key");
+        let host_key = read_host_key(host_key_path.as_path())?;
+        let tmp_swim_port = available_port()?;
+        let swim_addr_string = format!("127.0.0.1:{}", tmp_swim_port);
+        let swim_addr = to_socket_addr(&swim_addr_string)?;
+        let cluster = Arc::new(EpidemicCluster::new(host_key, swim_addr)?);
```
PSeitz, line 240: helper function

quickwit-search/src/client.rs
mosuka: oh, thanks!

quickwit-search/src/client.rs
mosuka: @PSeitz 
How about the name `SearchServiceClientImpl`?

PSeitz: yes that's better

quickwit-cluster/src/cluster.rs
mosuka: I changed the code a bit. Unnecessary variables have been removed.
It's not a helper function, but these variables are used in the assertions that follow. How about this one?

quickwit-search/src/client_pool/search_client_pool.rs
mosuka: I changed the code a bit. Unnecessary variables have been removed.
It's not a helper function, but these variables are used in the assertions that follow. How about this one?

quickwit-cluster/src/cluster.rs
```diff
@@ -363,28 +363,22 @@ mod tests {
...
-        let host_key3 = read_host_key(host_key_path3.as_path())?;
-        let tmp_port3 = available_port()?;
-        let addr_str3 = format!("127.0.0.1:{}", tmp_port3);
-        let remote_host3: SocketAddr = addr_str3.as_str().parse()?;
-        let cluster3 = Cluster::new(host_key3, remote_host3)?;
-        cluster3.add_peer_node(remote_host1);
+        let host_key1 = read_host_key(tmp_dir.path().join("host_key1").as_path())?;
+        let listen_addr1 =
+            SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), available_port()?);
+        let cluster1 = Arc::new(Cluster::new(host_key1, listen_addr1)?);
```
PSeitz, line 369: I still think this should be a utility function returning a test cluster, the listen address can be accesses via the the cluster object

mosuka: Thanks, I added a test_utils module.
Please take another look.
https://github.com/quickwit-inc/quickwit/pull/323/commits/5c2953f024e7e81b81b83e6409133fcc67f926f4

quickwit-search/src/client.rs
mosuka: It allows for calling directly a `SearchService` trait implementation, which is useful both for mocking but also can be used to bypass the transport when root node equals leaf node.
Right now the client uses this for `MockSearchService` for testing, but in the future, it will be possible to use the service directly without going through gRPC for leaf search to itself like an Elasticsearch.

quickwit-search/src/client_pool/search_client_pool.rs
mosuka: Indeed. I hadn't thought about control by feature flags yet.
I guess I need to think about what I want to control with feature flags.

quickwit-search/src/client.rs
PSeitz: Can't we mock the target node when using grpc?

PSeitz (approved):

quickwit-search/src/client.rs
fulmicoton: And mock transport? or you'd open a new server for every mock?

quickwit-search/src/client_pool/search_client_pool.rs
fulmicoton: I don't think this sounds like a feature flag setting, but more like something that is dynamic.

It is per-case discussion and it is difficult to judge, but in this specific case, I think one default behavior should fit
everyone 's liking.

quickwit-search/src/error.rs
```diff
@@ -18,16 +18,17 @@
...
 use tokio::task::JoinError;
 
+use quickwit_index_config::QueryParserError;
+use quickwit_metastore::MetastoreError;
+use quickwit_storage::StorageResolverError;
+
 /// Possible SearchError
 #[allow(missing_docs)]
-#[derive(Error, Debug, Serialize, Deserialize)]
+#[derive(Clone, Error, Debug, Serialize, Deserialize)]
```
fulmicoton, line 31: do we still need the clone?

fulmicoton (approved):

quickwit-search/src/error.rs
mosuka: Sorry, I forgot to put it back.

quickwit-search/src/client_pool/search_client_pool.rs
PSeitz: I think it would be better to remove the node and fail if the error rate is too high.

After a certain cluster size (what's the target cluster size btw) and client failure rate too many requests could fail.

fulmicoton: Absolutely. Although keep in mind this happens if the node is considered "healthy" by the SWIM cluster, but we somehow cannot connect to it.

(We could see that if some firewall refuses TCP connection but accepts UDP connection for instance... But it is not a super pathological case)

PR #326, opened 2021-07-21
Add garce-period flag to gc command

evanxg852000:
### Context / purpose
Closes https://github.com/quickwit-inc/quickwit/issues/320

### Description
- Added `update_timestamp`  member to `SplitMetadata` for tracking when the split was last updated in the metastore
- Added a `grace-period` flag to the `gc` command (default is `1h` 1 hour)
- Supported grace-period format: `[0-9]+[smhd]`
  - 2s -> 2 seconds (Not sure allowing seconds, but it was useful for testing)
  - 2m -> 2 minutes
  - 2h -> 2 hours
  - 2d -> 2 days

### How was this PR tested?
Run the gc command while poking around the metastore file
`./quickwit gc --index-uri file://home/evance/quickwit/data --grace-period 10m` 

### Checklist
- [x] tested locally
- [x] added unit tests
- [ ] included documentation (https://github.com/quickwit-inc/quickwit/issues/312)

### DIFF
--- a/quickwit-cli/src/cli.yaml
+++ b/quickwit-cli/src/cli.yaml
@@ -93,7 +93,7 @@ subcommands:
                 value_name: INDEX URI
                 required: true
             - grace-period:
-                help: Threshold period after which an intermediate file can be collected.
+                help: Threshold period after which intermediate files can be garbage collected.
                 long: grace-period
                 value_name: GRACE PERIOD
                 default_value: '1h'
--- a/quickwit-cli/src/cli.yaml
+++ b/quickwit-cli/src/cli.yaml
@@ -92,6 +92,11 @@ subcommands:
                 long: index-uri
                 value_name: INDEX URI
                 required: true
+            - grace-period:
+                help: Threshold period after which an intermediate file can be collected.
+                long: grace-period
+                value_name: GRACE PERIOD
+                default_value: '1h'
             - dry-run:
                 help: Executes the command in dry run mode and displays the list of files to remove
                 long: dry-run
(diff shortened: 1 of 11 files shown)


quickwit-cli/src/cli.yaml
```diff
@@ -92,6 +92,11 @@ subcommands:
                 long: index-uri
                 value_name: INDEX URI
                 required: true
+            - grace-period:
+                help: Threshold period after which intermediate files can be garbage collected.
```
fulmicoton, line 96: i did not know the term. I suggest not making that configurable though. 
1h is a good one size fits all parameter.

quickwit-cli/src/main.rs
```diff
@@ -276,17 +282,43 @@ fn about_text() -> String {
     about_text
 }
 
+/// Parse duration with unit.
+/// examples: 1s 2m 3h 5d
+pub fn parse_duration_with_unit(duration: &str) -> anyhow::Result<Duration> {
```
fulmicoton, line 287: no need to make the grace period configurable, so we can remove this. Let's keep the snippet and the test though. I am sure it will be useful

fulmicoton (changes requested):
Can we have unit test to check that GC spares files that are within the grace period?
(and deletes the others)

quickwit-cli/src/main.rs
fulmicoton: Marking as resovled as apparently it was requested by Adrien and Frnacois

evanxg852000:
@fulmicoton  I have added another test to check that gc does indeed ignore stage splits that are still within the grace period.
We can still remove the `grace-period` flag less code is always better.

guilload:
Having the grace period not hardcoded allows for testing the gc locally without having to wait for an hour, potentially integration testing, and cleaning up right away when messing up with an index. I'd rather keep it.

fulmicoton (approved):

PR #327, opened 2021-07-21
add retry mechanism

PSeitz:
add retry mechnism for failed splits
Retry nodes are either all nodes in case all requests failed or only the (partial successfull) nodes.
When there are issue regarding caching, this retry logic may not work well. When a node partially fails, the split will choose the same node again, because the same algorithm (rendezvous hashing) is used.

### Context / purpose
Describe the context and purpose of this PR. Link related issue(s) if any.

### Description

### How was this PR tested?

### Checklist
- [x] tested locally
- [x] added unit tests
- [x] included documentation

### DIFF
--- a/quickwit-cli/tests/cli.rs
+++ b/quickwit-cli/tests/cli.rs
@@ -28,7 +28,7 @@ use crate::helpers::{create_test_env, make_command, spawn_command};
 use anyhow::Result;
 use helpers::{TestEnv, TestStorageType};
 use predicates::prelude::*;
-use quickwit_cli::{create_index_cli, search_index, CreateIndexArgs, SearchIndexArgs};
+use quickwit_cli::{create_index_cli, CreateIndexArgs};
 use quickwit_common::extract_metastore_uri_and_index_id_from_index_uri;
 use quickwit_metastore::{MetastoreUriResolver, SplitState};
 use quickwit_storage::{localstack_region, S3CompatibleObjectStorage, Storage};
@@ -589,41 +589,5 @@ async fn test_all_with_s3_localstack_internal_api() -> Result<()> {
     let metadata_file_exist = object_storage.exists(Path::new("quickwit.json")).await?;
     assert_eq!(metadata_file_exist, false);
 
-    Ok(())
-}
-
-// TODO add failure test
-#[tokio::test]
-async fn test_failure_not() -> Result<()> {
-    let test_env = create_test_env(TestStorageType::LocalFileSystem)?;
-    create_logs_index(&test_env);
-
-    index_data(
-        &test_env.index_uri,
-        test_env.resource_files["logs"].as_path(),
-    );
-
-    // using piped input.
-    let log_path = test_env.resource_files["logs"].clone();
-    make_command(format!("index --index-uri {} ", test_env.index_uri).as_str())
-        .pipe_stdin(log_path)?
-        .assert()
-        .success()
-        .stdout(predicate::str::contains("Indexed"))
-        .stdout(predicate::str::contains("documents in"))
-        .stdout(predicate::str::contains(
-            "You can now query your index with",
-        ));
-
-    let args = SearchIndexArgs {
-        index_uri: test_env.index_uri.to_string(),
-        query: "level:info".to_string(),
-        max_hits: 10,
-        start_offset: 0,
-        ..Default::default()
-    };
-    let search_result = search_index(args).await.unwrap();
-
-    assert_eq!(search_result.num_hits, 4);
     Ok(())
 }
--- a/quickwit-cli/tests/cli.rs
+++ b/quickwit-cli/tests/cli.rs
@@ -28,7 +28,7 @@ use crate::helpers::{create_test_env, make_command, spawn_command};
 use anyhow::Result;
 use helpers::{TestEnv, TestStorageType};
 use predicates::prelude::*;
-use quickwit_cli::{create_index_cli, CreateIndexArgs};
+use quickwit_cli::{create_index_cli, search_index, CreateIndexArgs, SearchIndexArgs};
 use quickwit_common::extract_metastore_uri_and_index_id_from_index_uri;
 use quickwit_metastore::{MetastoreUriResolver, SplitState};
 use quickwit_storage::{localstack_region, S3CompatibleObjectStorage, Storage};
@@ -589,5 +589,41 @@ async fn test_all_with_s3_localstack_internal_api() -> Result<()> {
     let metadata_file_exist = object_storage.exists(Path::new("quickwit.json")).await?;
     assert_eq!(metadata_file_exist, false);
 
+    Ok(())
+}
+
+// TODO add failure test
+#[tokio::test]
+async fn test_failure_not() -> Result<()> {
+    let test_env = create_test_env(TestStorageType::LocalFileSystem)?;
+    create_logs_index(&test_env);
+
+    index_data(
+        &test_env.index_uri,
+        test_env.resource_files["logs"].as_path(),
+    );
+
+    // using piped input.
+    let log_path = test_env.resource_files["logs"].clone();
+    make_command(format!("index --index-uri {} ", test_env.index_uri).as_str())
+        .pipe_stdin(log_path)?
+        .assert()
+        .success()
+        .stdout(predicate::str::contains("Indexed"))
+        .stdout(predicate::str::contains("documents in"))
+        .stdout(predicate::str::contains(
+            "You can now query your index with",
+        ));
+
+    let args = SearchIndexArgs {
+        index_uri: test_env.index_uri.to_string(),
+        query: "level:info".to_string(),
+        max_hits: 10,
+        start_offset: 0,
+        ..Default::default()
+    };
+    let search_result = search_index(args).await.unwrap();
+
+    assert_eq!(search_result.num_hits, 4);
     Ok(())
 }
--- a/quickwit-proto/proto/search_api.proto
+++ b/quickwit-proto/proto/search_api.proto
@@ -79,6 +79,10 @@ message SearchResult {
   // Elapsed time to perform the request. This time is measured
   // server-side and expressed in microseconds.
   uint64 elapsed_time_micros = 3;
+
+  // The searcherrors that occured formatted as string.
+  repeated string errors = 4;
+
 }
 
 message SplitSearchError {
@@ -87,6 +91,9 @@ message SplitSearchError {
 
   // Split id that failed. 
   string split_id = 2;
+
+  // Flag to indicate if the error can be considered a retryable error
+  bool retryable_error = 3;
 }
 
 message LeafSearchRequest {
--- a/quickwit-proto/proto/search_api.proto
+++ b/quickwit-proto/proto/search_api.proto
@@ -146,7 +146,7 @@ message LeafSearchResult {
   // List of the best top-K candidates for the given leaf query.
   repeated PartialHit partial_hits = 2;
 
-  // The list of requests that failed. LeafSearchResult can be an aggregation of results, so there may be multiple.
+  // The list of splits that failed. LeafSearchResult can be an aggregation of results, so there may be multiple.
   repeated SplitSearchError failed_splits = 3;
 
   // Total number of splits the leaf(s) were in charge of.
--- a/quickwit-proto/proto/search_api.proto
+++ b/quickwit-proto/proto/search_api.proto
@@ -147,7 +147,7 @@ message LeafSearchResult {
   repeated PartialHit partial_hits = 2;
 
   // The list of requests that failed. LeafSearchResult can be an aggregation of results, so there may be multiple.
-  repeated SplitSearchError failed_requests = 3;
+  repeated SplitSearchError failed_splits = 3;
 
   // Total number of splits the leaf(s) were in charge of.
   // num_attempted_splits = num_successful_splits + num_failed_splits.
--- a/quickwit-search/src/client_pool.rs
+++ b/quickwit-search/src/client_pool.rs
@@ -49,6 +49,6 @@ pub trait ClientPool: Send + Sync + 'static {
     async fn assign_jobs(
         &self,
         jobs: Vec<Job>,
-        exclude_addresses: &Option<HashSet<SocketAddr>>,
+        exclude_addresses: &HashSet<SocketAddr>,
     ) -> anyhow::Result<Vec<(SearchServiceClient, Vec<Job>)>>;
 }
--- a/quickwit-search/src/client_pool.rs
+++ b/quickwit-search/src/client_pool.rs
@@ -44,9 +44,11 @@ pub struct Job {
 pub trait ClientPool: Send + Sync + 'static {
     /// Assign the given job to the clients.
     /// Returns a list of pair (SocketAddr, Vec<Job>)
+    ///
+    /// When exclude_addresses filters all clients it is ignored.
     async fn assign_jobs(
         &self,
         jobs: Vec<Job>,
-        address_whitelist: Option<HashSet<SocketAddr>>,
+        exclude_addresses: Option<HashSet<SocketAddr>>,
     ) -> anyhow::Result<Vec<(SearchServiceClient, Vec<Job>)>>;
 }
--- a/quickwit-search/src/client_pool.rs
+++ b/quickwit-search/src/client_pool.rs
@@ -21,6 +21,8 @@
 
 pub mod search_client_pool;
 
+use std::{collections::HashSet, net::SocketAddr};
+
 use async_trait::async_trait;
 
 use crate::SearchServiceClient;
@@ -45,5 +47,6 @@ pub trait ClientPool: Send + Sync + 'static {
     async fn assign_jobs(
         &self,
         jobs: Vec<Job>,
+        address_whitelist: Option<HashSet<SocketAddr>>,
     ) -> anyhow::Result<Vec<(SearchServiceClient, Vec<Job>)>>;
 }
--- a/quickwit-search/src/client_pool.rs
+++ b/quickwit-search/src/client_pool.rs
@@ -49,6 +49,6 @@ pub trait ClientPool: Send + Sync + 'static {
     async fn assign_jobs(
         &self,
         jobs: Vec<Job>,
-        exclude_addresses: Option<HashSet<SocketAddr>>,
+        exclude_addresses: &Option<HashSet<SocketAddr>>,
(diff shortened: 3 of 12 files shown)


quickwit-cli/tests/cli.rs
```diff
@@ -507,3 +507,39 @@ async fn test_all_with_s3_localstack_internal_api() -> Result<()> {
 
     Ok(())
 }
+
+// TODO add failure test
+#[tokio::test]
+async fn test_failure_not() -> Result<()> {
```
fulmicoton, line 513: what is this testing?

quickwit-search/src/client_pool/search_client_pool.rs
```diff
@@ -129,7 +130,11 @@ impl ClientPool for SearchClientPool {
             // TODO optimize the case where there are few jobs and many clients.
             let clients = self.clients.read().await;
 
-            for (grpc_addr, client) in clients.iter() {
+            for (grpc_addr, client) in clients.iter().filter(|(grpc_addr, _)| {
+                address_whitelist
+                    .as_ref()
+                    .map_or(true, |whilelist| whilelist.contains(grpc_addr))
```
fulmicoton, line 136: Why the hell does `map_or` exist      (   O_O  " )

quickwit-search/src/error.rs
```diff
@@ -46,6 +47,12 @@ pub fn parse_grpc_error(grpc_error: &tonic::Status) -> SearchError {
     }
 }
 
+impl From<TantivyError> for SearchError {
+    fn from(any_err: TantivyError) -> Self {
```
fulmicoton, line 51: nitpick: maybe rename tantivy_err

quickwit-search/src/root.rs
```diff
@@ -212,13 +221,32 @@ pub async fn root_search(
...
     // Perform the query phase.
-    let result_per_node_addr = execute_search(
+    let mut result_per_node_addr = execute_search(
         &assigned_leaf_search_jobs,
         search_request_with_offset_0.clone(),
     )
     .await?;
 
-    analyze_errors(&result_per_node_addr);
+    if let Some(retry_action) = analyze_errors(&result_per_node_addr) {
```
fulmicoton, line 230: very readable!

fulmicoton (changes requested):
Can you add unit tests for the root when different errors happen?

quickwit-cli/tests/cli.rs
PSeitz: nothing yet I was waiting for minoru's pr

quickwit-search/src/client_pool/search_client_pool.rs
PSeitz: on Option map inner value, or on none

quickwit-proto/proto/search_api.proto
```diff
@@ -87,6 +91,9 @@ message SplitSearchError {
 
   // Split id that failed. 
   string split_id = 2;
+
+  // Flag to indicate if the error can be considered a retryable error
+  bool retryable_error = 3;
```
fulmicoton, line 96: I think this is the right way indeed.

quickwit-proto/proto/search_api.proto
```diff
@@ -140,7 +147,7 @@ message LeafSearchResult {
   repeated PartialHit partial_hits = 2;
 
   // The list of requests that failed. LeafSearchResult can be an aggregation of results, so there may be multiple.
```
fulmicoton, line 149: s/requests/splits/ in the comment

quickwit-search/src/client_pool.rs
```diff
@@ -42,8 +44,11 @@ pub struct Job {
 pub trait ClientPool: Send + Sync + 'static {
     /// Assign the given job to the clients.
     /// Returns a list of pair (SocketAddr, Vec<Job>)
+    ///
+    /// When exclude_addresses filters all clients it is ignored.
     async fn assign_jobs(
         &self,
         jobs: Vec<Job>,
+        exclude_addresses: &Option<HashSet<SocketAddr>>,
```
fulmicoton, line 52: It would have been more natural to have `exclude_addresses` be a `HashSet<SocketAddr>`.
When empty we do not exclude any address.

The contract becomes simpler, and so does the code.
As you know an empty HashSet does not allocate so it is not more expensive either.

Also, this implementation does something extremely unexpected if all addresses are discarded.
Here are 3 options:
a) document it
b) remove it entirely if it is not useful. (we just return an error if no node is available)
c) remove it and move that logic to the caller.

I don't like a) because contracts should be as simple as possible. (Complexity should stay within boxes on which we slap on labels with a contract on it.)
b) seems reasonable but I might be missing some information.
c) is ok but might be tricky as it might require information about the pool.

So I suspect c) would lead to something even more complex.
b) is my favorite, and then a) is ok.

quickwit-search/src/collector.rs
```diff
@@ -317,7 +317,7 @@ fn merge_leaf_results(leaf_results: Vec<LeafSearchResult>, max_hits: usize) -> L
         .sum();
     let failed_requests = leaf_results
```
fulmicoton, line 318: Renamed failed_requests into failed_splits and reduce line 331 too

quickwit-search/src/root.rs
```diff
@@ -136,39 +142,60 @@ fn analyze_errors(search_result: &SearchResultsByAddr) -> Option<ErrorRetries> {
...
         .filter_map(|result| result.as_ref().ok())
         .flat_map(|res| {
             let results = res
-                .failed_requests
+                .failed_splits
                 .iter()
-                .map(move |failed_req| failed_req.split_id.to_string())
+                .filter(move |failed_splits| failed_splits.retryable_error) // only retry splits marked as retryable
+                .map(move |failed_splits| failed_splits.split_id.to_string())
                 .collect_vec();
```
fulmicoton, line 164: I suspect line 164 and 165 are useless?

quickwit-search/src/root.rs
```diff
@@ -110,21 +113,24 @@ pub struct ErrorRetries {
...
 /// - Did all requests fail? (Should we retry in that case?)
 fn analyze_errors(search_result: &SearchResultsByAddr) -> Option<ErrorRetries> {
-    let contains_errors = search_result.values().any(|res| {
-        res.is_err()
-            || res
-                .as_ref()
-                .map(|ok_res| ok_res.failed_requests.is_empty())
-                .unwrap_or(false)
+    let contains_retryable_errors = search_result.values().any(|res| {
+        res.as_ref().map_or(true, |ok_res| {
```
fulmicoton, line 117: nitpick: It is probably just me but I have trouble reading `map_or` code.
Also I don't see how it helps compared to `.map(...).or(...)`

In this specific case, I would prefer a 


```
if let Some(leaf_search_result) = leaf_search_result_opt {
    contains_retryable_error(leaf_search_result)
} else {
    true
}
```

quickwit-search/src/root.rs
```diff
@@ -110,21 +113,24 @@ pub struct ErrorRetries {
 /// - Is the node ok, but just the split failing?
 /// - Did all requests fail? (Should we retry in that case?)
 fn analyze_errors(search_result: &SearchResultsByAddr) -> Option<ErrorRetries> {
-    let contains_errors = search_result.values().any(|res| {
-        res.is_err()
-            || res
-                .as_ref()
-                .map(|ok_res| ok_res.failed_requests.is_empty())
-                .unwrap_or(false)
+    let contains_retryable_errors = search_result.values().any(|res| {
```
fulmicoton, line 116: I suspect you can simplify your code by computing `retry_split_ids` explicitly even on your happy path.

Then contains_retryable_errors comes simply as 
```
let contains_retryable_error = retry_split_ids.is_empty();
```

quickwit-search/src/root.rs
```diff
@@ -110,21 +113,24 @@ pub struct ErrorRetries {
...
+                .filter(|err| err.retryable_error)
+                .count()
+                > 0
+        })
     });
-    if !contains_errors {
+    if !contains_retryable_errors {
         return None;
     }
     let complete_failure_nodes: Vec<_> = search_result
```
fulmicoton, line 129: I suspect a single procedural loop would actually have been easier to understand here.

```rust 
let mut complete_failure_nodes = Vec::new();
let mut partial_or_no_failure_nodes = Vec::new();
for (addr, leaf_search_result) in search_result {
    match leaf_search_result {
        Ok(res) => {
           // ...
        }
        Err(err) => {
           complete_failure_nodes.push(addr);
        }
    }
}
```
The functional approach you used here makes it very hard to proofread that this is a effectively partitioning the nodes.
Of course, you can also use the partition method and add a helper function called `is_complete_failure` or something like that.

quickwit-search/src/root.rs
```diff
@@ -136,39 +142,60 @@ fn analyze_errors(search_result: &SearchResultsByAddr) -> Option<ErrorRetries> {
...
 
     Some(ErrorRetries {
-        retry_split_ids: failed_split_ids,
-        nodes_to_use: retry_nodes,
+        retry_split_ids,
+        nodes_to_avoid: nodes_to_avoid.into_iter().collect(),
     })
 }
 
+fn job_for_split(
```
fulmicoton, line 183: jobs_for_splits maybe?

quickwit-search/src/root.rs
```diff
@@ -209,13 +228,79 @@ pub async fn root_search(
...
     // Perform the query phase.
-    let result_per_node_addr = execute_search(
+    let mut result_per_node_addr = execute_search(
         &assigned_leaf_search_jobs,
         search_request_with_offset_0.clone(),
     )
     .await?;
 
-    analyze_errors(&result_per_node_addr);
+    let analyze_result = analyze_errors(&result_per_node_addr);
```
fulmicoton, line 237: Rename `retry_action_opt` maybe?

quickwit-search/src/root.rs
```diff
@@ -209,13 +228,79 @@ pub async fn root_search(
...
+            .iter()
+            .filter(|(_addr, res)| res.is_err())
+            .map(|(addr, _err)| *addr)
+            .collect_vec();
+        for err_addr in complete_errors {
+            result_per_node_addr.remove(&err_addr);
+        }
+        // remove partial, retryable errors
+        for result in result_per_node_addr.values_mut().flatten() {
+            let failed_splits: Vec<SplitSearchError> = std::mem::take(&mut vec![]);
```
fulmicoton, line 267: Wtf !?
```rust
mem::take(&mut vec![])
```

quickwit-search/src/root.rs
```diff
@@ -209,13 +228,79 @@ pub async fn root_search(
...
+            result.failed_splits = failed_splits
+                .into_iter()
+                .filter(|err| !err.retryable_error)
+                .collect_vec();
```
fulmicoton, line 271: TIL about collect_vec

quickwit-search/src/root.rs
```diff
@@ -209,13 +228,79 @@ pub async fn root_search(
...
+        }
+
+        for (addr, new_result) in result_per_node_addr_new {
+            if let Some(Ok(orig_res)) = result_per_node_addr.get_mut(&addr) {
```
fulmicoton, line 275: Was the borrow checker making `match` pattern impossible? (it often happens)

quickwit-search/src/root.rs
```diff
@@ -209,13 +228,79 @@ pub async fn root_search(
...
+                    orig_res
+                        .failed_splits
+                        .extend(err.split_ids.iter().map(|split_id| SplitSearchError {
+                            error: err.search_error.to_string(),
+                            split_id: split_id.to_string(),
+                            retryable_error: true,
+                        }));
+                }
+            } else if let Some(Err(err)) = result_per_node_addr.get_mut(&addr) {
+                panic!("unexpected error leftover: {:?}", err);
```
fulmicoton, line 297: Wait... When does that happen?

quickwit-search/src/root.rs
```diff
@@ -209,13 +228,79 @@ pub async fn root_search(
...
+            .await?;
+        // Perform the query phase.
+        let result_per_node_addr_new = execute_search(
+            &retry_assigned_leaf_search_jobs,
+            search_request_with_offset_0.clone(),
+        )
+        .await?;
+
+        // clean out old errors
+        let complete_errors = result_per_node_addr
```
fulmicoton, line 257: I'm having trouble understanding l.255 .. l.300. If you have ideas on how to simplify it would be awesome

quickwit-search/src/root.rs
```diff
@@ -136,39 +142,60 @@ fn analyze_errors(search_result: &SearchResultsByAddr) -> Option<ErrorRetries> {
...
+            split: split_id.clone(),
+            cost: compute_split_cost(split_metadata),
+        })
+        .collect();
+    leaf_search_jobs
+}
+
 /// Perform a distributed search.
 /// It sends a search request over gRPC to multiple leaf nodes and merges the search results.
 pub async fn root_search(
```
fulmicoton, line 201: I think you wrote somewhere a great explanation on your retry logic.

Can you copy paste it in the doc of root_search and shorten it if necessary.

If I understand correctly, 


After a first round of leaf_search requests, we identify the list of failed but retryable splits,
and retry them. The leaf nodes which did not return any result are suspected to be unhealthy and are excluded
from this retry round.

quickwit-search/src/root.rs
```diff
@@ -573,11 +716,526 @@ mod tests {
...
+            Arc::new(SearchClientPool::from_mocks(vec![Arc::new(mock_search_service1)]).await?);
+
+        let search_result = root_search(&search_request, &metastore, &client_pool).await;
+        assert!(search_result.is_err());
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_root_search_one_splits_two_nodes_but_one_is_failing_for_split(
```
fulmicoton, line 1063: great tests

fulmicoton (changes requested):
I put a lot of comments, but overall that's awesome.

Retry logic for that kind of stuff is quite muddy, and the logic you went for is quite solid and easy to understand...
Good unit testing.

Please have a look at the comment I added though. There was one WTF, and a bunch of rather minor stuff or improvement suggestions.

quickwit-search/src/root.rs
PSeitz: ah yeah good catch, it was necessary in a previous version

quickwit-search/src/client_pool.rs
PSeitz: Ah yeah removing the Option is the best way. It was previously a whitelist, there it made sense.

It is already documented, but on the trait. 
The problem with b is, that the current logic retries if all nodes fail, so this behaviour would be weird, because it depends on how many nodes are assigned to our jobs.

c) would be the cleanest solution since there is no implicit logic, but it adds more complexity.

quickwit-search/src/root.rs
PSeitz: I think map_or is a little hard to read because the parameter order is or and then map.

I adopted your proposal. They are both weird to read because of the error double layering and think they should be flattened.

quickwit-search/src/root.rs
PSeitz: That removes some code, so that's better.

Rust doesn't allocate for empty vecs? I didn't heard that before, where can I find something about that? 
Although I wouldn't be concerned about the performance impact here by one vec allocation.

quickwit-search/src/root.rs
PSeitz: I think the functional approach with the helper function is the most readable

quickwit-search/src/root.rs
PSeitz: hahaha, what is that

It solved previously a lifetime issue, not sure how it became this

quickwit-search/src/root.rs
PSeitz: another nice itertools utility

PSeitz: should be imo in the standard library

quickwit-search/src/root.rs
PSeitz: I wasn't sure, what's more readable, but it's a match now

quickwit-search/src/root.rs
PSeitz: never, that's why it's a panic. We remove retryable errors before we retry and complete NodeSearchErrors are considered retryable currently

quickwit-search/src/root.rs
PSeitz: I think the code is as simple as it gets, but I can add some explanation

quickwit-search/src/root.rs
PSeitz: yes

PSeitz:
@fulmicoton I think I addressed all comments

quickwit-search/src/root.rs
```diff
@@ -110,67 +121,85 @@ pub struct ErrorRetries {
...
+        .filter(|(_addr, result)| is_complete_failure(*result))
         .map(|(addr, _)| *addr)
         .collect();
     let partial_or_no_failure_nodes: Vec<_> = search_result
         .iter()
         .filter(|(_addr, result)| {
             result
                 .as_ref()
-                .map(|ok_res| ok_res.failed_requests.len() as u64 != ok_res.num_attempted_splits)
+                .map(|ok_res| ok_res.failed_splits.len() as u64 != ok_res.num_attempted_splits)
```
fulmicoton, line 159: isn't it `!is_complete_failure()`? If it is we should use it, or even use .partition(...).

quickwit-search/src/root.rs
```diff
@@ -209,13 +232,85 @@ pub async fn root_search(
...
+            .filter(|(_addr, res)| res.is_err())
+            .map(|(addr, _err)| *addr)
+            .collect_vec();
+        for err_addr in complete_errors {
+            result_per_node_addr.remove(&err_addr);
+        }
+        // Remove partial, retryable errors
+        for result in result_per_node_addr.values_mut().flatten() {
+            let failed_splits: Vec<SplitSearchError> = vec![];
+            result.failed_splits = failed_splits
```
fulmicoton, line 278: WTF?

fulmicoton (changes requested):
See comments

quickwit-search/src/root.rs
PSeitz: yes, that's much better

fulmicoton (approved):

PR #331, opened 2021-07-23
CLI doc for gc command

evanxg852000:
### Context / purpose
Closes https://github.com/quickwit-inc/quickwit/issues/312

### Description
Added section in cli reference for `gc` command

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -267,8 +267,8 @@ quickwit delete
 
 *Options*
 
-`--index-uri` (string) Location of the target index.
-`--dry-run` (boolean) Executes the command in dry run mode and displays the list of files subject to be deleted.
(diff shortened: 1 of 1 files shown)


docs/reference/cli.md
```diff
@@ -281,3 +281,47 @@ quickwit delete --index-uri s3://quickwit-indexes/catalog
...
+```bash
+quickwit gc
+    --index-uri <uri>
+    [--grace-period <duration>]
+    [--dry-run]
+```
+
+:::note
+
+Intermediate files are created while executing Quickwit commands. These intermediate files are always cleaned at the end of each successfully executed command. However, failed or interrupted commands can leave behind intermediate files that need to be removed.
```
PSeitz, line 302: adding a comment to a low grace period would be good

guilload (approved):

PR #336, opened 2021-07-28
consider some errors non-retryable closes #124

### LINKED ISSUE #124: Gracefully handle failing split search
Right now if a single split fails, we fail. We need to at least retry this split, or return partial results and point out the failed split.

PSeitz:
Consider missing index and invalid query as non retryable


fulmicoton (changes requested):
Can we have a unit test?

quickwit-search/src/leaf.rs
```diff
@@ -178,10 +178,19 @@ pub async fn leaf_search(
...
-        .extend(errors.iter().map(|(split_id, err)| SplitSearchError {
-            split_id: split_id.to_string(),
-            error: format!("{:?}", err),
-            retryable_error: true,
+        .extend(errors.iter().map(|(split_id, err)| {
+            let retryable_error = match err {
+                SearchError::IndexDoesNotExist { index_id: _ } | SearchError::InvalidQuery(_) => {
+                    false
+                }
+                _ => true,
```
fulmicoton, line 186: Can you put the list of items instead of using the default pattern?

fulmicoton (changes requested):
See comment on the matched enum

quickwit-search/src/leaf.rs
PSeitz: I moved it to a a method to the enum,  so it should be easier to spot

fulmicoton:
@fmassot @PSeitz Is this PR now obsolete?

PSeitz:
I think it's still valid

fmassot:
I would not keep the code at the search error level and define retry able errors at the retry policy level.

fulmicoton:
@PSeitz Closing the PR.

If this is valuable, discuss with Francois where to put that and **add unit tests**.

PR #337, opened 2021-07-29
Rename and document environment variables

guilload:
### Description
- Rename `DISABLE_QUICKWIT_TELEMETRY` env variable to `QUICKWIT_DISABLE_TELEMETRY`
- Document env variables in CLI reference

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -326,3 +326,13 @@ quickwit gc --index-uri s3://quickwit-indexes/catalog --dry-run
 ```bash
 quickwit gc --index-uri s3://quickwit-indexes/catalog --grace-period 5m
 ```
+
+## Environment Variables
+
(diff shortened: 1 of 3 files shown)


docs/reference/cli.md
```diff
@@ -281,3 +281,13 @@ quickwit delete --index-uri s3://quickwit-indexes/catalog
 ```bash
 quickwit delete --index-uri s3://quickwit-indexes/catalog --dry-run
 ```
+
+## Environment Variables
+
+### QUICKWIT_ENV
```
fmassot, line 287: I would point out that this is currently used only for testing purpose.

fmassot (approved):

docs/reference/cli.md
guilload: Good call.

PR #353, opened 2021-08-04
Add a health check endpoint on the search server.

### LINKED ISSUE #268: Add a healthcheck endpoint on search server
This is useful when you want to host search servers behind a load balancer. The load balancer generally expects an endpoint to check if the server is correctly running.

I think we did that for barrel.

mosuka:
### Context / purpose
Add a health check endpoint on the search server. https://github.com/quickwit-inc/quickwit/issues/268

### Description
Port the health check functionality from Barrel.

### How was this PR tested?

```
$ cargo run -- serve --host=0.0.0.0 --index-uri=file:///home/minoru/tmp/quickwit_data/indices/hdfslogs-idx --port=8080
```

```
$ curl 'http://0.0.0.0:8080/health/livez'
```

### Checklist
- [x] tested locally
- [x] added unit tests
- [x] included documentation

Close #268


fmassot:
Thanks @mosuka for the PR. But I can't merge that for the moment because:
- I think we don't need HTML but only a json response
- We can start by something very simple, like a 200 empty response is good 
- We need to define what health checks we want before

mosuka:
@fmassot 
I had mistakenly thought that it ported from barrel.
I’ll define the health check specification.
Thanks.

fmassot:
Can you remove this file?

mosuka:
@fmassot 
I have simplified the code by removing unnecessary code.
Please take another look.

quickwit-serve/src/rest.rs
```diff
@@ -528,6 +547,17 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn test_rest_search_api_health_check_livez() {
+        let rest_search_api_filter = liveness_check_handler();
+        let resp = warp::test::request()
+            .path("/health/livez")
+            .reply(&rest_search_api_filter)
+            .await;
+        println!("{:?}", resp);
```
fmassot, line 557: remove the print

fmassot (changes requested):

quickwit-serve/src/rest.rs
mosuka: Removed it. Thanks.

fmassot (approved):

PR #355, opened 2021-08-04
add BundleStorage

PSeitz:
### Context / purpose
Describe the context and purpose of this PR. Link related issue(s) if any.

### Description

### How was this PR tested?

### Checklist
- [ ] tested locally
- [ ] added unit tests
- [ ] included documentation


quickwit-directories/src/bundle_storage.rs
```diff
@@ -0,0 +1,289 @@
...
+struct BundleStorageMetadata {
+    pub(crate) files: HashMap<PathBuf, FileOffsets>,
+}
+
+impl BundleStorageMetadata {
+    fn open(data: &[u8]) -> io::Result<Self> {
+        let footer_size: u64 = BinarySerializable::deserialize(&mut &data[data.len() - 8..])?;
+
+        let footer_slice = &data[data.len() - 8 - footer_size as usize..data.len() - 8];
+        //println!("{}", String::from_utf8(footer_slice.to_owned()).unwrap());
```
fmassot, line 80: remove println

quickwit-directories/src/bundle_storage.rs
```diff
@@ -0,0 +1,289 @@
...
+            )
+            .await
+    }
+
+    async fn delete(&self, path: &Path) -> quickwit_storage::StorageResult<()> {
+        Err(unsupported_operation(path))
+    }
+
+    async fn exists(&self, path: &Path) -> quickwit_storage::StorageResult<bool> {
+        // also check if self.bundle_file_name exists ?
```
fmassot, line 141: Checking bundle_file_name seems strange as this is the bundle.

I would check only files in metadata.

quickwit-directories/src/bundle_storage.rs
```diff
@@ -0,0 +1,289 @@
...
+pub struct CreateBundleStorage {
+    metadata: BundleStorageMetadata,
+    current_offset: u64,
+    /// the offset in the file where the hotcache begins. This is used to read
+    /// the hotcache and footer in a single read.
+    pub hotcache_offset: u64,
+    bundle_file: File,
+}
+
+impl CreateBundleStorage {
```
fmassot, line 189: BundleStorageBuilder?

quickwit-directories/src/lib.rs
```diff
@@ -27,14 +27,17 @@ This crate contains all of the building pieces that make quickwit's IO possible.
 - The `HotDirectory` wraps another directory with a static cache.
 - The `CachingDirectory` wraps a Directory with a dynamic cache.
 - The `DebugDirectory` acts as a proxy to another directory to instrument it and record all of its IO.
+- The `BundleStorage` bundles together multiple files into a single file.
 */
 #![warn(missing_docs)]
 
+mod bundle_storage;
```
fmassot, line 34: I would put bundle_storage into quickwit-storage as here we only keep structs that implement `Directory` trait.

quickwit-directories/src/bundle_storage.rs
```diff
@@ -0,0 +1,289 @@
...
+            bundle_file: sink,
+            current_offset: 0,
+            hotcache_offset: 0,
+            metadata: Default::default(),
+        })
+    }
+
+    /// Appends a file to the bundle file.
+    ///
+    /// The hotcache needs to be the last file that is added, in order to be able to read
```
fmassot, line 204: I'm worried about the user that will add the hotcache at the beginning. 

What will happen if add it at the beginning and even if we forget to add the hotcache? Can we protect us against these two situations?

fmassot (commented):
Sounds already good, I put a few comments.

quickwit-directories/src/bundle_storage.rs
PSeitz: we could check if the bundle exists in the underlying storage, if  the bundle doesn't exist the file does not exist. but it's much more expensive

quickwit-directories/src/bundle_storage.rs
PSeitz: I added some checks, but did not yet commit them

quickwit-directories/src/lib.rs
PSeitz: I moved it

quickwit-common/src/lib.rs
```diff
@@ -24,6 +24,9 @@ use std::net::{SocketAddr, ToSocketAddrs};
 use once_cell::sync::Lazy;
 use regex::Regex;
 
+/// Filenames used for hotcache files.
+pub const HOTCACHE_FILENAME: &str = "hotcache";
```
fulmicoton, line 28: Why in common?
It would make more sense in quickwit-storage?

quickwit-directories/src/lib.rs
```diff
@@ -27,6 +27,7 @@ This crate contains all of the building pieces that make quickwit's IO possible.
 - The `HotDirectory` wraps another directory with a static cache.
 - The `CachingDirectory` wraps a Directory with a dynamic cache.
 - The `DebugDirectory` acts as a proxy to another directory to instrument it and record all of its IO.
+- The `BundleStorage` bundles together multiple files into a single file.
```
fulmicoton, line 30: Why in Directory? It's a storage.

quickwit-search/src/service.rs
```diff
@@ -130,6 +130,14 @@ impl SearchService for SearchServiceImpl {
                 index_id: search_request.index_id.clone(),
             })?;
         let index_metadata = metastore.index_metadata(&search_request.index_id).await?;
+        //let split_meta_data = metastore
```
fulmicoton, line 133: remove

quickwit-storage/Cargo.toml
```diff
@@ -16,8 +16,10 @@ md5 = '0.7'
 anyhow = '1'
 futures = '0.3'
 bytes = '1'
+serde_json = "1"
 base64 = '0.13'
 tracing = '0.1'
+tantivy = { git= "https://github.com/quickwit-inc/tantivy", rev="c1d1d84"}
```
fulmicoton, line 22: Let's not add tantivy as a deps just for BinarySerializable.

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,282 @@
...
+use std::io::{self, ErrorKind, Write};
+use std::ops::Range;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+use tantivy::common::BinarySerializable;
+use tantivy::HasLen;
+use tracing::error;
+
+/// Filename used for the bundle.
+pub const BUNDLE_FILENAME: &str = "bundle";
```
fulmicoton, line 43: We'll probably drop the concept of split directory now that everything is a file. It will make operations nicer.
(one will be able to ls the index directory and see what is in there)

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    async fn copy_to_file(&self, path: &Path, _output_path: &Path) -> crate::StorageResult<()> {
+        Err(unsupported_operation(path))
+    }
+
+    async fn get_slice(&self, path: &Path, range: Range<usize>) -> crate::StorageResult<Bytes> {
+        let file_offsets = self.metadata.get(path)?;
+        let new_range =
+            range.start + file_offsets.start as usize..range.end + file_offsets.start as usize;
```
fulmicoton, line 108: super nitpick... I'd commute the addition there.
Offset usually comes first.

No need to change if you disagree.

```
file_offsets.start as usize + range.start..file_offsets.start as usize +  range.end
```

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,282 @@
...
+        Ok(self.metadata.exists(path))
+    }
+
+    fn uri(&self) -> String {
+        self.storage.uri()
+    }
+}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+struct FileOffsets {
```
fulmicoton, line 139: Can you add a comment to explain this is a semi-open interval? 
It seems obvious but we've had a bug related to that ambiguity.

fulmicoton: Alternatively you can use Range<u64>. Apparently it is serializable.

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,282 @@
...
+            "BundleStorage({:?}, files={:?})",
+            &self.bundle_file_name, self.metadata
+        )
+    }
+}
+
+fn unsupported_operation(path: &Path) -> StorageError {
+    let msg = "Unsupported operation. BundleStorage only supports async reads";
+    error!(path=?path, msg);
+    io::Error::new(io::ErrorKind::Other, format!("{}: {:?}", msg, path)).into()
```
fulmicoton, line 163: `io::ErrorKind::Unsupported`

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// The footer includes the footer metadata as json encoded and the size of the footer in
+    /// bytes.
+    pub footer_offset: u64,
+    bundle_file: File,
+}
+
+impl BundleStorageBuilder {
+    /// Creates a new CreateBundleStorage, to which files can be added.
+    pub fn new(path: &Path) -> io::Result<Self> {
+        let sink = OpenOptions::new().create(true).append(true).open(path)?;
```
fulmicoton, line 184: Why append?

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// The footer offset, where the footer begins.
+    /// The footer includes the footer metadata as json encoded and the size of the footer in
+    /// bytes.
+    pub footer_offset: u64,
+    bundle_file: File,
+}
+
+impl BundleStorageBuilder {
+    /// Creates a new CreateBundleStorage, to which files can be added.
+    pub fn new(path: &Path) -> io::Result<Self> {
```
fulmicoton, line 183: Taking an io::Write would be nicer.

If you want to get rid of the generic, you can just make new generic, wrap it in a BufWriter<Box<dyn io::Write>>

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub hotcache_offset: u64,
+    /// The footer offset, where the footer begins.
+    /// The footer includes the footer metadata as json encoded and the size of the footer in
+    /// bytes.
+    pub footer_offset: u64,
+    bundle_file: File,
+}
+
+impl BundleStorageBuilder {
+    /// Creates a new CreateBundleStorage, to which files can be added.
```
fulmicoton, line 182: s/CreateBundleStorage/BundleStorageBuilder/

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,282 @@
...
+        })
+    }
+
+    /// Appends a file to the bundle file.
+    ///
+    /// The hotcache needs to be the last file that is added, in order to be able to read
+    /// the hotcache and the metadata in one continous read.
+    pub fn add_file(&mut self, path: &Path) -> io::Result<()> {
+        let mut file = File::open(path)?;
+        let bytes_written = io::copy(&mut file, &mut self.bundle_file)?;
```
fulmicoton, line 201: Nitpick. 

if you want to make your life easier for unit test, you can add another public function that takes a 

```
fn add_file_from_read(&Path, read: &mut io::Read) -> io::Result<()>;
```

And call it from your `add_file` method.

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,282 @@
...
+            self.hotcache_offset = self.current_offset;
+        }
+        self.metadata.files.insert(
+            file_name,
+            FileOffsets {
+                start: self.current_offset,
+                end: self.current_offset + bytes_written,
+            },
+        );
+        self.current_offset += bytes_written;
```
fulmicoton, line 218: That's nice... I didn't know io::copy returned the number of bytes written!

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,282 @@
...
+pub struct BundleStorageBuilder {
+    metadata: BundleStorageMetadata,
+    current_offset: u64,
+    /// The offset in the file where the hotcache begins. This is used to read
+    /// the hotcache and footer in a single read.
+    pub hotcache_offset: u64,
+    /// The footer offset, where the footer begins.
+    /// The footer includes the footer metadata as json encoded and the size of the footer in
+    /// bytes.
+    pub footer_offset: u64,
```
fulmicoton, line 177: *not a nitpick*.
I think we need to remove those `pub` and have finalize return the metadata object if you need it.

finalize could also consume self.

`fn finalize(self) -> io::Result<BundleStorageMetadata>`

fulmicoton (changes requested):
I added a bunch of comments. Most are nitpicks, feel free to ignore them.

quickwit-common/src/lib.rs
PSeitz: because it's also used in directories, the hotcache is a directory and we already have a directories -> storage dependency

quickwit-directories/src/lib.rs
PSeitz: I moved it to to storage, but forgot about that comment

quickwit-storage/Cargo.toml
PSeitz: tantivy is already a dependency of the whole package though.
What serializer should we use then?

quickwit-storage/src/bundle_storage.rs
PSeitz: what are the implications of dropping the concept of split directory?

When we `ls` the directory we just see the bundle file, or do you mean a virtual ls on the bundle metadata?

quickwit-storage/src/bundle_storage.rs
PSeitz: I switched to Range, but this is not an semi-open interval, it has start and end

quickwit-storage/src/bundle_storage.rs
PSeitz: true, File::create is enough

quickwit-storage/src/bundle_storage.rs
PSeitz: But in the creation step you don't want the BundleStorageMetadata, these offsets are only used for the single read and not used inside the bundle itself. we can return the offsets instead of pub, it's not much difference it think

quickwit-storage/src/bundle_storage.rs
PSeitz: I would keep `Path` for now, since we create the bundle in the local fs and then upload it, the creation and ownership can be in the builder

fmassot:
The bundle seems good for me. 

But currently, when I index some documents, this will produce a bundle but we don't use it. We can't merge that (and we need to remove some comments.

The plan was to merge a bundle but not plugged because @fulmicoton is working on indexing in // and it's quite a heavy part.
Or maybe we can reconsider this, @fulmicoton can you tell us your thoughts on this so that we merge now or plug the bundle right now ?

quickwit-storage/src/bundle_storage.rs
fulmicoton: Semi-Open or Half-Open means something like this
`start <= ... < end`
In US they write it `[start, end)`. In France it is usually `[start, end[`.

Range in rust is semi-open, while RangeInclusive is closed.

PSeitz: I think half-open is when the Range is open on one side, so there is only start or end `(start ..)` or `(..end)`, while `start..end` is a not inclusive closed range

quickwit-storage/src/bundle_storage.rs
fulmicoton: That's a bottom-up vs top-bottom debate.
It's tricky because there are Pros and Cons to both.

The cons of bottom-up are mild here:
- We don't really hide the intent too much.
- We don't introduce complicated abstraction that impact the complexity.

The pros are:
- The code reuse is easier. I agree we need to consider that one mild.
We are not closing any doors by using Path.
- Unit tests will be cleaner and won't write to disk.

quickwit-storage/src/bundle_storage.rs
fulmicoton: (I let you decide)

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,276 @@
...
+
+        let path_root = format!("file://{}", temp_dir.to_string_lossy());
+        let file_storage = LocalFileStorage::from_uri(&path_root)?;
+        let bundle_storage = BundleStorage {
+            metadata,
+            bundle_file_name,
+            storage: Arc::new(file_storage),
+        };
+        let f1_data = bundle_storage.get_all(Path::new("f1")).await?;
+        assert_eq!(f1_data, &vec![123, 76]);
```
fulmicoton, line 264: no need for vec!.

```
&[123, 76]
```

works I think

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,276 @@
...
+#[cfg(test)]
+mod tests {
+    use std::env::temp_dir;
+
+    use crate::LocalFileStorage;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn bundlestorage_test() -> anyhow::Result<()> {
```
fulmicoton, line 233: Can we add a silly test for what happens if the bundlestorage contains no file?

quickwit-storage/src/bundle_storage.rs
```diff
@@ -0,0 +1,276 @@
...
+            file_name,
+            self.current_offset..self.current_offset + bytes_written,
+        );
+        self.current_offset += bytes_written;
+        Ok(())
+    }
+
+    /// Writes the metadata into the footer.
+    ///
+    pub fn finalize(mut self) -> io::Result<(usize, usize)> {
```
fulmicoton, line 215: It is impossible for the reader to know what these two usize are.
Can we return a simple struct with good names?

Also I think it might be nice to return the overall file size as well.

quickwit-storage/src/bundle_storage.rs
fulmicoton: I let you decide.

quickwit-storage/src/bundle_storage.rs
fulmicoton: Ah yes.. We don't want the whole object.

Returning a pair of offset is not good though... Can we make it a dedicated struct with good name for the two offsets?
Shouldn't we add the file size as well?

PSeitz: yes, but then we are really close to just making the two fields pub :)

quickwit-storage/src/bundle_storage.rs
PSeitz: do you mean if we add no file to the bundle or if the underlying file does not exist?

fulmicoton: If we add no file to the bundle. NOt that I think there is a bug, but that is a good black box test to have.

quickwit-storage/Cargo.toml
fulmicoton: It is not a deps of quickwit storage. You only serialize 2 u64?
If so can you just do it directly with rust?

quickwit-storage/src/bundle_storage.rs
PSeitz: it returns a `BundleStorageOffsets` now

quickwit-cli/tests/cli.rs
```diff
@@ -322,7 +318,7 @@ async fn test_cmd_garbage_collect_spares_files_within_grace_period() -> Result<(
         .assert()
         .success()
         .stdout(predicate::str::contains(
-            "No dangling files to garbage collect",
+            "No dangling splits to garbage collect",
```
fulmicoton, line 321: nice you recalled doing this!

quickwit-index/src/actors/publisher.rs
```diff
@@ -72,7 +72,7 @@ impl AsyncActor for Publisher {
             .await
             .with_context(|| "Upload apparently failed")?; //< splits must be published in order, so one uploaded failing means we should fail entirely.
         self.metastore
-            .publish_splits(&uploaded_split.index_id, &[&uploaded_split.split_id])
+            .publish_splits(&uploaded_split.index_id, vec![uploaded_split.split])
```
fulmicoton, line 75: why vec?

quickwit-index/src/actors/uploader.rs
```diff
@@ -146,14 +126,14 @@ async fn put_split_files_to_storage(
...
 
     let elapsed_secs = start.elapsed().as_secs();
-    let file_statistics = manifest.file_statistics();
+    let file_statistics = split.file_statistics.clone();
     info!(
         min_file_size_in_bytes = %file_statistics.min_file_size_in_bytes,
         max_file_size_in_bytes = %file_statistics.max_file_size_in_bytes,
         avg_file_size_in_bytes = %file_statistics.avg_file_size_in_bytes,
-        split_size_in_megabytes = %manifest.split_size_in_bytes / 1000,
+        split_size_in_megabytes = %split.bundle_offsets.bundle_file_size / 1000,
```
fulmicoton, line 134: i don't understand the 1_000

fulmicoton: can we do the multiplication above:

let split_size_in_megabytes = ...;

and use the expression in throughput_mb_s as well?

quickwit-metastore/src/metastore.rs
```diff
@@ -63,6 +64,9 @@ pub struct SplitMetadata {
     /// JSON payloads.
     pub size_in_bytes: u64,
 
+    /// The bundle offsets. Used for single read of hotcache and bundle footer.
+    pub bundle_offsets: BundleStorageOffsets,
```
fulmicoton, line 68: so we store it in the SplitMetadata and not the filename or both?

fulmicoton:
I wish this was two PRs (I know I am not showing a great example with my gigantic PR).
The introduction of the bundle storage module, and then its insertion into the rest of the project.

Well sometimes this is difficult.

quickwit-metastore/src/metastore.rs
```diff
@@ -176,10 +187,10 @@ pub trait Metastore: Send + Sync + 'static {
     /// At this point, the split files are assumed to have already been uploaded.
     /// If the split is already published, this API call returns a success.
     /// An error will occur if you specify an index or split that does not exist in the storage.
-    async fn publish_splits<'a>(
+    async fn publish_splits(
         &self,
         index_id: &str,
-        split_ids: &[&'a str],
+        split_metadata: Vec<SplitMetadata>,
```
fulmicoton, line 193: let's not change publish.

staging happens after packaging.

quickwit-metastore/src/metastore.rs
```diff
@@ -193,6 +204,25 @@ pub trait Metastore: Send + Sync + 'static {
         time_range: Option<Range<i64>>,
     ) -> MetastoreResult<Vec<SplitMetadata>>;
 
+    /// Lists the splits.
+    /// Returns a list of splits that intersects the given `time_range`, `split_state` and
+    /// `split_ids`.
+    /// Regardless of the time range filter, if a split has no timestamp it is always returned.
+    /// An error will occur if an index that does not exist in the storage is specified.
+    async fn list_split_ids(
```
fulmicoton, line 212: Remove this method.

PR #359, opened 2021-08-05
Export data as csv [W.I.P]

evanxg852000:
### Context / purpose
Closes https://github.com/quickwit-inc/quickwit/issues/324

### Description
Very minimal implementation that need early feedbacks still need lot of clean up and tests 

### How was this PR tested?
- Run these set of commands to setup and index with an i64 field as fast_field
```bash
cargo run -- new --index-uri file://$(pwd)/movies --index-config-path ./movies-config.json
cargo run -- index --index-uri file://$(pwd)/movies --input-path ./movies.json
cargo run -- search --index-uri file://$(pwd)/movies --query avenger
cargo run -- serve --index-uri file://$(pwd)/movies 
```

- Point browser to `http://0.0.0.0:8080/api/v1/movies/export?query=spider&fastField=id` endpoint

The file mentioned in the script are available here https://github.com/evanxg852000/live-search/tree/master/quickwit

### Checklist
- [x] tested locally
- [ ] added unit tests
- [ ] included documentation


evanxg852000:
@fulmicoton @fmassot  I will need your opinion on how we can create a common type for the collector in order use the same type in many places without duplication. For now I settled with something simple and let you suggest the best design https://github.com/quickwit-inc/quickwit/pull/359/files#diff-1c9121e90e06ee53fb8c7972a2bc0b32d02a8be0faa3476f0c8e7abec3e97715R187

fmassot (commented):
Ok, as discussed we need to create a specific collector and the segment collector should return binaries. Then we should be able to stream results by split. I will push some code very soon.

quickwit-proto/proto/search_api.proto
```diff
@@ -167,3 +176,11 @@ message FetchDocsResult {
   // List of complete hits.
   repeated Hit hits = 1;
 }
+
+
+// -- Export -------------------
+
+message LeafExportResult {
+  // Row of data serialized  in bytes (DocID, fast_field_value)
```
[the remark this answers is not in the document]
evanxg852000: ok only one collumn (fast field)

quickwit-search/src/collector.rs
```diff
@@ -97,10 +101,11 @@ fn resolve_sort_by(
 
 /// PartialHitHeapItem order is the inverse of the natural order
 /// so that we actually have a min-heap.
-#[derive(Clone, Copy)]
+#[derive(Clone)]
 struct PartialHitHeapItem {
     sorting_field_value: u64,
     doc_id: DocId,
+    fast_field_value: Option<i64>,
```
fulmicoton, line 108: wtf

quickwit-search/src/collector.rs
```diff
@@ -219,6 +227,127 @@ impl SegmentCollector for QuickwitSegmentCollector {
     }
 }
 
+pub struct QuickwitSegmentExportCollector {
+    values: Vec<u8>,
```
fulmicoton, line 231: rename.. 

values is rarely a good name.

quickwit-proto/proto/search_api.proto
```diff
@@ -167,3 +170,47 @@ message FetchDocsResult {
...
+  // Index ID
+  string index_id = 1;
+
+  // Query
+  string query = 2;
+
+  // Fields to search on
+  repeated string  search_fields = 3;
+  
+  // Time filter
```
fulmicoton, line 188: Add comment:
The time filter is interpreted as a semi-open interval. [start, end)

quickwit-proto/proto/search_api.proto
```diff
@@ -167,3 +170,47 @@ message FetchDocsResult {
...
+  // Name of the fast field to export
+  string fast_field = 7;
+
+  // The output format
+  string output_format = 8;
+}
+
+message LeafExportRequest {
+  // Export request. This is a perfect copy of the original export request,
+  // that was sent to root apart from the start_offset & max_hits params.
```
fulmicoton, line 204: No start_offset I think...

quickwit-proto/proto/search_api.proto
```diff
@@ -167,3 +170,47 @@ message FetchDocsResult {
...
+
+  // Fields to search on
+  repeated string  search_fields = 3;
+  
+  // Time filter
+  optional int64 start_timestamp = 4;
+  optional int64 end_timestamp = 5;
+
+  // Maximum number of hits to return.
+  uint64 max_hits = 6;
```
fulmicoton, line 193: Do we really handle max_hits? That's non trivial.

quickwit-search/src/client.rs
```diff
@@ -121,6 +122,37 @@ impl SearchServiceClient {
...
+                    .map_err(|tonic_error| parse_grpc_error(&tonic_error))?
+                    .into_inner();
+
+                // TODO: returning stream instead of a channel may be better.
+                // But this seems to be difficult. Try it at your own expense.
+                let (export_sender, export_receiver) = tokio::sync::mpsc::channel(10);
+                while let Some(result) = export_result_stream
+                    .message()
+                    .await
+                    .map_err(|status| parse_grpc_error(&status))?
```
fulmicoton, line 145: That looks good but I am surprsied of the type here.

Could we have ended with 
```
crate::Result<ReceiverStream<Result<LeafExportResult, tonic::Status>>>
```

Looking at the impl I get the impression that it is what it was doing, so I am confused.

quickwit-search/src/error.rs
```diff
@@ -41,6 +41,25 @@ pub enum SearchError {
     InvalidQuery(String),
 }
 
+impl SearchError {
+    fn convert_to_tonic_status_code(search_error: &SearchError) -> tonic::Code {
```
fulmicoton, line 45: There was probably code doing that somewhere, is it removed?

quickwit-search/src/export/leaf.rs
```diff
@@ -0,0 +1,127 @@
...
+
+/// `leaf` step of export.
+// Note: we return a stream of a result with a tonic::Status error
+// to be compatible with the stream coming from the grpc client.
+// It would be better to have a SearchError but we need then
+// to process stream in grpc_adapater.rs to change SearchError
+// to tonic::Status as tonic::Status is required by the stream result
+// signature defined by proto generated code.
+pub async fn leaf_export(
+    query: Box<dyn Query>,
```
fulmicoton, line 42: I don't think we can pass a `Box<dyn Query>` here.
It assumes that all splits have the same schema.

It is a bummer but we want to parse the query for each split.
In the future it would be better to parse the query to an AST independent of the schema and pass that to each split.

quickwit-search/src/export/leaf.rs
```diff
@@ -0,0 +1,127 @@
...
+    fast_field_collector_builder: FastFieldCollectorBuilder,
+    split_ids: Vec<String>,
+    storage: Arc<dyn Storage>,
+    output_format: OutputFormat,
+) -> ReceiverStream<Result<LeafExportResult, tonic::Status>> {
+    let (result_sender, result_receiver) = tokio::sync::mpsc::channel(10);
```
fulmicoton, line 48: This channel is bounded, but since a task is spawned for every single split anyway, the fast that it is bounded does not create backpressure. It does not help with cpu usage nor memory.

I think you can make it unbounded.

quickwit-search/src/export/leaf.rs
```diff
@@ -0,0 +1,127 @@
...
+    warmup(
+        &*searcher,
+        query.as_ref(),
+        fast_field_collector_builder.fast_field_to_warm(),
+    )
+    .await?;
+    let mut buffer = Vec::new();
+    match fast_field_collector_builder.value_type() {
+        Type::I64 => {
+            let fast_field_collector = fast_field_collector_builder.build_i64();
```
fulmicoton, line 102: I suspect it feels a bit bad to have the same code copy paste.
I think this is perfectly fine here.

There are only 2 cases. Trying to do something smart would be overengineering.

quickwit-search/src/export/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+impl From<String> for OutputFormat {
+    fn from(spec: String) -> Self {
+        match spec.as_str() {
+            "rowbinary" => Self::RowBinary,
+            _ => Self::CSV,
+        }
+    }
+}
+
+pub struct ExportSerializer<'a> {
```
fulmicoton, line 70: I don't think a struct is helpful here.

```rust
pub fn serialize<TFastValue: FastValue>(vals: &[TFastValue], buffer: &mut Vec<u8>, format: OutputFormat) -> io::Result<()> {...}
fn serialize_csv<TFastValue: FastValue>(vals: &[TFastValue], buffer: &mut Vec<u8>) -> io::Result<()> {...}
fn serialize_row_binary<TFastValue: FastValue>(vals: &[TFastValue], buffer: &mut Vec<u8>) -> io::Result<()> {...}
```

quickwit-search/src/export/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    output_format: OutputFormat,
+    writer: &'a mut Vec<u8>,
+}
+
+impl<'a> ExportSerializer<'a> {
+    pub fn write_u64(&mut self, values: Vec<u64>) -> crate::Result<()> {
```
fulmicoton, line 76: you can remove the distinction u64/i64 by cheating a bit since FastValue has a `.as_u64()` method.

fulmicoton: I have several issues here:

This function seems lowlevel enough that it should not have any concern about SearchError.
If it had to return an error, it should probably be something like io::Result<()>. The high level calling code should then convert the error to a higher level error if needed.

BUT:
your code cannot error in reality. We are writing in a Vec anyway.

The trouble is coming from the fact that you are using io::Write.
A better way (the compiler might even spew better code although I doubt it) is to use extend.

Finally there is technically a bug in your code.
`.write()` can actually only write you content partially. You almost always want to call `.write_all()`.

So after rewrite

```rust
#[inline(never)]
fn serialize_row_binary<TFastValue: FastValue>(vals: &[TFastValue], buffer: &mut Vec<u8>) {
    buffer.clear(); // by the way we need to comment in the public function that it clears the buffer if it is not empty
    buffer.reserve_exact(std::mem::size_of::<TFastValue>() * vals.len());
    for &val in vals {
        buffer.extend(val.as_u64().to_le_bytes());
    }
}
```

fulmicoton: (I tested in godbolt, rust is unfortunately not smart enough to transform that in a memcopy. If the serialization is not negligeable, we can special case for little endian arch.

quickwit-search/src/client.rs
evanxg852000: Yes so the stream returns result

quickwit-search/src/error.rs
evanxg852000: Yes, I moved it here because of this need.

quickwit-search/src/export/root.rs
```diff
@@ -0,0 +1,194 @@
...
+                    search_error: parse_grpc_error(&status),
+                    split_ids: split_ids.clone(),
+                })?;
+                leaf_bytes.extend(leaf_data.data);
+            }
+            Result::<Vec<u8>, NodeSearchError>::Ok(leaf_bytes)
+        });
+        handles.push(handle);
+    }
+    let export_results = futures::future::try_join_all(handles).await?;
```
r-zenine, line 95: Is it possible to start streaming the result before all the clients are done ? 
Given how `row_binary` works, it should be doable. I see that this function returns a `Vec<u8>`, I don't know how easy it is to change that.

quickwit-search/src/export/root.rs
```diff
@@ -0,0 +1,194 @@
...
+    let mut errors = Vec::new();
+    // TODO: refactor this part...
+    let mut buffer = Vec::new();
+    for response in export_results {
```
r-zenine, line 99: It might be faster to traverse a first time the results, collect the total number of bytes that we should allocate and allocate it first. Then traverse a second time to write the `responses`

quickwit-search/src/export/root.rs
```diff
@@ -0,0 +1,194 @@
...
+use crate::ClientPool;
+use crate::SearchClientPool;
+use crate::SearchError;
+
+/// Perform a distributed export.
+pub async fn root_export(
+    export_request: &ExportRequest,
+    metastore: &dyn Metastore,
+    client_pool: &Arc<SearchClientPool>,
+) -> Result<Vec<u8>, SearchError> {
```
r-zenine, line 48: Do we have to return a `Vec<u8>` ? 
we could return a `Vec<Vec<u8>>` and save on the extra allocation line `103`

r-zenine (commented):
I just left some comment regarding the `root_export` function. I hope they will be helpful. 

Thanks a lot for the hard work :). It's my first review here, sorry if I overstepped in any way !

quickwit-search/src/export/leaf.rs
fmassot: let's open a github issue to change that for `search` & `export` as we are doing that for the search part too.

quickwit-proto/proto/search_api.proto
evanxg852000: I have removed this, let's add this when we actually make use of it

quickwit-search/src/export/root.rs
fmassot: Yes, we tested it and it works. But we need first to know how to handle errors during stream on the client side. Eventually, we can add a `stream` parameter in the request so it's the client to decide to stream at his own risks :)

quickwit-search/src/export/leaf.rs
fmassot: https://github.com/quickwit-inc/quickwit/issues/365

quickwit-search/src/export/root.rs
evanxg852000: Currently just returning the `Vec<Vec<u8>>` , and reserving capacity. we could also use partition on `export_result`.

fmassot:
> I just left some comment regarding the `root_export` function. I hope they will be helpful.
> 
> Thanks a lot for the hard work :). It's my first review here, sorry if I overstepped in any way !

that was very useful! we are still doing too many allocations and we will improve that in the next steps but for now we already return a `vec<vec<u8>>` and it has increased a little bit the performance, something around 5-10%, which is definitely very nice :)

quickwit-search/src/error.rs
fulmicoton: awesome

quickwit-search/src/search_stream/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+    io::{self, Write},
+};
+use tantivy::fastfield::FastValue;
+
+/// Output format for stream.
+#[derive(Deserialize, Debug, PartialEq, Eq, Clone, Copy)]
+#[serde(rename_all = "lowercase")]
+pub enum OutputFormat {
+    /// Format data by row in ClickHouse binary format. TODO: add link to format.
+    CHRowBinary,
```
fulmicoton, line 41: Let's explicitly call it ClickHouseRowBinary. I had trouble understanding what was CH about.

quickwit-search/src/search_stream/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+impl ToString for OutputFormat {
+    fn to_string(&self) -> String {
+        match &self {
+            Self::CHRowBinary => "ch-row-binary".to_string(),
+            Self::CSV => "csv".to_string(),
+        }
+    }
+}
+
+impl From<String> for OutputFormat {
```
fulmicoton, line 62: I wonder if it could be From<&str>

quickwit-search/src/search_stream/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+impl Default for OutputFormat {
+    fn default() -> Self {
+        OutputFormat::CHRowBinary
+    }
+}
+
+impl ToString for OutputFormat {
+    fn to_string(&self) -> String {
```
fulmicoton, line 54: nitpick: I agree it is nice to reuse trait off-the-shelf. Ideally however, we would have preferred to_str(), which does not exist.

It might be nice to make it a simple method.

quickwit-search/src/search_stream/mod.rs
fmassot: arf, I forget to rename it, `ClickHouseRowBinary` is better

quickwit-search/src/search_stream/root.rs
```diff
@@ -0,0 +1,186 @@
...
+use crate::SearchClientPool;
+use crate::SearchError;
+
+/// Perform a distributed search stream.
+pub async fn root_search_stream(
+    search_stream_request: &SearchStreamRequest,
+    metastore: &dyn Metastore,
+    client_pool: &Arc<SearchClientPool>,
+) -> Result<Vec<Vec<u8>>, SearchError> {
+    let start_instant = tokio::time::Instant::now();
```
fulmicoton, line 50: We could have use measure_time

quickwit-search/src/search_stream/root.rs
```diff
@@ -0,0 +1,186 @@
...
+        let handle = tokio::spawn(async move {
+            let mut receiver = search_client
+                .leaf_search_stream(leaf_request)
+                .await
+                .map_err(|search_error| NodeSearchError {
+                    search_error,
+                    split_ids: split_ids.clone(),
+                })?;
+
+            let mut leaf_bytes = Vec::new();
```
fulmicoton, line 84: that could be a Vec<Vec<u8>>

quickwit-serve/src/rest.rs
```diff
@@ -62,6 +65,15 @@ impl Default for Format {
     }
 }
 
+impl ToString for Format {
+    fn to_string(&self) -> String {
+        match &self {
+            Self::Json => "json".to_string(),
+            Self::PrettyJson => "prety-json".to_string(),
```
fulmicoton, line 72: s/prety/pretty

quickwit-serve/src/rest.rs
```diff
@@ -168,7 +180,94 @@ pub fn search_handler<TSearchService: SearchService>(
...
+    pub start_timestamp: Option<i64>,
+    /// If set, restrict search to documents with a `timestamp < end_timestamp``.
+    pub end_timestamp: Option<i64>,
+    /// The fast field to extract
+    #[serde(default)]
+    #[serde(rename(deserialize = "fastField"))]
+    pub fast_field: String,
+    /// The output format.
+    #[serde(default)]
+    #[serde(rename(deserialize = "outputFormat"))]
```
fulmicoton, line 208: camelCase does that for us already

quickwit-serve/src/rest.rs
```diff
@@ -168,7 +180,94 @@ pub fn search_handler<TSearchService: SearchService>(
...
+    #[serde(rename(deserialize = "searchField"))]
+    #[serde(deserialize_with = "from_simple_list")]
+    pub search_fields: Option<Vec<String>>,
+    /// If set, restrict search to documents with a `timestamp >= start_timestamp`.
+    pub start_timestamp: Option<i64>,
+    /// If set, restrict search to documents with a `timestamp < end_timestamp``.
+    pub end_timestamp: Option<i64>,
+    /// The fast field to extract
+    #[serde(default)]
+    #[serde(rename(deserialize = "fastField"))]
```
fulmicoton, line 204: camelCase does that for us already

quickwit-search/src/search_stream/mod.rs
fulmicoton: doing it

fulmicoton (approved):

PR #374, opened 2021-08-10
Export documentation

evanxg852000:
### Context / purpose
Closes https://github.com/quickwit-inc/quickwit/issues/366

### Description
Added documentation for export endpoint

### DIFF
--- a/docs/reference/search-api.md
+++ b/docs/reference/search-api.md
@@ -56,7 +56,7 @@ Search for documents matching a query in the given index `<index name>`.
 | Variable | Type | Description | Default value |
 |----------|------|-------------|---------------|
 | **query** | `String` | Query text. See the [query language doc](query-language.md) (mandatory) | |
-| **searchFields** | `[String]` | If set, specify the set of fields to operate the search on | |
+| **searchFields** | `[String]` | If set, specify the set of fields the search will performed on | |
 | **startTimestamp** | `i64` | If set, restrict search to documents with a `timestamp >= start_timestamp` | |
 | **endTimestamp** | `i64` | If set, restrict search to documents with a `timestamp < end_timestamp`` | |
 | **startOffset** | `Integer` | Number of documents to skip | `0` |
@@ -79,7 +79,7 @@ Search for documents matching a query in the given index `<index name>`.
 GET api/v1/indexes/<index name>/search/stream
 ```
 
-Export documents matching a search query in the given index `<index name>` in a specified output format.
+Export documents matching a search query in the given index `<index name>`, in a specified output format.
 
 The output format can be one of the following:
  -  [Comma Separated Values](https://datatracker.ietf.org/doc/html/rfc4180)
@@ -98,7 +98,7 @@ The output format can be one of the following:
 |----------|------|-------------|---------------|
 | **query** | `String` | Query text. See the [query language doc](query-language.md) (mandatory) | |
 | **fastField** | `String` | Name of fast field to retrieve from documents (mandatory)| |
-| **searchFields** | `[String]` | If set, specify the set of fields to operate the search on | |
+| **searchFields** | `[String]` | If set, specify the set of fields the search will performed on | |
 | **startTimestamp** | `i64` | If set, restrict search to documents with a `timestamp >= start_timestamp` | |
 | **endTimestamp** | `i64` | If set, restrict search to documents with a `timestamp < end_timestamp`` | |
 | **outputFormat** | `String` | Response output format. `csv` or `clickHouseRowBinary`  | `csv` | |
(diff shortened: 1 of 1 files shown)


docs/reference/search-api.md
```diff
@@ -53,13 +53,14 @@ Search for documents matching a query in the given index `<index name>`.
...
-| **startTimestamp**                 | `i64`              | If set, restrict search to documents with a `timestamp >= start_timestamp`                                                            |                                                                                |
-| **endTimestamp**                 | `i64`              | If set, restrict search to documents with a `timestamp < end_timestamp``                                                            |                                                                                     |
-| **startOffset**                | `Integer`              | Number of documents to skip                                                                | `0`                                                                                             |
-| **maxHits**                 | `Integer`              | Maximum number of hits to return (by default 20)                                                            | `20`                                                                                            |
+| Variable | Type | Description | Default value |
```
fmassot, line 56: why is there a diff here?

docs/reference/search-api.md
```diff
@@ -69,3 +70,40 @@ Search for documents matching a query in the given index `<index name>`.
...
+### Export search from an index
+
+```
+GET api/v1/indexes/<index name>/search/stream
+```
+
+Export documents matching a search query in the given index `<index name>`, in a specified output format.
+
+The output format can be one of the following:
+ -  [Comma Separated Values](https://datatracker.ietf.org/doc/html/rfc4180)
```
fmassot, line 85: CSV is better

docs/reference/search-api.md
```diff
@@ -69,3 +70,40 @@ Search for documents matching a query in the given index `<index name>`.
 | **hits**             | Results of the query           | `[hit]` |
 | **numHits**         | Total number of matches        |  `number`  |
 | **numMicrosecs**    | Processing time of the query   |  `number`  |
+
+
+
+### Export search from an index
```
fmassot, line 76: Let's call it search stream

docs/reference/search-api.md
```diff
@@ -69,3 +70,40 @@ Search for documents matching a query in the given index `<index name>`.
...
+
+```
+GET api/v1/indexes/<index name>/search/stream
+```
+
+Export documents matching a search query in the given index `<index name>`, in a specified output format.
+
+The output format can be one of the following:
+ -  [Comma Separated Values](https://datatracker.ietf.org/doc/html/rfc4180)
+ -  [ClickHouse Row Binary](https://clickhouse.tech/docs/en/interfaces/formats/#rowbinary)
```
fmassot, line 86: ClickHouse RowBinary?

docs/reference/search-api.md
```diff
@@ -69,3 +70,40 @@ Search for documents matching a query in the given index `<index name>`.
...
+| **fastField** | `String` | Name of fast field to retrieve from documents (mandatory)| |
+| **searchFields** | `[String]` | If set, specify the set of fields the search will performed on | |
+| **startTimestamp** | `i64` | If set, restrict search to documents with a `timestamp >= start_timestamp` | |
+| **endTimestamp** | `i64` | If set, restrict search to documents with a `timestamp < end_timestamp`` | |
+| **outputFormat** | `String` | Response output format. `csv` or `clickHouseRowBinary`  | `csv` | |
+
+
+### Response
+
+The response is a list of fast field values. The formatting is based on the specified output format.
```
fmassot, line 109: We can be more accurate. We return all the field values from documents that match the query. 
It works only if the field is marked as fast in the index config.

docs/reference/search-api.md
```diff
@@ -69,3 +70,40 @@ Search for documents matching a query in the given index `<index name>`.
...
+| ------------- | ------------- |
+| **index name** | The index name |
+
+
+#### Get parameters
+
+| Variable | Type | Description | Default value |
+|----------|------|-------------|---------------|
+| **query** | `String` | Query text. See the [query language doc](query-language.md) (mandatory) | |
+| **fastField** | `String` | Name of fast field to retrieve from documents (mandatory)| |
```
fmassot, line 100: It's the name of a field. And this field must be marked as "fast" in the config.

fmassot (changes requested):

docs/reference/search-api.md
fmassot: Fine, generally it's better to isolate the formatting from other changes because diff are useless then. But that's ok for this one ;)

evanxg852000: Ok, for cross check these are the lines I added:
```
| **searchFields** | `[String]` | If set, specify the set of fields the search will be performed on | |
| **format** | `String` | Response output format. `json` or `pretyjson`  | `pretyjson` |
```

fmassot (approved):

PR #375, opened 2021-08-10
Clean up code in field mapping parsing functions

guilload:
### Context / purpose
Code cleanup

### Description
I am not a fan of the following pattern, which needlessly creates an intermediate vector for error handling purposes:
```rust
values
.iter()
.map(...)
.collect::<Result<Vec<_>, _>>()?
.into_inter()
.flatten()
.collect()
```
Instead, let's rely on `process_results` from `itertools`. I also fixed some documentation along the way.

### How was this PR tested?
No functional changes were made in this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2719,6 +2719,7 @@ dependencies = [
  "base64",
  "chrono",
  "dyn-clone",
+ "itertools 0.10.1",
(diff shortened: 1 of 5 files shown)


fmassot (approved):
Nice, didn't know about process_results

PR #377, opened 2021-08-11
Try to guess AWS_REGION from ec2 instance metadata

r-zenine:
### Context / purpose
When running quickwit on an EC2, we still need to indicate the region `AWS_REGION` in order to properly use the IAM role associated with the EC2 machine. 
This PR, leverage the ec2 instance metadata to get the `AWS_REGION` automatically when running on EC2. 

### Description

### How was this PR tested?
We tested this PR on our EC2 instances that are running quickwit. We unset all the environment variables related to `AWS` and we try to Index data to `S3`

### Checklist
- [ X ] tested locally
- [ ] added unit tests
- [ ] included documentation

### DIFF
--- a/quickwit-storage/src/object_storage/s3_compatible_storage_uri_resolver.rs
+++ b/quickwit-storage/src/object_storage/s3_compatible_storage_uri_resolver.rs
@@ -20,7 +20,7 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-use std::sync::Arc;
+use std::{str::FromStr, sync::Arc};
 
 pub use rusoto_core::Region;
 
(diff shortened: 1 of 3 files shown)


quickwit-storage/src/object_storage/s3_compatible_storage_uri_resolver.rs
```diff
@@ -45,7 +45,10 @@ impl S3CompatibleObjectStorageFactory {
 
 impl Default for S3CompatibleObjectStorageFactory {
     fn default() -> Self {
-        S3CompatibleObjectStorageFactory::new(Region::default(), "s3")
+        let client = ec2_instance_metadata::InstanceMetadataClient::new();
+        let region_str = client.get().map(|e| e.region).unwrap_or_default();
+        let region = Region::from_str(region_str).unwrap_or(Region::default());
```
fulmicoton, line 50: unwrap_or_default

fmassot:
Great. One remark though: what if a user wants to connect to.a bucket in a different region from the ec2 instance? We won't be able to choose the region from the env variable anymore from what I understand.

Currently the default region is like this:
```
impl Default for Region {
    fn default() -> Region {
        match std::env::var("AWS_DEFAULT_REGION").or_else(|_| std::env::var("AWS_REGION")) {
            Ok(ref v) => Region::from_str(v).unwrap_or(Region::UsEast1),
            Err(_) => match ProfileProvider::region() {
                Ok(Some(region)) => Region::from_str(&region).unwrap_or(Region::UsEast1),
                _ => Region::UsEast1,
            },
        }
    }
}
``` 

I think we should do the additional check on ec2 instance region after checking env variable and profile variable but before using the default value `UsEast1`.

fulmicoton:
I opened a PR in quickwit with an extra commit that makes the logic more documented and adds Francois request.
https://github.com/quickwit-inc/quickwit/pull/379

PR #378, opened 2021-08-11
First stab at the indexer

fulmicoton:
This first PR introduce an actor framework and a draft of the indexing pipeline expressed using these actors.
- source
- indexer
- packager
- uploader
- publisher.


quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+}
+
+impl From<SendError> for ActorTermination {
+    fn from(_: SendError) -> Self {
+        ActorTermination::DownstreamClosed
+    }
+}
+
+/// An actor has an internal state and processes a stream of message.
```
PSeitz, line 70: I think it would be good to the add why we have this and where it's used (maybe also in the `README.md` in the project)

e.g. these actors are used only for indexing I guess, I that case `IndexingActor` would be a better name

fulmicoton: Right now it is only for indexing yes. But I prefer to separate the concerns and keep `quickwit-actors` entirely
agnostic. There are no real reasons why it shouldn't be... 

And if it is not killed by Darwinism, it will very likely be reused in other parts of quickwit too.

fulmicoton: I definitely need to add a README.

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub fn mailbox(&self) -> &Mailbox<A::Message> {
+        &self.self_mailbox
+    }
+
+    pub fn actor_instance_name(&self) -> &str {
+        &self.actor_instance_name
+    }
+
+    /// This function returns a guard that prevents any supervisor from identifying the
+    /// actor has dead.
```
PSeitz, line 162: as dead

quickwit-index/src/actors/indexer.rs
```diff
@@ -0,0 +1,364 @@
...
+            indexed_split.num_docs += 1;
+        }
+        counters.num_docs_in_split = indexed_split.num_docs;
+        Ok(())
+    }
+}
+
+pub struct Indexer {
+    immutable_state: ImmutableState,
+    sink: Mailbox<IndexedSplit>,
```
fmassot, line 137: replacing `sink` by `packager_mailbox` would be nicer

quickwit-actors/src/mailbox.rs
```diff
@@ -0,0 +1,271 @@
...
+    pub(crate) sender: flume::Sender<ActorMessage<Message>>,
+    command_sender: flume::Sender<Command<Message>>,
+    id: Uuid,
+    actor_name: String,
+}
+
+pub enum Command<Msg> {
+    Pause,
+    Stop(oneshot::Sender<()>),
+    Start,
```
fulmicoton, line 70: Rename resume

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+        self.actor_state.terminate();
+    }
+}
+
+impl<A: SyncActor> ActorContext<A> {
+    /// Sends a message to the actor being the mailbox.
+    ///
+    /// This method hides logic to prevent an actor from being identified
+    /// as frozen if the destination actor channel is saturated, and we
+    /// are simply experiencing back pressure.
```
PSeitz, line 217: We need to make sure to avoid cycles, a pipeline should only always go forward

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+        let _guard = self.protect_zone();
+        debug!(from=%self.self_mailbox.actor_instance_name(), to=%mailbox.actor_instance_name(), msg=?msg, "send");
+        mailbox.send_message_blocking(msg)
+    }
+
+    /// Sends a message to itself.
+    /// Since it is very easy to deadlock an actor, the behavior is quite
+    /// different from `send_message_blocking`.
+    ///
+    /// The method not
```
PSeitz, line 232: incomplete comment

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub fn send_self_message_blocking(&self, msg: A::Message) -> Result<(), crate::SendError> {
+        debug!(self=%self.self_mailbox.actor_instance_name(), msg=?msg, "self_send");
+        self.self_mailbox
+            .sender
+            .try_send(ActorMessage::Message(msg))
+            .map_err(|_| crate::SendError::WouldDeadlock)?;
```
PSeitz, line 238: keeping the original error would be good, e.g. an infinite call would not be a deadlock

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+        mailbox: &Mailbox<M>,
+        msg: M,
+    ) -> Result<(), crate::SendError> {
+        let _guard = self.protect_zone();
+        debug!(from=%self.self_mailbox.actor_instance_name(), send=%mailbox.actor_instance_name(), msg=?msg);
+        mailbox.send_message(msg).await
+    }
+
+    /// `async` version of `send_self_message`
+    pub async fn send_self_message(&self, msg: A::Message) -> Result<(), crate::SendError> {
```
PSeitz, line 268: send_self_command_message?

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+use std::fmt;
+/// An Actor Handle serves as an address to communicate with an actor.
+///
+/// It is lightweight to clone it.
+/// If all actor handles are dropped, the actor does not die right away.
+/// It will process all of the message in its mailbox before being terminated.
+///
+/// Because `ActorHandle`'s generic types are Message and Observable, as opposed
+/// to the actor type, `ActorHandle` are interchangeable.
+/// It makes it possible to plug different implementations, have actor proxy etc.use std::fmt;
```
PSeitz, line 13: remove use std::fmt;

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+                interval.tick().await;
+                if !ctx.progress().harvest_changes() {
+                    if ctx.get_state() == ActorState::Terminated {
+                        return;
+                    }
+                    error!(actor=%actor_instance_name, "actor-timeout");
+                    ctx.kill_switch().kill();
+                    return;
+                }
+            }
```
PSeitz, line 55: we could also use the slightly weird do while syntax

```
while {
   // do
    if !ctx.progress().harvest_changes() {
        if ctx.get_state() == ActorState::Terminated {
            return;
        }
        error!(actor=%actor_instance_name, "actor-timeout");
        ctx.kill_switch().kill();
        return;
    }
    interval.tick().await;
   // while
    ctx.kill_switch().is_alive()
} {}
```

quickwit-actors/src/progress.rs
```diff
@@ -0,0 +1,183 @@
...
+            }
+        }
+    }
+
+    /// This method mutates the state as follows and returns true if
+    /// the object was in the protected zone or had change registered.
+    /// - Updated -> NoUpdate, returns true
+    /// - NoUpdate -> Updated, returns true
+    /// - ProtectedZone -> ProtectedZone, returns true
+    pub fn harvest_changes(&self) -> bool {
```
PSeitz, line 114: had/registered_activity_since_last_call or similar would be a better name I think

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// the observable state of the actor after this.
+    ///
+    /// This method is mostly useful for tests.
+    ///
+    /// To actually observe the state of an actor for ops purpose,
+    /// prefer using the `.observe()` method.
+    ///
+    /// This method timeout if reaching the end of the message takes more than an HEARTBEAT.
+    /// Hence, it is only useful or unit tests.
+    pub async fn process_pending_and_observe(&self) -> Observation<A::ObservableState> {
```
PSeitz, line 74: since it is unit tests only, it could make sense to build a helper function using public methods, for a more realistic scenario

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+        let (tx, rx) = oneshot::channel();
+        let _ = self
+            .actor_context
+            .mailbox()
+            .send_command(Command::Stop(tx))
+            .await;
+        let _ = rx.await;
+    }
+
+    pub async fn join(self) -> Result<(ActorTermination, A::ObservableState), JoinError> {
```
PSeitz, line 107: it looks similar to finish, some comment to distinguish would be good

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+        let termination = self.join_handle.await?;
+        let observation = self.last_state.borrow().clone();
+        Ok((termination, observation))
+    }
+
+    /// Observe the current state.
+    ///
+    /// If a message is currently being processed, the observation will be
+    /// after its processing has finished.
```
PSeitz, line 116: /// The observation will be scheduled as a command message, therefore it will be executed
/// after the current active message and the current command queue have been processed.

quickwit-actors/src/actor_state.rs
```diff
@@ -0,0 +1,147 @@
...
+    Paused = 1,
+    Terminated = 2,
+}
+
+impl From<u32> for ActorState {
+    fn from(actor_u32: u32) -> Self {
+        match actor_u32 {
+            0 => ActorState::Running,
+            1 => ActorState::Paused,
+            _ => ActorState::Terminated,
```
PSeitz, line 37:
```
2 => ActorState::Terminated,
_ => panic!("could not parse actorstate"),
```
?

quickwit-actors/src/async_actor.rs
```diff
@@ -0,0 +1,160 @@
...
+        message: Self::Message,
+        ctx: &ActorContext<Self>,
+    ) -> Result<(), ActorTermination>;
+
+    async fn finalize(
+        &mut self,
+        _termination: &ActorTermination,
+        _ctx: &ActorContext<Self>,
+    ) -> anyhow::Result<()> {
+        Ok(())
```
PSeitz, line 49: todo?

quickwit-actors/src/mailbox.rs
```diff
@@ -0,0 +1,271 @@
...
+    pub rx: flume::Receiver<ActorMessage<Message>>,
+    command_rx: flume::Receiver<Command<Message>>,
+}
+
+#[derive(Debug)]
+pub enum ReceptionResult<M> {
+    Command(Command<M>),
+    Message(M),
+    None,
+    Disconnect, // was disconnect from either the command or the regular mailbox.
```
guilload, line 158: Disconnected?

quickwit-actors/src/async_actor.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+pub(crate) fn spawn_async_actor<A: AsyncActor>(
+    actor: A,
+    kill_switch: KillSwitch,
+    scheduler_mailbox: Mailbox<SchedulerMessage>,
```
PSeitz, line 56: is this the universes scheduler mailbox? or are there multiple scheduler?

quickwit-actors/src/mailbox.rs
```diff
@@ -0,0 +1,271 @@
...
+impl<Message: fmt::Debug> Inbox<Message> {
+    fn get_command_if_available(&self) -> Option<ReceptionResult<Message>> {
+        match self.command_rx.try_recv() {
+            Ok(command) => Some(ReceptionResult::Command(command)),
+            Err(TryRecvError::Disconnected) => Some(ReceptionResult::Disconnect),
+            Err(TryRecvError::Empty) => None,
+        }
+    }
+
+    pub(crate) async fn try_recv_msg(&self, message_enabled: bool) -> ReceptionResult<Message> {
```
guilload, line 170: I don't understand the concept of message enabled.

quickwit-actors/src/mailbox.rs
```diff
@@ -0,0 +1,271 @@
...
+pub struct Inbox<Message> {
+    pub rx: flume::Receiver<ActorMessage<Message>>,
+    command_rx: flume::Receiver<Command<Message>>,
+}
+
+#[derive(Debug)]
+pub enum ReceptionResult<M> {
+    Command(Command<M>),
+    Message(M),
+    None,
```
guilload, line 157: `Empty`, `NoMessage`?

quickwit-actors/src/mailbox.rs
```diff
@@ -0,0 +1,271 @@
...
+    }
+}
+
+pub struct Inbox<Message> {
+    pub rx: flume::Receiver<ActorMessage<Message>>,
+    command_rx: flume::Receiver<Command<Message>>,
+}
+
+#[derive(Debug)]
+pub enum ReceptionResult<M> {
```
guilload, line 154: Can we rename this enum `ReceiveResult` or `RecvResult`? `ReceptionResult` could be both interpreted as the result of trying to send or receive something. Looks like we're only using this enum for the latter here, so we might as well make it clearer.

quickwit-actors/src/progress.rs
```diff
@@ -0,0 +1,183 @@
...
+                Ordering::Relaxed,
+                Ordering::Relaxed,
+            )
+            .unwrap_or_else(|previous_value| previous_value)
+            .into();
+        previous_state != ProgressState::NoUpdate
+    }
+}
+
+pub struct ProtectZoneGuard(Arc<AtomicU32>);
```
guilload, line 129: ProtectedZoneGuard

quickwit-actors/src/universe.rs
```diff
@@ -0,0 +1,178 @@
...
+use crate::QueueCapacity;
+use crate::Scheduler;
+use crate::SyncActor;
+
+/// Universe serves as the top-level context in which Actor can be spawned.
+/// It is *not* a singleton. A typical application will usually have only one universe hosting all of the actors
+/// but it is not a requirement.
+///
+/// In particular, unit test all have their own universe and hence can be executed in parallel.
+pub struct Universe {
```
guilload, line 41: Fancy name ;)

guilload:
I left a few comments and suggestions, but I'm barely halfway through. I'll drop a commit on top of your PR to fix the typos if you don't mind.

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+        &self.self_mailbox
+    }
+
+    pub fn actor_instance_name(&self) -> &str {
+        &self.actor_instance_name
+    }
+
+    /// This function returns a guard that prevents any supervisor from identifying the
+    /// actor has dead.
+    /// The protection last as long as the `ProtectZoneGuard` it returns.
```
guilload, line 163:
```suggestion
    /// The protection ends when the `ProtectZoneGuard` is dropped.
```

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+    ///
+    /// In an ideal world, you should never need to call this function.
+    /// It is only useful in some corner like, like a calling a long blocking
+    /// from an external library that you trust.
+    pub fn protect_zone(&self) -> ProtectZoneGuard {
+        self.progress.protect_zone()
+    }
+
+    /// Get a copy of the actor kill switch.
```
guilload, line 172:
```suggestion
    /// Gets a copy of the actor kill switch.
```

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// For instance, when quitting from the process_message function, prefer simply
+    /// returning `Error(ActorTermination::Failure(..))`
+    pub fn kill_switch(&self) -> &KillSwitch {
+        &self.kill_switch
+    }
+
+    pub(crate) fn progress(&self) -> &Progress {
+        &self.progress
+    }
+    /// Record some progress.
```
guilload, line 183:
```suggestion
    /// Records some progress.
```

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// This function is only useful when implementing actors that may takes more than
```
guilload, line 184:
```suggestion
    /// This function is only useful when implementing actors that may take more than
```

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+            .mailbox()
+            .send_actor_message(ActorMessage::Observe(tx))
+            .await
+            .is_err()
+        {
+            error!("Failed to send message");
+        }
+        // TODO The timeout is required here. If the actor fails, the inbox is properly dropped but the send channel might actually
+        // prevent the onechannel Receiver from being dropped.
+        let observable_state_res = tokio::time::timeout(crate::HEARTBEAT, rx).await;
```
guilload, line 87: I guess that's fine but as is but wouldn't it be nicer to have the `observable_state_res` contain the observed state? I find this API a bit clunky. Plus, what if in between the two calls, the state of the actor is modified and `observe` has been called somewhere else.

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+        let state = self.last_state.borrow().clone();
```
guilload, line 88:
```suggestion
        let state = self.last_observation();
```

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+            .actor_context
+            .mailbox()
+            .send_command(Command::Observe(tx))
+            .await
+            .is_err()
+        {
+            error!("Failed to send message");
+        }
+        let observable_state_or_timeout = timeout(crate::HEARTBEAT, rx).await;
+        let state = self.last_state.borrow().clone();
```
guilload, line 129:
```suggestion
        let state = self.last_observation();
```

quickwit-actors/src/async_actor.rs
```diff
@@ -0,0 +1,160 @@
...
+use crate::{Actor, ActorContext, KillSwitch, Mailbox, ReceptionResult};
+use anyhow::Context;
+use async_trait::async_trait;
+use tokio::sync::watch::{self, Sender};
+use tracing::{debug, error};
+
+/// An async actor is executed on a regular tokio task.
+///
+/// It can make async calls, but it should not block.
+/// Actors doing CPU heavy work should implement `SyncActor` instead.
```
guilload, line 15:
```suggestion
/// Actors doing CPU-heavy work should implement `SyncActor` instead.
```

quickwit-actors/src/async_actor.rs
guilload: most likely a default impl.

quickwit-index/src/actors/indexer.rs
```diff
@@ -0,0 +1,364 @@
...
+    ) -> Result<(), ActorTermination> {
+        let indexed_split = self.get_or_create_current_indexed_split(current_split_opt, ctx)?;
+        for doc_json in batch.docs {
+            indexed_split.size_in_bytes += doc_json.len() as u64;
+            let doc = if let Some(doc) = self.parse_doc(&doc_json, counters) {
+                doc
+            } else {
+                continue;
+            };
+            if let Some(timestamp) = extract_timestamp_if_any(&doc, self.timestamp_field_opt) {
```
fmassot, line 123: note: if we have a timestamp field on the index, it could be nice to fail if the field is not the document.

quickwit-index/src/actors/packager.rs
```diff
@@ -0,0 +1,331 @@
...
+use quickwit_actors::QueueCapacity;
+use quickwit_actors::SyncActor;
+use quickwit_directories::write_hotcache;
+use quickwit_directories::HOTCACHE_FILENAME;
+use tantivy::SegmentId;
+use tantivy::SegmentMeta;
+use tracing::info;
+
+pub struct Packager {
+    sink: Mailbox<PackagedSplit>,
```
fmassot, line 40:
```suggestion
    uploader_mailbox: Mailbox<PackagedSplit>,
```

quickwit-actors/src/universe.rs
```diff
@@ -0,0 +1,178 @@
...
+            .scheduler_mailbox
+            .send_message(SchedulerMessage::SimulateAdvanceTime {
+                time_shift: TimeShift::ByDuration(duration),
+                tx,
+            })
+            .await;
+        let _ = rx.await;
+    }
+
+    pub fn spawn<A: AsyncActor>(&self, actor: A) -> (Mailbox<A::Message>, ActorHandle<A>) {
```
guilload, line 84: How does an actor spawn an actor? As you know, a Kafka source is an actor that manages a fleet of actors reading Kafka partitions. So we're going to need that.

quickwit-index/src/actors/uploader.rs
```diff
@@ -0,0 +1,316 @@
...
+use tantivy::SegmentId;
+use tokio::sync::oneshot::Receiver;
+use tracing::info;
+
+pub const MAX_CONCURRENT_SPLIT_UPLOAD: usize = 3;
+
+pub struct Uploader {
+    metastore: Arc<dyn Metastore>,
+    index_storage: Arc<dyn Storage>,
+    sink: Mailbox<Receiver<UploadedSplit>>,
```
fmassot, line 53:
```suggestion
    publisher_mailbox: Mailbox<Receiver<UploadedSplit>>,
```

quickwit-actors/src/mailbox.rs
fmassot: Yes, it's not obvious. My understanding is that it avoids making a if else when the actor is paused by passing a boolean to `try_recv_msg`. So whatever the actor state, we are just calling this function.
Personnally I would prefer to make the `if paused return None else try_recv_msg` but I'm ok to keep it like this.

fmassot (approved):

quickwit-actors/src/actor.rs
fulmicoton: I don't understand your comment.

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+        ctx: ActorContext<A>,
+    ) -> Self {
+        let mut interval = tokio::time::interval(crate::HEARTBEAT);
+        let actor_instance_name = ctx.actor_instance_name().to_string();
+        let ctx_clone = ctx.clone();
+        tokio::task::spawn(async move {
+            interval.tick().await;
+            while ctx.kill_switch().is_alive() {
+                interval.tick().await;
+                if !ctx.progress().harvest_changes() {
```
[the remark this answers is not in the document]
fulmicoton: You are absolutely right. This is code that we need to remove once we have a supervisor.

quickwit-actors/src/async_actor.rs
fulmicoton: yes

quickwit-actors/src/async_actor.rs
fulmicoton: One scheduler per universe

quickwit-actors/src/actor.rs
fulmicoton:
```suggestion
    /// actor as dead.
```

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub fn actor_instance_name(&self) -> &str {
+        &self.actor_instance_name
+    }
+
+    /// This function returns a guard that prevents any supervisor from identifying the
+    /// actor has dead.
+    /// The protection last as long as the `ProtectZoneGuard` it returns.
+    ///
+    /// In an ideal world, you should never need to call this function.
+    /// It is only useful in some corner like, like a calling a long blocking
```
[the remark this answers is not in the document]
fulmicoton:
```suggestion
    /// It is only useful in some corner case, like calling a long blocking
```

quickwit-actors/src/actor_handle.rs
fulmicoton:
```suggestion
/// It makes it possible to plug different implementations, have actor proxy etc.
```

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+/// - sync actors, executed on the blocking thread pool of tokio runtime.
+pub trait Actor: Send + Sync + 'static {
+    /// Type of message that can be received by the actor.
+    type Message: Send + Sync + fmt::Debug;
+    /// Piece of state that can be copied for assert in unit test, admin, etc.
+    type ObservableState: Send + Sync + Clone + fmt::Debug;
+    /// A name identifying the type of actor.
+    /// It does not need to be "instance-unique", and can be the name of
+    /// the actor implementation.
+    fn name(&self) -> String {
```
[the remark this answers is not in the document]
fulmicoton: Good idea. I added it as an individual commit you can check here.
https://github.com/quickwit-inc/quickwit/pull/378/commits/60ada2cfc7a382fd4d3b79324d8b1cd4c9a1a0b4

quickwit-actors/src/actor.rs
fulmicoton: That's a famous problem associated with actors in general yes.

quickwit-actors/src/actor.rs
fulmicoton: We can change this if there is an actual need for it.

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub fn send_self_message_blocking(&self, msg: A::Message) -> Result<(), crate::SendError> {
+        debug!(self=%self.self_mailbox.actor_instance_name(), msg=?msg, "self_send");
+        self.self_mailbox
+            .sender
+            .try_send(ActorMessage::Message(msg))
+            .map_err(|_| crate::SendError::WouldDeadlock)?;
+        Ok(())
+    }
+
+    pub fn schedule_self_msg_blocking(&self, after_duration: Duration, msg: A::Message) {
```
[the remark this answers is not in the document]
fulmicoton: I don't understand this message.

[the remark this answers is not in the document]
fulmicoton: Ah you are wondering why this method does not exist?
Well we don't need it yet.

quickwit-actors/src/actor_handle.rs
fulmicoton: I did not understand your comment.

This method can be useful outside of unit tests.

PSeitz: Then the comment is incorrect
`/// Hence, it is only useful or unit tests.`

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+    actor_state::{ActorState, AtomicState},
+    progress::{Progress, ProtectZoneGuard},
+    AsyncActor, KillSwitch, Mailbox, QueueCapacity, SendError, SyncActor,
+};
+
+// While the lack of message cannot pause a problem with heartbeating,  sending a message to a saturated channel
+// can be interpreted as a blocked actor.
+
+#[derive(Error, Debug)]
+pub enum ActorTermination {
```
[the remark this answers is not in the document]
fulmicoton: Ok for ActorExitStatus. (I'd rather avoid Code because we have more than a code when it is an enum

quickwit-actors/src/actor.rs
```diff
@@ -0,0 +1,316 @@
...
+    type Message: Send + Sync + fmt::Debug;
+    /// Piece of state that can be copied for assert in unit test, admin, etc.
+    type ObservableState: Send + Sync + Clone + fmt::Debug;
+    /// A name identifying the type of actor.
+    /// It does not need to be "instance-unique", and can be the name of
+    /// the actor implementation.
+    fn name(&self) -> String {
+        type_name::<Self>().to_string()
+    }
+    /// Actor mailbox queue capacity. It is set when the actor is spawned.
```
PSeitz, line 88:
```suggestion
    /// The Actor's incoming mailbox queue capacity. It is set when the actor is spawned.
```

quickwit-actors/src/actor.rs
PSeitz: It looked like it was scheduling a command message

quickwit-actors/src/actor_handle.rs
fulmicoton: Oh yes! there are two lines saying different things. Thanks.

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Process all of the pending messages, and returns a snapshot of
+    /// the observable state of the actor after this.
+    ///
+    /// This method is mostly useful for tests.
+    ///
+    /// To actually observe the state of an actor for ops purpose,
+    /// prefer using the `.observe()` method.
+    ///
+    /// This method timeout if reaching the end of the message takes more than an HEARTBEAT.
+    /// Hence, it is only useful or unit tests.
```
fulmicoton, line 68:
```suggestion
```

quickwit-actors/src/progress.rs
fulmicoton: agreed

quickwit-actors/src/progress.rs
fulmicoton: agreed

quickwit-index/src/actors/source/vec_source.rs
```diff
@@ -0,0 +1,110 @@
...
+        self.next_item_id += line_docs.len();
+        if line_docs.is_empty() {
+            return Err(ActorExitStatus::Success);
+        }
+        let batch = RawDocBatch {
+            docs: line_docs,
+            // checkpoint: Checkpoint::default(),
+        };
+        ctx.send_message(&self.sink, batch).await?;
+        ctx.send_self_message(()).await?;
```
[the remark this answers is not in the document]
guilload: I see, this is how you create a producer and keeps the actor spinning out messages.

quickwit-index/src/actors/source/vec_source.rs
```diff
@@ -0,0 +1,110 @@
...
+        let (mailbox, inbox) = create_test_mailbox();
+        let vec_source = VecSource::new(
+            std::iter::repeat_with(|| "{}".to_string())
+                .take(100)
+                .collect(),
+            3,
+            mailbox,
+        );
+        let (vec_source_mailbox, vec_source_handle) = universe.spawn(vec_source);
+        universe.send_message(&vec_source_mailbox, ()).await?;
```
guilload, line 102: and so this kickstarts the source.

quickwit-actors/src/mailbox.rs
fulmicoton: It was entirely remove in a refactoring that really simplified the mailbox.

quickwit-actors/src/mailbox.rs
fulmicoton: I am not sure this helped but I move the logic outside.

The idea is that when the actor is paused, we stop processing regular messages, but we keep processing command.
One glitchy spec is that we also process scheduled message. (I'm not sure this is a good thing....)

quickwit-actors/src/actor_handle.rs
fulmicoton: So join is similar to thread.join(). You wait for the actor to end by itself for one reason or another.

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+use crate::mailbox::Command;
+use crate::{Actor, ActorContext, ActorTermination, Observation};
+use std::fmt;
+/// An Actor Handle serves as an address to communicate with an actor.
+///
+/// It is lightweight to clone it.
+/// If all actor handles are dropped, the actor does not die right away.
+/// It will process all of the message in its mailbox before being terminated.
+///
+/// Because `ActorHandle`'s generic types are Message and Observable, as opposed
```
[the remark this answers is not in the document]
fulmicoton: Oh yes you are right. I fixed the comment.

quickwit-index/src/actors/indexer.rs
fulmicoton: done

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+        let observable_state_or_timeout = timeout(crate::HEARTBEAT, rx).await;
+        let state = self.last_state.borrow().clone();
+        match observable_state_or_timeout {
+            Ok(Ok(())) => Observation::Running(state),
+            Ok(Err(_)) => Observation::Terminated(state),
+            Err(_) => {
+                if self.actor_context.kill_switch().is_alive() {
+                    Observation::Timeout(state)
+                } else {
+                    self.join_handle.abort();
```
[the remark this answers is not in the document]
fulmicoton: Agreed

quickwit-actors/src/actor_handle.rs
fulmicoton: I address this. It is a bit quirky though as I did not want to add ObservableState to the mailbox generic parameter.
I ended up relying on `Box<dyn Any>`. (The casting is done by the `ActorHandle` so everything is safe)

quickwit-index/src/actors/indexer.rs
fulmicoton: Agreed and fixed.

quickwit-index/src/actors/source/vec_source.rs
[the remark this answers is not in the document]
fulmicoton: Yes, that's how you loop on a source.

quickwit-index/src/actors/source/vec_source.rs
fulmicoton: Yes, it is also possible to do that in the initialize function.

quickwit-actors/src/actor_handle.rs
```diff
@@ -0,0 +1,165 @@
...
+        if self
+            .actor_context
+            .mailbox()
+            .send_actor_message(ActorMessage::Observe(tx))
+            .await
+            .is_err()
+        {
+            error!("Failed to send message");
+        }
+        // TODO The timeout is required here. If the actor fails, the inbox is properly dropped but the send channel might actually
```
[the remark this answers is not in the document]
fulmicoton: yes

quickwit-actors/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+pub use kill_switch::KillSwitch;
+pub use observation::Observation;
+pub(crate) use scheduler::Scheduler;
+pub use sync_actor::SyncActor;
+pub use universe::Universe;
+
+/// Heartbeat used to verify that actors are progressing.
+///
+/// If an actor does not advertise a progress within an interval of duration `HEARTBEAT`,
+/// the killswith is hit, and all of the actors in this generation are killed.
```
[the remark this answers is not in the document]
fulmicoton: yes

quickwit-index/src/actors/indexer.rs
```diff
@@ -0,0 +1,364 @@
...
+    type Message = IndexerMessage;
+
+    type ObservableState = IndexerCounters;
+
+    fn observable_state(&self) -> Self::ObservableState {
+        self.counters.clone()
+    }
+
+    fn queue_capacity(&self) -> QueueCapacity {
+        QueueCapacity::Bounded(10)
```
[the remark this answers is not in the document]
fulmicoton: I think we need to test... but 10 seems okayish if the batches are not too big.

quickwit-index/src/actors/source/vec_source.rs
fulmicoton: I moved it in initialize for VecSource and FileSource. 

It uses a LoopMessage that cannot be initialized by the outside world to avoid bugs.

PR #379, opened 2021-08-11
From Ryad Zenine: try guess region on ec2 instance

fulmicoton:


fulmicoton:
Let's not squash this  one!

fmassot (approved):

PR #385, opened 2021-08-13
Draft to refactor query execution with retry 

fmassot:
Some notes I had in mind:
- we want a retry on 3 types of requests: fetch docs request, search request, search stream request
- we distribute those types of queries according to split id. Getting the list of queries to send to nodes is a sort of query plan
- it should be possible to share the code between those types of requests

### DIFF
--- a/quickwit-search/src/query_runner.rs
+++ b/quickwit-search/src/query_runner.rs
@@ -1,0 +1,184 @@
+use std::{collections::{HashMap, HashSet}, net::SocketAddr, sync::Arc};
+use async_trait::async_trait;
+use futures::{StreamExt, stream::FuturesUnordered};
+
+use itertools::{Itertools};
+use quickwit_metastore::{Metastore, SplitMetadata};
+use quickwit_proto::{FetchDocsRequest, FetchDocsResult, LeafSearchRequest, LeafSearchResult, PartialHit, SearchRequest};
+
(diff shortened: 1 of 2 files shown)


quickwit-search/src/query_runner.rs
```diff
@@ -0,0 +1,184 @@
...
+        }
+        let fetch_docs_req_jobs = request.partial_hits
+            .iter()
+            .map(|hit| Job {
+                split: hit.split_id.to_string(),
+                cost: 1,
+            })
+            .collect_vec();
+        let doc_fetch_jobs = self.client_pool
+            .assign_jobs(fetch_docs_req_jobs, &HashSet::default())
```
fulmicoton, line 167: The logic we have today is not like that, is it? Today we reassign the job to the same leaf that returned the result I think

fulmicoton:
I think I would remove the root out of the picture here and just focus on the problem of "how do I refactor the code that dispatch tasks to leaves, get back the results and retries if necessary".

The root stuff seems less easy to refactor, and also yields less benefit I believe.

... also I would ideally find another word than `ExecutionPlan`. That term has usually a different meaning.
Maybe QueryPlacement?

fmassot:
Yes, this is too early to refactor as is. With issue #422 it's clear that we need some scheduling and some query prioritization strategy.

fmassot:
Closing it in favor of #498

PR #392, opened 2021-08-16
Added checkpoint.

fulmicoton:
Quickwit indexes ordered messages queues (Pulsar, Kinesis, Kafka) as first class citizen.

All such product have in common the idea of a partitioned message queue.
Each message gets a position/offset/sequenceNumber within a partition and within a partition, messages are strictly ordered.

Quickwit will store these checkpoint on publish, in order to offer exactly once semantics for all sources of this type.
This PR introduces the idea of checkpoint, and checkpoint delta, in order to pass this position information through the indexer pipeline.


quickwit-index/src/actors/indexer.rs
```diff
@@ -160,6 +160,10 @@ impl ImmutableState {
         ctx: &ActorContext<Indexer>,
     ) -> Result<(), ActorExitStatus> {
         let indexed_split = self.get_or_create_current_indexed_split(current_split_opt, ctx)?;
+        indexed_split
+            .checkpoint_delta
+            .add(batch.checkpoint_delta)
+            .with_context(|| "Batch delta does not follow indexer checkpoint")?;
```
PSeitz, line 166: It would be better to keep the error details here

PSeitz: and add a test for this use case, they may be useful, when we have distributed indexing

quickwit-index/src/actors/source/file_source.rs
```diff
@@ -47,35 +51,44 @@ impl fmt::Debug for Loop {
     }
 }
 
+#[derive(Default, Clone, Debug, Eq, PartialEq)] //< TODO default should never be used I think? From<Checkpoint> is what we want.
```
PSeitz, line 54: some comments what it's used for would be good

quickwit-index/src/actors/source/file_source.rs
```diff
@@ -92,10 +105,10 @@ impl AsyncActor for FileSource {
         _message: Self::Message,
         ctx: &ActorContext<Self>,
     ) -> Result<(), ActorExitStatus> {
-        let limit_num_bytes = self.file_position.num_bytes + BATCH_NUM_BYTES_THRESHOLD;
+        let limit_num_bytes = self.counters.previous_offset + BATCH_NUM_BYTES_THRESHOLD;
```
PSeitz, line 108: I would add short comment on what the method is doing
// collecting batches of documents and sending the batch to the indexer

quickwit-metastore/src/checkpoint.rs
```diff
@@ -0,0 +1,457 @@
...
+/// The nature of the position may very depending on the source.
+/// Each source needs to encode it as a String in such a way that
+/// the lexicographical order matches the natural order of the
+/// position.
+///
+/// For instance, for u64 a 0-left-padded decimal representation
+/// can be used. Alternatively a base64 representation of their
+/// Big Endian representation can be used.
+#[derive(Debug, Default, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub struct Position(pub Arc<String>);
```
PSeitz, line 75: In which case do we need `String` for the partition offsets?

quickwit-metastore/src/checkpoint.rs
```diff
@@ -0,0 +1,457 @@
...
+    }
+}
+
+/// A checkpoint is a map of the last processed position for every partition.
+///
+/// If a partition is missing, it implicitely means that none of its message
+/// has been processed.
+#[derive(Default, Clone)]
+pub struct Checkpoint {
+    per_partition: BTreeMap<PartitionId, Position>,
```
PSeitz, line 109: why no hashmap?

quickwit-metastore/src/checkpoint.rs
```diff
@@ -0,0 +1,457 @@
...
+                    });
+                }
+            }
+        }
+        Ok(())
+    }
+
+    /// Try and apply a delta.
+    ///
+    /// `strict_mode` controls whether we accept a delta that does not chain perfectly with the
```
PSeitz, line 190: there is no strict_mode?

quickwit-metastore/src/checkpoint.rs
```diff
@@ -0,0 +1,457 @@
...
+    /// current position.
+    ///
+    ///    |    Checkpoint & Delta       | Outcome                     |
+    ///    |-----------------------------|-----------------------------|
+    ///    | ]?..a] ]a..c]               | Compatible                  |
```
PSeitz, line 195: can we use rust range syntax here?

quickwit-metastore/src/error.rs
```diff
@@ -84,6 +86,10 @@ pub enum MetastoreError {
         /// Split that should have been staged.
         split_id: String,
     },
+
+    /// The index is actually
```
PSeitz, line 90: it looks like this sentence is not

quickwit-index/src/actors/source/file_source.rs
```diff
@@ -106,11 +119,22 @@ impl AsyncActor for FileSource {
...
-            self.file_position.num_bytes += num_bytes as u64;
-            self.file_position.line_num += 1u64;
+            docs.push(doc_line);
+            self.counters.current_offset += num_bytes as u64;
+            self.counters.num_lines_processed += 1;
         }
-        if !raw_doc_batch.docs.is_empty() {
+        if !docs.is_empty() {
+            let mut checkpoint_delta = CheckpointDelta::default();
+            checkpoint_delta.add_partition(
```
PSeitz, line 128: so currently we merge all arriving messages into one one index?
Where would the distributor be to group partitions. On the sender side, as an actor step or here by having multiple batches?

quickwit-metastore/src/checkpoint.rs
fulmicoton: I don't think there is a syntax for that kind of interval.

quickwit-metastore/src/checkpoint.rs
fulmicoton: The source I know of (Kinesis, Kafka, Pulsar) are using either u64, u128.
We use `String` here to stay agnostic.

I guess we could take a risk and use u128, and hope this is ok.

quickwit-index/src/actors/source/file_source.rs
fulmicoton: One source actor can possibly work on several partitions.
Several source can end up in a single index, they just need to work on different partitions.

That's how we will handle indexing distribution.

The coordination (deciding how many source are indexing, which partitions belongs to whom, and where to spawn them) will be the role of one leader, materialized by an actor on a single server.

mosuka (approved):

quickwit-metastore/src/checkpoint.rs
```diff
@@ -0,0 +1,461 @@
...
+    }
+}
+
+/// A checkpoint is a map of the last processed position for every partition.
+///
+/// If a partition is missing, it implicitely means that none of its message
+/// has been processed.
+#[derive(Default, Clone)]
+pub struct Checkpoint {
+    per_partition: BTreeMap<PartitionId, Position>,
```
PSeitz, line 109: why no hashmap?

quickwit-metastore/src/checkpoint.rs
```diff
@@ -0,0 +1,461 @@
...
+            if !is_last {
+                f.write_str(" ")?;
+            }
+        }
+        f.write_str(")")?;
+        Ok(())
+    }
+}
+
+/// A partition delta represents an interval (from, to] over a partition of a source.
```
PSeitz, line 234: how do we cover the first element in a source with (from, to]?

quickwit-metastore/src/checkpoint.rs
```diff
@@ -0,0 +1,461 @@
...
+    /// We accept a delta as long as it comes after the current checkpoint,
+    /// for all partitions.
+    ///
+    /// We accept a delta that is not perfected chained after a checkpoint,
+    /// as gaps may happen. For instance, assuming a Kafka source, if the indexing
+    /// pipeline is down for more than the retention period.
+    ///
+    ///    |    Checkpoint & Delta        | Outcome                     |
+    ///    |------------------------------|-----------------------------|
+    ///    | (?..a] (b..c] with a = b     | Compatible                  |
```
PSeitz, line 199: what is ? meaning here

quickwit-metastore/src/checkpoint.rs
fulmicoton: For ordering in serialization.

quickwit-metastore/src/checkpoint.rs
fulmicoton: `Position::default()`. This is the empty string.

PSeitz (approved):
LGTM

PR #394, opened 2021-08-16
Isolate SingleFileMetastore unit tests as an independent test suite in Metastore.

### LINKED ISSUE #388: Isolate SingleFileMetastore as a metastore independent test suite
We have a lot of unit tests in the SingleFileMetastore.

These tests would be relevant to any other implementation of a metastore.

We should isolate it as a suite function

```
fn test_suite_for_metastore(metastore: Arc<dyn Metastore>) -> anyhow::Result<()> {
   test_single_file_metastore_publish_split(&metastore)?;
   //...
}

### PR DESCRIPTION
**Pull request by mosuka**
Isolate SingleFileMetastore unit tests as an independent test suite in Metastore.

### Context / purpose
Close #388

### Description
Isolate SingleFileMetastore unit tests as an independent test suite in Metastore.

### Checklist
- [x] tested locally
- [x] added unit tests

### CODE REVIEW

**Comment by fulmicoton** [comment]
First remark:
You should move the unit test related to the single file metastore into the single file metastore file.
(and keep the generic tests you added in the test file)

That is still not great because it requires a lot of work to test a new metastore.
Ideally adding a new test should mean adding a test in one file, and it should apply to all metastore implementation.
Adding a new implementation should mean adding one line.

You can either:

Solution 1. Create a single unit test that wraps all other unit test.

```
trait Meta

#[test]
async fn test_metastore_suite<MetastoreToTest: Metastore>() {
   // ...
  {
     let metastore = MetastoreToTest::for_test();
     super::test_metastore_create_index<MetastoreToTest>(&metastore).await;
   }
   // ...
   {
       let metastore = MetastoreToTest::for_test();
        super::test_metastore_index_metadata(&metastore).await;
   }
}
```
The cons with this approach is that if one test file, it makes it difficult to know which test fails.
Also if the suite is not fast enough, it will be detected as a slow test.

Solution 2. Use a macro.

```
generate_unit_tests!(SingleFileMetaStore);
```

If you use the second solution, let's do it in two steps. We can isolate the test suite test and their calls, and then we can work on the macro

fulmicoton (approved):

PR #424, opened 2021-08-18
Collecting tags during Indexing

evanxg852000:
### Context / purpose
Closes https://github.com/quickwit-inc/quickwit/issues/421

### Description
Implement extracting tags during indexing in both pipelines. 

### How was this PR tested?
run the commands using this index config
```json
{
    "store_source": true,
    "default_search_fields": ["event"],
    "timestamp_field": "ts",
    "tag_fields": ["device", "city"],
    "field_mappings": [
        {
            "name": "ts",
            "type": "i64",
            "fast": true
        },
        {
            "name": "event",
            "type": "text",
            "stored": true,
            "indexed": true
        },
        {
            "name": "device",
            "type": "text",
            "stored": false
        },
        {
            "name": "city",
            "type": "text",
            "stored": false
        }
    ]
}
```

### Checklist
- [x] tested locally
- [x] added unit tests
- [ ] included documentation


quickwit-core/src/indexing/split.rs
```diff
@@ -243,6 +250,29 @@ impl Split {
         let manifest = put_split_files_to_storage(&*self.storage, self).await?;
         Ok(manifest)
     }
+
```
fulmicoton, line 253: I think we have this in the indexer and the packager right now. Am I missing something?

fulmicoton: Ah I see, this is because you wanted to populate tags in the current pipeline and the new one!

quickwit-index/src/actors/packager.rs
```diff
@@ -223,8 +250,9 @@ impl SyncActor for Packager {
     ) -> Result<(), quickwit_actors::ActorExitStatus> {
         commit_split(&mut split, ctx)?;
         let segment_metas = merge_segments_if_required(&mut split, ctx)?;
+        let tags = extract_tags(&mut split)?;
```
fulmicoton, line 253: I think we can move that call into create_packaged_split and remove one argument.

quickwit-index-config/src/default_index_config/default_config.rs
```diff
@@ -95,6 +97,30 @@ impl DefaultIndexConfigBuilder {
             }
         }
 
+        // Resolve tag fields
```
fulmicoton, line 100: There was a misunderstanding here. Apologies, the spec you went for absolutely makes sense but this is not really what we want.

Tags are based on field existing in the schema. Country etc. 
They can be of any type. (not necessarily string).

We want to add an extra field in the schema, with the reserved name "_tags". (we already have a reserved field called source).

The _tags uses a raw tokenizer and is indexed like the schema you gave below.

When we create the tantivy document, we want to populate this extra field using the tag fields as defined in the index config. We populate it with the "field:field_value" template you used  on extract_tags.

extract_tags on the other hand, becomes simpler. We only need to iterate through the terms of the _tags field.

fulmicoton (changes requested):
There was a misunderstanding in the spec. this is not major. You can make the change only for the actor pipeline.

quickwit-core/src/indexing/split.rs
evanxg852000: Yes, currently there are two indexing pipelines. I am aware this one in split will be replaced so I did not bother duplicating just thing are in to write tests for it.

quickwit-index-config/src/default_index_config/default_config.rs
evanxg852000: Do we still support multiple fields for tagging? I guess if so, the `_tags` ends up with a string value like
`field1:x, field1:y, field2:z` then when we get the term value, we split by `,` and add them to the split tags. we will have to deduplicate.

quickwit-index-config/src/default_index_config/default_config.rs
```diff
@@ -130,6 +149,9 @@ impl DefaultIndexConfigBuilder {
         if self.store_source {
             builder.add_text_field(SOURCE_FIELD_NAME, STORED);
         }
+        if !self.tag_fields.is_empty() {
+            builder.add_text_field(TAGS_FIELD_NAME, STRING);
```
fulmicoton, line 153: I think we should always add it. One benefit  is that someone could populate it by actually specifying it in the document.
What does @fmassot think?

fulmicoton (approved):

quickwit-index-config/src/default_index_config/default_config.rs
evanxg852000: Yeah, it makes sense, I would have agreed if we were not providing a remapping of other fields to the special `_tags`.
But since we are, I think it's better to let the user requests it via remapping.   
I already have the small changes but I will hold till @fmassot  gives us his take.

fmassot: I like the idea of always adding it:
- it means that the user can activate very easily this features by adding new tags in the index config
- as @fulmicoton said it, it could be useful to add tags on one document, basically you can tag your document without even adding a new field in your schema, this can be very flexible
- there is no hidden cost of always adding this field, so there is no risk of always adding this field right now

quickwit-cli/tests/cli.rs
```diff
@@ -216,10 +216,22 @@ fn test_cmd_search() -> Result<()> {
...
     .stdout(predicate::function(|output: &[u8]| {
         let result: Value = serde_json::from_slice(output).unwrap();
-        result["numHits"] == Value::Number(Number::from(0i64))
+        result["numHits"] == Value::Number(Number::from(2i64))
     }));
 
-    // TODO add another case when we have mechanism for tagging splits
+    make_command(
+        format!(
+            "search --index-uri {} --query level:info --tags city:conakry",
```
fmassot, line 224: nice city

quickwit-core/src/indexing/split.rs
```diff
@@ -171,6 +176,8 @@ impl Split {
                 );
             }
         };
+
+        self.metadata.size_in_bytes += doc_size as u64;
```
fmassot, line 180: Do we really want to have the size in bytes of the raw json docs? From a user that is looking at this metadata information on the split, I think this is misleading.
I would expect something like `SegmentSpaceUsage` given by `tantivy`.
What do you think? we can add an issue about this and leave the code as is.

quickwit-index-config/src/default_index_config/default_config.rs
```diff
@@ -214,11 +253,20 @@ impl IndexConfig for DefaultIndexConfig {
...
+            let field_name = field_path.field_name();
             let field = self
                 .schema
-                .get_field(&field_path.field_name())
-                .ok_or_else(|| DocParsingError::NoSuchFieldInSchema(field_path.field_name()))?;
+                .get_field(&field_name)
+                .ok_or_else(|| DocParsingError::NoSuchFieldInSchema(field_name.clone()))?;
+            if self.tag_field_names.contains(&field_name) {
+                let tags_field = tags_field_opt.ok_or_else(|| {
+                    DocParsingError::NoSuchFieldInSchema(TAGS_FIELD_NAME.to_string())
```
fmassot, line 265: I would raise the exception immediately when doing `self.schema.get_field(TAGS_FIELD_NAME)``

evanxg852000: Hmm, I preferred the option so that the calling client decides what Error to raise since we tend to do field resolution in two places.

quickwit-core/src/indexing/split.rs
evanxg852000: Yes, I think `SegmentSpaceUsage` is relevant for the split and its segments,  but I think having the overall size of data that was ingested is equally relevant and possibly displaying `doc_size`  vs split_size (composed of segment space usage) gives more send to the user

fmassot (changes requested):
Sound goods, but there is not test on tags in `default_config.rs` file. It would be nice to at least test tags here and don't forget to test a binary field.
I'm also worried because I don't see any test to check that a split has the right tags, I see there's something at the cli level but that's all. Can you fix that too?

quickwit-core/src/indexing/split.rs
fmassot: mmm, but the doc_size can be very different from the split. I will open an issue to discuss that, this is minor.

quickwit-index-config/src/default_index_config/default_config.rs
fmassot: Not sure to get it :). Anyway this is very minor.

fmassot (approved):

PR #433, opened 2021-08-19
Only keep the footer offsets in the metastore

fulmicoton:
### Context / purpose

This PR refactors the organisation of the split file.
The split file is now the concatenation

- concatenated files (does not contain hotcache anymore)
- FOOTER - bundle meta data (file offsets)
- FOOTER - bundle meta data len
- FOOTER - hotcache
- FOOTER - hotcache len

The element marked footer are what we need to download in order to open a split in one fetch.

### DIFF
--- a/quickwit-index/src/actors/uploader.rs
+++ b/quickwit-index/src/actors/uploader.rs
@@ -129,7 +129,7 @@ async fn put_split_files_to_storage(
     let elapsed_secs = start.elapsed().as_secs();
     let elapsed_ms = start.elapsed().as_millis();
     let file_statistics = split.file_statistics.clone();
-    let split_size_in_megabytes = split.bundle_offsets.bundle_file_size / 1000000;
+    let split_size_in_megabytes = split.footer_offsets.end / 1000000;
     let throughput_mb_s = split_size_in_megabytes as f32 / (elapsed_ms as f32 / 1000.0);
     info!(
         min_file_size_in_bytes = %file_statistics.min_file_size_in_bytes,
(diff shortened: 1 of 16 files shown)


quickwit-storage/src/bundle_storage.rs
```diff
@@ -327,12 +251,9 @@ mod tests {
...
         create_bundle.add_file(&test_file1_name)?;
         create_bundle.add_file(&test_file2_name)?;
-        let offsets = create_bundle.finalize()?;
-        assert!(offsets.footer_offsets.start > offsets.hotcache_offset_start);
-
-        let bytes = std::fs::read(&bundle_file_name)?;
+        create_bundle.finalize()?;
 
-        let metadata = BundleStorageFileOffsets::open(&bytes).unwrap();
+        let metadata = BundleStorageFileOffsets::open(&buffer)?;
```
PSeitz, line 256: while `?` is nicer, it doesn't give good stacktraces on error. we could fix that with an error that carries the stacktraces

fulmicoton: Yes I agree it is a bit lame when the test fails.

mosuka (approved):

quickwit-index/src/actors/uploader.rs
```diff
@@ -193,27 +164,18 @@ impl AsyncActor for Uploader {
         ctx.send_message(&self.publisher_mailbox, split_uploaded_rx)
             .await?;
 
-        // The permit will be added back manually to the semaphore the task after it is finished.
-        let permit_guard = self.concurrent_upload_permits.acquire().await;
+        {
+            // The permit will be added back manually to the semaphore the task after it is finished.
+            let _permit_guard = self.concurrent_upload_permits.acquire().await;
```
PSeitz, line 169: oom issues can lead to leaks here right?

PR #438, opened 2021-08-21
Fix bug in `CheckpointDelta::add`

guilload:
### Description
Fix bug in `CheckpointDelta::add`. When a delta was already recorded for a given partition, the current delta was overridden instead of being updated.

### How was this PR tested?
Added unit test.

### DIFF
--- a/quickwit-metastore/src/checkpoint.rs
+++ b/quickwit-metastore/src/checkpoint.rs
@@ -144,7 +144,7 @@ impl<'de> Deserialize<'de> for Checkpoint {
 /// Error returned when trying to apply a checkpoint delta to a checkpoint that is not
 /// compatible. ie: the checkpoint delta starts from a point anterior to
 /// the checkpoint.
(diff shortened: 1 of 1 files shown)


quickwit-metastore/src/checkpoint.rs
```diff
@@ -310,19 +310,19 @@ impl CheckpointDelta {
...
+            let entry = self.per_partition.entry(partition_id);
             match entry {
                 Entry::Occupied(mut occupied_entry) => {
                     if occupied_entry.get().to == partition_delta.from {
-                        occupied_entry.insert(partition_delta);
+                        occupied_entry.get_mut().to = partition_delta.to
                     } else {
                         return Err(IncompatibleCheckpoint {
-                            partition_id,
+                            partition_id: occupied_entry.key().clone(),
```
fulmicoton, line 325: nitpick: i don't understand the point of modifying this line.

fulmicoton (approved):

Issue #389, opened 2021-08-16
PostgreSQL based metastore 


fulmicoton:
Implement a metastore based on postgresql.

(You can refer to barrel for some of the implementation details)

PR #440, opened 2021-08-21
Add metastore-uri option to CLI.

mosuka:
### Context / purpose
Preparation for the addition of a PostgreSQL-based metastore.  #389

### Description
Add `metastore-uri` option to CLI.

Example:
```
quickwit new --index-config-path=./examples/index_configs/wikipedia_index_config.json \
    --index-uri=file:///home/minoru/tmp/quickwit_data/indices/test-idx \
    --metastore-uri=file:///home/minoru/tmp/quickwit_data/metastore
```

```
quickwit serve --host=0.0.0.0 --port=8080 \
    --index-uri=file:///home/minoru/tmp/quickwit_data/indices/test-idx \
    --index-uri=file:///home/minoru/tmp/quickwit_data/indices/test2-idx \
    --metastore-uri=file:///home/minoru/tmp/quickwit_data/metastore
```

### Checklist
- [x] tested locally


fmassot:
Very good idea to use now a `metastore-uri`! 
But now `index-uri` is confusing because we only need an index identifier to identify the index! With a `metastore-uri` we only need an `index-id` or `index` cli parameter.

fmassot (changes requested):
See my comment about the index-uri / index-id

mosuka:
Thank you for your comment!

I added that `metastore-uri` is for users to explicitly specify whether to store metadata in the existing File-based metastore or PostgreSQL-based metastore when the PostgreSQL-based metastore is added.
So I understood that it would be better to have the index location specified separately in `index-uri`.

The following example shows how to create both an index and a metastore in S3.
```
quickwit new --index-config-path=./examples/index_configs/wikipedia_index_config.json \
    --index-uri=s3://mybucket/indexes/test-idx \
    --metastore-uri=s3://mybucket/metastore
```

When a PostgreSQL-based metastore is avairable in the future, we can create the index in S3 and the metastore in PostgreSQL, as shown in the following example.
```
quickwit new --index-config-path=./examples/index_configs/wikipedia_index_config.json \
    --index-uri=s3://mybucket/indexes/wikipedia-idx \
    --metastore-uri=postgresql://metastore.xxxxxxx.us-west-1.amazonaws.com:5432
```

Can you tell me what CLI options you are imagining for reference?

fmassot:
Ah yes you're right, I totally missed that... we should obviously keep`index-uri` with the `metastore-uri`, sorry for the confusion :/

I was thinking about a s3 metastore which is currently located at the parent directory of the index-uri. Let's discuss that offline.

fmassot (approved):
Looks good.

mosuka:
Thank you for confirming!

PR #452, opened 2021-08-25
Use futures::stream in leaf functions

fmassot:
### Context / purpose
It could be useful to return `futures::stream` in `leaf` functions to:
- easily limit the number of `leaf_search_stream_single_split` or `leaf_search_single_split` executed concurrently
- easily convert a stream item error `SearchError` to tonic `Status` and vice versa, currently `leaf_search_stream` returns an UnboundedReceiver with a result containing tonic `status`. We don't want to have this status at the `leaf_search_stream` level.

### Description
This PR shows how to use streams in `leaf_search_stream` function and the subsequent changes we must apply to make the grpc layer compatible.

### Caveats
I did not manage to make the `SearchClient` returns a stream because of trait bounds error: the stream has to be Send + Sync but I only have a Send stream from tonic response. So I keep the unbounded receiver.

### DIFF
--- a/quickwit-search/src/client.rs
+++ b/quickwit-search/src/client.rs
@@ -19,6 +19,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+use futures::StreamExt;
 use http::Uri;
 use quickwit_proto::LeafSearchStreamResult;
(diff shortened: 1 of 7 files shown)


fmassot:
closing PR in favor of #487

PR #453, opened 2021-08-25
Increased multipart target size for s3

evanxg852000:
Closes https://github.com/quickwit-inc/quickwit/issues/448

### How was this PR tested?
Test with the Wikipedia corpus (8.9GB uncompressed) generating 2 splits with the largest file being `xxx.store` (6.2GB)
With this change uploading takes a while and suspect that it because of the 5Gb part size. I am proposing we balance 
to 2 or 3GB to get more concurrent uploads

### Checklist
- [x] tested locally


fulmicoton:
Can we change the Default implementation instead?

evanxg852000:
Yes, but my guess was that this Default will be tight to AWS s3 rules if I do so. I think ideally we should have a default for each object storage. But to keep thing simple let me just put it in the default.

PR #476, opened 2021-08-28
Replacing artillery-core by an internal fork.

fulmicoton:


fmassot:
Would be nice to add some tests. And I think someone else can code the tests.

fulmicoton:
@fmassot I did test it locally.

fmassot:
> @fmassot I did test it locally.

I think we can merge now and add an issue to add some tests. Maybe there were some tests in the original project that we can copy.

fmassot (approved):

Issue #484, opened 2021-08-29
Remove warmup done twice for the same fast field in search stream


fmassot:
When `timestamp_field` and `fast_field_to_collect` are the same, the warmup is done twice for the same field.

PR #487, opened 2021-08-29
Fix #483, #484 and above all handle errors happening in search stream leaves.

### LINKED ISSUE #483: Use spawn_blocking in leaf stream search
To efficiently parallelize tasks with tokio, cpu heavy tasks need to be spawn using `task::spawn_blocking`.

fmassot:
Fixes #483 #484 

Bonus: 
- the leaf stream return a search error and not a tonic status
- add tests at the root and leaf levels
- add test on the cli (csv output)

Not included:
- tests on collector: some important changes are coming soon and I would prefer to implement tests at this moment
- tests on performance would be nice

### DIFF
--- a/quickwit-cli/tests/cli.rs
+++ b/quickwit-cli/tests/cli.rs
@@ -589,6 +589,92 @@ async fn test_cmd_dry_run_delete_on_s3_localstack() -> Result<()> {
 /// testing the api via cli commands
 #[tokio::test]
 #[serial]
+async fn test_all_local_index() -> Result<()> {
+    // Implicit index_id defined in test env struct.
+    // TODO: change that after the metastore uri refactoring.
+    let index_id = "data";
+    let test_env = create_test_env(TestStorageType::LocalFileSystem)?;
+    make_command(
+        format!(
+            "new --index-uri {} --metastore-uri {} --index-config-path {}",
+            test_env.index_uri(index_id),
+            test_env.metastore_uri,
+            test_env.resource_files["config"].display()
+        )
+        .as_str(),
+    )
+    .assert()
+    .success();
+
+    let metadata_file_exist = test_env.storage.exists(&Path::new(index_id).join("quickwit.json")).await?;
+    assert_eq!(metadata_file_exist, true);
+
+    index_data(
+        index_id,
+        test_env.resource_files["logs"].as_path(),
+        &test_env.metastore_uri,
+    );
+
+    // serve & api-search
+    let mut server_process = spawn_command(
+        format!(
+            "serve --metastore-uri {} --host 127.0.0.1 --port 8182",
+            test_env.metastore_uri,
+        )
+        .as_str(),
+    )
+    .unwrap();
+    sleep(Duration::from_secs(2)).await;
+    let mut data = vec![0; 512];
+    server_process
+        .stdout
+        .as_mut()
+        .expect("Failed to get server process output")
+        .read_exact(&mut data)
+        .expect("Cannot read output");
+    let process_output_str = String::from_utf8(data).unwrap();
+    let query_response = reqwest::get(format!(
+        "http://127.0.0.1:8182/api/v1/{}/search?query=level:info",
+        index_id
+    ))
+    .await?
+    .text()
+    .await?;
+
+    assert!(process_output_str.contains("http://127.0.0.1:8182"));
+    let result: Value =
+        serde_json::from_str(&query_response).expect("Couldn't deserialize response.");
+    assert_eq!(result["numHits"], Value::Number(Number::from(2i64)));
+
+    let search_stream_response = reqwest::get(format!(
+        "http://127.0.0.1:8182/api/v1/{}/search/stream?query=level:info&outputFormat=csv&fastField=ts",
+        index_id
+    ))
+    .await?
+    .text()
+    .await?;
+    assert_eq!(search_stream_response, "2\n13\n");
+    
+    server_process.kill().unwrap();
+
+    make_command(
+        format!(
+            "delete --index-id {} --metastore-uri {}",
+            index_id, test_env.metastore_uri
+        )
+        .as_str(),
+    )
+    .assert()
+    .success();
+    let metadata_file_exist = test_env.storage.exists(&Path::new(index_id).join("quickwit.json")).await?;
+    assert_eq!(metadata_file_exist, false);
+
+    Ok(())
+}
+
+/// testing the api via cli commands
+#[tokio::test]
+#[serial]
 #[cfg_attr(not(feature = "ci-test"), ignore)]
 async fn test_all_with_s3_localstack_cli() -> Result<()> {
     let index_id = "s3_index_1";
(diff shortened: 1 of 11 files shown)


quickwit-search/src/search_stream/root.rs
```diff
@@ -212,4 +211,54 @@ mod tests {
...
+        result_sender.send(Ok(quickwit_proto::LeafSearchStreamResult {
+            data: b"123".to_vec(),
+        }))?;
+        result_sender.send(Err(SearchError::InternalError("error".to_string())))?;
+        mock_search_service.expect_leaf_search_stream().return_once(
+            |_leaf_search_req: quickwit_proto::LeafSearchStreamRequest| {
+                Ok(UnboundedReceiverStream::new(result_receiver))
+            },
+        );
+        // The test will hang on indefinitely if we don't drop the receiver.
```
evanxg852000, line 255: did you meant drop `sender`?

evanxg852000 (approved):

quickwit-search/src/search_stream/leaf.rs
```diff
@@ -46,39 +55,52 @@ pub async fn leaf_search_stream(
...
-            result_sender_clone.send(leaf_split_result).map_err(|_| {
-                SearchError::InternalError(format!(
-                    "Unable to send leaf export result for split `{}`",
-                    &split.split_id
-                ))
-            })?;
-            Result::<(), SearchError>::Ok(())
-        });
-    }
+    let request_clone = request.clone();
```
guilload, line 60: We could "own" that request and not clone it as you did in some other parts of this PR.

quickwit-search/src/search_stream/leaf.rs
```diff
@@ -46,39 +55,52 @@ pub async fn leaf_search_stream(
...
+    tokio::spawn(async move {
+        let mut stream =
+            leaf_search_results_stream(request_clone, storage, splits, index_config).await;
+        while let Some(item) = stream.next().await {
+            if let Err(error) = result_sender.send(item) {
```
guilload, line 65: Just my personal preference, but I like the `map_err(...)?` pattern better.

quickwit-search/src/search_stream/leaf.rs
```diff
@@ -46,39 +55,52 @@ pub async fn leaf_search_stream(
...
+    futures::stream::iter(splits)
+        .map(move |split| {
+            (
+                split,
+                index_config.clone(),
+                request.clone(),
+                storage.clone(),
+            )
+        })
+        .map(|(split, index_config_clone, request_clone, storage)| {
```
guilload, line 92: Why the double `map`? Readability?

guilload (approved):
After reading this PR, I vote for changing `SearchError::InternalError(String)` to `SearchError::InternalError(anyhow::Error)`.

guilload:
I mostly made syntactical comments. The functional changes are great and good job on those tests.

quickwit-search/src/client.rs
```diff
@@ -138,25 +138,35 @@ impl SearchServiceClient {
...
-                        .map_err(|status| parse_grpc_error(&status))?
-                    {
-                        // We want to stop doing unnecessary work on the leaves as soon as
-                        // there is an issue sending the result.
-                        // Terminating the task will drop the `result_stream` consequently
-                        // canceling the gRPC request.
-                        result_sender.send(Ok(result)).map_err(|_| {
-                            SearchError::InternalError("Could not send leaf result".into())
-                        })?;
+                    loop {
```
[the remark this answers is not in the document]
guilload: I think this is actually going to swallow the last error :x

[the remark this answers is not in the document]
guilload: This is what I want https://github.com/rust-lang/rust/issues/62208. Anyway t_t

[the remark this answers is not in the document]
fmassot: the last element of the stream is `None` so I think it should work, I really like your version :)

quickwit-search/src/client.rs
```diff
@@ -138,25 +138,35 @@ impl SearchServiceClient {
...
+                        if grpc_result.is_err() {
+                            result_sender
+                                .send(Err(parse_grpc_error(&grpc_result.unwrap_err())))
+                                .map_err(|_| {
+                                    SearchError::InternalError(
+                                        "Sender closed, could not send leaf result.".into(),
+                                    )
+                                })?;
+                            // Stop the loop as soon as we receive a leaf error.
+                            break;
```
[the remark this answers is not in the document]
fmassot: indeed

fmassot (commented):
> After reading this PR, I vote for changing `SearchError::InternalError(String)` to `SearchError::InternalError(anyhow::Error)`.

We serialize `SearchError` to pass it through the network so I'm not sure how to handle it on the serialization side. We could serialize the error as a string but then we're back to square one. Nevertheless, we could at least have real errors on the root because these errors are not serialized.

Right now, I have the feeling that we already need to improve the errors we are handling (`SearchError`, `NodeSearchError`), I found it a little bit messy.

quickwit-cli/tests/cli.rs
```diff
@@ -586,6 +586,98 @@ async fn test_cmd_dry_run_delete_on_s3_localstack() -> Result<()> {
...
+    .assert()
+    .success();
+    let metadata_file_exist = test_env
+        .storage
+        .exists(&Path::new(index_id).join("quickwit.json"))
+        .await?;
+    assert_eq!(metadata_file_exist, false);
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
fulmicoton: I think SearchError gets serialized/deserialized... anyhow cannot be serialized, hence the msg.

quickwit-search/src/search_stream/leaf.rs
fulmicoton: +1... Generally speaking if ownership is required, it is better to reflect it in the prototype. It does not necessarily avoid the clone, but at least it might.

quickwit-search/src/search_stream/leaf.rs
fulmicoton: I prefer the `if let Err` here.

The body here is just doing logging (so a side effect).
`map_err` is more about converting an error isn't it?

fulmicoton (approved):

fulmicoton:
Glad to see unit tests on thjis part of the code!

quickwit-search/src/search_stream/leaf.rs
fmassot: readability and compiler issue I had but it's no more relevant.

quickwit-cli/tests/cli.rs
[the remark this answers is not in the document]
guilload: I like anyhow because it tends to do a better job at tracking the underlying error while we do a poor job at it. We could serialize the anyhow error as a string using the debug format, which displays the cause and a backtrace if one is present.

PR #490, opened 2021-08-30
Add PostgreSQL-based metastore.

### LINKED ISSUE #389: PostgreSQL based metastore

mosuka:
### Context / purpose
Add PostgreSQL-based metastore.
Close #389

### Description
Add PostgreSQL-based metastore.

### How was this PR tested?
```
$ cd quickwit-metastore

$ docker-compose up -d --force-recreate

$ cargo test --features=postgresql

$ docker-compose down
```

### Checklist
- [x] tested locally
- [x] added unit tests


mosuka:
@fulmicoton @fmassot 
I have a question on how to specify the max connection pool size.

Currently, the default value is specified, but I think it would be better to allow users to freely set the pool size.
However, in the current CLI, there is no way to pass the metastore-specific configuration values other than specifying the connection destination with `--metastore-uri`.

So, came up with the following methods:

Add an environment variable
```
POSTGRES_MAX_CONNECTION_POOL_SIZE=50
```

Extend the connection URI specified by `--metastore-uri`
```
postgres://quickwit-dev:quickwit-dev@localhost/quickwit-dev?max_connection_pool_size=50
```

Do you have any ideas to pass the metastore specific configurations?

quickwit-metastore/migrations/2021-08-29-122131_create_splits/up.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE splits (
+    split_id VARCHAR(50) PRIMARY KEY, 
+    split_state INTEGER NOT NULL,
```
fulmicoton, line 3: Let's use strings instead of INTEGER for the split state.

PostgreSQL also has enum that supposedly offer the best of two worlds, but they are a pain to modify.

quickwit-metastore/migrations/2021-08-29-122131_create_splits/up.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE splits (
+    split_id VARCHAR(50) PRIMARY KEY, 
+    split_state INTEGER NOT NULL, 
+    num_records BIGINT NOT NULL, 
+    size_in_bytes BIGINT NOT NULL, 
+    start_time_range BIGINT,
+    end_time_range BIGINT,
+    generation BIGINT NOT NULL,
+    update_timestamp BIGINT NOT NULL,
```
fulmicoton, line 9: Having a strict schema is can be great, but we will have more than one implementation, but as I commented on notion, we should rely on JSON here for most of this data.

We want to have as a field:

split_id,
index_id,
split_state,
start_time_range
end_time_range

Possibly tags as well, (which would then be an array).
The rest should be in a JSON field.

quickwit-metastore/src/error.rs
```diff
@@ -30,6 +30,9 @@ use crate::checkpoint::IncompatibleCheckpoint;
 #[allow(missing_docs)]
 #[derive(Debug, Error)]
 pub enum MetastoreError {
+    #[error("Connection error: `{message}`.")]
+    ConnectionError { message: String },
```
fulmicoton, line 34: good idea

quickwit-metastore/src/metastore.rs
```diff
@@ -130,6 +132,28 @@ impl Default for SplitState {
...
 
+impl FromPrimitive for SplitState {
+    fn from_i64(num_split_state: i64) -> Option<SplitState> {
+        match num_split_state {
+            0 => Some(SplitState::New),
+            1 => Some(SplitState::Staged),
+            2 => Some(SplitState::Published),
+            3 => Some(SplitState::ScheduledForDeletion),
+            _ => None,
+        }
```
fulmicoton, line 143: This is useless... 
- From<u64> should be sufficient
- if From<i64> was a good thing, it could still call from_u64.
- anyway let's use strings for enum.

quickwit-metastore/src/model.rs
```diff
@@ -0,0 +1,170 @@
...
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+use std::ops::{Range, RangeInclusive};
```
fulmicoton, line 23: we should put that in a postgresql module.  (as well as all of the files specific to postgres in this PR)

`quickwit-metastore/src/postgres/...`

Right now it is confusing...

fulmicoton (changes requested):

fulmicoton:
@mosuka For the moment, I think you can keep that a hardcoded constant. We have a lot of such constants and I think we need to think deeply about how to let the user configure them. (probably a config file)

quickwit-metastore/migrations/2021-08-29-122131_create_splits/up.sql
mosuka: I changed to use VARCHAR instead of INTEGER for split_state.

quickwit-metastore/migrations/2021-08-29-122131_create_splits/up.sql
mosuka: I left the following columns in the table and added a new `json` column.
The `json` column now contains a JSON string of `SplitMetadataAndFooterOffsets`.
This is so that the string in the `json` column can be easily deserialized to `SplitMetadataAndFooterOffsets`.

```
CREATE TABLE splits (
    split_id VARCHAR(50) PRIMARY KEY,
    split_state VARCHAR(30) NOT NULL,
    start_time_range BIGINT,
    end_time_range BIGINT,
    tags TEXT NOT NULL,
    json TEXT NOT NULL,
    index_id VARCHAR(50) NOT NULL,

    FOREIGN KEY(index_id) REFERENCES indexes(index_id)
);
```

quickwit-metastore/src/model.rs
mosuka: I did not fully understand the use of diesel.rs. I've fixed it now.

mosuka:
@fulmicoton 
I've reflected on your feedback. Please take another look if you have time.
Thanks.

quickwit-metastore/src/metastore.rs
```diff
@@ -20,15 +20,18 @@
...
 use std::collections::HashSet;
-use std::fmt::Debug;
+use std::fmt::{self, Debug};
 use std::ops::{Range, RangeInclusive};
+use std::str::FromStr;
 use std::{collections::HashMap, sync::Arc};
 
 use async_trait::async_trait;
 use chrono::Utc;
+// use num::FromPrimitive;
```
fulmicoton, line 34: remove

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1139 @@
...
+const MAX_CONNECTION_POOL_SIZE: u32 = 10;
+const CONNECTION_TIMEOUT: u64 = 10;
+const MAX_CONNECTION_RETRY_COUNT: u32 = 10;
+const CONNECTION_STATUS_CHECK_INTERVAL: u64 = 2;
+
+/// Establishes a connection to the given database URI.
+fn establish_connection(
+    database_uri: &str,
+) -> anyhow::Result<Pool<ConnectionManager<PgConnection>>> {
+    dotenv().ok();
```
fulmicoton, line 64: let's remove the use to dotenv.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1139 @@
...
+
+    let mut retry_cnt = 0;
+
+    loop {
```
fulmicoton, line 68: make it a for loop

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1139 @@
...
+
+                if retry_cnt > MAX_CONNECTION_RETRY_COUNT {
+                    anyhow::bail!(
+                        "The retry count has exceeded the limit ({})",
+                        MAX_CONNECTION_RETRY_COUNT
+                    );
+                }
+
+                retry_cnt += 1;
+                continue;
```
fulmicoton, line 90: continue is useless here.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1139 @@
...
+
+    embedded_migrations_res?;
+
+    Ok(())
+}
+
+/// Get the PostgreSQL-based metastore. The metastore is created as a singleton.
+pub async fn get_postgresql_metastore(database_uri: &str) -> &'static PostgresqlMetastore {
+    static POSTGRESQL_METASTORE: OnceCell<PostgresqlMetastore> = OnceCell::const_new();
+    POSTGRESQL_METASTORE
```
fulmicoton, line 128: This is incorrect.

It cannot be a single if it has an argument.

fulmicoton: HashMap<String, Weak<PostgresqlMetastore>> would be better.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1139 @@
...
+use crate::MetastoreResult;
+use crate::SplitMetadataAndFooterOffsets;
+use crate::SplitState;
+
+embed_migrations!("migrations/postgresql");
+
+const MAX_CONNECTION_POOL_SIZE: u32 = 10;
+const CONNECTION_TIMEOUT: u64 = 10;
+const MAX_CONNECTION_RETRY_COUNT: u32 = 10;
+const CONNECTION_STATUS_CHECK_INTERVAL: u64 = 2;
```
fulmicoton, line 58: You can have const duration.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1139 @@
...
+use crate::MetastoreFactory;
+use crate::MetastoreResolverError;
+use crate::MetastoreResult;
+use crate::SplitMetadataAndFooterOffsets;
+use crate::SplitState;
+
+embed_migrations!("migrations/postgresql");
+
+const MAX_CONNECTION_POOL_SIZE: u32 = 10;
+const CONNECTION_TIMEOUT: u64 = 10;
```
fulmicoton, line 56: you can have const Duration

quickwit-metastore/migrations/postgresql/2021-09-02-060554_create_splits/up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE splits (
+    split_id VARCHAR(50) PRIMARY KEY,
+    split_state VARCHAR(30) NOT NULL,
+    start_time_range BIGINT,
```
fulmicoton, line 4: I think it is ok not to have an index on start_time_ragne, end_time_range, and tags at the moment.

quickwit-metastore/migrations/postgresql/2021-09-02-060554_create_splits/up.sql
```diff
@@ -0,0 +1,11 @@
...
+    end_time_range BIGINT,
+    tags TEXT NOT NULL,
```
fulmicoton, line 6: tags could be an array field.
https://www.postgresql.org/docs/9.1/arrays.html

quickwit-metastore/migrations/postgresql/2021-09-02-060554_create_splits/up.sql
```diff
@@ -0,0 +1,11 @@
...
+    json TEXT NOT NULL,
```
fulmicoton, line 7: rename `split_metadata_json`  and comment that it contains the entire split metadata.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1139 @@
...
+
+        Ok(Arc::new((*metastore).clone()))
+    }
+}
+
+#[cfg(test)]
+#[async_trait]
+impl crate::tests::test_suite::DefaultForTest for PostgresqlMetastore {
+    async fn default_for_test() -> Self {
+        use std::env;
```
fulmicoton, line 1128: Can you put instruction somewhere about how to run tests locally?
Typically a docker compose file spinning a postgres and environment variable would be great.

dotenv with a committed .env file pointing at the docker-compose instance is ok, but please make the dotenv a dev-dependency.

quickwit-metastore/src/postgresql.rs
```diff
@@ -0,0 +1,4 @@
+#[allow(missing_docs)]
+pub mod model;
```
fulmicoton, line 2: postgresql/mod.rs please

quickwit-metastore/src/lib.rs
```diff
@@ -31,13 +31,22 @@ etc.
...
+#[macro_use]
+extern crate diesel;
+
 #[allow(missing_docs)]
 pub mod checkpoint;
 mod error;
 mod metastore;
 mod metastore_resolver;
+#[allow(missing_docs)]
+pub mod postgresql;
```
fulmicoton, line 46: #[feature(postresql)]

quickwit-metastore/src/metastore_resolver.rs
```diff
@@ -68,6 +69,7 @@ impl Default for MetastoreUriResolver {
             .register("file", SingleFileMetastoreFactory::default())
             .register("s3", SingleFileMetastoreFactory::default())
             .register("s3+localstack", SingleFileMetastoreFactory::default())
+            .register("postgresql", PostgresqlMetastoreFactory::default())
```
fulmicoton, line 72:
```
if cfg!(feature="postgresql") {
   // ...
}
```

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1120 @@
...
+            let conn_res = connection_pool_clone.get();
+            if conn_res.is_ok() {
+                break;
+            }
+            tokio::time::sleep(CONNECTION_STATUS_CHECK_INTERVAL).await;
+        }
+
+        match initialize_db(&*connection_pool) {
+            Ok(_) => info!("Database initialized"),
+            Err(err) => error!("Failed to initialize database {:?}", err),
```
fulmicoton, line 149: We should return an error here.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1120 @@
...
+
+/// PostgreSQL metastore implementation.
+#[derive(Clone)]
+pub struct PostgresqlMetastore {
+    uri: String,
+    connection_pool: Arc<Pool<ConnectionManager<PgConnection>>>,
+}
+
+impl PostgresqlMetastore {
+    /// Creates a meta store given a storage.
```
fulmicoton, line 124: the comment is wrong.

quickwit-metastore/src/metastore/postgresql_metastore.rs
mosuka: Sorry, I'll fix it.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1120 @@
...
+    pub async fn new(database_uri: &str) -> MetastoreResult<Self> {
+        let connection_pool = Arc::new(establish_connection(database_uri).map_err(|err| {
+            MetastoreError::ConnectionError {
+                message: err.to_string(),
+            }
+        })?);
+
+        let connection_pool_clone = connection_pool.clone();
```
fulmicoton, line 132: is cloning required here?

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1120 @@
...
+            debug!(
+                connections = connection_pool_status.connections,
+                idle_connections = connection_pool_status.idle_connections,
+                "Connection pool status"
+            );
+            let conn_res = connection_pool_clone.get();
+            if conn_res.is_ok() {
+                break;
+            }
+            tokio::time::sleep(CONNECTION_STATUS_CHECK_INTERVAL).await;
```
fulmicoton, line 144: Can we return an error if the connection could not be established after 3 attempts?

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1120 @@
...
+        let mut index_metadata =
+            model_index
+                .make_index_metadata()
+                .map_err(|err| MetastoreError::InternalError {
+                    message: "Failed to deserialize index metadata".to_string(),
+                    cause: anyhow::anyhow!(err),
+                })?;
+
+        // Apply checkpoint_delta
+        index_metadata
```
fulmicoton, line 557: I'd rather have the checkpoint check logic to be part of the transaction...

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1120 @@
...
+                })?;
+        if !index_exists {
+            return Err(MetastoreError::IndexDoesNotExist {
+                index_id: index_id.to_string(),
+            });
+        }
+
+        // Check for the inclusion of non-publishable split IDs.
+        // Except for SplitState::Staged and SplitState::Published, you cannot publish.
+        let non_publishable_split_ids = self
```
fulmicoton, line 655: That logic is not bad but is redundant with the logic existing within the transaction. I think we can remove it and make our error more specific in the transaction code.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1120 @@
...
+                for split_id in replaced_split_ids.iter() {
+                    if !mark_as_deleted_split_ids
+                        .iter()
+                        .any(|mark_as_deleted_split_id| mark_as_deleted_split_id == *split_id)
+                    {
+                        error_split_id = split_id;
+                        break;
+                    }
+                }
+                return Err(diesel::result::Error::NotFound);
```
fulmicoton, line 703: We are finding error_split_id and then not using it and returning an error.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1120 @@
...
+            }
+
+            // Mark as deleted.
+            let mark_as_deleted_split_ids =
+                self.mark_as_deleted(&conn, index_id, replaced_split_ids)?;
+
+            // Find out the ID of the update error and return an error if there is one.
+            if mark_as_deleted_split_ids.len() < replaced_split_ids.len() {
+                for split_id in replaced_split_ids.iter() {
```
fulmicoton, line 694: Editing split_error and returning an error, that will be converted afterwards is a very ugly pattern. If you have ideas for refactoring, it might be better.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1120 @@
...
+                    Ok(metadata) => metadata,
+                    Err(err) => {
+                        error!("Failed to make split metadata and footer offsets {:?}", err);
+                        continue;
+                    }
+                };
+
+            // The tags are empty (unspecified) or the tags are specified
+            // and the split has a matching tag, it will be added to the list.
+            let mut match_tag = false;
```
fulmicoton, line 796: Can you that logic in an external function?

/// ...
fn does_match_tag_filter(tags: &[String], tag_filter: &[String]) -> bool {
}


it could be factorized in the single file metastore too.

quickwit-metastore/src/metastore/postgresql_metastore.rs
mosuka: This cloning was not necessary...

quickwit-metastore/src/metastore/postgresql_metastore.rs
mosuka: Sorry, I'll fix it.

fulmicoton:
The following tests are failing for me.
```

---- metastore::postgresql_metastore::common_tests::test_metastore_delete_splits stdout ----
ng
thread 'metastore::postgresql_metastore::common_tests::test_metastore_delete_splits' panicked at 'assertion failed: matches!(result, MetastoreError :: IndexDoesNotExist { .. })', quickwit-metastore/src/tests.rs:897:13

---- metastore::postgresql_metastore::common_tests::test_metastore_split_update_timestamp stdout ----
ng
thread 'metastore::postgresql_metastore::common_tests::test_metastore_split_update_timestamp' panicked at 'called `Result::unwrap()` on an `Err` value: InternalError { message: "Failed to create index", cause: relation "indexes" does not exist }', quickwit-metastore/src/tests.rs:1685:14

---- metastore::postgresql_metastore::common_tests::test_metastore_replace_splits stdout ----
ng
thread 'metastore::postgresql_metastore::common_tests::test_metastore_replace_splits' panicked at 'assertion failed: matches!(result, MetastoreError :: IndexDoesNotExist { .. })', quickwit-metastore/src/tests.rs:631:13

---- metastore::postgresql_metastore::common_tests::test_metastore_stage_split stdout ----
ng
thread 'metastore::postgresql_metastore::common_tests::test_metastore_stage_split' panicked at 'assertion failed: matches!(result, MetastoreError :: IndexDoesNotExist { .. })', quickwit-metastore/src/tests.rs:188:9


failures:
    metastore::postgresql_metastore::common_tests::test_metastore_delete_splits
    metastore::postgresql_metastore::common_tests::test_metastore_replace_splits
    metastore::postgresql_metastore::common_tests::test_metastore_split_update_timestamp
    metastore::postgresql_metastore::common_tests::test_metastore_stage_split

```

mosuka:
@fulmicoton Let me check...

fulmicoton:
The test are passing correctly when run single threaded.
Please complete the test instructions, or tweak the DefaultForTest impl of PostgresMetastore to prevent concurrent execution of these tests.

mosuka:
@fulmicoton I tweaked the DefaultForTest implementation of PostgresMetastore to prevent concurrent execution of these tests.
https://github.com/quickwit-inc/quickwit/blob/8b460dfd684912e3668f8192e5a9d96edd4fbcaa/quickwit-metastore/src/metastore/postgresql_metastore.rs#L1104-L1134

quickwit-metastore/src/metastore/postgresql_metastore.rs
mosuka: https://github.com/quickwit-inc/quickwit/blob/8b460dfd684912e3668f8192e5a9d96edd4fbcaa/quickwit-metastore/src/metastore/postgresql_metastore.rs#L811-L819

https://github.com/quickwit-inc/quickwit/blob/8b460dfd684912e3668f8192e5a9d96edd4fbcaa/quickwit-metastore/src/metastore/single_file_metastore.rs#L386-L393

quickwit-metastore/src/metastore/postgresql_metastore.rs
mosuka: Fixed.

quickwit-metastore/src/metastore/postgresql_metastore.rs
mosuka: Fixed.

quickwit-metastore/src/metastore/postgresql_metastore.rs
mosuka: Fixed.

quickwit-metastore/src/metastore/postgresql_metastore.rs
mosuka: Fixed.

quickwit-metastore/src/metastore/postgresql_metastore.rs
mosuka: If I move the checkpoint logic to a transaction, I would expect to roll back the transaction if an error occurs in `Checkpoint::try_apply_delta`.
In `transaction` functions, it seems that if I want to roll back with user code other than DB operations, I have to return `diesel::result::Error:.RollbackTransaction`.
Then the `MetastoreError::IncompatibleCheckpointDelta` that the `publish_splits` function has to return cannot be known outside the transaction.
Is there a good way to know non-diesel::result::Error that occurs in a transaction outside of the transaction?

mosuka:
@fulmicoton 
I corrected the points that were pointed out. Please take another look.

fulmicoton (changes requested):
I requested a couple of changes.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1066 @@
...
+            let split_metadata_and_footer_offsets_json =
+                match serde_json::to_string(&split_metadata_and_footer_offsets) {
+                    Ok(json_str) => json_str,
+                    Err(err) => {
+                        error!(
+                            "Failed to serialize from JSON to SplitMetadataAndFooterOffsets \
+                             split_id={:?} {:?}",
+                            model_split.split_id, err
+                        );
+                        continue;
```
fulmicoton, line 247: You did not address my comments.. How is this ok to just log ?

mosuka:
@fulmicoton 
Thanks for the review.
The part about not returning an error is my mistake. I have fixed it.
I also fixed the other comments you made, so please check them.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1095 @@
...
+
+            if published_split_ids.len() < split_ids.len() {
+                // Return an error if there are any splits that could not be published.
+                check_all_splits_were_modified(
+                    split_ids,
+                    &published_split_ids
+                        .iter()
+                        .map(String::as_str)
+                        .collect::<Vec<_>>(),
+                )?
```
fulmicoton, line 723:
```suggestion
                )?;
```

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1095 @@
...
+
+            if published_split_ids.len() < new_split_ids.len() {
+                // Return an error if there are any splits that could not be published.
+                check_all_splits_were_modified(
+                    new_split_ids,
+                    &published_split_ids
+                        .iter()
+                        .map(String::as_str)
+                        .collect::<Vec<_>>(),
+                )?
```
fulmicoton, line 765:
```suggestion
                )?;
```

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -0,0 +1,1095 @@
...
+
+            if mark_as_deleted_split_ids.len() < replaced_split_ids.len() {
+                // Return an error if there are any splits that could not be marked as deleted.
+                check_all_splits_were_modified(
+                    replaced_split_ids,
+                    &mark_as_deleted_split_ids
+                        .iter()
+                        .map(String::as_str)
+                        .collect::<Vec<_>>(),
+                )?
```
fulmicoton, line 780:
```suggestion
                )?;
```

fulmicoton (approved):

Issue #480, opened 2021-08-29
Bad error message on `quickwit new` if the config file path is wrong


fulmicoton:
The user only gets the following error.
```
Failed to parse command arguments: No such file or directory (os error 2)
```

Ideally we should show the command it is complaining about and the filepath that was input.
It is especially confusing as the index-uri could be referring to a filepath too.

PR #491, opened 2021-08-30
Use the couple (metastore-uri, index-id) as CLI main arguments.

### LINKED ISSUE #459: metastore uri mess
We recently introduced the notion of a metastore uri...
With it, we broke the notion of a required index_uri, from which we were deducing
the index id and the metastore_uri.

Here is the problem.
The index uri is coming from (metastore_uri, index_id)....
So what is the point of the index_uri as a CLI argument then?

Let's define a spec and stick with it.

I think 
- index_uri 
or
- metastore_uri + index_id makes sense.

We could accepts both sets of cli parameters.

Alternatively, we could keep the concept of index_uri, and have the index_uri
mean metastore_uri + index_id.

This is close to what was happening in quickwit0.1.
Of course, it feels quirky for any store different than object storage.

fmassot:
Fix #459 and #480

### DIFF
--- a/quickwit-cli/src/lib.rs
+++ b/quickwit-cli/src/lib.rs
@@ -337,7 +337,7 @@ pub async fn garbage_collect_index_cli(args: GarbageCollectIndexArgs) -> anyhow:
         "{}MB of storage garbage collected.",
         deleted_bytes / 1_000_000
     );
-    println!("Index `{}` successfully garbage collected", args.index_id);
+    println!("Index `{}` successfully garbage collected.", args.index_id);
     Ok(())
 }
 
--- a/quickwit-cli/src/lib.rs
+++ b/quickwit-cli/src/lib.rs
@@ -69,9 +69,9 @@ const THROUGHPUT_WINDOW_SIZE: usize = 5;
 
 #[derive(Debug)]
 pub struct CreateIndexArgs {
+    metastore_uri: String,
     index_uri: String,
     index_config: Arc<dyn IndexConfig>,
-    metastore_uri: String,
     overwrite: bool,
 }
 impl PartialEq for CreateIndexArgs {
@@ -84,9 +84,9 @@ impl PartialEq for CreateIndexArgs {
 
 impl CreateIndexArgs {
     pub fn new(
+        metastore_uri: String,
         index_uri: String,
         index_config_path: PathBuf,
-        metastore_uri: String,
         overwrite: bool,
     ) -> anyhow::Result<Self> {
         let json_file = std::fs::File::open(index_config_path)?;
@@ -96,9 +96,9 @@ impl CreateIndexArgs {
         let index_config = Arc::new(builder.build()?) as Arc<dyn IndexConfig>;
 
         Ok(Self {
+            metastore_uri,
             index_uri,
             index_config,
-            metastore_uri,
             overwrite,
         })
     }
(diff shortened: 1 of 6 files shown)


quickwit-cli/src/lib.rs
```diff
@@ -84,61 +84,81 @@ impl PartialEq for CreateIndexArgs {
...
 impl CreateIndexArgs {
     pub fn new(
+        metastore_uri: String,
         index_uri: String,
         index_config_path: PathBuf,
-        metastore_uri: String,
         overwrite: bool,
     ) -> anyhow::Result<Self> {
-        let json_file = std::fs::File::open(index_config_path)?;
+        let json_file = std::fs::File::open(index_config_path.clone()).map_err(|error| {
```
guilload, line 92: `map_err(|err| anyhow::anyhow!(...))?` can be replaced with the `anyhow` helper `with_context(...)?`

quickwit-cli/src/lib.rs
```diff
@@ -221,7 +233,7 @@ pub async fn index_data_cli(args: IndexDataArgs) -> anyhow::Result<()> {
         start_statistics_reporting_loop(supervisor_handler, args.input_path.clone()).await?;
 
     if statistics.num_published_splits > 0 {
-        println!("You can now query your index with `quickwit search --index-uri {} --query \"barack obama\"`" , args.index_uri);
+        println!("You can now query your index with `quickwit search --index-id {} --query \"barack obama\"`" , args.index_id);
```
guilload, line 236: Is that gonna work? There's no metastore URI in the command.

quickwit-cli/src/lib.rs
```diff
@@ -331,10 +337,7 @@ pub async fn garbage_collect_index_cli(args: GarbageCollectIndexArgs) -> anyhow:
         "{}MB of storage garbage collected.",
         deleted_bytes / 1_000_000
     );
-    println!(
-        "Index successfully garbage collected at `{}`",
-        args.index_uri
-    );
+    println!("Index `{}` successfully garbage collected", args.index_id);
```
guilload, line 340:
```suggestion
    println!("Index `{}` successfully garbage collected.", args.index_id);
```

guilload (commented):
I'm also in favor of [capitalizing acronyms](https://www.grammarly.com/blog/abbreviations/):
- id -> ID
- uri -> URI

guilload:
With this PR, do we lose the ability to `serve` multiple indexes?

fmassot:
> With this PR, do we lose the ability to `serve` multiple indexes?

Definitely not. For a given index, its `index-uri` is stored in the metastore, index metadata is fetched given the `index-id`. If no index meta is found on the metastore, we return a 404.

This makes me think about a missing method on the metastore, we may want to list all the indexes registered in the metastore, the current metastore API does not provide such a method.

quickwit-cli/src/lib.rs
fmassot: nice catch, we need the metastore uri.

fmassot:
> I'm also in favor of [capitalizing acronyms](https://www.grammarly.com/blog/abbreviations/):
> 
> * id -> ID
> * uri -> URI

in documentation you mean?

guilload:
yep

evanxg852000 (approved):

PR #506, opened 2021-09-02
Fix #505 and add a test on a running server.

### LINKED ISSUE #505: Fix search stream grpc client
and add a test to confirm that @guilload is right :)

Currently, we are dropping errors coming from leaf responses because of this line: https://github.com/quickwit-inc/quickwit/pull/506/files#diff-f1b0efaf0dd1c97155a4ed98cb294688935792c9ebd97d9352157bf8f4d38c0dL143

fmassot:
Fix #505


guilload (approved):

quickwit-serve/src/lib.rs
```diff
@@ -150,3 +150,94 @@ pub async fn serve_cli(args: ServeArgs) -> anyhow::Result<()> {
...
+        result_sender.send(Err(SearchError::InternalError("error".to_string())))?;
+        mock_search_service.expect_leaf_search_stream().return_once(
+            |_leaf_search_req: quickwit_proto::LeafSearchStreamRequest| {
+                Ok(UnboundedReceiverStream::new(result_receiver))
+            },
+        );
+        // The test will hang on indefinitely if we don't drop the sender.
+        drop(result_sender);
+
+        let grpc_addr: SocketAddr = format!("127.0.0.1:{}", 10000).parse()?;
```
PSeitz, line 231: this should not be a static number to avoid collisions with other tests

PR #510, opened 2021-09-03
Storage cache

evanxg852000:
### Context / purpose
Ongoing implementation for multi-level cache storage 
should close:
-  https://github.com/quickwit-inc/quickwit/issues/400
-  https://github.com/quickwit-inc/quickwit/issues/408

### Description
Implemented the first-level cache from remote storage to local storage.

### How was this PR tested?
N/A

### Checklist
- [x] first level cache implementation 
- [x] Create an in-ram cache that can cache file slice (Path, Range).
- [x] Integrate in the Quicwit indexing, and search
- [x] tested locally
- [x] added unit tests (in progress)


evanxg852000:
@fulmicoton @guilload 
The idea is to have a peek and given potential pitfalls. 
The next step is to create an in-ram cache more like `quickwit/quickwit-storage/src/cache/in_ram_slice_cache.rs` and plug it in 
`LocalStorageCache`  to provide a better implementation of `get_slice`

quickwit-storage/src/storage.rs
```diff
@@ -121,4 +121,7 @@ pub trait Storage: Send + Sync + 'static {
 
     /// Returns an URI identifying the storage
     fn uri(&self) -> String;
+
+    /// Return the root path of this storage if appropriate.
```
fulmicoton, line 125: I really wonder why you would need that. the comment is not helping... is it to guess where the cache should be created?

quickwit-storage/src/cache_bis/storage_with_local_cache.rs
```diff
@@ -0,0 +1,632 @@
...
+    },
+};
+
+use super::{local_storage_cache::LocalStorageCache, Cache, CacheState, FULL_SLICE};
+
+/// A storage with a backing [`Cache`].
+pub struct StorageWithLocalStorageCache {
+    remote_storage: Arc<dyn Storage>,
+    cache: Mutex<Box<dyn Cache>>,
+    in_flight_download_requests: Mutex<HashSet<PathBuf>>,
```
fulmicoton, line 47: great name.

quickwit-storage/src/storage.rs
evanxg852000: Yes, mostly for SingleFileStorage, I will elaborate.

fulmicoton:
I think this is too complicated. 

Let's solve the 2 ticket independently and start by the easy one: the upload cache.
For that one, we only cache entire files.

You won't be able to use the `StorageWithCache` trait unfortunately so that means creating your own wrapper.
We can call it, `StorageWithUploadCache`.

It needs to work for the following workload.
- The cache entirely ignores slices. The wrapper directly calls the underlying storage. It also ignore get_all.
- `copy_to_file` is on the hand trying to hit the cache. 
When it is called, if the file is not in cache, just bypass the cache entirely. In other words, do not try to store the file into the cache.
- The cache is populated upon `put` calls and when we restart the process. For a same path, any reasonable behavior is acceptable for the put. We don't want a bug, but you can assume our workload does not put the same file twice, so for instance no need to try to optimize and debounce those calls etc.

That way, you can entirely dismiss the problem of "what do I do if two call to get are performed?", how do I debounce them?

evanxg852000:
No longer needed

PR #516, opened 2021-09-05
Add and apply rustfmt config

guilload:
I was looking into rustfmt options and found some interesting settings:
- enforce license header;
- organize comments and limit comment width;
- organize imports in groups (std, external, crate);
- remove trailing commas (or always add them if that's more to your liking).

This makes for a more consistent and eye-pleasing codebase. The major con is that some options are only available on the nightly channel. So, I added a `make fmt` command.

Let me know what you think and if we decide to go ahead with this, I can set this up in CI and update `make fmt` to display a nicer error message when nightly is not available with instructions on how to install it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
             target/
           key: ${{ runner.os }}-cargolint-${{ hashFiles('Cargo.lock') }}
       - name: rustfmt
-        run: cargo fmt --all -- --check
+        run: cargo +nightly fmt --all -- --check
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.license_header.txt
+++ b/.license_header.txt
@@ -1,0 +1,19 @@
+// Copyright (C) {\d+} Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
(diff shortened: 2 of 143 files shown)


.rustfmt.toml
```diff
@@ -0,0 +1,16 @@
...
+  "quickwit-swim",
+]
+
+comment_width = 120
+format_strings = true
+group_imports = "StdExternalCrate"
+imports_granularity = "Module"
+license_template_path = ".license_header.txt"
+normalize_comments = true
+trailing_comma = "Never"
```
PSeitz, line 13: I prefer the default behaivour, it's easier to extend

.rustfmt.toml
```diff
@@ -0,0 +1,16 @@
+ignore = [
+  "quickwit-proto/src/cluster.rs",
+  "quickwit-proto/src/quickwit.rs",
+  "quickwit-swim",
+]
+
+comment_width = 120
+format_strings = true
+group_imports = "StdExternalCrate"
+imports_granularity = "Module"
```
PSeitz, line 10: I think there is a real downside on increased merge conflicts and would not automatically group imports for that reason

.rustfmt.toml
guilload: Agreed. Do folks prefer "always" or "vertical"?
https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=#trailing_comma

.rustfmt.toml
guilload: Those conflicts are super easy to resolve though: 
1. remove the git markers;
2. run `fmt` again to get rid of duplicates;
3. in some cases, remove the unused imports.

I personally find that this overhead is preferable over hard-to-read nested imports and contributors arguing over how imports should be organized:
```rust
use foo::{self, bar::{self, qux::{self, ...}}
};
```
In my opinion, the more code formatting can be automated, the better.

.rustfmt.toml
guilload: (the default behavior being "vertical")

PSeitz: I assume it's vertical, but I don't really care about the difference between the two

.rustfmt.toml
PSeitz: It's still an extra step where you need to check if it's a semantic merge conflict or only imports.

I agree on more formatting is better, since it usually increases readability and reduces merge conflicts.
But in my experience this is not true for imports since readability on imports is not really important and merge conflicts can happen very frequent with that setting.

In practice I think our imports are just a flat list and not such a nested structure, since they are hard to type and maintain.

.rustfmt.toml
fmassot: vertical is better :)

.rustfmt.toml
fmassot: Indeed, merge conflicts coming from formatting are annoying. Currently, I did not really have to handle formatting merge conflicts or it was really minor. I don't think formatting the imports will change much so I think we can test it and rethink about it as the project grows.

fmassot:
sounds all good.

PR #519, opened 2021-09-08
Moving test utils to quickwit-indexing.

fulmicoton:
This PR moves test_utils from quickwit-core to quickwit-utils.

The idea is that this code is useful for testing quickwit_indexing, but depending on quickwit-indexing would create a loop dependency.


evanxg852000 (approved):
LGTM

PR #527, opened 2021-09-10
Rename cluster RPCs *Result -> *Response

guilload:
### Description
Rename gRPC structs such as `LeaveResult` to `LeaveResponse`. More renaming PRs will come. 

### How was this PR tested?
No new unit tests were added.


PSeitz (approved):
LGTM (why is license check so slow)

guilload:
> (why is license check so slow)

I think it's a GitHub bug. I removed the check because it's part of `cargo fmt` now but for some reason, it still appears in the checks.

Issue #425, opened 2021-08-19
Retry for stream API


fulmicoton:
Right now we have a retry in the search API, but we want a retry in the stream API as well.

fulmicoton:
Assigned to @fmassot temporarily.

fmassot:
duplicate of #369

Issue #380, opened 2021-08-11
Add retry for fetch doc.


fulmicoton:

fmassot:
closed with #530

PR #530, opened 2021-09-10
Refactor retry for search, docs and search stream

fmassot:
Aims at fixing #380 and #425

A few comments:
- the retry policies are very simple right now, one retry for all errors but this is easy to customize.
- on retry, we select a retry client, the choice is also very simple, select one client while excluding the failing one. For now, I would prefer to keep this part simple and improve the client pool to make it aware of failing clients in another PR.


quickwit-search/src/client.rs
```diff
@@ -149,47 +149,62 @@ impl SearchServiceClient {
...
+                } else {
+                    tokio::spawn(
+                        async move {
+                            let mut results_stream = tonic_result
+                                .unwrap()
+                                .into_inner()
+                                .map_err(|tonic_error| parse_grpc_error(&tonic_error));
+                            while let Some(search_result) = results_stream.next().await {
+                                result_sender.send(search_result).map_err(|_| {
+                                    SearchError::InternalError(
```
fulmicoton, line 187: What is the point of this error conversion?
The error is actually silently dropped.

quickwit-search/src/client.rs
```diff
@@ -149,47 +149,62 @@ impl SearchServiceClient {
...
+                if tonic_result.is_err() {
+                    // Receiver cannot be closed here, ignore error.
+                    let _ = result_sender.send(Err(tonic_result.unwrap_err()));
+                } else {
+                    tokio::spawn(
+                        async move {
+                            let mut results_stream = tonic_result
+                                .unwrap()
+                                .into_inner()
+                                .map_err(|tonic_error| parse_grpc_error(&tonic_error));
```
fulmicoton, line 184: Maybe I misunderstand the flow, but I suspect leaving the map_err within the loop would be actually easier to understand (provided it compiles of course).

It is always a tad difficult for the reader to know if that kind of map_err applies to a Result, or to the items within the stream.

quickwit-search/src/cluster_client.rs
```diff
@@ -0,0 +1,541 @@
...
+        }
+        (Ok(response), Err(_)) => Ok(response),
+        (Err(_), Ok(retry_response)) => Ok(retry_response),
+        (Err(error), Err(_)) => Err(error),
+    }
+}
+
+async fn forward_leaf_search_stream(
+    mut stream: UnboundedReceiverStream<Result<LeafSearchStreamResult, SearchError>>,
+    sender: UnboundedSender<Result<LeafSearchStreamResult, SearchError>>,
```
fulmicoton, line 158: Should we make both the stream and the sender Bounded instead?
(If it is sensible, we can put that in a separate ticket.)

The point would be to add some nice back pressure.

quickwit-search/src/cluster_client.rs
```diff
@@ -0,0 +1,541 @@
...
+    stop_on_error: bool,
+) -> Result<SuccessFullSplitIds, SendError<Result<LeafSearchStreamResult, SearchError>>> {
+    let mut sucessful_split_ids: Vec<String> = Vec::new();
+    while let Some(result) = stream.next().await {
+        match result {
+            Ok(response) => {
+                sucessful_split_ids.push(response.split_id.clone());
```
fulmicoton, line 165: successful_split_ids

quickwit-search/src/cluster_client.rs
```diff
@@ -0,0 +1,541 @@
...
+            };
+            Ok(merged_response)
+        }
+        (Ok(response), Err(_)) => Ok(response),
+        (Err(_), Ok(retry_response)) => Ok(retry_response),
+        (Err(error), Err(_)) => Err(error),
+    }
+}
+
+async fn forward_leaf_search_stream(
```
fulmicoton, line 156: We need doc on that function, especially explain the spec for stop_on_error...

quickwit-search/src/cluster_client.rs
```diff
@@ -0,0 +1,541 @@
...
+        placed_request: (LeafSearchStreamRequest, SearchServiceClient),
+    ) -> UnboundedReceiverStream<Result<LeafSearchStreamResult, SearchError>> {
+        let (request, mut client) = placed_request;
+        let (result_sender, result_receiver) = unbounded_channel();
+        let client_pool = self.client_pool.clone();
+        let retry_policy = LeafSearchStreamRetryPolicy {};
+        tokio::spawn(async move {
+            let results_stream = client.leaf_search_stream(request.clone()).await;
+            let forward_result =
+                forward_leaf_search_stream(results_stream, result_sender.clone(), true).await;
```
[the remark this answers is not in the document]
fulmicoton: Why do we stop on error for the first round?

Let's say a split is missing from the storage. Isn't it something we want to not stop on, and should keep getting the working splits?

quickwit-search/src/retry/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+
+use crate::client_pool::Job;
+use crate::{ClientPool, SearchClientPool, SearchServiceClient};
+
+/// A retry policy to evaluate if a request should be retried.
+pub trait RetryPolicy<Request, Response, Error>: Sized
+where Request: RequestOnSplit + Clone
+{
+    /// Checks the policy if a certain request should be retried.
+    fn should_retry(&self, req: &Request, result: Result<&Response, &Error>) -> bool;
```
fulmicoton, line 37: Alternatively we could have had a 

```
fn retry_request(&self, req: &Request, result: Result<&Response, &Error>) -> Option<Request>;
```

quickwit-search/src/retry/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+use quickwit_proto::{FetchDocsRequest, LeafSearchRequest, LeafSearchStreamRequest};
+
+use crate::client_pool::Job;
+use crate::{ClientPool, SearchClientPool, SearchServiceClient};
+
+/// A retry policy to evaluate if a request should be retried.
+pub trait RetryPolicy<Request, Response, Error>: Sized
+where Request: RequestOnSplit + Clone
+{
+    /// Checks the policy if a certain request should be retried.
```
fulmicoton, line 36: I was surprised by the fact that the prototype suggests that we might retry and `Ok(&Response)`.

I assume this is because potentially, the `Ok(&Response)` might contain a partial set of result?

quickwit-search/src/retry/search_stream.rs
```diff
@@ -0,0 +1,175 @@
...
+
+use super::RetryPolicy;
+use crate::SearchError;
+
+/// Retry policy for consuming the result stream of a LeafSearchStreamRequest.
+/// A retry is only made if there are some missing splits.
+/// As errors only come from a closed receiver, we ignore them.
+pub struct LeafSearchStreamRetryPolicy {}
+
+pub type SuccessFullSplitIds = Vec<String>;
```
fulmicoton, line 31: Nitpick: No need to fix, but typedef in rust are quite a bummer. They are like C++ typedef. It is just a type alias.

If you want static typing over it, you can do

```pub struct SuccessfullSplitIds(pub Vec<String>);```

Comment (fulmicoton) [reply quickwit-search/src/retry/search_stream.rs:31]: SuccessfulSplitIds

File: quickwit-search/src/retry/fetch_docs.rs [hunk 10]
```diff
@@ -0,0 +1,72 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+
+
+#[cfg(test)]
+mod tests {
+    use quickwit_proto::{FetchDocsRequest, FetchDocsResult, SplitIdAndFooterOffsets};
+    use super::FetchDocsRetryPolicy;
```
fulmicoton, line 25: I suspect `FetchDocsRetryPolicy` is defined nowhere and this file is not imported from retry/mod.rs

quickwit-search/src/retry/mod.rs
fmassot: exactly, we want to build a new request on failing splits

quickwit-search/src/retry/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+use std::sync::Arc;
+
+use itertools::Itertools;
+use quickwit_proto::{FetchDocsRequest, LeafSearchRequest, LeafSearchStreamRequest};
+
+use crate::client_pool::Job;
+use crate::{ClientPool, SearchClientPool, SearchServiceClient};
+
+/// A retry policy to evaluate if a request should be retried.
+pub trait RetryPolicy<Request, Response, Error>: Sized
```
fulmicoton, line 33: It was not clear when I read the code, but what happens on how to handle partial success with non-retryable errors or failed retried errors?

(Typically one split is corrupted)

Do we return nothing?
Do we return Ok(...) with whatever data we had and some information in the response to tell what split is missing?
Do we returen Ok(...) silently? 

Maybe the RetryPolicy trait could be a decent place to describe that spec.

fulmicoton (changes requested):
That's cool, especially the unit tests.

I had a couple of change request, quite a few questions, (and a bunch of nitpicks you should probably ignore)

quickwit-search/src/retry/mod.rs
fmassot: Good point. I will describe it more accurately.
I described the behaviour for the `LeafSearchRequest` [here](https://github.com/quickwit-inc/quickwit/pull/530/files#diff-721044fceec236ac1242484cf70dddddf96f672c0c7070421d2d48423799a69aR25-R27) but it's rather simple right now.

Basically we want to make a retry either on error or retryable split errors. 
On improvement that can make the code better is just to keep one method `retry_request` that return None if there is no retry and the retry request if we have retryable errors.

quickwit-search/src/retry/fetch_docs.rs
fmassot: yes, old file that needs to be deleted.

quickwit-search/src/client.rs
fmassot: yes this is useless

quickwit-search/src/cluster_client.rs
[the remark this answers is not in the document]
fmassot: actually that was the current behaviour, stop doing some work as soon as possible. On your kind of errors, this totally make sense to continue receiving results.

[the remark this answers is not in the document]
fmassot: I thought about passing the sender to `leaf_search_stream` but totally forget that, this can really simplify the code.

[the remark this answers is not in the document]
fmassot: For the extra mpsc piping, we need to filter out errors in order to return only leaf results if possible.
So the need is to create an intermediary channel to send only ok(response) on the first query and on retry send both ok(response) and errors.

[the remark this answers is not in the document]
fulmicoton: Can we add that in a comment? (I'm not sure what the right place could be though)

quickwit-search/src/retry/mod.rs
fulmicoton: I saw that comment, but I don't know what the end user sees if one split fails even after retry?

quickwit-search/src/retry/mod.rs
fmassot: super cool idea, this will make the trait clearer.

quickwit-search/src/retry/mod.rs
fmassot: Currently we have no HTTP streaming and as soon as we have an error, we stop and return an HTTP error.

With HTTP streaming, we will stream the result and on the first error do something so that the client will understand that something went wrong, I still need to look at that to see what the proper way to stop the stream response.

fmassot: In the `root_search` or `root_search_stream` functions, you have something like this:

```
    let result: Vec<Result> =
        futures::stream::iter(jobs.into_iter())
            .map(|(client, client_jobs)| {
                let leaf_request = jobs_to_leaf_request(client_jobs)
                cluster_client.leaf_search((leaf_request, client))
            })
            .buffer_unordered(MAX_CONCURRENT_LEAF_TASKS)
            .try_collect()
            .await?;
```

fmassot: The rationale is: once some errors reach the root level even after retry, just return propagate the error to the end user.

quickwit-search/src/cluster_client.rs
fmassot: I would prefer to handle that later when we enable HTTP streaming.

quickwit-search/src/client.rs
```diff
@@ -149,47 +149,59 @@ impl SearchServiceClient {
...
                             .map_err(|tonic_error| parse_grpc_error(&tonic_error));
-
                         while let Some(search_result) = results_stream.next().await {
-                            result_sender.send(search_result).map_err(|_| {
-                                SearchError::InternalError(
-                                    "Sender closed, could not send leaf result.".into(),
-                                )
-                            })?;
+                            let send_result = result_sender.send(search_result);
+                            // If we get a sending error, stop consuming the stream.
```
fulmicoton, line 186: Very useful comment

quickwit-search/src/cluster_client.rs
fulmicoton: great comment

quickwit-search/src/cluster_client.rs
fulmicoton: ah yes

fulmicoton (approved):
LGTM good job. The comments are really helpful and the unit test will help all of us sleep at night :D

quickwit-search/src/client_pool.rs
```diff
@@ -32,7 +31,7 @@ use crate::SearchServiceClient;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Job {
     /// SplitMetadataAndFooterOffsets
```
PSeitz, line 33: outdated comment

quickwit-search/src/retry/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+        }
+    }
+}
+
+// Select a new client from the client pool by the following oversimplified policy:
+// 1. Take the first split_id of the request
+// 2. Ask for a relevant client for that split while excluding the failing client.
+pub async fn retry_client<Request>(
+    client_pool: &Arc<SearchClientPool>,
+    failing_client: &SearchServiceClient,
```
PSeitz, line 62: shouldn't this be a list of failing clients to exclude all failing clients as alternative (except all are failing)

fmassot: it would be better yes, it's clearly oversimplified. The main advantage of this PR is: we do a very simple retry for every type of request while keeping the retry code limited. But it's a babystep clearly.

The plan is to improve that point in another PR.

But I need to think more deeply about this. I want also to think about how we want to execute requests. For example, we may want to retry a request if we think that a node is taking too much time. We would need some sort of tasks executor... I need to read some code from bigger projects to know what we should do there.

PR #532, opened 2021-09-10
Refactor cluster tests

guilload:
### Description
- added a `wait_for_members` method on `Cluster` that waits for the cluster to be in a state that satisfies a given predicate
- used that method to remove the ugly calls to `sleep` in tests
- added some checks (node is dropped vs. node leaves cluster)
- cleaned up code and tests
- renamed a few functions
- moved some functions to `common::net`

### How was this PR tested?
Ran `cargo test` locally.


quickwit-cluster/src/cluster.rs
```diff
@@ -205,16 +207,36 @@ impl Cluster {
...
+    /// members.
+    pub async fn wait_for_members<F>(
+        self: &Cluster,
+        mut predicate: F,
+        timeout_after: Duration,
+    ) -> anyhow::Result<()>
+    where
+        F: FnMut(&Vec<Member>) -> bool,
+    {
+        timeout(
```
fulmicoton, line 231: that's nice!

fulmicoton (approved):

PR #533, opened 2021-09-10
Prepend split ID with datetime

guilload:
### Description
Prepend split ID with UTC datetime. This is really nice when listing splits on a file system.

### How was this PR tested?
Added unit test and printed new split ID: `20210910T225401-01FF8YNWXWZE7N2P2VAX4MC62N`.


fulmicoton:
I agree that this nice to be human readable but that makes very long split ids.
The current split id are using ULID, so they already sort with date.

Also, while it is not human readable, someone who really wants the timestamp can extract them.
Note that both in your PR and in the ULID, it is a bit difficult to understand what this date is about.

This is the date of the creation of the split. For a merged split, it is not super useful information.

guilload:
Prefixing files with a datetime has been super useful at my previous gig. I could just go to S3 and quickly identify issues just doing `aws s3 ls <...>/20210911`. For any given time window, I could right away see whether the size of the files was right, whether compaction was happening, whether the GC was running correctly... With ULIDs, this becomes tedious very quickly.

We can get away with it if we build good tooling in the CLI though. How much of a problem are long split IDs?

guilload:
I'm going to close that PR. Let's build good tools instead! A measure of success will be having users not mentioning this PR or opening a similar issue in the future :)

PR #536, opened 2021-09-11
Simplified bundle directory

fulmicoton:
This PR simplifies the bundle directory.

This was required for the merge PR, as it avoids copying the data fetched into a `Vec<u8>`.
In the merge PR, we use this directory with a file slice backed by a Mmap, so that we entirely avoid the copy
(and the anonymous memory).


quickwit-storage/src/bundle_storage.rs
```diff
@@ -111,11 +111,8 @@ impl BundleStorageFileOffsets {
     }
 
     /// Return file offsets for given path.
```
guilload, line 113:
```suggestion
    /// Returns file offsets for given path.
```

quickwit-storage/src/bundle_storage.rs
```diff
@@ -111,11 +111,8 @@ impl BundleStorageFileOffsets {
...
-    pub fn get(&self, path: &Path) -> StorageResult<Range<usize>> {
-        self.files
-            .get(path)
-            .cloned()
-            .ok_or_else(|| io::Error::new(io::ErrorKind::NotFound, "not found in metadata").into())
+    pub fn get(&self, path: &Path) -> Option<Range<usize>> {
+        self.files.get(path).cloned()
     }
 
     /// Return if file exists in metadata.
```
guilload, line 118:
```suggestion
    /// Returns whether file exists in metadata.
```

guilload (approved):
LGTM.
Can you add a test for `exists`, which was lacking in the original PR?

fmassot (approved):

quickwit-storage/src/bundle_storage.rs
fulmicoton: Good point. I added a ticket to merge first, and improve later.

PR #537, opened 2021-09-11
Added UnionDirectory in preparation for merges.

fulmicoton:
### Description

Added an union directory.
It makes it possible to stack different directory implementation as different layers.

It is required for merging splits.

### DIFF
--- a/quickwit-directories/src/union_directory.rs
+++ b/quickwit-directories/src/union_directory.rs
@@ -1,0 +1,161 @@
+// Copyright (C) 2021 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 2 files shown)


evanxg852000 (approved):

quickwit-directories/src/union_directory.rs
```diff
@@ -0,0 +1,161 @@
...
+            Err(DeleteError::FileDoesNotExist(path.to_path_buf()))
+        }
+    }
+
+    fn atomic_write(&self, path: &Path, data: &[u8]) -> io::Result<()> {
+        self.directories[0].atomic_write(path, data)
+    }
+
+    fn watch(&self, _: tantivy::directory::WatchCallback) -> tantivy::Result<WatchHandle> {
+        Ok(WatchHandle::empty())
```
guilload, line 108: Should we prefer `unimplemented!("``UnionDirectory`` does not support watch.")`?

guilload (approved):
LGTM.
This PR could use a few more tests though (`exists`, `delete`, `write`).

quickwit-directories/src/union_directory.rs
fulmicoton: I think this is ok (and required).

PR #538, opened 2021-09-11
Publisher can replace splits

fulmicoton:
### Description
This PR makes it possible for the publisher to handle message requiring it to replace splits (instead of just publishing splits).
It also removes the tag_field from the IndexedSplit message, and gives it directly to the packager.


quickwit-indexing/src/models/indexed_split.rs
```diff
@@ -43,16 +45,18 @@ pub struct IndexedSplit {
     // invalid.
     pub docs_size_in_bytes: u64,
 
-    /// Instant of the instant of the first document in the split.
+    /// Instant of reception of the first document in the indexer.
+    ///
+    /// This is mostly useful to understand part of the time to search.
+    /// Note however, the document may have been waiting for a long time in the source
```
guilload, line 51:
```suggestion
    /// However, note that the document may have been waiting for a long time in the source
```

guilload, line 51: We could actually measure that time if we wanted to. Kafka and Kinesis, at least, provide the arrival time of a message in the queue.

guilload (approved):

PR #540, opened 2021-09-11
Adds a working merge pipeline.

### LINKED ISSUE #399: Merge pipeline
The merge pipeline comes and extends the original pipeline according to the following schema.

merge_pipeline exc

Note that the merge pipeline use the same publisher as the original pipeline.
It also use the same code as the original pipeline packager and the uploader but these in different actor instances.

The rationale is that we want as much as possible for the merge operations not to hurt the time to search. 

Within this pipeline, all actors are spawned and supervised by the supervisor just like it was before.

Note that the pipeline does not constitue a DAG anymore.
Indeed, the publisher needs to inform the merge planner that a new split is available, and the merge pipeline itself also send merged splits for the publisher to publish. 

For this reason, terminating the pipeline (upon end of source for instance) is therefore tricky.

We do this by adding a link from the uploader to the merge planner.
On its termination the uploader informs the merge planner that there won't be any fresh splits.

The merge planner then terminates successfully. The indexing pipeline will just wait for the termination of the running merge operation and then terminate successfully.

(The publisher will ignore the error received when sending to the merge planner.)

fulmicoton:
See detail on #399.
Closes #399

### DIFF
--- a/quickwit-indexing/src/actors/packager.rs
+++ b/quickwit-indexing/src/actors/packager.rs
@@ -34,7 +34,7 @@ use tantivy::schema::Field;
 use tantivy::{ReloadPolicy, SegmentId, SegmentMeta};
 use tracing::*;
 
-use crate::models::{IndexedSplit, PackagedSplit, ScratchDirectory};
+use crate::models::{IndexedSplit, MergePlannerMessage, PackagedSplit, ScratchDirectory};
 
 /// The role of the packager is to get an index writer and
 /// produce a split file.
@@ -49,14 +49,20 @@ use crate::models::{IndexedSplit, PackagedSplit, ScratchDirectory};
 /// The split format is described in `internals/split-format.md`
 pub struct Packager {
     uploader_mailbox: Mailbox<PackagedSplit>,
+    merge_planner_mailbox_opt: Option<Mailbox<MergePlannerMessage>>,
     /// The special field for extracting tags.
     tags_field: Field,
 }
 
 impl Packager {
-    pub fn new(tags_field: Field, uploader_mailbox: Mailbox<PackagedSplit>) -> Packager {
+    pub fn new(
+        tags_field: Field,
+        uploader_mailbox: Mailbox<PackagedSplit>,
+        merge_planner_mailbox_opt: Option<Mailbox<MergePlannerMessage>>,
+    ) -> Packager {
         Packager {
             uploader_mailbox,
+            merge_planner_mailbox_opt,
             tags_field,
         }
     }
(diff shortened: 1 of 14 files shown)


quickwit-indexing/src/actors/packager.rs
```diff
@@ -266,10 +272,25 @@ impl SyncActor for Packager {
...
     }
+
+    fn finalize(
+        &mut self,
+        _exit_status: &quickwit_actors::ActorExitStatus,
+        ctx: &ActorContext<Self::Message>,
+    ) -> anyhow::Result<()> {
+        if let Some(merge_planner_mailbox) = self.merge_planner_mailbox_opt.as_ref() {
+            // We are trying to stop the merge planner.
+            // If merge planner is already dead, this is not an error.
```
fmassot, line 283:
```suggestion
            // If the merge planner is already dead, this is not an error.
```

quickwit-indexing/src/actors/pipeline_supervisor.rs
```diff
@@ -240,7 +331,32 @@ impl IndexingPipelineSupervisor {
...
             match self.healthcheck() {
-                Health::Healthy => {}
+                Health::Healthy => {
+                    // We stop triggering new merges as soon as there cannot be more splits.
+                    if let Some(handlers) = self.handlers.as_mut() {
+                        // If the uploader has terminated (due to end of source) we want to stop
+                        // merging new splits.
+                        //
+                        // This is done by sending a message to the `MergePlanner`.
+                        // We already send this message in the uploader finalizer, but if the
```
fmassot, line 341: actually this is done in the packager finalizer.

quickwit-indexing/src/actors/pipeline_supervisor.rs
```diff
@@ -240,7 +331,32 @@ impl IndexingPipelineSupervisor {
...
+                        // uploader panicks, the finalizer may never be
+                        // called, so we defensively send the message if we
+                        // detect that the uploader is not running, while the merge planner is.
+                        if handlers.uploader.health() != Health::Healthy
```
fmassot, line 345: check the packager instead? 
I'm not sure what is the best: send the message in packager finalizer or in the uploader finalizer. Doing it as soon as possible sounds good for me, so we can keep it into the packager and update the comments and health check.

quickwit-indexing/src/actors/publisher.rs
```diff
@@ -120,6 +125,18 @@ impl AsyncActor for Publisher {
         };
         self.run_publish_operation(&publisher_message).await?;
 
+        let new_splits = publisher_message.operation.extract_new_splits();
+
+        // The merge planner is not necessarily awake and this is not an error.
+        // For instance, when a source reaches its end, and the last "new" split
+        // has been packaged, the packager finalizer sends a message to the merge
+        // in order to stop it.
```
fmassot, line 133:
```suggestion
        // The merge planner is not necessarily awake and this is not an error.
        // For instance, when a source reaches its end, and the last "new" split
        // has been packaged, the packager finalizer sends a message to the merge
        // planner in order to stop it.
```

fmassot (approved):
Sounds good, there is only one important thing to fix, the health check on the package instead of the check on the uploader.

quickwit-indexing/src/actors/pipeline_supervisor.rs
fulmicoton: Great point

PR #544, opened 2021-09-14
Cache storage for indexing

evanxg852000:
### Description
Implemented a local file cache for the indexing pipeline.
Closes  https://github.com/quickwit-inc/quickwit/issues/400

### How was this PR tested?
Run the indexing

### DIFF
--- a/quickwit-storage/src/local_file_storage.rs
+++ b/quickwit-storage/src/local_file_storage.rs
@@ -46,10 +46,25 @@ impl fmt::Debug for LocalFileStorage {
 
 impl LocalFileStorage {
     /// Creates a file storage instance given a uri
+    pub fn from_uri(uri: &str) -> StorageResult<LocalFileStorage> {
+        let root_pathbuf = Self::extract_root_path_from_uri(uri)?;
+        Ok(Self { root: root_pathbuf })
+    }
+
+    /// Validates if the provided uri is of the correct protocol & extracts the root path.
+    ///
     /// Both scheme `file:///{path}` and `file://{path}` are accepted.
     /// If uri starts with `file://`, a `/` is automatically added to ensure
     /// `path` starts from root.
-    pub fn from_uri(uri: &str) -> StorageResult<LocalFileStorage> {
+    pub fn extract_root_path_from_uri(uri: &str) -> StorageResult<PathBuf> {
+        if !uri.starts_with("file://") {
+            let err_msg = anyhow::anyhow!(
+                "{:?} is an invalid file storage uri. Only file:// is accepted.",
+                uri
+            );
+            return Err(StorageErrorKind::DoesNotExist.with_error(err_msg));
+        }
+
         let mut root_path = uri
             .split("://")
             .nth(1)
@@ -69,7 +84,7 @@ impl LocalFileStorage {
             return Err(StorageErrorKind::Io
                 .with_error(anyhow::anyhow!("Invalid uri, `..` is forbidden: {}", uri)));
         }
-        Ok(Self { root: pathbuf })
+        Ok(pathbuf)
     }
 }
 
--- a/quickwit-storage/src/local_file_storage.rs
+++ b/quickwit-storage/src/local_file_storage.rs
@@ -233,14 +233,6 @@ impl StorageFactory for LocalFileStorageFactory {
     }
 
     fn resolve(&self, uri: &str) -> StorageResult<Arc<dyn Storage>> {
-        if !uri.starts_with("file://") {
-            let err_msg = anyhow::anyhow!(
-                "{:?} is an invalid file storage uri. Only file:// is accepted.",
-                uri
-            );
-            return Err(StorageErrorKind::DoesNotExist.with_error(err_msg));
-        }
-
         let storage = LocalFileStorage::from_uri(uri)?;
         Ok(Arc::new(storage))
     }
--- a/quickwit-storage/src/local_file_storage.rs
+++ b/quickwit-storage/src/local_file_storage.rs
@@ -86,6 +86,14 @@ impl LocalFileStorage {
         }
         Ok(pathbuf)
     }
+
+    /// Moves a file from a source to a destination.
+    pub async fn move_to(&self, from: &Path, to: &Path) -> crate::StorageResult<()> {
+        let from_full_path = self.root.join(from);
+        let to_full_path = self.root.join(to);
+        fs::rename(from_full_path, to_full_path).await?;
+        Ok(())
+    }
 }
 
 /// Delete empty directories starting from `{root}/{path}` directory and stopping at `{root}`
--- a/quickwit-storage/src/storage.rs
+++ b/quickwit-storage/src/storage.rs
@@ -83,8 +83,8 @@ impl From<&'static [u8]> for PutPayload {
 /// object storage treat them. This means when directory separators a present
 /// in the storage operation path, the storage implementation should create and remove transparently
 /// these intermediate directories.
-#[cfg_attr(any(test, feature = "testsuite"), mockall::automock)]
 #[async_trait]
+#[cfg_attr(any(test, feature = "testsuite"), mockall::automock)]
 pub trait Storage: Send + Sync + 'static {
     /// Saves a file into the storage.
     async fn put(&self, path: &Path, payload: PutPayload) -> StorageResult<()>;
--- a/quickwit-storage/src/storage.rs
+++ b/quickwit-storage/src/storage.rs
@@ -83,8 +83,8 @@ impl From<&'static [u8]> for PutPayload {
 /// object storage treat them. This means when directory separators a present
 /// in the storage operation path, the storage implementation should create and remove transparently
 /// these intermediate directories.
-#[async_trait]
 #[cfg_attr(any(test, feature = "testsuite"), mockall::automock)]
+#[async_trait]
 pub trait Storage: Send + Sync + 'static {
     /// Saves a file into the storage.
     async fn put(&self, path: &Path, payload: PutPayload) -> StorageResult<()>;
--- a/quickwit-storage/src/storage_with_upload_cache.rs
+++ b/quickwit-storage/src/storage_with_upload_cache.rs
@@ -53,6 +53,15 @@ impl DiskCapacity {
     }
 }
 
+impl Default for DiskCapacity {
+    fn default() -> Self {
+        Self {
+            max_num_files: 1000,
+            max_num_bytes: 10_000_000_000, // 10GB
+        }
+    }
+}
+
 /// CacheState is a struct for serializing/deserializing the cache state.
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct CacheState {
@@ -75,7 +84,7 @@ impl CacheState {
 
 /// A struct to wrap the lru cache and its info.
 pub struct CacheWithMeta {
-    /// Underlying L.R.U cache.
+    /// Underlying LRU cache.
     pub cache: LruCache<PathBuf, usize>,
     /// Current number of bytes in the cache.
     pub num_bytes: usize,
@@ -173,7 +182,7 @@ impl StorageWithUploadCache {
             cache_storage_root,
             disk_capacity: cache_state.disk_capacity,
             disk_cache: Mutex::new(CacheWithMeta {
-                cache: LruCache::unbounded(),
+                cache: disk_cache,
                 num_bytes,
                 num_files,
             }),
@@ -210,11 +219,9 @@ impl Storage for StorageWithUploadCache {
         let mut locked_disk_cache = self.disk_cache.lock().await;
         let payload_length = payload.len().await? as usize;
         // Ignore storing an item that cannot fit in the cache.
-        if self
-            .disk_capacity
-            .exceeds_capacity(payload_length, locked_disk_cache.num_files)
-        {
-            return Ok(());
+        if payload_length > self.disk_capacity.max_num_bytes {
+            warn!("Failed to cache file: file size exceeds total cache capacity.");
+            return Ok(())
         }
 
         if let Some(item_num_bytes) = locked_disk_cache.cache.pop(&path.to_path_buf()) {
@@ -335,23 +342,6 @@ fn atomic_write(path: &Path, content: &[u8]) -> io::Result<()> {
     Ok(())
 }
 
-/// A struct embedding the cache parameters.
-#[derive(Debug, Clone)]
-pub struct CacheConfig {
-    /// The maximum number of files for the local file cache.
-    pub max_num_files: usize,
-    /// The maximum size in bytes allowed for the local file cache.
-    pub max_num_bytes: usize,
-}
-
-impl Default for CacheConfig {
-    fn default() -> Self {
-        Self {
-            max_num_files: 1000,
-            max_num_bytes: 10_000_000_000, // 10GB
-        }
-    }
-}
 
 /// Creates an instance of [`StorageWithUploadCache`].
 ///
@@ -361,7 +351,7 @@ pub fn create_storage_with_upload_cache(
     remote_storage: Arc<dyn Storage>,
     storage_uri_resolver: &StorageUriResolver,
     cache_dir: &Path,
-    config: CacheConfig,
+    capacity: DiskCapacity,
 ) -> crate::StorageResult<Arc<dyn Storage>> {
     let cache_state_file_path = cache_dir.join(CACHE_STATE_FILE_NAME);
     let storage = if cache_state_file_path.exists() {
@@ -374,8 +364,8 @@ pub fn create_storage_with_upload_cache(
         StorageWithUploadCache::create(
             remote_storage,
             cache_storage,
-            config.max_num_files,
-            config.max_num_bytes,
+            capacity.max_num_files,
+            capacity.max_num_bytes,
         )?
     };
 
--- a/quickwit-storage/src/storage_with_upload_cache.rs
+++ b/quickwit-storage/src/storage_with_upload_cache.rs
@@ -89,22 +89,23 @@ impl CacheState {
 }
 
 /// A struct to wrap the LRU cache and its info.
-pub struct CacheWithMeta {
+struct CacheWithMeta {
     /// Underlying LRU cache.
-    pub cache: LruCache<PathBuf, usize>,
+    cache: LruCache<PathBuf, usize>,
     /// Current number of bytes in the cache.
-    pub num_bytes: usize,
+    num_bytes: usize,
     /// Current number of files in the cache.
-    pub num_files: usize,
+    num_files: usize,
 }
 
-/// A storage with a backing file system cache.
+/// A storage keeping the most recently uploaded files on local disk.
 ///
-/// This storage is meant to be use during indexing where `put` and `copy_to_file`
-/// are most relevant.
-/// It only interract with the cache for `put`, `copy_to_file` and `delete`
-/// operation.
-/// Other operations directly interract with the underlying remote storage.
+/// This storage is meant to be used during indexing and merging where the `put` and `copy_to_file`
+/// methods are the most utilized.
+/// As a result, only the methods `put`, `copy_to_file`, and `delete` interact with the local file
+/// system.
+/// Other operations, such as `get_all` or `get_slice,` directly interact with the underlying remote
+/// storage.
 pub struct StorageWithUploadCache {
     /// The remote storage.
     remote_storage: Arc<dyn Storage>,
@@ -245,15 +246,20 @@ impl Storage for StorageWithUploadCache {
             locked_disk_cache.num_files -= 1;
         }
 
-        let mut undable_to_delete = HashSet::new();
+        // There is a risk of having one or a few items in the cache
+        // that cannot be deleted from disk. This will cause an infinite loop.
+        // This hashset is used to detect that we have looped around by checking if
+        // an item that failed deletion has already been seen.
+        let mut failed_to_delete_items = HashSet::new();
+
         while self.disk_capacity.exceeds_capacity(
             locked_disk_cache.num_bytes + payload_length,
             locked_disk_cache.num_files + 1,
         ) {
             if let Some((item_path, item_num_bytes)) = locked_disk_cache.cache.pop_lru() {
-                // check for cycle
-                if undable_to_delete.contains(&item_path) {
-                    // save cache state, because we may have succeeeded in deleting items
+                // Check if the failed to delete item has already been encountered.
+                if failed_to_delete_items.contains(&item_path) {
+                    // Save cache state, because we may have succeeeded in deleting some items
                     // while trying to make room.
                     locked_disk_cache.cache.put(item_path, item_num_bytes);
                     let _ = self.save_state(&locked_disk_cache).await;
@@ -262,7 +268,7 @@ impl Storage for StorageWithUploadCache {
                 if let Err(error) = self.cache_storage.delete(item_path.as_path()).await {
                     if error.kind() != StorageErrorKind::DoesNotExist {
                         warn!(file_path = %item_path.to_string_lossy(), "Could not remove file from cache.");
-                        undable_to_delete.insert(item_path.clone());
+                        failed_to_delete_items.insert(item_path.clone());
                         locked_disk_cache.cache.put(item_path, item_num_bytes);
                         continue;
                     }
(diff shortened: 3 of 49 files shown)


quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+    ///
+    /// It needs a folder `path` previously used by an instance of [`StorageWithUploadCache`].
+    /// Lastly, the file at `{path}/[`CACHE_STATE_FILE_NAME`]` should be present with valid content.
+    pub fn from_path(
+        remote_storage: Arc<dyn Storage>,
+        storage_uri_resolver: &StorageUriResolver,
+        path: &Path,
+    ) -> StorageResult<Self> {
+        let cache_state = CacheState::from_path(path)?;
+        let local_storage = storage_uri_resolver
```
fmassot, line 142: It feels strange to store the `local_storage_uri` in the cache state.
Why not remove `local_storage_uri` from the cache state and just use the path? 
By the way, renaming `path` by `cache_dir` would make things clearer.

Can we say that `local_storage` is the `cache_storage`?

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+}
+
+#[async_trait]
+impl Storage for StorageWithUploadCache {
+    async fn put(&self, path: &Path, payload: PutPayload) -> StorageResult<()> {
+        self.remote_storage.put(path, payload.clone()).await?;
+
+        let mut locked_disk_cache = self.disk_cache.lock().await;
+        let payload_length = payload.len().await? as usize;
+        if self.exceeds_capacity(payload_length, locked_disk_cache.num_files + 1) {
```
fmassot, line 224: We can add a comment like "Don't cache payload > max capacity". I don't really understand why there is this condition: "locked_disk_cache.num_files + 1".

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+            disk_capacity: cache_state.disk_capacity,
+            disk_cache: Mutex::new(WrappedCache {
+                lru: LruCache::unbounded(),
+                num_bytes,
+                num_files,
+            }),
+        })
+    }
+
+    /// Check if an incomming item will exceed the capacity once inserted.
```
fmassot, line 186: Comment is misleading. Here we need to pass the total num bytes and total num files to check if we exceed capacity.
/// Returns false if `num_bytes` or `num_files` are greater than disk capacity.

quickwit-storage/src/storage.rs
```diff
@@ -83,8 +83,8 @@ impl From<&'static [u8]> for PutPayload {
 /// object storage treat them. This means when directory separators a present
 /// in the storage operation path, the storage implementation should create and remove transparently
 /// these intermediate directories.
-#[cfg_attr(any(test, feature = "testsuite"), mockall::automock)]
 #[async_trait]
+#[cfg_attr(any(test, feature = "testsuite"), mockall::automock)]
```
fmassot, line 87: any reason for this change?

fmassot (commented):
Sound goods, check my comments, I still need to look at the tests more deeply.

quickwit-storage/src/storage_with_upload_cache.rs
evanxg852000: > Can we say that local_storage is the cache_storage?
yes

quickwit-storage/src/storage_with_upload_cache.rs
evanxg852000: Oh yes that's a bug.

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+        Ok(())
+    }
+
+    async fn copy_to_file(&self, path: &Path, output_path: &Path) -> StorageResult<()> {
+        let mut locked_disk_cache = self.disk_cache.lock().await;
+        if locked_disk_cache.lru.get(&path.to_path_buf()).is_none() {
+            return self.remote_storage.copy_to_file(path, output_path).await;
+        }
+
+        let _ = self.save_state(&locked_disk_cache).await;
```
fmassot, line 270: why do we need to save the state here?

quickwit-storage/src/storage.rs
evanxg852000: Yes, so placing the `mockall::` after make the async methods visible to the Mock, without this change, the async function are just generated in mockall as normal functions. I needed this for testing concurrency on the storage. this test has been scrapped because it's not needed, but I kept this as it convey what we do here.

quickwit-storage/src/storage_with_upload_cache.rs
fmassot: same remark for `get_slice`, `get_all`

evanxg852000: The lru internal structure has moved so is the list of items ordering.
But we can also leave out this and only save when something is added or evicted.

quickwit-storage/src/local_file_storage.rs
```diff
@@ -46,10 +46,25 @@ impl fmt::Debug for LocalFileStorage {
...
+    }
+
+    /// Validates if the provided uri is of the correct protocol & extracts the root path.
+    ///
     /// Both scheme `file:///{path}` and `file://{path}` are accepted.
     /// If uri starts with `file://`, a `/` is automatically added to ensure
     /// `path` starts from root.
-    pub fn from_uri(uri: &str) -> StorageResult<LocalFileStorage> {
+    pub fn extract_root_path_from_uri(uri: &str) -> StorageResult<PathBuf> {
+        if !uri.starts_with("file://") {
```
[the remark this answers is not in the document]
evanxg852000: fixed

quickwit-storage/src/storage_with_upload_cache.rs
fmassot: We need to clarify the usage of this cache layer. What is it for? It would be cool to explain that in the docs of `StorageWithUploadCache`
From my understanding, this is essentially to avoid a costly `copy_to_file`.  We don't want to hit the cache for other operations except for put. Right?

fmassot: If you agree, we should never touch the cache except for `put` and `copy_to_file`.

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+            LocalFileStorage::extract_root_path_from_uri(&cache_state.local_storage_uri)?;
+        if local_storage_root_path != path.to_path_buf() {
+            return Err(StorageErrorKind::InternalError.with_error(anyhow::anyhow!(
+                "Invalid root path: {}, but {} expected ",
+                path.to_string_lossy(),
+                local_storage_root_path.to_string_lossy()
+            )));
+        }
+
+        let num_files = cache_state.items.len();
```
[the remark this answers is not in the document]
evanxg852000: Not sure this is a great idea, num_files & num_bytes are used afterward. also this check happens only here

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+        let reader = std::io::BufReader::new(json_file);
+        serde_json::from_reader(reader)
+            .map_err(|err| StorageErrorKind::InternalError.with_error(err))
+    }
+}
+
+/// A struct to wrap the lru cache and its info.
+pub struct WrappedCache {
+    /// Underlying L.R.U cache.
+    pub lru: LruCache<PathBuf, usize>,
```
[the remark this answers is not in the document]
evanxg852000: Ok

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+        let file_path = path.to_path_buf().join(CACHE_STATE_FILE_NAME);
+        let json_file = std::fs::File::open(file_path)?;
+        let reader = std::io::BufReader::new(json_file);
+        serde_json::from_reader(reader)
+            .map_err(|err| StorageErrorKind::InternalError.with_error(err))
+    }
+}
+
+/// A struct to wrap the lru cache and its info.
+pub struct WrappedCache {
```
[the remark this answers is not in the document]
evanxg852000: Ok

quickwit-storage/src/storage_with_upload_cache.rs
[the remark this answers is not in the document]
fmassot: ah yes, I did not see the usage of num_bytes at the end.

quickwit-storage/src/storage_with_upload_cache.rs
evanxg852000: Yeah the `+1` is a bug

quickwit-storage/src/storage_with_upload_cache.rs
evanxg852000: I feel the comment is ok. but probably the params names
```rust
fn exceeds_capacity(&self, total_bytes_after_insert: usize, total_files_after_insert: usize) -> bool {
        self.disk_capacity.exceeds_capacity(total_bytes_after_insert, total_files_after_insert)
    }
```

quickwit-storage/src/storage.rs
fulmicoton: I remember you mentionned that in the chat :)

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+use bytes::Bytes;
+use lru::LruCache;
+use serde::{Deserialize, Serialize};
+use tempfile::Builder;
+use tokio::sync::{Mutex, MutexGuard};
+
+use crate::{
+    LocalFileStorage, PutPayload, Storage, StorageErrorKind, StorageResult, StorageUriResolver,
+};
+
```
fulmicoton, line 35: it still shows `cache-sate.json`

quickwit-storage/src/storage_with_upload_cache.rs
fulmicoton: I feel the function is not super helpful :)

quickwit-storage/src/storage_with_upload_cache.rs
evanxg852000: I think this is was a refresh issue. I also saw the change was not there but now I am seeing it.

quickwit-storage/src/storage_with_upload_cache.rs
fulmicoton: I don't think we want a `local_storage` at all for the local caching directory.

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+    /// Check if an incomming item will exceed the capacity once inserted.
+    fn exceeds_capacity(&self, num_bytes: usize, num_files: usize) -> bool {
+        self.disk_capacity.exceeds_capacity(num_bytes, num_files)
+    }
+
+    /// Persist the state of the entire cache.
+    ///
+    /// Takes a previously locked guard for easier mutext management.
+    async fn save_state<'a>(
```
fulmicoton, line 195: is the lifetime required?

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+    }
+
+    async fn get_all(&self, path: &Path) -> StorageResult<Bytes> {
+        let mut locked_disk_cache = self.disk_cache.lock().await;
+        if locked_disk_cache.lru.get(&path.to_path_buf()).is_none() {
+            return self.remote_storage.get_all(path).await;
+        }
+
+        let _ = self.save_state(&locked_disk_cache).await;
+        self.local_storage.get_all(path).await
```
fulmicoton: we don't care about get_all. Really we can focus on `copy_to_file` and `put`

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+        self.local_storage.get_slice(path, range).await
+    }
+
+    async fn get_all(&self, path: &Path) -> StorageResult<Bytes> {
+        let mut locked_disk_cache = self.disk_cache.lock().await;
+        if locked_disk_cache.lru.get(&path.to_path_buf()).is_none() {
+            return self.remote_storage.get_all(path).await;
+        }
+
+        let _ = self.save_state(&locked_disk_cache).await;
```
fulmicoton:
```suggestion
```

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+    }
+
+    async fn get_slice(&self, path: &Path, range: Range<usize>) -> StorageResult<Bytes> {
+        let mut locked_disk_cache = self.disk_cache.lock().await;
+        if locked_disk_cache.lru.get(&path.to_path_buf()).is_none() {
+            return self.remote_storage.get_slice(path, range).await;
+        }
+
+        let _ = self.save_state(&locked_disk_cache).await;
+        self.local_storage.get_slice(path, range).await
```
fulmicoton:
```suggestion
        self.remote_storage.get_slice(path, range).await
```

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+        let _ = self.save_state(&locked_disk_cache).await;
+        self.local_storage.get_all(path).await
+    }
+
+    async fn delete(&self, path: &Path) -> StorageResult<()> {
+        self.remote_storage.delete(path).await?;
+
+        let mut locked_disk_cache = self.disk_cache.lock().await;
+        if let Some(item_num_bytes) = locked_disk_cache.lru.pop(&path.to_path_buf()) {
+            if self.local_storage.delete(path).await.is_err() {
```
fulmicoton: Can you check the error and log if it is not a file missing error?

Also, if the error is about the file missing locally, we should update the state.

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,885 @@
...
+            locked_disk_cache.num_files + 1,
+        ) {
+            if let Some((item_path, item_num_bytes)) = locked_disk_cache.lru.pop_lru() {
+                if self
+                    .local_storage
+                    .delete(item_path.as_path())
+                    .await
+                    .is_err()
+                {
+                    return Ok(());
```
fulmicoton, line 248: Certainly returning `Ok(())` here is very wrong. We haven't save the file yet.

fulmicoton (changes requested):
A bunch of nitpicks and a bunch of bugs :)

quickwit-storage/src/storage_with_upload_cache.rs
evanxg852000: > I don't think we want a local_storage at all for the local caching directory.

Not sure about this comment. do you mean to get rid of the FileStorage and use `fs` routines?

quickwit-storage/src/storage_with_upload_cache.rs
evanxg852000: Well not sure of this. I can be more elaborate on handling the error for non-exiting files.
But not being able to proceed just means we cannot store the item in the cache and as long as we have it the remote storage, I think `Ok(())` result is fine from the `Storage` trait point of view.

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,881 @@
...
+impl Storage for StorageWithUploadCache {
+    async fn put(&self, path: &Path, payload: PutPayload) -> StorageResult<()> {
+        self.remote_storage.put(path, payload.clone()).await?;
+
+        let mut locked_disk_cache = self.disk_cache.lock().await;
+        let payload_length = payload.len().await? as usize;
+        // Ignore storing an item that cannot fit in the cache.
+        if self
+            .disk_capacity
+            .exceeds_capacity(payload_length, locked_disk_cache.num_files)
```
guilload, line 215: Did you want to write `.exceeds_capacity(payload_length, 1)`? With the current form, the logic is massively flawed, and your implementation acts as a cache for the first `max_num_files` and then does nothing:

Let assume cache capacity is 2 files, ignoring the file sizes.
file `#1`: exceeds_capacity(_, 0) -> false -> put file in cache
file `#2`: exceeds_capacity(_, 1) -> false -> put file in cache
file `#3`: exceeds_capacity(_, 2) -> true -> do nothing.

I suggest we go with something simpler:
```rust
if payload_length > self.disk_capacity.max_num_bytes {
  warn!("Failed to cache file: file size exceeds total cache capacity.");
  return Ok(())
}
```

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,1171 @@
...
+            locked_disk_cache.num_files -= 1;
+        }
+
+        let mut undable_to_delete = HashSet::new();
+        while self.disk_capacity.exceeds_capacity(
+            locked_disk_cache.num_bytes + payload_length,
+            locked_disk_cache.num_files + 1,
+        ) {
+            if let Some((item_path, item_num_bytes)) = locked_disk_cache.cache.pop_lru() {
+                // check for cycle
```
guilload, line 254: Can you expand on what kind of cycles can we expect and when they do happen?

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,1171 @@
...
+                if undable_to_delete.contains(&item_path) {
+                    // save cache state, because we may have succeeeded in deleting items
+                    // while trying to make room.
+                    locked_disk_cache.cache.put(item_path, item_num_bytes);
+                    let _ = self.save_state(&locked_disk_cache).await;
```
guilload, line 259: Is it reasonable to ignore a potential error here?

guilload (commented):
Making good progress.

The semantics of this cache are very specific since it only acts on `copy_to_file`.I'll leave it up to you but this is maybe worth a rename.

quickwit-storage/src/storage_with_upload_cache.rs
evanxg852000: Yes. I will explain in the comment.
we can end up with files that cannot be removed remaining in the cache and creating an infinite loop.

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,1171 @@
...
+            max_num_bytes: 10_000_000_000, // 10GB
+        }
+    }
+}
+
+/// CacheState is a struct for serializing/deserializing the cache state.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CacheState {
+    /// The disk capacity
+    pub disk_capacity: DiskCapacity,
```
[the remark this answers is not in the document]
evanxg852000: Yes if you had a previous state, we ignore the new capacity at the moment.

quickwit-storage/src/storage_with_upload_cache.rs
evanxg852000: To me yes, following the same semantic that we successfully stored in the remote storage. and the cache part failed.

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,1171 @@
...
+}
+
+/// A storage with a backing file system cache.
+///
+/// This storage is meant to be use during indexing where `put` and `copy_to_file`
+/// are most relevant.
+/// It only interract with the cache for `put`, `copy_to_file` and `delete`
+/// operation.
+/// Other operations directly interract with the underlying remote storage.
+pub struct StorageWithUploadCache {
```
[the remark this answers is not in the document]
evanxg852000: I started this refactoring, but given the semantics I am not so sure now if it's a good idea to leave out the notion of caching.
@fulmicoton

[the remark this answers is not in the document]
guilload: You don't have to name the cache file after the struct it represents (`quickwit.json`). Additionally, the semantics of a class API has nothing to do with the name of its private variable.

guilload:
Something that occurred to me when thinking about this PR this morning, is that the cache state file is pretty much useless here. We wanted to use a state file so we could store the most recent access time, which is not implemented in this PR. If we're not going to save that information, then we can get rid of the state file and list the files, and get their sizes from the fs upon instantiating the cache.

evanxg852000:
@guilload 
My reasoning was that having a state file helps only consider things that got in via our API (put, delete). 
At some point back I was even storing the remote storage URI and validating it. We decided it was just of to use it with any remote storage. So if working with the stored state does not add much value to the solution, I can remove it. I reckon that not having to worry about saving simplifies a lot.

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,630 @@
...
+    // The underlying cache.
+    cache_items: Mutex<HashMap<PathBuf, usize>>,
+    /// The parameters of the cache.
+    cache_params: CacheParams,
+}
+
+impl StorageWithUploadCache {
+    /// Create an instance of [`StorageWithUploadCache`]
+    ///
+    /// It needs both the remote to work with.
```
fmassot, line 82: you mean "It needs both the remote and local storage"?

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,630 @@
...
+        };
+
+        // Ingore storing when maximum number of files is reached.
+        if num_entries + 1 > self.cache_params.max_num_files {
+            warn!("Failed to cache file: maximum number of files exceeded.");
+            return Ok(());
+        }
+
+        let payload_length = payload.len().await? as usize;
+        // Ignore storing item whose size exeeds the size in bytes per item.
```
fmassot, line 162:
```suggestion
        // Ignore storing item whose size exceeds the max file size.
```

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,630 @@
...
+use bytes::Bytes;
+use serde::{Deserialize, Serialize};
+use tokio::sync::Mutex;
+use tracing::warn;
+
+use crate::{LocalFileStorage, PutPayload, Storage, StorageErrorKind, StorageResult};
+
+// TODO(asking): could we make it hidden with `.quickwit-cache`
+const INTERNAL_CACHE_DIR_NAME: &str = "quickwit-cache";
+const CACHE_TEMP_FILE_EXTENSION: &str = "_quickwit_cache_temp_file";
```
fmassot, line 36: I don't know which files are considered as `quickwit cache temp` files, can you give an example to better understand this case?

quickwit-storage/src/storage_with_upload_cache.rs
```diff
@@ -0,0 +1,630 @@
...
+        for dir_entry_result in fs::read_dir(local_storage_root.as_path())? {
+            let dir_entry = dir_entry_result?;
+            let path = dir_entry.path();
+            if path.is_file() {
+                let relative_file_path = path
+                    .strip_prefix(local_storage_root.as_path())
+                    .map_err(|err| StorageErrorKind::InternalError.with_error(err))?;
+                let file_size = dir_entry.metadata()?.len() as usize;
+
+                // TODO: check if special temp file and remove it
```
fmassot, line 101: here we need to remove files with `CACHE_TEMP_FILE_EXTENSION`?

evanxg852000: Yes

quickwit-storage/src/storage_with_upload_cache.rs
evanxg852000: this is a work in progress, I still need to est and add comment

fulmicoton:
Great job!

fulmicoton (changes requested):

fulmicoton (changes requested):
Minor comments and 1 bi problem:
I think the temp_child usage in the supervisor is wrong,

fulmicoton (approved):

PR #545, opened 2021-09-14
Tags is a mandatory field in the schema. Panic if it's not there.

### LINKED ISSUE #521: Schema constraint Schizophrenia
`IndexConfig::tags_field` seem to suggest that not having tag field is possible, but we actually panic in the indexer if there are no tag field.

I am not sure what we should do:
- Make the tag field really optional
- Add the tag field automagically in the schema generated by the indexconfig.
- Document somewhere that not having a fast field is an error.
- panic in `IndexConfig::tags_field` if none.

fmassot:
Fix #521

### DIFF
--- a/quickwit-index-config/src/config.rs
+++ b/quickwit-index-config/src/config.rs
@@ -108,7 +108,7 @@ pub trait IndexConfig: Send + Sync + Debug + DynClone + 'static {
     fn tags_field(&self, split_schema: &Schema) -> Field {
         split_schema
             .get_field(TAGS_FIELD_NAME)
-            .expect("Tags field must exist on the schema.")
+            .expect("Tags field must exist in the schema.")
     }
 }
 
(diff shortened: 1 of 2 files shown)


quickwit-index-config/src/config.rs
```diff
@@ -105,8 +105,10 @@ pub trait IndexConfig: Send + Sync + Debug + DynClone + 'static {
     }
 
     /// Returns the special tags field if any.
-    fn tags_field(&self, split_schema: &Schema) -> Option<Field> {
-        split_schema.get_field(TAGS_FIELD_NAME)
+    fn tags_field(&self, split_schema: &Schema) -> Field {
+        split_schema
+            .get_field(TAGS_FIELD_NAME)
+            .expect("Tags field must exist on the schema.")
```
guilload, line 111:
```suggestion
            .expect("Tags field must exist in the schema.")
```

PR #546, opened 2021-09-14
[DRAFT] Attempt to speed up CI

guilload:
### Description
Attempt to speed up CI by adding a second restore key.


guilload:
Moot.

PR #550, opened 2021-09-15
Improve error messages

guilload:
### Description
Got some errors while playing with the CLI and the error messages were not helpful. I was able to found the issues only after displaying the path of the metastore files.


fmassot (approved):
Nice catch.

PR #555, opened 2021-09-16
Enable test for PostgreSQL-based metastore on CI

mosuka:
### Description
Enable test for PostgreSQL-based metastore on CI.
Also, changed the feature name from postgresql to postgres.


quickwit-metastore/docker-compose.yml
```diff
@@ -1,9 +1,10 @@
-version: "3.9" 
+version: "3.9"
```
guilload, line 1: I recently committed a docker-compose file at the root of the project; we can work off that one and delete this one.

quickwit-metastore/.env
```diff
@@ -1,4 +1,5 @@
 POSTGRES_USER=quickwit-dev
 POSTGRES_PASSWORD=quickwit-dev
-DATABASE_URL=postgres://quickwit-dev:quickwit-dev@localhost/quickwit-dev
-TEST_DATABASE_URL=postgres://quickwit-dev:quickwit-dev@localhost/quickwit-dev
+POSTGRES_DB=quickwit-metastore
+DATABASE_URL=postgres://quickwit-dev:quickwit-dev@localhost:5432/quickwit-metastore
```
guilload, line 4:
```suggestion
DATABASE_URL=postgres://quickwit-dev:quickwit-dev@localhost:5432/quickwit-metastore-dev
```

quickwit-metastore/.env
```diff
@@ -1,4 +1,5 @@
...
+TEST_DATABASE_URL=postgres://quickwit-dev:quickwit-dev@localhost:5432/quickwit-metastore
```
guilload, line 5: Why do we need both `DATABASE_URL` and `TEST_DATABASE_URL` if they are the same? I'd rather keep `TEST_DATABASE_URL,` which has a more explicit name.

quickwit-metastore/README.md
```diff
@@ -69,7 +69,7 @@ $ docker-compose up --force-recreate -d
 Once PostgreSQL is up and running, you can run tests including PostgresqlMetastore with the following command.
 
 ```
-$ cargo test --features=postgresql
+$ cargo test --features=postgres
```
guilload, line 72: You'll have to modify this README to refer to the "main" docker-compose: `make docker-compose-up DOCKER_SERVICES=postgres`

guilload (changes requested):

quickwit-metastore/.env
mosuka: Yes, we don't need the `DATABASE_URL` here. I'll delete it.

mosuka:
@guilload 
Thank you for the review.
I've updated the branch, so please take another look.

quickwit-metastore/.env
PSeitz: If this is only for tests then the file location is not ideal

mosuka: @PSeitz 
You're right, it's better to put it in the same place as docker-compose.yml.

PR #557, opened 2021-09-16
Read `SourceConfig` from file

guilload:
### Description
Read `SourceConfig` from file. 

### How was this PR tested?
- added a unit test
- indexed data with Kafka source loading the Kafka source conf from local file

### DIFF
--- a/quickwit-cli/src/lib.rs
+++ b/quickwit-cli/src/lib.rs
@@ -20,6 +20,7 @@
 use std::collections::VecDeque;
 use std::convert::TryFrom;
 use std::env;
+use std::fs::File;
 use std::io::{stdout, Stdout, Write};
 use std::path::PathBuf;
 use std::sync::Arc;
@@ -102,6 +103,7 @@ pub struct IndexDataArgs {
     pub metastore_uri: String,
     pub index_id: String,
     pub input_path: Option<PathBuf>,
+    pub source_config_path: Option<PathBuf>,
     pub temp_dir: Option<PathBuf>,
     pub heap_size: Byte,
     pub overwrite: bool,
@@ -155,11 +157,13 @@ pub async fn create_index_cli(args: CreateIndexArgs) -> anyhow::Result<()> {
 }
 
 pub async fn index_data_cli(args: IndexDataArgs) -> anyhow::Result<()> {
-    debug!(args=?args, "indexing");
+    debug!(args = ?args, "index-data");
     quickwit_telemetry::send_telemetry_event(TelemetryEvent::IndexStart).await;
 
-    let input_path = args.input_path.clone();
-    let source_config = create_source_config_from_args(input_path.clone());
+    let source_config_path_opt = args.source_config_path.as_ref();
+    let input_path_opt = args.input_path.as_ref();
+    let source_config =
+        create_source_config_from_args(source_config_path_opt, input_path_opt).await?;
     let scratch_directory = if let Some(scratch_root_path) = args.temp_dir.as_ref() {
         ScratchDirectory::new_in_path(scratch_root_path.clone())
     } else {
@@ -195,7 +199,7 @@ pub async fn index_data_cli(args: IndexDataArgs) -> anyhow::Result<()> {
         universe.spawn_actor(indexing_supervisor).spawn_async();
 
     let is_stdin_atty = atty::is(atty::Stream::Stdin);
-    if args.input_path.is_none() && is_stdin_atty {
+    if args.source_config_path.is_none() && args.input_path.is_none() && is_stdin_atty {
         let eof_shortcut = match env::consts::OS {
             "windows" => "CTRL+Z",
             _ => "CTRL+D",
@@ -206,7 +210,6 @@ pub async fn index_data_cli(args: IndexDataArgs) -> anyhow::Result<()> {
             eof_shortcut
         );
     }
-
     let statistics =
         start_statistics_reporting_loop(supervisor_handler, args.input_path.clone()).await?;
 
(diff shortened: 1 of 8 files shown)


quickwit-indexing/src/source/mod.rs
```diff
@@ -161,7 +162,7 @@ pub fn quickwit_supported_sources() -> &'static SourceLoader {
     })
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
```
fulmicoton, line 165: PartialEq might be a little ambitious... You'd be surprised by the semantics of PartialEq of a json object in general, and `serde_json::Value` in particular.

quickwit-cli/src/lib.rs
```diff
@@ -220,15 +223,40 @@ pub async fn index_data_cli(args: IndexDataArgs) -> anyhow::Result<()> {
...
-    SourceConfig {
-        id: "cli_source".to_string(),
-        source_type: "file".to_string(),
-        params: serde_json::to_value(params).unwrap(),
+async fn create_source_config_from_args(
+    source_config_path_opt: Option<&PathBuf>,
+    input_path_opt: Option<&PathBuf>,
+) -> anyhow::Result<SourceConfig> {
+    if let Some(source_config_path) = source_config_path_opt {
+        let source_config_file = File::open(source_config_path).with_context(|| {
```
fulmicoton, line 231: nitpick.
`fs::read` works fine for small files. It avoid building the `BufReader`.

Another nice thing is that your entirely separate the io from the deserialization, making the source of the error clearer.
https://doc.rust-lang.org/stable/std/fs/fn.read.html

quickwit-cli/src/lib.rs
```diff
@@ -220,15 +223,40 @@ pub async fn index_data_cli(args: IndexDataArgs) -> anyhow::Result<()> {
...
-    };
-    SourceConfig {
-        id: "cli_source".to_string(),
-        source_type: "file".to_string(),
-        params: serde_json::to_value(params).unwrap(),
+async fn create_source_config_from_args(
+    source_config_path_opt: Option<&PathBuf>,
+    input_path_opt: Option<&PathBuf>,
+) -> anyhow::Result<SourceConfig> {
+    if let Some(source_config_path) = source_config_path_opt {
```
fulmicoton, line 230: We should return an error if both Option are "Some"

quickwit-cli/src/lib.rs
```diff
@@ -220,15 +223,40 @@ pub async fn index_data_cli(args: IndexDataArgs) -> anyhow::Result<()> {
...
 
-fn create_source_config_from_args(input_path_opt: Option<PathBuf>) -> SourceConfig {
-    let params = FileSourceParams {
-        filepath: input_path_opt,
-    };
-    SourceConfig {
-        id: "cli_source".to_string(),
-        source_type: "file".to_string(),
-        params: serde_json::to_value(params).unwrap(),
+async fn create_source_config_from_args(
```
fulmicoton, line 226: One line or too of doc to explain the intent could be nice.

quickwit-cli/src/lib.rs
```diff
@@ -492,6 +520,60 @@ impl ThroughputCalculator {
...
 
+#[cfg(test)]
+mod tests {
+    use serde_json::json;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_create_source_config_from_args() -> anyhow::Result<()> {
+        {
```
fulmicoton, line 531: nitpick. 2 unit tests would have been nicer.

fulmicoton (changes requested):
Can you put some doc somewhere with an example conf?

Probably the actually documentation is the best place.
(It is not published automatically so don't worry)

Another place is ok too... The main point is to save other devs 10mn of reverse engineering.

guilload:
I made the changes you requested. For now, I added the doc to `SourceConfig`. I'll update our primary documentation with all the updates right before our next release.

Finally, I took the liberty to rename `SourceConfig.id` to `SourceConfig.source_id` to be more consistent with our naming conventions (`index_id`, `split_id`).

quickwit-indexing/src/source/mod.rs
```diff
@@ -161,9 +162,27 @@ pub fn quickwit_supported_sources() -> &'static SourceLoader {
     })
 }
 
-#[derive(Clone)]
+/// A `SourceConfig` describes the properties of a source. A source config can be created
+/// dynamically or loaded from a file consisting of a JSON object with 3 mandatory properties:
+/// - `source_id`, a name identifying the source uniquely;
+/// - `source_type`, the type of the target source, for instance, `kafka` or `kinesis`;
```
fmassot, line 168: do we have a working kinesis source?

quickwit-indexing/src/source/mod.rs
```diff
@@ -161,9 +162,27 @@ pub fn quickwit_supported_sources() -> &'static SourceLoader {
...
+///         "group_id": "my-kafka-source-consumer-group",
+///         "topic": "my-kafka-source-topic"
+///     }
+/// }
+/// ```
+#[derive(Clone, Serialize, Deserialize)]
 pub struct SourceConfig {
-    pub id: String,
+    pub source_id: String,
     pub source_type: String,
```
fmassot, line 186: it seems strange to have a source type as a `String`.

fmassot (approved):

quickwit-indexing/src/source/mod.rs
guilload: No, not yet. I'll remove the mention to Kinesis.

PR #560, opened 2021-09-17
Fix single file metastore race condition

### LINKED ISSUE #411: Race conditon in the SingleFileMetastore
Right now there is a race condition pretty much everywhere in the single file metastore.

While we do not allow for several processes to edit the single file metastore concurrently, it is possible for several threads within a given process to use it concurrently.

(for instance, the uploader and the publisher)

Today, the interaction local caching happening in the single file metastore can lead to an operation erasing the effect of another operation.

The solution might be to hold a lock. Maybe make sure this is not one that can get poisoned...

mosuka:
### Description
Wraps the current implementation with a `Mutex`.

### How was this PR tested?

Close #411


quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -71,6 +72,7 @@ impl SingleFileMetastore {
         SingleFileMetastore {
             storage,
             cache: Arc::new(RwLock::new(HashMap::new())),
```
guilload, line 74: Do we still need that lock if all the ops are behind a mutex?

quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -691,4 +707,32 @@ mod tests {
             .unwrap();
         assert!(!split.is_empty());
     }
+
+    #[tokio::test]
+    async fn test_single_file_metastore_race_condition() {
```
guilload, line 712: I don't see how this test creates a potential race condition.

quickwit-metastore/src/metastore/single_file_metastore.rs
mosuka: `cache` is wrapped in `Arc` to safely share ownership between threads.
However, `cache` wrapped with `Arc` can only be accessed immutable. To be able to change the value from multiple threads, I think need to use `RwLock` or `Mutex`.
Do you know a better way?

quickwit-metastore/src/metastore/single_file_metastore.rs
mosuka: The test was still unfinished. It has just been updated.

quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -735,4 +751,70 @@ mod tests {
             MetastoreError::InternalError { .. }
         ));
     }
+
+    #[tokio::test]
+    async fn test_single_file_metastore_race_condition() {
```
guilload, line 756: Does that test fail (at least sometimes) without your fix?

quickwit-metastore/src/metastore/single_file_metastore.rs
fulmicoton: @guilload but the borrow checker will never accept that right?

So  should we use the rwlock as a the mutex?

guilload: Right, @fulmicoton. What I was trying to say is that we can do with one lock and a simple design to do so is to encapsulate the mutable state in some `inner` struct enclosed in a `Mutex`.

mosuka: @fulmicoton @guilload Thank you for the suggestion.

quickwit-metastore/src/metastore/single_file_metastore.rs
evanxg852000: @guilload I changed the test using threads and now it fails sporadically on the old implementation.
Sometimes `SplitDoesNotExit` error, other times published splits don't match the expected.

guilload (commented):
I think we can clean up the design a little bit and write everything that way:
```rust
struct Inner { ... }

impl Inner {
  async create_index(...) -> ... { ... }
  async delete_index(...) -> ... { ... }
  ...
}

struct SFMetastore { inner: Arc<Mutex<Inner>>, ... }

impl Metastore for SFMetastore {
  async create_index(...) -> ... { let inner = self.inner.lock().await; inner.create_index(...).await }
  ...
}
```
This way the awkward helpers are going to disappear and it'll be obvious that the outer implementation is using the lock correctly.

quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -51,10 +51,25 @@ fn is_disjoint(left: &Range<i64>, right: &RangeInclusive<i64>) -> bool {
...
+            cache: HashMap::new(),
+        }
+    }
+}
+
 /// Single file metastore implementation.
 pub struct SingleFileMetastore {
-    storage: Arc<dyn Storage>,
-    cache: Arc<RwLock<HashMap<String, MetadataSet>>>,
+    uri: String,
```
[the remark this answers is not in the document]
evanxg852000: the interface of `uri()` is not async and I needed to cache the URI value otherwise there I would have to deal with `await` during locking the `inner`

quickwit-metastore/src/metastore/single_file_metastore.rs
evanxg852000: So the tokio tasks was not failing at all. My guess was that tokio is kind of executing these tasks on the same thread and in order to get better randomness, I need at least to find a way to force tokio run it on a different threads or at least submit the task from different threads to increase the chance of random behavior. I went with submitting from different threads.
Thought not so helpful, but I also made a tour on this to learn a bit about the scheduler https://tokio.rs/blog/2019-10-scheduler

guilload: I see. Thanks for the explanation. Not that you should do it, but it'd be fun to try again with some tweaks: number of threads in the runtime, number of tasks spawned, adding some sleep here and there.

evanxg852000: @guilload  You were right in asking to try out new things. I found out that tokio during tests uses a `simple-scheduler`. So if we want this kind of behavior, we needed to request it with something like `#[tokio::test(flavor = "multi_thread", worker_threads = 3)]`

guilload (approved):
Nice refactor. This is looking good.

PR #561, opened 2021-09-17
Add http streaming for search stream.

fmassot:
When there is an error for a split result, we stop sending bytes and send an HTTP trailer `X-Stream-Error`.

It works almost, there is still an issue with the trailers that cannot be seen by a `curl` client. I don't know yet where the problem comes from.


r-zenine (commented):
Looks good for now, I'll test it and come back with more feedbacks :)

fmassot:
@r-zenine I updated the branch, partitioned search stream is working with HTTP streaming. I'm still afraid that the trailing errors will not be seen. Can you check that on your side? If this is ok, we can then merge this PR.

r-zenine:
@fmassot I'll setup a testing environment early next week with the bridge to test it :)

r-zenine:
@fmassot I can confirm your observation, I am not seeing the trailers from from curl and I am not seeing them in Go. 
 
What I did:
* I changed [this line](https://github.com/quickwit-inc/quickwit/pull/561/files#diff-b1c9444a7b852a159b49e3067fc34a98a607da53814672d929419e72e6af13a6R258 ) to make the call fail all the time
* I used the command below to test with `curl` and I don't see the trailers.
```bash
curl -isv  --raw --output file.out "http://localhost:8080/api/v1/index/search/stream?query=%22Add%20to%20Cart%22&fastField=session_id&partitionByField=project_id&outputFormat=clickHouseRowBinary"
```
* I did the same making a request using our `Go` service and specifically looking for Trailers and I don't see any.

r-zenine:
I think we can still merge this as it greatly reduce the charge on the search service. 
We could later switch to gprc if the issue is not solved by then in the dependency
Wdyt ?

fmassot:
I prefer to test it before merging :)
So I did it, I tried with a hyper client to make a get request and it's working... only if you use an `http2` client. Normally you will not get trailers with a `http1` client. The real question now is: do you an http2 go client that supports trailers....

fmassot:
oh I just saw your attemps with curl and a go client.

Here is how I tested with a hyper client, I also changed the same line as you to always send the trailers.

```rust
    #[tokio::test]
    async fn test_rest_search_stream_http2_trailers() -> anyhow::Result<()> {
        let client = hyper::Client::builder().http2_only(true).build_http::<hyper::Body>();
        let uri = "http://localhost:8080/api/v1/hdfs_logs/search/stream?outputFormat=csv&fastField=timestamp&query=error".parse()?;
        let mut resp = client.get(uri).await?;
        println!("{:?}", resp);
        println!("Headers: {:#?}\n", resp.headers());
        while let Some(next) = resp.data().await {
            let chunk = next?;
            stdout().write_all(&chunk)?;
        }
        let trailer = resp.trailers().await;
        print!("{:?}", trailer);
        assert!(false);
        Ok(())
    }
```

Which prints:
```
Response { status: 200, version: HTTP/2.0, headers: {"content-type": "text/csv", "date": "Mon, 04 Oct 2021 20:00:40 GMT"}, body: Body(Streaming) }
Headers: {
    "content-type": "text/csv",
    "date": "Mon, 04 Oct 2021 20:00:40 GMT",
}

Ok(Some({"x-stream-error": "search stream error"}))
```

fmassot:
But with `curl` it is definitely not working. To test it, you can try:

```
curl --http2-prior-knowledge --verbose http://localhost:8080/api/v1/hdfs_logs/search/stream\?outputFormat\=csv\&fastField\=timestamp\&query\=error
```

evanxg852000 (approved):
LGTM

Issue #565, opened 2021-09-20
Enable timestamp field on u64 


fmassot:
Currently we only accept `i64`.

Issue #486, opened 2021-08-29
Make the sort field configurable


fulmicoton:
Currently the sorting is not configurable and is defined by default by `SortBy::DocId`.

We should be able to sort on any fast field (constraint coming from `tantivy`) and configure it in the the config file with a new key `sort_by` which defines the `field_name` and the `order`.

Issue #360, opened 2021-08-06
Remove or fix the AllFlattenDocConfig


fulmicoton:
(I don't think it is usable right now.)

As far as I could tell one can only pass it a schema with a single source field.

fmassot:
yes let's remove it

PR #566, opened 2021-09-20
Add sort field configuration and fix config related issues

fmassot:
Fixes:
- [x] #486 
- [x] #565 
- [x] #564
- [x] #360


quickwit-index-config/src/default_index_config/default_config.rs
```diff
@@ -43,20 +43,36 @@ pub struct DefaultIndexConfigBuilder {
...
+}
+
+impl From<SortByConfig> for SortBy {
+    fn from(sort_by_config: SortByConfig) -> Self {
+        SortBy::SortByFastField {
+            field_name: sort_by_config.field_name,
+            order: sort_by_config.order,
+        }
+    }
+}
```
PSeitz, line 66: looks like SortBy::SortByFastField is the same as SortByConfig, so we could reuse it there
```
#[derive(Clone, Debug)]
pub enum SortBy {
    /// Sort by a specific field.
    SortByFastField (SortByFastFieldConfig),
    /// Sort by DocId
    DocId,
}
```

quickwit-indexing/src/actors/indexer.rs
```diff
@@ -93,8 +93,24 @@ enum PrepareDocumentOutcome {
...
     fn create_indexed_split(&self) -> anyhow::Result<IndexedSplit> {
         let schema = self.index_config.schema();
+        let mut index_settings = IndexSettings::default();
+        let sort_by_field = match self.index_config.sort_by() {
+            SortBy::DocId => None,
+            SortBy::SortByFastField { field_name, order } => {
+                let tantivy_order = match order {
+                    SortOrder::Asc => Order::Asc,
+                    SortOrder::Desc => Order::Desc,
+                };
```
PSeitz, line 103: replace with into?

PSeitz (approved):
LGTM except minor comments

quickwit-index-config/src/default_index_config/default_config.rs
fulmicoton: I prefer the way it is today.

quickwit-index-config/src/default_index_config/default_config.rs
```diff
@@ -87,26 +103,50 @@ impl DefaultIndexConfigBuilder {
...
                         timestamp_field_name
                     )
                 }
-            } else {
+            }
+        }
+
+        // Resolve sort field
+        let mut sort_by = None;
+        if let Some(sort_by_config) = self.sort_by {
```
fulmicoton, line 133: can we remove the `mut` ?

Either by using map, or by extracting the conversion in a function.

let sort_by = resolve_sort_field(...);

quickwit-index-config/src/default_index_config/default_config.rs
fmassot: @PSeitz this is the way :D 
I was hesitating and have found that it was slightly hurting readability and I like keeping away all the bloated config structs away from the core as much as possible. I would keep it as it is this for now.

PR #569, opened 2021-09-20
Garbage collect actor

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/457

### How was this PR tested?
- Run indexing, gc, and new command with `-overwrite`

### DIFF
--- a/quickwit-indexing/src/actors/garbage_collector.rs
+++ b/quickwit-indexing/src/actors/garbage_collector.rs
@@ -1,0 +1,336 @@
+// Copyright (C) 2021 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use std::path::Path;
+use std::sync::Arc;
+use std::time::Duration;
+
+use async_trait::async_trait;
+use futures::StreamExt;
+use quickwit_actors::{Actor, ActorContext, AsyncActor};
+use quickwit_metastore::{Metastore, SplitMetadataAndFooterOffsets, SplitState};
+use quickwit_storage::Storage;
+use tantivy::chrono::Utc;
+use tracing::{info, warn};
+
+// TODO: discuss & settle on best default for this
+// https://aws.amazon.com/premiumsupport/knowledge-center/s3-request-limit-avoid-throttling/
+const MAX_CONCURRENT_STORAGE_REQUESTS: usize = 1000;
+
+const RUN_INTERVAL_IN_SECONDS: u64 = 10 * 60; // 10 minutes
+const GRACE_PERIOD_IN_SECONDS: u64 = 2 * 60; // 2 minutes
+
+#[derive(Debug, Clone, Default)]
+pub struct GarbageCollectorCounters {
+    // Denotes the number of passes the garbage collector has performed.
+    pub num_passes: u64,
+}
+
+/// An actor for collecting garbage periodically from an index.
+pub struct GarbageCollector {
+    index_id: String,
+    index_storage: Arc<dyn Storage>,
+    metastore: Arc<dyn Metastore>,
+    counters: GarbageCollectorCounters,
+}
+
+impl GarbageCollector {
+    pub fn new(
+        index_id: String,
+        index_storage: Arc<dyn Storage>,
+        metastore: Arc<dyn Metastore>,
+    ) -> Self {
+        Self {
+            index_id,
+            index_storage,
+            metastore,
+            counters: GarbageCollectorCounters::default(),
+        }
+    }
+
+    pub async fn run_garbage_collect_operation(&mut self) -> anyhow::Result<()> {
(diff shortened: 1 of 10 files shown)


evanxg852000:
@fulmicoton The current code for garbage collection in GC actor has similarities with the `garbage_collect` in `quickwit-core`.
I first attempted to reuse the one from `quickwit-core` but this would create circular dependency, also it has more cli feature like `dry-run` that we don't need in the actor.
Please advise if it's ok as is or should we refactor all this in a common place.

Also, I had a discussion with @PSeitz  last time around this and I think we should settle on a good default value.
https://github.com/quickwit-inc/quickwit/pull/569/files#diff-db9404b546b4450d49144384d7d240612e625b92162aa5a91a2c282637bf00deR32

quickwit-indexing/src/actors/garbage_collector.rs
```diff
@@ -0,0 +1,336 @@
...
+}
+
+/// Delete a list of splits from the storage and the metastore.
+/// It should leave the index and the metastore in good state.
+///
+/// * `index_id` - The target index id.
+/// * `splits`  - The list of splits to delete.
+/// * `metastore` - The metastore managing the target index.
+/// * `storage - The storage managing the target index.
+pub async fn delete_splits_with_files(
```
fulmicoton, line 144: why pub?

quickwit-indexing/src/actors/garbage_collector.rs
```diff
@@ -0,0 +1,336 @@
...
+use quickwit_storage::Storage;
+use tantivy::chrono::Utc;
+use tracing::{info, warn};
+
+// TODO: discuss & settle on best default for this
+// https://aws.amazon.com/premiumsupport/knowledge-center/s3-request-limit-avoid-throttling/
+const MAX_CONCURRENT_STORAGE_REQUESTS: usize = 1000;
+
+const RUN_INTERVAL_IN_SECONDS: u64 = 10 * 60; // 10 minutes
+const GRACE_PERIOD_IN_SECONDS: u64 = 2 * 60; // 2 minutes
```
fulmicoton, line 37: Duration can be used as a const.

quickwit-indexing/src/actors/garbage_collector.rs
```diff
@@ -0,0 +1,336 @@
...
+
+#[derive(Debug, Clone, Default)]
+pub struct GarbageCollectorCounters {
+    // Denotes the number of passes the garbage collector has performed.
+    pub num_passes: u64,
```
fulmicoton, line 42: can we add the number of files that were removed?

quickwit-indexing/src/actors/garbage_collector.rs
```diff
@@ -0,0 +1,336 @@
...
+        }
+    }
+
+    pub async fn run_garbage_collect_operation(&mut self) -> anyhow::Result<()> {
+        info!(index = %self.index_id, "garbage-collect-operation");
+        self.counters.num_passes += 1;
+
+        // Select staged splits with staging timestamp older than grace period timestamp.
+        let grace_period_timestamp = Utc::now().timestamp() - GRACE_PERIOD_IN_SECONDS as i64;
+        let staged_splits = self
```
fulmicoton, line 73: Can you add the type here? Is it : Vec<SplitMetadata>

quickwit-indexing/src/actors/garbage_collector.rs
fulmicoton: (and the number of failures too?)

fulmicoton:
quickwit-indexing is a dep of quickwit-core today. Maybe we can pragmatically move the logic to quickwit-indexing and call it or reexport it in quickwit-core

quickwit-indexing/src/garbage_collection.rs
```diff
@@ -0,0 +1,177 @@
...
+
+use futures::StreamExt;
+use quickwit_metastore::{Metastore, SplitMetadataAndFooterOffsets, SplitState};
+use quickwit_storage::Storage;
+use tantivy::chrono::Utc;
+use tracing::warn;
+
+// TODO: discuss & settle on best default for this
+// https://aws.amazon.com/premiumsupport/knowledge-center/s3-request-limit-avoid-throttling/
+pub const MAX_CONCURRENT_STORAGE_REQUESTS: usize = if cfg!(test) { 2 } else { 1000 };
```
fulmicoton, line 32: it is ok to set it to something lower (100)

fulmicoton (approved):

quickwit-indexing/src/garbage_collection.rs
```diff
@@ -0,0 +1,175 @@
...
+                (
+                    meta.split_metadata.split_id.clone(),
+                    file_entry,
+                    delete_result.is_ok(),
+                )
+            }
+        })
+        .buffer_unordered(MAX_CONCURRENT_STORAGE_REQUESTS);
+
+    while let Some((split_id, file_entry, is_success)) = delete_splits_results_stream.next().await {
```
fmassot, line 155: just a personal nitpick: I would use `.collect()`, log when there is an error and filter out errors and ends with just a vec of `split_id`.

quickwit-indexing/src/actors/garbage_collector.rs
```diff
@@ -0,0 +1,278 @@
...
+
+    async fn process_message(
+        &mut self,
+        _: (),
+        ctx: &ActorContext<Self::Message>,
+    ) -> Result<(), quickwit_actors::ActorExitStatus> {
+        info!(index = %self.index_id, "garbage-collect-operation");
+        self.counters.num_passes += 1;
+
+        let deletion_stats = run_garbage_collect(
```
fmassot, line 95: is `run_garbage_collect` during more than a heartbeat? If yes, I'm worried about this actor becoming unhealthy.

evanxg852000: I don't get this question?

fmassot (approved):
Clean PR, I like that :)

quickwit-indexing/src/actors/garbage_collector.rs
fmassot: (discussed offline)

fulmicoton:
Merging. I created a separate issue to deal with the risk of timeout pointed out by fmassot.

PR #579, opened 2021-09-23
Publish Docker image to Docker Hub on push to main branch

guilload:
### Description
Publish Docker image to Docker Hub at `quickwit/quickwit:main` on push to main branch. 

### How was this PR tested?
- [x] Built Docker image locally.
- [x] Ensured image is correctly published

### DIFF
--- a/.github/workflows/publish_main_docker_image.yaml
+++ b/.github/workflows/publish_main_docker_image.yaml
@@ -1,0 +1,32 @@
+name: Publish Docker image to Docker Hub (main)
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM rust:bullseye AS builder
```
guilload, line 1: FYI, `bullseye` is the new Debian stable. `rust:bullseye` points to the latest version of Rust on `bullseye`.

.github/workflows/publish_main_docker_image.yaml
```diff
@@ -0,0 +1,32 @@
+name: Publish Docker image to Docker Hub (main)
+
+on:
+  push:
+    branches: [main]
```
fulmicoton, line 5: I'd rather have a manual trigger. What do you think?

Dockerfile
fulmicoton: thanks for the info

fulmicoton (approved):

.github/workflows/publish_main_docker_image.yaml
guilload: I like having an automatic build for the `main` branch. That way you know that `quickwit/quickwit:main` always reflects the latest version of `main`. Having a manual trigger for another tag makes sense though. I don't think we'd want our deployments to point to main.

PR #583, opened 2021-09-24
Allow for splits involving less than `merge_factor` splits.

### LINKED ISSUE #575: Avoid creating super large segment
Right now the number of segments involved in the last merge to reach maturity follows the same logic as any other merge.

This is lame, because with a  merge factor of 8, we might end up with segments that are almost 8 times larger than the target segment size.

We should have specific logic to allow for the merge of smaller number of segment for that last merge, and ensure that we create segment that are at < 2x merge_max_doc.

fulmicoton:
The point is to bound the resulting split to
`max_merge_docs* 2` instead of `max_merge_docs * merge_factor`.

Closes #575

### DIFF
--- a/quickwit-indexing/src/merge_policy.rs
+++ b/quickwit-indexing/src/merge_policy.rs
@@ -184,6 +184,10 @@ enum MergeCandidateSize {
     /// The split candidate is good to go.
     ValidSplit,
     /// We should not add an extra split in this candidate.
+    /// This can happen for any of the two following reasons:
+    /// - the number of splits involved already reached `merge_factor_max`.
+    /// - the overall number of docs that will end up in the merged segment already
+    /// exceeds `max_merge_docs`.
     OneMoreSplitWouldBeTooBig,
 }
 
@@ -208,17 +212,13 @@ impl StableMultitenantWithTimestampMergePolicy {
 
         let mut split_levels: Vec<Range<usize>> = Vec::new();
         let mut current_level_start_ord = 0;
-        let mut current_level_max_docs =
-            (splits[0].num_records * 3).clamp(self.min_level_num_docs, self.max_merge_docs);
+        let mut current_level_max_docs = (splits[0].num_records * 3).max(self.min_level_num_docs);
 
         for (split_ord, split) in splits.iter().enumerate() {
             if split.num_records >= current_level_max_docs {
                 split_levels.push(current_level_start_ord..split_ord);
                 current_level_start_ord = split_ord;
-                current_level_max_docs = (3 * split.num_records).min(self.max_merge_docs);
-                if split.num_records >= self.max_merge_docs {
-                    return split_levels;
-                }
+                current_level_max_docs = 3 * split.num_records;
             }
         }
         split_levels.push(current_level_start_ord..splits.len());
--- a/quickwit-indexing/src/merge_policy.rs
+++ b/quickwit-indexing/src/merge_policy.rs
@@ -148,20 +148,14 @@ impl MergePolicy for StableMultitenantWithTimestampMergePolicy {
         // Splits should naturally have an increasing num_merge
         let split_levels = self.build_split_levels(splits);
         for split_range in split_levels.into_iter().rev() {
-            if split_range.len() < self.merge_factor {
-                continue;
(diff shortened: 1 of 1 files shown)


quickwit-indexing/src/merge_policy.rs
```diff
@@ -204,19 +208,74 @@ impl StableMultitenantWithTimestampMergePolicy {
...
 
         for (split_ord, split) in splits.iter().enumerate() {
             if split.num_records >= current_level_max_docs {
                 split_levels.push(current_level_start_ord..split_ord);
                 current_level_start_ord = split_ord;
-                current_level_max_docs = 3 * split.num_records;
+                current_level_max_docs = (3 * split.num_records).min(self.max_merge_docs);
+                if split.num_records >= self.max_merge_docs {
+                    return split_levels;
+                }
```
fmassot, line 221: this early return sounds very strange and the code will never reach this part, right?

quickwit-indexing/src/merge_policy.rs
```diff
@@ -204,19 +208,74 @@ impl StableMultitenantWithTimestampMergePolicy {
...
+            == MergeCandidateSize::TooSmall
+        {
+            return None;
+        }
+        Some(merge_candidate_start..merge_candidate_end)
+    }
+
+    /// Returns true iff we should stop adding extra split into this
+    /// merge candidate.
+    fn merge_candidate_size(&self, splits: &[SplitMetadata]) -> MergeCandidateSize {
```
fmassot, line 254: From what I understand we are checking if we can merge the split candidates. This depends on several conditions: number of splits, number of documents. 
I would rename some things here:


```rust
enum MergeCandidateState {
    NotEnoughSplits,
    Valid,
    TooManySplits, // we can keep too big for both
    TooManyDocuments, // 
}
```

quickwit-indexing/src/merge_policy.rs
```diff
@@ -204,19 +208,74 @@ impl StableMultitenantWithTimestampMergePolicy {
...
+        }
+        if self.merge_candidate_size(&splits[merge_candidate_start..merge_candidate_end])
+            == MergeCandidateSize::TooSmall
+        {
+            return None;
+        }
+        Some(merge_candidate_start..merge_candidate_end)
+    }
+
+    /// Returns true iff we should stop adding extra split into this
```
fmassot, line 252:
```suggestion
    /// Returns `MergeCandidateSize` iff we should stop adding extra split into this
```

quickwit-indexing/src/merge_policy.rs
```diff
@@ -204,19 +208,74 @@ impl StableMultitenantWithTimestampMergePolicy {
...
+            return MergeCandidateSize::TooSmall;
+        }
+
+        // There are already enough splits in this merge.
+        if splits.len() >= self.merge_factor_max {
+            return MergeCandidateSize::OneMoreSplitWouldBeTooBig;
+        }
+        let num_docs_in_merge: usize = splits.iter().map(|split| split.num_records).sum();
+
+        // The resulting split will already exceed `max_merge_docs`.
```
fmassot, line 267:
```suggestion
        // The resulting split will exceed `max_merge_docs`.
```

fmassot (approved):
Sounds good, I would rename some things to make it clearer.

quickwit-indexing/src/merge_policy.rs
fulmicoton: ah you are absolutely right. It is the leftover of a failed refactorign attempt.

quickwit-indexing/src/merge_policy.rs
fulmicoton: From the caller side, the distinction TooManySplits or  TooManyDocuments does not matter though.

I'll add explanations in the rustdoc

fmassot: oh yes, I misunderstood the usage I would have expected a condition like `if splits.len() == self.merge_factor_max` to return `MergeCandidateSize::OneMoreSplitWouldBeTooBig`. 
I guess that the case `splits.len() > self.merge_factor_max` with the current caller implementation.

PR #584, opened 2021-09-24
Fix GC actor integration

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/573
Fixed GC Actor integration into pipeline

### How was this PR tested?
Run indexing

### DIFF
--- a/quickwit-indexing/src/actors/pipeline_supervisor.rs
+++ b/quickwit-indexing/src/actors/pipeline_supervisor.rs
@@ -375,8 +375,8 @@ impl IndexingPipelineSupervisor {
 
                         // When the publisher is dead, try to stop the garbage collector if not
                         // already.
-                        if handlers.publisher.health() != Health::Healthy
-                            && handlers.garbage_collector.health() == Health::Healthy
+                        if handlers.publisher.state() != ActorState::Running
+                            && handlers.garbage_collector.state() == ActorState::Running
                         {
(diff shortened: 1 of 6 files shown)


quickwit-indexing/src/actors/pipeline_supervisor.rs
```diff
@@ -355,6 +360,18 @@ impl IndexingPipelineSupervisor {
                                 .send_exit_with_success(handlers.merge_planner.mailbox())
                                 .await;
                         }
+
+                        // When the publisher is dead, try to stop the garbage collector if not
+                        // already.
+                        if handlers.publisher.health() != Health::Healthy
+                            && handlers.garbage_collector.health() == Health::Healthy
```
fulmicoton, line 367: There was a bug in the indexer logic.

Let me try to describe the problem it is a bit tricky.
`.health()` is actually mutating something. 

It uses a strange contraption in the Progress object that checks if a has_changed is set to true.
At every healthchec this flag is checked, and the actor is in charge of flipping it up until the next iteration of the loop.

Here we call health a second time right after the regular healthcheck call. 95% of the time the actor will not have modified it so that both the publisher and the garbage collector are not healthy.

The fix I added for the MergePlanner is checking for the state of the actors instead.

evanxg852000: Ok

fulmicoton (approved):

PR #585, opened 2021-09-24
Rename `FetchDocsResult` to `FetchDocsResponse`

guilload:
### Description
Rename `FetchDocsResult` to `FetchDocsResponse`.

### How was this PR tested?
Compiled locally.

### DIFF
--- a/quickwit-search/src/fetch_docs.rs
+++ b/quickwit-search/src/fetch_docs.rs
@@ -22,7 +22,7 @@ use std::sync::Arc;
 
 use anyhow::Context;
 use itertools::Itertools;
-use quickwit_proto::{FetchDocsResult, Hit, PartialHit, SplitIdAndFooterOffsets};
+use quickwit_proto::{FetchDocsResponse, Hit, PartialHit, SplitIdAndFooterOffsets};
 use quickwit_storage::Storage;
 use tantivy::{IndexReader, ReloadPolicy};
 use tracing::error;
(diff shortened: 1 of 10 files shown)


quickwit-search/src/fetch_docs.rs
```diff
@@ -149,18 +149,16 @@ async fn fetch_docs_in_split<'a>(
     split: &SplitIdAndFooterOffsets,
 ) -> anyhow::Result<Vec<(GlobalDocAddress<'a>, String)>> {
     let index_reader = get_searcher_for_split(global_doc_addrs.len(), index_storage, split).await?;
-    let mut doc_futures = Vec::new();
-    for global_doc_addr in global_doc_addrs {
+    let doc_futures = global_doc_addrs.into_iter().map(|global_doc_addr| {
```
guilload, line 152: `futures::future::try_join_all` accepts any iterable so often, when it does not hurt readability, it's better to pass an iterator rather than needlessly allocating a new vector.

fmassot: good point. Another remark on the error side, I found the usage `.context("searcher-doc-async")?` or `.with_context(|| "searcher-doc-async")?;` too poor for debugging purpose.

fmassot (approved):

quickwit-search/src/fetch_docs.rs
guilload: Yeah, those are usually not helpful without additional context.

PR #586, opened 2021-09-24
Fix and optimize code coverage CI workflow

guilload:
### Description
Fix and optimize code coverage CI workflow:
- add Postgres service
- add services health checks for faster startup
- avoid compiling cargo-llvm-cov and install from binary
- add artifacts caching

### How was this PR tested?
https://github.com/quickwit-inc/quickwit/runs/3704095533

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -2,34 +2,52 @@ name: Code coverage
 
 on:
   push:
-    branches: [main, run-coverage-workflow]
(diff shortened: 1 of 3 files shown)


fmassot (approved):
LGTM. So you divided by 2 the coverage time?

guilload:
Not really, but at least now it finishes successfully. The last run took `15m 30s`. I hope it'll take less than 10 minutes when we hit the cache.

PR #591, opened 2021-09-27
Define a default set of release-flags.

fulmicoton:
This works by adding a `release-feature-set` feature in quickwit-cli.

- `--all-features` does not work with `cargo install`
- Release binaries github actions now rely on `release-feature-set`
- Docker container also relies on `release-feature-set`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ WORKDIR /quickwit
 
 RUN echo "Building workspace with feature(s) '$CARGO_FEATURES' and profile '$CARGO_PROFILE'" \
     && cargo build \
-        $(echo "--features $CARGO_FEATURES") \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -24,7 +24,7 @@ WORKDIR /quickwit
 
 RUN echo "Building workspace with feature(s) '$CARGO_FEATURES' and profile '$CARGO_PROFILE'" \
     && cargo build \
-        $(test "$CARGO_FEATURES" = "all" && echo "--all-features" || echo "--features $CARGO_FEATURES") \
+        $(echo "--features $CARGO_FEATURES") \
```
guilload, line 27:
```suggestion
        --features $CARGO_FEATURES \
```

guilload (approved):
Thanks!

PR #601, opened 2021-09-29
Empty scratch directory on startup

### LINKED ISSUE #599: Clean up scratch on restart
Indexing requires writing a lot of data in a scratch dir.

We rely on `tempfile` for that. It works as expected, but in case of crash, or a kill, etc. at least on linux, The files are actually not deleted.

We should probably ensure that if a scratch dir is set, we start quickwit by trashing the existing files in it.
If it is tempfile, probably we should not do it... But eventually we should forbid tempfile, it is a terrible default.

guilload:
### Description
Empty scratch directory on startup. Fixes #599.

### How was this PR tested?
Added unit tests.


fulmicoton (changes requested):
That won't work because the quickwit-cache directory used fro the upload cache is a subdir of that dir.

We don't want to remove it. 
One fix could be to add create a dir called scratch on top of the indexing directory.


```
indexing-directory/
    scratch/
    quickwit-cache/
```

We might want to rename quickwit-cache split-cache too.

fulmicoton:
@guilload it was not merge when you fixed the bug :)

PR #602, opened 2021-09-30
Limit Kafka source batch size to 5mb

### LINKED ISSUE #598: Limit batch size in kafka source.
Right now the kafka source builds batches as large as can be consumed in HEARTBEAT * 4 /5

The result is that we send batches of ~500MB, that then take several seconds for the index writer to process.

Due to the messages being atomic, the indexer counters can only be observed outside of the processing of a message.
This means throughput metrics are non-null for long period of times.

It can also mean that we don't an accurate control of the indexer timeout.  

Finally it increases our memory footprint.

I think targeting batches of 10mb should be a decent one size fits all solution.
We could just measure the current batch size and break out of it once the current batch size exceeds 10mb.

fulmicoton:
Closes #598

### DIFF
--- a/quickwit-indexing/src/source/kafka_source.rs
+++ b/quickwit-indexing/src/source/kafka_source.rs
@@ -190,7 +190,7 @@ impl Source for KafkaSource {
         let mut docs = Vec::new();
         let mut checkpoint_delta = CheckpointDelta::default();
 
-        let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT.div_f32(2.0f32));
+        let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT / 2);
 
         let mut message_stream = Box::pin(self.consumer.stream().take_until(deadline));
 
--- a/quickwit-indexing/src/source/kafka_source.rs
+++ b/quickwit-indexing/src/source/kafka_source.rs
@@ -190,7 +190,7 @@ impl Source for KafkaSource {
         let mut docs = Vec::new();
         let mut checkpoint_delta = CheckpointDelta::default();
 
-        let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT * 4 / 5);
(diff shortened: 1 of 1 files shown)


quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -229,6 +244,11 @@ impl Source for KafkaSource {
             checkpoint_delta
                 .record_partition_delta(partition_id, previous_position, current_position)
                 .context("Failed to record partition delta.")?;
+
+            // Above
```
fulmicoton, line 248:
```suggestion
            // Above
```

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -180,8 +191,11 @@ impl Source for KafkaSource {
         let mut checkpoint_delta = CheckpointDelta::default();
 
         let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT * 4 / 5);
```
fulmicoton, line 193:
```suggestion
        let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT / 2);
```

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -179,9 +190,12 @@ impl Source for KafkaSource {
         let mut docs = Vec::new();
         let mut checkpoint_delta = CheckpointDelta::default();
 
-        let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT * 4 / 5);
+        let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT.div_f32(2.0f32));
```
fulmicoton, line 193:
```suggestion
        let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT / 2);
```

guilload (approved):

PR #614, opened 2021-09-30
Implement `get_conn()` method and remove duplicated code

guilload:
### Description
Implement `get_conn()` method and remove duplicated code. Performing this small refactor as I'm taking a look at a bug with the Postgres metastore.

### How was this PR tested?
```sh
cargo build --all-features
```

### DIFF
--- a/quickwit-metastore/src/metastore/postgresql_metastore.rs
+++ b/quickwit-metastore/src/metastore/postgresql_metastore.rs
@@ -119,7 +119,6 @@ fn check_all_splits_were_modified<'a>(
             });
         }
     }
-
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -130,6 +129,8 @@ pub struct PostgresqlMetastore {
     connection_pool: Arc<Pool<ConnectionManager<PgConnection>>>,
 }
 
+type Conn = PooledConnection<ConnectionManager<PgConnection>>;
```
guilload, line 132: This seems useless actually. Happy to remove it.

fulmicoton (approved):

PR #620, opened 2021-10-01
Optimizing indexing split store for write amplification

### LINKED ISSUE #619: Fix Upload cache
The following is an attempt to solve #616.

We should move away from the "transparent" caching approach to this problem and give more control to the client to decide what should be evinced from cache, what should really be cached, etc.

```
struct LocalSplitStore { ... }

impl LocalSplitStore {
   // Called by uploader, only if the split is not mature
   fn put_and_store(&self, split: &Path);
   // Open a split.
   // If the split is not in the store download it. 
   fn open_split(&self, scratch_directory: &Path) -> anyhow::Result<Box<dyn Directory>>();
   // Remove the split. It is called from GC.
   fn delete_split(&self, split: String);
   // Remove the splits that are not within the list.
   fn retain_only(&self, split_ids: Vec<String>) -> anyhow::Result<()> {} 
}
```

Writing into the LocalSplitStore will be done in the Uploader.
The uploader has access to the merge policy in order to decide whether it should be stored or not.

fulmicoton:
Closes #619


quickwit-storage/src/local_file_storage.rs
```diff
@@ -146,6 +154,12 @@ fn missing_file_is_ok(io_result: io::Result<()>) -> io::Result<()> {
     }
 }
 
+impl From<PathBuf> for LocalFileStorage {
```
evanxg852000, line 157: I think sometime back we wanted to only have the LocalFileStorage created with a check on whether the path is rooted or not `Self::extract_root_path_from_uri(uri)?;` to protect against. we could possibly have that check here or make add a comment for people using `LocalFileStorage::from` in another context.

quickwit-storage/src/local_file_storage.rs
```diff
@@ -88,10 +88,18 @@ impl LocalFileStorage {
...
+    /// from here is an external path, and to is an internal path.
+    pub async fn move_into(&self, from_external: &Path, to: &Path) -> crate::StorageResult<()> {
         let to_full_path = self.root.join(to);
-        fs::rename(from_full_path, to_full_path).await?;
+        fs::rename(from_external, to_full_path).await?;
+        Ok(())
+    }
+
+    /// Moves a file from a source to a destination.
+    /// from here is an external path, and to is an internal path.
```
evanxg852000, line 99:
```suggestion
    /// from here is an internal path, and to is an external path.
```

quickwit-storage/src/local_file_storage.rs
```diff
@@ -88,10 +88,18 @@ impl LocalFileStorage {
...
+    pub async fn move_out(&self, from_internal: &Path, to: &Path) -> crate::StorageResult<()> {
```
evanxg852000, line 100:
```suggestion
    pub async fn move_out(&self, from_external &Path, to: &Path) -> crate::StorageResult<()> {
```

evanxg852000, line 100:
```suggestion
    /// from here is an internal path, and to is an external path.
```

quickwit-storage/src/local_file_storage.rs
evanxg852000, line 99:
```suggestion
    /// from here is an internal path, and to is an external path.
```

quickwit-indexing/src/split_store/local_split_store.rs
```diff
@@ -0,0 +1,213 @@
...
+        })
+    }
+
+    #[cfg(test)]
+    pub fn inspect(&self) -> &HashMap<String, usize> {
+        &self.stored_splits
+    }
+
+    pub async fn remove_split(&mut self, split_id: &str) -> StorageResult<()> {
+        if !self.stored_splits.contains_key(split_id) {}
```
evanxg852000, line 102: I think we want an early return here right.
```suggestion
        if !self.stored_splits.contains_key(split_id) { return Ok(()); }
```

quickwit-indexing/src/split_store/indexing_split_store.rs
```diff
@@ -0,0 +1,550 @@
...
+    local_split_store: Option<Arc<Mutex<LocalSplitStore>>>,
+
+    /// The merge policy is useful to identify whether a split
+    /// should be stored in the local storage or not.
+    /// (mature splits do not need to be stored).
+    merge_policy: Arc<dyn MergePolicy>,
+}
+
+impl IndexingSplitStore {
+    /// Create an instance of [`StorageWithUploadCache`]
```
evanxg852000, line 87:
```suggestion
    /// Create an instance of [`IndexingSplitStore`]
```

quickwit-indexing/src/split_store/indexing_split_store.rs
```diff
@@ -0,0 +1,550 @@
...
+    ///
+    /// It needs the remote storage to work with.
+    /// Tries to create a folder `cache_dir/INTERNAL_CACHE_DIR_NAME` if it does not exist.
+    /// List the content of `cache_dir/INTERNAL_CACHE_DIR_NAME` as initial cache content.
+    /// Removes and ignores all files ending with `CACHE_TEMP_FILE_EXTENSION`
```
evanxg852000, line 92:
```suggestion
```

evanxg852000 (commented):
Looks good to me aside from few suggestions :+1:

evanxg852000 (approved):

quickwit-indexing/src/split_store/local_split_store.rs
fulmicoton: nice catch!

quickwit-indexing/src/split_store/local_split_store.rs
```diff
@@ -0,0 +1,213 @@
...
+        let split_path = PathBuf::from(quickwit_common::split_file(split_id));
+        if let Err(error) = self.local_storage.delete(&split_path).await {
+            warn!(file_path = %split_path.display(), error = %error, "Could not remove file from local storage.");
```
fulmicoton, line 105: hmmm we should probably return an error here too.

evanxg852000: I thought we wanted to be less strict on delete. but yes if we miss this delete, the file might never get a delete request anymore and even if we restart it will still be loaded

PR #621, opened 2021-10-01
Avoid cloning values when going from JSON `Value` to `Document`

guilload:
### Description
Modified parsing functions to work on owned values rather than references.

### How was this PR tested?
```sh
make docker-compose-up
QUICKWIT_ENV=local cargo test --all-features
```


fmassot (approved):
nice

PR #622, opened 2021-10-01
Draft indexing server and collect metrics for prometheus.

fmassot:
### Description
My first goal was to have Prometheus indexing metrics and I ended up drafting an indexing server.

The indexing server starts like this:
`cargo r serve-indexing --metastore-uri 's3://my-bucket/quickwit' --source-config-uri 's3://my-bucket/quickwit-sources' --index-id hdfs_logs,wikipedia`

When starting, we first look for source files of names `{index_id}.json` located at the `source-config-uri` and if found, we create the source config and instantiate the pipeline.

Concerning metrics, I used the `IndexingStatistics` and prometheus gauges but I think it would be more appropriate to have counters and directly increment them in actors.

 
To be discussed @fulmicoton @guilload

### DIFF
--- a/quickwit-indexing/src/lib.rs
+++ b/quickwit-indexing/src/lib.rs
@@ -186,6 +186,7 @@ async fn observe_supervisors(index_ids: &[&String], supervisor_handlers: Vec<Act
         if observations.iter().all(|observation| {
             observation.obs_type == ObservationType::PostMortem
         }) {
+            // TODO: funny things. I got post mortem even though one actor continues to publish splits.
             warn!("all supervisors are down, stopping observations.");
             break;
         }
--- a/quickwit-indexing/src/lib.rs
+++ b/quickwit-indexing/src/lib.rs
@@ -17,15 +17,21 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+use std::path::Path;
 use std::sync::Arc;
-
-use anyhow::bail;
-use quickwit_actors::Universe;
-use quickwit_metastore::Metastore;
-use quickwit_storage::StorageUriResolver;
+use std::time::Duration;
+use futures::future::join_all;
+use anyhow::{Context, bail};
+use counters::COUNTERS;
+use itertools::Itertools;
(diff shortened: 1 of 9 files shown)


quickwit-indexing/src/lib.rs
```diff
@@ -71,3 +80,126 @@ pub async fn index_data(
...
+                COUNTERS.num_staged_splits.with_label_values(&[index_id]).set(observation.num_staged_splits as i64);
+                COUNTERS.num_uploaded_splits.with_label_values(&[index_id]).set(observation.num_uploaded_splits as i64);
+                COUNTERS.total_bytes_processed.with_label_values(&[index_id]).set(observation.total_bytes_processed as i64);
+                COUNTERS.total_size_splits.with_label_values(&[index_id]).set(observation.total_size_splits as i64);
+            }   
+        }
+        if observations.iter().all(|observation| {
+            observation.obs_type == ObservationType::PostMortem
+        }) {
+            // TODO: funny things. I got post mortem even though one actor continues to publish splits.
```
fulmicoton, line 189: When you see unexpected stuff, can you filll a bug, with logs etc. This is precious information!

quickwit-indexing/src/lib.rs
```diff
@@ -71,3 +80,126 @@ pub async fn index_data(
...
+                )
+            })?;
+        debug!(index_id=?index_id, "Source config built.");
+        source_configs.push((index_id, source_config));
+    }
+    Ok(source_configs)
+}
+
+/// Start Quickwit indexing server.
+pub async fn serve_indexing_cli(args: ServeIndexingArgs) -> anyhow::Result<()> {
```
fulmicoton, line 111: Do you think it might b possible to move this to quickwit-cli?

I am mostly  uncomfortable to have the warp dependency on quickwit-indexing.

quickwit-indexing/src/lib.rs
```diff
@@ -71,3 +80,126 @@ pub async fn index_data(
 pub(crate) fn new_split_id() -> String {
     ulid::Ulid::new().to_string()
 }
+
+/// Fetch source configs from storage for each index id.
+/// The source path must have the form `{index_id}.json`.
```
fulmicoton, line 85: I think we can consider the config stuff as high priority now.

PR #626, opened 2021-10-04
add extract-split closes #625

### LINKED ISSUE #625: Extract Split Method
Add CLI method to extract split files

PSeitz:
### Description
add extract-split method to cli. closes #625

### How was this PR tested?
locally

### DIFF
--- a/quickwit-storage/src/bundle_storage.rs
+++ b/quickwit-storage/src/bundle_storage.rs
@@ -66,6 +66,11 @@ impl BundleStorage {
             metadata,
         })
     }
+
+    /// Returns Iterator over files contained in the bundle.
+    pub fn iter_files(&self) -> impl Iterator<Item = &PathBuf> {
+        self.metadata.files.keys()
+    }
 }
 
 #[derive(Debug, Error)]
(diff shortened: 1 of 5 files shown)


fulmicoton:
A better commit comment would be:
```
Add CLI command to extract split files.

Closes #625
```

quickwit-storage/src/bundle_storage.rs
```diff
@@ -132,8 +138,11 @@ impl Storage for BundleStorage {
         Err(unsupported_operation(path))
     }
 
-    async fn copy_to_file(&self, path: &Path, _output_path: &Path) -> crate::StorageResult<()> {
-        Err(unsupported_operation(path))
+    async fn copy_to_file(&self, path: &Path, output_path: &Path) -> crate::StorageResult<()> {
+        let file_content = self.get_all(path).await?;
```
fulmicoton, line 142: This loads the entire file in anonymous memory. 
That's often 10+ GB. 

Can you fix that (Note it is non trivial with the current storage API) or at least document that?

PSeitz: I changed it load chunks of 100MB. For a better solution with the storage API, we would need something like `append_from_slice`

fulmicoton (approved):

PR #628, opened 2021-10-04
Delay deletion of splits marked for deletion

### LINKED ISSUE #605: GC Logic is not good. (Incorrect and suboptimal)
Right now GC delete stage split that are older than the grace period.
GC deletes splits that marked for deletion right away.

This is incorrect for the following reason.
Merge ends in marking for deletes the replaced split ids.

A currently flying query might have already decided to use these splits.
We need a short grace period to deal with this problem (2mn is ok).

Ideally the deletion of staged split on the other hand should happen either by marking them for delete on restart
of the indexing pipeline ( start of the GC actor for instance).
If it is difficult or I am missing some bad side effect, we can also use a long grace period (it needs to take in account the upload time).

PSeitz:
Delay deletion of splits by 2 minutes.

fixes #605

### Description

### How was this PR tested?

### DIFF
--- a/quickwit-indexing/src/actors/garbage_collector.rs
+++ b/quickwit-indexing/src/actors/garbage_collector.rs
@@ -30,7 +30,16 @@ use tracing::info;
 use crate::run_garbage_collect;
 
 const RUN_INTERVAL: Duration = Duration::from_secs(60); // 1 minutes
+/// Staged files needs to be deleted if there was a failure.
+/// TODO ideally we want clean up all staged splits every time we restart the indexing pipeline, but the grace
+/// period strategy should do the job for the moment.
 const STAGED_GRACE_PERIOD: Duration = Duration::from_secs(60 * 60); // 1 hour
+/// We cannot safely delete splits right away as a in-flight queries could actually
+/// have selected this split.
+/// We deal this probably by introducing a grace period. A split is first marked as delete,
+/// and hence won't be selected for search. After a few minutes, once it reasonably safe to assume
+/// that all queries involving this split have terminated, we effectively delete the split.
+/// This duration is controlled by `DELETION_GRACE_PERIOD`.
 const DELETION_GRACE_PERIOD: Duration = Duration::from_secs(120); // 2 min
 
 #[derive(Debug, Clone, Default)]
--- a/quickwit-indexing/src/actors/garbage_collector.rs
+++ b/quickwit-indexing/src/actors/garbage_collector.rs
@@ -30,7 +30,8 @@ use tracing::info;
 use crate::run_garbage_collect;
 
 const RUN_INTERVAL: Duration = Duration::from_secs(60); // 1 minutes
-const GRACE_PERIOD: Duration = Duration::from_secs(60 * 60); // 1 hour
+const STAGED_GRACE_PERIOD: Duration = Duration::from_secs(60 * 60); // 1 hour
+const DELETION_GRACE_PERIOD: Duration = Duration::from_secs(120); // 2 min
 
 #[derive(Debug, Clone, Default)]
 pub struct GarbageCollectorCounters {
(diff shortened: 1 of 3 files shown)


quickwit-indexing/src/garbage_collection.rs
```diff
@@ -61,17 +61,22 @@ impl From<&SplitMetadataAndFooterOffsets> for FileEntry {
...
     storage: Arc<dyn Storage>,
     metastore: Arc<dyn Metastore>,
-    grace_period: Duration,
+    staged_grace_period: Duration,
+    deletion_grace_period: Duration,
     dry_run: bool,
 ) -> anyhow::Result<SplitDeletionStats> {
     // Select staged splits with staging timestamp older than grace period timestamp.
-    let grace_period_timestamp = Utc::now().timestamp() - grace_period.as_secs() as i64;
+    let grace_period_timestamp = Utc::now().timestamp() - staged_grace_period.as_secs() as i64;
```
fulmicoton, line 78: Utc have arithmetic.

quickwit-indexing/src/actors/garbage_collector.rs
```diff
@@ -30,7 +30,8 @@ use tracing::info;
 use crate::run_garbage_collect;
 
 const RUN_INTERVAL: Duration = Duration::from_secs(60); // 1 minutes
-const GRACE_PERIOD: Duration = Duration::from_secs(60 * 60); // 1 hour
+const STAGED_GRACE_PERIOD: Duration = Duration::from_secs(60 * 60); // 1 hour
+const DELETION_GRACE_PERIOD: Duration = Duration::from_secs(120); // 2 min
```
fulmicoton, line 34:
```suggestion
/// We cannot safely delete splits right away as a in-flight queries could actually
/// have selected this split.
/// We deal this probably by introducing a grace period. A split is first marked as delete,
/// and hence won't be selected for search. After a few minutes, once it reasonably safe to assume
/// that all queries involving this split have terminated, we effectively delete the split.
/// This duration is controlled by `DELETION_GRACE_PERIOD`.
const DELETION_GRACE_PERIOD: Duration = Duration::from_secs(120); // 2 min
```

quickwit-indexing/src/actors/garbage_collector.rs
```diff
@@ -30,7 +30,8 @@ use tracing::info;
 use crate::run_garbage_collect;
 
 const RUN_INTERVAL: Duration = Duration::from_secs(60); // 1 minutes
-const GRACE_PERIOD: Duration = Duration::from_secs(60 * 60); // 1 hour
+const STAGED_GRACE_PERIOD: Duration = Duration::from_secs(60 * 60); // 1 hour
```
fulmicoton, line 33:
```suggestion
/// Staged files needs to be deleted if there was a failure.
/// TODO ideally we want clean up all staged splits every time we restart the indexing pipeline, but the grace
/// period strategy should do the job for the moment.
const STAGED_GRACE_PERIOD: Duration = Duration::from_secs(60 * 60); // 1 hour
```

quickwit-indexing/src/garbage_collection.rs
```diff
@@ -105,10 +110,16 @@ pub async fn run_garbage_collect(
         .mark_splits_for_deletion(index_id, &split_ids)
         .await?;
 
-    // Select split to delete
+    // We wait another 2 minutes until the split is actually deleted.
+    let grace_period_deletion = Utc::now().timestamp() - deletion_grace_period.as_secs() as i64;
```
fulmicoton, line 114: same as above.

fulmicoton (changes requested):

fulmicoton (changes requested):
See comments

fulmicoton (approved):

Issue #520, opened 2021-09-08
Make Storage Cache Path configurable 


evanxg852000:
The storage cache used in indexing and search is currently using a temporary directory from `tempfile` crate. This makes it useless when the indexing/search server is restarted.
This issue should discuss the way the cache path should be configured and shared among service restarts.

PR #644, opened 2021-10-07
Create indexing directories in data dir

guilload:
### Description
For each index, create an indexing directory in the data dir at `<data dir>/<index name>`. The data dir path is specified by the user or a configuration setting.

An indexing directory consists of two subdirectories:
- a cache directory
- a scratch directory, emptied upon restart

### How was this PR tested?
- added unit tests
- ran unit and integration test suites


guilload:
Fixes #520.
Fixes #599.

PR #671, opened 2021-10-19
Added migration files for schema changes in #658.

fulmicoton:
### How was this PR tested
Using diesel on a local postgres. Revert has been tested too.
On an actual old indexed index with many splits

### DIFF
--- a/quickwit-metastore/src/metastore/mod.rs
+++ b/quickwit-metastore/src/metastore/mod.rs
@@ -93,6 +93,7 @@ pub struct SplitMetadata {
     pub tags: HashSet<String>,
 
     /// Number of demux operations this split has undergone.
+    #[serde(default)]
     pub demux_num_ops: usize,
 }
 


PSeitz:
@fulmicoton 
As per docs, they run in a transaction:
https://docs.rs/diesel_migrations/1.4.0/diesel_migrations/fn.run_pending_migrations.html

Looking at the code, there is a conditional, which seems to return always true currently
https://github.com/diesel-rs/diesel/blob/9afd87b51611aeb7eb485a71aef782accc593925/diesel_migrations/src/migration_harness.rs#L187-L191

There are some limits on the transaction guarantees on some DDL statements, but DDL transaction support looks pretty good on PostgreSQL (https://www.postgresql.org/docs/14/mvcc-caveats.html)

"Some DDL commands, currently only TRUNCATE and the table-rewriting forms of ALTER TABLE, are not MVCC-safe. This means that after the truncation or rewrite commits, the table will appear empty to concurrent transactions, if they are using a snapshot taken before the DDL command committed."

PR #675, opened 2021-10-19
Fix release builds

evanxg852000:
### Description
Fixed some release builds
- x86_64-apple-darwin
- x86_64-unknown-linux-gnu
- x86_64-unknown-linux-musl (statically linked)
- aarch64-unknown-linux-gnu

Added documentation

TODO: (In another task https://github.com/quickwit-inc/quickwit/issues/715)


fulmicoton:
@evanxg852000 I think you nailed the problem. We could also use this https://docs.rs/openssl-probe/0.1.4/openssl_probe/
to sniff the right path for OpenSSL cert.

fulmicoton (approved):

cross-images/libpq.sh
```diff
@@ -0,0 +1,47 @@
...
+        if ! dpkg -L $dep; then
+            apt-get install --no-install-recommends -y $dep
+            purge_list+=( $dep )
+        fi
+    done
+
+    td=$(mktemp -d)
+
+    pushd $td
+    curl -kfLO https://ftp.postgresql.org/pub/source/v$version/postgresql-$version.tar.gz
```
fulmicoton, line 28: Adding some comment about what all this is about would be awesome.

fulmicoton (approved):

cross-images/libpq.sh
evanxg852000: I have added comments. Also, it will need refactoring base on how this goes https://github.com/quickwit-inc/quickwit/issues/715

.github/workflows/release.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
```
guilload, line 4: Have you looked into simplifying this workflow by creating a generic workflow that builds a given target and then calling the generic workflow with a matrix of targets? Are there some subtleties between targets that prevent doing so?

Makefile
```diff
@@ -31,3 +31,14 @@ fmt:
...
 	QUICKWIT_ENV=local cargo test --all-features
+
+# This will build and push all custom cross images for cross-compilation.
+# You will need to login into dockerhub with `quickwitinc` account
+IMAGE_TAGS = x86_64-unknown-linux-gnu aarch64-unknown-linux-gnu x86_64-unknown-linux-musl aarch64-unknown-linux-musl 
+
+.PHONY: cross-images
+cross-images:
+	for tag in ${IMAGE_TAGS}; do \
+		docker build --tag quickwitinc/builder-$$tag --file ./cross-images/$$tag.dockerfile ./cross-images; \
```
guilload, line 42: Who owns this DockerHub account? You?

cross-images/aarch64-unknown-linux-gnu.dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM rustembedded/cross:aarch64-unknown-linux-gnu
```
guilload, line 1: Can we move the cross images to `build/cross-images`?

cross-images/aarch64-unknown-linux-gnu.dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+COPY libpq.sh /
+RUN bash /libpq.sh aarch64-linux-gnu
+
+RUN apt-get update -y && \
+    apt-get install -y libpq-dev
```
guilload, line 7:
```suggestion
    apt-get install -y libpq-dev && \
    rm -rf /var/lib/apt/lists/*
```

cross-images/aarch64-unknown-linux-gnu.dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+ENV X86_64_UNKNOWN_LINUX_MUSL_OPENSSL_DIR=/usr/local/musl/ \
```
guilload, line 9: nit: can we sort those alphabetically?

cross-images/aarch64-unknown-linux-gnu.dockerfile
guilload: Note that this is actually not necessary here given the purpose of this image, but it's just a good habit.

guilload (approved):
LGTM.

quickwit-metastore/Cargo.toml
```diff
@@ -43,4 +44,4 @@ tempfile = '3'
 [features]
 testsuite = ["mockall", "tempfile"]
 ci-test = []
-postgres = [ "diesel", "diesel_migrations" ]
+postgres = [ "diesel", "diesel_migrations", "openssl"]
```
guilload, line 47: Why don't use we `openssl/vendored` here?

quickwit-indexing/Cargo.toml
```diff
@@ -42,6 +44,7 @@ arc-swap = "1.4"
 [features]
 kafka = ["rdkafka"]
 kafka-external-service = []
+kafka-vendored = ["kafka", "libz-sys/static", "openssl/vendored"]
```
guilload, line 47: vendored-kafka

quickwit-cli/Cargo.toml
```diff
@@ -53,4 +54,6 @@ serial_test = "0.5.1"
 
 [features]
 ci-test = []
-release-feature-set = ["quickwit-metastore/postgres", "quickwit-indexing/kafka"]
+openssl-support = ["openssl-probe"]
```
guilload, line 57: openssl

Makefile
evanxg852000: We own it this is where our docker images are pushed

quickwit-cli/Cargo.toml
evanxg852000: The name of the feature will become the name of a dependency. `openssl`

.github/workflows/release.yml
evanxg852000: I tried this but GitHub action does not give much reusability unless we go with a full bash script that we can invoke. You will not lot of `if` at the en of each pipeline. I tried factoring this into reusable action without much success. composite actions don't get all the context of the top-level action.

quickwit-metastore/Cargo.toml
evanxg852000: This helped more like selecting the ssl version to be same with the one on `quickwit/indexing`, I did not try `openssl/vendored` tbh. I think this might even be a hint to make build for `arm-musl` Will test it on https://github.com/quickwit-inc/quickwit/issues/715

Makefile
guilload: There's also:
https://hub.docker.com/orgs/quickwit

We can figure that out later but we need to consolidate around one account.

.github/workflows/release.yml
guilload: Cool, just wanted to know if we tried and what the outcome was for developing future workflows. Thanks.

quickwit-cli/Cargo.toml
guilload: Ok. Makes sense. This is why some projects just call the feature `ssl` then.

Makefile
evanxg852000: I can definetly route all our images to this account if this is the official one. before merging this.
We just need someone with access to change the dockerhub access token on the github repository.

PR #682, opened 2021-10-20
Implement basic API for interacting with Kinesis data streams

guilload:
### Description
Implement basic API for interacting with Kinesis data streams. My Kinesis source branch was getting quite large so I decided to split my work into multiple PRs. Here is the first one.

### How was this PR tested?
Added tests and ran `make test-all`.

### DIFF
--- a/quickwit-indexing/src/source/kinesis/api.rs
+++ b/quickwit-indexing/src/source/kinesis/api.rs
@@ -1,0 +1,454 @@
+// Copyright (C) 2021 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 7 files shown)


quickwit-indexing/src/source/kinesis/api.rs
```diff
@@ -0,0 +1,449 @@
...
+            records: put_records_request_entries,
+        };
+        let response = kinesis_client.put_records(request).await?;
+
+        let mut sequence_numbers = HashMap::new();
+        for record in response.records {
+            sequence_numbers
+                .entry(record.shard_id.and_then(parse_shard_id).unwrap())
+                .or_insert_with(Vec::new)
+                .push(record.sequence_number.unwrap());
```
fulmicoton, line 282:
```suggestion
            if record.sequence_number.is_none() {
                error!("Missing sequence number for a kinesis record. Skipping the record");
                continue;
            }
            sequence_numbers
                .entry(record.shard_id.and_then(parse_shard_id).unwrap())
                .or_insert_with(Vec::new)
                .extend(record.sequence_number.unwrap());
```

fulmicoton, line 282: I don't think we should ever unwrap things that are coming from a third-party.
In an ideal world, rusoto should take care of returning an error if the field cannot be deserialized but they don't because it is generated from metadata that does not contain this information.

In this non ideal world, unwrapping here is as dangerous as unwrapping during deserialization in the rusoto library.

fulmicoton (approved):
Great job. I still have one simple comment but approves the PR to make the process as light as possible.

quickwit-indexing/src/source/kinesis/api.rs
guilload: I agree with you. I'm unwrapping here because this function is only used for unit testing.

PR #709, opened 2021-10-28
Add stats subcommands with 'demux' + use colored crate for printing colors in terminal 

fmassot:
### Description
Add subcommands `stats` and sub subcommand `demux`. I think we will need some very simple statistics on split size or something like that, hence the creation of the `stats` subcommand.

```bash
quickwit stats demux --index-id hdfs_logs
```

Note: I did not find a maintained simple crate for descriptive statistics, if you have any names, I would welcome it.

### How was this PR tested?
Locally only. No unit test added. Waiting first for feedback before adding some simple unit tests.

### Screenshot of output
Capture d’écran 2021-10-29 à 00 30 36


fmassot:
I decided to use the `colored` crate everywhere as:
- the API is really nice for setting colors
- by default, it detects automatically if the output is a terminal. The user can also disable color by setting the env variable `NO_COLOR`. 
- the crate follow nice standards: https://no-color.org/ and https://bixense.com/clicolors/

fmassot:
@evanxg852000 can you make the last check on this small PR? thanks!

PR #711, opened 2021-11-01
Added a controllable directory.

### LINKED ISSUE #704: Use the directory abstraction to record progress and respond to kill switch
A lot of the tantivy code is currently protected by a protect guard to pass health check.
This is suboptimal as it does not make it possible to detect a dead lock that might happen in tantivy.

In addition, if the killswitch is activated, we would rather be able to stop as soon as possible these blocking functions.

To do this, we cannot stop the threads and we'd rather avoid leaking our need into tantivy.
The idea would be to wrap our Directory in order to record progress everytime we write on disk.

In addition, we could check and return an error if the kill switch has been activated.

fulmicoton:
The directory records progress and checks for the kill switch on every
writes.

Closes #704. See details in the issue.

### Description

### How was this PR tested?

### DIFF
--- a/quickwit-actors/src/progress.rs
+++ b/quickwit-actors/src/progress.rs
@@ -85,7 +85,7 @@ impl Progress {
 
     pub fn protect_zone(&self) -> ProtectedZoneGuard {
         loop {
-            let previous_state: ProgressState = self.0.load(Ordering::SeqCst).into();
+            let previous_state: ProgressState = self.0.load(Ordering::Relaxed).into();
             let new_state = match previous_state {
                 ProgressState::NoUpdate | ProgressState::Updated => ProgressState::ProtectedZone(0),
                 ProgressState::ProtectedZone(level) => ProgressState::ProtectedZone(level + 1),
@@ -95,8 +95,8 @@ impl Progress {
                 .compare_exchange(
                     previous_state.into(),
                     new_state.into(),
-                    Ordering::SeqCst,
-                    Ordering::SeqCst,
+                    Ordering::Relaxed,
+                    Ordering::Relaxed,
                 )
                 .is_ok()
             {
@@ -129,7 +129,7 @@ pub struct ProtectedZoneGuard(Arc<AtomicU32>);
 
 impl Drop for ProtectedZoneGuard {
     fn drop(&mut self) {
-        let previous_state: ProgressState = self.0.fetch_sub(1, Ordering::SeqCst).into();
+        let previous_state: ProgressState = self.0.fetch_sub(1, Ordering::Relaxed).into();
         assert!(matches!(previous_state, ProgressState::ProtectedZone(_)));
     }
 }
--- a/quickwit-indexing/failpoints/mod.rs
+++ b/quickwit-indexing/failpoints/mod.rs
@@ -35,20 +35,27 @@
 //!
 //! Below we test panics at different steps in the indexing pipeline.
 
+use std::path::Path;
 use std::sync::{Arc, Mutex};
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use byte_unit::Byte;
 use fail::FailScenario;
-use quickwit_index_config::default_config_for_tests;
-use quickwit_indexing::actors::IndexerParams;
-use quickwit_indexing::index_data;
-use quickwit_indexing::models::{CommitPolicy, IndexingDirectory};
+use quickwit_actors::{create_test_mailbox, ActorExitStatus, Universe};
+use quickwit_common::split_file;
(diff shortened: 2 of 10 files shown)


fulmicoton:
I just tested the PR and I get packager timeouts.

quickwit-actors/src/progress.rs
```diff
@@ -85,7 +85,7 @@ impl Progress {
 
     pub fn protect_zone(&self) -> ProtectedZoneGuard {
         loop {
-            let previous_state: ProgressState = self.0.load(Ordering::SeqCst).into();
+            let previous_state: ProgressState = self.0.load(Ordering::Relaxed).into();
```
fmassot, line 88: Thinking about ordering is always a brainfuck. But in this simple case, I agree, I think we don't really care so `Relaxed` sounds good.

quickwit-indexing/failpoints/mod.rs
```diff
@@ -210,3 +217,119 @@ async fn aux_test_failpoints() -> anyhow::Result<()> {
     );
     Ok(())
 }
+
+const TEST_TEXT: &'static str = r#"His sole child, my lord, and bequeathed to my
```
fmassot, line 221: Great text.

quickwit-indexing/src/actors/merge_executor.rs
```diff
@@ -354,27 +377,22 @@ impl MergeExecutor {
             replaced_segments_demux_field_readers,
             demuxed_virtual_splits,
         );
-        info!("demux-build-mapping-end");
```
fmassot, line 357: you can remove this one yes, can you also remove `info!("demux-build-mapping-start")` to be consistent, I added these when I was trying to understand which demux part was taking time.

quickwit-indexing/src/controllable_directory.rs
```diff
@@ -0,0 +1,264 @@
...
+/// with functionalities such as
+/// - records progress everytime a write (Note there is however a buffer writer above it)
+/// - if the killswitch is activated, returns an error on the first write happening after it
+/// - in the future, record a writing speed, possibly introduce some throttling, etc.
+#[derive(Clone)]
+pub struct ControllableDirectory {
+    inner: Arc<Inner>,
+}
+
+impl ControllableDirectory {
```
fmassot, line 47: `ControlledDirectory` or `ControllableDirectory`?
`ControlledWrite` or `ControllableWrite` ? 
I would choose `controlled`. 
(not really important, you can ignore)

quickwit-indexing/src/controllable_directory.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    fn watch(&self, watch_callback: WatchCallback) -> tantivy::Result<WatchHandle> {
+        self.inner.controls.check_if_alive()?;
+        self.inner.underlying.watch(watch_callback)
+    }
+}
+
+impl TerminatingWrite for ControlledWrite {
+    #[inline]
+    fn terminate_ref(&mut self, token: AntiCallToken) -> io::Result<()> {
```
fmassot, line 202: didn't know about this trick  `AntiCallToken` in tantivy, interesting.

quickwit-indexing/src/controllable_directory.rs
```diff
@@ -0,0 +1,264 @@
...
+use tantivy::common::{AntiCallToken, TerminatingWrite};
+use tantivy::directory::error::{DeleteError, OpenReadError, OpenWriteError};
+use tantivy::directory::{FileHandle, WatchCallback, WatchHandle, WritePtr};
+use tantivy::Directory;
+
+/// Buffer capacity.
+///
+/// This is the current default for the BufWriter, but considering this constant
+/// will have a direct impact on health check, we'd better fix it.
+const BUFFER_NUM_BYTES: usize = 8_192;
```
fmassot, line 35: `8_192` sounds like a good starting value, do you want to update the buffer size regarding the number of bytes to write?

fmassot (approved):
Really nice piece of code and simple.

quickwit-indexing/src/controllable_directory.rs
fulmicoton: No. 8KB or 16KB is considered optimal. 8K is the default in rust btw

quickwit-indexing/src/controllable_directory.rs
fulmicoton: Went for Controlled!

quickwit-indexing/src/actors/merge_executor.rs
fulmicoton: I did not remove it on purpose actually. Thanks for spotting!

PR #714, opened 2021-11-01
stream split bundle on the fly

PSeitz:
### Description
Don't materialize the single split file in packager.rs, instead stream the compute bundle from the files in uploader.rs. This will save IOPS.

### How was this PR tested?
cargo test

### DIFF
--- a/quickwit-cli/src/lib.rs
+++ b/quickwit-cli/src/lib.rs
@@ -211,7 +211,11 @@ pub async fn extract_split_cli(args: ExtractSplitArgs) -> anyhow::Result<()> {
     let split_file = PathBuf::from(format!("{}.split", args.split_id));
     let (_, bundle_footer) = read_split_footer(index_storage.clone(), &split_file).await?;
 
-    let bundle_storage = BundleStorage::new(index_storage, split_file, &bundle_footer)?;
+    let (_hotcache_bytes, bundle_storage) = BundleStorage::open_from_split_data_with_owned_bytes(
+        index_storage,
+        split_file,
+        bundle_footer,
+    )?;
 
     std::fs::create_dir_all(args.target_folder.to_owned())?;
 
--- a/quickwit-common/src/lib.rs
+++ b/quickwit-common/src/lib.rs
@@ -22,6 +22,7 @@ pub mod metrics;
 
 use std::fmt::Debug;
 use std::ops::Range;
+use std::path::Path;
 use std::str::FromStr;
 
 pub use coolid::new_coolid;
@@ -29,6 +30,10 @@ use tracing::{error, info};
 
 /// Filenames used for hotcache files.
 pub const HOTCACHE_FILENAME: &str = "hotcache";
+
+pub fn get_hotcache_path() -> &'static Path {
+    Path::new(HOTCACHE_FILENAME)
+}
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum QuickwitEnv {
--- a/quickwit-common/src/lib.rs
+++ b/quickwit-common/src/lib.rs
@@ -22,18 +22,10 @@ pub mod metrics;
 
 use std::fmt::Debug;
 use std::ops::Range;
-use std::path::Path;
 use std::str::FromStr;
 
 pub use coolid::new_coolid;
 use tracing::{error, info};
-
-/// Filenames used for hotcache files.
-pub const HOTCACHE_FILENAME: &str = "hotcache";
-
-pub fn get_hotcache_path() -> &'static Path {
-    Path::new(HOTCACHE_FILENAME)
-}
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum QuickwitEnv {
--- a/quickwit-directories/src/bundle_directory.rs
+++ b/quickwit-directories/src/bundle_directory.rs
@@ -90,11 +90,10 @@ pub fn get_hotcache_from_split(data: OwnedBytes) -> io::Result<OwnedBytes> {
 
 impl BundleDirectory {
     /// Get files and their sizes in a split.
-    pub fn get_stats_split(data: OwnedBytes) -> io::Result<Vec<(PathBuf, usize)>> {
+    pub fn get_stats_split(data: OwnedBytes) -> io::Result<Vec<(PathBuf, u64)>> {
         let split_file = FileSlice::new(Box::new(data));
         let (body_and_bundle_metadata, hot_cache) = split_footer(split_file)?;
-        let file_offsets =
-            BundleStorageFileOffsets::open_from_file_slice(body_and_bundle_metadata)?;
+        let file_offsets = BundleStorageFileOffsets::open(body_and_bundle_metadata)?;
 
         let mut files_and_size: Vec<(_, _)> = file_offsets
             .files
@@ -102,7 +101,10 @@ impl BundleDirectory {
             .map(|(file, range)| (file, range.end - range.start))
             .collect();
 
-        files_and_size.push((PathBuf::from("hotcache".to_string()), hot_cache.len()));
+        files_and_size.push((
+            PathBuf::from("hotcache".to_string()),
+            hot_cache.len() as u64,
+        ));
 
         files_and_size.sort();
         Ok(files_and_size)
@@ -117,7 +119,7 @@ impl BundleDirectory {
 
     /// Opens a BundleDirectory, given a file containing the bundle data.
     pub fn open_bundle(file: FileSlice) -> io::Result<BundleDirectory> {
-        let file_offsets = BundleStorageFileOffsets::open_from_file_slice(file.clone())?;
+        let file_offsets = BundleStorageFileOffsets::open(file.clone())?;
         Ok(BundleDirectory { file, file_offsets })
     }
 }
@@ -133,7 +135,9 @@ impl Directory for BundleDirectory {
             .file_offsets
             .get(path)
             .ok_or_else(|| OpenReadError::FileDoesNotExist(path.to_path_buf()))?;
-        Ok(self.file.slice(byte_range))
+        Ok(self
+            .file
+            .slice(byte_range.start as usize..byte_range.end as usize))
     }
 
     fn atomic_read(&self, path: &Path) -> Result<Vec<u8>, OpenReadError> {
@@ -173,26 +177,18 @@ impl Directory for BundleDirectory {
 
 #[cfg(test)]
 mod tests {
-    use std::fs::{self, File};
+    use std::fs::File;
     use std::io::Write;
 
-    use quickwit_storage::BundleStorageBuilder;
-    use tantivy::common::CountingWriter;
+    use quickwit_storage::{get_split_payload_streamer, PutPayload};
 
     use super::*;
 
-    #[test]
-    fn test_bundle_directory_stats() -> anyhow::Result<()> {
+    #[tokio::test]
+    async fn test_bundle_directory_stats() -> anyhow::Result<()> {
         let temp_dir = tempfile::tempdir()?;
         let test_filepath1 = temp_dir.path().join("f1");
         let test_filepath2 = temp_dir.path().join("f2");
-        let test_bundle_path = temp_dir.path().join("le_bundle");
-
-        // Simulating split file which is created in packager.rs, which contains the bundle and
-        // appends the footer. The bundle directory can read the format created by
-        // packager.rs
-        let mut split_file = CountingWriter::wrap(File::create(test_bundle_path.to_owned())?);
-        let mut bundle_builder = BundleStorageBuilder::new(&mut split_file)?;
 
         let mut file1 = File::create(&test_filepath1)?;
         file1.write_all(&[123, 76])?;
@@ -200,42 +196,30 @@ mod tests {
         let mut file2 = File::create(&test_filepath2)?;
         file2.write_all(&[99, 55, 44])?;
 
-        bundle_builder.add_file(&test_filepath1)?;
-        bundle_builder.add_file(&test_filepath2)?;
-        bundle_builder.finalize()?;
+        let split_streamer = get_split_payload_streamer(
+            &[test_filepath1.clone(), test_filepath2.clone()],
+            &[
+                1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18,
+            ],
+        )?;
 
-        // append hotcache and hotcache len
-        let hotcache_offset_start = split_file.written_bytes();
-        split_file.write_all(&[
-            1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18,
-        ])?;
-
-        let hotcache_offset_end = split_file.written_bytes();
-        let hotcache_num_bytes = hotcache_offset_end - hotcache_offset_start;
-
-        split_file.write_all(&hotcache_num_bytes.to_le_bytes())?;
-
-        let buffer = fs::read(test_bundle_path)?;
+        let buffer = split_streamer.read_all().await?;
 
         // check stats
-        let stats = BundleDirectory::get_stats_split(OwnedBytes::new(buffer))?;
+        let stats = BundleDirectory::get_stats_split(buffer)?;
 
-        assert_eq!(stats[0], (PathBuf::from("f1".to_string()), 2_usize));
-        assert_eq!(stats[1], (PathBuf::from("f2".to_string()), 3_usize));
-        assert_eq!(stats[2], (PathBuf::from("hotcache".to_string()), 18_usize));
+        assert_eq!(stats[0], (PathBuf::from("f1".to_string()), 2_u64));
+        assert_eq!(stats[1], (PathBuf::from("f2".to_string()), 3_u64));
+        assert_eq!(stats[2], (PathBuf::from("hotcache".to_string()), 18_u64));
 
         Ok(())
     }
 
-    #[test]
-    fn test_bundle_directory() -> anyhow::Result<()> {
+    #[tokio::test]
+    async fn test_bundle_directory() -> anyhow::Result<()> {
         let temp_dir = tempfile::tempdir()?;
         let test_filepath1 = temp_dir.path().join("f1");
         let test_filepath2 = temp_dir.path().join("f2");
-        let test_bundle_path = temp_dir.path().join("le_bundle");
-
-        let mut bundle_file = File::create(test_bundle_path.to_owned())?;
-        let mut bundle_builder = BundleStorageBuilder::new(&mut bundle_file)?;
 
         let mut file1 = File::create(&test_filepath1)?;
         file1.write_all(&[123, 76])?;
@@ -243,18 +227,52 @@ mod tests {
         let mut file2 = File::create(&test_filepath2)?;
         file2.write_all(&[99, 55, 44])?;
 
-        bundle_builder.add_file(&test_filepath1)?;
-        bundle_builder.add_file(&test_filepath2)?;
-        bundle_builder.finalize()?;
+        let split_streamer = get_split_payload_streamer(
+            &[test_filepath1.clone(), test_filepath2.clone()],
+            &[
+                1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18,
+            ],
+        )?;
 
-        let buffer = fs::read(test_bundle_path)?;
-        let bundle_file_slice = FileSlice::from(buffer);
+        let buffer = split_streamer.read_all().await?;
 
-        let bundle_dir = BundleDirectory::open_bundle(bundle_file_slice)?;
+        let bundle_file_slice = FileSlice::from(buffer.to_vec());
+
+        let bundle_dir = BundleDirectory::open_split(bundle_file_slice)?;
 
         assert!(bundle_dir.exists(Path::new("f1")).unwrap());
         assert!(bundle_dir.exists(Path::new("f2")).unwrap());
         assert!(!bundle_dir.exists(Path::new("f3")).unwrap());
+
+        let f1_data = bundle_dir.atomic_read(Path::new("f1"))?;
+        assert_eq!(&*f1_data, &[123u8, 76u8]);
+
+        let f2_data = bundle_dir.atomic_read(Path::new("f2"))?;
+        assert_eq!(&f2_data[..], &[99, 55, 44]);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_stream_split_to_bundle_and_open() -> anyhow::Result<()> {
+        let temp_dir = tempfile::tempdir()?;
+        let test_filepath1 = temp_dir.path().join("f1");
+        let test_filepath2 = temp_dir.path().join("f2");
+
+        let mut file1 = File::create(&test_filepath1)?;
+        file1.write_all(&[123, 76])?;
+
+        let mut file2 = File::create(&test_filepath2)?;
+        file2.write_all(&[99, 55, 44])?;
+
+        let split_streamer = get_split_payload_streamer(
+            &[test_filepath1.clone(), test_filepath2.clone()],
+            &[1, 2, 3],
+        )?;
+
+        let data = split_streamer.read_all().await?;
+
+        let bundle_dir = BundleDirectory::open_split(FileSlice::from(data.to_vec()))?;
 
         let f1_data = bundle_dir.atomic_read(Path::new("f1"))?;
         assert_eq!(&*f1_data, &[123u8, 76u8]);
--- a/quickwit-directories/src/bundle_directory.rs
+++ b/quickwit-directories/src/bundle_directory.rs
@@ -180,7 +180,7 @@ mod tests {
     use std::fs::File;
     use std::io::Write;
 
-    use quickwit_storage::{get_split_payload_streamer, PutPayload};
+    use quickwit_storage::{PutPayload, SplitPayloadBuilder};
 
     use super::*;
 
@@ -196,7 +196,7 @@ mod tests {
         let mut file2 = File::create(&test_filepath2)?;
         file2.write_all(&[99, 55, 44])?;
 
-        let split_streamer = get_split_payload_streamer(
+        let split_streamer = SplitPayloadBuilder::get_split_payload(
             &[test_filepath1.clone(), test_filepath2.clone()],
             &[
                 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18,
@@ -227,7 +227,7 @@ mod tests {
         let mut file2 = File::create(&test_filepath2)?;
         file2.write_all(&[99, 55, 44])?;
 
-        let split_streamer = get_split_payload_streamer(
+        let split_streamer = SplitPayloadBuilder::get_split_payload(
             &[test_filepath1.clone(), test_filepath2.clone()],
             &[
                 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18,
@@ -265,7 +265,7 @@ mod tests {
         let mut file2 = File::create(&test_filepath2)?;
         file2.write_all(&[99, 55, 44])?;
 
-        let split_streamer = get_split_payload_streamer(
+        let split_streamer = SplitPayloadBuilder::get_split_payload(
             &[test_filepath1.clone(), test_filepath2.clone()],
             &[1, 2, 3],
         )?;
--- a/quickwit-indexing/src/actors/packager.rs
+++ b/quickwit-indexing/src/actors/packager.rs
@@ -18,10 +18,7 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::collections::HashSet;
-use std::fs::File;
 use std::io;
-use std::io::Write;
-use std::ops::Range;
 use std::path::{Path, PathBuf};
 
 use anyhow::Context;
@@ -29,8 +26,6 @@ use fail::fail_point;
 use itertools::Itertools;
 use quickwit_actors::{Actor, ActorContext, Mailbox, QueueCapacity, SyncActor};
 use quickwit_directories::write_hotcache;
-use quickwit_storage::{BundleStorageBuilder, BUNDLE_FILENAME};
-use tantivy::common::CountingWriter;
 use tantivy::schema::Field;
 use tantivy::{ReloadPolicy, SegmentId, SegmentMeta};
 use tracing::*;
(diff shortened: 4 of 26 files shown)


quickwit-storage/src/bundle_storage.rs
```diff
@@ -22,15 +22,14 @@ use std::convert::TryInto;
...
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use async_trait::async_trait;
 use quickwit_common::chunk_range;
 use serde::{Deserialize, Serialize};
-use tantivy::common::CountingWriter;
 use tantivy::directory::FileSlice;
 use tantivy::HasLen;
 use thiserror::Error;
```
fmassot, line 35: I think we can get rid of 
```rust
/// Filename used for the bundle.
pub const BUNDLE_FILENAME: &str = "bundle";
```

PSeitz: yep, removed

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+}
+
+#[async_trait]
+impl PutPayload for SplitStreamer {
+    async fn len(&self) -> io::Result<u64> {
+        Ok(self.split_size as u64)
+    }
+
+    async fn range_byte_stream(&self, range: Range<u64>) -> io::Result<ByteStream> {
```
fmassot, line 67: `range_byte_stream` is a little bit hard to understand. I think we could do better, ideally something like this could be very nice, I admit I would need more thoughts to be sure this is not stupid...

```
let bytes_futures = Vec::new()
for file_slice in self.get_file_slices(range) {
   bytes_futures.push(file_slice.read_bytes_async())
}
// wrap that into a bytes stream and returns.
```
The annoying stuff would still be there in `get_file_slices(range)` as it's here that we would need to handle the corner cases. I can spend some time tomorrow and try to come up with something that should be reliable, I admit that for the moment it's just my intuition that wants to simplify the function.

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,379 @@
...
+            files: metadata_with_fixed_paths,
+        })?;
+
+        footer_bytes.write_all(metadata_json.as_bytes())?;
+        let metadata_json_len = metadata_json.len() as u64;
+        footer_bytes.write_all(&metadata_json_len.to_le_bytes())?;
+        footer_bytes.write_all(hotcache)?;
+        footer_bytes.write_all(&hotcache.len().to_le_bytes())?;
+
+        let mut files_and_offsets = self
```
fmassot, line 209: `block_offsets` and `files_and_offsets` seem redundant, maybe if we manage to simplify the `range_byte_stream` we will manage to simplify this.

fmassot (commented):
thanks for this PR, I think the big blocks are good. I'm just worried about the naming and the `range_byte_stream` that could be improved. Let's discuss that tomorrow.

quickwit-storage/src/bundle_storage.rs
```diff
@@ -82,37 +86,47 @@ pub struct CorruptedData {
...
-    fn open(data: &[u8]) -> Result<Self, CorruptedData> {
-        let (body_and_footer, footer_num_bytes_data) =
-            data.split_at(data.len() - SPLIT_HOTBYTES_FOOTER_LENGTH_NUM_BYTES);
-        let footer_num_bytes: u64 = u64::from_le_bytes(footer_num_bytes_data.try_into().unwrap());
-        let footer_slice = &body_and_footer[body_and_footer.len() - footer_num_bytes as usize..];
-        let bundle_storage_file_offsets = serde_json::from_slice(footer_slice)?;
-        Ok(bundle_storage_file_offsets)
+    /// Split data includes to hotcache
+    /// Returns (Hotcache, Self)
+    fn open_from_split_data(file: FileSlice) -> io::Result<(FileSlice, Self)> {
```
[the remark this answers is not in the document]
fulmicoton: This function has the same prototype as the one below. The name does not help me understand what is the difference. The comment barely.

Can you work on the comment or find a better name?

quickwit-cli/Cargo.toml
```diff
@@ -38,6 +38,7 @@ tokio = { version = "1", features = ["full"] }
 tokio-util = { version = "0.6", features = ["full"] }
 crossterm = "0.20"
 atty = "0.2"
+tantivy = { git= "https://github.com/quickwit-inc/tantivy", rev="ff631ec", default-features=false, features = ["mmap", "lz4-compression"] }
```
fulmicoton, line 41: You had to add tantivy as a deps of quickwit-cli because of the construction of the FileSlice.
Let's either add a From<> implem in tantivy or add an extra constructor to BundleStorage.

quickwit-directories/src/bundle_directory.rs
```diff
@@ -264,4 +253,34 @@ mod tests {
...
+
+        let mut file1 = File::create(&test_filepath1)?;
+        file1.write_all(&[123, 76])?;
+
+        let mut file2 = File::create(&test_filepath2)?;
+        file2.write_all(&[99, 55, 44])?;
+
+        let split_streamer = get_split_streamer(
+            &[test_filepath1.clone(), test_filepath2.clone()],
+            &[1, 2, 3],
```
fulmicoton, line 271: Not even a nitpick. just a matter of taste:
 
For meaninless `&[u8]` in unit test, I like to use bytestrings : b"234234".

Merits
- compactness
- it tells the reader that this is a &[u8] (right now the array could be &[usize])

quickwit-indexing/src/actors/packager.rs
```diff
@@ -206,9 +181,9 @@ fn merge_segments_if_required(
     Ok(segment_metas_after_merge)
 }
 
-fn build_hotcache<W: io::Write>(split_path: &Path, split_file: &mut W) -> anyhow::Result<()> {
+fn build_hotcache<W: io::Write>(split_path: &Path, out: &mut W) -> anyhow::Result<()> {
```
fulmicoton, line 184: Good rename!

quickwit-indexing/src/split_store/local_split_store.rs
```diff
@@ -24,33 +24,40 @@ use std::{fs, io};
...
+        )
+    })?)?;
+    let split_fileslice = mmap_directory.open_read(Path::new(&split_file))?;
+    Ok(Box::new(BundleDirectory::open_split(split_fileslice)?))
+}
+
 #[derive(Clone, Debug)]
-pub struct BundledSplitFile {
+pub struct SplitFolder {
     pub path: PathBuf,
```
fulmicoton, line 49: That's not from your code, but we should either make path non-pub, or if it is not possible, remove the `new` constructor ?

quickwit-storage/src/payload.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+impl Clone for Box<dyn PutPayload> {
+    fn clone(&self) -> Box<dyn PutPayload> {
+        self.box_clone()
+    }
+}
+
+#[async_trait]
+impl PutPayload for Vec<u8> {
```
fulmicoton, line 74: Can we change this `Vec<u8>` for `Arc<[u8]>` or OwnedBytes?
(The goal is to make sure cloning is light for all implementation of PutPayload and avoid bad surprise.)

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,379 @@
...
+        } else {
+            0
+        };
+        let last_block_take =
+            range.end - self.block_offsets[end_index].start as u64 - same_block_skip;
+
+        // build data_sources, apply seek in first
+        let mut datasources: Vec<Box<dyn DataSource>> = vec![];
+        let files_len = self.files_and_offsets.len();
+        let mut first_ds = true;
```
fulmicoton, line 86: what does ds stand for? datasource?

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,379 @@
...
+    let offsets = split_offset_computer.finalize(hotcache)?;
+
+    Ok(offsets)
+}
+
+fn get_start_and_end_index(
+    blocks_ranges: &[Range<usize>],
+    range: Range<usize>,
+) -> io::Result<(usize, usize)> {
+    let get_start = blocks_ranges
```
fulmicoton, line 139: I like the optimisation for readability & correctness.

quickwit-storage/src/split.rs
fulmicoton: The logic associated to first_ds is way too complicated below.

That kind of sentinel is generally ugly... but I think it is ok to use it like this.

```rust
let mut first_el: bool = false;
for i in els {
    if first_el {
     //...
       first_el = false;
    }
    
} 

```

In your code, this is way more complex.

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,379 @@
...
+        };
+        let last_block_take =
+            range.end - self.block_offsets[end_index].start as u64 - same_block_skip;
+
+        // build data_sources, apply seek in first
+        let mut datasources: Vec<Box<dyn DataSource>> = vec![];
+        let files_len = self.files_and_offsets.len();
+        let mut first_ds = true;
+
+        if start_index < files_len {
```
fulmicoton, line 88: It seems like a lot of the complexity is coming from the fact that you need to special case the footer bytes.

I think a no brainer to simplify this code is to introduce a private method `get_data_source` to this struct that takes a `chunk id, and a range`.
The function should start with a defensive assert checking if the range matches the given chunk, and return the right data source.

The current method should then be trivial.

fulmicoton (changes requested):
See inline comments.

quickwit-storage/src/split.rs
PSeitz: yes ds = datasource

I don't like this either, I initially wanted to apply a seek afterwards on the array, but it's not possible because `take` does not return the `Seek` trait. So we cannot have 
`let mut datasources: Vec<Box<dyn SeekableDataSource>> = vec![];`

The first_el is doing exactly what you are proposing (except the initial value)

quickwit-storage/src/bundle_storage.rs
```diff
@@ -82,37 +86,47 @@ pub struct CorruptedData {
...
+            bundle_and_hotcache_bytes.split_from_end(hotcache_num_bytes as usize);
+        Ok((hotcache, Self::open_from_file_slice(bundle)?))
     }
 
     /// Read metadata from a file.
     pub fn open_from_file_slice(file: FileSlice) -> io::Result<Self> {
-        let (body_and_footer, footer_num_bytes_data) =
-            file.split_from_end(SPLIT_HOTBYTES_FOOTER_LENGTH_NUM_BYTES);
+        let (file_data, num_bytes_filemetadata) =
+            file.split_from_end(BUNDLEMETADATA_LENGTH_NUM_BYTES);
```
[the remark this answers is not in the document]
PSeitz: yes good point

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,379 @@
...
+use crate::{BundleStorageFileOffsets, PutPayload};
+
+trait DataSource: tokio::io::AsyncRead + Send + Sync + Unpin {}
+impl<T: tokio::io::AsyncRead + Send + Sync + Unpin> DataSource for T {}
+
+/// SplitStreamer to Stream the split via the PutPayload.
+/// Bundles together multiple files into a single file
+/// with some metadata
+#[derive(Clone)]
+pub struct SplitStreamer {
```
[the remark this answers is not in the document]
PSeitz: yep

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,379 @@
...
+    /// data of the footer, containng
+    /// [bundle metadata, bundle metadata len 8byte, hotcache, hotcache len 8 bytes]
+    pub footer_bytes: Vec<u8>,
+    /// The end offset of the data provider [File, .. File, SplitFooter]
+    pub block_offsets: Vec<Range<usize>>,
+}
+
+impl SplitStreamer {
+    fn get_start_and_end_index(&self, range: Range<usize>) -> io::Result<(usize, usize)> {
+        get_start_and_end_index(&self.block_offsets, range)
```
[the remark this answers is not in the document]
PSeitz: Generally I prefer free functions over methods, because 
1. you can easily reuse them elsewhere
2. the required data is clear when reading and not just "self"

quickwit-storage/src/split.rs
PSeitz: I did some refactoring which reduces complexity a little, but having file and bytes as datasource has some complexity

quickwit-storage/src/split.rs
PSeitz: I added the method, because of the seek there is still some complexity, but it's improved

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,379 @@
...
+}
+
+/// SplitOffsetComputer is used to bundle together multiple files into a single split file.
+#[derive(Default, Debug)]
+pub struct SplitOffsetComputer {
+    metadata: BundleStorageFileOffsets,
+    current_offset: usize,
+}
+
+impl SplitOffsetComputer {
```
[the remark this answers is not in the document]
PSeitz: yep

quickwit-storage/src/split.rs
PSeitz: it's redundant, we could use files_and_offsets, but like this it allows for a minimal api when searching for the block

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,379 @@
...
+use rusoto_core::ByteStream;
+use tokio::io::{AsyncReadExt, AsyncSeekExt};
+use tokio_util::io::ReaderStream;
+
+use crate::{BundleStorageFileOffsets, PutPayload};
+
+trait DataSource: tokio::io::AsyncRead + Send + Sync + Unpin {}
+impl<T: tokio::io::AsyncRead + Send + Sync + Unpin> DataSource for T {}
+
+/// SplitStreamer to Stream the split via the PutPayload.
```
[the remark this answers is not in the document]
PSeitz: yep that's better

quickwit-cli/Cargo.toml
PSeitz: I tried the From<> version but there is a conflict with the generic From impl
```
impl<B> From<B> for FileSlice
where
    B: StableDeref + Deref<Target = [u8]> + 'static + Send + Sync,
```

quickwit-indexing/src/split_store/local_split_store.rs
PSeitz: done

quickwit-storage/src/payload.rs
PSeitz: Currently this code is only used by tests and would make them more complex

quickwit-storage/src/split.rs
PSeitz: the new version fixes this uglyness

quickwit-directories/src/bundle_directory.rs
PSeitz: It can be nice, my only issue with it is that the exact bytes behind the string may not be clear

PSeitz:
@fmassot @fulmicoton I moved the hotcache in the bundle footer, now the bundle is the split. This removes some confusing parts

quickwit-indexing/src/actors/uploader.rs
```diff
@@ -263,7 +273,7 @@ mod tests {
     use crate::models::ScratchDirectory;
 
     #[tokio::test]
-    async fn test_uploader() -> anyhow::Result<()> {
+    async fn test_uploader_1() -> anyhow::Result<()> {
```
fulmicoton, line 276: why uploader_1?

PSeitz: to execute only this test with a filter

quickwit-storage/src/split_creation.rs
```diff
@@ -0,0 +1,404 @@
...
+        // build data_sources, apply seek in first
+        let mut datasources: Vec<Box<dyn DataSource>> = vec![];
+        for index in start_index..=end_index {
+            // seek in first datasource
+            let seek_offset = if index == start_index {
+                start_file_offset
+            } else {
+                0
+            };
+            let mut ds = self.get_datasource(index, seek_offset).await?;
```
fulmicoton, line 119: I think it would have been even simpler by pushing more of the logic in `get_datasource`.

```get_datasource(file_index: usize, range: Range<u64>) -> Box<dyn DataSource>```

File: quickwit-storage/src/split_creation.rs [hunk 20]
```diff
@@ -0,0 +1,404 @@
...
+
+#[async_trait]
+impl PutPayload for SplitPayload {
+    async fn len(&self) -> io::Result<u64> {
+        Ok(self.split_size as u64)
+    }
+
+    async fn range_byte_stream(&self, range: Range<u64>) -> io::Result<ByteStream> {
+        let (start_index, end_index) =
+            self.get_start_and_end_block_index(range.start as usize..range.end as usize)?;
```
fulmicoton, line 97: In the same function, you use the word start and end for `[start, end)` AND `[start, end]` semantics.

fulmicoton: If you can remove the use of inclusive range this is better. If you cannot, please return  a RangeInclusive object.

quickwit-storage/src/split_creation.rs
```diff
@@ -0,0 +1,404 @@
...
+        let start_file_offset = range.start - self.block_offsets[start_index].start as u64;
+        // If we have skip and take in the same block, we need to take the skip into account for
+        // the take calculation
+        let same_block_skip = if start_index == end_index {
+            start_file_offset
+        } else {
+            0
+        };
+        let last_block_take =
+            range.end - self.block_offsets[end_index].start as u64 - same_block_skip;
```
fulmicoton, line 108: See comment below...

You can make the logic so much simpler by changing the contract of get_datasource

quickwit-storage/src/split_creation.rs
```diff
@@ -0,0 +1,404 @@
...
+        if index < self.files_and_offsets.len() {
+            let (path, _) = &self.files_and_offsets[index];
+            let mut ds = Box::new(tokio::fs::File::open(path).await?);
+            if start_file_offset != 0 {
+                ds.seek(SeekFrom::Start(start_file_offset)).await?;
+            }
+            Ok(ds)
+        } else {
+            let mut ds = Box::new(Cursor::new(self.footer_bytes.to_vec()));
+            if start_file_offset != 0 {
```
fulmicoton, line 81: Why is this repeated within the two branches of the if statement?

quickwit-storage/src/split_creation.rs
```diff
@@ -0,0 +1,404 @@
...
+        let metadata_json = serde_json::to_string(&BundleStorageFileOffsets {
+            files: metadata_with_fixed_paths,
+            files_and_data: iter::once((
+                PathBuf::from(HOTCACHE_FILENAME.to_string()),
+                OwnedBytes::new(hotcache.to_owned()),
+            ))
+            .collect(),
+        })?;
+
+        footer_bytes.write_all(metadata_json.as_bytes())?;
```
fulmicoton, line 225: `.extend`

quickwit-storage/src/split_creation.rs
```diff
@@ -0,0 +1,404 @@
...
+        let metadata_json_len = metadata_json.len() as u64;
+        footer_bytes.write_all(&metadata_json_len.to_le_bytes())?;
```
fulmicoton, line 227: `.extend`...

(The difference is mostly that we explicitly know we never return an error.)

quickwit-storage/src/split_creation.rs
```diff
@@ -0,0 +1,404 @@
...
+
+        let mut files_and_offsets = self
+            .metadata
+            .files
+            .iter()
+            .map(|(file, range)| (file.to_owned(), range.start as usize..range.end as usize))
+            .collect::<Vec<_>>();
+        files_and_offsets.sort_by_key(|(_file, range)| range.start);
+
+        let mut block_offsets = self
```
fulmicoton, line 237: Adding the type of files_and_offsets and block_offsets would really help understanding the difference here.

quickwit-storage/src/split_creation.rs
fulmicoton: Ah so it is not doing exactly doing what I thought it was doing.

quickwit-storage/src/split_creation.rs
PSeitz: hmm, I don't think this makes sense a `range` can cover multiple datasources

quickwit-storage/src/split_creation.rs
fulmicoton: Your `get_start_and_end_block_index` function doc is very clear. Then implementation is also pretty easy.
The contract choice however is rather unexpected. I don't think people would expect that.

There is a clunky side effect here. The last datasource can be empty. I don't think it yields a bug (I'd like a unit test about it though), but we end up opening a file for no reason at all.

quickwit-storage/src/split_creation.rs
PSeitz: done

quickwit-storage/src/split_creation.rs
PSeitz: Because they are different types. The trait can't seek or we can't apply take later.

quickwit-storage/src/split_creation.rs
```diff
@@ -0,0 +1,404 @@
...
+    blocks_ranges: &[Range<usize>],
+    range: Range<usize>,
+) -> io::Result<(usize, usize)> {
+    let get_start = blocks_ranges
+        .iter()
+        .position(|block_range| block_range.end > range.start)
+        .ok_or_else(|| io::Error::new(ErrorKind::InvalidInput, "range is out of Range"))?;
+    let get_end = blocks_ranges
+        .iter()
+        .position(|block_range| block_range.end >= range.end)
```
fulmicoton, line 162: This does not match the function comment.

quickwit-storage/src/payload.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use async_trait::async_trait;
+use rusoto_core::ByteStream;
+use tantivy::directory::OwnedBytes;
+
+#[async_trait]
+/// PutPayload is used to upload data and support multipart.
+pub trait PutPayload: PutPayloadClone + Send + Sync {
+    /// Return the total length of the payload.
+    async fn len(&self) -> io::Result<u64>;
```
fulmicoton, line 31: Does not need to be async nor return io::Result anymore.

quickwit-storage/src/split_creation.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// [bundle metadata, bundle metadata len 8byte, hotcache, hotcache len 8 bytes]
+    pub footer_bytes: Vec<u8>,
+    /// The end offset of the data provider [File, .. File, SplitFooter]
+    pub block_offsets: Vec<Range<usize>>,
+}
+
+impl SplitPayload {
+    /// Get start and end block index for a range, where start of the Range is in the start block
+    /// and end of the range is in the end block.
+    fn get_start_and_end_block_index(&self, range: Range<usize>) -> io::Result<(usize, usize)> {
```
fulmicoton, line 60: This should probably not return an io::Result

fulmicoton (changes requested):
See comments.

quickwit-storage/src/split_creation.rs
```diff
@@ -0,0 +1,404 @@
...
+                })?);
+                Ok((file_name, range.start..range.end))
+            })
+            .collect::<Result<HashMap<_, _>, io::Error>>()?;
+
+        let metadata_json = serde_json::to_string(&BundleStorageFileOffsets {
+            files: metadata_with_fixed_paths,
+            files_and_data: iter::once((
+                PathBuf::from(HOTCACHE_FILENAME.to_string()),
+                OwnedBytes::new(hotcache.to_owned()),
```
fulmicoton, line 220: Wait... how does that work?

fulmicoton (changes requested):
Let's not change the split format :
- without discussion in an issue
- within a PR that is about something totally differnet.

fulmicoton:
Here is a suggestion on how to simplify the PR.
It is not mergeable as is!

For instance, Vec should not be cloned that way...
It si probably not correct.
etc.

The point is to show how the stacking logic can simplified by working over a 
Vec<Box<dyn PutPayload>>.


https://github.com/quickwit-inc/quickwit/commit/29d5775d4c3b922c1f31ffe9aedb5020b12eb56b

quickwit-storage/src/bundle_storage.rs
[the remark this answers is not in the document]
PSeitz: I updated the names and comments

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,469 @@
...
+#[async_trait]
+impl PutPayload for SplitPayload {
+    fn len(&self) -> u64 {
+        self.payloads.iter().map(|payload| payload.len()).sum()
+    }
+
+    async fn range_byte_stream(&self, range: Range<u64>) -> io::Result<ByteStream> {
+        range_byte_stream_from_payloads(&self.payloads, range).await
+
+        // let requested_payload_ranges = chunk_range(&absolute_payload_ranges, range.start as usize
```
fmassot, line 73: Remove commented code.

fmassot: nitpick: this is the unique call of `range_byte_stream_from_payloads` so you can directly put the implementation in the `SplitPayload.

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,469 @@
...
+        //} else {
+        //// This is the last chunk
+        // bytestreams.push(payload.range_byte_stream(range.start..range.end).await?);
+        // break;
+        //}
+    }
+}
+
+/// compute split offsets used by the SplitStreamer
+pub fn get_split_payload_streamer(
```
fmassot, line 107: with the refactoring and renaming I think `get_split_payload_streamer` is just building the `SplitPayload` and that should be done in the `SplitPayloadBuilder`.

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,469 @@
...
+
+/// SplitOffsetComputer is used to bundle together multiple files into a single split file.
+#[derive(Default, Debug)]
+pub struct SplitPayloadBuilder {
+    metadata: BundleStorageFileOffsets,
+    current_offset: usize,
+}
+
+impl SplitPayloadBuilder {
+    /// Creates a new SplitOffsetComputer, to which files and the hotcache can be added.
```
fmassot, line 155: Update docs.

quickwit-storage/src/split.rs
```diff
@@ -0,0 +1,469 @@
...
+        let relative_range_in_payload = current_range.start - payload_range.start
+            ..payload_range_absolute_end - payload_range.start;
+
+        ranges.push((payload.clone(), relative_range_in_payload));
+
+        current_range = payload_range.end..current_range.end;
+    }
+
+    ranges
+}
```
fmassot, line 274: I struggled a little bit to review this function, I feel that this is a little bit clearer to compute the absolute ranges in it directly so that we see exactly what we are doing in this function. What do you think?
You can update the docs too.

```rust
/// Returns payloads and their relative ranges that overlap with the input `range`.
fn chunk_payload_ranges(
    payloads: &[Box<dyn PutPayload>],
    range: Range<usize>,
) -> Vec<(Box<dyn PutPayload>, Range<usize>)> {
    let mut ranges = vec![];
    let mut current_payload_absolute_start = 0usize;
    for payload in payloads {
        let current_payload_absolute_end = current_payload_absolute_start + payload.len() as usize;
        let absolute_range_overlap = get_ranges_overlap(&(current_payload_absolute_start..current_payload_absolute_end), &range);
        if !absolute_range_overlap.is_empty() {
            // Push the range relative to this payload as we will read from it.
            ranges.push((payload.clone(), (absolute_range_overlap.start - current_payload_absolute_start)..(absolute_range_overlap.end - current_payload_absolute_start)));
        }
        current_payload_absolute_start = current_payload_absolute_end;
    }
    ranges
}
```

fmassot: (ps I tested this function)

fmassot (changes requested):
I tested merge/demux/indexing on this branch and everything is working well, good job. I added a bunch of comments.

quickwit-common/src/lib.rs
```diff
@@ -30,6 +31,10 @@ use tracing::{error, info};
 /// Filenames used for hotcache files.
 pub const HOTCACHE_FILENAME: &str = "hotcache";
 
+pub fn get_hotcache_path() -> &'static Path {
```
fmassot, line 34: This function is not used I think. You can put `HOTCACHE_FILENAME` directly in the bundle directory module as it's only used here, currently you don't use the variable `PathBuf::from("hotcache".to_string()),`

quickwit-storage/src/split.rs
PSeitz: yes this is easier to read

quickwit-common/src/lib.rs
PSeitz: they are  both not used anymore :)

fmassot:
@PSeitz one more thing :)

I noticed that the debug logs are not readable because you add the `hotcache_bytes` in the `PackagedSplit` struct. Can you add the implementation of `fmt::Debug` for this struct to avoid that?

Thanks!

PSeitz:
@fmassot Thanks, I fixed the `PackagedSplit` debugging

quickwit-cli/src/lib.rs
```diff
@@ -211,7 +211,11 @@ pub async fn extract_split_cli(args: ExtractSplitArgs) -> anyhow::Result<()> {
     let split_file = PathBuf::from(format!("{}.split", args.split_id));
     let (_, bundle_footer) = read_split_footer(index_storage.clone(), &split_file).await?;
 
-    let bundle_storage = BundleStorage::new(index_storage, split_file, &bundle_footer)?;
+    let (_hotcache_bytes, bundle_storage) = BundleStorage::open_from_split_data_with_owned_bytes(
```
fmassot, line 214: There is something wrong here. The docs of `open_from_split_data_with_owned_bytes` indicates 
```
/// The provided data must include the footer_bytes at the end of the slice, but it can have
/// more up front.
```
What is the `footer_bytes`? Should it include the hotcache or not? By reading the code, we see that this method call `open_from_split_data` which returns `(Hotcache, BundleStorage)` so now we understand what it returns.

But... `read_split_footer` returns `(split_footer, only_bundle_footer)` so this can't work.

It seems that we can remove `open_from_split_data_with_owned_bytes`. Can you update the docs too? 

The docs of `read_split_footer` is not very clear too: `/// Loads the split footer from a storage and path.` but it returns 2 owned bytes `split_footer, only_bundle_footer`, I need to read the code to understand what it is exactly.

fulmicoton (approved):
LGTM but @fmassot can you comment? I think you tested it in the last few days

fmassot (changes requested):
See comment.

fmassot (approved):

PR #718, opened 2021-11-02
Implement Kinesis shard consumer

guilload:
### Description
Implement shard consumer, next building for implementing a Kinesis source.

### How was this PR tested?
Added some unit tests.

### Current issues
The Kinesis mock providers do not behave correctly on shard merge or split. Those tests are currently disabled.

### DIFF
--- a/quickwit-indexing/src/source/kinesis/api.rs
+++ b/quickwit-indexing/src/source/kinesis/api.rs
@@ -20,68 +20,14 @@
 // TODO: Remove when `KinesisSource` is fully implemented.
 #![allow(dead_code)]
 
-use std::collections::BTreeSet;
-use std::time::Duration;
-
-use anyhow::Context;
 use rusoto_kinesis::{
-    CreateStreamInput, DeleteStreamInput, DescribeStreamSummaryInput, GetRecordsInput,
-    GetRecordsOutput, GetShardIteratorInput, Kinesis, ListShardsInput, ListStreamsInput, Shard,
-    StreamDescriptionSummary,
+    GetRecordsInput, GetRecordsOutput, GetShardIteratorInput, Kinesis, ListShardsInput, Shard,
 };
-
-/// Creates a Kinesis data stream.
-/// https://docs.aws.amazon.com/kinesis/latest/APIReference/API_CreateStream.html
-async fn create_stream(
-    kinesis_client: &dyn Kinesis,
-    stream_name: &str,
-    num_shards: usize,
-) -> anyhow::Result<()> {
-    let request = CreateStreamInput {
-        stream_name: stream_name.to_string(),
-        shard_count: num_shards as i64,
-    };
-    // TODO: Implement retry.
-    kinesis_client
-        .create_stream(request)
-        .await
-        .with_context(|| format!("Failed to create Kinesis data stream `{}`.", stream_name))?;
-    Ok(())
-}
-
-/// Deletes a Kinesis data stream. Only streams in `ACTIVE` state can be deleted.
-/// https://docs.aws.amazon.com/kinesis/latest/APIReference/API_DeleteStream.html
-async fn delete_stream(kinesis_client: &dyn Kinesis, stream_name: &str) -> anyhow::Result<()> {
-    let request = DeleteStreamInput {
-        stream_name: stream_name.to_string(),
-        ..Default::default()
-    };
-    // TODO: Implement retry.
-    kinesis_client
-        .delete_stream(request)
-        .await
-        .with_context(|| format!("Failed to delete Kinesis data stream `{}`.", stream_name))?;
-    Ok(())
-}
-
-/// Provides a summarized description of the specified Kinesis data stream without the shard list.
-/// https://docs.aws.amazon.com/kinesis/latest/APIReference/API_DescribeStreamSummary.html
-async fn describe_stream(
-    kinesis_client: &dyn Kinesis,
-    stream_name: &str,
-) -> anyhow::Result<StreamDescriptionSummary> {
-    let request = DescribeStreamSummaryInput {
-        stream_name: stream_name.to_string(),
-    };
-    // TODO: Implement retry.
-    let response = kinesis_client.describe_stream_summary(request).await?;
-    Ok(response.stream_description_summary)
-}
 
 /// Gets records from a Kinesis data stream's shard.
 /// https://docs.aws.amazon.com/kinesis/latest/APIReference/API_GetRecords.html
-async fn get_records(
-    kinesis_client: &dyn Kinesis,
+pub(crate) async fn get_records(
+    kinesis_client: &(dyn Kinesis + Send + Sync),
     shard_iterator: String,
 ) -> anyhow::Result<GetRecordsOutput> {
     let request = GetRecordsInput {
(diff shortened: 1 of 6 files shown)


quickwit-indexing/src/source/kinesis/api.rs
```diff
@@ -183,10 +187,46 @@ async fn list_streams(
...
+        adjacent_shard_to_merge: adjacent_shard_id.to_string(),
+    };
+    // TODO: Implement retry.
+    kinesis_client.merge_shards(request).await?;
+    Ok(())
+}
+
+/// Splits a shard into two new shards in the Kinesis data stream.
+/// https://docs.aws.amazon.com/kinesis/latest/APIReference/API_SplitShard.html
+pub(crate) async fn split_shard(
```
fulmicoton, line 210: Is this something we trigger and we actually will do on quickwit? Or is an event we need to react to?

quickwit-indexing/src/source/kinesis/api.rs
```diff
@@ -183,10 +187,46 @@ async fn list_streams(
     Ok(stream_names)
 }
 
+/// Merges two adjacent shards in a Kinesis data stream and combines them into a single shard.
+/// https://docs.aws.amazon.com/kinesis/latest/APIReference/API_MergeShards.html
+pub(crate) async fn merge_shards(
```
fulmicoton, line 192: Same question as below

quickwit-indexing/src/source/kinesis/shard_consumer.rs
```diff
@@ -0,0 +1,340 @@
...
+
+use rusoto_kinesis::{Kinesis, Record};
+use tokio::sync::mpsc;
+use tokio::time::Duration;
+
+use crate::source::kinesis::api::{get_records, get_shard_iterator};
+
+#[derive(Debug)]
+enum ShardConsumerMessage {
+    NewShards(Vec<String>),
```
fulmicoton, line 31: Can you tell what NewShards is about?

quickwit-indexing/src/source/kinesis/shard_consumer.rs
```diff
@@ -0,0 +1,340 @@
...
+    ShardClosed(String),
+    /// The consumer has reached the latest record in the shard.
+    ShardEOF(String),
+}
+
+struct ShardConsumer {
+    stream_name: String,
+    shard_id: String,
+    starting_sequence_number: Option<String>,
+    eof_enabled: bool,
```
fulmicoton, line 47: A comment on the semantics of eof_enabled could be useful (the rest is fairly obvious)

quickwit-indexing/src/source/kinesis/shard_consumer.rs
```diff
@@ -0,0 +1,340 @@
...
+    kinesis_client: Box<dyn Kinesis + Send + Sync>,
+    sink: mpsc::Sender<ShardConsumerMessage>,
+}
+
+impl ShardConsumer {
+    fn new(
+        stream_name: String,
+        shard_id: String,
+        starting_sequence_number: Option<String>,
+        eof_enabled: bool,
```
fulmicoton, line 57: A comment on the semantics of eof_enabled could be useful

quickwit-indexing/src/source/kinesis/shard_consumer.rs
```diff
@@ -0,0 +1,340 @@
...
+    eof_enabled: bool,
+    kinesis_client: Box<dyn Kinesis + Send + Sync>,
+    sink: mpsc::Sender<ShardConsumerMessage>,
+}
+
+impl ShardConsumer {
+    fn new(
+        stream_name: String,
+        shard_id: String,
+        starting_sequence_number: Option<String>,
```
fulmicoton, line 56: A comment telling if it starts at the first el >= starting_sequence_number or the first el  > starting_sequence_number would be very useful

fulmicoton (changes requested):
Minor comments.
(I did not LGTM just because the PR is marked as in progress, as you want to actorify it)

quickwit-indexing/src/source/kinesis/api.rs
guilload: This is an event we react to (see `NewShards` message). This is implemented here for testing purposes.

quickwit-indexing/src/source/kinesis/shard_consumer.rs
guilload: Yep. I'm going to rename this message `ChildShards` actually.

quickwit-indexing/src/source/kinesis/shard_consumer.rs
guilload: Good point.

guilload (commented):
Thanks for the review. You're right, documentation is really sparse. I'll improve upon that.

guilload:
I have addressed the CR comments and "actorified" the shard consumer. This PR is ready for a final round of reviews.

quickwit-indexing/src/source/kinesis/shard_consumer.rs
```diff
@@ -0,0 +1,451 @@
...
+        ctx: &ActorContext<Self::Message>,
+    ) -> Result<(), ActorExitStatus> {
+        self.state.next_shard_iterator = get_shard_iterator(
+            &*self.kinesis_client,
+            &self.stream_name,
+            &self.shard_id,
+            self.from_sequence_number_exclusive.clone(),
+        )
+        .await?;
+        self.process_message(Loop {}, ctx).await?;
```
fulmicoton, line 133: Super nitpick. Probably just a taste story.

`struct Loop` and `Loop` works too.

quickwit-indexing/src/source/kinesis/shard_consumer.rs
```diff
@@ -0,0 +1,451 @@
...
+                    // Filter out duplicate message when two shards are merged.
+                    .filter(|child| child.parent_shards.first() == Some(&self.shard_id))
+                    .map(|child| child.shard_id)
+                    .collect();
+                if !shard_ids.is_empty() {
+                    let message = ShardConsumerMessage::ChildShards(shard_ids);
+                    ctx.send_message(&self.sink, message).await?;
+                }
+            }
+            if self.eof_enabled && response.millis_behind_latest.unwrap_or(-1) == 0 {
```
fulmicoton, line 178: I'd rather have `== Some(0)` instead of the unwrap_or(-1)

quickwit-indexing/src/source/kinesis/shard_consumer.rs
```diff
@@ -0,0 +1,451 @@
...
+                let message = ShardConsumerMessage::ShardEOF(self.shard_id.clone());
+                ctx.send_message(&self.sink, message).await?;
+                return Err(ActorExitStatus::Success);
+            };
+            // The `GetRecords` API has a limit of 5 transactions per second. 1s / 5 + ε = 205ms.
+            let interval = Duration::from_millis(205);
```
fulmicoton, line 184: Do we want to increase the interval if empty? There is not cost component associated to the number of req?

quickwit-indexing/src/source/kinesis/api.rs
fulmicoton: I'm not sure I understand. Could it be #[cfg(test)] for instance?

Can we discuss it a little tomorrow maybe?

fulmicoton (approved):

quickwit-indexing/src/source/kinesis/api.rs
guilload: Yeah, I added a few `#[cfg(test)]` in my `KinesisSource` branch after removing the `#![allow(dead_code)]`.

guilload: Sorry about the confusion. I think we agree on this and will land on the same code, but I concede that the path I'm taking to get there is not the most straightforward.

quickwit-indexing/src/source/kinesis/shard_consumer.rs
guilload: Ok.

quickwit-indexing/src/source/kinesis/shard_consumer.rs
guilload: So, a shard is divided into multiple buckets. Each call to `GetRecords` reads from one bucket. As a result, a call to `GetRecords` can return no records because the bucket is empty, but the next bucket might not, so you still want to poll at the same frequency.

The Kinesis pricing seems to take that into account, and API calls to `GetRecords` are "free."

PR #721, opened 2021-11-02
Metastore tests enhancement

evanxg852000:
### Description
- Added fine grained checks on metastore tests https://github.com/quickwit-inc/quickwit/issues/670

### How was this PR tested?
Execute unit test for File & Postgres metastores


evanxg852000:
@guilload I have removed these tests as I think they are  already tested implicitly from what's left: 
 - `test_metastore_create_index` 
 - `test_metastore_stage_split` 
 - `test_metastore_mark_splits_for_deletion`

This is what I got from the coverage metrics:
before:
postgresql_metastore.rs	(64.75%)
single_file_metastore.rs(77.91%)
after:
postgresql_metastore.rs	(67.15%)
single_file_metastore.rs(80.00%)

Also while extracting the coverage metrics, I found a bug in the tests with `update_timestamp`. Basically, as things take  time within the instrumentation, `assert_eq!` were failing on the `update_timestamp` field, I fixed it using https://crates.io/crates/derivative

quickwit-metastore/Cargo.toml
```diff
@@ -22,9 +22,9 @@ thiserror = "1.0"
 tokio = { version = "1.6", features = ["full"]}
 tracing = "0.1"
 openssl = { version = "0.10.36", optional = true }
-
 quickwit-index-config = { version = "0.1.0", path = "../quickwit-index-config" }
 quickwit-storage = { version = "0.1.0", path = "../quickwit-storage" }
+derivative = "2.2.0"
```
guilload, line 27: Should this dependency only enabled when the feature `testsuite` is on?

quickwit-metastore/src/tests.rs
```diff
@@ -113,13 +92,17 @@ pub mod test_suite {
             .create_index(index_metadata.clone())
             .await
             .unwrap();
+        assert_eq!(metastore.list_all_splits(index_id).await.unwrap().len(), 0);
```
guilload, line 95: Do we need an `index_exists` in the public API?

quickwit-metastore/src/tests.rs
```diff
@@ -113,13 +92,17 @@ pub mod test_suite {
...
 
         // Delete an index
         let result = metastore.delete_index(index_id).await.unwrap();
         assert!(matches!(result, ()));
+
+        // check index is really gone
+        let result = metastore.list_all_splits(index_id).await.unwrap_err();
```
guilload, line 102: nit: this is no longer a result since we called `unwrap_err`.

quickwit-metastore/src/tests.rs
```diff
@@ -297,6 +227,16 @@ pub mod test_suite {
                 .await
                 .unwrap();
             assert!(matches!(result, ()));
+            assert_eq!(
+                metastore.list_all_splits(index_id).await.unwrap(),
+                vec![SplitMetadataAndFooterOffsets {
```
guilload, line 232: Maybe a helper function can help us build this vector and reduce the verbosity of these tests since this pattern is repeated quite a few times.

guilload (commented):
Implementing helper functions for creating test structs such as `IndexMetadata` or `SplitMetadata` would help decrease the verbosity of this test suite.

I spotted some more opportunities for slashing some redundant tests. Let's pair program next week.

quickwit-metastore/src/tests.rs
evanxg852000: Maybe in the future but right now there is nowhere in the use of Metastore where we need to check if e this implicitly will tell us if index does not exist.

fulmicoton (approved):

fulmicoton:
@evanxg852000 I think we are good to go? Can you rebase and merge?

guilload:
I took over this PR and... dropped the ball... It's not ready. I will resume my work after the release.

PR #729, opened 2021-11-05
Added env var for the fast field/split cache capacity.

### LINKED ISSUE #728: Make the fast field cache size and the split footer cache size configurable
As a temporary solution, we can set the fastfield/split footer cache size using an environment variable.

- `SPLIT_FOOTER_CACHE_CAPACITY`: defaults to 500MB (500000000)
- `FAST_CACHE_CAPACITY`: defaults to 3GB (3000000000)

fulmicoton:
Closes #728


quickwit-storage/src/cache/quickwit_cache.rs
```diff
@@ -166,6 +175,18 @@ mod tests {
...
+    #[test]
+    fn test_fast_cache_capacity_from_env() {
+        assert_eq!(
+            super::fast_cache_capacity(),
+            super::DEFAULT_FAST_CACHE_CAPACITY
+        );
+        std::env::set_var(super::FAST_CACHE_CAPACITY_ENV_KEY, "3000000001");
+        assert_eq!(super::fast_cache_capacity(), 3_000_000_001);
+        std::env::set_var(super::FAST_CACHE_CAPACITY_ENV_KEY, "invalidnumber");
+        assert_eq!(super::fast_cache_capacity(), 3_000_000_000);
```
PSeitz, line 187: we should print at least a error here or the user will not know the value is rejected

fulmicoton: Great point

evanxg852000 (approved):
LGTM :+1:

PR #743, opened 2021-11-08
add checks before indexing and serve

### LINKED ISSUE #701: Run connectivity test on startup
It seems like a good 80% of the friction experienced by quickwit users are related to the proper configuration & connectivity of third-part services (source, store, metastore).

Right now, we surface these error of configuration only when they occur. For instance, if the credentials for the storage are wrong, indexing will fail after running for 30s, with a message that is relatively explicit for us, but might lack a bit of context for the user. 

Instead, we should have a nice way to test connectivity first and output a report stating clearly what piece of the puzzle is not configured correctly.

For the CLI, it would materialize in a familiar checklist list UI for instance.

----
- source  ❌
  <some extra informaiton>
- storage ✔️ 
- metastore connection ✔️ 
- index configuration ❌
  <some extra information>
-----

And all error details. 

This functionality is VERY likely to be reused in the near-future in an admin UI.
A little bit of overengineer does make sense here.

Typically, we should have a function taking some configuration as an input and returning the connectivity results as some structured object containing all the information necessary for reporting.

Printing this object as checklist will be done in a different function.

llogiq:
### Description
This adds various checks for config, storage and metastore. This fixes #701.

### How was this PR tested?

`cargo test --all-features`

### DIFF
--- a/quickwit-cli/Cargo.toml
+++ b/quickwit-cli/Cargo.toml
@@ -14,6 +14,7 @@ name = "quickwit"
 path = "src/main.rs"
 
 [dependencies]
+async-trait = "0.1"
 anyhow = "1"
 byte-unit = "4"
 clap = { version = "2", features = ["yaml"] }
@@ -46,6 +47,7 @@ humansize = "1.1.1"
 openssl-probe = { version = "0.1.4", optional = true }
 itertools = "0.10.1"
 colored = "2.0.0"
+futures = "0.3"
 
 [dev-dependencies]
 assert_cmd = "1"
--- a/quickwit-cli/src/lib.rs
+++ b/quickwit-cli/src/lib.rs
@@ -43,7 +43,7 @@ use quickwit_indexing::actors::{
 use quickwit_indexing::models::{CommitPolicy, IndexingDirectory, IndexingStatistics};
 use quickwit_indexing::source::{FileSourceParams, SourceConfig};
 use quickwit_metastore::checkpoint::Checkpoint;
-use quickwit_metastore::{IndexMetadata, MetastoreUriResolver};
+use quickwit_metastore::{do_checks, IndexMetadata, MetastoreUriResolver};
 use quickwit_proto::{SearchRequest, SearchResponse};
 use quickwit_search::{single_node_search, SearchResponseRest};
 use quickwit_storage::{quickwit_storage_uri_resolver, BundleStorage, Storage};
@@ -286,6 +286,13 @@ pub async fn index_data_cli(args: IndexDataArgs) -> anyhow::Result<()> {
     let storage_uri_resolver = quickwit_storage_uri_resolver();
     let metastore_uri_resolver = MetastoreUriResolver::default();
     let metastore = metastore_uri_resolver.resolve(&args.metastore_uri).await?;
+
+    do_checks(vec![
+        ("source", Box::pin(source_config.check_source_available())),
+        ("metastore", metastore.check_connectivity()),
+        ("index", metastore.check_index_available(&args.index_id)),
+    ])
+    .await?;
 
     if args.overwrite {
         reset_index(
--- a/quickwit-indexing/src/source/kafka_source.rs
+++ b/quickwit-indexing/src/source/kafka_source.rs
@@ -305,6 +305,12 @@ fn previous_position_for_offset(offset: i64) -> Position {
     }
 }
 
+/// Checks if connecting with the given parameters works.
+pub(super) fn check(client_params: serde_json::Value) -> anyhow::Result<()> {
+    create_consumer(None, client_params)?;
+    Ok(())
+}
+
 /// Creates a new `KafkaSourceConsumer`.
 fn create_consumer(
     client_log_level: Option<String>,
--- a/quickwit-indexing/src/source/mod.rs
+++ b/quickwit-indexing/src/source/mod.rs
@@ -198,3 +198,28 @@ pub struct SourceConfig {
     pub source_type: String,
     pub params: serde_json::Value,
 }
+
+impl SourceConfig {
+    pub async fn check_source_available(&self) -> anyhow::Result<()> {
+        match &*self.source_type {
+            "file" => {
+                if let Some(path) = self.params.get("filepath").and_then(|v| v.as_str()) {
+                    if !std::path::Path::new(path).exists() {
+                        anyhow::bail!("Source `{}` does not exist", path)
+                    }
+                }
+                Ok(())
+            }
+            #[cfg(feature = "kafka")]
+            "kafka" => kafka_source::check(self.params.clone()),
+            "vec" => Ok(()),
+            unrecognized_source_type => {
+                tracing::warn!(
+                    "No check implemented for source type `{}`",
+                    unrecognized_source_type
+                );
+                Ok(())
+            }
+        }
+    }
+}
--- a/quickwit-metastore/Cargo.toml
+++ b/quickwit-metastore/Cargo.toml
@@ -16,12 +16,13 @@ async-trait = "0.1"
 chrono = "0.4"
 diesel = { version = "1.4", features = ["postgres", "chrono", "extras"], optional = true }
 diesel_migrations =  { version = "1.4", optional = true }
+futures = "0.3.17"
+openssl = { version = "0.10.36", optional = true }
 serde = { version = "1.0", features = ["derive", "rc"] }
 serde_json = "1.0"
 thiserror = "1.0"
 tokio = { version = "1.6", features = ["full"]}
 tracing = "0.1"
-openssl = { version = "0.10.36", optional = true }
 
 quickwit-index-config = { version = "0.1.0", path = "../quickwit-index-config" }
 quickwit-storage = { version = "0.1.0", path = "../quickwit-storage" }
--- a/quickwit-metastore/src/lib.rs
+++ b/quickwit-metastore/src/lib.rs
@@ -49,7 +49,11 @@ mod metastore_resolver;
 #[allow(missing_docs)]
 pub mod postgresql;
 
+use std::pin::Pin;
+
 pub use error::{MetastoreError, MetastoreResolverError, MetastoreResult};
+use futures::stream::{FuturesUnordered, StreamExt};
+use futures::{Future, FutureExt};
 #[cfg(feature = "postgres")]
 pub use metastore::postgresql_metastore::PostgresqlMetastore;
 pub use metastore::single_file_metastore::SingleFileMetastore;
@@ -59,3 +63,44 @@ pub use metastore::{
     IndexMetadata, MetadataSet, Metastore, SplitMetadata, SplitMetadataAndFooterOffsets, SplitState,
 };
 pub use metastore_resolver::{MetastoreFactory, MetastoreUriResolver};
+
+fn decorate_results_with_output<F: Future<Output = anyhow::Result<()>>>(
+    description: &str,
+    f: F,
+) -> impl Future<Output = Option<anyhow::Error>> {
+    let description = description.to_owned();
+    f.map(|result| match result {
+        Ok(()) => {
+            eprintln!("✅ {}", description);
+            None
+        }
+        Err(e) => {
+            eprintln!("❌ {}", description);
+            Some(e.context(description))
+        }
+    })
+}
+
+/// A future that can be called with [`do_checks`]
+pub type CheckFuture<'a> = Pin<Box<(dyn Future<Output = anyhow::Result<()>> + Send + 'a)>>;
+
+/// Run a list of early checks
+pub async fn do_checks(checks: Vec<(&'_ str, CheckFuture<'_>)>) -> anyhow::Result<()> {
+    let mut all_checks: FuturesUnordered<_> = checks
+        .into_iter()
+        .map(|(desc, f)| decorate_results_with_output(desc, f))
+        .collect();
+    let mut errors = Vec::new();
+    while let Some(check) = all_checks.next().await {
+        if let Some(error) = check {
+            errors.push(error);
+        }
+    }
+    if !errors.is_empty() {
+        for error in errors {
+            tracing::error!("{}", error);
+        }
+        anyhow::bail!("Some checks failed")
+    }
+    Ok(())
+}
--- a/quickwit-metastore/src/metastore/mod.rs
+++ b/quickwit-metastore/src/metastore/mod.rs
@@ -194,6 +194,15 @@ pub struct MetadataSet {
 #[cfg_attr(any(test, feature = "testsuite"), mockall::automock)]
 #[async_trait]
 pub trait Metastore: Send + Sync + 'static {
+    /// Checks if the metastore is available.
+    async fn check_connectivity(&self) -> anyhow::Result<()>;
+
+    /// Checks if the given index is in this metastore.
+    async fn check_index_available(&self, index_id: &str) -> anyhow::Result<()> {
+        self.index_metadata(index_id).await?;
+        Ok(())
+    }
+
     /// Creates an index.
     /// This API creates index metadata set in the metastore.
     /// An error will occur if an index that already exists in the storage is specified.
--- a/quickwit-storage/src/bundle_storage.rs
+++ b/quickwit-storage/src/bundle_storage.rs
@@ -134,6 +134,18 @@ impl BundleStorageFileOffsets {
 
 #[async_trait]
 impl Storage for BundleStorage {
+    async fn check(&self) -> anyhow::Result<()> {
+        if !self
+            .storage
+            .exists(&self.bundle_filepath)
+            .await
+            .unwrap_or(false)
+        {
+            anyhow::bail!("`{}` not found in storage", self.bundle_filepath.display())
+        }
+        Ok(())
+    }
+
     async fn put(
         &self,
         path: &Path,
--- a/quickwit-storage/src/local_file_storage.rs
+++ b/quickwit-storage/src/local_file_storage.rs
@@ -163,6 +163,13 @@ impl From<PathBuf> for LocalFileStorage {
 
 #[async_trait]
 impl Storage for LocalFileStorage {
+    async fn check(&self) -> anyhow::Result<()> {
+        if !self.root.exists() {
+            anyhow::bail!("Missing path `{}`", self.root.display())
+        }
+        Ok(())
+    }
+
     async fn put(
         &self,
         path: &Path,
--- a/quickwit-storage/src/object_storage/s3_compatible_storage.rs
+++ b/quickwit-storage/src/object_storage/s3_compatible_storage.rs
@@ -34,8 +34,8 @@ use rusoto_core::{ByteStream, HttpClient, HttpConfig, Region, RusotoError};
 use rusoto_s3::{
     AbortMultipartUploadRequest, CompleteMultipartUploadRequest, CompletedMultipartUpload,
     CompletedPart, CreateMultipartUploadError, CreateMultipartUploadRequest, DeleteObjectRequest,
-    GetObjectRequest, HeadObjectError, HeadObjectRequest, PutObjectError, PutObjectRequest,
-    S3Client, UploadPartRequest, S3,
+    GetObjectRequest, HeadObjectError, HeadObjectRequest, ListObjectsV2Request, PutObjectError,
+    PutObjectRequest, S3Client, UploadPartRequest, S3,
 };
 use tokio::fs::File;
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWriteExt, BufReader};
@@ -454,6 +454,17 @@ async fn download_all(byte_stream: &mut ByteStream, output: &mut Vec<u8>) -> io:
 
 #[async_trait]
 impl Storage for S3CompatibleObjectStorage {
+    async fn check(&self) -> anyhow::Result<()> {
+        self.s3_client
+            .list_objects_v2(ListObjectsV2Request {
+                bucket: self.bucket.clone(),
+                max_keys: Some(1),
+                ..Default::default()
+            })
+            .await?;
+        Ok(())
+    }
+
     async fn put(
         &self,
         path: &Path,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2168,11 +2168,13 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
+ "async-trait",
  "atty",
  "byte-unit",
  "chrono",
  "clap",
  "colored",
+ "futures",
  "humansize",
  "itertools",
  "json_comments",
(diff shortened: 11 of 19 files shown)


quickwit-cli/src/lib.rs
```diff
@@ -278,14 +278,23 @@ pub async fn index_data_cli(args: IndexDataArgs) -> anyhow::Result<()> {
...
-    let source_config =
-        create_source_config_from_args(source_config_path_opt, input_path_opt).await?;
+    let mut source_config_res =
+        create_source_config_from_args(source_config_path_opt, input_path_opt).await;
     let indexing_directory_path = args.data_dir_path.join(args.index_id.as_str());
     let indexing_directory = IndexingDirectory::create_in_dir(indexing_directory_path).await?;
     let storage_uri_resolver = quickwit_storage_uri_resolver();
     let metastore_uri_resolver = MetastoreUriResolver::default();
     let metastore = metastore_uri_resolver.resolve(&args.metastore_uri).await?;
+    check_all!(
```
fulmicoton, line 288: The macro is overkill.

llogiq: What would you suggest instead?

quickwit-metastore/src/lib.rs
```diff
@@ -59,3 +63,86 @@ pub use metastore::{
     IndexMetadata, MetadataSet, Metastore, SplitMetadata, SplitMetadataAndFooterOffsets, SplitState,
 };
 pub use metastore_resolver::{MetastoreFactory, MetastoreUriResolver};
+
+/// A trait to allow for our data model to check itself
+#[async_trait]
+pub trait Check {
+    /// Something that can be implemented
```
fulmicoton, line 70: This is not a very good comment, and possibly not a good name.
When I first read "description", I was expecting either a description of the resource that was checked, or a description of what was checked.

Looking at the URI of the metastore it is neither: 
You just return the URI for the metastore for instance.

```
/// A concise description resuming the setting that were used to build an instance of 
/// this resource. The type of the resource (e.g. Metastore) does not need to be repeated. 
/// If the resource is built from an URI, returning just the URI is best.
```

quickwit-metastore/src/lib.rs
```diff
@@ -59,3 +63,86 @@ pub use metastore::{
...
+    fn description(&self) -> String;
+
+    /// An async function that implements the check on this type
+    async fn check(&self) -> Result<(), anyhow::Error>;
```
fulmicoton, line 74: I wonder if anyhow::Error is the best type for reporting.

quickwit-metastore/src/lib.rs
llogiq: You are right. I had thought more information would be beneficial here, but later reverted that and just added `&'static str`s to the `check_all!` macro. I'm going to remove this function.

quickwit-metastore/src/lib.rs
llogiq: The idea here was a minimal implementation, and since the checks touch various parts of the system with different error types, `anyhow::Error` looked like the least common denominator.

Do you have a suggestion what to use instead?

quickwit-metastore/src/lib.rs
fmassot: URI is good to identify what has been checked, I think it's also important to add the object type too. Something like this:
- For a storage:  `Storage - s3://my-bucket/indexes: ✔️`
- For a metastore:  `Metastore - file:///my-metastore-dir: ✔️`
- For a source:  `Source - file:///my-metastore-dir: ✔️`

quickwit-metastore/src/lib.rs
```diff
@@ -59,3 +63,68 @@ pub use metastore::{
...
+pub fn checked<T>(r: &mut anyhow::Result<T>) -> Pin<Box<(dyn Future<Output = anyhow::Result<()>> + Send)>> {
+    let res = match r {
+        Ok(_) => Ok(()),
+        Err(e) => Err(std::mem::replace(e, anyhow::anyhow!("taken"))),
+    };
+    Box::pin(std::future::ready(res))
+}
+
+/// Decorate a check with nice output
+pub fn map_results<F: Future<Output = anyhow::Result<()>>>(
```
fulmicoton, line 96: This function takes a future, and prints stuff in the stderr when it gets resolved.

The name of the function does not tell anything about what it does.

fulmicoton: As mentionned in the issue, we will want to use this check operation in different contexts.
The output will be the CLI today. In the future, we will very likely want to display its output in an admin UI.

That's where I would have preferred to have a tiny bit of overengineering and have a function that returns the outcome of the checks... 
And another that displays the outcome of the checks.

Here this function decorates -in the sense of a python decorator- the check function future.
It has the benefit of giving a cool CLI experience where each check appears asynchronously one ofter the other on stderr, but it is not as nice from the code POV.

fulmicoton: I'm curious... Is there a reason, why this is not using async?

quickwit-metastore/src/lib.rs
```diff
@@ -59,3 +63,68 @@ pub use metastore::{
...
+        }
+        Err(e) => {
+            eprintln!("❌ {}", description);
+            Some(e.context(description))
+        }
+    })
+}
+
+/// Run a list of early checks
+pub async fn do_checks<'a>(checks: impl IntoIterator<Item = (&'_ str, Pin<Box<(dyn Future<Output = anyhow::Result<()>> + Send + 'a)>>)>) -> anyhow::Result<()> {
```
fulmicoton, line 114: Generics are overkill too :) 
Vec or array would have been ok.
&'static str might have been ok too.

quickwit-metastore/src/lib.rs
fulmicoton: I checked the doc of anyhow::Error. It might be nice actually. It is really meant to be able to display different amount of context. Display is probably what we want.

quickwit-metastore/src/metastore/mod.rs
```diff
@@ -193,7 +193,15 @@ pub struct MetadataSet {
 /// running search queries can complete.
 #[cfg_attr(any(test, feature = "testsuite"), mockall::automock)]
 #[async_trait]
-pub trait Metastore: Send + Sync + 'static {
+pub trait Metastore: Check + Send + Sync + 'static {
+    /// Check if the given index is in this metastore
+    async fn check_index(&self, index_id: &str) -> anyhow::Result<()> {
+        if let Err(e) = self.index_metadata(index_id).await {
+            anyhow::bail!(e.to_string())
```
fulmicoton, line 200: `if ... { anyhow::bail!(e.to_string()) }` instead of  
`self.index_metadata(index_id).await?;` is meant to reduce the error message length?

quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -454,6 +454,13 @@ async fn download_all(byte_stream: &mut ByteStream, output: &mut Vec<u8>) -> io:
 
 #[async_trait]
 impl Storage for S3CompatibleObjectStorage {
+    async fn check(&self) -> anyhow::Result<()> {
+        match self.s3_client.list_buckets().await {
+            Ok(_) => Ok(()),
+            Err(e) => Err(e.into()),
+        }
```
fulmicoton, line 461: nitpick

```suggestion
        self.s3_client.list_buckets().await?;
        Ok(())
```

quickwit-metastore/src/lib.rs
fulmicoton: I see you that for storage, you introduced a `check` method in the Storage trait and for metastore you introduce an independent Check trait. I am missing the reason for the different approach?

If both works, I like the 

check method in the Storage trait.
check method in the Metastore trait.
and possibly a check_index(&self) method in the Metastore trait.
etc.

quickwit-storage/src/bundle_storage.rs
```diff
@@ -134,6 +134,14 @@ impl BundleStorageFileOffsets {
 
 #[async_trait]
 impl Storage for BundleStorage {
+    async fn check(&self) -> anyhow::Result<()> {
+        if self.bundle_filepath.exists() {
+            Ok(())
+        } else {
+            anyhow::bail!("Missing path `{}`", self.bundle_filepath.display())
+        }
```
fulmicoton, line 142: Super Nitpick:

We tend to prefer early return in the codebase. This is really 
a matter of convention and you could not know of course :).
```
if !self.bundle_filepath.exists() {
    anyhow::bail!("Missing path `{}`", self.bundle_filepath.display())
}
Ok(())
```

quickwit-storage/src/local_file_storage.rs
```diff
@@ -163,6 +163,14 @@ impl From<PathBuf> for LocalFileStorage {
 
 #[async_trait]
 impl Storage for LocalFileStorage {
+    async fn check(&self) -> anyhow::Result<()> {
+        if self.root.exists() {
```
fulmicoton, line 167: Same as above.

quickwit-storage/src/bundle_storage.rs
```diff
@@ -134,6 +134,14 @@ impl BundleStorageFileOffsets {
 
 #[async_trait]
 impl Storage for BundleStorage {
+    async fn check(&self) -> anyhow::Result<()> {
+        if self.bundle_filepath.exists() {
```
fulmicoton, line 138: The check is incorrect I believe. 
We want the path to exist... in the `Storage`.

Which makes me realize our usage of the Path struct is probably an abuse...

quickwit-cli/src/lib.rs
```diff
@@ -278,14 +278,20 @@ pub async fn index_data_cli(args: IndexDataArgs) -> anyhow::Result<()> {
...
-        create_source_config_from_args(source_config_path_opt, input_path_opt).await?;
+    let mut source_config_res =
+        create_source_config_from_args(source_config_path_opt, input_path_opt).await;
     let indexing_directory_path = args.data_dir_path.join(args.index_id.as_str());
     let indexing_directory = IndexingDirectory::create_in_dir(indexing_directory_path).await?;
     let storage_uri_resolver = quickwit_storage_uri_resolver();
     let metastore_uri_resolver = MetastoreUriResolver::default();
     let metastore = metastore_uri_resolver.resolve(&args.metastore_uri).await?;
+    do_checks([
+        ("config", checked(&mut source_config_res)),
```
fulmicoton, line 289:
```suggestion
        ("source configuration", checked(&mut source_config_res)),
```

quickwit-cli/src/lib.rs
```diff
@@ -278,14 +278,20 @@ pub async fn index_data_cli(args: IndexDataArgs) -> anyhow::Result<()> {
...
+        ("metastore", metastore.check()),
+        ("index", metastore.check_index(&args.index_id)),
```
fulmicoton, line 291: Cool so you spotted how there was a specific need to check further with `check_index`

fulmicoton (changes requested):
See comments!

quickwit-metastore/src/lib.rs
```diff
@@ -59,3 +63,68 @@ pub use metastore::{
...
+        Ok(())
+    }
+}
+
+/// Decorate a result as a check (for use in [`do_checks`]).
+///
+/// This takes out the original error of the `Result` (if any), replacing it with "taken".
+/// This measure is unfortunately required by the fact that `anyhow::Error` is not `Clone`.
+/// As the checks should bail anyway, this should not be a problem.
+pub fn checked<T>(r: &mut anyhow::Result<T>) -> Pin<Box<(dyn Future<Output = anyhow::Result<()>> + Send)>> {
```
fulmicoton, line 87: A lot of the extra complexity we are dealing with here is coming from the fact that your check function takes Futures.
Maybe it could just take resolved `anyhow::Result<()>` directly?

The checked function would become irrelevant.
Granted: the CLI display experience will not exactly be as nice.

What do you think @fmassot  @llogiq ?

quickwit-metastore/src/metastore/mod.rs
llogiq: Exactly. On the other hand, perhaps that additional info is useful, so I'll use your suggestion instead.

quickwit-metastore/src/lib.rs
llogiq: I got a type error when trying to use `'static` here because it appears to get casted away by the async method lowering. I'll try using a Vec.

quickwit-storage/src/bundle_storage.rs
llogiq: What would the right check be there? Perhaps `self.metadata.files.iter().all(|(file, _offset)| file.exists())` to make sure all the referenced files are alive?

quickwit-cli/src/lib.rs
llogiq: Yeah, the original issue had an index check, but since the index is now part of the metastore, I added this method.

quickwit-metastore/src/lib.rs
llogiq: The idea behind `do_checks` is that we can run multiple futures concurrently, which will return results faster when e.g. waiting for a database or S3 instance. So we have this one (admittedly not so small) piece of complexity, but the API is workable (though I might add some more docs) and makes calling it fairly painless.

llogiq:
@fulmicoton I initially had the `Check` trait because I thought I'd be using trait objects, but I found that not to work well, so the trait can be removed, the function moving to the various other traits or types.

quickwit-metastore/src/lib.rs
fmassot: I like the async checks for the reason mentioned by @llogiq.
But... I found that the complexity is a consequence of mixing connectivity checks with parsing checks. Here the source config can be bad only if the source file is badly written. We have the same thing for index config.

What I would do here is checking only connectivity stuff in `do_checks` and leave the parsing checking as it is today. (we could also do a parsing report but I think the issue was not opened for that reason).

```
let source_config =
        create_source_config_from_args(source_config_path_opt, input_path_opt).await?;
do_checks([
            ("source", source_check),
            ("metastore", metastore.check()),
            ("index", metastore.check_index(&args.index_id)),
        ])
        .await?;
```

llogiq: Yes, that makes sense. This would let us remove the `checked` method as well as some rather complex code in `quickwit-cli/src/lib.rs`.

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -305,6 +305,11 @@ fn previous_position_for_offset(offset: i64) -> Position {
     }
 }
 
+/// Checks if connecting with the given parameters works.
+pub(super) fn check(client_params: serde_json::Value) -> anyhow::Result<()> {
+    create_consumer(None, client_params).map(drop)
```
fulmicoton, line 310: My understanding is that your intent is `.map(|_| ())`.

I think the most readable code for that is simply:

```suggestion
    create_consumer(None, client_params)?;
    Ok(())
```

quickwit-indexing/src/source/mod.rs
```diff
@@ -198,3 +198,25 @@ pub struct SourceConfig {
...
+                    if !std::path::Path::new(path).exists() {
+                        anyhow::bail!("Source `{}` does not exist", path)
+                    }
+                }
+                Ok(())
+            }
+            #[cfg(feature = "kafka")]
+            "kafka" => kafka_source::check(self.params.clone()),
+            "vec" => Ok(()),
+            x => {
```
fulmicoton, line 216: rename x for something more suggestive like `unrecognized_source_type`
Generally speaking we try to avoid 1 letter variable names.

quickwit-metastore/src/lib.rs
fulmicoton: @fmassot the object type is added by the calling function.

quickwit-storage/src/bundle_storage.rs
fulmicoton: The bundle storage is  a bit complicated. It is a directory that is encoded in a single file. 
The file itself is in a storage. In other words, bundle_filepath is not something that makes sense on the local filepath.
It is a Path meant to be used with self.storage.

A good low cost fix would be to call
`self.storage.exists(&self.bundle_filepath).await`

It is not an exhaustive check but that will be sufficient. That function will most likely never be called actually.

llogiq:
I've incorporated @fulmicoton 's suggested changes and rebased onto main.

quickwit-storage/src/bundle_storage.rs
llogiq: Ok, done.

quickwit-metastore/src/lib.rs
llogiq: Because the checks may wait on certain things like locks, database connections, etc., which is a good use of the async runtime we already have. We might even make it multithreaded, but I don't think this will win us much.

quickwit-cli/Cargo.toml
```diff
@@ -14,6 +14,7 @@ name = "quickwit"
 path = "src/main.rs"
 
 [dependencies]
+async-trait = "0.1.51"
```
guilload, line 17: nit: we tend to not pin a package up to the minor version: `async-trait = '0.1'`

quickwit-cli/Cargo.toml
```diff
@@ -46,6 +47,7 @@ humansize = "1.1.1"
 openssl-probe = { version = "0.1.4", optional = true }
 itertools = "0.10.1"
 colored = "2.0.0"
+futures = "0.3.17"
```
guilload, line 50: ditto

quickwit-indexing/src/source/mod.rs
```diff
@@ -198,3 +198,28 @@ pub struct SourceConfig {
...
+                    if !std::path::Path::new(path).exists() {
+                        anyhow::bail!("Source `{}` does not exist", path)
+                    }
+                }
+                Ok(())
+            }
+            #[cfg(feature = "kafka")]
+            "kafka" => kafka_source::check(self.params.clone()),
+            "vec" => Ok(()),
+            unrecognized_source_type => {
```
guilload, line 216: I'd return an error here: `Source type `{}` is not supported`. We'll move this check to `IndexConfig::validate` eventually (#757).

quickwit-metastore/Cargo.toml
```diff
@@ -16,12 +16,13 @@ async-trait = "0.1"
 chrono = "0.4"
 diesel = { version = "1.4", features = ["postgres", "chrono", "extras"], optional = true }
 diesel_migrations =  { version = "1.4", optional = true }
+futures = "0.3.17"
+openssl = { version = "0.10.36", optional = true }
```
guilload, line 20: My man, I like ordering my imports too :)

quickwit-metastore/src/metastore/mod.rs
```diff
@@ -194,6 +194,15 @@ pub struct MetadataSet {
 #[cfg_attr(any(test, feature = "testsuite"), mockall::automock)]
 #[async_trait]
 pub trait Metastore: Send + Sync + 'static {
+    /// Checks if the metastore is available.
+    async fn check(&self) -> anyhow::Result<()>;
+
+    /// Checks if the given index is in this metastore.
+    async fn check_index(&self, index_id: &str) -> anyhow::Result<()> {
```
guilload, line 201: Could we change this method to `index_exists(...) -> anyhow::Result<bool> ` instead and use it for the connectivity checks?

quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -454,6 +454,11 @@ async fn download_all(byte_stream: &mut ByteStream, output: &mut Vec<u8>) -> io:
 
 #[async_trait]
 impl Storage for S3CompatibleObjectStorage {
+    async fn check(&self) -> anyhow::Result<()> {
+        self.s3_client.list_buckets().await?;
```
guilload, line 458: `list_buckets` is not a great candidate for a connectivity check. Quickwit does not require the `ListAllBuckets` permission to work correctly and most large organizations won't give applications the right to list all their buckets.

We could go with `list_objects` limited to 1 result or `delete_object` with a very random key.

guilload (commented):
Looking good overall.

My only grief with this PR is the naming. IMO, `check` is very vague. The range of possible checks is wide: from checking the validity of arguments and configs to checking the integrity of the index data. So, I'd go with something like  `check_connectivity`, which conveys the intent more precisely.

quickwit-indexing/src/source/mod.rs
llogiq: My idea here was that the checks are done on a best effort basis to make the program fail fast, and are not relied upon for the further running of the application. So making this an error will require everyone who writes a new source type to extend that function immediately to be able to run their code. On the other hand, making this a warning means it can still be searched in the code, but the programmer writing the extension still have the decision when to tackle it.

quickwit-metastore/Cargo.toml
llogiq: :smiley:

quickwit-storage/src/object_storage/s3_compatible_storage.rs
llogiq: Ok, that makes sense.

llogiq: I changed it to `list_objects` with a `max_keys: Some(1)`.

quickwit-metastore/src/metastore/mod.rs
llogiq: Probably, but that's outside the scope of this PR. It would require specialized implementation for all metastore implementations. Also what would `Ok(false)` mean? That the metastore is available but the index isn't?

fulmicoton (changes requested):

fulmicoton (changes requested):
See comments

quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -454,6 +454,17 @@ async fn download_all(byte_stream: &mut ByteStream, output: &mut Vec<u8>) -> io:
 
 #[async_trait]
 impl Storage for S3CompatibleObjectStorage {
+    async fn check(&self) -> anyhow::Result<()> {
+        self.s3_client
+            .list_objects(ListObjectsRequest {
```
guilload, line 459: Sorry, I should have been more precise and asked you to use `list_objects_v2`.

quickwit-metastore/src/metastore/mod.rs
guilload: Yeah, `Ok(false)` would mean that the index does not exist.

quickwit-storage/src/object_storage/s3_compatible_storage.rs
llogiq: No problem! I updated the code.

llogiq:
rebased. @fulmicoton is there anything stopping you from merging this?

fmassot (approved):
Now we're good, I fixed the localstack tests, we only pass them for coverage, you could not have known that :)

PR #750, opened 2021-11-10
Updated split timestamps fields

evanxg852000:
Closes https://github.com/quickwit-inc/quickwit/issues/665

- replaced the field `update_timestamp` with more conventional fields `created_at`, `updated_at`
- use diesel auto update feature for postgress metastore on `updated_at` field

### How was this PR tested?
Manually tested with pg/file metastore


quickwit-metastore/src/metastore/mod.rs
```diff
@@ -85,8 +85,11 @@ pub struct SplitMetadata {
     /// The state of the split. This is the only mutable attribute of the split.
     pub split_state: SplitState,
 
-    /// Timestamp for tracking when the split state was last modified.
-    pub update_timestamp: i64,
+    /// Timestamp for tracking when the split was created.
+    pub created_at: i64,
+
+    /// Timestamp for tracking when the split was updated.
```
guilload, line 91:
```suggestion
    /// Timestamp for tracking when the split was last updated.
```

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -450,7 +442,14 @@ impl PostgresqlMetastore {
             let split_metadata_and_footer_offsets =
                 match model_split.make_split_metadata_and_footer_offsets() {
                     Ok(mut metadata) => {
-                        metadata.split_metadata.update_timestamp = model_split.update_timestamp;
+                        metadata.split_metadata.created_at = model_split
+                            .created_at
+                            .unwrap_or(NaiveDateTime::from_timestamp(0, 0))
```
guilload, line 447: The column is nullable?

quickwit-metastore/src/postgresql/schema.rs
```diff
@@ -30,7 +30,8 @@ table! {
         split_state -> Varchar,
         time_range_start -> Nullable<Int8>,
         time_range_end -> Nullable<Int8>,
-        update_timestamp -> Int8,
+        created_at -> Nullable<Timestamp>,
```
guilload, line 33: Here is my answer. Why are those timestamps nullable? Backward-compatibility?

quickwit-metastore/src/postgresql/model.rs
```diff
@@ -82,8 +83,10 @@ pub struct Split {
     pub time_range_start: Option<i64>,
     /// If a timestamp field is available, the max timestamp in the split.
     pub time_range_end: Option<i64>,
-    /// Timestamp for tracking when the split state was last modified.
-    pub update_timestamp: i64,
+    /// Timestamp for tracking when the split was created.
+    pub created_at: Option<NaiveDateTime>,
+    /// Timestamp for tracking when the split was updated.
```
guilload, line 88:
```suggestion
    /// Timestamp for tracking when the split was last updated.
```

quickwit-metastore/src/postgresql/schema.rs
evanxg852000: This is to allow Diesel do it's magic, by passing None during create/update, diesel does the auto update. otherwise the value you pass will always take precedence.

quickwit-metastore/src/metastore/postgresql_metastore.rs
evanxg852000: Yes

guilload:
Looking good overall.

@fmassot, @fulmicoton, I'd like your opinion on two things:
1. Naming: are you happy with `created_at` and `updated_at`? We can also consider `create_timestamp` and `update_timestamp`.
2. Chosen Postgres type for timestamps: we went with timestamp without time zone (but generated at UTC). We can also consider timestamp with time zone, which is more user-friendly. When querying the DB directly, the timestamps will be displayed in the local time, which might be helpful for forensics.

fmassot:
It seems a little bit a matter of taste here, I like `created_at` and `updated_at` when we have dates, I would choose `create_timestamp` and `update_timestamp` for timestamps.

Concerning time zone, I think it's important to have everything in UTC at this level. I would take into account time zone when we want to display something for the user.

fulmicoton:
I like both.  `{created/update}_at` is maybe a little bit nicer.

evanxg852000:
UTC with timezone is an option we can go with.
But when I read [the difference](https://stackoverflow.com/questions/5876218/difference-between-timestamps-with-without-time-zone-in-postgresql), my take was that what we need to enforce is to always save DateTime in UTC. This is currently what we do. By storing without time zone, there is no timezone conversion applied to the field when retrieving or when the DB timezone config changes. and we are sure to retrieve what we stored (in UTC) without any db or client TZ config applied to it.

guilload:
1. I like both namings, but I tend to agree with @fmassot's point about date vs. timestamp, so let's go with `create_timestamp` and `update_timestamp`. That's also more in line with our coding style.

2. It seems like there's a consensus about timestamps without timezone. Let's just make sure that, moving forward, they are always generated at UTC for all implementations. This is properly done in this PR.

quickwit-metastore/migrations/postgresql/2021-11-10-104441_change_timestamp_fields/up.sql
```diff
@@ -0,0 +1,25 @@
...
+DROP COLUMN update_timestamp;
+
+-- create new timestamp fields
+ALTER TABLE splits
+ADD create_timestamp TIMESTAMP NOT NULL DEFAULT (CURRENT_TIMESTAMP AT TIME ZONE 'UTC');
+ALTER TABLE splits
+ADD update_timestamp TIMESTAMP NOT NULL DEFAULT (CURRENT_TIMESTAMP AT TIME ZONE 'UTC');
+
+-- replace diesel helper function to make sure it uses UTC timestamp
+CREATE OR REPLACE FUNCTION diesel_set_updated_at() RETURNS trigger AS $$
```
guilload, line 12: `diesel_set_update_timestamp`?

quickwit-metastore/src/tests.rs
```diff
@@ -1733,7 +1753,7 @@ pub mod test_suite {
 
         current_timestamp = split_meta.update_timestamp;
 
-        // wait for 1s, mark split for deletion & check `update_timestamp`
+        // wait for 1s, mark split for deletion & check `updated_at`
```
guilload, line 1756:
```suggestion
        // wait for 1s, mark split for deletion & check `update_timestamp `
```

guilload (approved):

quickwit-metastore/migrations/postgresql/2021-11-10-104441_change_timestamp_fields/up.sql
evanxg852000: Yes I was hesitant on this, because we might just change the name of all the functions diesel provides for this setup

guilload: I see. Let's not interfere with the default diesel functions and triggers then. People used to diesel might get surprises.

Let's just roll our on trigger. `quickwit_set_update_timestamp`. I'm ok with the duplication if that makes thing more explicit/transparent.

quickwit-metastore/migrations/postgresql/2021-11-10-104441_change_timestamp_fields/up.sql
```diff
@@ -0,0 +1,35 @@
...
+ALTER TABLE splits
+DROP COLUMN update_timestamp;
+
+-- create new timestamp fields
+ALTER TABLE splits
+ADD create_timestamp TIMESTAMP NOT NULL DEFAULT (CURRENT_TIMESTAMP AT TIME ZONE 'UTC');
+ALTER TABLE splits
+ADD update_timestamp TIMESTAMP NOT NULL DEFAULT (CURRENT_TIMESTAMP AT TIME ZONE 'UTC');
+
+-- Setup helper function and trigger to automatically set the `update_timestamp` field
```
guilload, line 11: Nice.

guilload (approved):
LGTM.

PR #759, opened 2021-11-11
Implement `ServerConfig`

guilload:
Implement `ServerConfig`.


quickwit-config/resources/tests/server_config/server.toml
```diff
@@ -0,0 +1,13 @@
+metastore_uri = "postgres://username:password@host:port/db"
+
+[indexer]
+data_dir_path = "/opt/quickwit/data"
```
fmassot, line 4: we can set up here `IndexingSplitStoreParams` (there 2 params)

quickwit-config/resources/tests/server_config/server.toml
```diff
@@ -0,0 +1,13 @@
...
+
+[searcher]
+data_dir_path = "/opt/quickwit/data"
+peer_seeds = [ "quickwit-searcher-0.local", "quickwit-searcher-1.local" ]
```
fmassot, line 8: no port?

quickwit-config/resources/tests/server_config/server.toml
```diff
@@ -0,0 +1,13 @@
...
+fast_field_cache_capacity = "1G"
+split_footer_cache_capacity = "500M"
+
+[storage]
+s3 = { region = "us-east-1", endpoint = "https://s3.us-east-1.amazonaws.com" }
```
fmassot, line 13: what is this s3 endpoint for?

quickwit-config/resources/tests/server_config/server.toml
guilload: Will do.

quickwit-config/resources/tests/server_config/server.toml
guilload: I guess not in this example :) We should be able to rely on default values, right?

This makes me think I need to set up the default listening ports in the server config.

quickwit-config/resources/tests/server_config/server.toml
guilload: This is the URL of the AWS S3 API endpoint. Some accounts have custom endpoints (and the region is ignored).

quickwit-config/src/server_config.rs
```diff
@@ -0,0 +1,270 @@
...
+use byte_unit::Byte;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct IndexerConfig {
+    pub data_dir_path: PathBuf,
+}
+
+impl IndexerConfig {
+    pub fn validate(&self) -> anyhow::Result<()> {
```
fmassot, line 33: maybe `check` instead of `validate` to be consistent

quickwit-config/src/server_config.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+        Ok(())
+    }
+}
+
+// TODO: caching
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct SearcherConfig {
+    pub data_dir_path: PathBuf,
+    pub peer_seeds: Vec<String>,
```
fmassot, line 48: why not `SocketAddr` instead of `String` for peer ?

quickwit-config/src/server_config.rs
```diff
@@ -0,0 +1,270 @@
...
+
+impl SearcherConfig {
+    pub fn validate(&self) -> anyhow::Result<()> {
+        if !self.data_dir_path.exists() {
+            bail!(
+                "Data dir `{}` does not exist.",
+                self.data_dir_path.display()
+            );
+        }
+        if self.peer_seeds.is_empty() {
```
fmassot, line 61: mmm, this is ok to not have peer seeds.

quickwit-config/resources/tests/server_config/server.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[indexer]
+data_dir_path = "/opt/quickwit/data"
+split_store_max_num_bytes = "100G"
+split_store_max_num_files = 1_000
+
+[searcher]
+data_dir_path = "/opt/quickwit/data"
+peer_seeds = [ "quickwit-searcher-0.local", "quickwit-searcher-1.local" ]
+fast_field_cache_capacity = "1G"
```
fulmicoton, line 11: there is also a `split_footer_cache_capacity`. We may call it `hotcache_cache_capacity` or `split_turbo_cache_capacity` any cool name you can think of :)

guilload: it's right below :D

fulmicoton:
This is awesome work!

quickwit-config/src/server_config.rs
guilload: `SocketAddr` only accepts IPs but we want to be able to provide hostnames.

quickwit-config/src/server_config.rs
guilload: I find that `validate` comes up more frequently when it comes to validating configs or forms. Don't you agree?

fmassot (approved):

PR #765, opened 2021-11-12
Fix GC bug

evanxg852000:
### Description
Now returning error when one of the split cannot be deleted  

Closes https://github.com/quickwit-inc/quickwit/issues/604

### How was this PR tested?
Unit test & cli commands


evanxg852000:
@guilload  I am thinking of getting rid of `SplitDeletionStats` since we only have now: delete all or nothing. Alternatively, we can have an error that contains `SplitDeletionStats` if we want to propagate splits that were deleted.  
What's your take?

guilload:
Yes, I'm in favor of getting rid of `SplitDeletionStats.` Do we currently have a use case for propagating the IDs of the splits that were deleted? Since it's pretty easy to implement, I'd do it only if the answer to that question is yes.

quickwit-indexing/src/garbage_collection.rs
```diff
@@ -189,12 +189,19 @@ pub async fn delete_splits_with_files(
...
-        warn!(index_id = ?index_id, split_ids = ?failed_split_ids, "Failed to delete splits.");
-    }
+
     if !deleted_split_ids.is_empty() {
         let split_ids: Vec<&str> = deleted_split_ids.iter().map(String::as_str).collect();
         metastore.delete_splits(index_id, &split_ids).await?;
     }
+
+    if !failed_split_ids.is_empty() {
+        warn!(index_id = ?index_id, split_ids = ?failed_split_ids, "Failed to delete splits.");
```
guilload, line 199: Shouldn't this be `error!`?

quickwit-indexing/src/garbage_collection.rs
```diff
@@ -189,12 +189,19 @@ pub async fn delete_splits_with_files(
...
+        return Err(anyhow::anyhow!(
```
guilload, line 200: It would be nice to wrap and return at least of the original errors. WDYT?

quickwit-indexing/src/garbage_collection.rs
fulmicoton: +1

fmassot (approved):
LGTM

PR #772, opened 2021-11-14
Add new states to actors.

fmassot:
Two new states are now replacing the `Running` state:
- `Processing`: it corresponds to an actor that is processing messages, it has at least one message
- `Idle`: when an actor has no more message to process, it goes in `Idle` state.

## Purpose
It will be useful to define the right condition to terminate the indexing pipeline.

### DIFF
--- a/quickwit-actors/src/actor_state.rs
+++ b/quickwit-actors/src/actor_state.rs
@@ -22,17 +22,23 @@ use std::sync::atomic::{AtomicU32, Ordering};
 #[repr(u32)]
 #[derive(Clone, Debug, Copy, Eq, PartialEq)]
 pub enum ActorState {
-    Running = 0,
-    Paused = 1,
-    Exit = 2,
+    /// Processing implies that the actor has some message(s) to process.
+    Processing = 0,
+    /// Idle means that the actor is currently waiting for messages.
+    Idle = 1,
+    /// Pause means that the actor processes no message but can process commands.
+    Paused = 2,
+    /// Exit means that the actor exited and cannot return to any other states.
+    Exit = 3,
 }
 
 impl From<u32> for ActorState {
     fn from(actor_state_u32: u32) -> Self {
         match actor_state_u32 {
-            0 => ActorState::Running,
-            1 => ActorState::Paused,
-            2 => ActorState::Exit,
+            0 => ActorState::Processing,
+            1 => ActorState::Idle,
+            2 => ActorState::Paused,
+            3 => ActorState::Exit,
             _ => {
                 panic!(
                     "Found forbidden u32 value for ActorState `{}`. This should never happen.",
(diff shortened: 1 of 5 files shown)


quickwit-actors/src/async_actor.rs
```diff
@@ -131,13 +130,17 @@ async fn process_msg<A: Actor + AsyncActor>(
     }
 
     match command_or_msg_recv_res {
-        Ok(CommandOrMessage::Command(cmd)) => process_command(actor, cmd, ctx, state_tx),
+        Ok(CommandOrMessage::Command(cmd)) => {
+            ctx.start_processing();
+            process_command(actor, cmd, ctx, state_tx)
+        }
         Ok(CommandOrMessage::Message(msg)) => {
             let span = actor.message_span(msg_id, &msg);
```
fulmicoton, line 138: if your first message is message then I believe this is wrong.

quickwit-actors/src/sync_actor.rs
```diff
@@ -114,13 +113,17 @@ fn process_msg<A: Actor + SyncActor>(
         return Some(ActorExitStatus::Killed);
     }
     match command_or_msg_recv_res {
-        Ok(CommandOrMessage::Command(cmd)) => process_command(actor, cmd, ctx, state_tx),
+        Ok(CommandOrMessage::Command(cmd)) => {
+            ctx.start_processing();
```
fulmicoton, line 117: same as for async_actor.

quickwit-actors/src/actor_state.rs
```diff
@@ -49,28 +51,54 @@ impl From<ActorState> for AtomicState {
...
+        *self == ActorState::Idle || *self == ActorState::Processing
+    }
+}
+
 pub struct AtomicState(AtomicU32);
 
 impl Default for AtomicState {
     fn default() -> Self {
-        AtomicState(AtomicU32::new(ActorState::Running as u32))
+        AtomicState(AtomicU32::new(ActorState::Processing as u32))
```
fulmicoton, line 64: Yes exactly!

quickwit-actors/src/actor_state.rs
```diff
@@ -22,17 +22,19 @@ use std::sync::atomic::{AtomicU32, Ordering};
 #[repr(u32)]
 #[derive(Clone, Debug, Copy, Eq, PartialEq)]
 pub enum ActorState {
-    Running = 0,
-    Paused = 1,
-    Exit = 2,
+    Processing = 0,
```
fulmicoton, line 25:
```suggestion
    /// Processing implies that the actor has some message to process.
    Processing = 0,
    /// Idle means that the actor is currently waiting for messages.
```

fulmicoton (changes requested):
I think you need to set running in more places.

quickwit-actors/src/async_actor.rs
fmassot: oh yes totally wrong, I was not really awake when I committed this.

fulmicoton (approved):

PR #779, opened 2021-11-15
Added CODE_STYLE.md

fulmicoton:
First stab at a code style document.
It is not in marble.

The idea is for us to edit incrementally as we encounter
interesting point in code reviews.


PSeitz:
I think this is a very good guide. We can consider a block about minimal apis

e.g. prefer `&[u8]` over `&Vec<u8>`

fulmicoton:
Absuing my power to merge this.

PR #789, opened 2021-11-17
remove split_state from split_metadata

### LINKED ISSUE #698: Inconsistent split state in postgresql
We save twice the split state information: in a dedicated column and also in a text field (json).

Here is the output of a select on `splits` table:

```
split_state | split_metadata_json
Published   | {"split_metadata": {..."split_state":"Staged",...} }
```

It seems to be the case for many splits.

PSeitz:
Fixes #698

### Description

### How was this PR tested?

### DIFF
--- a/quickwit-cli/src/main.rs
+++ b/quickwit-cli/src/main.rs
@@ -20,7 +20,7 @@
 use std::env;
 use std::fmt::Debug;
 use std::net::SocketAddr;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::time::Duration;
 
 use anyhow::{bail, Context};
@@ -265,8 +265,7 @@ impl CliCommand {
             .context("'host-key-path-prefix' has a default  value")?
             .to_string();
         let host_key_path =
-            Path::new(format!("{}-{}-{}", host_key_path_prefix, host, port.to_string()).as_str())
-                .to_path_buf();
+            PathBuf::from(format!("{}-{}-{}", host_key_path_prefix, host, port).as_str());
         let mut peer_socket_addrs: Vec<SocketAddr> = Vec::new();
         if matches.is_present("peer-seed") {
             if let Some(values) = matches.values_of("peer-seed") {
--- a/quickwit-indexing/failpoints/mod.rs
+++ b/quickwit-indexing/failpoints/mod.rs
@@ -205,14 +205,14 @@ async fn aux_test_failpoints() -> anyhow::Result<()> {
     let mut splits = metastore
         .list_splits("test-index", SplitState::Published, None, &[])
         .await?;
-    splits.sort_by_key(|split| *split.split_metadata.time_range.clone().unwrap().start());
+    splits.sort_by_key(|split| *split.time_range.clone().unwrap().start());
     assert_eq!(splits.len(), 2);
     assert_eq!(
-        splits[0].split_metadata.time_range.clone().unwrap(),
+        splits[0].time_range.clone().unwrap(),
         1629889530..=1629889531
     );
     assert_eq!(
-        splits[1].split_metadata.time_range.clone().unwrap(),
+        splits[1].time_range.clone().unwrap(),
         1629889532..=1629889533
     );
     Ok(())
@@ -257,11 +257,8 @@ async fn test_merge_executor_controlled_directory_kill_switch() -> anyhow::Resul
     }
 
     let metastore = test_index_builder.metastore();
-    let splits_with_footer_offsets = metastore.list_all_splits(index_id).await?;
-    let splits: Vec<SplitMetadata> = splits_with_footer_offsets
-        .into_iter()
-        .map(|split_and_footer_offsets| split_and_footer_offsets.split_metadata)
-        .collect();
+    let split_infos = metastore.list_all_splits(index_id).await?;
+    let splits: Vec<SplitMetadata> = split_infos.into_iter().collect();
     let merge_scratch_directory = ScratchDirectory::for_test()?;
 
     let downloaded_splits_directory =
@@ -269,7 +266,7 @@ async fn test_merge_executor_controlled_directory_kill_switch() -> anyhow::Resul
     let storage = test_index_builder.index_storage(index_id)?;
     let mut tantivy_dirs: Vec<Box<dyn Directory>> = vec![];
     for split in &splits {
-        let split_filename = split_file(&split.split_id);
+        let split_filename = split_file(split.split_id());
         let dest_filepath = downloaded_splits_directory.path().join(&split_filename);
         storage
             .copy_to_file(Path::new(&split_filename), &dest_filepath)
--- a/quickwit-metastore/src/metastore/mod.rs
+++ b/quickwit-metastore/src/metastore/mod.rs
@@ -30,7 +30,7 @@ use quickwit_index_config::IndexConfig;
 use serde::{Deserialize, Serialize};
 
 use crate::checkpoint::{Checkpoint, CheckpointDelta};
-use crate::{MetastoreResult, SplitMetadataAndFooterOffsets, SplitState};
+use crate::{MetastoreResult, SplitInfo, SplitMetadata, SplitState};
 
 /// An index metadata carries all meta data about an index.
 #[derive(Clone, Debug, Serialize, Deserialize)]
@@ -108,7 +108,7 @@ pub trait Metastore: Send + Sync + 'static {
     async fn stage_split(
         &self,
         index_id: &str,
-        split_metadata: SplitMetadataAndFooterOffsets,
+        split_metadata: SplitMetadata,
     ) -> MetastoreResult<()>;
 
     /// Publishes a list splits.
@@ -143,14 +143,11 @@ pub trait Metastore: Send + Sync + 'static {
         split_state: SplitState,
         time_range: Option<Range<i64>>,
         tags: &[String],
-    ) -> MetastoreResult<Vec<SplitMetadataAndFooterOffsets>>;
+    ) -> MetastoreResult<Vec<SplitInfo>>;
 
     /// Lists the splits without filtering.
     /// Returns a list of all splits currently known to the metastore regardless of their state.
-    async fn list_all_splits(
-        &self,
-        index_id: &str,
-    ) -> MetastoreResult<Vec<SplitMetadataAndFooterOffsets>>;
+    async fn list_all_splits(&self, index_id: &str) -> MetastoreResult<Vec<SplitInfo>>;
 
     /// Marks a list of splits for deletion.
     /// This API will change the state to `ScheduledForDeletion` so that it is not referenced by the
--- a/quickwit-metastore/src/metastore/single_file_metastore.rs
+++ b/quickwit-metastore/src/metastore/single_file_metastore.rs
@@ -35,7 +35,7 @@ use crate::checkpoint::CheckpointDelta;
 use crate::metastore::match_tags_filter;
 use crate::{
     IndexMetadata, Metastore, MetastoreError, MetastoreFactory, MetastoreResolverError,
-    MetastoreResult, SplitMetadata, SplitMetadataAndFooterOffsets, SplitState,
+    MetastoreResult, SplitInfo, SplitMetadata, SplitState,
 };
 
 /// Metadata file managed by [`SingleFileMetastore`].
@@ -47,7 +47,7 @@ pub struct MetadataSet {
     /// Metadata specific to the index.
     pub index: IndexMetadata,
     /// List of splits belonging to the index.
-    pub splits: HashMap<String, SplitMetadataAndFooterOffsets>,
+    pub splits: HashMap<String, SplitInfo>,
 }
 
 /// Creates a path to the metadata file from the given index ID.
@@ -203,15 +203,15 @@ impl InnerSingleFileMetastore {
                 }
             };
 
-            match metadata.split_metadata.split_state {
+            match metadata.split_state {
                 SplitState::Published => {
                     // Split is already published. This is fine, we just skip it.
                     continue;
                 }
                 SplitState::Staged => {
                     // The split state needs to be updated.
-                    metadata.split_metadata.split_state = SplitState::Published;
-                    metadata.split_metadata.update_timestamp = Utc::now().timestamp();
+                    metadata.split_state = SplitState::Published;
+                    metadata.update_timestamp = Utc::now().timestamp();
                 }
                 _ => {
                     split_not_staged_ids.push(split_id.to_string());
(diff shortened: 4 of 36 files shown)


quickwit-metastore/src/metastore/single_file_metastore.rs
```diff
@@ -28,16 +28,25 @@ use quickwit_storage::{
...
+    MetastoreResult, SplitMetadata, SplitMetadataAndFooterOffsets, SplitState,
 };
 
+/// A MetadataSet carries an index metadata and its split metadata.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MetadataSet {
+    /// Metadata specific to the index.
+    pub index: IndexMetadata,
+    /// List of splits belonging to the index.
+    pub splits: HashMap<String, (SplitMetadataAndFooterOffsets, SplitState)>,
```
evanxg852000, line 47: Not necessarily in this PR, but I think it's a good time to discuss.
Maybe a struct type to wrap the tuple `(SplitMetadataAndFooterOffsets, SplitState)` as we also need to pull out `create_timestamp` & `update_timestamp`

guilload (commented):
Unsure this is the best strategy for fixing the issue. @fmassot, @fulmicoton, please, weigh in.

fmassot:
I will start with my current mental model of the metastore:
- an index is defined by its splits
- in our code, an index is represented by its metadata `IndexMetadata`
- a split by `SplitMetadataAndFooterOffsets` that wraps the real metadata stuff: `SplitMetadata`. Here I immediately think that we should merge these 2 structs into 1 but that's another problem and we may have good reasons to keep both.

A split has a state, this is fundamental to understand on which split a query will be executed. I think a split metadata should explicitly contains its state because it owns it. Everything else is related to how we serialize things in a JSON file or in a column-oriented database. 

To fix the issue, I think a simple solution could be: when serializing metadata in the text column, remove info that is duplicated in other dedicated columns to avoid inconsistencies or having to update multiple columns on an update. When we change the state, it's just easier to say "update only this column".

fulmicoton:
@guilload @fmassot. I suggested this approach to @pseitz.

The point is to isolate the mutable stuff from the immutable stuff.
The problem is not really specific to postgresl. It is specific to all metastore that are not the singlefilemetastore.

The split_state is also redundant (we need to double check for it) in all API but list_all_splits.

I won't have time to look at this PR today.
Can we wait 24h before making changes to it?

fulmicoton:
Another pros of isolating split_state.
We need to be backward compatible for the deserialization of SplitMetaData. 
We already messed it up 3 times so I think we need plumbing to check that we don't break it.

By introducing postgressplitmetadata we will have this constraint on 2 objects.

quickwit-metastore/src/metastore/single_file_metastore.rs
fulmicoton: I think we might want to pull out update_timestamp. (create_timestamp too if it helps)

fulmicoton: I'd rather have a struct than a tuple here.

quickwit-metastore/src/metastore/mod.rs
```diff
@@ -60,6 +60,27 @@ pub struct SplitMetadataAndFooterOffsets {
 }
 
 /// A split metadata carries all meta data about a split.
+///
+/// Be careful about changes in this struct. It is serialized as JSON into the database, so some
```
fulmicoton, line 64: Thank you for this comment!

fulmicoton (changes requested):
I think we should fix the format of the MetadataSet in the single file metastore, and consider extracting update timestamp (and possibly create_timestamp)

PSeitz:
> To fix the issue, I think a simple solution could be: when serializing metadata in the text column, remove info that is duplicated in other dedicated columns to avoid inconsistencies or having to update multiple columns on an update. When we change the state, it's just easier to say "update only this column".

@guilload @fmassot  @fulmicoton 

Nulling duplicated information in the JSON, or having the splitting of the metadata in postregsql was my initial proposal on the discussion with Paul.
On the other side this issue may affect multiple metastores and splitting the metadata in all of those stores could cause a lot of code duplication. 
Splitting the data in immutable and mutable can be valuable for those stores. But this not a good approach when thinking about separation of concerns, since we change our structure in the code because of the metastores ability (or inability :) to mutate JSON. 

Simply nulling mutable information in the serialized JSON may not have such a serious downside. What do you think?

fulmicoton:
I think the approach in the PR is the right one.

But there are a lot of different points there.

- SplitMetadataAndFooterOffsets vs SplitMetadata... We do we have two of these. We only know the offsets after creating the hotcache. An alternative would be to have the offset in the split metadata object and keep them uninitialized.
The two approach have both pros and cons. I like the current approach as static typing helps us avoid a bunch of mistakes.
If the serialization format feels clumsy (I personally think it is fine) we can use serde(flatten).

- Why not fix the issue on the postgresql side only: We want backward compatibility. We judged (rightfully I believe) that one good strategy to manage backward compatibility was to have some nice logic for a complex JSON object: `SplitMetadataAndFooterOffsets`, and have the different store serialize this object directly.
Concretely: this means we need to have logic to deserialize any version of that object and serialize only the last version.
Introducing a PostgresqlSplitmetadata would ruin this plan.

- The format of serialization sucks:
I agree the pair tuple in the SingleFileMetastore is a bad idea we need to fix that.

- Extract more mutable state: I think we need to extract more mutable state. (update_time?). 

- But does that mean we want to remove the state from the API response?
We need to return the state one way or another for list_all_splits. 
For the write calls, the split_state is just weird.
For list_split it is useless, but I agree the redundancy might be a good thing here.

```rust
// This struct is only built by metastores
struct SplitInfo {
  split_metadata: SplitMetadataAndOffsets,
  split_state,
  update_time,
}

struct SplitMetadataAndOffsets {
   //...
   // like today
}

struct SplitMetadata {
   // ... 
   // like today but without split state and update_time
}
```

What do you think?

PSeitz:
Currently update_time is created in the quickwit layer, but I think this is overwritten in the metastore layer anyhow.

From a high level viewpoint what we want to do is very common. We want to push some data and then update some parts of that data (update_timestamp, split_state). One possible way to do it is to push the object, calculate the delta and apply a list of `ChangeOperation` on a split_id, e.g. `ChangeOperation {field: split_state, value: Published }` (typing may be hard). 
But for delta operations you need to read the original object to compare.

Generally I'm fine with adding `SplitInfo`, but from an readability or API standpoint it's not ideal to save `SplitMetadata` and get `SplitInfo`.

I think there are also no major points against nulling mutable fields? Except since they are part of the Serialization they are part of the versioning.

**Versioning**

I think storing the version is good, but I think we won't necessarily use it in serialization? E.g. a rename could be covered by a serde_attribute; we would not need a custom serializer that reads the version first. 

I was hoping that there would be a solid versioning crate around serde, but I couldn't find anything.

fulmicoton:
See #804 for a solution of versionning.

guilload:
@fulmicoton, let me iterate on your approach to get rid of `SplitMetadataAndOffsets`:

```rust
struct  SplitMetadataBuilder {
  // Same fields as usual minus `split_state`, `create_timestamp`, `update_timestamp`, `footer_offsets`
  [...]
}

impl SplitMetadataBuilder {
  fn tags(&mut self, ...) -> Self;
  fn time_range(&mut self, ...) -> Self;
  [...]
  fn build(self, footer_offsets: Range<u64>) -> SplitMetadata {
    SplitMetadata {
      footer_offsets,
      [...]
  }
}

struct SplitMetadata {
  // Same fields as usual minus `split_state`, `update_timestamp`
  create_timestamp: i64,
  footer_offsets: Range<u64>,
  [...]
}

async fn stage_split(
        &self,
        index_id: &str,
        split_metadata: SplitMetadata,
    ) -> MetastoreResult<()>;
```

guilload:
This is less DRY (repeated attributes between `SplitMetadataBuilder` and `SplitMetadata`). However, the builder pattern is generally well understood.

guilload:
Also, do we want to flatten `SplitInfo`? Might be a little nicer to use.

Read path:
```rust
struct Split {
  split_id: String,
  split_state: SplitState,
  tags: BTreeSet<String>,
  time_range: Option<RangeInclusive<i64>>,
  create_timestamp: i64,
  update_timestamp: i64,
  footer_offsets: Range<u64>,
  [...]
}

async fn list_splits(
        &self,
        index_id: &str,
        split_state: SplitState,
        time_range: Option<Range<i64>>,
        tags: &[String],
    ) -> MetastoreResult<Vec<Split>>;

async fn list_all_splits(
        &self,
        index_id: &str,
    ) -> MetastoreResult<Vec<Split>>;
```

guilload:
To be clear, I'm not proposing to serialize and store a flat `Split`but rather to deserialize `SplitMetadata` into a flat struct.

guilload:
We haven't touched on the issue of duplicated values stored in `SplitMetadata` and dedicated columns in SQL metastores (`tags`, `time_range`). From a functional perspective, this is fine. However, users are inevitably going to be confused and ask questions about those. It'd be ideal to not serialize those fields in SQL implementations.

Conceptually:
```
`SplitMetadata` -> `serde_json::Value` -> delete duplicated fields -> store
load -> `serde_json::Value` -> populate missing fields -> `SplitMetadata`
```

-> [custom serialization](https://serde.rs/custom-serialization.html)

guilload:
Actually, reading the code in `quickwit-indexing`, we already have "builders" with `IndexedSplit` and `PackagedSplit`. A `PackagedSplit` gives birth directly to `SplitMetadataAndOffsets`. We should be able to go directly from `PackagedSplit` to a flat `SplitMetadata`, which should be serialized as `SplitMetadataV1`. `SplitMetadataAndOffsets` is our `SplitMetadataV0`.

fulmicoton:
@PSeitz I called @guilload  on Saturday to discuss the ticket.
Consensus is:
We want a SplitMetadata instance that would replace the current SplitMetadataAndFooterOffset.
It would contain the footer offsets in an inline fashion.

We have intermediary builder like object that contain a subset of the SplitMetadata properties.
Let's keep that. It is ok if those rely on nesting objects.

We keep the "separate mutable stuff" (right now the modification time and the split state)
separately. 

The API `list_all_splits` and `list_splits` should return this mutable state too.

We can introduce a `Split` instance for their responses that would wrap the SplitMetadata object and the mutable state.

```rust
struct Split {
   // mutable stuff.
   ...
   split_metadata: SplitMetadata
}
```

It will a bit clumsy to not have access at least to split_id at the top level of `Split`.
I suggest we either add an impl to `AsRef<SplitMetadata>`, or add a `pub fn id(&self)` function.

Putting the footer offsets and the rest of the metadata at the same level will break backward compatibility and will require
to create a new version.
#806 contains instructions on how to deal with the backward compatibility.

quickwit-cli/src/main.rs
```diff
@@ -265,8 +265,7 @@ impl CliCommand {
             .context("'host-key-path-prefix' has a default  value")?
             .to_string();
         let host_key_path =
-            Path::new(format!("{}-{}-{}", host_key_path_prefix, host, port.to_string()).as_str())
-                .to_path_buf();
+            Path::new(format!("{}-{}-{}", host_key_path_prefix, host, port).as_str()).to_path_buf();
```
fulmicoton, line 268:
```suggestion
            PathBuf::from(format!("{}-{}-{}", host_key_path_prefix, host, port));
```

quickwit-indexing/failpoints/mod.rs
```diff
@@ -258,18 +258,15 @@ async fn test_merge_executor_controlled_directory_kill_switch() -> anyhow::Resul
 
     let metastore = test_index_builder.metastore();
     let splits_with_footer_offsets = metastore.list_all_splits(index_id).await?;
```
fulmicoton, line 260: we should rename all footer_offsets stuff. Can you grep for them?

quickwit-metastore/src/split_metadata_version.rs
```diff
@@ -0,0 +1,215 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::split_metadata::utc_now_timestamp;
+use crate::{SplitMetadata, SplitState};
+
+/// Carries split metadata and footer offsets.
+/// This struct can deserialize older format automatically
+/// but can only serialize to the last version.
+#[derive(Clone, Eq, PartialEq, Default, Debug, Serialize, Deserialize)]
+pub struct SplitMetadataV0 {
```
fulmicoton, line 32:
```suggestion
struct SplitMetadataV0 {
```

quickwit-metastore/src/split_metadata_version.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Contains the range of bytes of the footer that needs to be downloaded
+    /// in order to open a split.
+    ///
+    /// The footer offsets
+    /// make it possible to download the footer in a single call to `.get_slice(...)`.
+    pub footer_offsets: Range<u64>,
+}
+
+#[derive(Clone, Eq, PartialEq, Default, Debug, Serialize, Deserialize)]
+pub(crate) struct SplitMetadataAndFooterV0 {
```
fulmicoton, line 116:
```rust
struct SplitMetadataAndFooterV0 {
```

fulmicoton, line 116:
```suggestion
struct SplitMetadataAndFooterV0 {
```

quickwit-metastore/src/split_metadata.rs
```diff
@@ -17,97 +17,62 @@
...
     }
 }
 
-/// A split metadata carries all meta data about a split.
-#[derive(Clone, Eq, PartialEq, Default, Debug, Serialize, Deserialize)]
+/// Carries immutable split metadata.
+/// This struct can deserialize older format automatically
+/// but can only serialize to the last version.
+#[derive(Clone, Eq, PartialEq, Default, Debug, Serialize)]
+#[serde(into = "VersionedSplitMetadataDeserializeHelper")]
```
fulmicoton, line 75: ah yes! That's better.

quickwit-metastore/src/split_metadata_version.rs
```diff
@@ -0,0 +1,215 @@
...
+use std::ops::{Range, RangeInclusive};
+
+use serde::{Deserialize, Serialize};
+
+use crate::split_metadata::utc_now_timestamp;
+use crate::{SplitMetadata, SplitState};
+
+/// Carries split metadata and footer offsets.
+/// This struct can deserialize older format automatically
+/// but can only serialize to the last version.
```
fulmicoton, line 30:
```suggestion
```

quickwit-metastore/src/split_metadata_version.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// The footer offsets
+    /// make it possible to download the footer in a single call to `.get_slice(...)`.
+    pub footer_offsets: Range<u64>,
+}
+
+#[derive(Clone, Eq, PartialEq, Default, Debug, Serialize, Deserialize)]
+pub(crate) struct SplitMetadataAndFooterV0 {
+    split_metadata: SplitMetadataV0,
+    footer_offsets: Range<u64>,
+}
```
fulmicoton, line 119: Can you reorder the types in this file to have all V0 stuff in one place and after all of the V1 stuff.

fulmicoton (changes requested):
Some minor comments inline

quickwit-metastore/src/split_metadata_version.rs
PSeitz: done

quickwit-indexing/failpoints/mod.rs
PSeitz: I think I found all places now (we also have split_id_and_footer_offsets)

quickwit-metastore/src/split_metadata_version.rs
PSeitz: pub(crate) is required

fulmicoton (approved):

PR #792, opened 2021-11-17
Fix failing build

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/761
Reason of the failure: https://github.com/alexcrichton/cmake-rs/issues/131

Added convenience make commands in the meantime. You need to install [cross](https://github.com/rust-embedded/cross)
- `make build-x86_64-unknown-linux-gnu`
-  `make build-aarch64-unknown-linux-gnu`
-  `make build-x86_64-unknown-linux-musl`

### How was this PR tested?
Run the build on CI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ IMAGE_TAGS = x86_64-unknown-linux-gnu aarch64-unknown-linux-gnu x86_64-unknown-l
 
 .PHONY: cross-images
 cross-images:
-	for tag in ${IMAGE_TAGS}; do \
+	@for tag in ${IMAGE_TAGS}; do \
 		docker build --tag quickwit/cross:$$tag --file ./build/cross-images/$$tag.dockerfile ./build/cross-images; \
 		docker push quickwit/cross:$$tag; \
 	done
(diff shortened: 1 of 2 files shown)


evanxg852000:
@fmassot when running a build with `--features release-feature-set`, you will need a CMake  >= 3.12.1 .

Makefile
```diff
@@ -42,3 +42,17 @@ cross-images:
 		docker build --tag quickwitinc/builder-$$tag --file ./build/cross-images/$$tag.dockerfile ./build/cross-images; \
 		docker push quickwitinc/builder-$$tag:latest; \
 	done
+
+
+# TODO: To be replaced by https://github.com/quickwit-inc/quickwit/issues/237
+.PHONY: build-x86_64-unknown-linux-gnu
```
guilload, line 48: I think I'd like `make build ARCH=x86_64-unknown-linux-gnu` better? The default value could be that of the hosts' arch. WDYT?

guilload (approved):

Makefile
evanxg852000: I think this is better, because I also felt someone always has to look this up and since this can be a good default `ARCH=x86_64-unknown-linux-gnu` let's go your option

fmassot (approved):

Makefile
evanxg852000: @guilload  I settled with `TARGET` instead of `ARCH` as the former is what we are actually referring to here.

guilload: Yeah, `TARGET` is great.

guilload:
One last thing, for `make fmt`, we display a nice error message if the nightly toolchain is not installed. Can we do the same for `cross`?

Makefile
```diff
@@ -42,3 +42,14 @@ cross-images:
...
 		docker push quickwit/cross:$$tag; \
 	done
+
+# TODO: to be replaced by https://github.com/quickwit-inc/quickwit/issues/237
+TARGET ?= x86_64-unknown-linux-gnu
+.PHONY: build
+build:
+	@echo "Building binary for target=${TARGET}"
+	if [ "${TARGET}" = "x86_64-unknown-linux-musl" ]; then \
+        cross build --release --features release-feature-set --target ${TARGET}; \
```
guilload, line 52: nit: can you fix the indentation?

evanxg852000: Just wondering what could be the origin of this, my local branch is fine and git says it's in sync with remote. 🤔

guilload: Maybe it's the GitHub rendering. Don't worry too much about it.

evanxg852000: It was my vscode, not Gihub

PR #796, opened 2021-11-17
Waiting for memory to be available before running leaf requests.

fulmicoton:


fmassot:
Food for thoughts.

I looked at how elasticsearch and druid are handling concurrent queries:
- elastic: for search queries, they have a dedicated thread pool with a configurable fixed size + queue size of 1000 by default
- druid: the broker (our root node) has query lanes (or queues), which makes it easy to prioritize some workloads. The concurrent query count is limited by a configured number of threads, by default it is set to `num_cores - 1`. There is the same limitation on leaf nodes.

From what I see there is no real limitation by memory usage.

Some remarks about that:
- for quickwit, we expect higher IO latency, and intuitively, I don't expect to be CPU bound with `num_cores` concurrent queries ;
- I don't think it's easy to track memory usage but in quickwit, we can use the directory abstraction. 

The missing piece for us is the distinction between a root node and a search node.

fulmicoton:
@fmassot I suggest we merge this PR as is and complete it later.
I tested it and compared against main.
main dies with a concurrency of 10.
This branch behaves as expected with a concurrency of 20 (everything is slowly but returns).
You can find the program I used here. 

https://github.com/quickwit-inc/qbench

PR #798, opened 2021-11-18
Renaming qspawn_blocking.

fulmicoton:
`quickwit_spawn_blocking` was a good candidate too.
I am a tad afraid that the difference in semantic is
not really a detail.

Contrary to spawn_blocking, with this function,
if noone await the result, no work is done.

### DIFF
--- a/quickwit-search/src/thread_pool.rs
+++ b/quickwit-search/src/thread_pool.rs
@@ -36,9 +36,9 @@ fn search_thread_pool() -> &'static rayon::ThreadPool {
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub struct Panicked;
 
-/// Function similar to tokio::spawn_blocking.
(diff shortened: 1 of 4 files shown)


quickwit-search/src/thread_pool.rs
```diff
@@ -36,9 +36,9 @@ fn search_thread_pool() -> &'static rayon::ThreadPool {
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub struct Panicked;
 
-/// Function similar to tokio::spawn_blocking.
+/// Function similar to `tokio::spawn_blocking`.
 ///
-/// Here are some difference however.
+/// Here are two important difference however.
```
guilload, line 41:
```suggestion
/// Here are two important differences, however.
```

guilload (approved):

PR #800, opened 2021-11-18
Renaming records to docs

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/534

### How was this PR tested?
unit test


quickwit-metastore/src/metastore/mod.rs
```diff
@@ -68,9 +68,9 @@ pub struct SplitMetadata {
     /// in the storage URI resolver: for instance, the Amazon S3 region.
     pub split_id: String,
 
-    /// Number of records (or documents) in the split.
+    /// Number of documents in the split.
     /// TODO make u64
-    pub num_records: usize,
+    pub num_docs: usize,
```
fulmicoton, line 73: That won't be backward compatible

fulmicoton:
This change requires to PR #804

quickwit-metastore/src/metastore/mod.rs
evanxg852000: Yes, and this is why I was mentioning migration script option [here](https://github.com/quickwit-inc/quickwit/issues/799#issuecomment-972601268). You have already been thinking of this so I will get those changes here to see.

guilload: `#[serde(alias = "num_records")]` to the rescue.

fulmicoton (approved):

PR #805, opened 2021-11-19
Test tokio console

fmassot:
See the instructions in contributing doc.

This is not mergeable as is as it relies on a local tokio console repository (see cargo.toml).

@PSeitz could you look at it? This is very rough and I did not dig into the console information, I'm sure that you're interested :)

### DIFF
--- a/quickwit-cli/Cargo.toml
+++ b/quickwit-cli/Cargo.toml
@@ -48,6 +48,8 @@ openssl-probe = { version = "0.1.4", optional = true }
 itertools = "0.10.1"
 colored = "2.0.0"
 futures = "0.3"
+console-subscriber = { path = "./../../console/console-subscriber/" }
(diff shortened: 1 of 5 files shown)


quickwit-cli/Cargo.toml
```diff
@@ -48,6 +48,8 @@ openssl-probe = { version = "0.1.4", optional = true }
 itertools = "0.10.1"
 colored = "2.0.0"
 futures = "0.3"
+console-subscriber = { path = "./../../console/console-subscriber/" }
```
guilload, line 51: This should be behind a feature flag.

fmassot: yes and the relative path is ugly too..

fulmicoton:
Fixed in #825

PR #810, opened 2021-11-22
Limit the overall amount of running leaf requests with

### LINKED ISSUE #422: Better behavior under load
Right now quickwit try to  address any query it receives right away.
If many queries come at the same time, all queries end up slower due simple queue theory and context switches.
Not only that, but because we rely a lot on loading data in RAM to run a query,  it can results in running out of memory.

We need to find a mechanism to deal with this issue.

A common solution is to have a request queue. Usually this is done on the application side but this is not very user friendly.

@r-zenine offered a solution based on limiting the number of leaf being executed at the same time.
https://github.com/quickwit-inc/quickwit/pull/416

fulmicoton:
Limit the overall amount of running leaf requests with
SPLIT_STREAM_CONCURRRENCY.

Closes #422


fmassot (commented):
So simple!

PR #813, opened 2021-11-24
Aarch64-musl build

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/715

Builds https://github.com/quickwit-inc/quickwit/actions/runs/1516584781

### How was this PR tested?
Tested on AWS arm64 based ec2 instance


build/cross-images/x86_64-unknown-linux-musl.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ekidd/rust-musl-builder:latest
+FROM evanxg852000/cross-x86_64-unknown-linux-musl:latest
```
fmassot, line 1: seems that we need to remove `evanxg852000`? Do we?

fmassot:
Wow, you did it! it's a real "tour de force" :)

build/cross-images/x86_64-unknown-linux-musl.dockerfile
evanxg852000: So the original image has not been updated to last rust version. this triggers a build error for `edition 2021`. 
I created an issue on the repo to update but it seems he is not been interested. In the first place, I made a script to build the image alongside but my fear is that they could just update the repo one day and our build will start failing. So I forked the repo and pushed the image on my dockerhub. Until he publishes a new version of the image.
The is nothing wrong with his repo from my check he seems just not to have time anymore to update the image.
https://github.com/emk/rust-musl-builder/issues/130

fulmicoton: Can we push this image somewhere on quickwit's dockerhub then? (and it's dockerfile on our github)

evanxg852000: Ok we have two images like this, I will move all of them into Quickwit repos then

fulmicoton: Yes forking sounds good to me.

build/cross-images/x86_64-unknown-linux-gnu.dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM evanxg852000/cross-x86_64-unknown-linux-gnu:latest
+FROM quickwit/cross-base:x86_64-unknown-linux-gnu
```
guilload, line 1: I see three repositories on [DockerHub](https://hub.docker.com/repositories): `cross`, `cross-base` and `cross-builder`. Can we consolidate everything in `cross` and remove the other two repositories. `Cross.toml` already references imagines `quickwit/cross`.

guilload (approved):

build/cross-images/x86_64-unknown-linux-gnu.dockerfile
evanxg852000: `cross-builder` can be safely removed it's the old one, I think it's best you remove it. 
`quickwit/cross-base` exists as an intermediate repository where we push slightly modified base images of `rust-embemded/cross` and others. I thought it's best to have them in a separate repository since we have cloned the git repos.

PR #818, opened 2021-11-26
URI normalization

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/548
 added URI normalization enhancements

### How was this PR tested?
Tested cli commands

### DIFF
--- a/quickwit-cli/src/lib.rs
+++ b/quickwit-cli/src/lib.rs
@@ -21,7 +21,7 @@ use std::collections::VecDeque;
 use std::convert::TryFrom;
 use std::fs::File;
 use std::io::{stdout, Stdout, Write};
-use std::path::PathBuf;
+use std::path::{Component, Path, PathBuf};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 use std::{env, fmt, io};
@@ -57,6 +57,9 @@ const THROUGHPUT_WINDOW_SIZE: usize = 5;
 
 /// This environment variable can be set to send telemetry events to a jaeger instance.
 pub const QUICKWIT_JAEGER_ENABLED_ENV_KEY: &str = "QUICKWIT_JAEGER_ENABLED";
+
+/// Default file protocol `file://`
+const FILE_PROTOCOL: &str = "file";
 
 #[derive(Debug, Eq, PartialEq)]
 pub struct ExtractSplitArgs {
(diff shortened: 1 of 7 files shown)


quickwit-cli/src/lib.rs
```diff
@@ -654,6 +657,77 @@ impl ThroughputCalculator {
     }
 }
 
+/// Normalizes a URI
+/// A `file://` protocol is assumed if not specified.
+/// File URIs are resolved (normalised) to absolute path.
+/// Handles special path characters like (~, ., ..)
+pub fn normalize_uri(uri: &str) -> anyhow::Result<String> {
+    let parts: Vec<&str> = uri.split("://").collect();
```
fulmicoton, line 665: nitpick: `split_once` is quite cool for this.

quickwit-cli/src/lib.rs
```diff
@@ -654,6 +657,77 @@ impl ThroughputCalculator {
...
+            let home_dir_path = home::home_dir()
+                .context("Could nto get home directory.")?
+                .to_string_lossy()
+                .to_string();
+
+            path.replace_range(0..1, &home_dir_path);
+        }
+
+        if !path.starts_with('/') {
+            path.insert(0, '/');
```
fulmicoton, line 690: for file://my_path I agree resolving from root might be a good idea. But otherwise resolving from CWD might be a better spec.

quickwit-cli/src/lib.rs
```diff
@@ -654,6 +657,77 @@ impl ThroughputCalculator {
...
+            .to_string_lossy()
+            .to_string();
+    }
+
+    Ok(format!("{}://{}", protocol, path))
+}
+
+/// Normalizes a path without checking file existence in contrary to `Path::canonicalize`.
+/// https://github.com/rust-lang/cargo/blob/fede83ccf973457de319ba6fa0e36ead454d2e20/src/cargo/util/paths.rs#L61
+fn normalize_path(path: &Path) -> PathBuf {
```
fulmicoton, line 703: What is this for?
I don't quite understand...

evanxg852000: This does the same thing as `Path::canonicalize`, that is normalise the path by converting things like `/../.` into the actual path. But the version with Path::canonicalize will check if the file exist and blow otherwise. All I want here is a king of simple path normalization, whether the path is good or not will be resolved by the metastore or whatever happens to use it. I will elaborate this in the comment.

fulmicoton: Ah got it. No need to edit the comment. I was a bit slow there.

quickwit-cli/src/main.rs
```diff
@@ -886,4 +886,24 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_normalize_uri() -> anyhow::Result<()> {
+        assert_eq!(
+            normalize_uri("home/hommer/docs/dognuts")?,
+            String::from("file:///home/hommer/docs/dognuts")
+        );
+
+        // assert_eq!(
```
fulmicoton, line 897: remove?

evanxg852000: Yes I will, I actually need more edge cases

quickwit-cli/src/lib.rs
evanxg852000: I have switched to `CWD`  given we have a CLI tool and that's what I would expect as an end user. Let's see what @guilload thinks about it.

guilload (approved):
I wonder if we should use this PR as an opportunity to introduce a `URI` type. Then, we can start with something as simple as `struct Uri(String)` and go from there. WDYT, folks?

quickwit-cli/src/lib.rs
```diff
@@ -654,6 +657,74 @@ impl ThroughputCalculator {
...
 }
 
+/// Normalizes a URI
+/// A `file://` protocol is assumed if not specified.
+/// File URIs are resolved (normalised) relative to the current working directory
+/// unless an absolute path is specified.
+/// Handles special characters like (~, ., ..)
+pub fn normalize_uri(uri: &str) -> anyhow::Result<String> {
+    let (protocol, mut path) = match uri.split_once("://") {
+        None => (FILE_PROTOCOL, String::from(uri)),
```
guilload, line 667: We tend to use `.to_string()` in the codebase.

quickwit-cli/src/lib.rs
```diff
@@ -654,6 +657,74 @@ impl ThroughputCalculator {
     }
 }
 
+/// Normalizes a URI
+/// A `file://` protocol is assumed if not specified.
+/// File URIs are resolved (normalised) relative to the current working directory
+/// unless an absolute path is specified.
+/// Handles special characters like (~, ., ..)
+pub fn normalize_uri(uri: &str) -> anyhow::Result<String> {
```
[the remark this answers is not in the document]
evanxg852000: Sure

evanxg852000:
@fmassot  @fulmicoton  We need your take on the idea of having a Uri type `Uri(String)`

guilload:
@evanxg852000, you can go ahead and merge this. I'll open a separate issue for the URI type.

fmassot (commented):
LGTM!

guilload:
I'm super stoked on this btw; it's a major usability upgrade.

PR #819, opened 2021-11-29
Terminate indexing pipeline once all merge operations have been done

fmassot:
Accurate termination condition:
- source + indexer + packager + uploader + publisher have exited successfully
- all actors of the merge pipeline are in `IDLE` state

## Test
I need to test it on a real dataset. For the moment, I did not find a nice way to add a unit test on the pipeline termination :/

### DIFF
--- a/quickwit-indexing/src/actors/pipeline_supervisor.rs
+++ b/quickwit-indexing/src/actors/pipeline_supervisor.rs
@@ -23,8 +23,8 @@ use std::time::Duration;
 use async_trait::async_trait;
 use itertools::Itertools;
 use quickwit_actors::{
-    create_mailbox, Actor, ActorContext, ActorExitStatus, ActorHandle, AsyncActor, Health,
-    KillSwitch, QueueCapacity, Supervisable,
+    create_mailbox, Actor, ActorContext, ActorExitStatus, ActorHandle, ActorState, AsyncActor,
+    Health, KillSwitch, QueueCapacity, Supervisable,
 };
 use quickwit_metastore::{Metastore, SplitState};
 use quickwit_storage::StorageUriResolver;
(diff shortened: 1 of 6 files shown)


quickwit-indexing/src/actors/pipeline_supervisor.rs
```diff
@@ -416,7 +414,40 @@ impl IndexingPipelineSupervisor {
...
+                        // This is true when all merging pipeline actors are all in `IDLE` state
+                        // and indexing actors have exited.
+                        if handlers.source.state() == ActorState::Exit
+                            && handlers.indexer.state() == ActorState::Exit
+                            && handlers.packager.state() == ActorState::Exit
+                            && handlers.uploader.state() == ActorState::Exit
+                            && handlers.publisher.state() == ActorState::Exit
+                            && handlers.merge_planner.state() == ActorState::Idle
+                            && handlers.merge_split_downloader.state() == ActorState::Idle
+                            && handlers.merge_executor.state() == ActorState::Idle
```
fulmicoton, line 430: I think there is a small race condition. (Maybe we don't care).
There are many ways for this condition to be true at one point of time, and not true later.

fulmicoton: Again better than nothing, but I think it is important to be aware of that.

fmassot: Indeed, I was not very confident when I wrote this code, many equality checks on shifting sands.
I will leave it a few days and see if I come up with better ideas or at least some docs to explain the limitations of this approach.

fmassot: To make the shutdown more reliable, here is what we can do when there is only the merge pipeline running:
- keep a boolean to track the fact that the merge pipeline was idle
- on supervise
  - if merge pipeline is idle then: if bool is true, exit if not set bool to true
  - if not, set bool to false

fmassot: @fulmicoton I added a state variable on the indexing pipeline to terminate the pipeline after 2 consecutive idle merge pipeline states.

fmassot:
Closing this PR for now.

PR #822, opened 2021-11-30
Test `IndexMetadata` deserialization for backward compatibility

guilload:
I figure this would be easier to do before merging my work on [index metadata v1]( https://github.com/quickwit-inc/quickwit/tree/guilload--index-metadata-v1).


quickwit-backward-compat/build.rs
```diff
@@ -39,6 +61,16 @@ fn sample_split_metadata_for_regression() -> SplitMetadataAndFooterOffsets {
 }
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let index_metadata = sample_index_metadata_for_regression();
```
fulmicoton, line 64: the duplicated code is fine but can we just isolate the two parts (split and index) into two different functions?

fulmicoton (approved):

PR #833, opened 2021-12-02
Split pruning by tags

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/823
- Implemented split running by tag. 
Possible refactoring need to be discussed with @fulmicoton as currently we kind of call `tantivy_query_grammar::parse_query(query)` many times during a request. 

### How was this PR tested?
- unit test added

### DIFF
--- a/quickwit-index-config/src/lib.rs
+++ b/quickwit-index-config/src/lib.rs
@@ -35,6 +35,7 @@ pub use default_index_config::{
     DefaultIndexConfig, DefaultIndexConfigBuilder, DocParsingError, FieldMappingEntry,
 };
 pub use error::QueryParserError;
+pub use query_builder::extract_tags_from_query;
 pub use wikipedia_config::WikipediaIndexConfig;
 
 /// Field name reserved for storing the source document.
--- a/quickwit-index-config/src/query_builder.rs
+++ b/quickwit-index-config/src/query_builder.rs
@@ -21,7 +21,7 @@ use quickwit_proto::SearchRequest;
 use tantivy::query::{Query, QueryParser, QueryParserError as TantivyQueryParserError};
 use tantivy::schema::{Field, Schema};
 use tantivy::tokenizer::TokenizerManager;
-use tantivy_query_grammar::{UserInputAst, UserInputLeaf};
+use tantivy_query_grammar::{UserInputAst, UserInputLeaf, UserInputLiteral};
 
 use crate::QueryParserError;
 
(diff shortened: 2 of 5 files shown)


fmassot:
Oh didn't know you were already working on this! I had a short discussion with @r-zenine yesterday on that.

I see one problem but you may already have this in mind: you are currently collecting tags but... what will happen with a query like `some_id:1000 OR body:text`? We don't want to prune splits in this case.

r-zenine (approved):

r-zenine:
Suggestion: 
- Should we remove the `tags` query parameter from the search endpoints, given that we are automatically extracting them now. It will make the api simpler to explain 
-

evanxg852000:
@r-zenine  I still need to sort out the `OR` case. I will open the PR for your feedback when those changes are added.
Thanks

fmassot:
> Should we remove the tags query parameter from the search endpoints, given that we are automatically extracting them now. It will make the api simpler to explain

@r-zenine we should not remove the tags. In a given split, we can still have multiple values so we need to keep it.

fmassot:
> Should we remove the tags query parameter from the search endpoints, given that we are automatically extracting them now. It will make the api simpler to explain

That sounds a good idea and it will avoid conflicts between settings tags in two different parameters, I can already see a user querying with `tags=some_id:100&query=some_id:102`. I don't see any downside in removing it.

quickwit-index-config/src/query_builder.rs
```diff
@@ -49,6 +49,43 @@ pub(crate) fn build_query(
...
+        .map_err(|_| TantivyQueryParserError::SyntaxError)?;
+
+    Ok(collect_tag_filters(user_input_ast, tag_field_names))
+}
+
+fn collect_tag_filters(user_input_ast: UserInputAst, tag_field_names: &[String]) -> Vec<String> {
+    let mut fields = vec![];
+    match user_input_ast {
+        UserInputAst::Clause(sub_queries) => {
+            for (_, sub_ast) in sub_queries {
```
fulmicoton, line 68: The `_` is actually important, the Occur information.
That's what decides whether the term should be or must be here. (NOT is also possible.

In the boolean logic world, is it A AND B or A OR B.

fulmicoton (changes requested):
The current logic does not work as it does not handle the `Occur::{Must, Should, Not}` object.

r-zenine:
Hello everyone,

I tried another approach this week-end. 
I derive an AST  of tag pruning conditions from the query. 
It correctly handles AND/OR conditions but it's not able for now to handle Not conditions correctly. 

Going further in that direction would imply changing the MetastoreTrait ( function list_relevant_splits ) 
Its in the branch  https://github.com/quickwit-inc/quickwit/tree/fix-823-ryad

Can you take a look and tell me what you think about it ? The way I see it we would : 
- Update the `list_relevant_splits` in the MetastoreTrait
- Remove the tag query param from the endpoints 
- Generate relevant conditions for the PosgresSQL and File Megastore

fulmicoton:
@r-zenine I created a PR and commented on site. You can liberally create PR and mark them as draft for that kind of "ongoing work".

Generally speaking, I really like the idea to have an explicit TagsFilterAST.
In another ticket, we try to workaround tags having a high cardinality.
Over N terms, we only store "tag:*" in the metastore, which means we need to do some AST manipulation to replace
each tag leaf by `(tag_field:my_tag OR tag_field:*)`.
The AST will make that work trivial.

On the metastore side, passing the ast itself can make things a little bit complicated. 
Still it should not be too bad.

I don't think your implementation is correct however. Or maybe I misunderstood something.
Copy pasting my comment here.

```
[(Must, "a"), (Must, "b")] here ends up interpreted as or([ and(["a"]), and(["b"]) ])
when it should actually be and([a. b]).
Although to be honest, it may have been the intended spec.As long as the predicate returned is implied by the query, the filter is technically correct.

If this is the intended behavior, some comments could have been helpful here.
```

The relationship between Clause and boolean formula may have been misleading. 
I borrowed the idea from Lucene of course, and while it was confusing at first it ended up making sense.

The "clause" formulation is more suited to search because of scoring.
Let's consider the query:

`(+ipod cheap)  == ( [(Must, ipod), (Should, cheap)] )`
 
 It's equivalent as a boolean expression does not exist.
When asking tantivy to interpret this query and not caring about computing scores, tantivy just interprets it as 
ipod. 

Indeed the set of documents matching this queries is exactly the set of document matching ipod. "cheap" has one effect however: it has an impact on scores.

quickwit-index-config/src/lib.rs
```diff
@@ -35,6 +35,7 @@ pub use default_index_config::{
     DefaultIndexConfig, DefaultIndexConfigBuilder, DocParsingError, FieldMappingEntry,
 };
 pub use error::QueryParserError;
+pub use query_builder::extract_tags_from_query;
```
fulmicoton, line 38: There is no great place to put the tag pruning logic because it is required by both search and indexing.
However index-config is probably the worst place :)

If possible `quickwit-common` would be ok but I think it requires adding a deps to tantivy_query_grammar which would be lame.

Maybe `quickwit-indexing` might be nicer? That will add a deps from search to `quickwit-indexing` though.
An alternative could be to create a dedicated quickwit-query project to avoid that dependency. No need to do that in this PR though.

fulmicoton:
@r-zenine I added a commit on your branch that adds some unit tests that were failing, and modified your code to handle them. 
One simple failing example was 

(tag:a AND tag:b)

fulmicoton:
Closing this PR as this effort is now on #886

PR #843, opened 2021-12-06
Limit tag split values & escaping

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/437
Related to https://github.com/quickwit-inc/quickwit/issues/823

### How was this PR tested?
unittest

### DIFF
--- a/quickwit-index-config/src/config.rs
+++ b/quickwit-index-config/src/config.rs
@@ -59,40 +59,12 @@ use crate::{DocParsingError, QueryParserError, SortBy, TAGS_FIELD_NAME};
 /// Converts a field (name, value) into a tag string `name:value`.
 pub fn convert_tag_to_string(field_name: &str, field_value: &Value) -> String {
     let field_value_str = tantivy_value_to_string(field_value);
-    if field_value_str == TOO_MANY_TAG_VALUES {
-        return format!(
-            "{}{}{}{}",
-            field_name, TAG_FIELD_VALUE_SEPARATOR, TAGS_VALUE_ESCAPE, field_value_str
-        );
-    }
-    format!(
-        "{}{}{}",
-        field_name, TAG_FIELD_VALUE_SEPARATOR, field_value_str
-    )
+    make_tag_value(field_name, &field_value_str)
 }
 
 /// Returns true if tag_string is of form `{field_name}:any_value`.
 pub fn match_tag_field_name(field_name: &str, tag_string: &str) -> bool {
     tag_string.starts_with(&format!("{}{}", field_name, TAG_FIELD_VALUE_SEPARATOR))
-}
-
-/// Extracts the field name from a tag value of the form `{field_name}:any_value`.
-pub fn extract_field_name_from_tag_value(tag_value: &str) -> Option<String> {
-    tag_value
-        .split_once(TAG_FIELD_VALUE_SEPARATOR)
-        .map(|(field_name, _)| field_name.to_string())
-}
-
-/// Escapes the tag value by prefixing the value part of this format `{field_name}:any_value`
-/// by `\` if `any_value` is the wildcard value `*`.
-pub fn escape_tag_value(tag_value: &str) -> String {
-    match tag_value.split_once(TAG_FIELD_VALUE_SEPARATOR) {
-        Some((field_name, value)) if value == TOO_MANY_TAG_VALUES => format!(
-            "{}{}{}{}",
-            field_name, TAG_FIELD_VALUE_SEPARATOR, TAGS_VALUE_ESCAPE, value
-        ),
-        _ => tag_value.to_string(),
-    }
 }
 
 /// Creates the wildcard tag value for a field name: `{field_name}:*`.
@@ -105,6 +77,12 @@ pub fn make_too_many_tag_value(field_name: &str) -> String {
 
 /// Creates a tag value for a field name of this format `{field_name}:value`.
 pub fn make_tag_value(field_name: &str, field_value: &str) -> String {
+    if field_value == TOO_MANY_TAG_VALUES {
+        return format!(
+            "{}{}{}{}",
+            field_name, TAG_FIELD_VALUE_SEPARATOR, TAGS_VALUE_ESCAPE, field_value
+        );
+    }
     format!("{}{}{}", field_name, TAG_FIELD_VALUE_SEPARATOR, field_value)
 }
 
--- a/quickwit-index-config/src/config.rs
+++ b/quickwit-index-config/src/config.rs
@@ -31,7 +31,7 @@ use crate::{DocParsingError, QueryParserError, TAGS_FIELD_NAME};
 pub const TAG_FIELD_VALUE_SEPARATOR: &str = ":";
 
 /// Wilcard value use to collapse too many tag values into one.
-pub const MANY_TAG_VALUES: &str = "*";
+pub const TOO_MANY_TAG_VALUES: &str = "*";
 
 /// Character use to escape tag value when there is collision with the wilcard
 /// tag values.
@@ -59,7 +59,7 @@ use crate::{DocParsingError, QueryParserError, SortBy, TAGS_FIELD_NAME};
 /// Converts a field (name, value) into a tag string `name:value`.
 pub fn convert_tag_to_string(field_name: &str, field_value: &Value) -> String {
     let field_value_str = tantivy_value_to_string(field_value);
-    if field_value_str == MANY_TAG_VALUES {
+    if field_value_str == TOO_MANY_TAG_VALUES {
         return format!(
             "{}{}{}{}",
             field_name, TAG_FIELD_VALUE_SEPARATOR, TAGS_VALUE_ESCAPE, field_value_str
@@ -87,7 +87,7 @@ pub fn extract_field_name_from_tag_value(tag_value: &str) -> Option<String> {
 /// by `\` if `any_value` is the wildcard value `*`.
 pub fn escape_tag_value(tag_value: &str) -> String {
     match tag_value.split_once(TAG_FIELD_VALUE_SEPARATOR) {
-        Some((field_name, value)) if value == MANY_TAG_VALUES => format!(
+        Some((field_name, value)) if value == TOO_MANY_TAG_VALUES => format!(
             "{}{}{}{}",
             field_name, TAG_FIELD_VALUE_SEPARATOR, TAGS_VALUE_ESCAPE, value
         ),
@@ -99,8 +99,13 @@ pub fn escape_tag_value(tag_value: &str) -> String {
 pub fn make_too_many_tag_value(field_name: &str) -> String {
     format!(
         "{}{}{}",
-        field_name, TAG_FIELD_VALUE_SEPARATOR, MANY_TAG_VALUES
+        field_name, TAG_FIELD_VALUE_SEPARATOR, TOO_MANY_TAG_VALUES
     )
+}
+
+/// Creates a tag value for a field name of this format `{field_name}:value`.
+pub fn make_tag_value(field_name: &str, field_value: &str) -> String {
+    format!("{}{}{}", field_name, TAG_FIELD_VALUE_SEPARATOR, field_value)
 }
 
 /// Converts a [`tantivy::Value`] to it's [`String`] value.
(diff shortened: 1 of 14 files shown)


quickwit-index-config/src/config.rs
```diff
@@ -72,12 +82,48 @@ pub enum SortBy {
...
 
 /// Returns true if tag_string is of form `{field_name}:any_value`.
 pub fn match_tag_field_name(field_name: &str, tag_string: &str) -> bool {
-    tag_string.starts_with(&format!("{}:", field_name))
+    tag_string.starts_with(&format!("{}{}", field_name, TAG_FIELD_VALUE_SEPARATOR))
+}
+
+/// Extracts the field name from a tag value of the form `{field_name}:any_value`.
+/// It will return the whole string if the separator is not present (to discuss: should we error?).
+pub fn extract_field_name_from_tag_value(tag_value: &str) -> String {
```
fmassot, line 102: why not return `None` if it does not match? I found quite normal to return an option here.

quickwit-indexing/src/actors/packager.rs
```diff
@@ -205,7 +208,7 @@ fn create_packaged_split(
         .sum();
 
     // Extracts tag values from `_tags` special fields.
-    let mut tags = BTreeSet::default();
+    let mut field_name_to_tag_values: HashMap<String, HashSet<String>> = HashMap::default();
```
fmassot, line 211: ok we should not do that :)

you should use the inverted index to get the number of tokens.

quickwit-index-config/src/config.rs
```diff
@@ -28,6 +28,16 @@ use tantivy::{Document, Order};
...
 
+/// Separator used to format tags into `{field_name}:{value}`
+pub const TAG_FIELD_VALUE_SEPARATOR: &str = ":";
+
+/// Wilcard value use to collapse too many tag values into one.
+pub const MANY_TAG_VALUES: &str = "*";
+
+/// Character use to escape tag value when there is collision with the wilcard
+/// tag values.
+pub const TAGS_VALUE_ESCAPE: &str = "\\";
```
fmassot, line 39: I'm not sure to understand why we need two backslashes.

quickwit-index-config/src/config.rs
```diff
@@ -72,12 +82,48 @@ pub enum SortBy {
 
 /// Convert a field (name, value) into a tag string `name:value`.
 pub fn convert_tag_to_string(field_name: &str, field_value: &Value) -> String {
-    format!("{}:{}", field_name, tantivy_value_to_string(field_value))
+    let mut field_value_str = tantivy_value_to_string(field_value);
```
fmassot, line 85: `format!` is expensive, I would call it only once

fmassot (changes requested):
I added some comments.

quickwit-index-config/src/config.rs
evanxg852000: I am thinking of switching to `char`. `\` alone in a string need to be escaped by itself

quickwit-index-config/src/config.rs
evanxg852000: The implication I wanted to clear may be more than this return type. So right now we have not been enforcing tags to have this format `{field}:{value}` and this current implementation follows that by **trying** at best to parse if not successful we just return the whole string. Now if we return option/error it means we may need to make sure to error out if the format of tag is not correct.
I still think the correct semantic is what you are hinting for so I will do that.

quickwit-index-config/src/config.rs
fmassot: ah yes indeed. Maybe a char would be nice

quickwit-indexing/src/actors/packager.rs
```diff
@@ -205,17 +219,47 @@ fn create_packaged_split(
...
+                        Some(field_name)
+                    } else {
+                        None
+                    }
+                })
+            {
+                tags.insert(make_too_many_tag_value(&field_name));
+            } else {
+                tags.insert(tag_value);
+            }
```
fmassot, line 262: this part is annoying. Why not iterating of the field tokens directly? This may sound a little bit stupid because we need to rebuild the tag but it would be more readable.

fmassot: At the end, I would do something like that:
```rust
for (tag_field_name, &tag_field) in tag_fields_list {
  if segments.iter().map(|seg| seg.inv_index(tag_field).terms().num_terms()).sum()> 1000 {
      continue;
  }
  // code to add tokens
}

```

quickwit-index-config/src/config.rs
```diff
@@ -70,14 +80,51 @@ pub enum SortBy {
...
+pub fn extract_field_name_from_tag_value(tag_value: &str) -> Option<String> {
+    tag_value
+        .split_once(TAG_FIELD_VALUE_SEPARATOR)
+        .map(|(field_name, _)| field_name.to_string())
+}
+
+/// Escapes the tag value by prefixing the value part of this format `{field_name}:any_value`
+/// by `\` if `any_value` is the wildcard value `*`.
+pub fn escape_tag_value(tag_value: &str) -> String {
+    match tag_value.split_once(TAG_FIELD_VALUE_SEPARATOR) {
```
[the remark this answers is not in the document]
evanxg852000: Yes, but we wanted to keep this wildcard as const so to change it in case we favored `_too_many_value`. This is why I am using all these variable interpolations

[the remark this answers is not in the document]
fmassot: Sorry my comment was not accurate at all! I will rewrite it.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -330,8 +333,21 @@ impl PostgresqlMetastore {
 
         if let Some(tags) = tags_opt {
             if !tags.is_empty() {
-                select_statement =
-                    select_statement.filter(schema::splits::dsl::tags.overlaps_with(tags));
+                // Rebuild split tags by escaping values & taking into account
```
fmassot, line 336: One question here: do we need to add this piece of code for each metastore implementation?

fmassot: Ok nevermind, I saw the common function `match_tags_filter ` so forget my comment.

evanxg852000: This is a good point will attempt to lift this, let's discuss options once I gather them.

quickwit-metastore/src/metastore/mod.rs
```diff
@@ -172,9 +172,9 @@ pub trait Metastore: Send + Sync + 'static {
 
 /// Finds out if a split matches a tag filter expression.
 ///
-/// `filter_tags` tags is an array representing a boolean expression
+/// `filter_tags` is an array representing a boolean expression
 /// of the form `[ [A OR B] AND  [C OR D] ]`
-/// Returns true if filter_tags is empty, or if the split's tags matches
+/// Returns true if filter_tags is empty, or if the split's tags matche
```
fmassot, line 177: `matche` => `match`

quickwit-metastore/src/metastore/mod.rs
```diff
@@ -172,9 +172,9 @@ pub trait Metastore: Send + Sync + 'static {
 
 /// Finds out if a split matches a tag filter expression.
 ///
-/// `filter_tags` tags is an array representing a boolean expression
+/// `filter_tags` is an array representing a boolean expression
```
fmassot, line 175: I think we can describe a little bit better, `filter_tags` is an array of array of tags representing
a boolean expression 
`of the form [ [A OR B OR...] AND [C OR D OR ...] AND ...]

quickwit-index-config/src/config.rs
```diff
@@ -70,14 +80,34 @@ pub enum SortBy {
     DocId,
 }
 
-/// Convert a field (name, value) into a tag string `name:value`.
+/// Converts a field (name, value) into a tag string `name:value`.
 pub fn convert_tag_to_string(field_name: &str, field_value: &Value) -> String {
-    format!("{}:{}", field_name, tantivy_value_to_string(field_value))
+    let field_value_str = tantivy_value_to_string(field_value);
+    make_tag_value(field_name, &field_value_str)
```
fmassot, line 86: tiny nitpick: I think we use more `build` than `make`

quickwit-indexing/src/actors/packager.rs
```diff
@@ -48,20 +49,21 @@ use crate::models::{
...
+    /// A list of tag fields specified in the index config and their
+    /// corresponding schema field.
+    tag_fields_list: Vec<(String, Field)>,
 }
 
 impl Packager {
     pub fn new(
         actor_name: &'static str,
-        tags_field: Field,
+        tag_fields_list: Vec<(String, Field)>,
```
fmassot, line 60: I would introduce a struct `NamedField` struct to make it clearer

fmassot: with `tag_fields` variable

quickwit-indexing/src/actors/packager.rs
```diff
@@ -205,17 +207,36 @@ fn create_packaged_split(
...
         .reader_builder()
         .reload_policy(ReloadPolicy::Manual)
         .try_into()?;
-    for reader in index_reader.searcher().segment_readers() {
-        let inv_index = reader.inverted_index(tags_field)?;
-        let mut terms_streamer = inv_index.terms().stream()?;
-        while let Some((term_data, _)) = terms_streamer.next() {
-            tags.insert(String::from_utf8_lossy(term_data).to_string());
+    let mut tags = BTreeSet::default();
+    for (tag_field_name, tag_field) in tag_fields_list {
```
fmassot, line 216: mmmm, I think `Field` is specific to a segment, not an index. The field resolving needs to be made at the segment level

evanxg852000: Are you implying segments within the same index can have a different structure or is this taking into account schema change?

quickwit-indexing/src/actors/packager.rs
fmassot:
```rust
pub fn new(
    ...
    tag_fields: Vec<NamedField>,
    ...
)

```

quickwit-indexing/src/actors/packager.rs
fmassot: Ok, here it is safe, let's keep what you did.

quickwit-indexing/src/actors/packager.rs
```diff
@@ -48,20 +49,21 @@ use crate::models::{
 pub struct Packager {
     actor_name: &'static str,
     uploader_mailbox: Mailbox<PackagedSplitBatch>,
-    /// The special field for extracting tags.
-    tags_field: Field,
+    /// A list of tag fields specified in the index config and their
```
fmassot, line 52: /// List of tag fields `(name: str, field: Field)` defined in the index config.

quickwit-indexing/src/actors/packager.rs
fmassot: it's ok if the docs is better as I suggested above.

fmassot (changes requested):
Seems good, see comments.

quickwit-indexing/src/actors/packager.rs
evanxg852000: I went with NamedField, This is not the first time we need something like this.

quickwit-indexing/src/actors/packager.rs
evanxg852000: /// List of tag fields ([Vec<NamedField>]) defined in the index config.

quickwit-indexing/src/actors/mod.rs
```diff
@@ -40,3 +41,12 @@ pub use self::merge_split_downloader::MergeSplitDownloader;
 pub use self::packager::Packager;
 pub use self::publisher::{Publisher, PublisherCounters};
 pub use self::uploader::{Uploader, UploaderCounters};
+
+/// A struct to wrap a tantivy field with it's name.
```
fmassot, line 45: /// A struct to wrap a tantivy field with its name.

fmassot (approved):
Good for me!

PR #851, opened 2021-12-08
add retry for failing spawn_pipeline

### LINKED ISSUE #463: Better supervisor
Right now the pipeline supervisor retry logic has two problems.

First, it just retries right away. We should probably have some type of exponential backoff...
Maybe we should log a warning on the first issue, and an error on the second to help users configure pagers?
(@guilload what is the right practise here?)

Second (which is way more important)
Right now, if an error happens during the pipeline creation, we just fail.
This should also be part of the respawn process. 

The reason why it is deactivated is because, for the current CLI experience, we want to exit if there is a connectivity problem. There is a consensus however that this should be solved in #701.

PSeitz:
fixes #463

### Description
add spawn command for supervisor with retry

### How was this PR tested?

### DIFF
--- a/quickwit-indexing/src/actors/indexing_pipeline.rs
+++ b/quickwit-indexing/src/actors/indexing_pipeline.rs
@@ -34,7 +34,7 @@ use quickwit_metastore::checkpoint::Checkpoint;
 use quickwit_metastore::{IndexMetadata, Metastore, SplitState};
 use quickwit_storage::Storage;
 use tokio::join;
-use tracing::{debug, error, info, info_span, instrument, warn, Span};
+use tracing::{debug, error, info, info_span, instrument, Span};
 
 use crate::actors::merge_split_downloader::MergeSplitDownloader;
 use crate::actors::{
@@ -421,36 +421,36 @@ impl IndexingPipeline {
         Ok(())
     }
 
+    /// retry_count, wait_time_sec
+    /// 1   1
+    /// 2   4
+    /// 3   8
+    /// 4   16
+    fn wait_duration_before_retry(retry_count: usize) -> Option<Duration> {
+        let mut wait_time_sec = (retry_count + 1).pow(2);
+        wait_time_sec = wait_time_sec.min(60 * 5);
+        Some(Duration::from_secs(wait_time_sec as u64))
+    }
+
     async fn process_spawn(
         &mut self,
-        ctx: &ActorContext<Msg>,
+        ctx: &ActorContext<IndexingPipelineMessage>,
         retry_count: usize,
     ) -> Result<(), ActorExitStatus> {
-        match retry_count {
-            0 | 1 | 2 => {
-                let duration = if retry_count == 0 {
-                    Duration::from_millis(1000)
-                } else if retry_count == 1 {
-                    Duration::from_millis(2000)
-                } else {
-                    Duration::from_millis(5000)
-                };
+        if let Err(spawn_error) = self.spawn_pipeline(ctx).await {
+            if let Some(duration_before_retry) = Self::wait_duration_before_retry(retry_count) {
+                error!(err=?spawn_error, "Error while spawn_pipeline, waiting 1000ms");
+                ctx.schedule_self_msg_blocking(
+                    duration_before_retry,
+                    IndexingPipelineMessage::Spawn {
+                        retry_count: retry_count + 1,
+                    },
+                );
+            } else {
+                return Err(ActorExitStatus::Failure(Arc::new(spawn_error)));
+            }
+        }
 
-                if let Err(spawn_error) = self.spawn_pipeline(ctx).await {
-                    warn!(err=?spawn_error, "Error while spawn_pipeline, waiting 1000ms");
-                    self.terminate().await;
-                    ctx.schedule_self_msg_blocking(
-                        duration,
-                        Msg::Spawn {
-                            retry_count: retry_count + 1,
-                        },
-                    );
-                }
-            }
-            _ => {
-                self.spawn_pipeline(ctx).await?;
-            }
-        };
         Ok(())
     }
 
@@ -518,7 +518,9 @@ impl AsyncActor for IndexingPipeline {
         match message {
             IndexingPipelineMessage::Observe => self.process_observe(ctx).await?,
             IndexingPipelineMessage::Supervise => self.process_supervise(ctx).await?,
-            Msg::Spawn { retry_count } => self.process_spawn(ctx, retry_count).await?,
+            IndexingPipelineMessage::Spawn { retry_count } => {
+                self.process_spawn(ctx, retry_count).await?
+            }
         }
         Ok(())
     }
@@ -572,7 +574,7 @@ mod tests {
     use quickwit_config::IndexingSettings;
     use quickwit_index_config::default_config_for_tests;
     use quickwit_metastore::checkpoint::Checkpoint;
-    use quickwit_metastore::MockMetastore;
+    use quickwit_metastore::{MetastoreError, MockMetastore};
     use quickwit_storage::RamStorage;
     use serde_json::json;
 
@@ -595,20 +597,15 @@ mod tests {
             .expect_index_metadata()
             .withf(|index_id| index_id == "test-index")
             .returning(move |_| {
-                if num_fails != 0 {
-                    num_fails -= num_fails.saturating_sub(1);
-                    Err(MetastoreError::ConnectionError {
-                        message: "MetastoreError Alarm".to_string(),
-                    })
-                } else {
-                    let index_metadata = IndexMetadata {
-                        index_id: "test-index".to_string(),
-                        index_uri: "ram://test-index".to_string(),
-                        index_config: Arc::new(quickwit_index_config::default_config_for_tests()),
-                        checkpoint: Default::default(),
-                    };
-                    Ok(index_metadata)
+                if num_fails == 0 {
+                    let index_metadata =
+                        IndexMetadata::for_test("test-index", "ram://indexes/my-index");
+                    return Ok(index_metadata);
                 }
+                num_fails -= 1;
+                Err(MetastoreError::ConnectionError {
+                    message: "MetastoreError Alarm".to_string(),
+                })
             });
         metastore
             .expect_stage_split()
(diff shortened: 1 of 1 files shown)


quickwit-indexing/src/actors/pipeline_supervisor.rs
```diff
@@ -405,16 +407,41 @@ impl IndexingPipelineSupervisor {
...
+                            retry_count: retry_count + 1,
+                        },
+                    );
+                }
+            }
+            _ => {
+                self.spawn_pipeline(ctx).await?;
+            }
+        };
+        Ok(())
```
fulmicoton, line 440: This is too convoluted.

```rust
if let Err(spawn_error) = self.spawn_pipeline(ctx).await {
   if let Some(duration_before_retry) = wait_duration_before_retry(retry_count) {
      ctx.schedule_self_msg_blocking(
             duration,
             Msg::Spawn {
                 retry_count: retry_count + 1,
             },
     );
   } else {
      return Err(spawn_error);
   }
}
Ok(())
```

or with a functional style

```rust
self.spawn_pipeline(ctx).await
     .or_else(|err| {
          let duration_before_retry = wait_duration_before_retry(retry_count)
              .unwrap_or_else(|| Err(err))?;
           ctx.schedule_self_msg_blocking(
                    duration,
                    Msg::Spawn {
                        retry_count: retry_count + 1,
                    });
            });
            Ok(())
    })
```

quickwit-indexing/src/actors/pipeline_supervisor.rs
```diff
@@ -492,14 +521,106 @@ mod tests {
...
+        metastore
+            .expect_mark_splits_for_deletion()
+            .returning(|_, _| Ok(()));
+
+        metastore
+            .expect_index_metadata()
+            .withf(|index_id| index_id == "test-index")
+            .returning(move |_| {
+                if num_fails != 0 {
+                    num_fails -= num_fails.saturating_sub(1);
```
fulmicoton, line 549:
```suggestion
                    num_fails -= 1;
```

quickwit-indexing/src/actors/pipeline_supervisor.rs
```diff
@@ -492,14 +521,106 @@ mod tests {
...
+                    Err(MetastoreError::ConnectionError {
+                        message: "MetastoreError Alarm".to_string(),
+                    })
+                } else {
```
fulmicoton, line 553: Early return would be nicer here.

quickwit-indexing/src/actors/pipeline_supervisor.rs
```diff
@@ -405,16 +407,41 @@ impl IndexingPipelineSupervisor {
...
 
+    async fn process_spawn(
+        &mut self,
+        ctx: &ActorContext<Msg>,
+        retry_count: usize,
+    ) -> Result<(), ActorExitStatus> {
+        match retry_count {
+            0 | 1 | 2 => {
+                let duration = if retry_count == 0 {
+                    Duration::from_millis(1000)
```
fulmicoton, line 418: There is a Duration::from_secs.

quickwit-indexing/src/actors/pipeline_supervisor.rs
```diff
@@ -405,16 +407,41 @@ impl IndexingPipelineSupervisor {
...
+                } else if retry_count == 1 {
+                    Duration::from_millis(2000)
+                } else {
+                    Duration::from_millis(5000)
+                };
+
+                if let Err(spawn_error) = self.spawn_pipeline(ctx).await {
+                    warn!(err=?spawn_error, "Error while spawn_pipeline, waiting 1000ms");
```
fulmicoton, line 426: Warning is incorrect.

quickwit-indexing/src/actors/pipeline_supervisor.rs
```diff
@@ -405,16 +407,41 @@ impl IndexingPipelineSupervisor {
...
+                    ctx.schedule_self_msg_blocking(
+                        duration,
+                        Msg::Spawn {
+                            retry_count: retry_count + 1,
+                        },
+                    );
+                }
+            }
+            _ => {
+                self.spawn_pipeline(ctx).await?;
```
fulmicoton, line 437: I think actually we want to keep retrying forever, at least for the moment.

Soooo:
- retry forever
- exponential backoff until the retry goes up to say 5mn, after which we stagnate to 5mn. (It actually simpliy the code too)

fulmicoton (changes requested):
See comments inline.

fulmicoton (changes requested):
see comments.

quickwit-indexing/src/actors/pipeline_supervisor.rs
PSeitz: if it's just for now, I would keep the logic like it is and just always return Some(duration)

fulmicoton: Makes sense.

fulmicoton:
@PSeitz I think most of my comments were lost and left unaddressed. 
Should I readd them?

PSeitz:
> @PSeitz I think most of my comments were lost and left unaddressed. Should I readd them?

Ah sorry, yeah please readd them

quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -419,19 +421,44 @@ impl IndexingPipeline {
...
 
+    async fn process_spawn(
+        &mut self,
+        ctx: &ActorContext<Msg>,
+        retry_count: usize,
+    ) -> Result<(), ActorExitStatus> {
+        match retry_count {
+            0 | 1 | 2 => {
+                let duration = if retry_count == 0 {
+                    Duration::from_millis(1000)
```
fulmicoton, line 432: from_secs exists.

quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -419,19 +421,44 @@ impl IndexingPipeline {
...
+                } else if retry_count == 1 {
+                    Duration::from_millis(2000)
```
fulmicoton, line 434: This is too convoluted.

```rust
if let Err(spawn_error) = self.spawn_pipeline(ctx).await {
   if let Some(duration_before_retry) = wait_duration_before_retry(retry_count) {
      ctx.schedule_self_msg_blocking(
             duration,
             Msg::Spawn {
                 retry_count: retry_count + 1,
             },
     );
   } else {
      return Err(spawn_error);
   }
}
Ok(())
```

or with a functional style

```rust
self.spawn_pipeline(ctx).await
     .or_else(|err| {
          let duration_before_retry = wait_duration_before_retry(retry_count)
              .unwrap_or_else(|| Err(err))?;
           ctx.schedule_self_msg_blocking(
                    duration,
                    Msg::Spawn {
                        retry_count: retry_count + 1,
                    });
            });
            Ok(())
    })
```

quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -550,6 +579,98 @@ mod tests {
...
+        metastore
+            .expect_mark_splits_for_deletion()
+            .returning(|_, _| Ok(()));
+
+        metastore
+            .expect_index_metadata()
+            .withf(|index_id| index_id == "test-index")
+            .returning(move |_| {
+                if num_fails != 0 {
+                    num_fails -= num_fails.saturating_sub(1);
```
fulmicoton, line 599: !?

fulmicoton:
```suggestion
                    num_fails -= 1;
```

fulmicoton (changes requested):
See comments.

quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -419,19 +421,44 @@ impl IndexingPipeline {
         Ok(())
     }
 
+    /// retry_count, wait_time_sec
+    /// 1   1
+    /// 2   4
+    /// 3   8
+    /// 4   16
```
fulmicoton, line 428: /// ...
/// >= 9    5mn

fulmicoton, line 428:
```suggestion
    /// retry_count, wait_time
    /// 0   1s
    /// 1   2s
    /// 2   4s
    /// 3   8s
    /// ...
    /// >=8   5mn
```

quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -419,19 +421,44 @@ impl IndexingPipeline {
...
+    fn wait_duration_before_retry(retry_count: usize) -> Option<Duration> {
```
fulmicoton, line 429: unit test on the mapping please.

quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -419,19 +421,44 @@ impl IndexingPipeline {
...
-        if self.handlers.is_none() {
-            // TODO Accept errors in spawning. See #463.
-            self.spawn_pipeline(ctx).await?;
-            // if let Err(spawn_error) = self.spawn_pipeline(ctx).await {
-            //    // only retry n-times.
-            //    error!(err=?spawn_error, "Error while spawning");
-            //    self.terminate().await;
-            // }
-        } else {
+        if self.handlers.is_some() {
```
fulmicoton, line 461: I am lost... When do we respawn if workers have failed now?

fulmicoton (changes requested):
Question: When do we respawn if the worker fail?

quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -543,13 +574,94 @@ mod tests {
...
+            .returning(|_, _, _, _| Ok(Vec::new()));
+        metastore
+            .expect_mark_splits_for_deletion()
+            .returning(|_, _| Ok(()));
+
+        metastore
+            .expect_index_metadata()
+            .withf(|index_id| index_id == "test-index")
+            .returning(move |_| {
+                if num_fails == 0 {
```
fulmicoton, line 600: I wonder if it is possible to do that in mockall with "times".

Basically register the n failure first, and then register the success.

fulmicoton (changes requested):
see comments

quickwit-indexing/src/actors/indexing_pipeline.rs
PSeitz: I added it to the observer now

fulmicoton (approved):

PR #868, opened 2021-12-10
Use gRPC and discovery ports relative to REST port

guilload:
### Description
- Use gRPC and discovery ports relative to REST port
- Rename `split_max_num_docs` to `split_num_docs_target`
- Simplify `get_resource_path`

### How was this PR tested?
`make test-all`

### DIFF
--- a/quickwit-config/src/server_config.rs
+++ b/quickwit-config/src/server_config.rs
@@ -301,14 +301,11 @@ mod tests {
     use super::*;
 
     fn get_resource_path(resource_filename: &str) -> String {
-        PathBuf::from(env!("CARGO_MANIFEST_DIR"))
-            .join("resources")
-            .join("tests")
-            .join("server_config")
-            .join(resource_filename)
-            .to_str()
-            .expect("Invalid resource file name.")
-            .to_string()
+        format!(
+            "{}/resources/tests/server_config/{}",
+            env!("CARGO_MANIFEST_DIR"),
+            resource_filename
+        )
     }
 
     fn set_data_dir_path(server_config: &mut ServerConfig, path: PathBuf) {
--- a/quickwit-config/src/server_config.rs
+++ b/quickwit-config/src/server_config.rs
@@ -45,10 +45,6 @@ pub struct IndexerConfig {
     pub rest_listen_address: String,
     #[serde(default = "IndexerConfig::default_rest_listen_port")]
     pub rest_listen_port: u16,
-    #[serde(default = "default_listen_address")]
-    pub grpc_listen_address: String,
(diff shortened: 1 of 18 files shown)


quickwit-config/src/server_config.rs
```diff
@@ -175,19 +147,35 @@ impl SearcherConfig {
...
+            self.rest_listen_port + 2
         ))
     }
 
     pub fn peer_socket_addrs(&self) -> anyhow::Result<Vec<SocketAddr>> {
-        self.peer_seeds.iter().map(socket_addr_from_str).collect()
+        let peer_socket_addrs: Vec<SocketAddr> = self
+            .peer_seeds
+            .iter()
+            .flat_map(|peer_seed| {
```
guilload, line 167: De we want to log each address that does not resolve?

quickwit-serve/src/lib.rs
```diff
@@ -96,13 +96,7 @@ pub async fn run_searcher(
         host_key,
         searcher_config.discovery_socket_addr()?,
     )?);
-    let rest_socket_addr = searcher_config.rest_socket_addr()?;
-    for peer_socket_addr in searcher_config
-        .peer_socket_addrs()?
-        .into_iter()
-        .filter(|peer_socket_addr| *peer_socket_addr != rest_socket_addr)
```
guilload, line 103: I moved the check if in `Cluster::add_peer_node`.

quickwit-serve/src/lib.rs
```diff
@@ -96,13 +96,7 @@ pub async fn run_searcher(
...
-        .map(http_addr_to_swim_addr)
```
guilload, line 104: The translation is now performed in the accessor `peer_socket_addrs()`.

quickwit-config/src/server_config.rs
fulmicoton: yes probably.

fulmicoton: So here we force the peer to use the same listen port as "self".
It will be painful when we want to test stuff.

I suggest we all putting a port in here. 
That port would be interpreted as the baseport.

It feels lame since, the first way this address will be used is the UDP swim stuff.
We could do 
base port = REST port = UDP Port. 

(Technically we could do GRPC  on the base port too and route based on the application type.)

fulmicoton (dismissed):
See comments

quickwit-serve/src/lib.rs
fulmicoton: love it

fmassot (approved):
LGTM

PR #870, opened 2021-12-10
Removed special _tags field

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/850

- Removed special _tags fields
- Adde restrictions to tags field type

### How was this PR tested?
unit test


fmassot (changes requested):
Can you handle also the restriction on field types? I added a comment on the related issue.

fulmicoton (approved):

fmassot (approved):

PR #882, opened 2021-12-10
Add a function to run connectivity checks for an given index.

### LINKED ISSUE #865: Refactor connectivy checks usage
Currently, our connectivity checks are not really easy to use.

Firstly, we always need to instantiate a metastore before doing anything else. It must be valid otherwise we cannot fetch the index metadata. 
So now we fetch the metadata and it means 2 things: the metastore is working and the index exists.

Put it simply, we can remove the metastore and the check on the existence of the index from the checklist. Once we have the index metadata, we can check the storage and finally the sources.

Another question arises when using the CLI, do we really want to print the nice report every time for every command?  As suggested by @guilload it could make sense to print the report in a dedicated command like this `quickwit index check --index-id`.

fmassot:
Fix #865

### DIFF
--- a/quickwit-cli/src/lib.rs
+++ b/quickwit-cli/src/lib.rs
@@ -21,6 +21,11 @@ use std::time::Duration;
 
 use anyhow::bail;
 use once_cell::sync::Lazy;
+use quickwit_common::run_checklist;
+use quickwit_config::SourceConfig;
+use quickwit_indexing::check_source_connectivity;
+use quickwit_metastore::MetastoreUriResolver;
+use quickwit_storage::quickwit_storage_uri_resolver;
 use regex::Regex;
 
 pub mod index;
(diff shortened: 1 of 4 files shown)


quickwit-cli/src/lib.rs
```diff
@@ -55,6 +60,43 @@ pub fn parse_duration_with_unit(duration_with_unit_str: &str) -> anyhow::Result<
     };
 }
 
+/// Run connectivity checks for a given metastore_uri and index id.
```
guilload, line 63:
```
/// Runs connectivity checks for a given `metastore_uri` and `index_id`.
/// Optionaly, it takes a `SourceConfig` that will be checked instead
/// of the index's sources.
```

guilload (approved):

PR #884, opened 2021-12-11
Better errors when checking source connectivity when quickwit not com…

### LINKED ISSUE #859: Unrecognized source
When using a kafka source on a quickiwt that is  not compiled with the kafka feature, we return an error that says "No check implemented for source type `{}`",`

A nicer error could be to return that quickwit has not been compiled with the right option.

fmassot:
…piled with kafka feature and fail when source type is unknown.

Fix #859

### DIFF
--- a/quickwit-config/resources/tests/server_config/server.yaml
+++ b/quickwit-config/resources/tests/server_config/server.yaml
@@ -1,5 +1,5 @@
 version: 0
-metastore_uri: file:///Users/fmassot/Documents/quickwit/repos/quickwit-2
+metastore_uri: postgres://username:password@host:port/db
 indexer:
   data_dir_path: /opt/quickwit/data
   rest_listen_address: 0.0.0.0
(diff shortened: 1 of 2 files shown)


quickwit-config/resources/tests/server_config/server.yaml
```diff
@@ -1,5 +1,5 @@
 version: 0
-metastore_uri: postgres://username:password@host:port/db
+metastore_uri: file:///Users/fmassot/Documents/quickwit/repos/quickwit-2
```
fulmicoton, line 2: ?

fulmicoton, line 2:
```suggestion
metastore_uri: postgres://username:password@host:port/db
```

fulmicoton (approved):
Approved but you need to revert the change in the conf. 

Also: it might be nice to keep these goodfirstissue issues.

PR #886, opened 2021-12-13
Fix 823 ryad

fulmicoton:
### Description

### How was this PR tested?

### DIFF
--- a/quickwit-index-config/src/tag_pruning.rs
+++ b/quickwit-index-config/src/tag_pruning.rs
@@ -1,0 +1,223 @@
+/// Copyright (C) 2021 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+use tantivy::query::QueryParserError as TantivyQueryParserError;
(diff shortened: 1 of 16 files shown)


quickwit-index-config/src/tag_pruning.rs
```diff
@@ -0,0 +1,223 @@
...
+    match user_input_ast {
+        UserInputAst::Clause(sub_queries) => {
+            let conditions: Vec<TagFiltersAST> = sub_queries
+                .iter()
+                .flat_map(|(occ, cond)| {
+                    collect_tag_filters(cond, tag_field_names)
+                        .and_then(|c| wrap_with_occur(&occ.unwrap_or(Occur::Should), c))
+                })
+                .collect();
+            Some(TagFiltersAST::Or(conditions))
```
fulmicoton, line 147: I don't think this code is correct.

`[(Must, "a"), (Must, "b")]` here ends up interpreted as  `or([   and(["a"]), and(["b"]) ])`
when it should actually be `and([a. b])`

fulmicoton: ... although to be honest, it may have been the intended spec.As long as the predicate returned is implied by the query, the filter is technically correct. 

If this is the intended behavior, some comments could have been helpful here.

r-zenine: @fulmicoton, 
You are right in your diagnosis the `[(Must, "a"), (Must, "b")]` should yield an `and([a. b])`. 
I missed that point. I'll work on it to fix that. 

Thank you for spotting it

fmassot:
great stuff here. @evanxg852000 can you take over the metastore part? 
My approach would be: 
- do that in one commit
- change the metastore signature so that we can get splits with an argument of`TagFiltersAST`
- only cares about the PostgreSQL implementation because it's the most annoying part.

fulmicoton:
@evanxg852000  @r-zenine I implemeted the ast -> diesel predicate function, and fixed the compilation.

Remaining is:
- [ ] Switch to the ideas explained in #887.
It simplifies the extraction of the tags from the userinputast.
We do not need to remove the literal that are not tags, and I think we don't need the Uninformative option I think after the change. The rest of the logic stays identical.
It changes the tag list in the metastore on the packaging side.
It changes the expansion logic on the query side.
- [ ] implement the evaluation of the TagFiltersAST against a list of tags.
- [ ] make sure everything is plugged together (See TODO)
- [ ] clean up where necessary
- [ ] more unit tests if necessary

fulmicoton:
Closing ... The PR is moved to #910 that has been squashed and rebased.
(Original authors are still attached.)

PR #888, opened 2021-12-13
Replace String & Path with Uri type

evanxg852000:
### Description
create Uri type to replace String & Path
closes https://github.com/quickwit-inc/quickwit/issues/834 https://github.com/quickwit-inc/quickwit/issues/854

### How was this PR tested?
unittest and cli runs


evanxg852000:
@fmassot @fulmicoton Please see the discussion on https://github.com/quickwit-inc/quickwit/issues/834

quickwit-common/src/uri.rs
```diff
@@ -18,46 +18,89 @@
...
+        let (protocol, mut path) = match uri.split_once(PROTOCOL_SEPARATOR) {
+            None => (FILE_PROTOCOL, uri.to_string()),
+            Some((protocol, path)) => (protocol, path.to_string()),
+        };
+
+        if protocol == FILE_PROTOCOL {
+            let current_dir =
+                env::current_dir().context("Could not fetch the current directory.")?;
+
+            if path.starts_with('~') {
```
fulmicoton, line 53: We don't handle `~username` correctly here.

quickwit-common/src/uri.rs
```diff
@@ -18,46 +18,89 @@
...
+        let (_, path) = self
+            .uri
+            .split_once(PROTOCOL_SEPARATOR)
+            .expect("expect Uri to have a protocol.");
+        Path::new(path)
     }
 
-    Ok(format!("{}://{}", protocol, path))
+    /// Resturn the full uri.
+    pub fn to_str(&self) -> &str {
```
fulmicoton, line 95: We could impl AsRef

quickwit-common/src/uri.rs
```diff
@@ -18,46 +18,89 @@
...
-    if protocol == FILE_PROTOCOL {
-        if path.starts_with('~') {
-            let home_dir_path = home::home_dir()
-                .context("Could not fetch the home directory.")?
+const PROTOCOL_SEPARATOR: &str = "://";
+
+/// Encapsulates the URI type.
+#[derive(Debug, PartialEq, Eq)]
+pub struct Uri {
+    uri: String,
```
fulmicoton, line 34: I like Adrien's idea to keep the protocol separator.

quickwit-common/src/uri.rs
```diff
@@ -18,46 +18,89 @@
...
+            .split_once(PROTOCOL_SEPARATOR)
+            .expect("expect Uri to have a protocol.");
+        protocol
+    }
 
-        path = normalize_path(Path::new(&path))
-            .to_string_lossy()
-            .to_string();
+    /// Returns the uri path.
+    pub fn path(&self) -> &Path {
```
fulmicoton, line 86: We should return an `Option<Path>` here

fulmicoton (changes requested):
Minor changes requested

quickwit-common/src/uri.rs
```diff
@@ -18,46 +18,97 @@
...
+            let current_dir =
+                env::current_dir().context("Could not fetch the current directory.")?;
+
+            if path.starts_with('~') {
+                let home_dir_path = home::home_dir()
+                    .context("Could not fetch the home directory.")?
+                    .to_string_lossy()
+                    .to_string();
+
+                let user_name_dir = format!("~{}", whoami::username());
```
fulmicoton, line 60: I think the ~username stuff is still not correct.
Note that we do not need to support it... but I don't want us to do random stuff on it
so your initial code was doing 

if path.startswith("~") {
   // ... logic to do the replace.
}
that would produce very weird and unpredictable result if someone used the syntax above.
~evance/ on my computer would become /home/fulmicotonevance /home/fulmicoton/evance depending on the body.
Your new code only handles ~current_user
We can just keep the previous code, and return an explicit error ("This path syntax: <user_input> is not supported.") if the input starts by ~ and is not followed by a separator.

quickwit-common/src/uri.rs
```diff
@@ -99,73 +150,100 @@ mod tests {
...
 
         assert_eq!(
-            normalize_uri("file://..")?,
+            Uri::try_new("file://..")?.to_string(),
             format!("file://{}", current_dir.parent().unwrap().display())
         );
 
         assert_eq!(
-            normalize_uri("s3://home/hommer/docs/dognuts")?,
+            Uri::try_new("s3://home/hommer/docs/dognuts")?.to_string(),
```
fulmicoton, line 241: nitpick.
s/hommer/homer/g :D

quickwit-common/src/uri.rs
```diff
@@ -18,46 +18,97 @@
...
+    }
+
+    /// Returns the uri protocol.
+    pub fn protocol(&self) -> &str {
+        &self.uri[..self.protocol_idx]
+    }
+
+    /// Returns the uri path.
+    /// If protocol is `file://`, returns [`Some(Path)`] otherwise returns [`None`].
+    pub fn path(&self) -> Option<&Path> {
```
fulmicoton, line 93: Maybe we should rename it filepath? (path is also the name of a component in an uri.)

fulmicoton (changes requested):
inline comment

quickwit-common/src/uri.rs
evanxg852000: :+1: Yes and that's why I was suggesting it should always return something rather than Option. But with `filepath` it's very clear what we want.

quickwit-common/src/uri.rs
```diff
@@ -18,46 +18,93 @@
...
+        };
+
+        if protocol == FILE_PROTOCOL {
+            let current_dir =
+                env::current_dir().context("Could not fetch the current directory.")?;
+
+            if path.starts_with('~') {
+                if path != "~" && !path.starts_with("~/") {
+                    bail!("This path syntax `{}` is not supported.", uri);
+                }
```
fmassot, line 57: I'm struggling to understand the condition.

Here we need to sastify 3 conditions to raise the error:
- `path.starts_with('~')`
- `path != "~"`
- `!path.starts_with("~/")`

But I don't have a better idea without using a regex.... can you add a comment just to explain the condition?

evanxg852000:
```rust
// We only accept `~` and `~/` which are aliases to the home directory.
// If there is something following `~` and it's not `/`, we bail out.
if path.chars()
    .nth(1)
    .map(|second_character| second_character != '/')
    .unwrap_or(false) {
    bail!("This path syntax `{}` is not supported.", uri);
}
```

fmassot (approved):

fulmicoton (approved):

PR #898, opened 2021-12-15
Docs tutorial section

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/876

### How was this PR tested?
run through the steps


evanxg852000:
@fmassot I found it hard for having an out-of-the-box config for `hdfs` tutorial, so I went with the user creating config files directly from the commands. If that sounds ok to you, we can remove the files:
- config/tutorials/hdfs-logs/searcher-1.yaml
- config/tutorials/hdfs-logs/searcher-2.yaml
- config/tutorials/hdfs-logs/searcher-3.yaml

docs/tutorials/tutorial-hdfs-logs.md
```diff
@@ -30,76 +29,62 @@ Let's download and install Quickwit.
...
-            ]
-        }
-    ]
-}
-```
+Let's make use of the default configuration file provided in quickwit installation throughout this guide by setting the `QW_CONFIG` environment variable.
 
-Now we can create the index with the new command (assuming you create it in your current directory):
+```bash
+export QW_CONFIG=./config/quickwit.yaml
```
fmassot, line 83: Add a small comment just to say that all commands need quickwit config and it's a good practice to have it in this env variable.

config/tutorials/hdfs-logs/searcher-1.yaml
```diff
@@ -1,5 +1,7 @@
 version: 0
 node_id: searcher-1
+metastore_uri: s3://path/to/bucket/indexes # change this to your S3 bucket
```
fmassot, line 3: I prefer to commit only things that can work out of the box. 
We can add `metastore_uri` and `default_index_root_uri` in comment.

docs/tutorials/tutorial-hdfs-logs-distributed-search-aws-s3.md
```diff
@@ -45,116 +46,152 @@ Let's create an index configured to receive these logs.
...
+  store_source: true
+
+indexing_settings:
+  timestamp_field: timestamp
+
+search_settings:
+  default_search_fields: [severity_text, body]
+```
 
+To index our dataset and serve it on a Quickwit cluster, we first need to create a few search nodes configuration files. Please, define your S3 bucket path now to ease the config files creation process.
```
fmassot, line 86: I think it's important to do things step by step for the new user.
Here there is something confusing: we are defining searcher's conf files but we want to first index. 

I think the good way to go is to say: let's first create and ingest our index on S3. For that we need to define a quickwit config file. We should not talk about searchers immediately.

Once we have indexed it then we can think about how to set up a search cluster.

The tutorial was about configuring an cluster on AWS but your configuration will not work with `rest_listen_address: 127.0.0.1` and the different ports used here. I added these config files for a local search cluster :)

If we are on several machines, we can even think about having only one config file named `quickwit.yaml` that sets the right IPs directly.

fmassot (changes requested):
We need to rethink a little bit the aws distributed search. 

I think we should use the searcher config 1,2,3 for the local hdfs logs tutorial so that we can show how easy it is to create a cluster.

config/quickwit.yaml
```diff
@@ -15,9 +15,9 @@ version: 0
 #
 # Path to directory where data is persisted. Default to `./qwdata`.
 #
-#data_dir: /path/to/data
+#data_dir_path: /path/to/data
```
fmassot, line 18: I think we should keep here `data_dir` to remain consistent with the env variable too. We need to modify `QuickwitConfig` so that it can parse `data_dir`

config/tutorials/hdfs-logs/searcher-1.yaml
evanxg852000: I think, my proposition is that to even get rid of these files. what do you think? right now it's not used in the tutorial

config/quickwit.yaml
evanxg852000: ok

config/tutorials/hdfs-logs/searcher-1.yaml
evanxg852000: That's why I proposed their removal but it seems we have found a use case for them in the local hdfs cluster.

fmassot: Yes exactly :)

docs/getting-started/installation.md
```diff
@@ -18,6 +18,24 @@ Downloads `.tar.gz`:
 
 Checkout all builds on [github](https://github.com/quickwit-inc/quickwit/releases)
 
+### Note on external dependencies
+
+Quickwit depends on the following external libraries to work properly:
```
fmassot, line 23:
```suggestion
Quickwit depends on the following external libraries to work correctly:
```

docs/getting-started/installation.md
```diff
@@ -18,6 +18,24 @@ Downloads `.tar.gz`:
...
+- `libpq`: the postgres client library.
```
fmassot, line 24:
```suggestion
- `libpq`: the Postgres client library.
```

docs/getting-started/installation.md
```diff
@@ -18,6 +18,24 @@ Downloads `.tar.gz`:
...
+- `libssl`: the industry defacto cryptography library. 
+These libraries can be installed on your system using the native package manager. 
+On ubuntu for instance, you can install these dependencies using the following command:
```
fmassot, line 27:
```suggestion
On Ubuntu for instance, you can install these dependencies using the following command:
```

docs/getting-started/installation.md
```diff
@@ -18,6 +18,24 @@ Downloads `.tar.gz`:
...
+
+```bash
+apt-get -y update && apt-get -y install libpq-dev libssl-dev
+```
+
+:::note
+
+Quickwit static binary packages are also provide as `musl` builds. These packages don't require you to install any external library and can be automatically picked during installation on your system if the required libc version is not present. You can also download and manually install a static binary package.
```
fmassot, line 35:
```suggestion
Quickwit static binary packages are also provided as `musl` builds. These packages don't require you to install any external library and can be automatically picked during installation on your system if the required libc version is not present. You can also download and manually install a static binary package. 
```

docs/tutorials/tutorial-hdfs-logs.md
```diff
@@ -163,6 +163,55 @@ curl -v 'http://127.0.0.1:7280/api/v1/hdfslogs/search?query=severity_text:ERROR&
 
 It should return 6 hits faster as Quickwit will query fewer splits.
 
+## Distributed search
+
+Now that we have indexed our dataset and can do a local search, let's show how easy it is to start a distibuted search cluster with Quickwit. We will be launching a three nodes cluster.
+First let's download the ssearche node configuration files:
```
fmassot, line 171:
```suggestion
First, let's download the Searcher's configuration files:
```

docs/tutorials/tutorial-hdfs-logs.md
```diff
@@ -163,6 +163,55 @@ curl -v 'http://127.0.0.1:7280/api/v1/hdfslogs/search?query=severity_text:ERROR&
 
 It should return 6 hits faster as Quickwit will query fewer splits.
 
+## Distributed search
+
+Now that we have indexed our dataset and can do a local search, let's show how easy it is to start a distibuted search cluster with Quickwit. We will be launching a three nodes cluster.
```
fmassot, line 170:
```suggestion
Now that we have indexed our dataset and can do a local search, let's show how easy it is to start a distributed search cluster with Quickwit. We will be launching a three-node cluster.
```

docs/tutorials/tutorial-hdfs-logs.md
```diff
@@ -163,6 +163,55 @@ curl -v 'http://127.0.0.1:7280/api/v1/hdfslogs/search?query=severity_text:ERROR&
...
+First let's download the ssearche node configuration files:
+
+```bash
+for i in {1..3}; do 
+curl -o searcher-$i.yaml https://raw.githubusercontent.com/quickwit-inc/quickwit/main/config/tutorials/hdfs-logs/searcher-$i.yaml
+done
+```
+
+Once the configuartion files are downloade, start a searcher node for each of the respective configuration file in a different terminal window.
```
fmassot, line 179:
```suggestion
Once the configuration files are downloaded, start a searcher node for each of the respective configuration files in a different terminal window.
```

docs/tutorials/tutorial-hdfs-logs.md
```diff
@@ -163,6 +163,55 @@ curl -v 'http://127.0.0.1:7280/api/v1/hdfslogs/search?query=severity_text:ERROR&
...
+
+:::
+
+Let's execute a simple query that returns only `ERROR` entries on field `severity_text` on one of our searcher node:
+
+```bash
+curl -v 'http://127.0.0.1:7280/api/v1/hdfs_logs/search?query=severity_text:ERROR
+```
+
+Now that you have a search cluster running, ideally, you will want to load balance external requests to it. This can quickly be done using appropriate tool. with a load balancer added, you can now play with your cluster, kill processes randomly, add/remove new instances.
```
fmassot, line 214: I don't think we need to add that as this is just a local cluster. This makes more sense on AWS.

docs/getting-started/installation.md
```diff
@@ -18,6 +18,24 @@ Downloads `.tar.gz`:
 
 Checkout all builds on [github](https://github.com/quickwit-inc/quickwit/releases)
 
+### Note on external dependencies
```
fmassot, line 21: Nice section. It would be nice to put the link of the static binary packages above.

docs/tutorials/tutorial-hdfs-logs-distributed-search-aws-s3.md
```diff
@@ -27,144 +27,201 @@ Before using Quickwit with an object storage, check out our [advice](../administ
...
 
 ```bash
-quickwit new --index-uri s3://path-to-your-bucket/hdfs_logs --index-config-path ./hdfslogs_index_config.json
+./quickwit index create --index hdfslogs --index-config ./hdfslogs_index_config.yaml --config ./config.yaml
 ```
 
 :::note
 
-This step can be executed on your local machine. The `new` command creates the index locally and will then only upload a json file `quickwit.json` to your bucket at `s3://path-to-your-bucket/hdfs_logs/quickwit.json`. 
+This step can also be executed on your local machine. The `create` command creates the index locally and then uploads a json file `quickwit.json` to your bucket at `s3://path-to-your-bucket/hdfslogs/quickwit.json`.
```
fmassot, line 99: now it's `metastore.json` I think, no more `quickwit.json`

docs/tutorials/tutorial-hdfs-logs-distributed-search-aws-s3.md
```diff
@@ -27,144 +27,201 @@ Before using Quickwit with an object storage, check out our [advice](../administ
...
+# configuration for our first node
+echo "version: 0
+node_id: searcher-1
+metastore_uri: ${S3_PATH}
+default_index_root_uri: ${S3_PATH}
+searcher:
+  rest_listen_address: 0.0.0.0
+  rest_listen_port: 7100
+  peer_seeds:
+    - ${IP_NODE_2}:7200 # searcher-2
```
fmassot, line 160: I would avoid changing the ports. It's clearer to have all instances on the same ports.

fmassot: by the way, using the default port would be nice, so that we don't need to add the port in the peer_seeds.

docs/tutorials/tutorial-hdfs-logs-distributed-search-aws-s3.md
```diff
@@ -27,144 +27,201 @@ Before using Quickwit with an object storage, check out our [advice](../administ
...
+# configuration for our second node
+echo "version: 0
+node_id: searcher-2
+metastore_uri: ${S3_PATH}
+default_index_root_uri: ${S3_PATH}
+searcher:
+  rest_listen_address: 0.0.0.0
+  rest_listen_port: 7200
+  peer_seeds:
+    - ${IP_NODE_1}:7100 # searcher-1
```
fmassot, line 190: if you want to simplify the tutorial, you don't need to declare all peer_seeds. Declaring peer seeds on node 1 is sufficient.

fmassot: And ideally, do not change `rest_listen_port`

docs/tutorials/tutorial-hdfs-logs-distributed-search-aws-s3.md
```diff
@@ -27,144 +27,201 @@ Before using Quickwit with an object storage, check out our [advice](../administ
...
+You will see in the terminal the confirmation that the instance has joined the existing cluster. Example of such a log:
+
+```
+INFO quickwit_cluster::cluster: Joined. node_id="searcher-1" remote_host=Some(18.222.142.42:7100)
+```
+
+Now we can query one of our instance directly by issuing http requests to the node rest API endpoint.
+
+```
+curl -v 'http://{second-ec2-instance-public-ip}:7200/api/v1/hdfslogs/search?query=severity_text:ERROR
```
fmassot, line 219: nice reuse of env variable!

docs/tutorials/tutorial-hdfs-logs.md
```diff
@@ -30,76 +29,64 @@ Let's download and install Quickwit.
...
+curl -o hdfslogs_index_config.yaml https://raw.githubusercontent.com/quickwit-inc/quickwit/main/config/tutorials/hdfs-logs/index-config.yaml
 ```
 
-The index config defines 5 fields: `timestamp`, `severity_text`, `body`, and two object fields
-for the nested values `resource` and `attributes`. 
-It also sets the [default search field](../reference/index-config.md) and a timestamp field. 
-This timestamp field will be used by Quickwit for [splits pruning](../overview/architecture.md) at query time to boost search speed. Check out the [index config docs](../reference/index-config.md) for details.
+The index config defines four fields: `timestamp`, `severity_text`, `body`, and one object field
+for the nested values `resource.service` . It also sets the `default_search_fields`, the `tag_fields`, and the `timestamp_field`.The `timestamp_field` and `tag_fields` are used by Quickwit for [splits pruning](../overview/architecture.md) at query time to boost search speed. Check out the [index config docs](../reference/index-config.md) for more details.
```
fmassot, line 45:
```suggestion
for the nested values `resource.service` . It also sets the `default_search_fields`, the `tag_fields`, and the `timestamp_field`. The `timestamp_field` and `tag_fields` are used by Quickwit for [splits pruning](../overview/architecture.md) at query time to boost search speed. Check out the [index config docs](../reference/index-config.md) for more details.
```

docs/tutorials/tutorial-hdfs-logs.md
```diff
@@ -30,76 +29,64 @@ Let's download and install Quickwit.
...
-                    "name": "class",
-                    "type": "text",
-                    "tokenizer": "raw"
-                }
-            ]
-        }
-    ]
-}
-```
+Let's make use of the default configuration file provided in quickwit installation throughout this guide by setting the `QW_CONFIG` environment variable.
```
fmassot, line 80:
```suggestion
Let's make use of the default configuration file provided in Quickwit installation throughout this guide by setting the `QW_CONFIG` environment variable.
```

fmassot (approved):
Sounds good, you can check the minor comments.

docs/tutorials/tutorial-hdfs-logs-distributed-search-aws-s3.md
evanxg852000: Hmm, my intent was more like the user to replace with the real IP since he want to run this on his host, not on ec2 instance. I will switch to `${IP_NODE_2}`

Issue #852, opened 2021-12-08
Update Packaging & installation script


fmassot:
# Old way 
`curl -L https://install.quickwit.io | sh` just download quickwit binary and that's it.

# New way

We want to do more today to help the user starts a server quickly with a basic config.

## User journey 
1. `curl -L https://install.quickwit.io | sh` 
2. Ask the user if he is ok with the default installation or change default parameters
3. Install
  - create directory `$HOME/$QUICKWIT_HOME`, by default it's `quickwit-version-number`
  - in the directory, we put the following files:
    - the binary `bin/quickwit`
    - a default server config `config/server.toml`. The default datadir can be `/var/lib/quickwit/data` and logs in `/var/log/quickwit`
    - a tutorials directory which contains the tutorials config files 

The default installation parameters are:
-  $QUICKWIT_HOME = $HOME/`quickwit-version-number` 
- Add `$QUICKWIT_HOME/bin` in his path

guilload:
Looking at what other programs are doing in a user's home directory (`cargo`, `rustup`, `steam`, ...): I'd like to amend the spec:
- Quickwit home directory: `$HOME/.quickwit`
- Bin directory: `$HOME/.quickwit/bin`
- Config directory: `$HOME/.quickwit/config`
- Install release x.y in `$HOME/.quickwit/quickwit-x.y`
- Symlink  `$HOME/.quickwit/quickwit-x.y/quickwit` to `$HOME/.quickwit/bin/quickwit`

Optionally:
- Remove earlier versions
- Offer to symlink `$HOME/.quickwit/bin/quickwit` to `/usr/local/bin/quickwit`. I don't think our binary is so important that it deserves to be added to the PATH.

fmassot:
I like your modifications. I suggested at the beginning to go for the hidden directory but I found hidden directory makes sense especially for mainstream binaries.

fulmicoton:
I'll add we need a strong story for the user who wants to run two server on the same machine. This is very common for testing / development.

I'd go for the following: what we deliver is a .tar.gz package that can be unpacked anywhere.


```
quickwit-VERSION/ <--- That's quickwit home
      bin/
           quickwit   <--- the quickwit binary. 
                                 If QUICKWIT_HOME env variable is not set, use its `own path/../..`
      conf/    <--- I would have put the node local config stuff. What you have today in ServerConfig, IndexerConfig. The conf could give path relative to the QUICKWIT_HOME.
[...]

PR #899, opened 2021-12-15
Install script update

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/852 

Updates the installation script and the way we bundle the binary with config files.

### How was this PR tested?
Tested the installation with temporary release


.github/workflows/release.yml
```diff
@@ -76,20 +75,19 @@ jobs:
...
-      - name: Rename artifact
+      - name: Bundle archive
         if: ${{ github.ref != 'refs/heads/make-binaries-from-ci' }}
-        run: mv target/x86_64-unknown-linux-gnu/release/quickwit ${{ env.ASSET_FULL_NAME }}
-      - name: Compress artifact
-        if: ${{ github.ref != 'refs/heads/make-binaries-from-ci' }}
-        run: tar -czvf ${{ env.ASSET_FULL_NAME }}.tar.gz ${{ env.ASSET_FULL_NAME }}
-      - name: Upload artifact
+        run: |
+          make archive ARCHIVE_NAME=${{ env.ASSET_FULL_NAME }}.tar.gz \
```
fmassot, line 88: cool!!! you found a way to use make!

install.sh
```diff
@@ -52,6 +52,42 @@ main() {
     install_from_archive
 }
 
+confirm_install_options() {
+    local _version="$1"
```
fmassot, line 56: As you may see the ongoing dicussion here #852, we have changed a little bit the installation script. We are going to simplify this and follow elasticsearch `tar.gz` install.

fmassot: Let's wait until tomorrow before updating the script.

Makefile
```diff
@@ -58,5 +58,15 @@ build:
 		;; \
 	esac
 
+.PHONY: archive
+archive:
```
fmassot, line 62: A comment would be great to understand we need several env variables

fmassot (approved):
Great! I did not test the install script but we can do it in a few days.

Makefile
fmassot: Add the version and the right name for the folder would be great too.

Issue #839, opened 2021-12-03
Search servers should (optionally?) poll the S3 metastore for updates


fulmicoton:
Right now, if we have 
- one indexer
- search servers
and we are using the S3 metastore, the search server will load the S3 metastore state on startup and never update it.

We could have an option to reload it at regular interval
(once every 10s induce a cost of $0.1 per month per index).

PR #900, opened 2021-12-16
Refactoring of the single file metastore to prepare for #839.

fulmicoton:
### Description
- The lock is now per index.
- Factored together the mutation and the view logic to improve
  proofreadability.
- Discarding entry in the cache after an error.

### How was this PR tested?
Unit test

### Pointers for the code review

All of model logic with no concurrency concern is in the Metadataset object.
It is a copy paste of the existing code.

The mutation and the read logic are then factorized in the `mutate` and `read` methods.

We have two-level locks over our state cache.
One on the map itself, and one per index.

The outter one is almost always accessed as "read" since it is only used to
edit the content of the map.

The index-level one is used to run operation targeting a specific index.
That way, there are no contention when updating different indices, and the polling task 
that we we want to implement in #839 can own the index-level Arc<Mutex>.

In addition, the previous logic was difficult to proofread and contain an intricate bug:
while if a storage returns a success, we are guarantedd that the write was successful, 
a storage returning a failure does not necessarily guarantee that the write did not go through.
In this case, we were ending up with a cache inconsistent with the storage state.
In this PR, we solve the problem by simply invalidating the cache if a write fails.


quickwit-metastore/src/metastore/single_file_metastore/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+
+    /// Returns a valid metadataset that is locked.
+    ///
+    /// This function guarantees that the metadataset has not been
+    /// marked as discarded.
+    async fn metadata_set_locked(
+        &self,
+        index_id: &str,
+    ) -> MetastoreResult<OwnedMutexGuard<MetadataSet>> {
+        loop {
```
evanxg852000, line 120: I think we could  avoid the busy loop with tokio's equivalent of `CondVar` https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html
It will be a bit more complex and less readable though.

evanxg852000, line 120: I thought of busy loop here, but then realized we are `awaiting` which does the same thing as `CondVar`

quickwit-metastore/src/metastore/single_file_metastore/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+                return Ok(metadata_set_lock);
+            }
+        }
+    }
+
+    /// Returns an IndexView for the given index_id.
+    ///
+    /// If this is the first call during this instance for this
+    /// `index_id`, a fetch to the storage will be initiated
+    /// and might triggered an error.
```
evanxg852000, line 133:
```suggestion
    /// and might trigger an error.
```

evanxg852000, line 133:
```suggestion
    /// and might trigger an error.
```

quickwit-metastore/src/metastore/single_file_metastore/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+        // We pick the outer lock here, so that we enter a critical section.
+        let mut per_index_metastores_wlock = self.per_index_metastores.write().await;
+
+        let delete_res = delete_metadata_set(&*self.storage, index_id).await;
+
+        match &delete_res {
+            Ok(()) |
+            // If the index file does not exist, we still need to return an error,
+            // but it makes sense to ensure that the cache is up to date.
+            Err(MetastoreError::IndexDoesNotExist { .. }) => {
```
evanxg852000, line 219: we are guarding against the index file being deleted by an external process right?

quickwit-metastore/src/metastore/single_file_metastore/store_operations.rs
```diff
@@ -0,0 +1,151 @@
...
+    let metadata_path = meta_path(index_id);
+    let exists = storage
+        .exists(&metadata_path)
+        .await
+        .map_err(|storage_err| convert_error(index_id, storage_err))?;
+    Ok(exists)
+}
+
+/// Serializes the metadata set and stores the data on the storage.
+pub(crate) async fn put_metadata_set_to_path(
```
evanxg852000, line 86: Nitpick: since we pass index_id & make the path inside
```suggestion
pub(crate) async fn put_metadata_set(
```

quickwit-metastore/src/metastore/single_file_metastore/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+    delete_metadata_set, fetch_metadata_set, index_exists, put_metadata_set,
+};
+use crate::{
+    IndexMetadata, Metastore, MetastoreError, MetastoreResult, Split, SplitMetadata, SplitState,
+};
+
+/// Single file metastore implementation.
+pub struct SingleFileMetastore {
+    storage: Arc<dyn Storage>,
+    // try a rwlock
```
evanxg852000, line 45:
```suggestion
```

quickwit-metastore/src/metastore/single_file_metastore/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+        }
+
+        let put_result = put_metadata_set(&*self.storage, &metadata_set).await;
+        match put_result {
+            Ok(()) => {
+                *metadata_set_lock = metadata_set;
+                Ok(())
+            }
+            err @ Err(_) => {
+                // metadata_set_mutex.discard();
```
evanxg852000, line 90:
```suggestion
```

quickwit-metastore/src/metastore/single_file_metastore/mod.rs
fulmicoton: nice catch

quickwit-metastore/src/metastore/single_file_metastore/mod.rs
fulmicoton: yes :)

quickwit-metastore/src/metastore/single_file_metastore/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+        // to make sure we return the same instance.
+        //
+        // If there is an error here, note we do not return right away.
+        // That's because we want to observe the property that after one success
+        // all subsequent calls will succeed.
+        let metadata_set_result = fetch_metadata_set(&*self.storage, index_id).await;
+
+        // Here we retake the lock, still no io ongoing.
+        let mut per_index_metastores_wlock = self.per_index_metastores.write().await;
+        if let Some(metadata_set) = per_index_metastores_wlock.get(index_id) {
```
evanxg852000, line 156: Can we comment that at this point the `metadata_set` might be put by another client?

quickwit-metastore/src/metastore/single_file_metastore/mod.rs
fulmicoton: Technically we do not support two processes writing on the singlefilemetastore.

Still it is nice to have a proper definition of a trait like the `Metastore` on that kind of corner case that "should never happen".

evanxg852000 (commented):
It looks good to me & nice work :clap: . locks are always hard to manage.
Just left a few comments and a question.

evanxg852000 (approved):

quickwit-metastore/src/metastore/single_file_metastore/metadata_set.rs
```diff
@@ -0,0 +1,294 @@
+// Copyright (C) 2021 Quickwit, Inc.
```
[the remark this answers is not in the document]
fulmicoton: Agreed it sucks.

It's an object that serves as a model of the all the data associated with an Index and its splits.
It gets serialized into a single file.

Not sure what the perfect name should be...

quickwit-metastore/src/metastore/single_file_metastore/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+};
+use crate::{
+    IndexMetadata, Metastore, MetastoreError, MetastoreResult, Split, SplitMetadata, SplitState,
+};
+
+/// Single file metastore implementation.
+pub struct SingleFileMetastore {
+    storage: Arc<dyn Storage>,
+    // try a rwlock
+    per_index_metastores: RwLock<HashMap<String, Arc<Mutex<MetadataSet>>>>,
```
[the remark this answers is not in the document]
fulmicoton: Not really no. 
The silliness is like it was before.

What we call SingleFileMetastore actually deals with more than one index, and we do not have any mechanism (yet) to ensure that we do not have more than two instances for the same index in the same process. That's on my task list.

What changed is that the thing that holds the data for a single index can now be detached.

quickwit-metastore/src/metastore/single_file_metastore/metadata_set.rs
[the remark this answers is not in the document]
fulmicoton: Opened #909 for the renaming

quickwit-metastore/src/metastore/single_file_metastore/metadata_set.rs
```diff
@@ -0,0 +1,294 @@
...
+                return DeleteSplitOutcome::SplitNotFound;
+            }
+        };
+        match metadata.split_state {
+            SplitState::ScheduledForDeletion | SplitState::Staged => {
+                // Only `ScheduledForDeletion` and `Staged` can be deleted
+                self.splits.remove(split_id);
+                DeleteSplitOutcome::Success
+            }
+            SplitState::Published | SplitState::New => {
```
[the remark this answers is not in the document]
fulmicoton: Oh great point. I was wondering what this was about. That makes sense.

quickwit-metastore/src/metastore/single_file_metastore/mod.rs
fulmicoton: Adding a comment

PR #907, opened 2021-12-16
Fix never read field warning

guilload:


fmassot (approved):
LGTM

guilload:
I think I'm the only one to see this warning because I'm using a freshly cloned repo. The field is used only in the `new` constructor.

Screenshot_20211216_145439

PR #916, opened 2021-12-17
[DISCUSS] Write host key in data dir

guilload:
### Description
To reduce the number of things to configure, I thought we could directly writhe the "host key" in the data dir. Let me know what you think.

I'm also wondering if there is a better way to identify a host than writing an ID in a file?


fulmicoton:
The host key is the thing we use in quickwit-cluster? I think this is a good idea.

tbh I am not sure i ever understand the implication of this key...

fmassot:
> The host key is the thing we use in quickwit-cluster? I think this is a good idea.
tbh I am not sure i ever understand the implication of this key...

The only thing that this `host_id` brings is to have a stable `uuid` after a reboot of a node in a cluster, thus it improves the readability of membership logs. Looking at this makes me think that it's almost useless to store the uuid in a file.

fmassot:
Let's wait until tomorrow before merging this PR. 
In order to set up a local search cluster easily, we would need to add an extra argument to the CLI to override the default data dir.

```bash
# searcher 1
quickwit service run searcher --data-dir data/search-1
# searcher 2
quickwit service run searcher --data-dir data/search-2
```

fulmicoton:
@fmassot I am lost here... With the new packaging system, isn't it just two directories and 
`quickwit service run searcher` run twice with 2 different CWDs?

fulmicoton:
> The only thing that this host_id brings is to have a stable uuid after a reboot of a node in a cluster, thus it improves the readability of membership logs. Looking at this makes me think that it's almost useless to store the uuid in a file.

It would be a tiny bit more than this, if we used this id for cache consistency... but we don't (we use the grpc addr for this).
By cache consistency I mean, if an instance is a tiny bit flaky for a reason and gets disconnected an reconnected in 10s, it is a nice thing to have this instance takes its previous place in the request assignment logic (this thing using rdv hashing today), because the caches  probably have not been entirely evinced yet.

Also... We could consider in the future letting the user labelling the instances in their conf.
(I am not too sure however how to prevent collisions due to a misconfiguration.)

I suggest we don't have a UUID, and we don't have a user defined id, and only use our rpc_address.


What do I mean by external rpc_address?
I think we will have soon to have the user define two socket addr. (well more due to the 3 different protocols, but bear with me).

- listen_address: the interface we listen too
- rpc_address: the address other nodes will call

By default rpc_address refers to the listen address.
If we use a proxy, or docker, ... we need to explicitly state the rpc_address, which should be accessible for all of the other nodes of the network.

We can use the remote addr as a (at least locally in time) unique identity.

fmassot:
> It would be a tiny bit more than this, if we used this id for cache consistency... 

yes! I thought at the beginning that it was used for caching but it's not.

> That's for the future. For the present, what about we just use a cool id to help correlate the different logs?

Something like `{hostname}-{adjective}-{slug}`? So for correlation we have a stable hostname and that would be enough for now.

fmassot:
> @fmassot I am lost here... With the new packaging system, isn't it just two directories and
> `quickwit service run searcher` run twice with 2 different CWDs?

Maybe I'm overthinking this but I thought it's overkill to have multiple directories when you test a local search cluster. Using the same quickwit binary and just passing a host id or a `datadir` sounds sufficient for this use case. And the `datadir` is really useful for the indexing.

fulmicoton:
Ok, so the data dir is the thing that is truely instance specific. I'm ok with that.

fmassot:
Closing as it is addressed by #938

PR #918, opened 2021-12-17
Added update_timestamp field to Index

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/797

### How was this PR tested?
Unit tests

### DIFF
--- a/quickwit-metastore/src/metastore/mod.rs
+++ b/quickwit-metastore/src/metastore/mod.rs
@@ -143,6 +143,7 @@ impl IndexMetadata {
                 "attributes.server.status".to_string(),
             ],
         };
+        let now_timestamp = Utc::now().timestamp();
         Self {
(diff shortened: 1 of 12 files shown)


quickwit-metastore/src/metastore/mod.rs
```diff
@@ -148,6 +151,7 @@ impl IndexMetadata {
             search_settings,
             sources: Vec::new(),
             create_timestamp: Utc::now().timestamp(),
+            update_timestamp: Utc::now().timestamp(),
```
fulmicoton, line 154: nitpick: Can we call now once and use the same value on create_timestamp and update_timestamp?
It could be nice for debugging to be able to identify index htat have never been modified.

(I know we can't have that sweet logic on the default of deserialization on older version but that's ok)

fulmicoton (approved):

PR #920, opened 2021-12-19
Fix tags collection for  numbers (u64, i64, and f64)

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/912
- fixed tag collection logic to properly handle u64, i64, and f64 types

### How was this PR tested?
unit tests


fulmicoton (approved):
good job!

PR #925, opened 2021-12-20
Interprets query terms as a conjunction (AND).

### LINKED ISSUE #924: Conjunction by default?
Tantivy interprets several keywords as a disjunction (OR) by default. 

This makes sense in a document search setting where ranking is supposed to save the day.
This is  a setting where user want the top 10 best documents, and they are fine if the count of matched is pretty useless.

For logs however, people usually use full text search as a filter.
Conjunction would be a more sensible default setting.

(People can still express disjunction using boolean keywords)

fulmicoton:
Closes #924


fulmicoton:
@fmassot coudl you review?

fmassot:
@fulmicoton see my comment here -> https://github.com/quickwit-inc/quickwit/issues/924#issuecomment-997849372

fmassot (approved):

PR #927, opened 2021-12-20
Added polling for the SingleFileMetastore.

### LINKED ISSUE #839: Search servers should (optionally?) poll the S3 metastore for updates

fulmicoton:
This adds the possibility to configure a polling period for a given SingleFileMetastore by appending a uri fragment to tis uri like this: `s3://my_bucket/blabla#period=30s`

Closes #839

### DIFF
--- a/quickwit-metastore/src/metastore/single_file_metastore/single_file_metastore_factory.rs
+++ b/quickwit-metastore/src/metastore/single_file_metastore/single_file_metastore_factory.rs
@@ -18,9 +18,12 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::sync::Arc;
+use std::time::Duration;
 
 use async_trait::async_trait;
+use once_cell::sync::OnceCell;
 use quickwit_storage::{quickwit_storage_uri_resolver, StorageResolverError, StorageUriResolver};
+use regex::Regex;
 
 use crate::{
     Metastore, MetastoreError, MetastoreFactory, MetastoreResolverError, SingleFileMetastore,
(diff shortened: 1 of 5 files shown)


quickwit-metastore/src/metastore/single_file_metastore/single_file_metastore_factory.rs
```diff
@@ -40,27 +43,81 @@ impl Default for SingleFileMetastoreFactory {
     }
 }
 
+fn extract_polling_period_from_uri(uri: &str) -> (String, Option<Duration>) {
```
guilload, line 46: I'm not fond of this. Unfortunately, this adds another way to configure quickwit, and we can't detect issues at config parsing time unless we add some additional validation logic.

I hope in the future we can get rid of this and resolve `Metastore` and `Storage` with an optional config object:
`Metastore::resolve("s3://...", Some(MetastoreConfig { ... }))`.

quickwit-metastore/src/metastore/single_file_metastore/single_file_metastore_factory.rs
```diff
@@ -40,27 +43,81 @@ impl Default for SingleFileMetastoreFactory {
...
+    static URI_FRAGMENT_PATTERN: OnceCell<Regex> = OnceCell::new();
+    if let Some(captures) = URI_FRAGMENT_PATTERN
+        .get_or_init(|| Regex::new("(.*)#period=([1-9][0-9]{0,8})s").unwrap())
```
guilload, line 49: `polling_period`?

guilload (approved):
LGTM.

A few words on naming. `polling_interval` is more commonly used: a search on `polling_interval` vs. `polling_period` on all GitHub yields [837k hits](https://github.com/search?q=polling_interval&type=code) vs. [33k](https://github.com/search?q=polling_period&type=code).

quickwit-metastore/src/metastore/single_file_metastore/single_file_metastore_factory.rs
fulmicoton: Yes it is not great. Right now if the users gets the DSL wrong we don't spot anything & don't log anything.

quickwit-metastore/src/metastore/single_file_metastore/single_file_metastore_factory.rs
fulmicoton: renamed polling_interval

fulmicoton:
@guilload Renaming to polling_interval

quickwit-metastore/src/metastore/single_file_metastore/single_file_metastore_factory.rs
fulmicoton: renaming polling_interval

PR #934, opened 2021-12-20
unplug sorting from index config

### LINKED ISSUE #908: Disable sorting the doc id index.
For the 0.2 release, it seems that it is wise to postpone to the next release the configuration of index sorting. 

Pros:
- we don't want the user to pay extra cost for sorting as there are no clear usage of it today

Cons:
- as @PSeitz there is one thing that we lose: it is the compression of the sorted fast field values. We have no pressure coming from users for that right now, so this is reasonable to delay index sorting to the next release.

PSeitz:
fixes #908

### Description

### How was this PR tested?


quickwit-metastore/src/metastore/mod.rs
```diff
@@ -281,15 +277,11 @@ impl From<UnversionedIndexMetadata> for IndexMetadata {
...
-        let (sort_field, sort_order) = match unversioned.index_config.sort_by {
-            SortBy::DocId => (None, None),
-            SortBy::FastField { field_name, order } => (Some(field_name), Some(order)),
-        };
         let indexing_settings = IndexingSettings {
             demux_field: unversioned.index_config.demux_field_name,
             timestamp_field: unversioned.index_config.timestamp_field_name,
-            sort_field,
-            sort_order,
+            sort_field: None,
```
fmassot, line 283: I suggest to remove them from the indexing settings

quickwit-indexing/src/actors/indexer.rs
```diff
@@ -404,7 +404,7 @@ mod tests {
     use std::time::Duration;
 
     use quickwit_actors::{create_test_mailbox, Universe};
```
fmassot, line 406: in the indexer actor, we can keep the sort attribute in the indexer state but always set it to None.

fmassot (changes requested):
See my 2 comments, let's clean a little bit more the sorting stuff :)

quickwit-config/Cargo.toml
```diff
@@ -16,6 +16,8 @@ json_comments = "0.2"
 quickwit-common = { version = "0.1.0", path = "../quickwit-common" }
 quickwit-index-config = { version = "0.1.0", path = "../quickwit-index-config" }
 quickwit-storage = { version = "0.1.0", path = "../quickwit-storage" }
+quickwit-proto = { version = "0.1.0", path = "../quickwit-proto" }
```
fulmicoton, line 19: I'm surprised: how do we end up adding dependencies when removing a feature?

PR #936, opened 2021-12-20
Implement lightweight indexing server

guilload:
### Description
First stab at an indexing server that spawns and observes multiple indexing pipelines.

### How was this PR tested?
In progress

### Todo
- [x] Implement healthcheck
- [x] Implement supervisor
- [x] Implement `IndexingServerClient::join_indexing_pipeline`
- [x] Populate observable state with statistics
- [x] Add tests
- [x] Misc. fixes

### DIFF
--- a/quickwit-cli/src/index.rs
+++ b/quickwit-cli/src/index.rs
@@ -28,23 +28,22 @@ use chrono::Utc;
 use clap::ArgMatches;
 use colored::Colorize;
 use itertools::Itertools;
-use quickwit_actors::{ActorExitStatus, ActorHandle, ObservationType, Universe};
+use quickwit_actors::{ActorHandle, ObservationType};
 use quickwit_common::uri::Uri;
 use quickwit_common::GREEN_COLOR;
 use quickwit_config::{IndexConfig, IndexerConfig, SourceConfig};
 use quickwit_core::{create_index, delete_index, garbage_collect_index, reset_index};
 use quickwit_index_config::tag_pruning::match_tag_field_name;
-use quickwit_indexing::actors::{IndexingPipeline, IndexingPipelineParams};
+use quickwit_indexing::actors::{IndexingPipeline, IndexingServer};
 use quickwit_indexing::models::IndexingStatistics;
 use quickwit_indexing::source::FileSourceParams;
-use quickwit_indexing::{index_data, STD_IN_SOURCE_ID};
+use quickwit_indexing::STDIN_SOURCE_ID;
 use quickwit_metastore::checkpoint::Checkpoint;
 use quickwit_metastore::{quickwit_metastore_uri_resolver, IndexMetadata, Split, SplitState};
 use quickwit_proto::{SearchRequest, SearchResponse};
 use quickwit_search::{single_node_search, SearchResponseRest};
 use quickwit_storage::{load_file, quickwit_storage_uri_resolver};
 use quickwit_telemetry::payload::TelemetryEvent;
-use serde_json::json;
 use tracing::{debug, info, Level};
 
 use crate::stats::{mean, percentile, std_deviation};
@@ -364,7 +363,6 @@ pub async fn describe_index_cli(args: DescribeIndexArgs) -> anyhow::Result<()> {
     debug!(args = ?args, "describe");
     let metastore_uri_resolver = quickwit_metastore_uri_resolver();
     let quickwit_config = load_quickwit_config(args.config_uri, args.data_dir).await?;
-    info!(quickwit_config = ?quickwit_config);
     let metastore = metastore_uri_resolver
         .resolve(&quickwit_config.metastore_uri)
         .await?;
@@ -587,7 +585,6 @@ pub async fn create_index_cli(args: CreateIndexArgs) -> anyhow::Result<()> {
     quickwit_telemetry::send_telemetry_event(TelemetryEvent::Create).await;
 
     let quickwit_config = load_quickwit_config(args.config_uri, args.data_dir).await?;
-    info!(quickwit_config = ?quickwit_config);
     let index_uri = if let Some(index_uri) = args.index_uri {
         index_uri.as_ref().to_string()
     } else {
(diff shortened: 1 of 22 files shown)


quickwit-cli/src/index.rs
```diff
@@ -637,11 +636,10 @@ pub async fn ingest_docs_cli(args: IngestDocsArgs) -> anyhow::Result<()> {
...
         reset_index(&index_metadata, metastore.clone(), storage.clone()).await?;
     }
-    let pipeline_params =
-        IndexingPipelineParams::try_new(index_metadata, indexer_config, metastore, storage).await?;
-    let pipeline = IndexingPipeline::new(pipeline_params);
-    let universe = Universe::new();
-    let (_pipeline_mailbox, pipeline_handle) = universe.spawn_actor(pipeline).spawn_async();
+    let client = IndexingServer::spawn(indexer_config, metastore);
+    client
+        .spawn_pipeline(args.index_id.clone(), Some(source_config))
```
guilload, line 641: In this PR, I'm handling both `quickwit index ingest` and `quickwit service indexer` commands with the same `IndexingServer`. 

Consequently, most indexing server APIs accept an optional source parameter. `None` means we launch an indexing pipeline for each source defined for the target index.

quickwit-cli/src/index.rs
```diff
@@ -703,22 +706,19 @@ pub async fn merge_or_demux_cli(
...
-    let storage = storage_uri_resolver.resolve(&index_metadata.index_uri)?;
-    index_metadata.sources = vec![source_config];
     let indexer_config = IndexerConfig {
         data_dir_path: args.data_dir_path,
         ..Default::default()
     };
-    index_metadata.indexing_settings.demux_enabled = demux_enabled;
-    index_metadata.indexing_settings.merge_enabled = merge_enabled;
-    index_data(index_metadata, indexer_config, metastore, storage).await?;
+    // FIXME
```
guilload, line 718: Supporting this is actually going to suck. I'm going to have to propagate those settings all the way down.

quickwit-cli/src/index.rs
```diff
@@ -799,7 +801,10 @@ pub async fn start_statistics_reporting_loop(
         report_interval.tick().await;
         // Try to receive with a timeout of 1 second.
         // 1 second is also the frequency at which we update statistic in the console
-        let observation = pipeline_handle.observe().await;
+        let observations = client
+            .observe_pipeline(index_id.to_string(), Some(source_id.to_string()))
```
guilload, line 805: Since there's one indexing pipeline per source, the `observe_pipeline` API becomes quite quirky. If no source ID is specified, it returns a map from (`index_id`, `source_id)` -> `Observation<IndexingStatistics>` for each source defined for the index. Otherwise, it returns a map with one value if the source exists.

quickwit-cli/src/index.rs
```diff
@@ -799,7 +801,10 @@ pub async fn start_statistics_reporting_loop(
...
+            .await?;
+        let observation = observations.get(source_id).expect("");
```
guilload, line 807: Meh.

quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -550,23 +546,28 @@ pub struct IndexingPipelineParams {
...
+        indexing_settings: IndexingSettings,
+        data_dir_path: PathBuf,
+        split_store_max_num_bytes: usize,
+        split_store_max_num_splits: usize,
         metastore: Arc<dyn Metastore>,
         storage: Arc<dyn Storage>,
     ) -> anyhow::Result<Self> {
         let doc_mapper = index_metadata.build_doc_mapper()?;
-        let indexing_directory_path = indexer_config.data_dir_path.join(&index_metadata.index_id);
+        let indexing_directory_path = data_dir_path
```
guilload, line 558: The path for an indexing directory is now `{data_source_path}/{index_id}/{source_id}`.

quickwit-indexing/src/actors/indexing_server.rs
```diff
@@ -0,0 +1,250 @@
...
+    handle: ActorHandle<IndexingServer>,
+}
+
+impl IndexingServerClient {
+    pub async fn spawn_pipeline(
+        &self,
+        index_id: String,
+        source_opt: Option<SourceConfig>,
+    ) -> anyhow::Result<()> {
+        let (sender, receiver) = oneshot::channel();
```
guilload, line 48: The request/response pattern is not supported by the actor framework so I'm relying on one-shot channels.

quickwit-indexing/src/actors/indexing_server.rs
```diff
@@ -0,0 +1,250 @@
...
+        self.universe.send_message(&self.mailbox, message).await?;
+        receiver.await?
+    }
+
+    pub async fn join_pipeline(
+        &self,
+        index_id: String,
+        source_id: String,
+    ) -> anyhow::Result<(ActorExitStatus, IndexingStatistics)> {
+        unimplemented!()
```
guilload, line 78: I'm hoping that an `ActorHandle` is `Send + Sync` so I can detach it from the indexing server and perform the join here.

quickwit-indexing/src/actors/indexing_server.rs
```diff
@@ -0,0 +1,250 @@
...
+            handle,
+        }
+    }
+
+    async fn observe_indexing_pipeline(
+        &mut self,
+        _ctx: &ActorContext<IndexingServerMessage>,
+        index_id: String,
+        source_id_opt: Option<String>,
+    ) -> anyhow::Result<HashMap<String, Observation<IndexingStatistics>>> {
```
guilload, line 129: The not-so-great signature I was mentionning.

quickwit-indexing/src/actors/indexing_server.rs
```diff
@@ -0,0 +1,250 @@
...
+        Ok(observations)
+    }
+
+    async fn spawn_indexing_pipeline(
+        &mut self,
+        ctx: &ActorContext<IndexingServerMessage>,
+        index_id: String,
+        source_opt: Option<SourceConfig>,
+    ) -> anyhow::Result<()> {
+        let index_metadata = self.metastore.index_metadata(&index_id).await?;
```
guilload, line 163: If the metastore takes too long to respond, the actor will timeout. Shall add a record progress right before?

quickwit-indexing/src/actors/indexing_server.rs
```diff
@@ -0,0 +1,250 @@
...
+            let indexing_pipeline_handle = self
+                .indexing_pipeline_handles
+                .get(&(index_id.clone(), source_id.clone()))
+                .with_context(|| {
+                    format!(
+                        "Indexing pipeline `{}` for source `{}` does not exist.",
+                        index_id, source_id
+                    )
+                })?;
+            let observation = indexing_pipeline_handle.observe().await;
```
guilload, line 142: Would we be happy with the last observed state?

quickwit-cli/src/index.rs
guilload: We can also split the API in two:
```
observe_pipeline(index_id, source_id) -> Option<IndexingStatistics>
observe_pipelines(index_id) -> HashMap<String, IndexingStatistics>
```

Same for the spawn API: `spawn_pipeline`, `spawn_pipeline_with_source`.

quickwit-indexing/src/actors/indexing_server.rs
PSeitz: it is, you can check the AutoTraits section in the docs

quickwit-cli/src/index.rs
fmassot: indeed, I would not use the indexing server in this case, it's annoying but I think it's acceptable for now. I guess we will have some code to launch jobs later.

quickwit-cli/src/index.rs
fulmicoton: Could it be two methods then?

quickwit-indexing/src/actors/indexing_server.rs
```diff
@@ -0,0 +1,291 @@
...
+    }
+}
+
+pub struct IndexingServer {
+    data_dir_path: PathBuf,
+    split_store_max_num_bytes: usize,
+    split_store_max_num_splits: usize,
+    metastore: Arc<dyn Metastore>,
+    storage_uri_resolver: StorageUriResolver,
+    indexing_pipeline_handles: BTreeMap<(String, String), ActorHandle<IndexingPipeline>>,
```
fulmicoton, line 106:
```suggestion
    // The key here is (IndexId, Source) in that order.
    indexing_pipeline_handles: BTreeMap<(String, String), ActorHandle<IndexingPipeline>>,
```

quickwit-indexing/src/actors/indexing_server.rs
```diff
@@ -0,0 +1,291 @@
...
+                        "Indexing pipeline `{}` for source `{}` does not exist.",
+                        index_id, source_id
+                    )
+                })?;
+            let observation = indexing_pipeline_handle.observe().await;
+            observations.insert(source_id, observation);
+        } else {
+            for ((_, source_id), indexing_pipeline_handle) in self
+                .indexing_pipeline_handles
+                .range((index_id.clone(), "".to_string())..)
```
fulmicoton, line 173: That's quite cryptic. 

A `HashMap<String, HashMap<String, _>>` might be less riddly.
Otherwise we can add a method called 

```fn source_ids(&self, index_id: &str) -> impl Iterator<Item=&str>```

to encapsulate that logic.

quickwit-indexing/src/actors/indexing_server.rs
fulmicoton: record_progress never hurt, yes.

We might even consider a protect_guard here.

fulmicoton:
I would remove the stats CLI output when we are running the indexer service mode.

quickwit-cli/src/index.rs
guilload: As discussed offline, I will implement a dedicated command for this use case.

guilload:
I still have to write some tests but this is ready for review.

fmassot (approved):
Great PR, nothing to say on my side.

Issue #932, opened 2021-12-20
Fail to ingest data on a new index if the index directory is not created by the user itself


fmassot:
## Steps to reproduce

```bash
cargo r --release index create --index-config-uri examples/index_configs/wikipedia_index_config.yaml --overwrite
Index `wikipedia` successfully created.
cat wiki-articles-10000.json | cargo r --release index ingest --index-id wikipedia --data-dir-path ./data
---------------------------------------------------
 Connectivity checklist
 ✔ metastore
 ✖ storage
 ✔ stdin-source
---------------------------------------------------
 Error Details

 ✖ storage
Missing path `/Users/fmassot/Documents/quickwit/indexes/wikipedia`
```

fmassot:
Instead of checking if a directory exists or not in the `check` function, we can just create the directories and see if we have the right permissions.

Issue #915, opened 2021-12-17
Setup nice default config values for easing the use of the CLI


fmassot:
Following discussions on #852, here are the modifications I would like to make on the CLI:
Assuming the CLI is run in `$QUICKWIT_HOME` with binary in  `$QUICKWIT_HOME/bin`:
- by default, the CLI fetches a config file in `$QUICKWIT_HOME/config/quickwit.yaml`. If there is no config file there, the user must pass the argument `config-uri`. We could also use an environment variable for that.
- the default `metastore_uri` is `file:///$QUICKWIT_HOME/metastore`
- the default data-dir is `$QUICKWIT_HOME/data`
- the default index uri root dir is `file:///$QUICKWIT_HOME/data/indexes`

With these defaults, a user can untar quickwit package and start like this:

## User story 1: test quickwit locally
- `cd quickwit-{VERSION}`
- `./bin/quickwit index create --index-id gh-archive`
- `./bin/quickwit index ingest --index-id gh-archive --input-path ./tutorials/gh-archive/gh-archive.json`
- `./bin/quickwit service run searcher`

## User story 2: test quickwit with index data on s3
- `./bin/quickwit index create --index-id gh-archive --index-uri s3://....`

## User story 3: test quickwit with PostgreSQL metastore
The user edit `./config/quickwit.yaml` with the right `metastore_uri`
- `./bin/quickwit index create --index-id gh-archive --index-uri s3://....`

Or the user has another config file with the right `metastore_uri` and do
- `./bin/quickwit index create --index-id gh-archive --config-uri ./config/postgresql-quickwit.yaml`

fulmicoton:
`${QUICKWIT_HOME}` is defined as $BINARY_PATH/../ 
Confs act as template, and we interpolate them as we read them so that 
the conf can refer to the ${QUICKWIT_HOME} path.

Issue #896, opened 2021-12-14
Pass index-uri in CLI argument when creating an index (and index-id?)


fmassot:
I'm writing the GitHub archive tutorial and one thing that is annoying is the fact that you need to put the `index-uri` in the index config. The user can't download the index config file and just create its index. He will have to customize the file to make it work locally. 

1) First solution: a practical way to avoid that is to pass the `index-uri` in argument. I was thinking of adding also the `index-id`.
2) Second solution: we can use a bash command to quickwit replace the `index-uri` value with an env variable.

I like the 1.

evanxg852000:
- Another option I used was to have the user creates the config file with commands like below assuming he is running through the tutorial from the same directory.
```bash
echo "version: 0
metastore_uri: file://$(pwd)/indexes
searcher:
  data_dir_path: file://$(pwd)/datadir
  host_key_path: file://$(pwd)/datadir" > server_config.yaml
```

- Also if this is only for tutorial files we are bundling with the installation, we could have a script in QUICKWIT_HOME/init_tutorial.sh that will ideally create all the files needed for the tutorial in an argument passed folder. Then we can instruct the user to walk through the tutorial from that folder.

guilload:
I've also been struggling with this. This makes reusing index configurations more difficult. So I vote to get rid of `index_id` and `index_uri` in `IndexConfig` and pass them explicitly in the `quickwit index create` command.

PR #938, opened 2021-12-20
Ease quickwit configuration

fmassot:
## Issues closed by this PR

- #915
- #896
- #932

## Description
The goal of this PR is to make quickwit configuration easy for a new user, it's very user-friendly oriented.

### 1. A quickwit config file for all commands

This PR introduces a quickwit config file that corresponds to the previous `server_config.yaml`. Its primary goal is to define the main variables used by all or almost commands:
- the `metastore_uri``
- the `data_dir_path`

I moved the `data_dir_path` as a global variable as I considered it useful for many commands.

The config file contains also the indexer/searcher config parameters if the user wants to start these services.

Moreover, to make it easy to start with quickwit (and similar to other big project config setup), the user can run the following commands:
```
quickwit service run searcher # <-- by default it will load the config file in ./config/quickwit.yaml
quickwit service run searcher --config-uri ./config/quickwit-node-1.yaml
QUICKWIT_CONFIG_URI=./config/quickwit-node-1.yaml quickwit service run searcher
``` 

### 2. A default `metastore_uri` and `index_root_uri` useful at index creation

When the user wants to create an index, it's good to have nice defaults AND to have a config independent from local environment.

The defaults are:
- `metastore_uri: CWD/metastore` 
- `default_index_root_uri: CWD/data/indexes`: I'm hesitating with this default, putting indexes dir at root level is also ok. I like having one data directory. I don't think this is a big issue as tutorial will end up putting indexes in dedicated directories on s3.

The create command in its simplest form is:
```
quickwit index create --index-id wikipedia
```

You can also specify the `index-uri` if needed.

### 3. Node id

As discussed here #916, I changed `host_key` into `node_id` specified in the config file. So don't need to save it in the data dir.

### 4. A conf file to start with.

The goal of this config file is to shipped in the `tar.gz`: https://github.com/quickwit-inc/quickwit/blob/7a9ab3a91ad3497b2a7cbecf6c650436dcf3662a/config/quickwit.yaml

### DIFF
--- a/config/quickwit.yaml
+++ b/config/quickwit.yaml
@@ -1,0 +1,58 @@
+# ============================ Quickwit Configuration ==============================
+#
+# Website: https://quickwit.io
+# Docs: https://quickwit.io/docs/
+#
+# -------------------------------- General settings --------------------------------
+#
+# Config file version.
+#
+version: 0
+#
+# Node ID must be unique in your cluster. If not set, a random ID is given at each boot.
+#
+#node_id: node-1
+#
+# Path to directory where data is persisted. Default to `./qwdata`.
+#
+#data_dir: /path/to/data
+#
+# Metastore URI. Default to `data_dir/indexes` which is a file backed metastore
+# and mostly convenient for testing. A cluster would require a metastore backed by S3 or
+# postgreqsl.
+#
+#metastore_uri: s3://your-bucket/indexes
+#metastore_uri: postgres://username:password@host:port/db
+#
+# Default index root URI which define the location where index data (splits) is stored
+# following the scheme `{default_index_root_uri}/{index-id}`. Default to `data_dir/indexes`.
+#
+#default_index_root_uri: s3://your-bucket/indexes
+#
+#
+
+#
+# -------------------------------- Indexer settings --------------------------------
+#
+# By default an indexer is only accessible on localhost on http port 7280.
+#
+#indexer:
+#  rest_listen_address: 127.0.0.1
+#  rest_listen_port: 7280
+#  split_store_max_num_bytes: 200G
+#  split_store_max_num_splits: 10000
+#
+# -------------------------------- Searcher settings --------------------------------
+#
+# By default a searcher is only accessible on localhost on http port 7280.
+#
+#searcher:
+#  rest_listen_address: 127.0.0.1
+#  rest_listen_port: 7280
+#  peer_seeds:
+#    - quickwit-searcher-0.local
+#    - quickwit-searcher-1.local
+#  fast_field_cache_capacity: 10G
+#  split_footer_cache_capacity: 1G
+#
+
--- a/quickwit-cli/src/cli.yaml
+++ b/quickwit-cli/src/cli.yaml
@@ -12,17 +12,31 @@ subcommands:
                 about: Creates an index.
                 display_order: 1
                 args:
-                    - metastore-uri:
-                        about: Location of the metastore.
-                        long: metastore-uri
-                        value_name: METASTORE URI
+                    - index:
+                        about: ID of the target index.
+                        long: index
+                        value_name: INDEX
                         required: true
-                        env: QUICKWIT_METASTORE_URI
-                    - index-config-uri:
+                    - index-config:
                         about: Location of the index config file.
-                        long: index-config-uri
-                        value_name: INDEX CONFIG URI
+                        long: index-config
+                        value_name: INDEX CONFIG
                         required: true
+                    - index-uri:
+                        about: Index data (or splits) storage URI.
+                        long: index-uri
+                        value_name: INDEX URI
+                    - config:
+                        about: Quickwit config file.
+                        long: config
+                        value_name: CONFIG
+                        env: QW_CONFIG
+                        required: true
+                    - data-dir:
+                        about: Where data is persisted. Override data-dir defined in config file, default is `./qwdata`.
+                        long: data-dir
+                        value_name: DATA DIR
+                        env: QW_DATA_DIR
                     - overwrite:
                         about: Overwrites pre-existing index.
                         long: overwrite
@@ -30,26 +44,26 @@ subcommands:
                 display_order: 2
                 about: Indexes JSON documents read from a file or streamed from stdin.
                 args:
-                    - index-id:
+                    - index:
                         about: ID of the target index.
-                        long: index-id
-                        value_name: INDEX ID
+                        long: index
+                        value_name: INDEX
                         required: true
-                    - metastore-uri:
-                        about: Location of the metastore.
-                        long: metastore-uri
-                        value_name: METASTORE URI
+                    - config:
+                        about: Quickwit config file.
+                        long: config
+                        value_name: CONFIG
+                        env: QW_CONFIG
                         required: true
-                        env: QUICKWIT_METASTORE_URI
+                    - data-dir:
+                        about: Where data is persisted. Override data-dir defined in config file, default is `./qwdata`.
+                        long: data-dir
+                        value_name: DATA DIR
+                        env: QW_DATA_DIR
                     - input-path:
                         about: Location of the input file.
                         long: input-path
                         value_name: INPUT PATH
-                    - data-dir-path:
-                        about: Creates cached and intermediate data structures in this local directory.
-                        long: data-dir-path
-                        value_name: DATA DIR PATH
-                        required: true
                     - overwrite:
                         about: Overwrites pre-existing index.
                         long: overwrite
@@ -57,32 +71,42 @@ subcommands:
                 display_order: 3
                 about: Display descriptive statistics about the index.
                 args:
-                    - index-id:
+                    - index:
                         about: ID of the target index.
-                        long: index-id
-                        value_name: INDEX ID
+                        long: index
+                        value_name: INDEX
                         required: true
-                    - metastore-uri:
-                        about: Location of the metastore.
-                        long: metastore-uri
-                        value_name: METASTORE URI
+                    - config:
+                        about: Quickwit config file.
+                        long: config
+                        value_name: CONFIG
+                        env: QW_CONFIG
                         required: true
-                        env: QUICKWIT_METASTORE_URI
+                    - data-dir:
+                        about: Where data is persisted. Override data-dir defined in config file, default is `./qwdata`.
+                        long: data-dir
+                        value_name: DATA DIR
+                        env: QW_DATA_DIR
             - search:
                 display_order: 4
                 about: Searches an index.
                 args:
-                    - index-id:
+                    - index:
                         about: id of the target index.
-                        long: index-id
-                        value_name: INDEX ID
+                        long: index
+                        value_name: INDEX
                         required: true
-                    - metastore-uri:
-                        about: Location of the metastore.
-                        long: metastore-uri
-                        value_name: METASTORE URI
+                    - config:
+                        about: Quickwit config file.
+                        long: config
+                        value_name: CONFIG
+                        env: QW_CONFIG
                         required: true
-                        env: QUICKWIT_METASTORE_URI
+                    - data-dir:
+                        about: Where data is persisted. Override data-dir defined in config file, default is `./qwdata`.
+                        long: data-dir
+                        value_name: DATA DIR
+                        env: QW_DATA_DIR
                     - query:
                         about: Query expressed in Tantivy syntax.
                         long: query
@@ -115,57 +139,62 @@ subcommands:
                 display_order: 4
                 about: Merges an index.
                 args:
-                    - index-id:
-                        about: Index ID.
-                        long: index-id
-                        value_name: INDEX ID
+                    - index:
+                        about: INDEX.
+                        long: index
+                        value_name: INDEX
                         required: true
-                    - metastore-uri:
-                        about: Location of the metastore.
-                        long: metastore-uri
-                        value_name: METASTORE URI
+                    - config:
+                        about: Quickwit config file.
+                        long: config
+                        value_name: CONFIG
+                        env: QW_CONFIG
                         required: true
-                        env: QUICKWIT_METASTORE_URI
-                    - data-dir-path:
-                        about: Creates indexing directories in this local directory.
-                        long: data-dir-path
-                        value_name: DATA DIR PATH
-                        required: true
+                    - data-dir:
+                        about: Where data is persisted. Override data-dir defined in config file, default is `./qwdata`.
+                        long: data-dir
+                        value_name: DATA DIR
+                        env: QW_DATA_DIR
             - demux:
                 display_order: 5
                 about: Demuxes an index.
                 args:
-                    - index-id:
+                    - index:
                         about: ID of the target index.
-                        long: index-id
-                        value_name: INDEX ID
+                        long: index
+                        value_name: INDEX
                         required: true
-                    - metastore-uri:
-                        about: Location of the metastore.
-                        long: metastore-uri
-                        value_name: METASTORE URI
+                    - config:
+                        about: Quickwit config file.
+                        long: config
+                        value_name: CONFIG
+                        env: QW_CONFIG
                         required: true
-                        env: QUICKWIT_METASTORE_URI
-                    - data-dir-path:
-                        about: Creates indexing directories in this local directory.
-                        long: data-dir-path
-                        value_name: DATA DIR PATH
-                        required: true
+                    - data-dir:
+                        about: Where data is persisted. Override data-dir defined in config file, default is `./qwdata`.
+                        long: data-dir
+                        value_name: DATA DIR
+                        env: QW_DATA_DIR
             - gc:
                 display_order: 6
                 about: Garbage collects stale staged splits and splits marked for deletion.
                 args:
-                    - index-id:
+                    - index:
                         about: ID of the target index.
-                        long: index-id
-                        value_name: INDEX ID
+                        long: index
+                        value_name: INDEX
                         required: true
-                    - metastore-uri:
-                        about: Location of the metastore.
-                        long: metastore-uri
-                        value_name: METASTORE URI
+                    - config:
+                        about: Quickwit config file.
+                        long: config
+                        value_name: CONFIG
+                        env: QW_CONFIG
                         required: true
-                        env: QUICKWIT_METASTORE_URI
+                    - data-dir:
+                        about: Where data is persisted. Override data-dir defined in config file, default is `./qwdata`.
+                        long: data-dir
+                        value_name: DATA DIR
+                        env: QW_DATA_DIR
                     - grace-period:
                         about: Threshold period after which stale staged splits are garbage collected.
                         long: grace-period
@@ -178,17 +207,22 @@ subcommands:
                 display_order: 7
                 about: Delete an index.
                 args:
-                    - index-id:
+                    - index:
                         about: ID of the target index.
-                        long: index-id
-                        value_name: INDEX ID
+                        long: index
+                        value_name: INDEX
                         required: true
-                    - metastore-uri:
-                        about: Location of the metastore.
-                        long: metastore-uri
-                        value_name: METASTORE URI
+                    - config:
+                        about: Quickwit config file.
+                        long: config
+                        value_name: CONFIG
+                        env: QW_CONFIG
                         required: true
-                        env: QUICKWIT_METASTORE_URI
+                    - data-dir:
+                        about: Where data is persisted. Override data-dir defined in config file, default is `./qwdata`.
+                        long: data-dir
+                        value_name: DATA DIR
+                        env: QW_DATA_DIR
                     - dry-run:
                         about: Executes the command in dry run mode and only displays the list of splits candidate for deletion.
                         long: dry-run
@@ -201,16 +235,10 @@ subcommands:
             - list:
                 about: Lists the splits of an index.
                 args:
-                    - metastore-uri:
-                        about: Location of the metastore.
-                        long: metastore-uri
-                        value_name: METASTORE URI
-                        required: true
-                        env: QUICKWIT_METASTORE_URI
-                    - index-id:
+                    - index:
                         about: ID of the target index.
-                        long: index-id
-                        value_name: INDEX ID
+                        long: index
+                        value_name: INDEX
                         required: true
                     - states:
                         about: |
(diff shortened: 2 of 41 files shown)


fmassot:
@fulmicoton @guilload stable version. Have a look.

quickwit-cli/src/cli.yaml
```diff
@@ -12,17 +12,25 @@ subcommands:
...
                     - index-config-uri:
                         about: Location of the index config file.
                         long: index-config-uri
                         value_name: INDEX CONFIG URI
                         required: true
+                    - index-uri:
+                        about: URI of the target index.
+                        long: index-uri
+                        value_name: INDEX URI
+                    - config-uri:
```
evanxg852000, line 29: nitpick: maybe we could refer to it `quickwit-config-uri` since this is what we are calling it in `.rs` files

quickwit-cli/src/index.rs
```diff
@@ -575,15 +585,24 @@ pub async fn create_index_cli(args: CreateIndexArgs) -> anyhow::Result<()> {
         create_timestamp: Utc::now().timestamp(),
         update_timestamp: Utc::now().timestamp(),
     };
-    create_index(&args.metastore_uri, index_metadata).await?;
+    create_index(&quickwit_config.metastore_uri, index_metadata.clone()).await?;
     println!("Index `{}` successfully created.", index_config.index_id);
+
+    // Add a uri is accessible.
```
evanxg852000, line 591: comments?

quickwit-cli/tests/cli.rs
```diff
@@ -730,24 +716,21 @@ async fn test_all_with_s3_localstack_cli() -> Result<()> {
...
+            "index search --index-id {} --query level:info --config-uri {}",
+            test_env.index_id,
+            test_env.resource_files["server_config"].display(),
         )
         .as_str(),
     )
     .assert()
     .success()
     .stdout(predicate::function(|output: &[u8]| {
+        // println!("{:?}", from_utf8(output));
```
evanxg852000, line 733: comment can be removed.

quickwit-config/src/config.rs
```diff
@@ -79,20 +94,13 @@ impl IndexerConfig {
     }
 
     pub fn validate(&self) -> anyhow::Result<()> {
-        if !self.data_dir_path.exists() {
```
evanxg852000, line 82: We could remove this.

quickwit-config/src/config.rs
```diff
@@ -224,19 +216,56 @@ pub struct StorageConfig {
...
+    )
+}
+
+fn default_quickwit_config_path() -> io::Result<String> {
+    env::current_dir().map(|path| path.join("quickwit.yaml").to_string_lossy().to_string())
+}
+
+impl QuickwitConfig {
+    // Load quickwit config from the given path or from the default config file
+    // and validate.
```
evanxg852000, line 254:
```suggestion
    // Loads quickwit config from the given path or from the default config file
    // and validates.
```

quickwit-config/src/index_config.rs
```diff
@@ -203,54 +204,83 @@ pub struct SourceConfig {
...
     pub indexing_settings: IndexingSettings,
+    #[serde(default)]
     pub search_settings: SearchSettings,
     #[serde(default)]
     pub sources: Vec<SourceConfig>,
 }
 
 impl IndexConfig {
-    pub async fn from_file(path: &str) -> anyhow::Result<Self> {
+    // Load config from a given path and validate it.
```
evanxg852000, line 221:
```suggestion
    // Loads config from a given path and validates it.
```

evanxg852000 (commented):
LGTM :+1:  just left a few comments.

evanxg852000 (approved):

quickwit-cluster/src/error.rs
```diff
@@ -41,14 +41,14 @@ pub enum ClusterError {
...
     /// Read host key error.
     #[error("Failed to read host key: `{message}`")]
-    ReadHostKeyError {
+    ReadHostIdError {
         /// Underlying error message.
         message: String,
     },
 
     /// Write host key error.
     #[error("Failed to write host key: `{message}`")]
```
guilload, line 50:
```suggestion
    #[error("Failed to write host ID: `{message}`")]
```

quickwit-cluster/src/error.rs
```diff
@@ -41,14 +41,14 @@ pub enum ClusterError {
 
     /// Read host key error.
     #[error("Failed to read host key: `{message}`")]
```
guilload, line 43:
```suggestion
    #[error("Failed to read host ID: `{message}`")]
```

quickwit-config/resources/tests/config/quickwit.json
```diff
@@ -1,17 +1,16 @@
 # Comments are supported.
 {
     "version": 0,
+    "node_id": "0",
```
guilload, line 4:
```suggestion
    "node_id": "my-unique-node-id",
```

quickwit-config/resources/tests/config/quickwit.yaml
```diff
@@ -1,14 +1,13 @@
 version: 0
+node_id: 0
```
guilload, line 2: Serde will parse that as a string?

quickwit-config/src/config.rs
```diff
@@ -20,17 +20,35 @@
...
+use quickwit_common::new_coolid;
 use quickwit_storage::load_file;
 use serde::{Deserialize, Serialize};
-use tracing::warn;
+use tracing::{info, warn};
 
 fn default_data_dir_path() -> PathBuf {
-    PathBuf::from("/var/lib/quickwit/data")
+    env::current_dir()
+        .expect("Current working directory value is invalid")
```
guilload, line 36:
```suggestion
        .expect("Failed to resolve current working directory: dir does not exist or insufficient permissions.")
```

quickwit-config/src/config.rs
```diff
@@ -224,19 +212,56 @@ pub struct StorageConfig {
...
+
+impl QuickwitConfig {
+    // Loads quickwit config from the given path or from the default config file
+    // and validates.
+    pub async fn load(path: Option<String>) -> anyhow::Result<Self> {
+        let config_path = if let Some(config_path) = path {
+            config_path
+        } else {
+            let config_path = default_quickwit_config_path()?;
+            info!("Use default config file: {}", config_path);
```
guilload, line 256: Do we display a good error message when we fail down the line if that file does not exist? I wonder if we should fail ealier.

quickwit-config/src/config.rs
```diff
@@ -377,84 +395,65 @@ mod tests {
...
-            assert!(indexer_config.validate().is_err());
-        }
-        {
-            let indexer_config = IndexerConfig {
-                data_dir_path: env::current_dir().unwrap(),
-                ..Default::default()
-            };
-            assert!(indexer_config.validate().is_ok());
+            let minimal_config_yaml = "version: 0";
+            let server_config =
```
guilload, line 438: `s/server_config/config`?

quickwit-config/src/index_config.rs
```diff
@@ -188,8 +188,9 @@ impl Default for IndexingSettings {
     }
 }
 
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Default)]
```
guilload, line 191: I think `Default` on `SearchSettings` is dangerous. That means you can index some docs with the minimal config and have all the subsequent searches on that index come back empty because the default search fields are not set.

quickwit-config/src/index_config.rs
```diff
@@ -427,6 +460,27 @@ mod tests {
...
+        let index_config_filepath = get_resource_path("hdfs-logs-create-config.yaml");
+        // The config file does not set index_id and index_uri.
+        assert!(IndexConfig::load(&index_config_filepath).await.is_err());
+        let index_config = IndexConfig::create(
+            "hdfs-logs-test".to_string(),
+            "s3://quickwit-indexes/hdfs-logs-test".to_string(),
+            &index_config_filepath,
+        )
+        .await;
+        println!("{:?}", index_config);
```
guilload, line 474: remove

quickwit-serve/src/lib.rs
```diff
@@ -48,28 +45,6 @@ use crate::grpc_adapter::cluster_adapter::GrpcClusterAdapter;
 use crate::grpc_adapter::search_adapter::GrpcSearchAdapter;
 use crate::rest::start_rest_service;
 
-fn display_help_message(rest_socket_addr: SocketAddr) -> anyhow::Result<()> {
```
guilload, line 51: I think that was actually useful.

quickwit-config/resources/tests/config/quickwit.yaml
guilload: Nvm.

quickwit-config/src/index_config.rs
```diff
@@ -203,54 +204,86 @@ pub struct SourceConfig {
...
+    }
+
+    async fn from_file(path: &str) -> anyhow::Result<Self> {
         let parser_fn = match Path::new(path).extension().and_then(OsStr::to_str) {
             Some("json") => Self::from_json,
             Some("toml") => Self::from_toml,
             Some("yaml") | Some("yml") => Self::from_yaml,
             Some(extension) => bail!(
-                "Failed to read index config file: file extension `.{}` is not supported. \
+                "Failed to read index config file {}: file extension `.{}` is not supported. \
```
guilload, line 237:
```suggestion
                "Failed to read index config file `{}`: file extension `.{}` is not supported. \
```

quickwit-config/src/index_config.rs
```diff
@@ -203,54 +204,86 @@ pub struct SourceConfig {
...
                  Supported file formats and extensions are JSON (.json), TOML (.toml), and YAML \
                  (.yaml or .yml).",
+                path,
                 extension
             ),
             None => bail!(
-                "Failed to read index config file: file extension is missing. Supported file \
-                 formats and extensions are JSON (.json), TOML (.toml), and YAML (.yaml or .yml)."
+                "Failed to read index config file {}: file extension is missing. Supported file \
```
guilload, line 244:
```suggestion
                "Failed to read index config file `{}`: file extension is missing. Supported file \
```

config/quickwit.yaml
```diff
@@ -0,0 +1,53 @@
...
+#searcher:
+#  rest_listen_address: 127.0.0.1
+#  rest_listen_port: 7280
+#  peer_seeds:
+#    - quickwit-searcher-0.local
+#    - quickwit-searcher-1.local
+#  fast_field_cache_capacity: 10G
+#  split_footer_cache_capacity: 1G
+#
+# -------------------------------- Storage settings --------------------------------
```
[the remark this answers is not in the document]
fmassot: Ok, I did not take a look at it, let's remove that.

config/quickwit.yaml
```diff
@@ -0,0 +1,53 @@
...
+#
+#indexer:
+#  rest_listen_address: 127.0.0.1
+#  rest_listen_port: 7280
+#  split_store_max_num_bytes: 200G
+#  split_store_max_num_splits: 10000
+#
+# -------------------------------- Searcher settings --------------------------------
+# 
+# By default a searcher is only accessible on localhost on http port 7280.
```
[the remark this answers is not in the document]
fmassot: I was waiting to merge your PR on indexing server.

quickwit-cli/src/index.rs
```diff
@@ -341,7 +336,11 @@ impl IndexCliCommand {
 pub async fn describe_index_cli(args: DescribeIndexArgs) -> anyhow::Result<()> {
     debug!(args = ?args, "describe");
     let metastore_uri_resolver = MetastoreUriResolver::default();
-    let metastore = metastore_uri_resolver.resolve(&args.metastore_uri).await?;
+    let quickwit_config = QuickwitConfig::load(args.quickwit_config_uri).await?;
+    info!(quickwit_config = ?quickwit_config);
```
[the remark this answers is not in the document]
fmassot: I like having that in info. It shows the default config, it's important as there are some default values and it's super useful to have that in info.

quickwit-cli/src/index.rs
```diff
@@ -341,7 +336,11 @@ impl IndexCliCommand {
 pub async fn describe_index_cli(args: DescribeIndexArgs) -> anyhow::Result<()> {
     debug!(args = ?args, "describe");
     let metastore_uri_resolver = MetastoreUriResolver::default();
-    let metastore = metastore_uri_resolver.resolve(&args.metastore_uri).await?;
+    let quickwit_config = QuickwitConfig::load(args.quickwit_config_uri).await?;
```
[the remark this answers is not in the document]
fmassot: I started with this pattern... but I wanted to print some logs each time we choose the default config file.

quickwit-cli/src/index.rs
```diff
@@ -559,10 +558,21 @@ fn print_descriptive_stats(values: &[usize]) {
...
     quickwit_telemetry::send_telemetry_event(TelemetryEvent::Create).await;
-    let index_config = IndexConfig::from_file(&args.index_config_uri).await?;
 
+    let quickwit_config = QuickwitConfig::load(args.quickwit_config_uri).await?;
+    info!(quickwit_config = ?quickwit_config);
+    let index_uri = args
+        .index_uri
+        .unwrap_or_else(|| default_quickwit_index_uri(&quickwit_config, args.index_id.clone()));
+    let index_config =
+        IndexConfig::create(args.index_id, index_uri, &args.index_config_uri).await?;
```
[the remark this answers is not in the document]
fmassot: Oh yes, I totally missed that, I thought I had to keep them.

quickwit-cli/tests/cli.rs
```diff
@@ -636,33 +624,28 @@ async fn test_all_local_index() -> Result<()> {
...
-        .expect("Failed to capture process stdout.")
-        .take(800)
-        .read_to_string(&mut process_stdout_output)
-        .expect("Failed to read 800 bytes from process stdout.");
-    assert!(process_stdout_output.contains("http://127.0.0.1:"));
+        .expect("Failed to get server process output")
+        .take(700)
+        .read_to_string(&mut process_output_str)
+        .expect("Cannot read output");
+    assert!(process_output_str.contains("rest_addr=127.0.0.1:"));
```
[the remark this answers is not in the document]
fmassot: I was annoyed by the print "curl 'http://127.0.0.1:40742/api/v1/my-index/search?query=my+query'" which is not very useful, it's a wrong index and a wrong query example.
I would be in favor of putting "Server started on http://127.0.0.1:40742/" in info.

quickwit-config/src/config.rs
fmassot: I'm renaming everything, will commit that soon.

guilload:
Putting more thoughts into this, I'd remove the default env var `QUICKWIT_CONFIG_URI` in the CLI and check it as part of the default routine:

```python
fn default_config_uri() -> Uri {
  if en.get('"QUICKWIT_CONFIG_URI"):
    info("Using config file `{}` from env var")
    return ...
  
  for extension in ["json", "yaml", "toml"] {
    if cwd.join(f"quickwit.{extension}").exists():
      info("Using config file `{}`")
      return ...
  }

  return Exception("Unable de locate configuration. Please, consult the documentation for ..." + Link to docs.")
}

File: quickwit-config/src/index_config.rs (hunk 14 above)
Comment (fmassot) [reply quickwit-config/src/index_config.rs:191]: I don't really agree. I would be happy to create an index without this parameter, no default search field is fine, it's not q a required setting.

**Comment by fmassot** [comment]
> Putting more thoughts into this, I'd remove the default env var `QUICKWIT_CONFIG_URI` in the CLI and check it as part of the default routine:

I was wondering how to do it better because I had issues with a `QUICKWIT_CONFIG_URI` set but empty, it's nice to check that in a dedicated routine and log that.

Thanks for the very helpful review.

File: quickwit-config/src/config.rs (hunk 12 above)
Comment (fmassot) [reply quickwit-config/src/config.rs:256]: Good point, we can improve a little, will have a look.

File: quickwit-serve/src/lib.rs (hunk 16 above)
Comment (fmassot) [reply quickwit-serve/src/lib.rs:51]: Sounds overkill and will be flooded by other log messages. I'm ok to show something in info like "Server started on".

File: quickwit-config/src/index_config.rs (hunk 14 above)
Comment (guilload) [reply quickwit-config/src/index_config.rs:191]: I don't understand how our search works then. A `SearchRequest` has an optional `search_fields` field; what happens if this field is not specified by the user?

File: quickwit-config/src/config.rs (hunk 12 above)
Comment (fmassot) [reply quickwit-config/src/config.rs:256]: This issue should be treated separately. The file metastore will create the necessary directories if it has the permissions to do it. To avoid getting an error super late, I modified the "check" function in the file backed metastore, it now creates the directories to check permissions. It's "borderline" but I found it acceptable.

File: quickwit-config/src/index_config.rs (hunk 14 above)
Comment (fmassot) [reply quickwit-config/src/index_config.rs:191]: the typical search query is like this `my_field:value`

File: quickwit-config/src/config.rs (hunk 12 above)
Comment (fmassot) [reply quickwit-config/src/config.rs]: Done.

File: quickwit-cli/src/index.rs (hunk 23 above)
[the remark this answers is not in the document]
Comment (fmassot) [reply quickwit-cli/src/index.rs]: Done.

**Comment by fmassot** [comment]
I think we are not far merging... 

I will set the default metastore uri in `data/metastore`. So now to quickly start, we only need 2 paths:
- the config file path
- the data dir path

With that, we can deduce easily everything else. Basically I'm already happy with that.

But In this PR, I also use the CWD path to guess the config file and the data dir if they are not given. @guilload is not very fond of it so we can first explicitly define the config file path (`--config-uri` or `QUICKWIT_CONFI_URI`).
For the data path, we can also be explicit and possibly override the config file by setting it with `--data-dir` or `QUICKWIT_DATA_DIR` and we are good.

**Comment by fmassot** [comment]
Looking at tikv or meilisearch, they are taking the CWD, still hesitating....

File: quickwit-cli/src/cli.yaml [hunk 25]
```diff
@@ -12,17 +12,24 @@ subcommands:
...
+                        long: index-id
+                        value_name: INDEX ID
                         required: true
-                        env: QUICKWIT_METASTORE_URI
                     - index-config-uri:
                         about: Location of the index config file.
                         long: index-config-uri
                         value_name: INDEX CONFIG URI
                         required: true
+                    - index-uri:
```
fulmicoton, line 25: do we have a default for the index-uri?

quickwit-cli/src/cli.yaml
```diff
@@ -12,17 +12,24 @@ subcommands:
...
+                        about: URI of the target index.
```
fulmicoton, line 26: I think we need to explain this is the `storage of the index's splits`.
Index-uri might sound like a URI to access (read & search) the index.

quickwit-cli/src/cli.yaml
fulmicoton: Also could we put that in the index config file?

quickwit-cli/src/cli.yaml
```diff
@@ -12,17 +12,24 @@ subcommands:
...
-                        about: Location of the metastore.
-                        long: metastore-uri
-                        value_name: METASTORE URI
+                    - index-id:
+                        about: ID of the target index.
+                        long: index-id
+                        value_name: INDEX ID
                         required: true
-                        env: QUICKWIT_METASTORE_URI
                     - index-config-uri:
```
fulmicoton, line 20: It is true that we accept an uri here. But considering the main usage is just to pass a file from the local file system, 
should we call the CLI option simply `index-config` ?

quickwit-cli/src/cli.yaml
fmassot: yes there is but not very explicit though. Adding a default_index_root_uri in the config file sound not bad, it has the merit of saying "ok if you don't set the URI of your index, it will be at `default_index_root_uri/{index-id}`.

quickwit-cli/src/cli.yaml
```diff
@@ -12,17 +12,24 @@ subcommands:
...
                         long: index-config-uri
                         value_name: INDEX CONFIG URI
                         required: true
+                    - index-uri:
+                        about: URI of the target index.
+                        long: index-uri
+                        value_name: INDEX URI
+                    - config-uri:
+                        about: Quickwit config file. Default to ${CURRENT_WORKING_DIR}/quickwit.yaml.
+                        long: config-uri
```
fulmicoton, line 31: Same as above. I suggest calling it config

quickwit-cli/src/cli.yaml
fulmicoton: (We could even make that a positional argument.)

fulmicoton (approved):
Approved. That's really neat to see the  default conf file.

I think we should remvoe the `-uri` from the cli arguments that are mostly local files to avoid surprising users.

Generally speaking I think we are needless verbose in many arguments.
`index-id` -> `index`.

fmassot:
Last stab made after some interesting thoughts from @guilload 

- `--config` is now mandatory. That is the unique required CLI argument. We can provide a `quickwit.sh` that will set the env variable if needed.
- `data` directory was renamed into `qwdata` following what Materialize is doing. I like this choice as it make the name pretty uniq
- env variables are renamed to to follow Materialize pattern, there are two variables: `QW_CONFIG`and `QW_DATA_DIR`
- the data dir is by default `./qwdata`. I would be very happy to keep that for 2 reasons: first no need to set it, and secondly it is very nice in the code to because I deduce several settings from it like the metatstore. Having it as `./qwdata` is really nice because it cannot fail.
- a `default_index_root_uri` setting has been added to the config file
- `index-config-uri` -> `index-config` and `config-uri` -> `config` which is very nice too
- For the moment I kept the toml/yaml/json.

quickwit-cli/src/cli.yaml
fmassot: Done.

quickwit-cli/src/cli.yaml
fmassot: `index-config` is good.

quickwit-cli/src/cli.yaml
fmassot: done.

quickwit-cli/src/cli.yaml
fmassot: I add `default_index_root_uri `.

fmassot:
> Generally speaking I think we are needless verbose in many arguments. `index-id` -> `index`.

Oh yes, I like this one. @guilload ok for you?

PR #952, opened 2021-12-22
Reenable backpressure in the indexing pipeline

### LINKED ISSUE #949: Reenable some back pressure in the indexer
Right now there is no back pressure... That was a way for us to test if our work was well-balanced.
In reality we want back pressure.

fulmicoton:
Closes #949

### DIFF
--- a/quickwit-indexing/src/actors/merge_planner.rs
+++ b/quickwit-indexing/src/actors/merge_planner.rs
@@ -19,7 +19,7 @@
 
 use std::sync::Arc;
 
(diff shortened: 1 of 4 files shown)


quickwit-indexing/src/actors/merge_planner.rs
```diff
@@ -46,6 +46,10 @@ impl Actor for MergePlanner {
     fn name(&self) -> String {
         "MergePlanner".to_string()
     }
+
+    fn queue_capacity(&self) -> QueueCapacity {
+        QueueCapacity::Bounded(3)
```
fulmicoton, line 51:
```suggestion
        QueueCapacity::Bounded(0)
```

PR #955, opened 2021-12-22
Singletonning the file backed metastore

### LINKED ISSUE #926: Only one SingleFileMetastore instance per process / directory

fulmicoton:
It is necessary to be able to run several
sources targetting the same index for instance.

The metastore uri resolver is a singleton now too.

Closes #926


guilload (approved):
LGTM

PR #961, opened 2021-12-23
use config instead of env

### LINKED ISSUE #905: Cache size shoudl be conf and not ENV variables.
Make the various caches size configurable via the config (currently configured via env variable).

PSeitz:
### Description
Use global config with once cell. When accessing there is a fallback to the default config. So the global config SEARCHER_CONFIG_INSTANCE needs to be set before that.
Break up circular dependency between config and storage crate.

Fixes #905

### How was this PR tested?


quickwit-config/src/config.rs
```diff
@@ -315,7 +323,10 @@ mod tests {
                     "quickwit.{}",
                     stringify!($file_extension)
                 )))?;
-                let config = QuickwitConfig::from_uri(&config_uri).await?;
+                let server_config_filepath =
```
fmassot, line 326: `server_config_filepath` can be renamed in `config_filepath`

fmassot:
The circular dependency is very annoying :/
This means that `quickwit-config` cannot use the storage, it seems that something is wrong here.

Basically we need to give the search config to `open_index(storage, &split)` which requires some caching.
This means that we need to add an extra argument to it and modify the signature of a bunch of functions: `fetch_docs`, `fetch_docs_in_split`, `get_searcher_for_split`, `leaf_search_single_split`, `leaf_search`.

When looking at the code, it does not seem unreasonable to do that now even though we want to release it in a few days. What do you think @guilload, @PSeitz ?

PSeitz:
> The circular dependency is very annoying :/ This means that `quickwit-config` cannot use the storage, it seems that something is wrong here.

I think the config should be a leaf node without storage dependency

> 
> Basically we need to give the search config to `open_index(storage, &split)` which requires some caching. This means that we need to add an extra argument to it and modify the signature of a bunch of functions: `fetch_docs`, `fetch_docs_in_split`, `get_searcher_for_split`, `leaf_search_single_split`, `leaf_search`.
> 

Why not also use the global config file?

fmassot:
> I think the config should be a leaf node without storage dependency
I tend to agree with that.
But now the `QuickwitConfig` API is not in line with that. For example, `from_uri` sounds strange now.

Anyway, let's keep things simple for the release.

fmassot (approved):

quickwit-config/src/index_config.rs
```diff
@@ -216,13 +215,13 @@ pub struct IndexConfig {
 
 impl IndexConfig {
     // Loads config from a given path and validates it.
```
fmassot, line 217: Update the comment.

PSeitz:
> > I think the config should be a leaf node without storage dependency
> 
> I tend to agree with that. But now the `QuickwitConfig` API is not in line with that. For example, `from_uri` sounds strange now.
> 
> Anyway, let's keep things simple for the release.

When we only support one input format, we can just pass the config content and change the function name/signature

PR #963, opened 2021-12-23
Multisource Checkpoints

fulmicoton:
Still missing
- [x] unit tests
- [x] when ingesting a file, we should publish a checkpoint with a source that makes sense.
The easiest way today is to go source_id = f(canonical_path), and keep the checkpoint as is.
A better way would be to add an "ingest_file" source and put the filename in the partition key.

That would require making a change in the FileSource however.


quickwit-metastore/src/metastore/file_backed_metastore/index.rs
```diff
@@ -151,19 +158,13 @@ impl Index {
...
 
-        if is_modified {
-            self.index.update_timestamp = now_timestamp;
-        }
         Ok(is_modified)
     }
 
-    /// Helper to publish a list of splits.
-    pub(crate) fn publish_splits<'a>(
+    pub(crate) fn mark_splits_as_published<'a>(
```
guilload, line 164: I think this one can be private.

quickwit-metastore/src/metastore/file_backed_metastore/index.rs
```diff
@@ -205,10 +206,21 @@ impl Index {
             });
         }
 
-        self.index.update_timestamp = now_timestamp;
         Ok(())
     }
 
+    /// Helper to publish a list of splits.
```
guilload, line 212: I believe the helper function is actually the one above.

quickwit-metastore/src/metastore/file_backed_metastore/index.rs
```diff
@@ -205,10 +206,21 @@ impl Index {
...
+    pub(crate) fn publish_splits<'a>(
```
guilload, line 213: Can we call the helper function `publish_splits_helper` or `publish_splits_inner`? When I quickly glanced over the file initially, I couldn't tell which function was calling the other. That might be because I'm reviewing a diff and not the full file.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -200,7 +200,7 @@ impl PostgresqlMetastore {
 
     /// Publish splits.
     /// Returns the successful split IDs.
-    fn publish_splits(
+    fn mark_splits_as_published(
```
guilload, line 203: I see, you consistently renamed the private functions `mark_splits_as_published`.

quickwit-metastore/src/metastore/index_metadata.rs
```diff
@@ -0,0 +1,414 @@
...
+#[derive(Clone, Debug, Serialize)]
+#[serde(into = "VersionedIndexMetadata")]
+pub struct IndexMetadata {
+    /// Index ID, uniquely identifies an index when querying the metastore.
+    pub index_id: String,
+    /// Index URI, defines the location of the storage that holds the split files.
+    pub index_uri: String,
+    /// Checkpoint relative to a source or a set of sources. It expresses up to which point
+    /// documents have been indexed.
+    pub source_checkpoints: HashMap<String, Checkpoint>, //< The key here is a source_id
```
guilload, line 47: `Checkpoint` -> `SourceCheckpoint`
`HashMap<String, Checkpoint>` -> `struct IndexCheckpoint(HashMap<String, SourceCheckpoint>)`
?

quickwit-metastore/src/metastore/index_metadata.rs
```diff
@@ -0,0 +1,414 @@
...
+            indexing_settings,
+            search_settings,
+            sources: Vec::new(),
+            create_timestamp: now_timestamp,
+            update_timestamp: now_timestamp,
+        }
+    }
+
+    /// Returns the checkpoint associated with a source.
+    pub fn get_checkpoint(&self, source_id: &str) -> Checkpoint {
```
guilload, line 180: I think it would be better to insert an empty source checkpoint when creating an index or adding a new source and fail here if the `source_id` does not exist.

quickwit-metastore/src/metastore/index_metadata.rs
```diff
@@ -0,0 +1,414 @@
...
+    fn from(index_metadata: IndexMetadata) -> Self {
+        VersionedIndexMetadata::V1(index_metadata.into())
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub(crate) struct IndexMetadataV0 {
+    pub index_id: String,
+    pub index_uri: String,
+    #[serde(default)]
```
guilload, line 305: We would have to remove those two lines.

quickwit-metastore/src/metastore/index_metadata.rs
guilload: Maybe that doesn't work for ad-hoc sources for the ingest command.
t_t

guilload: Maybe a source config should come with an "init checkpoint". That would also be useful for starting for a particular point for a Kafka source or prefix for an S3 source.

quickwit-metastore/src/metastore/file_backed_metastore/index.rs
fulmicoton: Rename as _helper and added a comment

quickwit-metastore/src/metastore/file_backed_metastore/index.rs
fulmicoton: good point

quickwit-metastore/src/metastore/file_backed_metastore/index.rs
fulmicoton: good point

quickwit-metastore/src/metastore/index_metadata.rs
fulmicoton: discussed offline. we don't want a default.

fulmicoton:
@guilload I tried but failed to get a good missing checkpoint story that worked properly with the ingest CLI.

quickwit-metastore/src/metastore/index_metadata.rs
fulmicoton: Done.

quickwit-metastore/src/metastore/index_metadata.rs
fulmicoton: Opened #991

fulmicoton:
@guilload ready for your review.

A couple of important changes...
The ingest stuff now gets a specific source id.
It is the same source id regardless of whether it is a file or stdin. Stdin is not a problem here since the fileSource the ignore checkpoint and do not emit checkpoints.

The filepath are canonicalized before creating a partition id out of it.

guilload (approved):

quickwit-metastore/src/checkpoint.rs
```diff
@@ -137,16 +137,80 @@ impl<'a> From<&'a str> for Position {
     }
 }
 
-/// A checkpoint is a map of the last processed position for every partition.
+#[derive(Default, Clone, PartialEq, Serialize, Deserialize)]
+pub struct IndexCheckpoint {
```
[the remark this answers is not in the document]
fulmicoton: Well if we have a problem we can still fix that by modifying the stuff that owns IndexCheckpoint.

PR #967, opened 2021-12-23
Improve serialization format of `quickwit.json` file

### LINKED ISSUE #266: Slight improvement of quickwit.json for readability
It would be nice to have a pretty json and avoid repeating the split id (both in key and object value).

guilload:
### Description
Improve serialization format of `quickwit.json` file. Fixes #266.
```json
{
  "index": {},
  "splits": [{}, {}],
  "version": 0
}
```
### How was this PR tested?
- Added backward-compatibility tests
- Ran `make test-all` command

### DIFF
--- a/quickwit-backward-compat/build.rs
+++ b/quickwit-backward-compat/build.rs
@@ -25,7 +25,47 @@ use quickwit_config::{
 };
 use quickwit_index_config::SortOrder;
 use quickwit_metastore::checkpoint::{Checkpoint, CheckpointDelta, PartitionId, Position};
-use quickwit_metastore::{IndexMetadata, SplitMetadata};
+use quickwit_metastore::file_backed_metastore::FileBackedIndex;
+use quickwit_metastore::{IndexMetadata, Split, SplitMetadata, SplitState};
+
+fn sample_file_backed_index_for_regression() -> FileBackedIndex {
+    let index_metadata = sample_index_metadata_for_regression();
+    let split_metadata = sample_split_metadata_for_regression();
+    let split = Split {
+        split_state: SplitState::Published,
+        split_metadata,
+        update_timestamp: 1789,
+    };
+    let splits = vec![split];
+    FileBackedIndex::new(index_metadata, splits)
+}
+
+fn save_file_backed_index_test_files() -> anyhow::Result<()> {
+    let index = sample_file_backed_index_for_regression();
+    let index_value = serde_json::to_value(&index)?;
+    let version: &str = index_value
+        .as_object()
+        .unwrap()
+        .get("version")
+        .expect("Key `version` is missing from index object.")
+        .as_str()
+        .expect("Value for key `version` should be a string.");
+    let mut index_json = serde_json::to_string_pretty(&index_value)?;
+    index_json.push('\n');
+
+    let md5_digest = md5::compute(&index_json);
+
+    for extension in [".json", ".expected.json"] {
+        std::fs::write(
+            format!(
+                "test-data/file-backed-index/v{}-{:x}{}",
+                version, md5_digest, extension
+            ),
+            index_json.as_bytes(),
+        )?;
+    }
+    Ok(())
+}
 
 /// Creates a new [`IndexMetadata`] object against which backward compatibility tests will be run.
 fn sample_index_metadata_for_regression() -> IndexMetadata {
@@ -193,6 +233,7 @@ fn save_split_metadata_test_files() -> anyhow::Result<()> {
 }
 
 fn main() -> anyhow::Result<()> {
+    save_file_backed_index_test_files()?;
     save_index_metadata_test_files()?;
     save_split_metadata_test_files()?;
     Ok(())
--- a/quickwit-backward-compat/tests/test_backward_compatibility.rs
+++ b/quickwit-backward-compat/tests/test_backward_compatibility.rs
@@ -20,7 +20,26 @@
 use std::fs;
 use std::path::Path;
 
-use quickwit_metastore::{IndexMetadata, SplitMetadata};
+use quickwit_metastore::{FileBackedIndex, IndexMetadata, SplitMetadata};
+
+fn read_file_backed_index(path: &Path) -> anyhow::Result<FileBackedIndex> {
+    let index_bytes = std::fs::read(path)?;
+    let index: FileBackedIndex = serde_json::from_slice(&index_bytes)?;
+    Ok(index)
+}
+
+fn test_file_backed_index_deser(path: &Path) -> anyhow::Result<()> {
+    println!(
+        "---\nTest deserialization of file backed index/{}",
+        path.display()
+    );
+    let index = read_file_backed_index(path)?;
+    let expected_path = path.to_string_lossy().replace(".json", ".expected.json");
+    let expected_index = read_file_backed_index(Path::new(&expected_path))?;
+    test_index_metadata_eq(index.metadata(), expected_index.metadata());
+    assert_eq!(index.splits(), expected_index.splits());
+    Ok(())
+}
 
 fn read_index_metadata(path: &Path) -> anyhow::Result<IndexMetadata> {
     let index_metadata_bytes = std::fs::read(path)?;
@@ -36,6 +55,11 @@ fn test_index_metadata_deser(path: &Path) -> anyhow::Result<()> {
     let index_metadata = read_index_metadata(path)?;
     let expected_path = path.to_string_lossy().replace(".json", ".expected.json");
     let expected_index_metadata = read_index_metadata(Path::new(&expected_path))?;
+    test_index_metadata_eq(&index_metadata, &expected_index_metadata);
+    Ok(())
+}
+
+fn test_index_metadata_eq(index_metadata: &IndexMetadata, expected_index_metadata: &IndexMetadata) {
     assert_eq!(index_metadata.index_id, expected_index_metadata.index_id);
     assert_eq!(index_metadata.index_uri, expected_index_metadata.index_uri);
     assert_eq!(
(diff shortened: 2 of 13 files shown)


quickwit-backward-compat/test-data/index/v0-9c83835c4bae2d3e57f712e2ab9e2c30.expected.json
```diff
@@ -0,0 +1,107 @@
+{
+  "metadata": {
```
guilload: I tried flattening this object, but the two `version` properties collided.

quickwit-metastore/src/metastore/file_backed_metastore/index.rs
```diff
@@ -29,28 +30,83 @@ use crate::{IndexMetadata, MetastoreError, MetastoreResult, Split, SplitMetadata
 
 /// A `Index` object carries an index metadata and its split metadata.
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub(crate) struct Index {
+#[serde(into = "VersionedIndex")]
+#[serde(from = "VersionedIndex")]
+pub struct Index {
     /// Metadata specific to the index.
-    index: IndexMetadata,
+    pub metadata: IndexMetadata,
```
guilload, line 37: Was there a particular reason for keeping the fields `index` and `splits` private? I can also leave them private and write a constructor.

quickwit-metastore/src/metastore/file_backed_metastore/index.rs
```diff
@@ -29,28 +30,83 @@ use crate::{IndexMetadata, MetastoreError, MetastoreResult, Split, SplitMetadata
 
 /// A `Index` object carries an index metadata and its split metadata.
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub(crate) struct Index {
```
guilload, line 32: Necessary for writing the backward-compatibility test.

fmassot:
You didn't change `META_FILENAME` into `metastore.json`?

quickwit-metastore/src/metastore/file_backed_metastore/index.rs
guilload: I renamed `index` to `metadata` because `index.index` was awkward. The serialized version is still called `index`.

fmassot:
I don't think we need to support the V0 format (we have only users with the PostgreSQL metastore).

guilload:
V0 is the new format.

fmassot:
> V0 is the new format.

oh, I think I misunderstand something. 

This code 

```
impl From<Index> for IndexV0 {
    fn from(index: Index) -> Self {
        Self {
            index: index.metadata,
            splits: index
                .splits
                .into_values()
                .sorted_by_key(|split| split.update_timestamp)
                .collect(),
        }
    }
}

impl From<IndexV0> for Index {
    fn from(index: IndexV0) -> Self {
        Self {
            metadata: index.index,
            splits: index
                .splits
                .into_iter()
                .map(|split| (split.split_id().to_string(), split))
                .collect(),
            discarded: false,
        }
    }
}
```

made me think you needed to convert from `IndexV0` to `Index`.

I don't really understand the `discarded` either. Well I will ping you tomorrow, I think I need an update on this.

guilload:
`Index` is the in-memory format. `VersionedIndex` is a "serialization/deserialization" proxy. On the serialization path, we always choose the "highest" `VersionedIndex` variant. On the deserialization path, serde deserializes the right variant thanks to the version tag. Then, we tell serde how to perform the conversion from `IndexVx` to `Index`.

fmassot:
> `Index` is the in-memory format. `VersionedIndex` is a "serialization/deserialization" proxy. On the serialization path, we always choose the "highest" `VersionedIndex` variant. On the deserialization path, serde deserializes the right variant thanks to the version tag. Then, we tell serde how to perform the conversion from `IndexVx` to `Index`.

ok, super clear, thanks.

Then this line `index: index.metadata,` is surprising.

quickwit-metastore/src/metastore/file_backed_metastore/index.rs
fulmicoton: > I renamed index to metadata because index.index was awkward. The serialized version is still called index.

It was awkward. But now the discrepancy between serialization and the object is awkward. That's a pick your poison story.

fulmicoton: > Was there a particular reason for keeping the fields index and splits private? I can also leave them private and write a constructor.

I suspect the goal is to ensure people go through the methods to change index and splits?
It enforces a bunch of cool stuff on the state.

quickwit-metastore/src/lib.rs
```diff
@@ -52,7 +52,7 @@ mod metastore_resolver;
 pub mod postgresql;
 
 pub use error::{MetastoreError, MetastoreResolverError, MetastoreResult};
-pub use metastore::file_backed_metastore::FileBackedMetastore;
+pub use metastore::file_backed_metastore::{FileBackedMetastore, Index};
```
fulmicoton, line 55: I don't think we should expose it as quickwit_metastore::Index...
At least under this name. This will make things very confusing.

This Index is an internal detail of the file backed metastore. 
If it wasn't for regression tests, it should only be visible in the file_backed_metastore module.

quickwit-metastore/src/metastore/file_backed_metastore/index.rs
fulmicoton: Yeah that's a bummer but that makes sense.

fulmicoton (changes requested):
Overall LGTM, and this is a very nice improvement.
A non-nitpick i have is that we really need to find a way to NOT expose quickwit_metastore::Index like that.

quickwit-backward-compat/build.rs
```diff
@@ -17,15 +17,58 @@
...
+    splits.insert(split.split_id().to_string(), split);
+
+    Index {
+        metadata: index,
+        splits,
+        discarded: false,
+    }
+}
+
+fn save_index_test_files() -> anyhow::Result<()> {
```
fulmicoton, line 49: I think we need to have `file_backed` or something in that kind of method name. It is very hard for the reader to undertsand the different between this thing and the index_metadat thing

quickwit-metastore/src/lib.rs
guilload: Can I make `file_backed_metastore` public without making `metastore` public, though?

quickwit-backward-compat/tests/test_backward_compatibility.rs
```diff
@@ -36,6 +55,11 @@ fn test_index_metadata_deser(path: &Path) -> anyhow::Result<()> {
     let index_metadata = read_index_metadata(path)?;
     let expected_path = path.to_string_lossy().replace(".json", ".expected.json");
     let expected_index_metadata = read_index_metadata(Path::new(&expected_path))?;
+    test_index_metadata_eq(&index_metadata, &expected_index_metadata);
+    Ok(())
+}
+
+fn test_index_metadata_eq(index_metadata: &IndexMetadata, expected_index_metadata: &IndexMetadata) {
```
guilload, line 62: This is brittle and will break when a field is added to `IndexMetadata`.

guilload: We decided not to derive `PartialEq` on`IndexMetadata` because `SourceConfig` carries a `serde_json::Value`.

fulmicoton: serde_json::Value implements PartialEq and the default map implementation is a BTreeMap.
Was there a problem with using this directly?

quickwit-cli/Cargo.toml
```diff
@@ -46,7 +46,7 @@ tempfile = "3"
 chrono = "0.4"
 humansize = "1.1.1"
 openssl-probe = { version = "0.1.4", optional = true }
-itertools = "0.10.1"
+itertools = "0.10"
```
fulmicoton, line 49: Contrary to what i might have said before.
I recommend leaving it at 0.10.1. 

First
`0.10.1` actually stands for `^0.10.1`

So `0.10.2` is actually matching the requirement.
Second, there might be an actual good reason to put 0.10.1 and not 0.10.0.

For instance, 0.10.1 may have added extra functionality compared to 0.10.
So it is fairly possible that quickwit does not compile with 0.10.0.

A contributor actually spotted a couple of such example and convinced me to change our dependancy version for that reason. (I don't know if itertools was one of them or not).

quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index.rs
```diff
@@ -17,40 +17,94 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+//! [`FileBackedIndex`] module. It is public so that the crate `quickwit-backward-compat` can
+//! import [`FiledBackedIndex`] and run backward-compatibility tests. You should not have to import
```
fulmicoton, line 21: good comment.

fulmicoton (approved):

fulmicoton:
@guilload  approved but had 2 comments!

quickwit-backward-compat/tests/test_backward_compatibility.rs
guilload: Discussed off GitHub: a few months ago, we decided that even if `serde_json::Value` implements `PartialEq`, that was a bit sketched. Today we are changing our minds.

quickwit-cli/Cargo.toml
guilload: Sounds good to me. Will revert.

quickwit-backward-compat/tests/test_backward_compatibility.rs
guilload: No go. `DocMapping` does not implement `PartialEq`.

PR #969, opened 2021-12-24
Fix void source connectivity check and add partial testing (not all s…

fmassot:
…ources are tested) on this.


guilload (approved):
nit: might as well do `vec`.

PR #973, opened 2021-12-24
Rename `quickwit.json` to `metastore.json`

guilload:
### Description
Rename `quickwit.json` to `metastore.json`. Something we discussed with @fmassot. WDYT @fulmicoton?


PSeitz:
I think that's better since quickwit.json doesn't tell anything

fulmicoton:
I prefer quickwit.json but I don't care all that much.

fulmicoton (approved):

guilload:
> I prefer quickwit.json but I don't care all that much.

We also wanted to avoid any confusion with `quickwit.yaml`.

fulmicoton:
Ah yes, Metastore.json is much better

PR #978, opened 2021-12-26
Rename old index config into doc mapper and add index config validation

fmassot:
This PR is lame as it is only doing the minimal stuff for the release: I mostly renamed old index config stuff into doc mapper stuff.
I change also 3 other things:
- define a function `build_doc_mapper` that is building a doc mapper from 3 structs `DocMapping`, `SearchSettings` and `IndexingSettings`. These structs are needed for what the `DocMapper` is responsible for: converting a JSON into a `Document` and building a query from a `SearchRequest`
- set `store_source` at `false` by default as we are already storing field values by default
- complete index config validate and add tests on it.  

I also plan to rename the cargo project `quickwit-index-config` into `quickwit-doc-mapper`, a kind of "déjà vu".

This lame choice was motivated essentially by 2 points:
- the `DocMapper` trait still seems to be relevant with 2 responsibilities: `doc_from_json` and `query`. For that, it needs basically the settings present in `QuickwitConfig` struct (thus the signature of `build_doc_mapper`)
- Doing some heavy refactoring seems premature as we are going to work partly on that in the next release (via schemaless feature)

I'm still hesitating on changing the `DocMapping` struct in this PR. Currently we have:

```rust
pub struct DocMapping {
    pub field_mappings: Vec<FieldMappingEntry>,
    #[serde(default)]
    pub tag_fields: BTreeSet<String>,
    #[serde(default)]
    pub store_source: bool,
}
```

The doc mapper is doing the real validation and is responsible also for generating the right index schema. Here `DocMapping` is not doing that much: it is checking 3 attributes and most of the work is made with the building of `field_mappings`. And as you may have noticed, this struct is quite redundant with the `DefaultDocMapper` which, nonetheless, is doing a little bit more by validating other settings (`demux_field`, `timestamp`, ...).

I will sleep on this and ping you to discuss that.


quickwit-backward-compat/test-data/index-metadata/unversioned.json
```diff
@@ -2,7 +2,7 @@
   "checkpoint": {
     "0000000000": "0000000042"
   },
-  "index_config": {
+  "doc_mapper": {
```
guilload, line 5: I didn't see any `serde(alias = ...)` anywhere, so are we breaking backward compatibility?

fmassot:
@guilload I fixed the backward compatibility and add tests on validation (spot one bug by the way ;)).

quickwit-index-config/src/default_doc_mapper/default_mapper.rs
```diff
@@ -738,7 +734,7 @@ mod tests {
             ]
         }"#;
 
-        let builder = serde_json::from_str::<DefaultIndexConfigBuilder>(index_config)?;
+        let builder = serde_json::from_str::<DefaultDocMapperBuilder>(index_config)?;
```
fulmicoton, line 737: Ultra nitpick.

I tend to prefer the style:
```
let builder: DefaultDocMapperBuilder = ...
```

It looks a little bit cleaner, but also it gives you the type of the variable.

guilload (approved):

quickwit-index-config/src/default_doc_mapper/default_mapper.rs
fmassot: Yes it's better.

PR #981, opened 2021-12-27
update API docs

PSeitz:
### Description

### How was this PR tested?

### DIFF
--- a/docs/reference/search-api.md
+++ b/docs/reference/search-api.md
@@ -53,16 +53,15 @@ Search for documents matching a query in the given index `<index name>`.
 
 #### Get parameters
 
-| Variable | Type | Description | Default value |
-|----------|------|-------------|---------------|
-| **query** | `String` | Query text. See the [query language doc](query-language.md) (mandatory) | |
-| **searchFields** | `[String]` | If set, specify the set of fields the search will be performed on | |
-| **startTimestamp** | `i64` | If set, restrict search to documents with a `timestamp >= start_timestamp` | |
-| **endTimestamp** | `i64` | If set, restrict search to documents with a `timestamp < end_timestamp`` | |
-| **startOffset** | `Integer` | Number of documents to skip | `0` |
-| **maxHits** | `Integer` | Maximum number of hits to return (by default 20) | `20` |
-| **format** | `String` | Response output format. `json` or `pretyjson`  | `pretyjson` |
-| **tags** | `[String]` | If set, the search is restricted to only splits having one of the tags | |
+| Variable                  | Type                 | Description                                                                                       | Default value                                                                                   |
+| ------------------------- | -------------------- | ------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------- |
+| **query**                  | `String`           | Query text. See the [query language doc](query-language.md) (mandatory)                                          |                                                                                                |
+| **startTimestamp**         | `i64`    		 	    | If set, restrict search to documents with a `timestamp >= start_timestamp`                                                            |                                                                                |
+| **endTimestamp**           | `i64`       		    | If set, restrict search to documents with a `timestamp < end_timestamp``                                                            |                                                                                     |
+| **startOffset**            | `Integer`     	    | Number of documents to skip                                                                | `0`                                                                                             |
+| **maxHits**                | `Integer`          | Maximum number of hits to return (by default 20)                                                            | `20`                                                                                            |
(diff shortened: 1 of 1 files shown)


docs/reference/search-api.md
```diff
@@ -53,16 +53,15 @@ Search for documents matching a query in the given index `<index name>`.
...
+| **endTimestamp**           | `i64`       		    | If set, restrict search to documents with a `timestamp < end_timestamp``                                                            |                                                                                     |
+| **startOffset**            | `Integer`     	    | Number of documents to skip                                                                | `0`                                                                                             |
+| **maxHits**                | `Integer`          | Maximum number of hits to return (by default 20)                                                            | `20`                                                                                            |
+| **searchFields**           | `String`      		  | Fields to search on. Comma-separated list, e.g. "field1,field2" |                                                                                             |
```
fmassot, line 63: We can add the fact that if it's empty, the default_search_feilds of the index will be used.

docs/reference/search-api.md
```diff
@@ -53,16 +53,15 @@ Search for documents matching a query in the given index `<index name>`.
...
+| ------------------------- | -------------------- | ------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------- |
+| **query**                  | `String`           | Query text. See the [query language doc](query-language.md) (mandatory)                                          |                                                                                                |
+| **startTimestamp**         | `i64`    		 	    | If set, restrict search to documents with a `timestamp >= start_timestamp`                                                            |                                                                                |
+| **endTimestamp**           | `i64`       		    | If set, restrict search to documents with a `timestamp < end_timestamp``                                                            |                                                                                     |
```
fmassot, line 60: remove extra `

fmassot:
Some wording can be improved like:

```
Streams field values from ALL documents matching a search query in the given index <index name>, in a specified output format among the following:
- CSV
- ClickHouse RowBinary
```

You have updated the the working for `searchFields` only on the search enpoint, you can do the same for the search stream endpoint.


Last but not least, it would be nice to add extra information about the search stream response. This is not a classic response as we are streaming the results. Can you document that? You can have a look at this issue https://github.com/quickwit-inc/quickwit/issues/637 to understand why it's important.

docs/reference/search-api.md
guilload: I don't think the default search fields are plugged in, though.

fmassot: will check that

fmassot: it's correclty plugged :)

fmassot:
This is no more valid with the stream API.
```
The body of responses is always a JSON object, and their content type is always `application/json; charset=UTF-8.`
```
Can you update that?

docs/reference/search-api.md
```diff
@@ -107,13 +102,16 @@ The endpoint will return 10 million values if 10 million documents match the que
...
-| **tags** | `[String]` | If set, the search is restricted to only splits having one of the tags | |
 
 
 ### Response
 
-The response is a list of all the field values from documents matching the query. The field must be marked as "fast" in the index config for this to work. The formatting is based on the specified output format. 
+The response is a list of all the field values from documents matching the query. The field must be marked as "fast" in the index config for this to work. 
+The formatting is based on the specified output format. 
+
+On error a "X-Stream-Error" header will be sent via the the trailers channel with information about the error and the stream will be closed. Depending on the client, the trailer header may not be shown.
```
fmassot, line 116: This is hard to understand. We need to be more accurate and say that we are using a`Streaming HTTP protocol`.

It's important to talk about the HTTP trailer and also about the `send.abort()`.

Moreover I think HTTP trailers are only compatible with HTTP2.

If a client cannot see the trailer, what will happen? You can maybe add a note on the logging error.

You have to assume that a user has little knowledge about this and if we are not accurate, he will think there is a bug and quit.

PSeitz: As I understand, we are using the chunked transfer encoding, which is not supported by HTTP2.
So I don't understand how it can work with the HTTP trailers which are only HTTP2

fmassot: Maybe we are doing something wrong :/
Can you check that ?

PSeitz: Trailers are HTTP1.1 https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer.
Not sure how they behave if the client switches to HTTP2 though

fmassot (approved):
LGTM

PR #982, opened 2021-12-27
Update Docker images

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/972
- update docker image for new installation layout, 
- added/overridden image metadata 

### How was this PR tested?
Used custom build & install script to test.


fmassot:
>  Though we don't build nightly, this image is currently used this way

I don't really get what you mean? The script is called each time there is a new commit on the main branch?

evanxg852000:
I mean: The image is built every time we merge to `main`, this is not exactly nightly (cron scheduled build) but it's ok.

fmassot (approved):
LGTM.

a nit that you can ignore: I would prefer keeping main in the script name but I may not fully understand why you wanted to call it nightly

docker/Dockerfile
```diff
@@ -1,14 +1,22 @@
...
-LABEL maintainer="Quickwit, Inc. <hello@quickwit.io>"
+FROM alpine:3.14 AS builder
 
 # Install tools
 RUN apk add --no-cache curl
-
-# Install Quickwit
 RUN curl -L https://install.quickwit.io | sh
-RUN mv ./quickwit /usr/local/bin/quickwit
+RUN mv  ./quickwit-*/* ./
```
guilload, line 6: Wonder if we should `sed` the config here to set the listen address to `0.0.0.0`.
@fmassot ?

guilload (approved):

docker/Dockerfile
evanxg852000: From the feedback on the other PR, we have a docker flavor of the config. We can copy that file too?

docker/Dockerfile
```diff
@@ -1,14 +1,22 @@
-FROM alpine:3.14
-LABEL maintainer="Quickwit, Inc. <hello@quickwit.io>"
+FROM alpine:3.14 AS builder
```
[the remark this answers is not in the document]
evanxg852000: Well spotted :+1:

docker/Dockerfile
fmassot: @guilload I suggested having a default config file suited for docker. 
So in this case we have the right config file from the start. Sed is fine too, I'm ok with both solutions.

PR #984, opened 2021-12-27
Update/Generate CLI docs

PSeitz:
### Description
The cli docs are generated from the clap data. It's easier to maintain them, but they are less extensive, since the source is the yaml file.


fmassot:
@PSeitz how do you generate it? 

The result is not bad but very rough, we need to customize it. 

Here is another example of nice CLI documentation: https://materialize.com/docs/cli/

fmassot:
My 2 cents on this.

Auto-generating CLI docs is not as easy as it seems. You have of course a strong pros that is to have a consistent CLI and docs. But you have some cons too: Having lots of writings in a toml file is not super readable and harder to customize. We can look at what other projects are doing, for example `ripgrep` is using a template like this: https://github.com/BurntSushi/ripgrep/blob/master/doc/rg.1.txt.tpl

What I would do for the release is to auto-generate a first version like you did and then customize it by editing manually the markdown file and don't code now a way to automate everything as it will take too long and will probably be hard to maintain. Quickwit is just at the beginning and I would take care of auto-generated docs a little later when things are more stable.

docs/reference/cli.md
```diff
@@ -44,314 +47,433 @@ The line `Telemetry enabled` disappears when you disable it.
...
 
 
+### Syntax
 
-### New
+The cli is structured into highlevel commands with subcommands.
+`quickwit [command] [subcommand] [args]`.
 
-*Description*
+* `command`:index, split, service, source.
```
fmassot, line 55:
```suggestion
* `command`:index, split, source, service. 
```

docs/reference/cli.md
```diff
@@ -44,314 +47,433 @@ The line `Telemetry enabled` disappears when you disable it.
...
-*Options*
-
-`--index-uri` (string) Defines the index location.<br />
-`--index-config-path` (string) Defines the index config path.<br />
-`--overwrite` (boolean) Overwrites existing index.
 
-*Examples*
 
-*Creating a new index on local file system*
+# index
```
fmassot, line 62:
```suggestion
## index
```

docs/reference/cli.md
```diff
@@ -44,314 +47,433 @@ The line `Telemetry enabled` disappears when you disable it.
...
+Index operation commands such as `create`, `index`, or `search`...
 
-```bash
-quickwit new --index-uri file:///quickwit-indexes/catalog --index-config-path ~/quickwit-conf/index_config.json
-```
+### index create
 
-Creating a new index on Amazon S3*
+Creates an index at `index-uri` configured by a yaml file located at `index-config-uri`. The command fails if an index already exists at `index-uri` unless `overwrite` is passed.
```
fmassot, line 67: I don't think this sentence is accurate: `Creates an index at index-uri` 

This was the case for 0.1 but now the reference is the index ID. Index URI can be passed as an optional args.

docs/reference/cli.md
```diff
@@ -44,314 +47,433 @@ The line `Telemetry enabled` disappears when you disable it.
 `quickwit --version` displays the version. It is useful for reporting bugs.
 
 
+### Syntax
```
fmassot, line 50: It would be nice to introduce early in the doc the fact that we can use the env variable `QW_CONFIG` to set the config once for all.

docs/reference/cli.md
```diff
@@ -44,314 +47,433 @@ The line `Telemetry enabled` disappears when you disable it.
...
+*Create a new index with metastore in the local `metastore` folder.*
+```bash
+quickwit index create --index-config-uri ./wikipedia_index_config.yaml  --config=quickwit.yaml
+```
 
-Indexes a dataset consisting of newline-delimited JSON objects located at `input-path` or read from *stdin*. The data is appended to the target index specified by `index-uri` unless `overwrite` is passed. `input-path` can be a file or another command output piped into stdin. Currently, only local datasets are supported. By default, tantivy's indexer will work with a heap of 1 GiB of memory, but this can be set with the `heap-size` options. This does not directly reflect the overall memory usage of `quickwit index`, but doubling this value should give a fair approximation.
+### index ingest
 
+Indexes a dataset consisting of newline-delimited JSON objects located at `input-path` or read from *stdin*. 
+The data is appended to the target index specified by `index-uri` unless `overwrite` is passed. `input-path` can be a file or another command output piped into stdin.
```
fmassot, line 102: we now use the index id and not the `index-uri`

docs/reference/cli.md
```diff
@@ -44,314 +47,433 @@ The line `Telemetry enabled` disappears when you disable it.
...
+```bash
+quickwit index create --index-config-uri ./wikipedia_index_config.yaml  --config=quickwit.yaml
+```
 
-Indexes a dataset consisting of newline-delimited JSON objects located at `input-path` or read from *stdin*. The data is appended to the target index specified by `index-uri` unless `overwrite` is passed. `input-path` can be a file or another command output piped into stdin. Currently, only local datasets are supported. By default, tantivy's indexer will work with a heap of 1 GiB of memory, but this can be set with the `heap-size` options. This does not directly reflect the overall memory usage of `quickwit index`, but doubling this value should give a fair approximation.
+### index ingest
 
+Indexes a dataset consisting of newline-delimited JSON objects located at `input-path` or read from *stdin*. 
+The data is appended to the target index specified by `index-uri` unless `overwrite` is passed. `input-path` can be a file or another command output piped into stdin. 
+Currently, only local datasets are supported.
```
fmassot, line 103: `Currently, only local datasets are supported.`. I think this has changed, it can work with a file sitting on S3.

docs/reference/cli.md
```diff
@@ -44,314 +47,433 @@ The line `Telemetry enabled` disappears when you disable it.
...
+```
 
-Indexes a dataset consisting of newline-delimited JSON objects located at `input-path` or read from *stdin*. The data is appended to the target index specified by `index-uri` unless `overwrite` is passed. `input-path` can be a file or another command output piped into stdin. Currently, only local datasets are supported. By default, tantivy's indexer will work with a heap of 1 GiB of memory, but this can be set with the `heap-size` options. This does not directly reflect the overall memory usage of `quickwit index`, but doubling this value should give a fair approximation.
+### index ingest
 
+Indexes a dataset consisting of newline-delimited JSON objects located at `input-path` or read from *stdin*. 
+The data is appended to the target index specified by `index-uri` unless `overwrite` is passed. `input-path` can be a file or another command output piped into stdin. 
+Currently, only local datasets are supported. 
+By default, tantivy's indexer will work with a heap of 2 GiB of memory, but this can be set with the `heap-size` option in the indexing_settings.resources section in the config.
```
fmassot, line 104: Let's name it "quickwit's indexer".

docs/reference/cli.md
```diff
@@ -44,314 +47,433 @@ The line `Telemetry enabled` disappears when you disable it.
...
-### Search
+*Options*
 
-*Description*
+`--index` ID of the target index.    
+`--config` Quickwit config file.    
+### index search
 
-Searches the index stored at `index-uri` and returns the documents matching the query specified with `query`. The offset of the first hit returned and the number of hits returned can be set with the `start-offset` and `max-hits` options. Given the query doesn't explicitly contains fields, it's possible to restrict the search on specified fields using the `search-fields` option. Search can also be limited to a time range using the `start-timestamp` and `end-timestamp` options. These timestamp options can particularly be useful in boosting query performance when using a time series dataset and only need to query a particular window.
+Searches the index stored at `index-uri` and returns the documents matching the query specified with `query`.
```
fmassot, line 159: Better to talk about the index id first

fmassot (changes requested):
I made a first review and added some comments. 

There is no "long_about" for the indexer and I think this is one of the most important commands.

docs/reference/cli.md
PSeitz: internally this is a PathBuf, it also uses `exists()`

```
    pub data_dir_path: PathBuf,
```

fmassot: I was not talking about `data_dir_path` which is different.
Anyway, I just checked that and the answer is `FileSource` only works with a local file.

quickwit-cli/src/main.rs
```diff
@@ -79,6 +79,155 @@ fn setup_logging_and_tracing(level: Level) -> anyhow::Result<()> {
     Ok(())
 }
 
+#[allow(dead_code)]
```
fmassot, line 82: We can't merge that. This code should be probably put somewhere else (in a script?)

docs/reference/cli.md
PSeitz: yeah it's more complicated now, I update the docs

quickwit-cli/src/main.rs
PSeitz: I moved it to its own binary

fmassot (approved):
Go for merge

PR #987, opened 2021-12-27
Implement delete source CLI command

guilload:
### Description
Implement delete source CLI command.

### How was this PR tested?
- added unit test
- ran `make test-all` command


fmassot (approved):
LGTM. We only test the args parsing for several CLI commands, I'm worried about regression on this. We will need to ease CLI tests because this will be an issue for the coming months.

guilload:
Yes, that's right. A good time to do so is when we add the REST API version of those commands. Most of the CLI commands will have to be refactored so we can share their code with the REST version.

PR #996, opened 2021-12-29
Removing the generated protobuf file from git

fulmicoton:


fulmicoton:
That does nto work >.< bummer

fulmicoton:
The technology is just not here yet.

PR #998, opened 2021-12-29
Implement add source CLI command

guilload:
### Description
Implement add source CLI command

### How was this PR tested?
- added unit test
- ran `make test-all` command

### DIFF
--- a/quickwit-cli/src/cli.yaml
+++ b/quickwit-cli/src/cli.yaml
@@ -398,7 +398,7 @@ subcommands:
                         value_name: SOURCE TYPE
                         required: true
                     - params:
-                        about: Parameters for the source formatted as a JSON object. The object can be passed inline or via a file.
+                        about: Parameters for the source formatted as a JSON object passed inline or via a file. Parameters are source-specific. Please, refer to the source's documentation for more details.
                         long: params
                         value_name: PARAMS
                         required: true
--- a/quickwit-cli/src/cli.yaml
+++ b/quickwit-cli/src/cli.yaml
@@ -379,6 +379,35 @@ subcommands:
         settings:
             - ArgRequiredElseHelp
         subcommands:
+            - add:
+                about: Adds a new source.
+                args:
+                    - index:
+                        about: ID of the target index.
+                        long: index
+                        value_name: INDEX
+                        required: true
+                    - id:
+                        about: ID of the source.
+                        long: id
+                        value_name: SOURCE ID
+                        required: true
+                    - type:
+                        about: "Type of the source. Available types are: `file` and `kafka`."
+                        long: type
+                        value_name: SOURCE TYPE
+                        required: true
+                    - params:
+                        about: Parameters for the source formatted as a JSON object. The object can be passed inline or via a file.
+                        long: params
+                        value_name: PARAMS
+                        required: true
+                    - config:
+                        about: Quickwit config file.
+                        long: config
+                        value_name: CONFIG
+                        env: QW_CONFIG
+                        required: true
             - delete:
                 about: Deletes a source.
                 args:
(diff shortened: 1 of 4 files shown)


quickwit-cli/src/cli.yaml
```diff
@@ -379,6 +379,35 @@ subcommands:
...
+                        about: ID of the source.
+                        long: id
+                        value_name: SOURCE ID
+                        required: true
+                    - type:
+                        about: "Type of the source. Available types are: `file` and `kafka`."
+                        long: type
+                        value_name: SOURCE TYPE
+                        required: true
+                    - params:
```
PSeitz, line 400: The required structure/fields for params is unclear

guilload: That's right. `params` is source specific. I can try to make that clearer.

quickwit-cli/src/source.rs
```diff
@@ -275,6 +334,20 @@ fn flatten_json(value: Value) -> Vec<(String, Value)> {
     acc
 }
 
+/// Tries to read a JSON object from a string, assuming the string is an inline JSON object or a
+/// path to a file holding a JSON object.
+async fn sniff_params(params: &str) -> anyhow::Result<Value> {
```
PSeitz, line 339: Can we use typing here, e.g. a enum of `KafkaParams` or `FileParams`

guilload: For now, I added a source connectivity check that ensure that the JSON value is a valid `KafkaSourceParams` or `FileParams`. I agree with you at some we need a proper enum `SourceParams`.

quickwit-cli/src/source.rs
```diff
@@ -275,6 +336,20 @@ fn flatten_json(value: Value) -> Vec<(String, Value)> {
...
+    if let Ok(object @ Value::Object(_)) = serde_json::from_str(params) {
+        return Ok(object);
+    }
+    let params_uri = Uri::try_new(params)?;
+    let params_bytes = load_file(&params_uri).await?;
+    if let Ok(object @ Value::Object(_)) = serde_json::from_slice(params_bytes.as_slice()) {
```
fmassot, line 347: first time I saw `@` in our base code, really cool :)

quickwit-cli/src/cli.yaml
```diff
@@ -379,6 +379,35 @@ subcommands:
...
         subcommands:
+            - add:
+                about: Adds a new source.
+                args:
+                    - index:
+                        about: ID of the target index.
+                        long: index
+                        value_name: INDEX
+                        required: true
+                    - id:
```
fmassot, line 390: there's a little inconsistency with the similar command `quickwit index create --index`.

Using `--id` seems relevant also for `index create`. Concerning create/add verb, I'm hesitating to  keep both.
- `quickwit index create --id <INDEX ID>`.
- `quickwit source create --id <SOURCE ID>`.

or 

- `quickwit index add --id <INDEX ID>`.
- `quickwit source add --id <SOURCE ID>`.

guilload: I think `add` makes sense for a source because you're really adding/appending the source to an index. When creating a new index, you're not really adding the index to anything. Although you could argue, you add the index to the metastore :)

If we're willing to sacrifice some semantic accuracy for more consistency, I'd go with `create`.

guilload: As far as command parameter names, yeah `index create --id` seems better. I can make that change f you want.

fmassot: > I think add makes sense for a source because you're really adding/appending the source to an index. When creating a new index, you're not really adding the index to anything. Although you could argue, you add the index to the metastore :)

Yeah, I made the same rationale. Let's keep `add` for this minor command.


> As far as command parameter names, yeah index create --id seems better. I can make that change f you want.

Not sure if it's worth it :/. Let's do that later in another PR.

fmassot (approved):
LGTM

PR #1021, opened 2022-01-03
check SocketAddr when node joins cluster

### LINKED ISSUE #1018: Restarted process joining cluster ends up detected as 2 nodes
To recreate
- create a cluster with 2 nodes. Let's call the node 1 and node 2. And let's assum they start with the cool id red and green
- kill node 1 aka red's process.
- red is detected down and removed from the cluster on green
- restart quickwit on the node 1. This time it will get another node id. say blue.

Expected outcome:
- green should see the following cluster members [blue, green]

Actual outcome:
- green sees blue joined, but also detects that red is back. We end up with [blue, green, red]

PSeitz:
Check SocketAddr when node joins cluster to avoid duplicates when node-id changes

Fixes #1018

### Description

### How was this PR tested?

### DIFF
--- a/quickwit-cli/src/generate_markdown.rs
+++ b/quickwit-cli/src/generate_markdown.rs
@@ -62,17 +62,14 @@ fn markdown_for_subcommand(
         }
         val_opt
     };
-    let long_about_opt: Option<&str> = subcommand_ext
-        .map(|el| el.get("long_about").map(|el| el.as_str()).flatten())
-        .flatten();
+    let long_about_opt: Option<&str> =
+        subcommand_ext.and_then(|el| el.get("long_about").and_then(|el| el.as_str()));
 
-    let note: Option<&str> = subcommand_ext
-        .map(|el| el.get("note").map(|el| el.as_str()).flatten())
-        .flatten();
+    let note: Option<&str> =
+        subcommand_ext.and_then(|el| el.get("note").and_then(|el| el.as_str()));
 
-    let examples_opt: Option<&Vec<Value>> = subcommand_ext
-        .map(|el| el.get("examples").map(|el| el.as_array()).flatten())
-        .flatten();
+    let examples_opt: Option<&Vec<Value>> =
+        subcommand_ext.and_then(|el| el.get("examples").and_then(|el| el.as_array()));
 
     if let Some(about) = long_about_opt {
         if !about.trim().is_empty() {
--- a/quickwit-swim/src/membership.rs
+++ b/quickwit-swim/src/membership.rs
@@ -5,7 +5,6 @@ use std::net::SocketAddr;
 use std::time::Duration;
 
 use rand::prelude::SliceRandom;
-use tracing::info;
 
 use crate::member::{self, ArtilleryMember, ArtilleryMemberState, ArtilleryStateChange};
 
@@ -122,24 +121,23 @@ impl ArtilleryMemberList {
         (suspect_members, down_members)
     }
 
-    // make sure node id is set correctly for the host (in case it changes)
-    pub fn ensure_node_id(&mut self, src_addr: SocketAddr, node_id: &str) {
-        for member in &mut self.members {
-            if member.remote_host() == Some(src_addr) {
-                member.node_id = node_id.to_string();
-            }
+    // Set node id for the host (in case it has changed).
+    pub fn set_node_id(&mut self, src_addr: SocketAddr, node_id: &str) {
+        if let Some(member) = self.get_mut_member_for_host(src_addr) {
+            member.node_id = node_id.to_string();
         }
     }
+
+    // Returns member if found via host and not alive
     pub fn mark_node_alive(
         &mut self,
         src_addr: &SocketAddr,
         node_id: String,
     ) -> Option<ArtilleryMember> {
-        self.ensure_node_id(*src_addr, &node_id);
-        // remove self duplicates
-        let my_node_id = self.current_node_id();
-        self.members
-            .retain(|member| !(member.node_id() == my_node_id && member.remote_host().is_some()));
+        if self.current_node_id() == node_id {
+            return None;
+        }
+        self.set_node_id(*src_addr, &node_id);
         for member in &mut self.members {
             if member.remote_host() == Some(*src_addr)
                 && member.state() != ArtilleryMemberState::Alive
@@ -201,10 +199,6 @@ impl ArtilleryMemberList {
                     .unwrap();
                 let update_member = update_member.member_by_changing_host(new_host);
 
-                if existing_member.node_id != member_change.node_id() {
-                    info!(old_node_id=?existing_member.node_id(), new_node_id=?member_change.node_id(), "updating node id from state change.");
-                    existing_member.node_id = member_change.node_id();
-                }
                 if update_member.state() != existing_member.state() {
                     existing_member.set_state(update_member.state());
                     existing_member.incarnation_number = update_member.incarnation_number;
@@ -275,9 +269,46 @@ impl ArtilleryMemberList {
         self.members.retain(|member| member.node_id() != id)
     }
 
+    /// `get_mut_member_for_host` will return artillery member if the given host is matched with any
+    /// of the member in the cluster.
+    pub fn get_mut_member_for_host(&mut self, host: SocketAddr) -> Option<&mut ArtilleryMember> {
+        self.members
+            .iter_mut()
+            .find(|m| m.remote_host() == Some(host))
+    }
+
     /// `get_member` will return artillery member if the given uuid is matched with any of the
     /// member in the cluster.
     pub fn get_member(&self, id: &str) -> Option<ArtilleryMember> {
         self.members.iter().find(|&m| m.node_id() == *id).cloned()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn membership_test() {
+        let current = ArtilleryMember::current("myid".to_string());
+        let mut members = ArtilleryMemberList::new(current);
+        let address = "127.0.0.1:8080".parse().unwrap();
+        members.add_member(ArtilleryMember::new(
+            "other_node".to_string(),
+            address,
+            0,
+            ArtilleryMemberState::Suspect,
+        ));
+
+        members.mark_node_alive(&address, "myid".to_string());
+
+        assert_eq!(
+            members
+                .available_nodes()
+                .iter()
+                .filter(|node| node.node_id() == "myid")
+                .count(),
+            1
+        );
     }
 }
--- a/quickwit-swim/src/membership.rs
+++ b/quickwit-swim/src/membership.rs
@@ -1,4 +1,6 @@
 use std::collections::HashSet;
+use std::fmt;
+use std::fmt::{Debug, Formatter};
 use std::net::SocketAddr;
 use std::time::Duration;
 
@@ -9,6 +11,14 @@ use crate::member::{self, ArtilleryMember, ArtilleryMemberState, ArtilleryStateC
 pub struct ArtilleryMemberList {
     members: Vec<ArtilleryMember>,
     periodic_index: usize,
+}
+
+impl Debug for ArtilleryMemberList {
+    fn fmt(&self, fmt: &mut Formatter) -> fmt::Result {
+        fmt.debug_struct("ArtilleryEpidemic")
+            .field("members", &self.members)
+            .finish()
+    }
 }
 
 impl ArtilleryMemberList {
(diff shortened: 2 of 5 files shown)


quickwit-swim/src/membership.rs
```diff
@@ -133,55 +125,74 @@ impl ArtilleryMemberList {
...
+    fn find_member<'a>(
+        members: &'a mut [ArtilleryMember],
+        node_id: &str,
+        addr: SocketAddr,
+    ) -> Option<&'a mut ArtilleryMember> {
+        // ugly hack for borrowchecker
+        if members.iter().any(|member| member.node_id() == node_id) {
+            members
+                .iter_mut()
+                .find(|member| member.node_id() == node_id)
```
fulmicoton, line 137: why?

quickwit-cli/src/generate_markdown.rs
```diff
@@ -62,17 +62,14 @@ fn markdown_for_subcommand(
         }
         val_opt
     };
-    let long_about_opt: Option<&str> = subcommand_ext
-        .map(|el| el.get("long_about").map(|el| el.as_str()).flatten())
-        .flatten();
+    let long_about_opt: Option<&str> =
```
fmassot, line 65: please don't mix as much as you can nonrelated change with your PR.

quickwit-swim/src/state.rs
```diff
@@ -329,10 +329,15 @@ impl ArtilleryEpidemic {
         use Request::*;
 
         if message.cluster_key == self.config.cluster_key {
-            // We want to abort if a new member has the same node_id of an existing member.  
+            // We want to abort if a new member has the same node_id of an existing member.
```
fmassot, line 332: I'm wondering why there is a change in this file. Is this coming from `make fmt`?

quickwit-swim/src/membership.rs
```diff
@@ -133,55 +125,74 @@ impl ArtilleryMemberList {
...
         let mut changed_nodes = Vec::new();
         let mut new_nodes = Vec::new();
 
-        let my_node_id = self.mut_myself().node_id();
+        let my_node_id = self.current_node_id();
 
         for state_change in state_changes {
-            let new_member_data = state_change.member();
-            let old_member_data = current_members.entry(new_member_data.node_id());
+            let member_change = state_change.member();
```
fmassot, line 158: `updated_member` seems better

quickwit-swim/src/membership.rs
```diff
@@ -133,55 +125,74 @@ impl ArtilleryMemberList {
         None
     }
 
+    fn find_member<'a>(
+        members: &'a mut [ArtilleryMember],
+        node_id: &str,
+        addr: SocketAddr,
+    ) -> Option<&'a mut ArtilleryMember> {
+        // ugly hack for borrowchecker
+        if members.iter().any(|member| member.node_id() == node_id) {
```
fulmicoton, line 134: nitpick... I'd use position here.

```
let pos_matching_node_id: Option<usize> =
      members.iter().position(|member| member.node_id() == node_id);
let pos_matching_address: Option<usize> = 
      members.iter().position(|member| member.remote_host() == Some(addr));
let chosen_position: usize = pos_matching_node_id.or(pos_matching_address)?;
&mut members[chosen_position]
```

quickwit-swim/src/membership.rs
```diff
@@ -133,55 +125,74 @@ impl ArtilleryMemberList {
...
+                &mut current_members,
+                &member_change.node_id(),
+                member_change.remote_host().unwrap_or(*from),
+            ) {
+                let update_member =
+                    member::most_uptodate_member_data(member_change, existing_member).clone();
+                let new_host = update_member
+                    .remote_host()
+                    .or_else(|| existing_member.remote_host())
+                    .unwrap();
```
fulmicoton, line 175: I'm super unfamiliar with this code base.
why is this unwrap safe?

fulmicoton (approved):

quickwit-swim/src/membership.rs
```diff
@@ -133,55 +125,74 @@ impl ArtilleryMemberList {
...
+                let update_member = update_member.member_by_changing_host(new_host);
+
+                if update_member.state() != existing_member.state() {
+                    existing_member.set_state(update_member.state());
+                    existing_member.incarnation_number = update_member.incarnation_number;
```
fmassot, line 180: don't really understand why you are setting the incarnation number

quickwit-swim/src/membership.rs
```diff
@@ -133,55 +125,74 @@ impl ArtilleryMemberList {
...
-                            entry.insert(new_member.clone());
-                            changed_nodes.push(new_member);
-                        }
+            } else if let Some(existing_member) = Self::find_member(
+                &mut current_members,
+                &member_change.node_id(),
+                member_change.remote_host().unwrap_or(*from),
+            ) {
+                let update_member =
+                    member::most_uptodate_member_data(member_change, existing_member).clone();
```
fmassot, line 171: When looking at `most_uptodate_member_data` this will not work. `update_member` will just keep the same node_id because in the common case where you kill one node and restart immediately, `member_change` and `existing_member` will still have the same state and same incarnation number and in this case, `most_uptodate_member_data` will return the `existing_member`. And now you lose the new `node_id`

quickwit-swim/src/membership.rs
```diff
@@ -133,55 +125,74 @@ impl ArtilleryMemberList {
...
-                            .remote_host()
-                            .or_else(|| entry.get().remote_host())
-                            .unwrap();
-                        let new_member = new_member.member_by_changing_host(new_host);
-
-                        if new_member.state() != entry.get().state() {
-                            entry.insert(new_member.clone());
-                            changed_nodes.push(new_member);
-                        }
+            } else if let Some(existing_member) = Self::find_member(
```
fmassot, line 165: Existing member node_id should be updated.

fmassot:
Currently, the problem is that we don't update the `node_id` of the existing member. This leads to weird logs because we continue to see old node_id in `quickwit-cluster/src/cluster.rs` and I guess this can lead to weird stuff because we continue to have members with the wrong `node_id` in `ArtilleryMemberList`.

The shortest path to fix this is to update the `existing_member` `node_id`.

We could go one step further by adding a event type in `ArtilleryMemberEvent` like `NodeIdChange` that we can send only when the existing member and new member are in state `Alive`, same host, and with a `node_id` change. This way we will be able to have nice logs.

quickwit-swim/src/membership.rs
PSeitz: because the method above may select the one with the higher number
```
                let update_member =
                    member::most_uptodate_member_data(member_change, existing_member).clone();
```

quickwit-swim/src/state.rs
PSeitz: probably because of the nighlty fmt

quickwit-swim/src/membership.rs
PSeitz: nice I never thought about using `?` on an Option

quickwit-swim/src/membership.rs
```diff
@@ -133,55 +152,75 @@ impl ArtilleryMemberList {
...
+            ) {
+                let update_member =
+                    member::most_uptodate_member_data(member_change, existing_member).clone();
+                let new_host = update_member
+                    .remote_host()
+                    .or_else(|| existing_member.remote_host())
+                    .unwrap();
+                let update_member = update_member.member_by_changing_host(new_host);
+
+                if update_member.state() != existing_member.state() {
```
fmassot, line 205: This means that you are updating the `existing_member` `node_id` only on a state change. If a node restart is fast, I think we will get the same state.
Why not consider an update of the `node_id` when the state is the same `Alive`? (I did it locally this morning and was happy with that ;))

quickwit-swim/src/membership.rs
PSeitz: This code was there before.
This works because all members except self have a host assigned and there is a self check.

```
pub fn is_current(&self) -> bool {
    self.remote_host.is_none()
}
```

quickwit-swim/src/membership.rs
PSeitz: yep

quickwit-swim/src/membership.rs
PSeitz: the node id update should be covered by Heartbeat(node_id) -> Ack(node_id)

quickwit-swim/src/membership.rs
```diff
@@ -133,55 +153,84 @@ impl ArtilleryMemberList {
...
+        addr: SocketAddr,
+    ) -> Option<&'a mut ArtilleryMember> {
+        let pos_matching_node_id: Option<usize> = members
+            .iter()
+            .position(|member| member.node_id() == node_id);
+        let pos_matching_address: Option<usize> = members
+            .iter()
+            .position(|member| member.remote_host() == Some(addr));
+        let chosen_position: usize = pos_matching_node_id.or(pos_matching_address)?;
+        Some(&mut members[chosen_position])
```
fmassot, line 168: So `find_member` is always returning a member? Why keeping the option?

quickwit-swim/src/state.rs
```diff
@@ -220,10 +238,14 @@ impl ArtilleryEpidemic {
     }
 
     fn enqueue_seed_nodes(&self) {
-        for seed_node in &self.seed_queue {
+        for seed_node in self
+            .seed_queue
+            .iter()
+            .filter(|addr| !self.members.has_member(addr))
```
fmassot, line 244: nice

fmassot:
I tested it with 3 nodes and it seems to work, it's really hard to test these things :/

2 observations:
- when there is a node that is restarting several times in a row, we will have temporary several members with the same host in the cluster, after a few seconds everything is back to normal. This is not what we want... but as we are relying on a map of socket addresses on the search client pool, this has no bad consequence on the search side because the search client pool only sees the right number of clients.
- I saw added the removal of down nodes, this was not the case before. The pro is that we don't keep a growing list of dead nodes and as you keep the list of seeds, we can keep pinging them even if there are down.

Looks good to me.

quickwit-swim/src/membership.rs
```diff
@@ -119,7 +122,24 @@ impl ArtilleryMemberList {
         (suspect_members, down_members)
     }
 
-    pub fn mark_node_alive(&mut self, src_addr: &SocketAddr) -> Option<ArtilleryMember> {
+    // make sure node id is set correctly for the host (in case it changes)
```
fmassot, line 125:
```suggestion
    // Set node id for the host (in case it has changed).
```

quickwit-swim/src/membership.rs
```diff
@@ -119,7 +122,24 @@ impl ArtilleryMemberList {
...
+    pub fn ensure_node_id(&mut self, src_addr: SocketAddr, node_id: &str) {
+        for member in &mut self.members {
+            if member.remote_host() == Some(src_addr) {
+                member.node_id = node_id.to_string();
+            }
+        }
+    }
+    pub fn mark_node_alive(
```
fmassot, line 133:
```suggestion

    pub fn mark_node_alive(
```

fmassot (approved):
LGTM

quickwit-swim/src/membership.rs
```diff
@@ -119,7 +122,24 @@ impl ArtilleryMemberList {
         (suspect_members, down_members)
     }
 
-    pub fn mark_node_alive(&mut self, src_addr: &SocketAddr) -> Option<ArtilleryMember> {
+    // make sure node id is set correctly for the host (in case it changes)
+    pub fn ensure_node_id(&mut self, src_addr: SocketAddr, node_id: &str) {
```
fulmicoton, line 126: Can this method be called with our own `src_addr` and do what happens if it happens?

fulmicoton, line 126: `ensure` suggests there is some kind of check in your method.
`set` would be better.

`set_node_id(_, _)`

quickwit-swim/src/membership.rs
```diff
@@ -119,7 +122,24 @@ impl ArtilleryMemberList {
...
+    pub fn mark_node_alive(
+        &mut self,
+        src_addr: &SocketAddr,
+        node_id: String,
+    ) -> Option<ArtilleryMember> {
+        self.ensure_node_id(*src_addr, &node_id);
+        // remove self duplicates
+        let my_node_id = self.current_node_id();
+        self.members
+            .retain(|member| !(member.node_id() == my_node_id && member.remote_host().is_some()));
```
fulmicoton, line 142: I don't understand this. How does this remove self duplicates?

fulmicoton (changes requested):
Can you write the list of invariant you want to protect on the state,
and enforce them on accessors?

It is too difficult to proofread right now.

quickwit-swim/src/membership.rs
PSeitz: Remove anything that has our id and a host set, because members with hosts are not self.

But I think this should be removed, we can avoid adding it in the first place.

quickwit-swim/src/membership.rs
PSeitz: yes, this can happen currently. We already have self member set with `remote_host: None`.

the "remove self duplicates" block removed the duplicate, but I will add a guard here to avoid this.

fulmicoton (approved):

PR #1028, opened 2022-01-04
Better release workflow

evanxg852000:
### Description
Closes https://github.com/quickwit-inc/quickwit/issues/990

- [x] Refactored release 
- [x] Implemented nightly build for `quickwit/quickwit-nightly` on  `everyday @ 5:00am`
- [x] Removed old way of testing binary build  (`make-binaries-from-ci`)
- [x] Add a `--allow-any-latest-version` flag to install script for pulling any latest release

### How was this PR tested?
Executing the workflow


docker/quickwit.yaml
```diff
@@ -0,0 +1,58 @@
+# ============================ Quickwit Configuration ==============================
```
fmassot, line 1: it's true that copying the config file is quite ugly too :/

fmassot (approved):
LGTM

PR #1029, opened 2022-01-04
Complete CLI docs.

fmassot:
Ready for review.

Wait for #1047 before merging.


docs/reference/cli.md
```diff
@@ -115,18 +94,17 @@ quickwit index create
 
 *Examples*
 
-*Create a new index with metastore in the local `metastore` folder.*
+*Create a new index with the default local metastore.*
```
fmassot, line 97: This is not exact.

docs/reference/cli.md
```diff
@@ -24,33 +24,15 @@ Before using Quickwit with an object storage, check out our [advice](../administ
 
 `quickwit <command name> --help` displays the documentation for the command and a usage example.
 
-## Environment Variables
```
PSeitz, line 27: In the last review you asked to put the QW_CONFIG env at the top, since it simpliyfies the whole usage. I think that makes sense

docs/reference/cli.md
```diff
@@ -63,34 +45,31 @@ The line `Telemetry enabled` disappears when you disable it.
...
 
-* `command`:index, split, source, service. 
-
-
-
-
+* `command`: `index`, `split`, `source` and `service`. 
 
 
+[comment]: <> (Insert auto generated CLI docs from here.)
```
PSeitz, line 59: is this ignored by docusaurus?

docs/reference/cli.md
fmassot: Yes indeed, but I only wanted to talk about `QW_CONFIG` which is the most important and leave other env variables at the end.

docs/reference/cli.md
```diff
@@ -115,18 +94,17 @@ quickwit index create
 
 *Examples*
 
-*Create a new index with metastore in the local `metastore` folder.*
+*Create a new index with the default local metastore.*
 ```bash
-quickwit index create --index-config-uri ./wikipedia_index_config.yaml  --config=quickwit.yaml
+quickwit index create --index wikipedia --index-config wikipedia_index_config.yaml  --config=./config/quickwit.yaml
```
PSeitz, line 99: I prefer it without subfolder since it's simpler for the user
`--config=quickwit.yaml`

docs/reference/cli.md
```diff
@@ -115,18 +94,17 @@ quickwit index create
...
 ```
 
 ### index ingest
 
 Indexes a dataset consisting of newline-delimited JSON objects located at `input-path` or read from *stdin*. 
-The data is appended to the target index specified by `index` unless `overwrite` is passed. `input-path` can be a file or another command output piped into stdin. 
-Currently, only local datasets are supported. 
-By default, quickwit's indexer will work with a heap of 2 GiB of memory, but this can be set with the `heap-size` option in the indexing_settings.resources section in the config. 
-This does not directly reflect the overall memory usage of `quickwit index ingest`, but doubling this value should give a fair approximation.
```
PSeitz, line 129: this is useful information, or is it outdated?

docs/reference/cli.md
```diff
@@ -388,12 +364,13 @@ Starts a service. Currently, the only services available are `indexer` and `sear
 `quickwit service run [args]`
 ### service run searcher
 
-Starts a web server that exposes the [Quickwit REST API](search-api.md). 
-The node can optionally join a cluster using the `peer_seed` parameter (quickwit.yaml). 
-This list of node addresses is used to discover the remaining peer nodes in the cluster through the use of a gossip protocol (SWIM).
+Starts a web server at `rest_listing_address:rest_list_port` that exposes the [Quickwit REST API](search-api.md)
+where `rest_listing_address` and `rest_list_port` are defined in Quickwit config file (quickwit.yaml).
```
PSeitz, line 368:
```suggestion
where `listen_address` and `rest_listen_port` are defined in the Quickwit config file (quickwit.yaml).
```

docs/reference/cli.md
```diff
@@ -419,24 +396,30 @@ quickwit service run searcher
...
 ```bash
- curl -L --http2-prior-knowledge "http://127.0.0.1:7280/api/v1/myindex/search/stream?query=clinton&searchField=body,title&fastField=number&outputFormat=clickHouseRowBinary"  
+curl -L --http2-prior-knowledge "http://127.0.0.1:7280/api/v1/myindex/search/stream?query=clinton&searchField=body,title&fastField=number&outputFormat=clickHouseRowBinary"  
 ```
 
 ### service run indexer
 
-Starts a indexing server that consumes the sources from the `quickwit source add` command. The server can be terminated when all sources are consumed.
+Starts an indexing server that consumes the sources of index IDs passed in `--indexes` argument.
+If all sources of all indexes are exhausted, the server will stop.
```
PSeitz, line 422: Last time I checked the server did just idle forever

docs/reference/cli.md
```diff
@@ -457,12 +440,9 @@ quickwit service run indexer
...
 *Examples*
 
-*Add and consume source*
+*Start an Indexer*
 ```bash
- 
-quickwit source add --id wikipedia_data_1 --index wikipedia --params '{"filepath":"wiki-articles_1.json"}' --type file
-quickwit service run indexer --indexes wikipedia
-
+quickwit service run indexer --config=./config/quickwit.yaml
```
PSeitz, line 445: I prefer the examples to be more complete, and indexer without sources doesn't do anything

docs/reference/cli.md
```diff
@@ -491,6 +471,33 @@ quickwit source add
...
+{
+  "topic": "wikipedia",
+  "client_params": {
+    "bootstrap.servers": "localhost:9092",
+    "group.id": "my-group-id",
+    "security.protocol": "SSL"
+  }
+}
+EOF
+quickwit source add --index wikipedia --id wikipedia-source --type kafka --params wikipedia-source.json
```
PSeitz, line 497: I think replacing `id` with `source` would be more consistent
e.g.
`quickwit index create --index wikipedia` (not id)

docs/reference/cli.md
```diff
@@ -178,13 +156,22 @@ quickwit index describe
 
 `--index` ID of the target index.    
 `--config` Quickwit config file.    
+
+*Examples*
+
+*Displays descriptive statistics of your index*
+```bash
+quickwit index describe --index wikipedia --config ./config/quickwit.yaml
```
PSeitz, line 164: Not sure, but we could add example output?

```suggestion
quickwit index describe --index wikipedia --config ./config/quickwit.yaml

1. General infos
===============================================================================
Index id:                           wikipedia
Index uri:                          file:///home/quickwit-indices/qwdata/indexes/wikipedia
Number of published splits:         1
Number of published documents:      300000
Size of published splits:           448 MB

2. Statistics on splits
===============================================================================
Document count stats:
Mean ± σ in [min … max]:            300000 ± 0 in [300000 … 300000]
Quantiles [1%, 25%, 50%, 75%, 99%]: [300000, 300000, 300000, 300000, 300000]

Size in MB stats:
Mean ± σ in [min … max]:            448 ± 0 in [448 … 448]
Quantiles [1%, 25%, 50%, 75%, 99%]: [448, 448, 448, 448, 448]


```

docs/reference/cli.md
```diff
@@ -204,9 +191,9 @@ quickwit index search
 
 *Options*
 
-`--index` id of the target index.    
+`--index` ID of the target index.    
 `--config` Quickwit config file.    
-`--query` Query expressed in Tantivy syntax.    
+`--query` Query expressed in natural query language (barack AND obama) OR "president of united states").
```
PSeitz, line 196: I think we need a link here or reference tantivy syntax, or we let the user guess the actual syntax

docs/reference/cli.md
fmassot: Hope so, will see that very soon.

docs/reference/cli.md
fmassot: I assume the user have a fresh install of quickwit and thus has the YAML file in the config dir. I think both are ok.

docs/reference/cli.md
fmassot: I prefer to have the details on the index config page to avoid duplication. Though the writing is not really good here, I will improve it.

docs/reference/cli.md
fmassot: good idea

docs/reference/cli.md
fmassot: We need to check that :/

docs/reference/cli.md
fmassot: good point, I will keep it.

docs/reference/cli.md
fmassot: I had this discussion with @guilload but we decided to not touch it for now

docs/reference/cli.md
fmassot: I put a link to the query language in the description.

docs/reference/cli.md
PSeitz: I don't understand how a new install is related to having the config file in the subfolder

docs/reference/cli.md
PSeitz: I think it would be good to have it also in the cli for users that explore the cli without the docs

docs/reference/cli.md
PSeitz: Yes it's still the case

fmassot: Yes, I read the code and the expected behavior is to wait until the `IndexingServer` actor exits. And this will not happen unless there is an error happens when processing a message. I'm removing the sentence.

docs/reference/cli.md
fmassot: Will do.

docs/reference/rest-api.md
```diff
@@ -61,7 +61,7 @@ Search for documents matching a query in the given index `<index name>`.
...
 | **startOffset**            | `Integer`     	    | Number of documents to skip                                                                | `0`                                                                                             |
 | **maxHits**                | `Integer`          | Maximum number of hits to return (by default 20)                                                            | `20`                                                                                            |
-| **searchFields**           | `String`      		  | Fields to search on. Comma-separated list, e.g. "field1,field2" | index_config.search_settings.default_search_fields                                                                                             |
+| **searchFields**           | `String`      		  | Fields to search on if no field name is specified in the query. Comma-separated list, e.g. "field1,field2" | index_config.search_settings.default_search_fields                                                                                             |
```
PSeitz, line 64: If the behavior is like this, this may be a security issue, since most will have query from user input and they can't be constrained to some fields (except by parsing the query)

fmassot: yeah nice spotted, this is wrong, I fixed the wording on the CLI but not here :/

fmassot: Docs updated.

CONTRIBUTING.md
```diff
@@ -55,3 +55,27 @@ We also have nightly builds that are pushed to dockerhub. This helps continiousl
...
+
+## Generating the CLI docs.
+
+The [CLI doc page](docs/reference/cli.md) is partly generated by a script.
+To update it, first run the script:
+```bash
+cargo run --bin generate_markdown > docs/reference/cli_insert.md
+```
+
+Then edit manually edit the [doc page](docs/reference/cli.md) to update it.
```
evanxg852000, line 72:
```suggestion
Then manually edit the [doc page](docs/reference/cli.md) to update it.
```

docs/reference/cli.md
```diff
@@ -3,13 +3,13 @@ title: CLI Reference
 sidebar_position: 1
 ---
 
-Quickwit is a single binary that makes it easy to index and search structured or unstructured data from the command line. It consumes datasets consisting of newline-delimited JSON objects with arbitrary keys. It produces indexes that can be stored locally or remotely on an object storage such as Amazon S3 and queried with subsecond latency.
+Quickwit command line tool lets you create, ingest, search, start search and indexer servers. For configuration, `quickwit` needs a [config file path](config.md) that you can configure with `QW_CONFIG` environment variable: `export QW_CONFIG=./config/quickwit.yaml`.
```
evanxg852000, line 6:
```suggestion
Quickwit command line tool lets you create, ingest, search, start search and indexer servers. For configuration, `quickwit` needs a [config file path](config.md) that you can specify with `QW_CONFIG` environment variable: `export QW_CONFIG=./config/quickwit.yaml`.
```

evanxg852000 (approved):

PR #1030, opened 2022-01-05
Refactor config to put host and port at the root of quickwit config.

### LINKED ISSUE #1020: Putting host and port in common config settings
Currently, the config file lets the user define for the searcher and for the indexer the `rest_listen_address` and `rest_listen_port`. Moreover, we define implicitly the grpc service on `rest_listen_port` + 1 for the searcher. 

What I have in mind for Quickwit is something like this:
- if a user wants to start several services like searcher + indexer on the same machine, I would want to define once for all `rest_listen_address` and `rest_listen_port`. Grpc service is on `rest_listen_port` + 1 by default. I would not want different ports.
- if a user wants to start a single service like the searcher one, having a common `rest_listen_address` and `rest_listen_port` is fine too. 
- there is the special case of UI. In general, people expects a front and a back and I like this separation, so I would have a specific port for the UI.

There is also the gossip configuration that is currently linked to `rest_listen_address` and `rest_listen_port` defined for the searcher. But the gossip protocol will be used for every quickwit instance service (searcher, indexer, metastore server) and consequently we should detach its configuration from the searcher config.

Finally, I would target a config file like this:

```yaml
server_config:
  listen_address: 127.0.0.1
  rest_listen_port: 7280
  grpc_listen_port: 7281

gossip_swim:
  node_id: searcher-1
  listen_address: 7280
  peers: 
     - 172.1.0.12
     - 172.1.0.13

```

What do you think?

You can have a look at Loki config spec: https://grafana.com/docs/loki/latest/configuration/#server_config

fmassot:
Fix #1020

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,12 +32,10 @@ RUN echo "Building workspace with feature(s) '$CARGO_FEATURES' and profile '$CAR
     && mkdir -p /quickwit/bin \
     && find target/$CARGO_PROFILE -maxdepth 1 -perm /a+x -type f -exec mv {} /quickwit/bin \;
 
-# Change the default configuration file in order to make the searcher
-# and indexer services accessible from outside of docker.
+# Change the default configuration file in order to make the rest,
+# grpc servers and gossip accessible from outside of docker.
 COPY ./config/quickwit.yaml ./config/quickwit.yaml
-RUN sed -i 's/#indexer/indexer/g' ./config/quickwit.yaml \
-    && sed -i 's/#searcher/searcher/g' ./config/quickwit.yaml  \
-    && sed -i 's/#  rest_listen_address: 127.0.0.1/ rest_listen_address: 0.0.0.0/g' ./config/quickwit.yaml
+RUN sed -i 's/#listen_address: 127.0.0.1/listen_address: 0.0.0.0/g' ./config/quickwit.yaml
 
 FROM debian:bullseye-slim AS quickwit
 
--- a/config/tutorials/hdfs-logs/searcher-1.yaml
+++ b/config/tutorials/hdfs-logs/searcher-1.yaml
@@ -1,8 +1,7 @@
 version: 0
 node_id: searcher-1
-searcher:
-  rest_listen_address: 127.0.0.1
-  rest_listen_port: 7280
-  peer_seeds:
-    - 127.0.0.1:7290 # searcher-2
-    - 127.0.0.1:7300 # searcher-3
+listen_address: 127.0.0.1
+rest_listen_port: 7280
+peer_seeds:
+  - 127.0.0.1:7290 # searcher-2
+  - 127.0.0.1:7300 # searcher-3
--- a/config/tutorials/hdfs-logs/searcher-2.yaml
+++ b/config/tutorials/hdfs-logs/searcher-2.yaml
@@ -1,8 +1,7 @@
 version: 0
 node_id: searcher-2
-searcher:
-  rest_listen_address: 127.0.0.1
-  rest_listen_port: 7290
-  peer_seeds:
-    - 127.0.0.1:7280 # searcher-1
-    - 127.0.0.1:7300 # searcher-3
+listen_address: 127.0.0.1
+rest_listen_port: 7290
+peer_seeds:
+  - 127.0.0.1:7280 # searcher-1
+  - 127.0.0.1:7300 # searcher-3
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,11 +5,9 @@ RUN apk add --no-cache curl
 RUN curl -L https://install.quickwit.io | sh
 RUN mv  ./quickwit-*/* ./ && ls ./config/
 
-# Change the default configuration file in order to make the searcher
-# and indexer services accessible from outside of docker.
-RUN sed -i 's/#indexer/indexer/g' ./config/quickwit.yaml \
-    && sed -i 's/#searcher/searcher/g' ./config/quickwit.yaml  \
-    && sed -i 's/#  rest_listen_address: 127.0.0.1/ rest_listen_address: 0.0.0.0/g' ./config/quickwit.yaml
+# Change the default configuration file in order to make the rest,
+# grpc servers and gossip accessible from outside of docker.
+RUN sed -i 's/#listen_address: 127.0.0.1/listen_address: 0.0.0.0/g' ./config/quickwit.yaml
 
 FROM alpine:3.15
 LABEL org.opencontainers.image.title="Quickwit"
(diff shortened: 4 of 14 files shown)


Dockerfile
```diff
@@ -32,12 +32,10 @@ RUN echo "Building workspace with feature(s) '$CARGO_FEATURES' and profile '$CAR
...
 
-# Change the default configuration file in order to make the searcher
-# and indexer services accessible from outside of docker.
+# Change the default configuration file in order to make the rest,
+# grpc servers and gossip accessible from outside of docker.
 COPY ./config/quickwit.yaml ./config/quickwit.yaml
-RUN sed -i 's/#indexer/indexer/g' ./config/quickwit.yaml \
-    && sed -i 's/#searcher/searcher/g' ./config/quickwit.yaml  \
-    && sed -i 's/#  rest_listen_address: 127.0.0.1/ rest_listen_address: 0.0.0.0/g' ./config/quickwit.yaml
+RUN sed -i 's/#listen_address: 127.0.0.1/listen_address: 0.0.0.0/g' ./config/quickwit.yaml
```
fmassot, line 38: @evanxg852000 I really would prefer to copy a file instead of playing with `sed` :)

docker/Dockerfile
```diff
@@ -5,11 +5,9 @@ RUN apk add --no-cache curl
...
 RUN mv  ./quickwit-*/* ./ && ls ./config/
 
-# Change the default configuration file in order to make the searcher
-# and indexer services accessible from outside of docker.
-RUN sed -i 's/#indexer/indexer/g' ./config/quickwit.yaml \
-    && sed -i 's/#searcher/searcher/g' ./config/quickwit.yaml  \
-    && sed -i 's/#  rest_listen_address: 127.0.0.1/ rest_listen_address: 0.0.0.0/g' ./config/quickwit.yaml
+# Change the default configuration file in order to make the rest,
+# grpc servers and gossip accessible from outside of docker.
+RUN sed -i 's/#listen_address: 127.0.0.1/listen_address: 0.0.0.0/g' ./config/quickwit.yaml
```
fmassot, line 10: annoying `sed`:P

Dockerfile
evanxg852000: ok let's have a docker dedicated cinfig file then.
Should I add update it in my PR?

fmassot: We could do that after the release, what do you think?

fulmicoton: I think it is ok though. We could really use host interpolation in the conf though. :-S

0.0.0.0 seems a tiny bit dangerous.
What do official image do? Should we listen to 172.17.42.1 for instance?

evanxg852000: I mostly see images with app server binding to `0.0.0.0` within docker (vector, Postgres). probably because it's just easy to run the image this way. I think the relational is docker is ran within a host and should take care of closing the gate to the world. The user can still change that.

config/tutorials/hdfs-logs/searcher-1.yaml
```diff
@@ -1,8 +1,7 @@
...
-searcher:
-  rest_listen_address: 127.0.0.1
-  rest_listen_port: 7280
-  peer_seeds:
-    - 127.0.0.1:7290 # searcher-2
-    - 127.0.0.1:7300 # searcher-3
+listen_address: 127.0.0.1
+rest_listen_port: 7280
+seeds:
+  - 127.0.0.1:7290 # searcher-2
```
evanxg852000, line 6: As you suggested yesterday, we can make it a good practice to specify less seed node (one in this case). we can document to add more seeds as the cluster gets larger thus reducing the likelihood of try to contact a seed that is down.

fmassot: good point

config/tutorials/hdfs-logs/searcher-2.yaml
```diff
@@ -1,8 +1,7 @@
...
 node_id: searcher-2
-searcher:
-  rest_listen_address: 127.0.0.1
-  rest_listen_port: 7290
-  peer_seeds:
-    - 127.0.0.1:7280 # searcher-1
-    - 127.0.0.1:7300 # searcher-3
+listen_address: 127.0.0.1
+rest_listen_port: 7290
+seeds:
```
evanxg852000, line 5: nitpick: `peer_seeds` sound more into our tradition of being clear in naming.

guilload: Yes, I like `peer_seeds` better too.

config/tutorials/hdfs-logs/searcher-1.yaml
guilload: Yes, I think we should demonstrate best practices whenever possible. Best practice for adding peer seeds is the more the merrier.

quickwit-serve/src/lib.rs
```diff
@@ -92,12 +91,13 @@ pub async fn run_searcher(
...
 
-    let grpc_addr = searcher_config.grpc_socket_addr()?;
+    let grpc_addr = quickwit_config.grpc_socket_addr()?;
     let grpc_search_service = GrpcSearchAdapter::from(search_service.clone());
     let grpc_cluster_service = GrpcClusterAdapter::from(cluster_service.clone());
     let grpc_server = start_grpc_service(grpc_addr, grpc_search_service, grpc_cluster_service);
 
-    let rest_socket_addr = searcher_config.rest_socket_addr()?;
+    let rest_socket_addr = quickwit_config.rest_socket_addr()?;
+    println!("rest_socket_addr {:?}", rest_socket_addr);
```
evanxg852000, line 100: let's remove this print

quickwit-config/src/config.rs
```diff
@@ -58,23 +58,19 @@ fn default_listen_address() -> String {
     "127.0.0.1".to_string()
 }
 
+fn default_rest_listen_port() -> u16 {
+    7180
```
evanxg852000, line 62: The default is `7280` 
```suggestion
    7280
```

evanxg852000 (approved):
Looks good overall :+1:  Please fix the default port to `7280` before merging.

config/tutorials/hdfs-logs/searcher-2.yaml
fmassot: Ok I will make a concession with time, but be aware that it cost me a lot.

config/tutorials/hdfs-logs/searcher-1.yaml
fmassot: ok let's keep it as is

Dockerfile
fmassot: let's keep sed then.

Issue #887, opened 2021-12-13
Handling "too many tags" in a way that allows for schema config change of the set of fieldset.


fulmicoton:
We addressed the too many tags problem by simply adding the tag `tag_field:*`, and expanding all tags in the tag filter ast using the rule:

`tag_field:tag_value -> tag_field:tag_value OR tag_field:*`

This works.

However I think there is a smarter way to do this.
What about do not add anything in the metastore if the cardinality of the field exceeds a given limit.

If a tag for a given field is registered on the other hand, we register `tag_field!`.

Finally we expand the tag filter ast using the following rule.

`tag_field:tag_value -> (NOT tag_field! OR tag_field:tag_value)`.

The benefit here is that this approach allow for a change of tag fields in the schema. 

What we lose is the abiilty to prune over the existence of a single doc containing the given field (which is probably not something that we care much about... but we can introduce yet another magic tag to get that back)

fmassot:
Let's take an example with 3 splits and one tag field `tag_field`
- `split_1`: no tag values => metadata contains `tag_field!` 
- `split_2`: one value `tag_field:value`  metadata contains `tag_field:value`
- `split_3`: too many tag values => metadata tags contains only `tag_field:value`

We can see the immediate benefit by adding a tag field like `tag_field_3`:
- `new_tag_field:value` => `(NOT new_tag_field! OR new_tag_field:value)` will match all splits and that's what we want

I did not follow you on this part:
> What we lose is the abiilty to prune over the existence of a single doc containing the given field (which is probably not something that we care much about... but we can introduce yet another magic tag to get that back)

Can you give an example?

fulmicoton:
I would add
- `split_4`: `tag_field` was not configured as a tag field at that point. metadata does not contain `tag_field!`. 

> Can you give an example?
Forget my point. It was not relevant actually.

fmassot:
All right. Let's do that in the 0.2.

fulmicoton:
@r-zenine is making some proper AST for the tag filtering. It should be relatively trivial to do that kind of manipulation.

fulmicoton:
@r-zenine is on a cruisade to handle negation too. The negation will only appear on the leaves.

It is compatible with the solution described in this ticket but 
the transformation rule is NOT the negation of the positive negation rule.

We want:

`not field_a:val_a -> (not field_a!) or (not field_a:val_a)`

PR #1031, opened 2022-01-05
Update demux maturity implementation with the new tag system.

fmassot:
I update the merge policy so that we can correctly identify the splits to demux.

Few notes:
I was very surprised by the usage of the special tag `tag_field!` which indicates that, for a given split, the field is registered as `tagged`, I was expecting a special tag like "indicates that there are no tags in a split" like described in #887 examples. It's working well and I understand why this is like this, it was just me that did not understand correctly the usage.
Just to be clear, I will rewrite the example I wrote in #887:

`split_1`: no tag values => metadata contains `tag_field!`
`split_2`: contains one value `tag_field:value` => metadata contains `tag_field:value` and `tag_field!`
`split_3`: too many tag values => metadata tags is empty
`split_4`: `tag_field` was not configured as a tag field at that point. metadata does not contain `tag_field!`.

@fulmicoton can you confirm that this is the implementation you have in mind when you open issue #887?

I added some documentation as it's not super easy to understand.

### DIFF
--- a/quickwit-indexing/src/merge_policy.rs
+++ b/quickwit-indexing/src/merge_policy.rs
@@ -22,7 +22,7 @@ use std::fmt;
 use std::ops::Range;
 
 use itertools::Itertools;
-use quickwit_doc_mapper::tag_pruning::match_tag_field_name;
+use quickwit_doc_mapper::tag_pruning::{field_tag, match_tag_field_name};
 use quickwit_metastore::SplitMetadata;
 use tracing::debug;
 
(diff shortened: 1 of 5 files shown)


quickwit-indexing/src/merge_policy.rs
```diff
@@ -260,29 +260,30 @@ impl StableMultitenantWithTimestampMergePolicy {
...
-            .tags
-            .iter()
-            .filter(|tag| match_tag_field_name(demux_field_name, tag))
-            .count()
-            < 2;
-        split_tags_contains_less_than_2_demux_values
-            || (split.num_docs < self.split_num_docs_target || split.demux_num_ops > 0)
+        let demux_field_name = self.demux_field_name.as_ref().unwrap();
+        // Note: split metadata tags is empty when demux values have a cardinality
+        // is strictly superior to [`MAX_VALUES_PER_TAG_FIELD`]. When inferior,
```
evanxg852000, line 272:
```suggestion
        // strictly superior to [`MAX_VALUES_PER_TAG_FIELD`]. When inferior,
```

quickwit-indexing/src/merge_policy.rs
```diff
@@ -260,29 +260,30 @@ impl StableMultitenantWithTimestampMergePolicy {
...
+        // tags always contains the special tag `field_tag(demux_field_name)`.
+        // This is why we have to check first if this special tag is present and then
+        // consider the split mature if we have less that one demux value.
```
evanxg852000, line 275:
```suggestion
        // consider the split mature if we have less than one demux value.
```

quickwit-metastore/src/split_metadata.rs
```diff
@@ -78,7 +78,14 @@ pub struct SplitMetadata {
     #[serde(default = "utc_now_timestamp")]
     pub create_timestamp: i64,
 
-    /// A set of tags for categorizing and searching group of splits.
+    /// Set of unique tags values of form `{field_name}:{field_value}`.
+    /// The set is filled at indexing with values from each field registered
+    /// in the [`DocMapping`] `tag_fields` attribute and only when cardinality
+    /// of a given field is less or equal to [`MAX_VALUES_PER_TAG_FIELD`].
+    /// A additional special tag of form `{field_name!}` is added to the set
```
evanxg852000, line 85:
```suggestion
    /// An additional special tag of the form `{field_name}!` is added to the set
```

evanxg852000 (commented):
I left comments on minor things

fulmicoton:
@fmassot this is correct.

evanxg852000 (approved):

quickwit-indexing/src/merge_policy.rs
fmassot: I still need to count() all tags like `field_name:value`. The count can be 0 or 1.

PR #1038, opened 2022-01-07
Support both postgres and postgresql as protocol in the

### LINKED ISSUE #1015: Support both postgres and postgresql as a protocol for the metastore

fulmicoton:
PostgresMetastore.

Both seem to be supported by other tools like psql.

Closes #1015.


fmassot (approved):
Good for me!

Issue #1004, opened 2021-12-31
Confusing log


fulmicoton:
I successfully indexed a dataset but the last warning line is confusing.
```
2021-12-31T01:14:56.387Z  INFO {actor=quickwit_indexing::actors::indexing_server::IndexingServer}:{msg_id=1}::{index=gh-archive gen=0}:{actor=Packager}:{msg_id=549 num_splits=1}: quickwit_indexing::actors::packager: commit-split split_id=01FR70SWCC1KW0REBEX1PDGSE0
2021-12-31T01:14:56.392Z  INFO {actor=quickwit_indexing::actors::indexing_server::IndexingServer}:{msg_id=1}::{index=gh-archive gen=0}:{actor=Indexer}:{msg_id=6871}: quickwit_indexing::actors::indexer: new-split split_id=01FR70VQ032WR8EZK4SRV2AN12
2021-12-31T01:14:56.544Z  INFO {actor=quickwit_indexing::actors::indexing_server::IndexingServer}:{msg_id=1}:: quickwit_actors::async_actor: actor-exit actor_id=IndexingPipeline-fragrant-rTvx exit_status=Success
 Num docs 3615659 Parse errs     0 PublSplits   1 Input size  3409MB Thrghput 19.92MB/s Time 00:02:01
 Num docs 3615659 Parse errs     0 PublSplits   1 Input size  3409MB Thrghput 17.14MB/s Time 00:02:01

Indexed 3615659 documents in 2.02min.
Now, you can query the index with the following command:
quickwit index search --index gh-archive --config ./config/quickwit.yaml --query "my query"
2021-12-31T01:14:57.343Z  WARN quickwit_actors::kill_switch: kill-switch-activated
```

guilload:
That's probably when the universe is dropped.

fmassot:
this is quite ugly

guilload:
I'll take a stab at it.

PR #1042, opened 2022-01-07
Set log level to debug for kill switch activation message

guilload:
As discussed, setting log level to debug for kill switch activation message.


guilload:
Fixes #1004.

fulmicoton (approved):

Issue #1046, opened 2022-01-07
`index_uri` is not validated at index creation


fmassot:
We can do 2 validations:
- check that the URI has the form `{protocol}://{any-path}`
- check that the protocol is supported. Today we support 3 protocols: `file`, `s3`, `s3+localstack`

fmassot:
closed by #1047

Issue #1041, opened 2022-01-07
Put the index-uri parameter in the IndexConfig


fulmicoton:
Right now we set the idnex uri in the CliArgs. Move it into the index config.

PR #1047, opened 2022-01-07
Put index-uri and index-id in IndexConfig, update --id to --source fo…

fmassot:
…r source add CLI, use data_dir in config.

This PR fixes:
- [x] #1041 
- [x] #1046 

It does not fix the docs, it's on purpose and it will be done in #1029 

Notes:
1. I finally put the `index_id` in the config as it would be easier to introduce an `apply` command
2. I put `index_uri` as an Option in `IndexConfig` to be able to parse the index config file valid even without `index_uri` and if absent, resolve the index URI to the default root URI + index id. I'm not satisfied by this choice as... an `IndexConfig` should always have an `index-uri`. But I did not find a simple solution to avoid this option unless having a dedicated struct for creation only or by implementing a custom deserializer.

### DIFF
--- a/quickwit-cli/src/cli.yaml
+++ b/quickwit-cli/src/cli.yaml
@@ -12,20 +12,11 @@ subcommands:
                 about: Creates an index.
                 display_order: 1
                 args:
-                    - index:
-                        about: ID of the target index.
-                        long: index
-                        value_name: INDEX
-                        required: true
                     - index-config:
                         about: Location of the index config file.
                         long: index-config
                         value_name: INDEX CONFIG
                         required: true
-                    - index-uri:
-                        about: Index data (or splits) storage URI.
-                        long: index-uri
-                        value_name: INDEX URI
                     - config:
                         about: Quickwit config file.
                         long: config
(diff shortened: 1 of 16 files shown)


quickwit-config/src/config.rs
```diff
@@ -167,6 +167,7 @@ pub struct QuickwitConfig {
     #[serde(default = "default_metastore_and_index_root_uri")]
     pub default_index_root_uri: String,
     #[serde(default = "default_data_dir_path")]
+    #[serde(rename = "data_dir")]
```
fmassot, line 170: I thought we were already using `data_dir` in the conf but it wasn't. I like keeping the short name for the config and the long name in the code. It's also in line with the env variable `QW_DATA_DIR`.

quickwit-cli/src/cli.yaml
```diff
@@ -372,9 +363,9 @@ subcommands:
                         long: index
                         value_name: INDEX
                         required: true
-                    - id:
+                    - source:
```
fmassot, line 366: @PSeitz: renamed it to `source` to remain consistent on the CLI side.

fulmicoton:
Ok. For the record, the case against internalizing `index_id` was alignment with the REST endpoint. We can find a workaround of course.

fulmicoton (approved):

PR #1057, opened 2022-01-10
update cloud env

PSeitz:
### Description

### How was this PR tested?

### DIFF
--- a/docs/administration/cloud-env.md
+++ b/docs/administration/cloud-env.md
@@ -1,25 +1,40 @@
 ---
-title: Operating in the cloud
+title: Cloud operation costs
 sidebar_position: 1
 ---
 
-Quickwit has only been tested on AWS S3. This page sums up what we have learned from that.
+Quickwit has been tested on Amazon S3. This page sums up what we have learned from that experience.
 
 ## Data transfers costs and latency
 
-Cloud providers charge for data transfers in and out of their networks. In addition, querying an index from a remote machine adds some extra latency. For those reasons, we recommend that you test and use the Quickwit from an instance located within your cloud provider's network.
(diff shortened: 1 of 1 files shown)


docs/administration/cloud-env.md
```diff
@@ -1,25 +1,40 @@
...
+During indexing, Quickwit uploads new splits on Amazon S3 and progressively merges them until they reach 10 million documents that we call “mature splits”. Such splits have a typical size between 1GB and 10GB and will usually require 2 PUT requests to be uploaded (1 PUT request / 5GB).
+
+With default indexing parameters `commit_timeout_secs` of 60 seconds and `merge_policy.merge_factor` of 10 and assuming you want to ingest 1 million documents every minute, this will cost you less than $1 / month.
+
+### GET requests
+
+When querying, Quickwit needs to make multiple GET requests:
+
+```jsx
+#num requests = #num splits * ((#num search fields * #num terms * 3) + #num fast fields)
```
fmassot, line 32: I would prefer to not put fast field number as today I think we are only getting the timestamp fast field values, right? So I would put only 1.

PSeitz: Yes you are right, timestamp and sort_by

```

/// Extracts all fast field names.
fn extract_fast_field_names(doc_mapper: &dyn DocMapper) -> HashSet<String> {
    let mut fast_fields = HashSet::new();
    if let Some(timestamp_field) = doc_mapper.timestamp_field_name() {
        fast_fields.insert(timestamp_field);
    }
    if let SortBy::FastField { field_name, .. } = doc_mapper.sort_by() {
        fast_fields.insert(field_name);
    }
    fast_fields
}

```

PR #1058, opened 2022-01-10
Use an enum rather than JSON value for storing source params

guilload:
### Description
Use an enum rather than JSON value for storing source params, which indirectly validates the content of the params provided by the user. To get this PR out in a few hours for the release, I took a few shortcuts :/

### How was this PR tested?
In progress

### DIFF
--- a/quickwit-config/src/index_config.rs
+++ b/quickwit-config/src/index_config.rs
@@ -33,7 +33,7 @@ use quickwit_doc_mapper::{
 use serde::{Deserialize, Serialize};
 
 use crate::source_config::SourceConfig;
-use crate::SourceType;
+use crate::SourceParams;
 
 // Note(fmassot): `DocMapping` is a struct only used for
 // serialization/deserialization of `DocMapper` parameters.
@@ -279,7 +279,7 @@ impl IndexConfig {
 
         // We want to forbid file source with no filepath.
         for source in self.sources.iter() {
-            if let SourceType::File(file_source_params) = &source.source_type {
+            if let SourceParams::File(file_source_params) = &source.source_params {
                 if file_source_params.filepath.is_none() {
                     bail!(
                         "Source `{}` of type `file` must contain a `filepath`",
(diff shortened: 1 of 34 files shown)


quickwit-config/src/index_config.rs
```diff
@@ -275,6 +277,18 @@ impl IndexConfig {
             bail!("Index config contains duplicate sources.")
         }
 
+        // We want to forbid file source with no filepath.
```
fulmicoton, line 280:
```suggestion
Refactor
```

fulmicoton, line 280:
```suggestion
        // TODO refactor. This should be a method on `SourceType`.
        // We want to forbid file source with no filepath.
```

quickwit-config/src/source_config.rs
```diff
@@ -0,0 +1,174 @@
...
+
+use quickwit_common::uri::Uri;
+use serde::de::Error;
+use serde::{Deserialize, Deserializer, Serialize};
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct SourceConfig {
+    pub source_id: String,
+    #[serde(flatten)]
+    pub source_type: SourceType,
```
fulmicoton, line 30: I expect `SourceType` to be a `#[derive(Clone,Copy)] enum`.

Not something that holds a param object.
Maybe rename `SourceParams` ?

quickwit-config/src/source_config.rs
```diff
@@ -0,0 +1,174 @@
...
+pub struct FileSourceParams {
+    /// Path of the file to read. Assume stdin if None.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    #[serde(deserialize_with = "absolute_filepath_from_str")]
+    pub filepath: Option<PathBuf>, //< If None read from stdin.
+}
+
+// Deserializing a filepath string into an absolute filepath.
+fn absolute_filepath_from_str<'de, D>(deserializer: D) -> Result<Option<PathBuf>, D::Error>
```
fulmicoton, line 98: here it is canonical.
absolute is slightly different from canonical.
canonical resolve links.

It's raises an interesting question. Here we might actually prefer absolute.

quickwit-config/src/source_config.rs
fmassot: yes indeed

fulmicoton (approved):

PR #1068, opened 2022-01-11
Making the index reference doc up to date.

fulmicoton:
We recently integrated index-id and index-uri in
the index-reference.

In addition this PR makes more minor change like
adding a source in the example configuration.

### DIFF
--- a/docs/reference/index-config.md
+++ b/docs/reference/index-config.md
@@ -5,8 +5,9 @@ position: 4
 
 This page describes how to configure an index.
 
-The index configuration lets you define four items:
+In addition to the `index_id`, the index configuration lets you define five items:
 
+- The **index-uri**: it defines where the index files should be stored.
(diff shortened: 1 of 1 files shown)


docs/reference/index-config.md
```diff
@@ -40,13 +46,39 @@ doc_mapping:
...
+     client_params:
+       bootstrap.servers: localhost:9092
+       group.id: quickwit-consumer-group
+       security.protocol: SSL
 ```
 
+## Index uri
+
+The index-uri defines where the index files (also called splits) should be stored.
+This parameter expected a [storage uri](storage-uri).
```
evanxg852000, line 70:
```suggestion
This parameter expects a [storage uri](storage-uri).
```

evanxg852000 (approved):

fmassot (approved):
Well spotted, thanks!

PR #1080, opened 2022-01-11
Fix Rest API docs

PSeitz:
improve GET docs
Remove Format block with reference to POST/PUT, there are no POST/PUT
rename searchFields parameter to searchField

### DIFF
--- a/docs/reference/rest-api.md
+++ b/docs/reference/rest-api.md
@@ -7,14 +7,6 @@ position: 9
 
 All the API endpoints start with the `api/v1/` prefix. `v1` indicates that we are currently using version 1 of the API.
 
-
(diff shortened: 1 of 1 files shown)


docs/reference/rest-api.md
```diff
@@ -7,14 +7,6 @@ position: 9
 
 All the API endpoints start with the `api/v1/` prefix. `v1` indicates that we are currently using version 1 of the API.
 
-
-## Format
```
fmassot, line 11: indeed it was obsolete. It's still the case for the section `Error handling`. Can you update it? this  error handling is only for non stream request.

fmassot (approved):

PR #1081, opened 2022-01-11
fix startOffset

### LINKED ISSUE #1079: Bad error in rest api when retrieving results outside startOffset
```
curl -L "http://127.0.0.1:7280/api/v1/hdfs/search?query=cliid&startOffset=10000"
{
  "error": "Search error. Internal error: `Spawned task in root join failed: panic`.."
}⏎
```

### Expected behaviour
Normal result, but empty hits array

PSeitz:
fixes #1079

fix `drain` parameter


PSeitz:
> can you add a test on that?

Not quickly, there is no existing proper test I could build upon

quickwit-search/src/collector.rs
```diff
@@ -289,10 +289,14 @@ impl Collector for QuickwitCollector {
         // We compute the overall [0..start_offset + max_hits) documents ...
         let num_hits = self.start_offset + self.max_hits;
         let mut merged_leaf_response = merge_leaf_responses(segment_fruits, num_hits);
-        // ... and drop the first [..start_offets) hits.
+        // ... and drop the first [..start_offsets) hits.
         merged_leaf_response
             .partial_hits
-            .drain(0..self.start_offset)
+            .drain(
```
fulmicoton, line 295: TIL !
nice catch.

fulmicoton:
We need a unit test.

PR #1090, opened 2022-01-13
add test for startOffset out of bounds

### LINKED ISSUE #1085: Add unit test to test startOffset.
Search for `test_root_` for similar unit tests.

PSeitz:
closes #1085

### DIFF
--- a/quickwit-search/src/root.rs
+++ b/quickwit-search/src/root.rs
@@ -288,6 +288,90 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_root_search_offset_out_of_bounds_1085() -> anyhow::Result<()> {
+        let search_request = quickwit_proto::SearchRequest {
+            index_id: "test-idx".to_string(),
+            query: "test".to_string(),
+            search_fields: vec!["body".to_string()],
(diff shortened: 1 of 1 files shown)


quickwit-search/src/root.rs
```diff
@@ -287,6 +287,64 @@ mod tests {
...
+                    "file:///path/to/index/test-idx",
+                ))
+            });
+        metastore.expect_list_splits().returning(
+            |_index_id: &str, _split_state: SplitState, _time_range: Option<Range<i64>>, _tags| {
+                Ok(vec![mock_split("split1")])
+            },
+        );
+        let mut mock_search_service = MockSearchService::new();
+        mock_search_service.expect_leaf_search().returning(
```
fulmicoton, line 317: I think we'd better have two leaves for this test.

fulmicoton (changes requested):
add another leaf please

quickwit-search/src/root.rs
PSeitz: yep, updated for 2 node leaves

fulmicoton (approved):

PR #1106, opened 2022-01-24
ISSUE-1086 Check and validate the query against the last schema before dispatching

linxGnu:
This PR is first attempt to fix issue-1086 where query should be validated against current/last schema before dispatching to leaf node. It also tries to refactor/optimize code a bit to be (a bit) safer and faster.

What is not covered:
- Stream API (I think it's better to make it in another PR)

### DIFF
--- a/quickwit-search/src/root.rs
+++ b/quickwit-search/src/root.rs
@@ -153,9 +153,8 @@ pub async fn root_search(
 
     // Merge the fetched docs.
     let hits = fetch_docs_responses
-        .iter()
-        .map(|response| response.hits.clone())
-        .flatten()
+        .into_iter()
+        .flat_map(|response| response.hits)
         .sorted_by(|hit1, hit2| {
             let value1 = if let Some(partial_hit) = &hit1.partial_hit {
                 partial_hit.sorting_field_value
@@ -220,9 +219,8 @@ fn jobs_to_leaf_request(
         search_request: Some(request_with_offset_0),
         split_metadata: jobs
             .iter()
-            .map(|job| {
-                extract_split_and_footer_offsets(split_metadata_map.get(&job.split_id).unwrap())
-            })
+            .filter_map(|job| split_metadata_map.get(&job.split_id))
+            .map(extract_split_and_footer_offsets)
             .collect(),
         doc_mapper: doc_mapper_str.to_string(),
         index_uri: index_uri.to_string(),
@@ -238,12 +236,12 @@ fn jobs_to_fetch_docs_request(
 ) -> FetchDocsRequest {
     let partial_hits = jobs
         .iter()
-        .map(|job| partial_hits_map.remove(&job.split_id).unwrap())
+        .filter_map(|job| partial_hits_map.remove(&job.split_id))
         .flatten()
         .collect_vec();
     let splits_footer_and_offsets = jobs
         .iter()
-        .map(|job| split_metadata_map.get(&job.split_id).unwrap())
+        .filter_map(|job| split_metadata_map.get(&job.split_id))
         .map(extract_split_and_footer_offsets)
         .collect_vec();
 
--- a/quickwit-search/src/root.rs
+++ b/quickwit-search/src/root.rs
@@ -215,18 +215,13 @@ fn jobs_to_leaf_request(
     request_with_offset_0.start_offset = 0;
     request_with_offset_0.max_hits += request.start_offset;
 
-    // collect split_id and footer offsets
-    let split_metadata: Vec<_> = jobs
-        .iter()
-        .filter_map(|job| split_metadata_map.get(&job.split_id))
-        .map(extract_split_and_footer_offsets)
-        .collect();
-    // we expect job's metadata existence (thus metadata_map does not miss any split_id)
-    debug_assert_eq!(jobs.len(), split_metadata.len());
-
     LeafSearchRequest {
         search_request: Some(request_with_offset_0),
-        split_metadata,
+        split_metadata: jobs
+            .iter()
+            .map(|job| split_metadata_map.get(&job.split_id).expect(&job.split_id))
+            .map(extract_split_and_footer_offsets)
+            .collect(),
         doc_mapper: doc_mapper_str.to_string(),
         index_uri: index_uri.to_string(),
     }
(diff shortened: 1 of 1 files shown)


quickwit-search/src/root.rs
```diff
@@ -149,9 +153,8 @@ pub async fn root_search(
 
     // Merge the fetched docs.
     let hits = fetch_docs_responses
-        .iter()
-        .map(|response| response.hits.clone())
-        .flatten()
+        .into_iter()
+        .flat_map(|response| response.hits)
```
linxGnu, line 157: Note: no need to clone the hits to improve performance

quickwit-search/src/root.rs
```diff
@@ -216,9 +219,8 @@ fn jobs_to_leaf_request(
         search_request: Some(request_with_offset_0),
         split_metadata: jobs
             .iter()
-            .map(|job| {
-                extract_split_and_footer_offsets(split_metadata_map.get(&job.split_id).unwrap())
-            })
+            .filter_map(|job| split_metadata_map.get(&job.split_id))
```
linxGnu, line 222: Note: unwrap is unsafe. It's better to check with filter*

fulmicoton (approved):

linxGnu:
@fulmicoton thank you so much for reviewing.

Yeah, I also notice about streaming API is not covered by this PR (also mention it in the description)

> What is not covered:
> Stream API (I think it's better to make it in another PR)

I would like to make it in another PR. HDYT

quickwit-search/src/root.rs
```diff
@@ -65,13 +65,17 @@ pub async fn root_search(
...
+    let _query = doc_mapper.query(doc_mapper.schema(), search_request)?;
+
     let doc_mapper_str = serde_json::to_string(&doc_mapper).map_err(|err| {
         SearchError::InternalError(format!("Failed to serialize doc mapper: Cause {}", err))
     })?;
     let split_metadata_list = list_relevant_splits(search_request, metastore).await?;
     let split_metadata_map: HashMap<String, SplitMetadata> = split_metadata_list
         .into_iter()
-        .map(|metadata| (metadata.split_id().to_string(), metadata))
+        .map(|mut metadata| (std::mem::take(&mut metadata.split_id), metadata))
```
linxGnu, line 78: Note: no need to copy here

quickwit-search/src/root.rs
fulmicoton: I might surprise you but I think I prefer the unwrap here  :).
But you are right, maybe something else would be even better.

So by unsafe I think you meant it can panic. (unsafe usually means something else).
This is right. If one of the key in that map is missing, then we get a panic... which is better 
than silently returning search results, without spotting the bug.

So the role of that unwrap is exactly the same as an `assert`.
It is a bit lame though because it is impossible for the reader to know if it is a "valid" unwrap or a  "sloppy" unwrap.
(The assert is about an invariant entirely enforced in the root so it is justified.)

Maybe we could replace it by an expect with a message explaining the situation, 
or use filter_map and add an explicit assert on the length of split_metadata.

```
let split_metadata = jobs.iter().filter_map(|job| {} /*...*/
assert_eq!(split_metadata.len(), jobs.len());
```

quickwit-search/src/root.rs
```diff
@@ -234,12 +236,12 @@ fn jobs_to_fetch_docs_request(
 ) -> FetchDocsRequest {
     let partial_hits = jobs
         .iter()
-        .map(|job| partial_hits_map.remove(&job.split_id).unwrap())
```
fulmicoton, line 237: same comment as above.

quickwit-search/src/root.rs
fulmicoton: Let's not do this. This probably has no impact at all on performance and it is harmful:
- the code is less simple and less readable
- you are modifying an object that was immutable.

Consumer of the metadata will  assume that it is properly populated and find an empty string.
Note that this value can be very handy for debugging for instance.

quickwit-search/src/root.rs
linxGnu: I love the idea of having `assert` here for 2 reasons:
- Resilient to user when no need for debugging
- Easier to troubleshoot when turning on debug-mode
Let me update the code

quickwit-search/src/root.rs
linxGnu: Yeah you are right. Let me revert this :heart:

linxGnu:
@fulmicoton I addressed your comments. Thank you so much for great catching

quickwit-search/src/root.rs
```diff
@@ -212,14 +215,18 @@ fn jobs_to_leaf_request(
...
     request_with_offset_0.max_hits += request.start_offset;
 
+    // collect split_id and footer offsets
+    let split_metadata: Vec<_> = jobs
+        .iter()
+        .filter_map(|job| split_metadata_map.get(&job.split_id))
+        .map(extract_split_and_footer_offsets)
+        .collect();
+    // we expect job's metadata existence (thus metadata_map does not miss any split_id)
+    debug_assert_eq!(jobs.len(), split_metadata.len());
```
fulmicoton, line 225:
```suggestion
    assert_eq!(jobs.len(), split_metadata.len());
```

quickwit-search/src/root.rs
```diff
@@ -232,16 +239,21 @@ fn jobs_to_fetch_docs_request(
...
     jobs: &[Job],
 ) -> FetchDocsRequest {
+    let partial_hits_map_size = partial_hits_map.len();
     let partial_hits = jobs
         .iter()
-        .map(|job| partial_hits_map.remove(&job.split_id).unwrap())
+        .filter_map(|job| partial_hits_map.remove(&job.split_id))
         .flatten()
         .collect_vec();
+    debug_assert_eq!(partial_hits_map_size - partial_hits_map.len(), jobs.len());
```
fulmicoton, line 248: Why not :
```suggestion
    assert_eq!(partial_hits.len(), jobs.len());
```

quickwit-search/src/root.rs
```diff
@@ -232,16 +239,21 @@ fn jobs_to_fetch_docs_request(
...
+
     let splits_footer_and_offsets = jobs
         .iter()
-        .map(|job| split_metadata_map.get(&job.split_id).unwrap())
+        .filter_map(|job| split_metadata_map.get(&job.split_id))
         .map(extract_split_and_footer_offsets)
         .collect_vec();
+    // we expect job's metadata existence (thus metadata_map does not miss any split_id)
+    debug_assert_eq!(jobs.len(), splits_footer_and_offsets.len());
```
fulmicoton, line 256:
```suggestion
    assert_eq!(jobs.len(), splits_footer_and_offsets.len());
```

fulmicoton:
> I have no strong point of resilient or not. Just IMHO, I prefer of not making user confuse due to runtime conditions. It's better to log it out and continue on.
> So if you make the decision of keeping assert/panic on missing split_id, then I would like to update the PR again instead of using debug_assert*
>We could even return error in this case thinking

So we have a bunch of actions we can take right?
log, return an error, assert, debug_assert.

Why do we not return an error on that?
Well the key part is that this is an invariant of the code we are checking.
It is supposed to never trigger... But we want things to catch fire if this invariant is broken.
This assert breaking is, philosophically, similar to wanting addition overflow to panic
vs wrapping around.

One important part is that this invariant is enforced locally. For instance, it is not enforced "assuming the request parameters"
are correct. If it was the latter, we would have to return an error.

quickwit-search/src/root.rs
linxGnu: Because `partial_hits_map` value is not flatten. Partial hits might have different size from jobs. The correct one is checking number of removed key.

linxGnu:
@fulmicoton I see what you aims for. And I think that using `unwrap` is better here. 

In this way, how about adding more context:
```rust
partial_hits_map.remove(&job.split_id).expect(&job.split_id)
```

In this way, panic is caught with missing split_id, user could have more choice to handle metadata.

fulmicoton:
@linxGnu actually the one call for fetch_docs is problematic... The split_id is coming from a leaf response, so the invariatn is coming from the external world.

linxGnu:
Thank you for pointing out. Could you please take a look again @fulmicoton

fulmicoton:
@linxGnu  FYI I refactored away all the unwrap but 1, and encapsulated the logic
justifying the last one in a function.
https://github.com/quickwit-inc/quickwit/pull/1108

linxGnu:
Thank you very much

PR #1117, opened 2022-01-27
First stab at quickwit UI.

fmassot:


fulmicoton (approved):
what a beast O_o
I could go in depth but I went through everything rapidly and did not find anything unexpected, and had no trouble navigating through the code. Awesome job.

I think we can merge this already. What do you think Francois?

quickwit-ui/app/src/utils/models.ts
```diff
@@ -0,0 +1,215 @@
...
+  numHits: 100,
+}
+
+export type SearchResponse = {
+  count: number;
+  hits: Array<RawDoc>;
+  numMicrosecs: number;
+}
+
+export type IndexMetadata = {
```
PSeitz, line 137: we should consider generating these types

https://github.com/Aleph-Alpha/ts-rs

quickwit-ui/app/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
+    ],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
+    "strict": true,
```
PSeitz, line 13: strict does not cover all strict options on newer versions automatically

noPropertyAccessFromIndexSignature
noUncheckedIndexedAccess
noUnusedLocals
noUnusedParameters

quickwit-ui/app/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "typescript": "^4.4.2",
+    "web-vitals": "^2.1.0"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject"
+  },
+  "eslintConfig": {
```
PSeitz, line 33: config, but no eslint in the dependencies?

quickwit-ui/app/src/utils/models.ts
fmassot: oh nice! Will definitely give it a try. I will put that in a separate issue.

quickwit-ui/app/tsconfig.json
fmassot: Will take a look on that.

fmassot:
> I think we can merge this already. What do you think Francois?

Let me make a final pass on it and we could do it. By the way, I want to remove the subfolder app as the rust code related to the REST endpoints will be in the quickwit-server crate.

quickwit-ui/app/tsconfig.json
fmassot: I set everything to true

quickwit-ui/app/package.json
fmassot: no ideas, I'm so lazy with the js stack...

PSeitz: we should add an eslinter

PR #1121, opened 2022-01-31
ISSUE-1118 Check collector logic at root and optimize

linxGnu:
This PR also makes some small optimizations

### DIFF
--- a/quickwit-search/src/collector.rs
+++ b/quickwit-search/src/collector.rs
@@ -284,6 +284,12 @@ impl Collector for QuickwitCollector {
         &self,
         segment_fruits: Vec<LeafSearchResponse>,
     ) -> tantivy::Result<Self::Fruit> {
+        if segment_fruits.is_empty() {
+            return Ok(LeafSearchResponse {
+                ..Default::default()
+            });
+        }
+
         // We want the hits in [start_offset..start_offset + max_hits).
         // All leaves will return their top [0..max_hits) documents.
         // We compute the overall [0..start_offset + max_hits) documents ...
@@ -304,7 +310,7 @@ impl Collector for QuickwitCollector {
 
 /// Merges a set of Leaf Results.
 fn merge_leaf_responses(
-    leaf_responses: Vec<LeafSearchResponse>,
+    mut leaf_responses: Vec<LeafSearchResponse>,
     max_hits: usize,
 ) -> LeafSearchResponse {
     // Optimization: No merging needed if there is only one result.
@@ -320,13 +326,13 @@ fn merge_leaf_responses(
         .map(|leaf_response| leaf_response.num_hits)
         .sum();
     let failed_splits = leaf_responses
-        .iter()
-        .flat_map(|leaf_response| leaf_response.failed_splits.iter().cloned())
+        .iter_mut()
+        .flat_map(|leaf_response| std::mem::take(&mut leaf_response.failed_splits))
         .collect_vec();
-    let all_partial_hits: Vec<PartialHit> = leaf_responses
+    let all_partial_hits = leaf_responses
         .into_iter()
         .flat_map(|leaf_response| leaf_response.partial_hits)
-        .collect();
+        .collect_vec();
     // TODO optimize
     let top_k_partial_hits = top_k_partial_hits(all_partial_hits, max_hits);
     LeafSearchResponse {
--- a/quickwit-search/src/collector.rs
+++ b/quickwit-search/src/collector.rs
@@ -324,10 +324,10 @@ fn merge_leaf_responses(
         .flat_map(|leaf_response| leaf_response.failed_splits.iter())
         .cloned()
         .collect_vec();
-    let all_partial_hits = leaf_responses
+    let all_partial_hits: Vec<PartialHit> = leaf_responses
         .into_iter()
         .flat_map(|leaf_response| leaf_response.partial_hits)
-        .collect_vec();
+        .collect();
     // TODO optimize
     let top_k_partial_hits = top_k_partial_hits(all_partial_hits, max_hits);
     LeafSearchResponse {
--- a/quickwit-search/src/collector.rs
+++ b/quickwit-search/src/collector.rs
@@ -342,7 +342,7 @@ fn merge_leaf_responses(
 ///
 /// TODO we could possibly optimize the sort away (but I doubt it matters).
 fn top_k_partial_hits(mut partial_hits: Vec<PartialHit>, num_hits: usize) -> Vec<PartialHit> {
-    partial_hits.sort_by(|left, right| {
+    partial_hits.sort_unstable_by(|left, right| {
         let left_key = partial_hit_sorting_key(left);
         let right_key = partial_hit_sorting_key(right);
         left_key.cmp(&right_key)
--- a/quickwit-search/src/collector.rs
+++ b/quickwit-search/src/collector.rs
@@ -284,12 +284,6 @@ impl Collector for QuickwitCollector {
         &self,
         segment_fruits: Vec<LeafSearchResponse>,
     ) -> tantivy::Result<Self::Fruit> {
-        if segment_fruits.is_empty() {
-            return Ok(LeafSearchResponse {
-                ..Default::default()
-            });
-        }
-
         // We want the hits in [start_offset..start_offset + max_hits).
         // All leaves will return their top [0..max_hits) documents.
         // We compute the overall [0..start_offset + max_hits) documents ...
@@ -310,7 +304,7 @@ impl Collector for QuickwitCollector {
 
 /// Merges a set of Leaf Results.
 fn merge_leaf_responses(
-    mut leaf_responses: Vec<LeafSearchResponse>,
+    leaf_responses: Vec<LeafSearchResponse>,
     max_hits: usize,
 ) -> LeafSearchResponse {
     // Optimization: No merging needed if there is only one result.
@@ -326,8 +320,9 @@ fn merge_leaf_responses(
         .map(|leaf_response| leaf_response.num_hits)
         .sum();
     let failed_splits = leaf_responses
-        .iter_mut()
-        .flat_map(|leaf_response| std::mem::take(&mut leaf_response.failed_splits))
+        .iter()
+        .flat_map(|leaf_response| leaf_response.failed_splits.iter())
+        .cloned()
         .collect_vec();
     let all_partial_hits = leaf_responses
         .into_iter()
--- a/quickwit-search/src/leaf.rs
+++ b/quickwit-search/src/leaf.rs
@@ -267,10 +267,10 @@ pub async fn leaf_search(
                 Err(err) => Either::Right(err),
             });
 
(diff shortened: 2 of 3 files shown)


quickwit-search/src/collector.rs
```diff
@@ -320,13 +326,13 @@ fn merge_leaf_responses(
         .map(|leaf_response| leaf_response.num_hits)
         .sum();
     let failed_splits = leaf_responses
-        .iter()
-        .flat_map(|leaf_response| leaf_response.failed_splits.iter().cloned())
+        .iter_mut()
+        .flat_map(|leaf_response| std::mem::take(&mut leaf_response.failed_splits))
```
linxGnu, line 330: No need to cloned each `failed_splits` reported

quickwit-search/src/collector.rs
```diff
@@ -342,7 +348,7 @@ fn merge_leaf_responses(
 ///
 /// TODO we could possibly optimize the sort away (but I doubt it matters).
 fn top_k_partial_hits(mut partial_hits: Vec<PartialHit>, num_hits: usize) -> Vec<PartialHit> {
-    partial_hits.sort_by(|left, right| {
+    partial_hits.sort_unstable_by(|left, right| {
```
linxGnu, line 351: I wonder if we need `stable` sort here

quickwit-search/src/leaf.rs
```diff
@@ -267,12 +266,17 @@ pub async fn leaf_search(
...
-            .context("Failed to merge split search responses.")??;
+
+    let mut merged_search_response = {
+        let span = info_span!("merge_search_responses");
+        let _enter = span.enter();
+
+        merge_collector
+            .merge_fruits(split_search_responses)
+            .context("Failed to merge split search responses.")?
+    };
```
linxGnu, line 279: Spawning task into tokio runtime in this case might introduce a little overhead for scheduler. In this way, it might be better to run merging task right away.

quickwit-search/src/root.rs
```diff
@@ -167,13 +166,14 @@ pub async fn root_search(
...
-        spawn_blocking(move || merge_collector.merge_fruits(leaf_search_responses))
-            .await?
-            .map_err(|merge_error: TantivyError| {
-                crate::SearchError::InternalError(format!("{}", merge_error))
-            })?;
+    let leaf_search_response = merge_collector
+        .merge_fruits(leaf_search_responses)
+        .map_err(|merge_error: TantivyError| {
+            crate::SearchError::InternalError(format!("{}", merge_error))
+        })?;
```
linxGnu, line 176: ditto

quickwit-search/src/root.rs
```diff
@@ -222,7 +222,7 @@ pub async fn root_search(
         .into_iter()
         .flat_map(|response| response.hits)
         .collect();
-    hits.sort_by_key(|hit| {
+    hits.sort_unstable_by_key(|hit| {
```
linxGnu, line 225: Do we need `stable` sort here?

quickwit-search/src/leaf.rs
fulmicoton: We avoid running any blocking operation that are not trivially taking less than 500 microseconds.

quickwit-search/src/collector.rs
fulmicoton: unstable is ok because partia_hit_sorting_key is a total order.

quickwit-search/src/collector.rs
```diff
@@ -284,6 +284,12 @@ impl Collector for QuickwitCollector {
         &self,
         segment_fruits: Vec<LeafSearchResponse>,
     ) -> tantivy::Result<Self::Fruit> {
+        if segment_fruits.is_empty() {
```
fulmicoton, line 287: These extra lines that are not helping.

quickwit-search/src/collector.rs
fulmicoton: You are correct, but the previous code is simpler to read.

The clone you are trying to avoid is tiny and only happens if we have failed splits. This is not something we need to optimize. "Premature optimization is the root of all evil."

linxGnu: You are right. Let me revert and make it a bit easier to read

quickwit-search/src/leaf.rs
linxGnu: Previous code looks like:
```
 let mut merged_search_response =
        spawn_blocking(move || merge_collector.merge_fruits(split_search_responses))
            .instrument(info_span!("merge_search_responses"))
            .await
            .context("Failed to merge split search responses.")??;
```
spawn task into runtime, let other/same thread run it. Current task is awaited for spawned task to be done. 

Hmm, I do not see the benefit here.

quickwit-search/src/collector.rs
linxGnu: ah yeah, you are right. Following up logic are safe even if input is empty

quickwit-search/src/leaf.rs
fulmicoton: spawn_blocking is different from spawn. It spawns a tasks a different thread pool than the tokio runtime thread pool.

The tokio runtime is meant to run a lot of IO operations and assumes cooperative tasks.
When we block it for say 1 second, io intensive operations will be blocked for no reason at all.

quickwit-search/src/root.rs
fulmicoton: no unstable is fine.

quickwit-search/src/leaf.rs
linxGnu: Oh if the metrics show that merging is cpu-bound tasks and often takes long, I think you are right to spawn it into Tokio's blocking threads.

Will revert the changes. I just read the code and does not know the reason why doing this. Let me add comment also, in this way, other developer does not need to ask like me :heart:

quickwit-search/src/leaf.rs
```diff
@@ -267,7 +267,11 @@ pub async fn leaf_search(
                 Err(err) => Either::Right(err),
             });
 
+    // create a collector which merges responses into one
     let merge_collector = make_merge_collector(request);
+
+    // merging might be heavy cpu-bound task and take time.
```
fulmicoton, line 273:
```suggestion
    // Merging might be heavy cpu-bound task and take time.
```

quickwit-search/src/leaf.rs
```diff
@@ -267,7 +267,11 @@ pub async fn leaf_search(
                 Err(err) => Either::Right(err),
             });
 
+    // create a collector which merges responses into one
```
fulmicoton, line 270:
```suggestion
    // Create a collector which merges responses into one
```

quickwit-search/src/root.rs
```diff
@@ -167,7 +167,12 @@ pub async fn root_search(
             }),
     )
     .await?;
+
+    // create a collector which merges responses into one
```
fulmicoton, line 171:
```suggestion
    // Create a collector which merges responses into one
```

quickwit-search/src/root.rs
```diff
@@ -167,7 +167,12 @@ pub async fn root_search(
...
     let merge_collector = make_merge_collector(search_request);
+
+    // merging might be heavy cpu-bound task and take time.
```
fulmicoton, line 174:
```suggestion
    // Merging might be heavy cpu-bound task and take time.
```

quickwit-search/src/collector.rs
```diff
@@ -321,12 +321,13 @@ fn merge_leaf_responses(
         .sum();
     let failed_splits = leaf_responses
         .iter()
-        .flat_map(|leaf_response| leaf_response.failed_splits.iter().cloned())
+        .flat_map(|leaf_response| leaf_response.failed_splits.iter())
+        .cloned()
         .collect_vec();
-    let all_partial_hits: Vec<PartialHit> = leaf_responses
+    let all_partial_hits = leaf_responses
```
fulmicoton, line 327:
```suggestion
    let all_partial_hits: Vec<PartialHit> = leaf_responses
```

quickwit-search/src/collector.rs
```diff
@@ -321,12 +321,13 @@ fn merge_leaf_responses(
...
         .into_iter()
         .flat_map(|leaf_response| leaf_response.partial_hits)
-        .collect();
+        .collect_vec();
```
fulmicoton, line 330:
```suggestion
        .collect();
```

fulmicoton: The `collect -> collect_vec()` change does not bring anything positive.

Writing the type explicitly from time to time in the code helps people reader from github for instance.

There is a reason why editor add an overlay with types.

quickwit-search/src/leaf.rs
fulmicoton:
```suggestion
    // Creates a collector which merges responses into one
```

quickwit-search/src/leaf.rs
fulmicoton:
```suggestion
    // Merging is a cpu-bound task.
```

quickwit-search/src/leaf.rs
```diff
@@ -267,7 +267,11 @@ pub async fn leaf_search(
                 Err(err) => Either::Right(err),
             });
 
+    // Create a collector which merges responses into one
```
fulmicoton, line 270:
```suggestion
    // Creates a collector which merges responses into one
```

quickwit-search/src/leaf.rs
```diff
@@ -267,7 +267,11 @@ pub async fn leaf_search(
...
     let merge_collector = make_merge_collector(request);
+
+    // Merging might be heavy cpu-bound task and take time.
```
fulmicoton, line 273: Merging is cpu-bound.

fulmicoton, line 273:
```suggestion
    // Merging is a cpu-bound task.
```

quickwit-search/src/collector.rs
```diff
@@ -321,7 +321,8 @@ fn merge_leaf_responses(
         .sum();
     let failed_splits = leaf_responses
         .iter()
-        .flat_map(|leaf_response| leaf_response.failed_splits.iter().cloned())
+        .flat_map(|leaf_response| leaf_response.failed_splits.iter())
```
fulmicoton, line 324: It really does not matter does it?

quickwit-search/src/leaf.rs
fulmicoton: Sure. It is a very common pitfall when using tokio. I think we have a point about it in our code style.
https://github.com/quickwit-oss/quickwit/blob/main/CODE_STYLE.md

fulmicoton (approved):

quickwit-search/src/collector.rs
linxGnu: Yes, it just make code easier to extend in the future in case: in the future, we want to perform post-action right after flat_map

fulmicoton (approved):

PR #1124, opened 2022-02-03
ISSUE-1112 Improve leaf node assignment algorithm

linxGnu:
Implementation of balancing algorithm suggested by Francois:
- Compute the overall cost.
- Compute a target cost per node. It could be something like ceil(overall_cost / number of node) * (1 + margin).
- For each split, sorted in decreasing order, assign the split with the highest affinity.
- Remove node from list as its load exceed the target cost

### DIFF
--- a/quickwit-search/src/search_client_pool.rs
+++ b/quickwit-search/src/search_client_pool.rs
@@ -251,7 +251,6 @@ impl SearchClientPool {
 
         for job in jobs {
             sort_by_rendez_vous_hash(&mut nodes, job.split_id());
-
             // choose one of the the first two nodes based on least loaded
             let chosen_node_index: usize = if nodes.len() >= 2 {
                 if nodes[0].load > nodes[1].load {
--- a/quickwit-search/src/search_client_pool.rs
+++ b/quickwit-search/src/search_client_pool.rs
@@ -33,6 +33,10 @@ use tracing::*;
 use crate::rendezvous_hasher::sort_by_rendez_vous_hash;
 use crate::{swim_addr_to_grpc_addr, Cost, SearchServiceClient};
 
+// allow load of a node to exceed 25% threshold
+const LOAD_MARGIN: f64 = 0.25;
+const MIN_LOAD_THRESHOLD: u64 = 16;
+
 /// Create a SearchServiceClient with SocketAddr as an argument.
 /// It will try to reconnect to the node automatically.
 async fn create_search_service_client(
@@ -73,7 +77,7 @@ pub struct SearchClientPool {
     /// Search clients.
     /// A hash map with gRPC's SocketAddr as the key and SearchServiceClient as the value.
     /// It is not the cluster listen address.
-    clients: Arc<RwLock<HashMap<SocketAddr, SearchServiceClient>>>, /* TODO: it's possible to
+    clients: Arc<RwLock<HashMap<SocketAddr, SearchServiceClient>>>, /* TODO possible to
                                                                      * use arc_swap here instead
                                                                      * of RWLock */
 }
@@ -249,8 +253,15 @@ impl SearchClientPool {
             job_order_key(left).cmp(&job_order_key(right))
         });
 
+        // Compute load (per node) threshold
+        let total_cost: Cost = jobs.iter().map(|v| v.cost()).sum();
+        let load_per_node_threshold =
+            ((total_cost as f64 / nodes.len() as f64) * (1.0 + LOAD_MARGIN)).ceil() as u64;
+        let load_per_node_threshold = load_per_node_threshold.max(MIN_LOAD_THRESHOLD);
+
         for job in jobs {
             sort_by_rendez_vous_hash(&mut nodes, job.split_id());
+
             // choose one of the the first two nodes based on least loaded
             let chosen_node_index: usize = if nodes.len() >= 2 {
                 if nodes[0].load > nodes[1].load {
@@ -264,6 +275,18 @@ impl SearchClientPool {
 
             // update node load for next round
             nodes[chosen_node_index].load += job.cost() as u64;
+
+            // check if node's load exceeds the threshold -> remove node from list
+            //
+            // note: computing load per node might result in saturation (in case cost data is
+            // malform). In this way, all nodes might be removed from list. Thus, having condition
+            // len > 1 to ensure that there is (at least) one node could serve remaining jobs
+            if nodes.len() > 1 && nodes[chosen_node_index].load > load_per_node_threshold {
+                // Removes an element from the vector and returns it.
+                // The removed element is replaced by the last element of the vector.
+                // O(1) complexity
+                nodes.swap_remove(chosen_node_index);
+            }
 
             let chosen_leaf_grpc_addr: SocketAddr = nodes[chosen_node_index].peer_grpc_addr;
             splits_groups
--- a/quickwit-search/src/search_client_pool.rs
+++ b/quickwit-search/src/search_client_pool.rs
@@ -247,8 +247,10 @@ impl SearchClientPool {
             }
         }
 
-        // Sort job
-        jobs.sort_by(|left, right| {
+        // Sort job by cost desc, split_id asc.
+        // by keeping split_id in comparision, we minimize cost for next rendez_vous sort when
+        // two consecutive split_id(s) are the same.
+        jobs.sort_unstable_by(|left, right| {
             // sort_by_key does not work here unfortunately
             job_order_key(left).cmp(&job_order_key(right))
         });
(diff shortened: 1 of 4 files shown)


quickwit-search/src/search_client_pool.rs
```diff
@@ -247,8 +253,15 @@ impl SearchClientPool {
...
+        let total_cost: Cost = jobs.iter().map(|v| v.cost()).sum();
+        let load_per_node_threshold =
+            ((total_cost as f64 / nodes.len() as f64) * (1.0 + LOAD_MARGIN)).ceil() as u64;
+        let load_per_node_threshold = load_per_node_threshold.max(MIN_LOAD_THRESHOLD);
+
         for job in jobs {
             sort_by_rendez_vous_hash(&mut nodes, job.split_id());
+
             // choose one of the the first two nodes based on least loaded
             let chosen_node_index: usize = if nodes.len() >= 2 {
```
fulmicoton, line 266: I think the whole point is to ditch this logic here.

The new algorithm will increase the chance of picking the node with the highest affinity.

linxGnu: Hi, I would like to discuss a bit about the strategy here.

IMHO, there is no perfect load-balancer algorithm. I really think the idea of Francois by introducing cost threshold works, in terms of performance and simplicity.

Let me summarize a bit:
- Job is sorted desc by cost (same as previous strategy)
- The difference between old-new algorithm is applying a threshold = total_cost / num_node
- If any node reaches the threshold -> remove from selection

There are many cases of imbalance just like knapsack problem. So, I think we should focus on 2 popular cases:
- Cost is similar between split_id
- There are several super split_id(s) that each of them contains lot of documents, cost is high

In the first case, the algorithm works, just like old one. 
In the second case, if there are super split_id(s), most likely that there is only one node would pick 1 or 2 split_id, then reaches threshold, then be removed from next selection.

Additionally, new algorithm still respects `rendez_vous_hash` order for cache locality at leaf-node (IMHO, using `rendez_vous_hash` order is pretty good approach :+1: )

fulmicoton: I don't think you want to pick the node that is the least loaded out of the first two. This is useless. 

Thanks to Francois's solution you can pick the best one.

fulmicoton: In other words, `chosen_node_index = 0` all of the time.

linxGnu: Hmm, I think code looks like:

```rust
// choose one of the the first two nodes based on least loaded
let chosen_node_index: usize = if nodes.len() >= 2 {
            if nodes[0].load > nodes[1].load {
                    1
                } else {
                    0
                }
            } else {
                0
            };
```
For each job -> node is sorted by `rendez_vous_hash`, not load. So the first two nodes (after sorted) does not mean lowest load.

fulmicoton: Why not simply take `nodes[0]`?

linxGnu: I think take best one of two is good heuristic (take lower load from two). 

That's why I keep it. It's also harmless IMHO

linxGnu:
Currently, job is ordered by cost desc, then split_id.

How about: just ordered by cost desc, no need to involve split_id. hdyt @fulmicoton

fulmicoton:
I don't think it is harmless.

Can you think of a metric to measure if an algorithm is good or not?

linxGnu:
Excuse me, you mean the whole algorithm or the part `best of 2` @fulmicoton 

And yeah, to measure an heuristic algorithm is a hard question I think. There is edge case which makes the load pretty good or pretty bad.

quickwit-search/src/search_client_pool.rs
fulmicoton: I don't think this is harmless at all.

fulmicoton:
The best of 2 part is useless. It is hurting the algorithm performance.

Now for how to measure the performance of the algorithm. Well we want 2 properties:
- (P1) We want to assign splits to the server with the highest affinity 
- (P2) We want to make sure that the work is balanced between the servers

It is possible to build a cost function combining the two, or it is possible to express P2 as a constraint, and try to optimize a cost function expressing P1.

The best of 2 part is hurting P1 greatly, and is not necessary anymore.
P2 is already well covered by removing nodes that exceed their budget.

linxGnu:
Oh, I don't know: `The best of 2 part is hurting P1 greatly, and is not necessary anymore`. It will be great help if we could provide some metric like: cpu usage, trace, response latency.

Why I think best of 2 is fine for both case: normally, there might be 1-3 nodes will receive same `split_id` job due to old/new algorithm. 

Anw, I think you have full metrics on a lot of datasets. Let me update the code :+1:

fulmicoton:
The best of 2 thing will hit the node with the best affinity with a probability of 50% .
Francois algorithm without the best of 2, will hit the node with the best affinity with something much higher.
It depends on the query, but probably consistently above 90%.

linxGnu:
Yeah, I know the distribution probability when use best of 2. Just don't know how the performance is impacted when having 2 diff node execute same `split_id` job.

Thank you for figuring it out. I addressed your comment. PTAL

fulmicoton:
@linxGnu The idea is to that raising this probability increase our cache hit rate.
For instance we cache a so-called hotcache. If we assume we have just the capacity to store num split/ num servers, we will end up with a cache hit rate about 90% without the best of 2, and around 50% with the best of 2.

fulmicoton (approved):

quickwit-search/src/search_client_pool.rs
```diff
@@ -241,33 +247,44 @@ impl SearchClientPool {
...
+        jobs.sort_unstable_by(|left, right| {
             // sort_by_key does not work here unfortunately
             job_order_key(left).cmp(&job_order_key(right))
         });
 
+        // Compute load (per node) threshold
+        let total_cost: Cost = jobs.iter().map(|v| v.cost()).sum();
+        let load_per_node_threshold =
+            ((total_cost as f64 / nodes.len() as f64) * (1.0 + LOAD_MARGIN)).ceil() as u64;
+        let load_per_node_threshold = load_per_node_threshold.max(MIN_LOAD_THRESHOLD);
```
fulmicoton, line 263: So we have an edge case here if we have a lot of nodes and 1 large split.

The `MIN_LOAD_THRESHOLD` seems to be helping here but it does not seem very good at that.
I suggest we change the MIN_LOAD_THRESHOLD to the largest cost within the splits.

quickwit-search/src/search_client_pool.rs
```diff
@@ -241,33 +247,44 @@ impl SearchClientPool {
...
-                } else {
-                    0
-                }
-            } else {
-                0
-            };
 
             // update node load for next round
-            nodes[chosen_node_index].load += job.cost() as u64;
+            nodes[0].load += job.cost() as u64;
```
fulmicoton, line 269: The resulting load may be much higher than the load_per_node threshold.

quickwit-search/src/search_client_pool.rs
fulmicoton: I wonder if we can have a more solid math background here.

For instance if we define,

`load_per_node_threshold = ceil([total_cost / nodes] + MAX_NODE_COST)`

and we never add a split to a node if it would result to a load exceeding load_per_node_threshold, I think we have some interesting property:

Without extra checks, the algorithm would be able to assign all split to nodes.
Proof:

Let's say that we are left with a split that is unassigned.
It means that all nodes are loaded with at  `load > load_per_node_threshold - missing_node`.
(Otherwise we could allocate the missing node).

In other word, the overall load in the nodes would be greater than 
```
    (load_per_node_threshold - missing_node_load) * num_nodes
> total_cost + num_nodes * (max_nodes  - missing_node_load)
> total_cost
```
which is impossible...

We can even probably relax that condition, since we sorted the node per cost.
For instance, I suggest the following formula

```
base_load_per_node_threshold = (total_cost as f64) / nodes.len() as f64;
optimality_margin = base_load_per_node_threshold * 0.2
feasability_margin = splits[min(num_nodes, splits.len - 1)].cost # 

load_per_node_threshold
    =  max(
             max_node_cost
             base_load_per_node_threshold + max(optimality_margin, feasability_margin)
    )
     
```

fulmicoton: @fmassot @linxGnu  what do you think?

quickwit-search/src/search_client_pool.rs
```diff
@@ -241,33 +247,44 @@ impl SearchClientPool {
...
                 .entry(chosen_leaf_grpc_addr)
                 .or_insert_with(Vec::new)
                 .push(job);
+
+            // check if node's load exceeds the threshold -> remove node from list
+            //
+            // note: computing load per node might result in saturation (in case cost data is
+            // malform). In this way, all nodes might be removed from list. Thus, having condition
+            // len > 1 to ensure that there is (at least) one node could serve remaining jobs
+            if nodes.len() > 1 && nodes[0].load > load_per_node_threshold {
```
fulmicoton, line 282: With the formula above, I think this condition can be removed.

quickwit-search/src/search_client_pool.rs
fulmicoton: with the formula I suggest above, I think we need to add a check on the load not exceeding the threshold after the assignment here.

fulmicoton:
@fmassot I'd like your opinion, see comment inline.

fulmicoton:
@fmassot I'd like your opinion, see comment inline.
@linxGnu I think we need a bunch of unit test for important edge case (I think your implem cover them correctly already).

In particular, I would like to have a unit test for 1000 nodes, 1 split. (the point is that one node need to get the split, even though its cost is much large than total_load / num_nodes)

quickwit-search/src/search_client_pool.rs
linxGnu: @fulmicoton I think your suggestion is good in case of having several split with "super" cost. This would make imbalance cost between node.

There is edge case for the formula above:
```
costs: 5000, 2500, 100, 100, 100, 100, 100, 100
total_cost: 8100
num_nodes: 4
base: 8100 / 4 = 2025

-> load_per_node_threshold
    =  max(
             max_node_cost
             base_load_per_node_threshold + max(optimality_margin, feasability_margin)
    )
    = max(5000, 2025 + max(2025 * 0.2, 100)) = 5000

Distribution:
node[0].cost = 5000
node[1].cost = 2500 + 100 + 100 = 2700
node[2].cost = 100 + 100 + 100 + 100 = 400
```

While original algorithm by fmassot (looks better in my point of view):
```
load_per_node = 2025 + 2025 * 0.2 = 2430

node[0].cost = 5000
node[1].cost = 2500
node[2].cost = 600
```

Actually, all are heuristic and have edge cases imho. How about modify your formula a bit:
```
load_per_node_threshold
    =  base_load_per_node_threshold + max(optimality_margin, feasability_margin)
```
This one removes the impact of super cost (max_node_cost, like above, it's 5000).

Issue #1116, opened 2022-01-27
The curl install throws an error


larksoftware:
**Describe the bug**
I tried to use the curl install but it just throws an error:
```
$ curl -L https://install.quickwit.io | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   134  100   134    0     0    375      0 --:--:-- --:--:-- --:--:--   374
sh: 199: Syntax error: "(" unexpected (expecting "fi")
 87 15202   87 13253    0     0  23965      0 --:--:-- --:--:-- --:--:-- 23965
curl: (23) Failure writing output to destination
```

**Steps to reproduce (if applicable)**
1. Run the curl command
2.

**Expected behavior**
App would get installed

**System configuration:**
Release version (or git SHA-1), OS, ...
Pop!_OS 21.10

**Additional context**

evanxg852000:
@larksoftware we will look into this but it seems your bash environment does not support some constructs used in the install script.
You can try `curl -L https://install.quickwit.io | bash` in the meantime.
We will investigate the offending bash syntax.

larksoftware:
I don't know if it makes a difference but I run `zsh`.  That said I would think the shebang would send it to the proper shell.

PR #1126, opened 2022-02-03
Fix install script

evanxg852000:
Fix install script for shell interpreters that don't support array syntax.

https://github.com/quickwit-oss/quickwit/issues/1116

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -186,6 +186,7 @@ semverLT() {
 
 # Returns the tag of the latest stable release (in terms of semver and not of release date)
 get_latest_version() {
+    need_cmd sed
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -186,6 +186,7 @@ semverLT() {
 
 # Returns the tag of the latest stable release (in terms of semver and not of release date)
 get_latest_version() {
+    need_cmd sed
```
guilload, line 189: Though it should be fine here, I'm always wary of using `sed` for a multi-platform purpose because it behaves slightly differently on OSX.

guilload (approved):
LGTM.

PR #1136, opened 2022-02-14
Sets default for store_source to false.

### LINKED ISSUE #1132: fix store source defaults or fix the hdfs log tutorial?
After, following the hdfs log tutorial I end up with fields stored twice.

Once individually, and once due to store_source.

fulmicoton:
`store_source` set to true makes a lot of sense...
However we already set each individual field store attribute in the
schema to true.

Having both set to true is redundant and needlessly inflates the
index size.

Closes #1132


fmassot (approved):
Agreed. I thought the default was false....

PR #1163, opened 2022-03-11
Search and indexing can run together, and do so as a default Issue/1160

### LINKED ISSUE #1160: Non-exclusive service and rest API for everyone.
We want indexing and search to be able to coexist in the same process.
We also want to have a rest API for indexing servers too (for monitoring etc.)

# Change in usage

- `quickwit services run` starts both searcher and indexer 
- `quickwit services run [--service <servicelist>]+` runs a subset of services

# Rest, gRPC and service

Regardless of which services is running, the command starts a gRPC server and a Rest server.

However, if the searcher service is not available for instance, the search endpoint will however not be available and return an explicit error message. 

Cluster membership is always started.
We will use scuttlebutt to advertise the list of service available for each nodes.

fulmicoton:
This PR makes it possible to run indexing and search together.

They are now "services"...
The gRPC and the rest server are always running and make all running services available.

In order to make it possible, we will be temporarily relying on a specific unpublished rev of tonic,
to benefit from the following PR. 
TLDR the generic in tonic's router have been removed very recently.
In other words, the type of a gRPC does not depend on the subset of service that are being mounted.

https://github.com/hyperium/tonic/pull/830

- This PR also makes a couple of code improvement. ClusterServiceImpl is removed as it was useless.
- Useless generics have been removed in quickwit-serve.

Closes #1160

### DIFF
--- a/quickwit-indexing/src/lib.rs
+++ b/quickwit-indexing/src/lib.rs
@@ -21,13 +21,15 @@ use std::path::PathBuf;
 use std::sync::Arc;
 
 use anyhow::bail;
-use quickwit_actors::Universe;
-use quickwit_config::{IndexerConfig, SourceConfig};
+use quickwit_actors::{Mailbox, Universe};
+use quickwit_config::{IndexerConfig, QuickwitConfig, SourceConfig};
 use quickwit_metastore::Metastore;
 use quickwit_storage::StorageUriResolver;
+use tracing::info;
 
+pub use crate::actors::IndexingServerError;
 use crate::actors::{IndexingPipeline, IndexingPipelineParams, IndexingServer};
-use crate::models::{DetachPipeline, IndexingStatistics, SpawnPipeline};
+use crate::models::{DetachPipeline, IndexingStatistics, SpawnPipeline, SpawnPipelinesForIndex};
 pub use crate::split_store::{
     get_tantivy_directory_from_split_bundle, IndexingSplitStore, IndexingSplitStoreParams,
     SplitFolder,
@@ -45,7 +47,7 @@ mod test_utils;
 pub use test_utils::{mock_split, mock_split_meta, TestSandbox};
 
 pub use self::garbage_collection::{delete_splits_with_files, run_garbage_collect, FileEntry};
-pub use self::merge_policy::{MergePolicy, StableMultitenantWithTimestampMergePolicy};
+use self::merge_policy::{MergePolicy, StableMultitenantWithTimestampMergePolicy};
 pub use self::source::check_source_connectivity;
 
 pub async fn index_data(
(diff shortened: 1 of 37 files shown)


quickwit-proto/src/cluster.rs
```diff
@@ -4,36 +4,39 @@
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct Member {
     //// Member ID.　A string of the UUID.
-    #[prost(string, tag = "1")]
+    #[prost(string, tag="1")]
```
fmassot, line 7: seems that something changed in the proto auto generated files. Is this on purpose?

quickwit-serve/src/search_api/rest_handler.rs
```diff
@@ -166,44 +166,59 @@ fn search_post_filter(
...
 ///
 /// Parses the search request from the
-pub fn search_get_handler<TSearchService: SearchService>(
-    search_service: Arc<TSearchService>,
+pub fn search_get_handler(
+    search_service_opt: Option<Arc<dyn SearchService>>,
 ) -> impl Filter<Extract = impl warp::Reply, Error = Rejection> + Clone {
     search_get_filter()
-        .and(warp::any().map(move || search_service.clone()))
+        .and(require(search_service_opt))
```
fmassot, line 202: nice! I did not think about using this trick to always return same impl.

I was thinking about adding/removing handlers depending on where the service is active. Of course, there is one small cons: every hander must be something like `BoxFilter<dyn Replay>`.

quickwit-serve/src/lib.rs
```diff
@@ -21,77 +21,104 @@ mod args;
...
-    let rest_server = start_rest_service(rest_socket_addr, search_service, cluster_service);
-    info!(
-        "Searcher ready to accept requests at http://{}/",
-        rest_socket_addr
-    );
+}
+struct QuickwitServices {
+    pub cluster_service: Arc<dyn ClusterService>,
+    pub search_service: Option<Arc<dyn SearchService>>,
+    pub indexer_service: Option<Arc<IndexingServerClient>>,
```
fmassot, line 72: we could rename `IndexingServerClient` in `IndexingService`

quickwit-serve/src/lib.rs
```diff
@@ -21,77 +21,104 @@ mod args;
...
-        quickwit_config.node_id.clone(),
-        quickwit_config.gossip_socket_addr()?,
-    )?);
-    for seed_socket_addr in quickwit_config.seed_socket_addrs()? {
-        // If the peer seed address is specified,
-        // it joins the cluster in which that node participates.
-        debug!(peer_seed_addr = %seed_socket_addr, "Add peer seed node.");
-        cluster.add_peer_node(seed_socket_addr).await;
+#[derive(Debug, PartialEq, Eq, Copy, Hash, Clone)]
+pub enum QService {
```
fmassot, line 51: I would probably choose `QuickwitService` as we tend to use struct with `Quickwit` prefix.

fmassot (approved):
I think you nailed it with this `require`. It will work well with new services.

Can you check why the formatting of proto auto-generated files changed?

quickwit-serve/src/lib.rs
fulmicoton: Absolutely. I want to do that soon.

quickwit-indexing/src/lib.rs
```diff
@@ -68,3 +70,24 @@ pub async fn index_data(
...
+    metastore: Arc<dyn Metastore>,
+    storage_uri_resolver: StorageUriResolver,
+) -> anyhow::Result<IndexingServerClient> {
+    let client = IndexingServer::spawn(
+        &config.data_dir_path,
+        &config.indexer_config,
+        metastore,
+        storage_uri_resolver,
+    );
+    // for index_id in args.index_ids {
```
fmassot, line 85: we need that.

quickwit-proto/src/cluster.rs
fulmicoton: This is a side effect of updating tonic / prost.

quickwit-serve/src/search_api/rest_handler.rs
fulmicoton: correct.

quickwit-serve/src/lib.rs
fulmicoton: I opened #1206 accordingly.

Issue #1141, opened 2022-02-22
Scuttlebutt integration


fulmicoton:
I have started integrating ScuttleButt into quickwit cluster and here are few things I need your take on.
- Currently  in quicwit, we have  Cluster::leave  ? I am not quiet sure i understand it's role since it's exposed on the GRPC service I would have expected to have a join method as well since the node is still kept around.  In ScuttleButt we already have the same concept as shutdown that gracefully stops the node. 
- To make quick progress, I am using pulling to keep the members list up to date in Quickwit, A stream that broadcast only member changes is needed ScuttleButt::membership_change_receiver(&self)
- We need to settle on what NodeId is https://github.com/quickwit-oss/scuttlebutt/issues/1  I am in favor of this suggestion NodeId{uniq_id, listen_udp_addr, generation} 
On the last point, we will need somewhere to store last-know generation , use it when restarting. I am thinking this makes our node statefull

fulmicoton:
> Currently in quicwit, we have Cluster::leave ? 

You can remove it for the moment if it is tricky to implement.
The RPC call is actually quite useful. For instance, if you want to upgrade one node, or just replace it, you might want for a node to report that it is leaving the clustering to avoid any disruption. 


 - To make quick progress, I am using pulling to keep the members list up to date in Quickwit, A stream that broadcast only member changes is needed ScuttleButt::membership_change_receiver(&self)

This makes sense. The stream is only here to notify events isn't it? It does not send elements does it?

- For the last point, can you write a little design doc on what your NodeId implies... What happens when a node leaves and rejoins the cluster etc.

evanxg852000:
>The RPC call is actually quite useful. For instance, if you want to upgrade one node, or just replace it, you might want for a node to report that it is leaving the clustering to avoid any disruption.
Then I think the shutdown method in ScuttleButt does the same. (We might need a chat around this)

> The stream is only here to notify events isn't it? The channel contains () or Vec<Member>
Yes only reporting the change with a `Vec<Member>` to that Quickwit::Cluster::update_memebers

> For the last point, can you write a little design doc on what your NodeId implies... What happens when a node leaves and rejoins the cluster etc.
Sure

fulmicoton:
@evanxg852000 can we close this?

PR #1164, opened 2022-03-11
Scuttlebutt integration W.I.P

evanxg852000:
Implements features for scuttlebutt integration within Quickwit

https://github.com/quickwit-oss/quickwit/issues/1141

### How was this PR tested?

### DIFF
--- a/quickwit-cluster/Cargo.toml
+++ b/quickwit-cluster/Cargo.toml
@@ -23,8 +23,8 @@ tokio-stream = { version = "0.1", features = [ "sync" ] }
 tonic = "0.6"
 tracing = "0.1.29"
 uuid = "0.8"
-
 quickwit-proto = { version = "0.2.0", path = "../quickwit-proto"}
+scuttlebutt = { git = "https://github.com/quickwit-oss/scuttlebutt" }
 
 [dev-dependencies]
 itertools = '0.10'
--- a/quickwit-cluster/Cargo.toml
+++ b/quickwit-cluster/Cargo.toml
@@ -23,7 +23,7 @@ tonic = "0.6"
 tracing = "0.1.29"
 uuid = "0.8"
 quickwit-proto = { version = "0.2.0", path = "../quickwit-proto"}
-scuttlebutt = { git = "https://github.com/quickwit-oss/scuttlebutt", rev="c0cdf1a" }
+scuttlebutt = { git = "https://github.com/quickwit-oss/scuttlebutt" }
 
 [dev-dependencies]
 itertools = '0.10'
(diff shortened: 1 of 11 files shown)


quickwit-cluster/Cargo.toml
```diff
@@ -14,16 +14,15 @@ anyhow = "1.0"
...
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 thiserror = "1.0"
 tokio = { version = "1.7", features = [ "full" ] }
 tokio-stream = { version = "0.1", features = [ "sync" ] }
 tracing = "0.1.29"
 uuid = "0.8"
-
 quickwit-proto = { version = "0.2.1", path = "../quickwit-proto"}
+scuttlebutt = { git = "https://github.com/quickwit-oss/scuttlebutt", branch="main" }
```
fulmicoton, line 25: you need a version here, otherwise we will not be able to publish quickwit to Cargo. Also I'd prefer to have rev instead of branch. You can have all of these coexist I believe.
`git, rev, version`

quickwit-proto/proto/cluster.proto
```diff
@@ -42,6 +42,9 @@ message Member {
 
   /// If true, it means self.
   bool is_self = 3;
+
+  /// member reincarnation 
+  int64 generation = 4;
```
fulmicoton, line 47: uint64 maybe?

quickwit-search/src/lib.rs
```diff
@@ -67,10 +67,13 @@ pub use crate::search_stream::root_search_stream;
...
 use crate::thread_pool::run_cpu_intensive;
 
-/// Compute the gRPC port from the SWIM port.
-/// Add 1 to the SWIM port to get the gRPC port.
-pub fn swim_addr_to_grpc_addr(swim_addr: SocketAddr) -> SocketAddr {
-    SocketAddr::new(swim_addr.ip(), swim_addr.port() + 1)
+/// Compute the gRPC port from the Scuttlebutt gossip port.
+/// Add 1 to the Scuttlebutt gossip port to get the gRPC port.
+pub fn scuttlebutt_gossip_addr_to_grpc_addr(scuttlebutt_gossip_addr: SocketAddr) -> SocketAddr {
+    SocketAddr::new(
```
fulmicoton, line 73: Can you open a ticket to fix our port story?

Right now we hardcoded udp -> grpc because we did not have any way to advertise it.
Now we can use scuttlebut to let other nodes know what is our grpc port.

Spec should be:
- rest port /  gossip port / grpc port are configured in the conf
- rest port has a hardcoded default value that can be overriden
- grpc and gossip port default value are computed from the rest port
- clients use scuttlebutt to know the grpc port of their peers.

fulmicoton (approved):
LGTM. I left an important comment about one follow up.

quickwit-proto/proto/cluster.proto
evanxg852000: We are using timestamps (that can be negative) for generation.

PR #1166, opened 2022-03-13
docs: fix a format error

dzvon:
If we want to include a literal backtick character within a code span, we should use multiple backticks as the delimiter.

Before the fix:

image

After:

image


fmassot:
oh you're right! nicely spotted! thanks @dzvon

PR #1169, opened 2022-03-15
Metastore API

fmassot:
Add `list_indexes_metadatas` API to metastore.


quickwit-metastore/src/metastore/file_backed_metastore/lazy_file_backed_index.rs
```diff
@@ -0,0 +1,137 @@
...
+use crate::MetastoreResult;
+
+/// Lazy `FileBackedIndex`. It loads a `FileBackedIndex`
+/// on demand and optionaly spawns a task to poll
+/// regularly the storage and update the index.
+pub struct LazyFileBackedIndex {
+    index_id: String,
+    storage: Arc<dyn Storage>,
+    polling_interval_opt: Option<Duration>,
+    index: OnceCell<Arc<Mutex<FileBackedIndex>>>,
```
fulmicoton, line 42: nitpick: I would rename `index->lazy_index` to highlight the purpose of the `OnceCell<...>`

quickwit-metastore/src/metastore/file_backed_metastore/lazy_file_backed_index.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+impl LazyFileBackedIndex {
+    /// Create `LazyFileBackedIndex`.
+    pub fn new(
+        storage: Arc<dyn Storage>,
+        index_id: String,
+        polling_interval_opt: Option<Duration>,
+        file_backed_index: Option<FileBackedIndex>,
```
fulmicoton, line 51: is it ever called with file_backed_index not set to None?

quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
```diff
@@ -172,20 +203,21 @@ impl FileBackedMetastore {
         }
     }
 
-    /// Returns an IndexView for the given index_id.
+    /// Returns a FileBackedIndex for the given index_id.
     ///
-    /// If this is the first call during this instance for this
-    /// `index_id`, a fetch to the storage will be initiated
-    /// and might trigger an error.
+    /// If `index_id` is in a transitioning state `Creating` or `Deleting`, it will
```
fulmicoton, line 208: Why does creating yield an error?

quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
```diff
@@ -239,44 +269,109 @@ impl FileBackedMetastore {
...
 
-        let exists = per_index_metastores_wlock.contains_key(&index_metadata.index_id)
-            || index_exists(&*self.storage, &index_metadata.index_id).await?;
-
-        if exists {
-            return Err(MetastoreError::IndexAlreadyExists {
-                index_id: index_metadata.index_id.clone(),
+        // Checking if index already exists is a bit tedious:
+        // - first we check the index state: if it's `Alive`, return `IndexAlreadyExists` error, and
+        //   if it's `Creating` or `Deleting`, it's ok to override them as these are transitioning
```
fulmicoton, line 280: I'm not sure we should consider Creating as transitioning.

quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
```diff
@@ -239,44 +269,109 @@ impl FileBackedMetastore {
...
             });
         }
 
-        let index = FileBackedIndex::from(index_metadata);
+        // Set state to Creating` and rollback on metastore error.
+        per_index_metastores_wlock.insert(index_id.clone(), IndexState::Creating);
+        put_indexes_states(&*self.storage, &per_index_metastores_wlock)
+            .await
+            .map_err(|error| {
+                per_index_metastores_wlock.remove(&index_id);
```
fulmicoton, line 305: nitpick:
I think it is unexpected to put side-effects in a `map` function.

I prefer a if let Err for that.
```
if let Err(err) {
   er_index_metastores_wlock.remove(&index_id);
   return Err(err);
}
```

fulmicoton (approved):
Awesome job.

quickwit-metastore/src/metastore/file_backed_metastore/lazy_file_backed_index.rs
fmassot: In 2 cases yes: when creating the index and in the weird case where we need to do a little dance to handle potential error when fetching the metadata and taking a write lock.
In both cases, we could avoid that and just instantiate with None, but it's a bit annoying too. I think it's ok as is.

quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
fmassot: Let's do the if let.

PR #1192, opened 2022-03-16
Actor refactoring

fulmicoton:
```
    This refactoring includes different changes:
    
    - Instead of having a Message associated type, actor now can
    have receive different Message by implementing a `Handler`
    trait.
    One hidden benefit is that some of the message can be private.
    (e.g. loop messages).
    
    - Message can yield a typed reply. The type is an associated type
    of the Handler trait.
    
    - There are no more Sync and Async actors. All handlers require
    to implement an `async` method. What used to be SyncActors need to
    specify that want to be executed on an independent runtime, on a
    dedicated thread, usign the `Actor::runner` method.
    
    The refactoring was mainly motivated by the introduction of services in
    quickwit: the IndexingService (currently named IndexingServer),
    and the upcoming PushAPIService.
```


quickwit-actors/src/envelope.rs
```diff
@@ -0,0 +1,102 @@
...
+use async_trait::async_trait;
+use tokio::sync::oneshot;
+use tracing::Instrument;
+
+use crate::{Actor, ActorContext, ActorExitStatus, Handler};
+
+/// An `Envelope` is just a way to capture the handler
+/// of a message and hide its type.
+///
+/// Message can have different types but need to somehow pushed to a
```
evanxg852000, line 32:
```suggestion
/// Message can have different types but somehow need to  be pushed to a
```

quickwit-indexing/src/actors/garbage_collector.rs
```diff
@@ -52,6 +52,9 @@ pub struct GarbageCollectorCounters {
     pub num_deleted_bytes: usize,
 }
 
+#[derive(Debug)]
```
evanxg852000, line 55: Maybe this could be exported from actor crate as a dedicated utility

quickwit-indexing/src/source/mod.rs
```diff
@@ -60,13 +60,12 @@
 mod file_source;
 #[cfg(feature = "kafka")]
 mod kafka_source;
-#[cfg(feature = "kinesis")]
-mod kinesis;
+// #[cfg(feature = "kinesis")]
```
evanxg852000, line 63: I guess this is no ready yet in this PR

evanxg852000 (commented):
LGTM :+1:  just left some questions.

quickwit-indexing/src/actors/garbage_collector.rs
fulmicoton: The loop message you mean? That might make sesnse but I actually like it to be private to the actor.

quickwit-indexing/src/source/mod.rs
fulmicoton: great catch

quickwit-actors/src/actor.rs
```diff
@@ -146,19 +154,40 @@ pub trait Actor: Send + Sync + Sized + 'static {
...
+    }
+
+    /// Hook  that can be set up to define what should happen upon actor exit.
+    /// This hook is called only once.
+    ///
+    /// It is always called regardless of the reason why the actor exited.
+    /// The exit status is passed as an argument to make it possible to act conditionnally
+    /// upon it.
+    ///
+    /// It is recommended to do as little work as possible on a killed actor.
```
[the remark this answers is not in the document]
fulmicoton: Killed was accurate here. It is related to the sentence above.
I'll try to rephrase.

Issue #1191, opened 2022-03-16
Fix REST, gRPC, and UDP ports config


evanxg852000:
The previous cluster implementation could not allow us to share node ports.
Now we can use scuttlebutt to let other nodes know what is our gRPC port.

The Specs are as follow:
- REST,  Gossip (udp),  and gRPC ports are configured from the configuration file.
- REST port has a hardcoded default value that can be overridden from the configuration file.
- gRPC and gossip (UDP) ports default values are computed from the RESP port.
- A client node should use scuttlebutt to know the gRPC port of its peers.

Also, expose a `ClusterState` method via the `ClusterService` gRPC service. This method should display the cluster state from a node's perspective in JSON format.

fulmicoton:
Closing...

PR #1202, opened 2022-03-17
Fix REST, gRPC, and UDP ports

evanxg852000:
### Description
- implemented fixes for our service ports story 
- implemented cluster state gRPC method

### How was this PR tested?
- unit test
- tested with CLI

https://github.com/quickwit-oss/quickwit/issues/1191


quickwit-config/src/config.rs
```diff
@@ -543,6 +569,68 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_socket_addr_ports() {
```
fulmicoton, line 573: Exactly the unit test I wanted to see.

fulmicoton (approved):

PR #1205, opened 2022-03-20
reduce mem allocations

PSeitz:
remove field mapping allocation per doc

Spotted in heaptrack,  vec of fastfield mappings is created _per_ document
heaptrack


quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
```diff
@@ -45,29 +45,48 @@ pub struct FieldMappingEntry {
...
+            mapping_type,
+            field_entries,
+            fast_field_entries,
+        }
     }
 
     /// Creates a new root [`FieldMappingEntry`].
     pub fn root(mapping_type: FieldMappingType) -> Self {
+        let name = "".to_string();
+        let field_entries = Self::compute_field_entries(name.as_str(), &mapping_type);
```
fulmicoton, line 71: wha't's the deal with name?

fulmicoton:
@PSeitz nice catch. Can you double check your code though before sending a CR? I didn't dig much but the  `"".to_string().as_str()` seemed fishy.
@fmassot can you review?

quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
PSeitz: "" is the name for a root entry

PSeitz:
> @PSeitz nice catch. Can you double check your code though before sending a CR? I didn't dig much but the `"".to_string().as_str()` seemed fishy. @fmassot can you review?

there is no `"".to_string().as_str()`, but something similar where the value is used later

fulmicoton:
@fmassot can you have a look?

fmassot (approved):
Well spotted!

PR #1210, opened 2022-03-21
Updated readme: UI & content

maximkpa:
Updated readme: UI & content


fmassot:
@fulmicoton Almost good, can you check if there is no blunder in it?

evanxg852000:
@maximkpa @fmassot  Please, Can we have a screenshot of the search UI with the browser frame around the screenshot seems to be cut off though it's not. 
Something like this with transparent background

fmassot:
@evanxg852000 good point, we can definitely improve this.

fmassot:
we plan to make a small video / gif, so we might skip the browser frame part.

PR #1215, opened 2022-03-23
Add option for fieldnorm

### LINKED ISSUE #1134: Remove fieldnorm
Tantivy now makes it possible to not index fieldnorm. This is perfectly useless for logs so we might as well disable it in quickwit. This should make all index a few percent smaller.

infiniteregrets:
### Description

This PR makes the fieldnorm feature optional as it was done in the tantivy library (refer: https://github.com/quickwit-oss/tantivy/pull/1001). The index configuration now supports `fieldnorms` field which can be set to a boolean value and by default is set to false. 
Closes #1134 

### How was this PR tested?
- Created a custom wikipedia index with the `fieldnorms` option enabled and found that the configuration was being serialized without any errors. 
- Wrote a test to check if the case when the FieldMappingType is Text, whether the index options for fieldnorms are set to true or not for the passed json with fieldnorms set to true.

Signed-off-by: Mehul Arora <aroram18@mcmaster.ca>


fmassot:
@infiniteregrets cool! To test it completely, that would be awesome to follow these steps:
- index Wikipedia datasets with fieldnorms
- index Wikipedia datasets without fieldnorms
- use the CLI to extract a split and look at the different tantivy files. You should find a .fieldnorm file and the size should be almost 0 without fieldnorms.

infiniteregrets:
Okay sure, I will test it out. I am actually missing the default fieldnorm setting for different types like i64, u64 etc. so I will fix that first

fmassot:
sounds good. @infiniteregrets did you try to index with and without fieldnorms?

fulmicoton:
Thank you @infiniteregrets

PR #1231, opened 2022-03-28
Add index management & UI REST API

fmassot:
Ready for review.


quickwit-core/Cargo.toml
```diff
@@ -33,7 +34,9 @@ serde_json = "1.0"
 tracing = "0.1.29"
 tokio-stream = "0.1"
 tempfile = '3'
+chrono = "0.4"
 
 [dev-dependencies]
 mockall = "0.11"
 quickwit-metastore = { version = "0.2.1", path = "../quickwit-metastore", features=["testsuite"]}
+serde_yaml = "0.8"
```
evanxg852000, line 42: Add an empty line at EOF, maybe configure your Text Editor to do it on save

quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -0,0 +1,305 @@
...
+
+async fn get_index_metadata(
+    index_id: String,
+    index_service: Arc<IndexService>,
+) -> Result<impl warp::Reply, Infallible> {
+    info!(index_id = %index_id, "get-index");
+    let index_metadata = index_service
+        .get_index(&index_id)
+        .await
+        .map_err(SearchError::from);
```
evanxg852000, line 58: Maybe we could have `IndexError` struct that just implement 
```rust 
struct IndexError;
impl warp::reject::Reject for IndexError {}
```
Maybe we could also rename SearchError into something more generic since we now have more services.
I suggest we create a small issue for this rather than you handling it here.

quickwit-serve/src/push_api/rest_handler.rs
```diff
@@ -50,7 +50,7 @@ pub fn ingest_handler(
 }
 
 fn ingest_filter() -> impl Filter<Extract = (String, String), Error = Rejection> + Clone {
-    warp::path!("api" / "v1" / String / "ingest")
+    warp::path!(String / "ingest")
```
evanxg852000, line 53: Any reason this is being unversioned?

quickwit-serve/src/rest.rs
```diff
@@ -44,8 +45,8 @@ pub(crate) async fn start_rest_server(
     let metrics_service = warp::path("metrics")
         .and(warp::get())
         .map(metrics::metrics_handler);
-    let rest_routes = liveness_check_handler()
-        .or(cluster_handler(quickwit_services.cluster_service.clone()))
+    let api_v1_root_url = warp::path!("api" / "v1" / ..);
```
evanxg852000, line 48: Nice refactoring :100:

quickwit-serve/src/rest.rs
```diff
@@ -58,28 +59,43 @@ pub(crate) async fn start_rest_server(
...
-    match rejection.find::<serde_qs::Error>() {
-        Some(err) => {
-            let err_msg = err.to_string();
-            Ok(Format::PrettyJson.make_reply_for_err(FormatError {
-                code: ServiceErrorCode::BadRequest,
-                error: err_msg,
-            }))
-        }
-        None => Ok(Format::PrettyJson.make_reply_for_err(FormatError {
+    if rejection.is_not_found() {
```
evanxg852000, line 79: Nitpick: I think the early return style would be better here given we might have more custom error to handle.

evanxg852000 (approved):
Looks good, just left some small comments/suggestions

quickwit-serve/src/rest.rs
fmassot: Actually, I hesitated to make early returns but here I think readability is quite the same with or without.

quickwit-serve/src/index_api/rest_handler.rs
fmassot: Good point.

fmassot: I created an custom `IndexServiceError`. I used the same pattern as the one used for other service errors.

Issue #1239, opened 2022-03-29
Implement a Push REST API


fulmicoton:
We want a rest push api :)

Actually we probably want 2 apis.
One compatible with elastic's bulk api, and one specific to quickwit.
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html

Quickwit is append-only so the "verbs" (index/create) is meaningless for quickwit...
we want to treat both as a new document.
We also want to return an error if an unsupported operation is sent in the bulk API. (e.g delete)

evanxg852000:
@fulmicoton @fmassot 
1. Right now I think a separate bulk API for Quickwit might not be needed as this could be done by the ingest API.
My reasoning is that we support `append-only` operation so there is no need to have the same API spec as elastic bulk API.
2. For the elastic bulk API, we want to specify how the source input should be structured for the operations: create, update and delete. right ?

evanxg852000:
Regarding certain error handling, it would have been nice to have RocksDB transactions available in the wrapper crate we are using :disappointed: . 
I think we will have to `undo` when an error occurs (ugly). 
There is `WriteBatch` but I am concerned about making a bunch of `create_cf` atomic which is not available on `WriteBatch`
There is a PR but not sure it's gonna land soon https://github.com/rust-rocksdb/rust-rocksdb/pull/565

fulmicoton:
> 2. For the elastic bulk API, we want to specify how the source input should be structured for the operations: create, update and 
   delete. right ?
   
Just return an error for update/delete, and treat both create and index like a create.

fulmicoton:
```
Regarding certain error handling, it would have been nice to have RocksDB transactions exposed from the wrapper crate we are using disappointed .
I think we will have to undo when an error occurs (ugly).
There is WriteBatch but I am concerned about making a bunch of create_cf atomic which is not available on WriteBatch
There is a PR but not sure it's gonna land soon https://github.com/rust-rocksdb/rust-rocksdb/pull/565
```

@evanxg852000 For the moment let's not add this complexity.

First, this is only a concert for the ES API. We have only one possible targetted in a single ingest call with the regular API.
Second, we need to progress on #1229 , but I suspect atomically creating  of cf is not much of a problem.

I suspect we can get away with:
The metastore is the single source of truth for the existence or the inexistence of an index.
cf are created and opened lazily upon the first ingestion after a restart.

[...]

PR #1244, opened 2022-03-31
Bulk PushAPI

evanxg852000:
Implements a bulk push API on:
- `api/v1/{index_id}/bulk`: native ndjson document 
- `api/v1/_bulk`: elastic bulk API compatible with support only for `Create` and `Index`

Closes https://github.com/quickwit-oss/quickwit/issues/1239


quickwit-pushapi/src/errors.rs
```diff
@@ -31,6 +31,8 @@ pub enum PushApiError {
     QueueAlreadyExists { queue_id: String },
     #[error("PushAPI service is down")]
     PushAPIServiceDown,
+    #[error("The following indexes `{queue_ids:?}` don't exist.")]
+    IndexDoesNotExist { queue_ids: Vec<String> },
```
fulmicoton, line 35: There was already an option called IndexDoesNotExist. Should we rename

QueueDoes(Not)?Exist to IndexDoes(Not)?Exist

quickwit-pushapi/src/push_api_service.rs
```diff
@@ -18,28 +18,63 @@
...
+        Ok(PushApiService { queues, metastore })
     }
 
-    fn ingest(&mut self, request: IngestRequest) -> crate::Result<()> {
+    async fn ingest(&mut self, request: IngestRequest) -> crate::Result<()> {
+        // Check all indexes exist.
+        let tasks = request.doc_batches.iter().map(|batch| {
+            let moved_metastore = self.metastore.clone();
+            let index_id = batch.index_id.clone();
+            tokio::spawn(async move {
```
fulmicoton, line 52: I don't think spawning a task is required here.... is it?

quickwit-pushapi/src/push_api_service.rs
```diff
@@ -18,28 +18,63 @@
...
+            .filter(|(_, result)| result.is_err())
+            .map(|(index_id, _)| index_id)
+            .collect::<Vec<_>>();
+        if !queue_ids.is_empty() {
+            return Err(PushApiError::IndexDoesNotExist { queue_ids });
+        }
+
         for doc_batch in &request.doc_batches {
+            // Attempt to create the queue if it does not exist.
+            if let Err(error) = self.queues.create_queue(&doc_batch.index_id) {
```
fulmicoton, line 71: I'm not fond of using error handling as logic. It makes the code harder to understand...

Could we use the following instead?

```rust
if !self.queues.exists() {
    self.queues.create_queue(&doc_batch.index_id)?;
}
```

quickwit-pushapi/src/push_api_service.rs
```diff
@@ -18,28 +18,63 @@
...
+        metastore: Arc<dyn Metastore>,
+    ) -> crate::Result<Self> {
         let queues = Queues::open(queue_path)?;
-        Ok(PushApiService { queues })
+        Ok(PushApiService { queues, metastore })
     }
 
-    fn ingest(&mut self, request: IngestRequest) -> crate::Result<()> {
+    async fn ingest(&mut self, request: IngestRequest) -> crate::Result<()> {
+        // Check all indexes exist.
```
fulmicoton, line 48: a call to the metastore every single time will be too expensive...

I think we want to skip the existence check if the queue exists.

... which will leaves us with a need to propage index deletion.
That part can be done in a separate ticket.

fulmicoton (changes requested):
see comments inline

quickwit-pushapi/src/push_api_service.rs
evanxg852000: I was trying to avoid that extra call since we are in a loop. but you are right let's make it readable first.

evanxg852000: I was try to avoid that extra call to `self.queue_exists(queue_id)` inside `create_queue`, but you are right let's make it readable first

quickwit-pushapi/src/push_api_service.rs
evanxg852000: This has to do with `async` block. I will extract it into a function to avoid this.

quickwit-pushapi/src/errors.rs
evanxg852000: We can rename but we lose precision between:
- Index missing error (metastore)
- queue missing error (queues)
Are you good with renaming `IndexDoesNotExist` ?

fulmicoton: I'm good with it

fulmicoton: Both for Exist and DoestNotExist

quickwit-pushapi/src/push_api_service.rs
evanxg852000: rust_analyser was giving errors on this, but build worked.

quickwit-pushapi/src/errors.rs
evanxg852000: I kept "IndexAlreadyExists" instead of `IndexDoesExists` is visually too close to `IndexDoesNotExists` and can lead to a hard time debugging.

fulmicoton (approved):

PR #1253, opened 2022-04-04
Update dockerfile with new native dependencies

xvello:
### Description

`docker build .` fails on `main` due to the new native dependencies for `librocksdb-sys`. This PR adds them to fix the build.

### How was this PR tested?

`docker build -t qw . && docker run -it qw run` locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,10 +11,12 @@ LABEL org.opencontainers.image.licenses="AGPL-3.0"
 
 RUN apt-get -y update \
     && apt-get -y install ca-certificates \
+                          clang \
                           cmake \
                           libpq-dev \
                           libpq5  \
                           libssl-dev \
+                          llvm \
     && rm -rf /var/lib/apt/lists/*
 
 # Required by tonic


fulmicoton:
Sweet... was llvm/clang really due to rocksdb or due to protobuf? If the latter, `apt install -y protobuf-compiler`  might be lighter

xvello:
Yes, here is the error I get on `main`:

```
error: failed to run custom build command for `librocksdb-sys v0.6.1+6.28.2`

Caused by:
  process didn't exit successfully: `/quickwit/target/release/build/librocksdb-sys-f46e47f1547d4828/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'Unable to find libclang: "couldn't find any valid shared libraries matching: ['libclang.so', 'libclang-*.so', 'libclang.so.*', 'libclang-*.so.*'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [])"', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/bindgen-0.59.2/src/lib.rs:2144:31
```

evanxg852000 (approved):
Indeed these packages are needed for building RocksDB.
LGTM :+1:

PR #1256, opened 2022-04-05
Record progress after each message received

guilload:
### Description
Replace the convoluted `stream.take_until(...)` with the more standard `tokio::select! {...}`.

### How was this PR tested?
`make test-all`

### DIFF
--- a/quickwit-indexing/src/source/kafka_source.rs
+++ b/quickwit-indexing/src/source/kafka_source.rs
@@ -175,7 +175,7 @@ impl Source for KafkaSource {
         let mut docs = Vec::new();
         let mut checkpoint_delta = CheckpointDelta::default();
 
-        let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT / 2);
+        let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT);
         let mut message_stream = Box::pin(self.consumer.stream().take_until(deadline));
 
         let mut batch_num_bytes = 0;
@@ -232,6 +232,7 @@ impl Source for KafkaSource {
             if batch_num_bytes >= TARGET_BATCH_NUM_BYTES {
                 break;
             }
+            ctx.record_progress();
         }
         if !checkpoint_delta.is_empty() {
             let batch = RawDocBatch {


fulmicoton:
Why is this an improvement?

fulmicoton (commented):
I don't understand why this is an improvement. Isn't the new code more complicated than the old one?

vector for instance, went the other way
https://github.com/vectordotdev/vector/pull/5173

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -172,65 +172,76 @@ impl Source for KafkaSource {
...
+        let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT);
+        tokio::pin!(deadline);
+
+        let mut stream = self.consumer.stream();
+
+        loop {
+            tokio::select! {
+                message_opt = stream.next() => {
+                    if let Some(message_res) = message_opt {
+                        let message = match message_res {
```
PSeitz, line 188: I think the indentation reaches a level here where is hurts readability

guilload:
Back when we implemented the Kafka source, we had some concerns about the correctness of using `take_until` with a stream. More recently, as we were discussing the implementation of the Kinesis source, which uses `tokio::select!`, I suggested that we use the same pattern in the Kafka source so we no longer have that doubt. You seemed to agree.

I'm happy to revert to `take_until` but I'd still like to add the call `ctx.record_progress();` in the loop. WDYT?

fulmicoton:
@guilload I cannot recall the conversation unfortunately. One problem that I can think of is the possibility that by stopping and reopening a stream could end up skipping some messages or reemitting some messages.

The problem is the same whether we use take_until or select, and depend on the implementation of the Kafka consumer.

fulmicoton:
The take until stuff looks ok.

https://docs.rs/futures-util/latest/src/futures_util/stream/stream/take_until.rs.html#119-138
"In poll_next, on the code path where we call poll_next on the stream, if an item is available we always return it."

fulmicoton:
Same for rdkafka... If poll is called, we return the item straight away. There is no buffering happening for instance.

quickwit-indexing/src/source/kafka_source.rs
guilload: This is relevant now since we're not going to move forward with this PR as is, but I agree with you :)

fulmicoton (approved):

Issue #1238, opened 2022-03-29
k8s support


fulmicoton:
We want quickwit to be easy to deploy on k8s.

One issue that we have is how do we start the cluster formation via scuttlebutt.

It requires to get connected to a small set of peers that we call peer nodes.

It is tempting to put a k8s service in front of our nodes and set the peer node IP to the service IP.
The default of this approach is that the service might connect us to ourselves, creating a bunch of 
partition in our cluster.

Another solution might be to relly on the k8s API to get our peers.

Another solution could be to rely on a headless service and list DNS entires.

Another problem that we have is that we may have nodes dedicated to a specific service.
For instance: an indexing only node and 2 search only node.

How should we deal with this? Should we have a quickwit headless service tied will all quickwit nodes, and then 
several specific services tied to specific deployments?

etc.

guilload:
Duplicate of #1658.

Issue #1011, opened 2021-12-31
Specify and implement a generic way of overriding config parameters


guilload:
For instance:
`quickwit service run searcher --config-override node-id=my-node-id searcher.listen-address=0.0.0.0`
or
`QW_NODE_ID=my-node-id QW_SEARCHER:LISTEN_ADDRESS=0.0.0.0 quickwit service run searcher`.

This is just an example. Let's discuss whether we support overrides through CLI and/or env variables, priority between overrides, syntax, implementation.

Food for thoughts:
https://materialize.com/docs/cli/#command-line-flags
```
if a command line flag takes an argument, you can alternatively set that flag via an environment variable named after the flag. If both the environment variable and command line flag are specified, the command line flag takes precedence.

The process for converting a flag name to an environment variable name is as follows:

Convert all characters to uppercase
Replace all hyphens with underscores
add an MZ_ prefix.
For example, the --data-directory command line flag corresponds to the MZ_DATA_DIRECTORY environment variable.
```

fulmicoton:
Addressing a similar use case...

We could do string interpolation with the environment variables and have people define their own template.

PSeitz:
If we support both, then the priorities should be:
* cli
* env
* config

But a generic solution with clap may be not so easy.

fulmicoton:
Developing on my previous comment as it was a bit too succinct.

We could let users reference environment variables in their config.

For instance:

quickwit.yaml

```yaml
rest_listen_port: ${REST_LISTEN_PORT}
```

We can use an off-the-shelve template engine, pick the environment variable dictionary and do the render quite simply.
It is easy to implement, and perfect for the k8s use case at least.

fulmicoton:
Developing on my previous comment as it was a bit too succinct.

We could let users reference environment variables in their config.

For instance:

quickwit.yaml

```yaml
rest_listen_port: ${REST_LISTEN_PORT}
```

We can use an off-the-shelve template engine, pick the environment variable dictionary and do the render quite simply.
It is easy to implement, and perfect for the k8s use case at least.

If we go this route (it has my votes), it would probably require displaying the resolved conf in the admin interface to let people debug.

PR #1260, opened 2022-04-06
Bootstrap kubernetes-discovery RFC

xvello:
This Request For Comment discusses the context, target and risks and proposed changes to implement https://github.com/quickwit-oss/quickwit/issues/1238

Once the proposed solution is discussed and agreed upon, it can progressively be implemented in separate PRs


quickwit-cluster/docs/kubernetes-discovery.md
```diff
@@ -0,0 +1,195 @@
...
+Alternatives are discussed in the next section, with a pros/cons comparison.
+
+### 1a. Source the sibling pods as seeds via DNS enumeration on startup
+
+A new `peer_seed_service` configuration option is added to Quickwit. It is exclusive to the existing `peer_seeds`,
+and setting both will result on a validation error. `quickwit-cluster::start_cluster_service` is updated to check
+this option. If it is not empty:
+
+- a DNS enumeration request is issued, that will return the IP addresses of all the live pods in the cluster,
+- if the DNS request fails or if the result is empty, Quickwit fails to start. Kubernetes will restart the
```
fulmicoton, line 115: interesting. Is it common for the result to be empty? I think our own IP address will be there usually but I guess this is not guaranteed.

quickwit-cluster/docs/kubernetes-discovery.md
```diff
@@ -0,0 +1,195 @@
...
+  container, allowing to retry the request when more pods are created. In case of repeated failures (misconfiguration,
+  or DNS error), the CrashLoopBackoff mechanism will engage, reducing the pressure on the DNS server by keeping
+  pods down longer between retries,
+- if not empty, the response is used to populate the `seed_nodes` argument passed to `scuttlebutt`.
+
+Due to the eventually consistent nature of the k8s control plane, there will be a delay before all pods are exposed
+in the DNS record. We do have guarantees that the responses to the different pods will be consistent. Pods will
```
fulmicoton, line 122: If we have pod A and pod B that starts at the same time, can you confirm that it is impossible for k8s DNS to deliver "eventually consistent results in the sense of ":

- node A receives only node A IP
- node B receives only node B IP.

Hence creating a split brain.

quickwit-cluster/docs/kubernetes-discovery.md
```diff
@@ -0,0 +1,220 @@
...
+
+If an existing node restarts (due to an out-of-memory error for example), it will request up-to-date seeds again
+from the DNS servers, discarding its previous state. This is better than running the DNS enumeration in an
+[init container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/), which would not
+update the seeds on container restart.
+
+### 2. Gate cluster membership on new a `cluster_name` property
+
+Redis, Cassandra and Elasticsearch all had to solve the IP reuse problem before. They all settled on the simplest and
+most resilient solution: adding a `cluster_name` property to the nodes, and forbidding nodes to join if their
```
fulmicoton, line 138: that sounds great.

quickwit-cluster/docs/kubernetes-discovery.md
```diff
@@ -0,0 +1,220 @@
...
+advertised cluster name does not match the other peers. This mechanism can be implemented in `scuttlebutt` with
+the following changes:
+
+- adding a new required argument `cluster_name` to the `ScuttleServer::spawn` function (String type)
+- copying that value as a new  `cluster_name` field to the `Digest` object
```
fulmicoton, line 143: Syn or Digest?

quickwit-cluster/docs/kubernetes-discovery.md
xvello: When the chart is first applied and the first pods come up (which can be fast if capacity is already available in the cluster), they could run the request before the DNS entry is created and propagated to the servers.

In that case, failing and retrying later will allow the cluster to eventually bootstrap without human intervention. We could decide to implement a retry logic, but we would be adding complexity for the same outcome. I prefer to rely on k8s's crash loop backoff mechanism, as it is more visible to cluster operators.

quickwit-cluster/docs/kubernetes-discovery.md
xvello: `Syn` could be enough, but implementing the check at the `Digest` level also allows nodes to avoid being "poisoned" by a bad `SynAck` message from a buggy release.

quickwit-cluster/docs/kubernetes-discovery.md
```diff
@@ -0,0 +1,220 @@
...
+
+### 3. Use the pod name as default `node_id`
+
+Pods are assigned a unique name by the system on creation. Reusing this pod name as the `node_id` will make it easier
+for cluster operators to navigate between the Quickwit admin and their infrastructure monitoring. This name is available
+in the `$POD_NAME` environment variable and can be injected in arbitrarily named environment variables using a 
+[fieldRef](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/).
+
+The best practice for Kubernetes-native services is to allow overriding configuration values with envvar.
+We will update `quickwit-config` to lookup the `QUICKWIT_NODE_ID` envvar, and use its value if not empty.
```
fulmicoton, line 160: This is something that often comes up in discussions.

I wonder if we should offer string interpolation in the quickwit config, using environment variables as a context `${QUICKWIT_NODE_ID}`, possibly with default values `${QUICKWIT_CLUSTER_ID:"test_id"}` maybe a bit like bash does.

I feel like one of the benefit is to leave an explicit trail about the population of a value.
"How did I set the cluster_id? Oh it is a environment value called blabla"

As opposed to having to look at the doc.

quickwit-cluster/docs/kubernetes-discovery.md
```diff
@@ -0,0 +1,220 @@
...
+  adding load to these servers when alternative solutions exist can be frowned upon
+- these endpoints expose a lot of information about the infrastructure and could be used by intruders for
+  lateral movement and privilege escalation
+
+Also, the DNS solution could be used in other containerized environments (Nomad/Consul or Mesos for example),
+which we might have to support at some point in the future.
+
+### 1c. Use a stable "seed" pool
+
+I have seen distributed systems use a set of three idle machines as their seed pool. Their single job is to
```
fulmicoton, line 207: The service discovery at indeed is a naked mongo db that works like this.

fulmicoton (commented):
See comments inline

quickwit-cluster/docs/kubernetes-discovery.md
xvello: The Kubernetes control plane is implemented as a series of single-threaded loops (called controllers) to avoid race conditions. This ensures that pods will be registered in the [Endpoints](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#endpoints-v1-core) entity in a consistent order.

That `Endpoints` entity is the source of truth for the DNS record. We might encounter a pathological DNS layer with inconsistent responses, but the common expectation is for the record to fully reflect one version of the `Endpoints` entity (not necessarily the latest though, due to propagation).

If A is registered in the `Endpoints` entity before B, we should  get either A, or A+B, never B alone.

quickwit-cluster/docs/kubernetes-discovery.md
```diff
@@ -0,0 +1,220 @@
...
+
+Defining the topology of the official Quickwit chart is out of scope for this document (especially choosing between
+deployments and statefulsets), but we will consider this potential topology (inspired from Bitnami's
+[Cassandra chart](https://github.com/bitnami/charts/tree/master/bitnami/cassandra/templates) and my experience)
+as an illustration for discussions on the proposed solution:
+
+- a `$release-indexing` deployment for an elastic pool of indexer nodes, that can be quickly auto-scaled based on volume
+- a `$release-indexing` service, to provide a DNS endpoint for the push API. This service uses a label selector to 
+  expose pods of the `$release-indexing` deployment once [they are ready](https://blog.colinbreck.com/kubernetes-liveness-and-readiness-probes-how-to-avoid-shooting-yourself-in-the-foot/)
+- a `$release-search` statefulset, for a pool of searcher nodes. Statefulset would allow prioritizing disk caching
```
fulmicoton, line 93: There are no disk caching on the search side at the moment. We do have RAM caching for the hotcache and column-oriented storage. Maybe we can stick to a simple deployment at the moment? Would that be more reassuring for users?

We would like to have some persistent identity however, but we just need it to be an id. 
The point there is that a node that would just be restarted would regain its node id, and take the place of its
previous instance in term of affinity with the splits. If possible, that should improve our cache efficiency.

e.g. 
Let's imagine all we need cached is cached over N searcher.
One stateful node goes down. To simplify we imagine no query is executed (but the story still holds).
The stateful node goes up but with the same node id.
All leaf request splits running on the new node will be a cache miss, but all the other nodes will have cache hits. 

With a random node id, 
All leaf request splits running on the new node will be have numsplits/N cache miss where N is the number of servers... And all splits on the new node will also be cache miss.

quickwit-cluster/docs/kubernetes-discovery.md
fulmicoton: And we are certain the DNS will always serve a view consistent with a past state of the control plane?

Could there be a silly situation where a cloud provider uses an adhoc distributed DNS caching service that uses a replication with eventual consistency semantics that could yield this situation?
(no sure if this is a dumb question or not.)

fulmicoton: Hmmm there a lot DNS caching solutions... But they all act ... as a cache. I'm overthinking this.

quickwit-cluster/docs/kubernetes-discovery.md
```diff
@@ -0,0 +1,195 @@
...
+  - with a label selector matching both the `$release-indexing` and `$release-search` pods, as a unified cluster
+
+## Proposed solution
+
+The following action items are ordered by priority, and could be independently implemented in several releases.
+Alternatives are discussed in the next section, with a pros/cons comparison.
+
+### 1a. Source the sibling pods as seeds via DNS enumeration on startup
+
+A new `peer_seed_service` configuration option is added to Quickwit. It is exclusive to the existing `peer_seeds`,
```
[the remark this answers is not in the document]
xvello: I have no strong opinion on the topic, and both a fine. My personal preference is for the intent to be as clear as possible on this kind of options, to protect my 3am self from doing a dumb mistake during an incident ; which is why I propose a separate option.

quickwit-cluster/docs/kubernetes-discovery.md
xvello: Adding string interpolation and setting `node_id: $POD_NAME` in the example k8s manifests would be a quick win here, but having automatic envvar mapping would be a cheap way to gain cookie points with seasoned k8s users.

The Datadog agent has its main config as a yaml file, but every option is automatically mapped to the `DD_OPTION_NAME` envvar, which allows to run simple deployments without even creating a configmap to provide the configuration.

quickwit-cluster/docs/kubernetes-discovery.md
xvello: It is too early to settle on this question, and a deployment would definitely work fine. I decided to use both types in my example topology to make sure the solution we discuss works with both, instead of assuming behaviours of deploy or sts.

Please note that although deployments result in random pod names, pods created from statefulsets will have a deterministic `$sts_name-$ordinal` pod name. This is one of the reasons I suggested to use the pod name as `node_id`. When the pod comes back from its update/restart, it will join back with the same `node_id` prefix, and use it's persisted disk cache to serve queries.

quickwit-cluster/docs/kubernetes-discovery.md
fulmicoton: (Receiving an Ack message is what triggers a write.)
Let s simply put it on all message then maybe?

quickwit-cluster/docs/kubernetes-discovery.md
xvello: The k8s ecosystem is so fragmented (especially in the networking part), that it is difficult to give a definitive answer on that. My anecdotal experience with the two most popular solutions ([kube-dns](https://github.com/kubernetes/dns) and [coredns](https://github.com/coredns/coredns/tree/master/plugin/kubernetes)) makes me pretty confident on this solution, but users could run alternative DNS servers with a buggy implementation.

Other solutions make this assumptions, so I would expect the risk for this to be undetected to be low. If that were to happen, there are easy mitigations we can suggest:

- deploy new clusters with a single pod (`replicas: 1`), then manually upscale once the DNS entry is created,
- add an entrypoint or an init container delaying the start of the Quickwit process for a couple of seconds, to ensure the DNS has settled.

quickwit-cluster/docs/kubernetes-discovery.md
fulmicoton: SGTM

quickwit-cluster/docs/kubernetes-discovery.md
fulmicoton: and there is another option, that Adrien likes, which is override via CLI arguments.
( " ^_^ )

guilload: For reference:
https://github.com/quickwit-oss/quickwit/issues/1011

quickwit-cluster/docs/kubernetes-discovery.md
fulmicoton: My question was probably unclear.

I don't think deployment allow that,
but is it possible to have statefulsets and enjoy a sticky name, while oddly not having to configure a persisted volume.
(I think it is not possible)?

That's an easy to address but odd requirement that does not seem supported by k8s.

xvello: Definitely, the [k8s docs]() state:

> 
> StatefulSets[ ](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#using-statefulsets)are valuable for applications that require one or more of the following.
> 
>  -   Stable, unique network identifiers.
>  -   Stable, persistent storage.
>  -   Ordered, graceful deployment and scaling.
>  -   Ordered, automated rolling updates.

I think that it's OK to define a statefulset without any `volumeClaimTemplates`, but we can try it out to confirm.

quickwit-cluster/docs/kubernetes-discovery.md
chhetripradeep: Using just cluster_name doesn't sound too resilient when someone misconfigures cluster_name. I agree cluster_name should be there but i would like to suggest using some kind of mTLS based communication so that a staging env node can't join production cluster.

xvello: Thanks for the feedback @chhetripradeep.
Allowing network traffic between staging and prod is usually discouraged (as it can open a privilege escalation path to an intruder) but we cannot assume all users have effective network segmentation.

Adding mTLS on top should be doable in a subsequent release, provided the cluster operators provide the cert and key via an init-container, or a secret.

Issue #1235, opened 2022-03-29
PushAPI Source


fulmicoton:
We need a push api source

evanxg852000:
looking at this, my guess is that we need a source that takes as input the HTTP request body and emits batch of docs. 
I am starting to think it could probably be better to have the elastic specific parsing in a dedicated source `ElasticBulkApiSource` rather than having it in the endpoint handler. WDYT?

fulmicoton:
I think we need 2 rest API endpoint.
- One compatible with elastic batch endpoint.
- One that just takes ndjson
We then need to store the data in rocksdb, one doc per record, in a way that is agnostic of the rest ingest endpoint. 

We then need a single source.

evanxg852000:
Currently, I am making push service (Mailbox<PushApiService>) a member of source.
```rust
pub struct PushApiSource {
    push_api_mailbox: Mailbox<PushApiService>,
   ...
}
```

Making it not possible to get an instance of `Mailbox<PushApiService>`  inside  
```rust 
impl TypedSourceFactory for PushApiSource {
      async fn typed_create_source(params: PushApiSourceParams, checkpoint: SourceCheckpoint,) {
           .... 
      }
```

@fulmicoton couple of solutions:
- Make the push_api_mailbox globally available via a function 
- Not support constructing the source via config (I don't think end-user needs to see/configure this). Do not implement `TypedSourceFactory` at all if that is the case
- Accept the Queues path and Open a Readable RockDB instance from PushApiSource, (very clunky)

I was in favor of solution (2) but you said we might need it to be able to instantiate `PushApiSource` via Factory.

fulmicoton:
> I was in favor of solution (2) but you said we might need it to be able to instantiate PushApiSource via Factory.

Just to clarify. This is not me doing overengineering here. I think we do need a factory to spawn an indexing pipeline, so you might need right after this ticket. 
You can double check that statement by doing a find usage.

It should lead you to these lines.
https://github.com/quickwit-oss/quickwit/blob/main/quickwit-indexing/src/actors/indexing_pipeline.rs#L384-L386

Solution one seems to be the only one possible.

Actix does that as part of the framework, offering a register for a specific class of service actors (that can be instantiated without  extra params). Fetching an actor can then spawn them in a lazy manner if they are not there.
We might add in the future too.

For the moment, can you add two method in the push-api crate that returns such a singleton.

```rust
[...]

Issue #1229, opened 2022-03-28
Create Push APi Source on create index


fulmicoton:

evanxg852000:
@fulmicoton  in relation to the closed [task](https://github.com/quickwit-oss/quickwit/issues/1399), we will not have any determinism when index is deleted. We could just use the Gc actor to remove the queue when the index is gone. 
An issue thought is the metastore fetch error will not necessarily mean index is deleted, it can be other things

fulmicoton:
Suggestion (you are closer to the problem than me).

To do GC, we need to have a function that lists queues lists queues, list indexes (in that order) does the diff, terminates  existing pipelines and deletes the extra numerous queues.

For this version, maybe we can do that in a GC actor that would run once every hour.

No need to do something super great.

---

Tricky stuff to handle:
- what happens to a running pipeline?

PR #1263, opened 2022-04-06
PushAPI Source

evanxg852000:
Implements a PushAPI source

### How was this PR tested?
- unittests
- ingested wikipedia data set to S3

Closes https://github.com/quickwit-oss/quickwit/issues/1235

Adding a pushapi source
```bash
cargo run -- source add --index wikipedia --source pushapi --type pushapi --params "{ \"index_id\": \"wikipedia\" }"
```


quickwit-indexing/src/source/push_api_source.rs
```diff
@@ -0,0 +1,322 @@
...
+            .await
+            .map_err(|err| anyhow::anyhow!(err.to_string()))?;
+
+        let FetchResponse {
+            first_position: first_position_opt,
+            doc_batch: doc_batch_opt,
+        } = fetch_resp;
+
+        if first_position_opt.is_none() {
+            return Ok(());
```
evanxg852000, line 109: @fulmicoton 
should we return `ActorExitStatus::Success`? 
This will exit this SourceActor and we currently have no way of recreating the SourceActor when new docs arrive.

fulmicoton: This is related to #1229.

For the moment, can you write something there before finishing this ticket maybe?
My 50c is for this iteration we don't think to be simple. 

In other words, the actor does not return Success if it reaches the end.
It keeps polling for new message.

We need however a way to sleep for 1s or so.
This probably means tweaking the SourceActor a little bit.

One improvement suggested by @trinity-1686a was to have actors return a Duration to inform the SourceActor how long it should sleep.

In the future, we can be smarter and have the push API source actor sleep properly.

quickwit-config/src/source_config.rs
```diff
@@ -285,4 +297,17 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn test_push_api_source_params_serialization() {
+        {
```
fulmicoton, line 303: can we remove the extra `{}`

fulmicoton, line 303:
```suggestion
```

quickwit-config/src/source_config.rs
```diff
@@ -285,4 +297,17 @@ mod tests {
...
+            let yaml = r#"
+                index_id: wikipedia
+                batch_num_bytes_threshold: 200000
+            "#;
+            let push_api_params = serde_yaml::from_str::<PushApiSourceParams>(yaml).unwrap();
+            assert_eq!(push_api_params.index_id, "wikipedia");
+            assert_eq!(push_api_params.batch_num_bytes_threshold, Some(200000))
+        }
```
fulmicoton, line 311:
```suggestion
```

quickwit-indexing/src/source/mod.rs
```diff
@@ -216,7 +222,12 @@ impl Handler<Loop> for SourceActor {
     type Reply = ();
 
     async fn handle(&mut self, _message: Loop, ctx: &SourceContext) -> Result<(), ActorExitStatus> {
-        self.source.emit_batches(&self.batch_sink, ctx).await?;
+        let wait_for_opt = self.source.emit_batches(&self.batch_sink, ctx).await?;
+        if let Some(wait_for) = wait_for_opt {
+            // wait_for time can be more than a HEARTBEAT timeout.
+            ctx.protect_zone();
+            tokio::time::sleep(wait_for).await;
```
fulmicoton, line 229: could we use the self_schedule message instead? (as a replacement to  the send_self_message).

There are several benefit:
- simulating time in tests
- be able to send commands to the source
- it will help us profiling as the actor will be properly idle when it waits.

fulmicoton: Also can we return `Duration` and rely on `Duration` being 0 instead of `None`.

fulmicoton (changes requested):
See inline comments.

quickwit-indexing/src/source/file_source.rs
```diff
@@ -104,7 +105,7 @@ impl Source for FileSource {
             ctx.send_exit_with_success(batch_sink).await?;
             return Err(ActorExitStatus::Success);
         }
-        Ok(())
+        Ok(None)
```
fulmicoton, line 108:
```suggestion
        Ok(Duration::default())
```

quickwit-indexing/src/source/push_api_source.rs
```diff
@@ -0,0 +1,355 @@
...
+        let partition_id = PartitionId::from(params.index_id.clone());
+        let offset = if let Some(Position::Offset(offset_str)) =
+            checkpoint.position_for_partition(&partition_id).cloned()
+        {
+            offset_str.parse::<u64>()?
+        } else {
+            0
+        };
+
+        ensure_queue_exist_for_index(&push_api_mailbox, params.index_id.clone()).await?;
```
fulmicoton, line 65: Why do we need to ensure that the queue exists here?

quickwit-indexing/src/source/push_api_source.rs
```diff
@@ -0,0 +1,355 @@
...
+            .await
+            .map_err(|err| anyhow::anyhow!(err.to_string()))?;
+
+        let FetchResponse {
+            first_position: first_position_opt,
+            doc_batch: doc_batch_opt,
+        } = fetch_resp;
+
+        if first_position_opt.is_none() {
+            // TODO: We are currently waiting for a constant time.
```
fulmicoton, line 135: can we comment about what this if-statement.

Is it what happens when the fetch response is empty and there no message available?

quickwit-indexing/src/source/push_api_source.rs
```diff
@@ -0,0 +1,355 @@
...
+            // Think of better way, maybe increment this (i.e wait longer) as time
+            // goes on without receiving docs.
+            return Ok(Some(Duration::from_secs(1)));
+        }
+
+        let doc_batch = doc_batch_opt.unwrap();
+        let first_position = first_position_opt.unwrap();
+        let docs = iter_doc_payloads(&doc_batch)
+            .map(|buff| String::from_utf8_lossy(buff).to_string())
```
fulmicoton, line 144: do we check for utf8 on the rest side, I forgot...?

quickwit-indexing/src/source/push_api_source.rs
```diff
@@ -0,0 +1,355 @@
...
+            .collect::<Vec<_>>();
+
+        if !docs.is_empty() {
```
fulmicoton, line 147: is it possible docs to be empty here?

quickwit-indexing/src/source/push_api_source.rs
```diff
@@ -0,0 +1,355 @@
...
+                    Position::from(self.counters.previous_offset),
+                    Position::from(self.counters.current_offset),
+                )
+                .unwrap();
+
+            let raw_doc_batch = RawDocBatch {
+                docs,
+                checkpoint_delta,
+            };
+            self.counters.previous_offset = self.counters.current_offset;
```
fulmicoton, line 165: Can we group the logic updating the counters in a separate method?

quickwit-pushapi/src/lib.rs
```diff
@@ -36,6 +37,26 @@ use tracing::info;
...
+
+/// Initializes the PushApiService single instance.
+pub fn init_push_api(
+    universe: &Universe,
+    queue_path: &Path,
+    metastore: Arc<dyn Metastore>,
+) -> anyhow::Result<Mailbox<PushApiService>> {
+    let push_api_service = PUSH_API_SERVICE_INSTANCE
+        .get_or_try_init(|| spawn_push_api_actor(universe, queue_path, metastore))
+        .map_err(|error| anyhow::anyhow!(error))?;
```
fulmicoton, line 50:
```suggestion
        .context("Failed to intiialize the PushApi")?;
```

fulmicoton: (An import will be required)

quickwit-serve/src/push_api/rest_handler.rs
```diff
@@ -154,7 +154,7 @@ async fn tail_endpoint(
 }
 
 fn bulk_filter() -> impl Filter<Extract = (String, String), Error = Rejection> + Clone {
-    warp::path!("api" / "v1" / String / "bulk")
```
fulmicoton, line 157: api/v1 is already mounted?

fulmicoton (changes requested):
See comments

quickwit-indexing/src/source/push_api_source.rs
evanxg852000: On the rest api we check from_utf8

quickwit-serve/src/push_api/rest_handler.rs
evanxg852000: Yes, a nice refactor from @fmassot

quickwit-indexing/src/source/push_api_source.rs
evanxg852000: No, not really. first_position_opt check is enough

quickwit-indexing/src/source/push_api_source.rs
evanxg852000: Maybe I am missing your point here. a portion of that update needs to be done before saving the Checkpoint. That is two lines above and this line. Should we then move all the checkpointing stuff with counters update into another method?

quickwit-indexing/src/source/push_api_source.rs
```diff
@@ -0,0 +1,387 @@
...
+        .ask_for_res(queue_exists_req)
+        .await
+        .map_err(|err| anyhow::anyhow!(err.to_string()))?
+    {
+        return Ok(());
+    }
+
+    let create_queue_req = CreateQueueRequest { queue_id: index_id };
+    push_api_mailbox
+        .ask_for_res(create_queue_req)
```
fulmicoton, line 120: there is a race condition here.

quickwit-indexing/src/source/push_api_source.rs
```diff
@@ -0,0 +1,387 @@
...
+        .await
+        .map_err(|err| anyhow::anyhow!(err.to_string()))
```
fulmicoton, line 122: why anyhow::anyhow!(err.to_string()) ?

Wouldn't
```
.await?;
Ok(())
```
(although the race condition is actually a problem, see above)

quickwit-indexing/src/source/push_api_source.rs
fulmicoton: I think it is better to not create the queue here.
I am not sure it is helping in any way, and it makes the code more error prone.

By leaving the creation to the push side, we end up creating it in only one place of the code, which makes your mental model simpler.
It avoid the current race condition: there exists a timing for which `ensure_queue_exists_for_index` returns an error, just because someone managed to push a message in the middle of this function call, hence creating the queue right before.

quickwit-indexing/src/source/push_api_source.rs
evanxg852000: Yes, you are right. I just tried to avoid using the error condition of CreateQueueRequest as success like you advise last time without thinking about the implication of message ordering.
Can I create a custom message for this operation? `EnsureQueueExistRequest` ?

quickwit-indexing/src/source/push_api_source.rs
evanxg852000: Ok

fulmicoton (approved):

Issue #1024, opened 2022-01-04
[DISCUSS] Index config update


guilload:
Currently, once an index is created, its config is immutable. We want to provide an API for updating an index config.

### CLI API
_Creating an index_
`quickwit index apply --id my-index --uri s3://my-indexes/my-index --config-file path/to/my/index/config.yaml`

_Updating an index config_
`quickwit index apply --index my-index --config-file path/to/my/index/config.yaml`

_Fetching an index config_
`quickwit index get --index my-index`

### REST API
TDB

### Metastore API
The following API already exists for creating an index:
```rust
metastore.create_index(&self, index_metadata: IndexMetadata) -> MetastoreResult<()>;
```

```rust
metastore.apply_index_config(&self, index_id: &str, index_config: IndexConfig) -> MetastoreResult<()>;
```

### Implementation
Ideally, the CLI and REST API both rely on the

### Out of scope:
- Index ID update
- Index URI update
- DocMapping update

guilload:
### CLI API alternative
_Creating an index_ (same as before)
`quickwit index create --id my-index --uri s3://my-indexes/my-index --config-file path/to/my/index/config.yaml`

_Updating an index config_
`quickwit index config apply --index my-index --config-file path/to/my/index/config.yaml`

_Fetching an index config_
`quickwit index config get --index my-index`

guilload:
See #1071 for a discussion regarding source creation and source config update.

rdettai:
Completed by https://github.com/quickwit-oss/quickwit/pull/5078.

Even updating doc mappings is now possible thanks to https://github.com/quickwit-oss/quickwit/pull/5253.

PR #1270, opened 2022-04-08
Rework command for adding a new source to an indexing

### LINKED ISSUE #1071: Add source CLI experience inconsistent with add an index [Papercut]
The current add source cli experience require to pass a source-id, source type, and source params independently in the cli.

In addition the source params has to be JSON.

This is quite inconsistent with the rest of our user experience:
- we use yaml everywehre for conf
- the index creation cli experience includes the index_id in the file that is passed.

I suggest we pass a single yaml file that deserializes to our `SourceConfig` object.
The only other CLI params will then be `--index`.

guilload:
### Description
The new command is `quickwit source create --index my-index --source-config my-source-config.yaml`. As we may revisit how me manage objects in a near future (#1024), I chose to leave the index out of the config so I did not have to write some custom parsing logic.

Fixes #1071

### How was this PR tested?
`make test-all`

### DIFF
--- a/quickwit-cli/src/source.rs
+++ b/quickwit-cli/src/source.rs
@@ -21,12 +21,12 @@ use anyhow::{bail, Context};
 use clap::{arg, ArgMatches, Command};
 use itertools::Itertools;
 use quickwit_common::uri::Uri;
-use quickwit_config::{SourceConfig, SourceParams};
+use quickwit_config::SourceConfig;
 use quickwit_indexing::check_source_connectivity;
 use quickwit_metastore::checkpoint::SourceCheckpoint;
 use quickwit_metastore::{quickwit_metastore_uri_resolver, IndexMetadata};
 use quickwit_storage::load_file;
-use serde_json::{Map, Value};
+use serde_json::Value;
 use tabled::{Table, Tabled};
 
 use crate::{load_quickwit_config, make_table};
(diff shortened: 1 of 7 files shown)


quickwit-common/src/uri.rs
```diff
@@ -28,6 +29,26 @@ const FILE_PROTOCOL: &str = "file";
 
 const PROTOCOL_SEPARATOR: &str = "://";
 
+#[derive(Debug, PartialEq)]
+pub enum Extension {
```
guilload, line 33: I got carried away trying to refactor and commonize the loading logic of `QuickwitConfig`, `IndexConfig`, and `SourceConfig`. Those are the remains of the aborted refactor. I thought that having a `.extension` method on `Uri` might be helpful in the future, but I would be happy to remove it if judged premature.

quickwit-cli/src/source.rs
```diff
@@ -35,22 +35,20 @@ pub fn build_source_command<'a>() -> Command<'a> {
...
+            Command::new("create")
+                .about("Adds a new source to an index.")
                 .args(&[
-                    arg!(--config <CONFIG> "Quickwit config file").env("QW_CONFIG"),
-                    arg!(--index <INDEX> "ID of the target index"),
-                    arg!(--params <PARAMS> "Parameters for the source formatted as a JSON object passed inline or via a file. Parameters are source-specific. Please, refer to the source's documentation for more details."),
-                    arg!(--source <SOURCE_ID> "ID of the source."),
-                    arg!(--type <TYPE> "Type of the source. Available types are: `file` and `kafka`."),
+                    arg!(--config <CONFIG> "Path to Quickwit config file").env("QW_CONFIG"),
+                    arg!(--index <INDEX_ID> "ID of the target index"),
```
fulmicoton, line 42: I thought you were heading for index_id integrated in the source file?

fulmicoton (approved):
I wasn't able to go too deep on how much of your refactoring was applied or useless.
I leave you judge to what to do there. If there is a follow-up needed, please create a ticket so we do not forget.

quickwit-cli/src/source.rs
guilload: As we may revisit how we manage objects in a near future (https://github.com/quickwit-oss/quickwit/issues/1024), I chose to leave the index out of the source config so I did not have to write some custom parsing logic.

quickwit-cli/src/source.rs
```diff
@@ -35,22 +35,20 @@ pub fn build_source_command<'a>() -> Command<'a> {
     Command::new("source")
         .about("Manages sources.")
         .subcommand(
-            Command::new("add")
-                .about("Adds a new source.")
+            Command::new("create")
+                .about("Adds a new source to an index.")
```
fmassot, line 39: "Creates"? :)

fmassot (approved):
LGTM

Issue #1007, opened 2021-12-31
Define and enforce naming convention for index ID, source ID, ...


guilload:
`[a-zA-Z0-9_-]{1,255}` ?

Following https://github.com/quickwit-oss/quickwit/pull/1272/files we need to make sure that source_id do not contain any of the kafka group_id unallowed characters, or fix the kafka source.

fmassot:
no dots?

guilload:
Periods are fine with me.

FYI, the field names regex is
`[_a-zA-Z][_\.\-a-zA-Z0-9]`.

Maybe we could use the same naming convention everywhere so that there's only one rule to remember. I also suggest we start with something restrictive first, that can revisit to be more lenient later.

PSeitz:
dots could conflict to address nested documents.

e.g. `index_config.search_settings.default_search_fields`

fulmicoton:
Is this already fixed?

PR #1272, opened 2022-04-08
Auto-generate Kafka group ID

### LINKED ISSUE #1069: Better error message if missing group.id in kafka source [Papercut]
If the `group.id`  parameter is missing, quickwit will fail adding the source with a non-helpful error.

The doc does not make it clear that this is a required parameter.

Also, I wonder it is not clear what this parameter is and why it is required. Could we set it to quickwit by default?
librdkafka suggest just passing a uuid for user like us.

guilload:
### Description
Auto-generate a group ID derived from the source ID for debugging purposes.

Fixes #1069.

### How was this PR tested?
`make test-all`

### DIFF
--- a/quickwit-indexing/src/source/kafka_source.rs
+++ b/quickwit-indexing/src/source/kafka_source.rs
@@ -28,6 +28,7 @@ use backoff::ExponentialBackoff;
 use futures::{StreamExt, TryFutureExt};
 use itertools::Itertools;
 use quickwit_actors::{ActorExitStatus, Mailbox};
+use quickwit_common::new_coolid;
 use quickwit_config::KafkaSourceParams;
 use quickwit_metastore::checkpoint::{CheckpointDelta, PartitionId, Position, SourceCheckpoint};
 use rdkafka::config::{ClientConfig, RDKafkaLogLevel};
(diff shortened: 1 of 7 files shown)


quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -292,18 +301,24 @@ fn previous_position_for_offset(offset: i64) -> Position {
...
 /// Creates a new `KafkaSourceConsumer`.
 fn create_consumer(
+    source_id: &str,
     client_log_level: Option<String>,
     client_params: serde_json::Value,
 ) -> anyhow::Result<Arc<KafkaSourceConsumer>> {
-    let log_level = parse_client_log_level(client_log_level)?;
     let mut client_config = parse_client_params(client_params)?;
+    // We assign partitions manually: we always want one consumer per consumer group.
+    let group_id = new_coolid(&format!("quickwit-{}", source_id));
```
fulmicoton, line 318: There is a restriction on group_id. Can you project the source id to that restriction?

`can be up to 255 characters in length, and can include the following characters: a-z, A-Z, 0-9, . (dot), _ (underscore), and - (dash).`

fulmicoton (approved):

fulmicoton (changes requested):
See comment inline

quickwit-indexing/src/source/kafka_source.rs
guilload: Good catch. I'm going to assume that #1007 will be implemented shortly and merely truncate the generated group ID to here 255 characters.

fulmicoton (approved):

fulmicoton (approved):

PR #1282, opened 2022-04-12
Refactoring the doc mapper.

fulmicoton:
The current implementation is relying on a struct with the union of all
of the field options, and some dynamic logic to ensure that the enterred
options are all there and correct.

Functionally, there are two modifications:
- it forbids fast fields for `array<bytes>` as tantivy does not support array<bytes> + fastfield.
- some of the error message are relying on `serde`,
and are therefore different. One class of error message have some
regression: if `tokenizer` is added to an int field, we detect and
return a valid error, but we do not mention that this option is only
valid for TextOptions any more.

In term of refactoring we also:
- move the field_mapping, and required fast fields out of the field mapping entry object, as there is no point to 
cache the earlier at all, and the latter are only useful at the default_doc_mapper level
- remove support for older version. They introduce complexity that is not useful.
- remove fast from text fields, as they are not supported yet.


quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -402,15 +428,15 @@ impl std::fmt::Debug for DefaultDocMapper {
...
 #[typetag::serde(name = "default")]
 impl DocMapper for DefaultDocMapper {
-    fn doc_from_json(&self, doc_json: String) -> Result<Document, DocParsingError> {
+    fn doc_from_json(&self, doc_json: &str) -> Result<Document, DocParsingError> {
         let mut document = Document::default();
-        let json_obj: JsonValue = serde_json::from_str(&doc_json).map_err(|_| {
+        let json_obj: JsonValue = serde_json::from_str(doc_json).map_err(|_| {
             // FIXME: the error contains some useful information (line, column, ...) that we could
             // surface to the user.
-            let doc_json_sample = format!("{:?}...", &doc_json[0..doc_json.len().min(20)]);
```
fmassot, line 410: This one yes...

fmassot: I don't remember this one though :/

fmassot:
@fulmicoton should I review it now? all tests are failing

fulmicoton:
@fmassot yes, I reverted one minor change late last night and it broke the test. It is ready for your review.

quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -138,16 +139,35 @@ impl DefaultDocMapperBuilder {
...
-        let field_mappings = FieldMappingEntry::root(FieldMappingType::Object(self.field_mappings));
+        let root_field_mapping =
+            FieldMappingEntry::root(FieldMappingType::Object(QuickwitObjectOptions {
+                field_mappings: self.field_mappings,
+            }));
+
+        let field_entries = root_field_mapping.compute_field_entries();
+        let required_fast_fields: Vec<FieldPath> = field_entries
+            .into_iter()
+            .filter(|(_, field_type)| match field_type {
```
PSeitz, line 150: the current tantivy master has `field_type.is_fast`, but this version is not yet used in quickwit

quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -354,15 +377,18 @@ pub struct DefaultDocMapper {
...
     /// Schema generated by the store source and field mappings parameters.
     #[serde(skip_serializing)]
     schema: Schema,
     /// List of field names used for tagging.
     pub tag_field_names: BTreeSet<String>,
     /// Demux field name.
     pub demux_field_name: Option<String>,
+    /// List of fast fields. This list is used to reject Document if they
+    /// are missing a fast field.
+    pub required_fast_fields: Vec<FieldPath>,
```
PSeitz, line 391: I don't understand this, why do we require fast fields to be set. fast fields can handle non-existent values

fulmicoton: Not in quickwit apparently.

quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
```diff
@@ -492,262 +465,223 @@ impl FieldPath {
...
+        Ok(FieldMappingEntry::new(value.name, mapping_type))
+    }
+}
 
-    fn int_options(&self) -> anyhow::Result<NumericOptions> {
-        self.check_no_text_options()?;
-        let mut options = NumericOptions::default();
-        if self.stored {
-            options = options.set_stored();
+fn extract_map_from_json(
```
PSeitz, line 643: Could be named closer to [as_object](https://docs.serde.rs/serde_json/value/enum.Value.html#method.as_object)

quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
fmassot: AFAIK, if there is a missing value, tantivy will put a default value in the fast field... 0 for integer/float. As I did not want a user to have unexpected results on aggregations, I rejected missing values.

fulmicoton: Generally speaking I prefer the spec in quickwit than the one in tantivy. I like the cardinality stuff too.

fmassot: Ok, I opened https://github.com/quickwit-oss/tantivy/issues/1338 on tantivy to track the problem of null values, I think this should be fixed on tantivy side.

quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
fmassot: oh nice!

quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
fulmicoton: Thanks!

PSeitz: > AFAIK, if there is a missing value, tantivy will put a default value in the fast field... 0 for integer/float. As I did not want a user to have unexpected results on aggregations, I rejected missing values.

That's a good point, I think we should improve tantivy here, so the user doesn't need to provide a value for null. We can e.g. compute it or handle it with a bitvec

fmassot (approved):
Really cool refactoring. I did not check all cases... but I think test coverage is nice and you spotted the bytes array thing.

quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -138,16 +139,35 @@ impl DefaultDocMapperBuilder {
 
         // Build the root mapping entry, it has an empty name so that we don't prefix all
         // field name with it.
```
saroh, line 141:
```suggestion
```
I feel this would be a more interesting info in the class doc than here where we're a click away from the implem

quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -354,15 +377,18 @@ pub struct DefaultDocMapper {
     pub timestamp_field_name: Option<String>,
     /// Sort field name and order.
     pub sort_by: SortBy,
-    /// List of field mappings which defines how a json field is mapped to index fields.
-    pub field_mappings: FieldMappingEntry,
+    /// Root field mapping.
```
saroh, line 380:
```suggestion
    /// Root field mapping. It has an empty name so that we don't prefix all field names with it. See [FieldMappingEntry::root]
```
unsure if the link works :D

PR #1284, opened 2022-04-12
Add new cluster_id config option

### LINKED ISSUE #1283: Update scuttlebutt version and add configurable cluster_name
In https://github.com/quickwit-oss/scuttlebutt/pull/36 we added a cluster_name to scuttlebutt.

This is meant to avoid having two clusters merge due to a misconfiguration or k8s IP recycling.
We need to update scuttlebutt, and set the cluster_name properly (probably add that in the quickwit_config.)

xvello:
Fixes #1283

### Description
- Updates scuttlebutt to use https://github.com/quickwit-oss/scuttlebutt/pull/36
- Add the optional `cluster_name` config param, with a default value of "Test cluster"
- Add a new unit test to confirm that four nodes with different cluster names settle correctly into 2x2

### How was this PR tested?

Only unit test for now, we should test it a bit more manually before the v0.3 release

### Additional notes

- We should expose the cluster name in the UI too
- Do we want to print a warn log if using the default value? I would, but I'm not sure whether that's in line with the project's logging policy.

### DIFF
--- a/docs/reference/quickwit-config.md
+++ b/docs/reference/quickwit-config.md
@@ -17,6 +17,7 @@ A commented example is accessible here: [quickwit.yaml](https://github.com/quick
 | --- | --- | --- |
 | version | Config file version. 0 is the only available value. |  |
 | node_id | Node ID of the instance (searcher or indexer). It must be unique in your cluster. If not set, a random ID is generated at each boot. |  |
+| cluster_name | Name for the cluster this node will be joining. Should be set to a unique name to ensure clusters do not accidentally merge together. | "Test cluster" |
 | listen_address | The IP address or hostname that Quickwit service binds to for starting REST and GRPC server and connecting this node to other nodes. By default, Quickwit binds itself to 127.0.0.1 (localhost). This default is not valid when trying to form a cluster. | 127.0.0.1 |
 | rest_listen_port | The port which to listen for HTTP REST API. | 7280 |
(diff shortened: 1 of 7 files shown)


docs/reference/quickwit-config.md
```diff
@@ -17,6 +17,7 @@ A commented example is accessible here: [quickwit.yaml](https://github.com/quick
 | --- | --- | --- |
 | version | Config file version. 0 is the only available value. |  |
 | node_id | Node ID of the instance (searcher or indexer). It must be unique in your cluster. If not set, a random ID is generated at each boot. |  |
+| cluster_name | Name for the cluster this node will be joining. Should be set to a unique name to ensure clusters do not accidentally merge together. | "Test cluster" |
```
guilload, line 20: I wonder whether we should rename cluster_name to cluster_id to remain consistent with `index_id`, `node_id`, and `source_id`.

We probably also want to enforce the same naming convention (see #1007). This does not have to be tackled in this PR though.

quickwit-config/src/config.rs
```diff
@@ -50,6 +50,10 @@ fn default_metastore_and_index_root_uri(data_dir_path: &Path) -> String {
         .to_string()
 }
 
+fn default_cluster_name() -> String {
+    "Test cluster".to_string()
```
fulmicoton, line 54:
```suggestion
    "quickwit-test-cluster".to_string()
```

docs/reference/quickwit-config.md
fulmicoton: ah yes. can we make it `cluster_id` everywhere? .

fulmicoton (changes requested):
see comments inline

PR #1295, opened 2022-04-14
Add `quickwit split mark-for-deletion` command

guilload:
### Description
Add `quickwit split mark` command

### How was this PR tested?
Add unit test

### DIFF
--- a/quickwit-cli/src/split.rs
+++ b/quickwit-cli/src/split.rs
@@ -42,7 +42,7 @@ pub fn build_split_command<'a>() -> Command<'a> {
         .about("Operations (list, add, delete, describe...) on splits.")
         .subcommand(
             Command::new("list")
-                .about("List the splits of an index.")
+                .about("Lists the splits of an index.")
                 .args(&[
                     arg!(--config <CONFIG> "Quickwit config file").env("QW_CONFIG"),
                     arg!(--index <INDEX> "ID of the target index"),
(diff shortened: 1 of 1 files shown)


quickwit-cli/src/split.rs
```diff
@@ -89,6 +89,18 @@ pub fn build_split_command<'a>() -> Command<'a> {
                         .required(false),
                 ])
             )
+        .subcommand(
+            Command::new("delete")
+                .about("Removes one or multiple splits from an index.")
+                .args(&[
+                    arg!(--config <CONFIG> "Quickwit config file").env("QW_CONFIG"),
+                    arg!(--index <INDEX> "ID of the target index"),
+                    arg!(--splits <SPLITS> "Comma-separated list of split IDs.")
```
guilload, line 98: Really, this should be `<SPLIT IDs>` but clap is not happy with whitespace here.

quickwit-cli/src/split.rs
```diff
@@ -291,6 +336,25 @@ async fn list_split_cli(args: ListSplitArgs) -> anyhow::Result<()> {
...
+        .resolve(&quickwit_config.metastore_uri())
+        .await?;
+    let split_ids: Vec<&str> = args
+        .split_ids
+        .iter()
+        .map(|split_id| split_id.as_ref())
+        .collect();
+    metastore
+        .mark_splits_for_deletion(&args.index_id, &split_ids)
+        .await?;
```
guilload, line 354: Note that we don't run GC here, so "delete" here is misleading. Some options:
- rename the option `--mark-as-deleted`
- run the GC
- run the GC and add a `--no-gc` option
- your awesome idea here :)

Let me know what you think and I'll make the appropriate changes.

guilload:
PTAL

quickwit-cli/src/split.rs
guilload: I'm going to rename the command `mark-for-deletion` and provide a `mark` alias. The gc will delete the splits later. One can always run the gc manually with `quickwit index gc`  and a short grace period if necessary.

fmassot (approved):

Issue #1290, opened 2022-04-13
Document parsing is needlessly slow.


fulmicoton:
There is a significant indexing performance gain that we could get from optimizing indexing.
```
simple-json-to-doc/simple-json-to-doc-donothing                                                                             
                        time:   [47.552 us 47.697 us 47.845 us]
                        thrpt:  [1.7065 GiB/s 1.7118 GiB/s 1.7171 GiB/s]
                 change:
                        time:   [+2.8327% +3.6207% +4.2879%] (p = 0.00 < 0.05)
                        thrpt:  [-4.1116% -3.4942% -2.7546%]
                        Performance has regressed.
Benchmarking simple-json-to-doc/simple-json-to-doc: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 7.9s, enable flat sampling, or reduce sample count to 50.
simple-json-to-doc/simple-json-to-doc                                                                             
                        time:   [1.5471 ms 1.5501 ms 1.5539 ms]
                        thrpt:  [53.807 MiB/s 53.939 MiB/s 54.044 MiB/s]
                 change:
                        time:   [-1.6838% -1.5192% -1.3615%] (p = 0.00 < 0.05)
                        thrpt:  [+1.3803% +1.5427% +1.7127%]
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  2 (2.00%) low mild
  2 (2.00%) high mild
  5 (5.00%) high severe
simple-json-to-doc/simple-json-to-doc-tantivy                                                                            
                        time:   [713.68 us 714.82 us 715.92 us]
                        thrpt:  [116.79 MiB/s 116.97 MiB/s 117.15 MiB/s]
                 change:
                        time:   [+1.5761% +1.8180% +2.0980%] (p = 0.00 < 0.05)
                        thrpt:  [-2.0549% -1.7855% -1.5517%]
                        Performance has regressed.
```

Two observation:
- Document parsing is in the same realm as indexing (20MB/s -- 60MB/s).
- Tantivy's document parser does not use zero-copy either and is more than 2x faster.

We can probably do much better.

One low hanging fruit is removing the needless step

`json -> Vec<(FieldPath, Value)> -> Vec<(Field, Value)>`

The resolution `FieldPath -> Field` is not really useful.
We could for instance hold a `Field` member within the FieldMappingEntry object (or a parallel structure).

[...]

PR #1303, opened 2022-04-18
Added JSON support

### LINKED ISSUE #1234: Add Schemaless/ JSON support.
We want support for schemaless. 

The plan is to rely on tantivy's JSON field, in two ways:
- [x] json type.
- [ ] a catchall json field that captures all of the unmatched fields, if dynamic is set to true.

fulmicoton:
Dynamic mapping in quickwit.
    
    - Adds the JSON field type. This feature relies heavily on tantivy#1270.
    - More unit test for deserialization of the doc mapper.
    - Introduce the notion of mode in the default_mapper to decide how
      unmapped fields should be handled. (lenient, strict, dynamic)
      - In dynamic mode, unmapped fields are stored in a _dynamic field.
      - The DefaultDocMapper is refactored, yielding a substantial
      performance improvement in indexing when the data contains many short
      or untokenized fields. (Closes Document parsing is needlessly slow. #1290)
      - field_mapping can be absent of the index conf. (Closes Allow missing field mappings in conf #1324)
      - Error on deserializaton for unknown field in the default doc mapper.
      - When an error is spotted on doc parsing, the error contains the entire field path from the root instead of just the
      faulty leaf field name.
    
    Closes #1234
    
    Json output will be addressed in a separate ticket. #1325


quickwit-config/src/index_config.rs
```diff
@@ -306,17 +311,21 @@ pub fn build_doc_mapper(
     search_settings: &SearchSettings,
     indexing_settings: &IndexingSettings,
 ) -> anyhow::Result<Arc<dyn DocMapper>> {
-    let mut builder = DefaultDocMapperBuilder::new();
```
fulmicoton, line 309: That change may seem weird... 

The way we build DefaultDocMapper is actually quite unexpected.
We both build it through deserializing and through a `DocMapping` object.

For this reason it is quite easy to forget modify both location if we add a field for instance.

Relying on struct initialization makes this kind of mistakes  harder.

fmassot: Thanks for cleaning that, I think it comes from the 0.2 hurry release.

quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -369,42 +83,39 @@ impl From<DefaultDocMapper> for DefaultDocMapperBuilder {
 #[derive(Serialize, Deserialize, Clone)]
 #[serde(try_from = "DefaultDocMapperBuilder", into = "DefaultDocMapperBuilder")]
 pub struct DefaultDocMapper {
-    /// Store the json source in a text field _source.
-    pub store_source: bool,
+    /// Field in which the source should be stored.
+    pub(crate) source_field: Option<Field>,
```
fmassot, line 87: I was a bit surprised to see a tantivy `Field` as usually we don't want to store it directly, it's only relevant to the `DefaultDocMapper` schema and not for an external schema. I would avoid having putting a `pub` here and enrich a bit the docs `/// Field in which the source should be stored, can only be used with a `DefaultDocMapper` schema.`

quickwit-doc-mapper/src/default_doc_mapper/mod.rs
```diff
@@ -18,28 +18,32 @@
...
 
 /// Validates a field mapping name.
 /// Returns `Ok(())` if the name can be used for a field mapping. Does not check for reserved field
 /// mapping names such as `_source`.
 ///
 /// A field mapping name:
 /// - may only contain uppercase and lowercase ASCII letters `[a-zA-Z]`, digits `[0-9]`, hyphens
-///   `-`, periods `.`, and underscores `_`;
-/// - must start with an uppercase or lowercase ASCII letter `[a-zA-Z]`, or an underscore `_`;
+///   `-`, and underscores `_`;
```
fmassot, line 44: I'm ok to ban field names that start with `_`. I'm expecting to see users with fields starting `_` to be happy with json field.

quickwit-doc-mapper/src/doc_mapper.rs
```diff
@@ -133,4 +137,103 @@ mod tests {
...
+            sort_by_field: None,
+            aggregation_request: None,
+        };
+        let query = doc_mapper.query(schema, &search_request).unwrap();
+        assert_eq!(
+            format!("{:?}", query),
+            r#"BooleanQuery { subqueries: [(Should, TermQuery(Term(type=Json, field=0, path=toto, vtype=U64, 5))), (Should, TermQuery(Term(type=Json, field=0, path=toto, vtype=Str, "5")))] }"#
+        );
+    }
 }
```
fmassot, line 239: Currently, we can query a text field `body` with the query `body.wrong_subpath:hello`. 

Here is the test that is passing:

```rust
#[test]
    fn test_doc_mapper_accept_sub_field_query_on_non_json_field() {
        let mut doc_mapper_builder = DefaultDocMapperBuilder::default();
        doc_mapper_builder.field_mappings.push(FieldMappingEntry {
            name: "json_field".to_string(),
            mapping_type: FieldMappingType::Json(
                QuickwitJsonOptions::default(),
                Cardinality::SingleValue,
            ),
        });
        doc_mapper_builder.field_mappings.push(FieldMappingEntry {
            name: "body".to_string(),
            mapping_type: FieldMappingType::Text(
                QuickwitTextOptions::default(),
                Cardinality::SingleValue,
            ),
        });
        
        let doc_mapper = doc_mapper_builder.build().unwrap();
        let schema = doc_mapper.schema();
            index_id: "quickwit-index".to_string(),
            query: "body.wrong_field:hello".to_string(),
            search_fields: vec![],
            start_timestamp: None,
            end_timestamp: None,
            max_hits: 10,
            start_offset: 0,
            sort_order: None,
            sort_by_field: None,
            aggregation_request: None,
        };
        let query = doc_mapper.query(schema, &search_request).unwrap();
        assert_eq!(
            format!("{:?}", query),
            r#"TermQuery(Term(type=Json, field=1, vtype=Str, "hello"))"#
        );
    }
```

quickwit-doc-mapper/src/error.rs
```diff
@@ -17,6 +17,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+use tantivy::schema::DocParsingError as TantivyDocParsingError;
 use thiserror::Error;
 
 // TODO improve me and my error messages :)
```
fmassot, line 23: Either ditch this old comment or let's create an issue so that we can track it.

fmassot: (I will create the issue related to query parsing error I think)

fmassot (approved):
Great work. I saw some typos in the documentation, I will add a commit to yours in a couple of hours or if it's already merged I will open a PR on that.

quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
fulmicoton: Good point.
I added the comment and made everything private.

I also opened #1340 to see if we can simplify the weird DocMapper contract.

PR #1306, opened 2022-04-18
Logrotate source

### LINKED ISSUE #1159: Investigate: File as a source
We already have a file source but right now it is only good as a way to index a bunch of static files.

It would be great however to have a source that handles log rotation.

For instance, if a system does log rotation in a way that keeps files in order,
we could use whatever we want as a partition id and use 
`lastindexed_filename @ offset in file` as a position.

This ticket requires first investigating a bunch of system doing log rotation, seeing if there is a common denominator
that might be used to flesh out of a possible specification and its configuration.

trinity-1686a:
### Description
This pr adds a new source to fix #1159. There is no support for compressed files, and it requires logrotate to keep files in lexicographic order, which is not the default configuration for logrotate, but can be enabled by setting `dateext`. It's the default for some other systems like log4j RollingFile, provided no more than 10 file are generated over a give period, usually a day (after 10 files, files are no longer in lexicographic order as 1 < 10 < 2, 10 won't be processed if 2 is already processed itself).

Todo: add some documentation to docs/reference/source-config.md

### How was this PR tested?
This pr was tested by simulating log rotation, and with unit tests.

<details><summary>simulation description</summary>

```bash
export PATH="$PATH":../quickwit/target/debug
export QW_CONFIG=quickwit.yaml
export RUST_LOG=info

rm -r qwdata/
mkdir qwdata

quickwit index create --index-config ./wikipedia_index_config.yaml # 

quickwit run --config=quickwit.yaml

# add entries to files with echo >>, and wait for quickwit_indexing::actors::indexing_server::IndexingServer
# to say in logs it indexed something, then verify things where indexed properly with
quickwit source describe --index wikipedia --source my-source-id
quickwit index search --index wikipedia --config=quickwit.yaml --query "teststring"
```

wikipedia_index_config.yaml:
```yml
#
# Index config file for wikipedia dataset.
#

version: 0

index_id: wikipedia

doc_mapping:
  field_mappings:
    - name: title
      type: text
      tokenizer: default
      record: position
      stored: true
    - name: body
      type: text
      tokenizer: default
      record: position
      stored: true
    - name: url
      type: text
      tokenizer: raw

search_settings:
  default_search_fields: [title, body]

sources:
  - source_id: my-source-id
    source_type: logrotate
    params:
      current_file: /tmp/datasets/log/latest.json
      name_pattern: rotated.*.json
```

</details>

### DIFF
--- a/quickwit-config/src/source_config.rs
+++ b/quickwit-config/src/source_config.rs
@@ -96,7 +96,7 @@ impl SourceConfig {
                 }
                 Ok(())
             }
-            SourceParams::Kafka(_) | SourceParams::Kinesis(_) => {
+            SourceParams::Kafka(_) | SourceParams::Kinesis(_) | SourceParams::LogRotate(_) => {
                 // TODO consider any validation opportunity
                 Ok(())
             }
@@ -109,6 +109,7 @@ impl SourceConfig {
             SourceParams::File(_) => "file",
             SourceParams::Kafka(_) => "kafka",
             SourceParams::Kinesis(_) => "kinesis",
+            SourceParams::LogRotate(_) => "logrotate",
             SourceParams::Vec(_) => "vec",
             SourceParams::Void(_) => "void",
             SourceParams::PushApi(_) => "pushapi",
@@ -121,6 +122,7 @@ impl SourceConfig {
             SourceParams::File(params) => serde_json::to_value(params),
             SourceParams::Kafka(params) => serde_json::to_value(params),
             SourceParams::Kinesis(params) => serde_json::to_value(params),
+            SourceParams::LogRotate(params) => serde_json::to_value(params),
             SourceParams::Vec(params) => serde_json::to_value(params),
             SourceParams::Void(params) => serde_json::to_value(params),
             SourceParams::PushApi(params) => serde_json::to_value(params),
@@ -138,6 +140,8 @@ pub enum SourceParams {
     Kafka(KafkaSourceParams),
     #[serde(rename = "kinesis")]
     Kinesis(KinesisSourceParams),
+    #[serde(rename = "logrotate")]
+    LogRotate(LogRotateSourceParams),
     #[serde(rename = "vec")]
     Vec(VecSourceParams),
     #[serde(rename = "void")]
@@ -166,12 +170,12 @@ pub struct FileSourceParams {
     /// Path of the file to read. Assume stdin if None.
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(default)]
-    #[serde(deserialize_with = "absolute_filepath_from_str")]
+    #[serde(deserialize_with = "absolute_filepath_from_str_opt")]
     pub filepath: Option<PathBuf>, //< If None read from stdin.
 }
 
 // Deserializing a filepath string into an absolute filepath.
-fn absolute_filepath_from_str<'de, D>(deserializer: D) -> Result<Option<PathBuf>, D::Error>
+fn absolute_filepath_from_str_opt<'de, D>(deserializer: D) -> Result<Option<PathBuf>, D::Error>
 where D: Deserializer<'de> {
     let filepath_opt: Option<String> = Deserialize::deserialize(deserializer)?;
     if let Some(filepath) = filepath_opt {
@@ -180,6 +184,12 @@ where D: Deserializer<'de> {
     } else {
         Ok(None)
     }
+}
+
+fn absolute_filepath_from_str<'de, D>(deserializer: D) -> Result<PathBuf, D::Error>
+where D: Deserializer<'de> {
+    absolute_filepath_from_str_opt(deserializer)
+        .and_then(|path| path.ok_or_else(|| D::Error::custom("missing file path")))
 }
 
 impl FileSourceParams {
@@ -252,6 +262,37 @@ impl TryFrom<KinesisSourceParamsInner> for KinesisSourceParams {
             region_or_endpoint,
             shutdown_at_stream_eof: value.shutdown_at_stream_eof,
         })
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct LogRotateSourceParams {
+    /// Path of the current log file.
+    #[serde(deserialize_with = "absolute_filepath_from_str")]
+    pub current_file: PathBuf,
+    /// Path of the directory where rotated logs are stored. If None, assumes it's
+    /// the directory where `latest_file` is.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
+    #[serde(deserialize_with = "absolute_filepath_from_str_opt")]
+    archive_dir: Option<PathBuf>,
+    /// glob pattern of logs that already got rotated
+    pub name_pattern: String,
+}
+
+impl LogRotateSourceParams {
+    pub fn archive_dir(&self) -> PathBuf {
+        if let Some(dir) = &self.archive_dir {
+            dir.clone()
+        } else {
+            let mut dir = self.current_file.clone();
+            dir.pop();
+            dir
+        }
+    }
+    pub fn set_archive_dir(&mut self, dir: PathBuf) {
+        self.archive_dir = Some(dir);
     }
 }
 
--- a/quickwit-config/src/source_config.rs
+++ b/quickwit-config/src/source_config.rs
@@ -170,12 +170,12 @@ pub struct FileSourceParams {
     /// Path of the file to read. Assume stdin if None.
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(default)]
-    #[serde(deserialize_with = "absolute_filepath_from_str_opt")]
+    #[serde(deserialize_with = "absolute_filepath_opt_from_str")]
     pub filepath: Option<PathBuf>, //< If None read from stdin.
 }
 
 // Deserializing a filepath string into an absolute filepath.
-fn absolute_filepath_from_str_opt<'de, D>(deserializer: D) -> Result<Option<PathBuf>, D::Error>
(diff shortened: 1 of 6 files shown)


quickwit-config/src/source_config.rs
```diff
@@ -255,6 +265,37 @@ impl TryFrom<KinesisSourceParamsInner> for KinesisSourceParams {
...
+    /// glob pattern of logs that already got rotated
+    pub name_pattern: String,
+}
+
+impl LogRotateSourceParams {
+    pub fn archive_dir(&self) -> PathBuf {
+        if let Some(dir) = &self.archive_dir {
+            dir.clone()
+        } else {
+            let mut dir = self.current_file.clone();
```
fulmicoton, line 289: [file_name](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.file_name)
does that I think?

quickwit-config/src/source_config.rs
```diff
@@ -166,12 +170,12 @@ pub struct FileSourceParams {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(default)]
-    #[serde(deserialize_with = "absolute_filepath_from_str")]
+    #[serde(deserialize_with = "absolute_filepath_from_str_opt")]
     pub filepath: Option<PathBuf>, //< If None read from stdin.
 }
 
 // Deserializing a filepath string into an absolute filepath.
-fn absolute_filepath_from_str<'de, D>(deserializer: D) -> Result<Option<PathBuf>, D::Error>
+fn absolute_filepath_from_str_opt<'de, D>(deserializer: D) -> Result<Option<PathBuf>, D::Error>
```
fulmicoton, line 178:
```suggestion
fn absolute_filepath_opt_from_str<'de, D>(deserializer: D) -> Result<Option<PathBuf>, D::Error>
```

quickwit-config/src/source_config.rs
```diff
@@ -255,6 +265,37 @@ impl TryFrom<KinesisSourceParamsInner> for KinesisSourceParams {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct LogRotateSourceParams {
+    /// Path of the current log file.
+    #[serde(deserialize_with = "absolute_filepath_from_str")]
+    pub current_file: PathBuf,
```
fulmicoton, line 273: So users are required to pass a `current_file`, and may or may not pass an `archive_dir`?

What happens if `archive_dir` does not contain the file?

quickwit-config/src/source_config.rs
```diff
@@ -255,6 +265,37 @@ impl TryFrom<KinesisSourceParamsInner> for KinesisSourceParams {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct LogRotateSourceParams {
+    /// Path of the current log file.
+    #[serde(deserialize_with = "absolute_filepath_from_str")]
```
fulmicoton, line 272: `current_file` is required?

fulmicoton: Ah I see, current_file is the file being written. "latest" for instance.

quickwit-indexing/src/source/logrotate_source.rs
```diff
@@ -0,0 +1,984 @@
...
+        while read < BATCH_NUM_BYTES_THRESHOLD {
+            let mut doc_line = String::new();
+            let num_bytes = self
+                .reader
+                .read_line(&mut doc_line)
+                .await
+                .map_err(|io_err: io::Error| anyhow::anyhow!(io_err))?;
+            if num_bytes == 0 {
+                let should_read = self.position.next_file().await?;
+                if should_read {
```
fulmicoton, line 419: This code can be simplified using early exit style I think...

quickwit-indexing/src/source/logrotate_source.rs
```diff
@@ -0,0 +1,984 @@
...
+
+/// Cut a new batch as soon as we have read BATCH_NUM_BYTES_THRESHOLD.
+const BATCH_NUM_BYTES_THRESHOLD: u64 = 500_000u64;
+
+/// Default name when no log file has been rotated yet. It must be before any name
+/// any log file may have. This works under the assumption only printable name will
+/// be used. Alternatively we could use \0 but it may mess with some debug print.
+const START_FILE: &str = "        ";
+
+/// Time to sleep when there is nothing to do.
```
fulmicoton, line 51: Can we make "nothing to do" more explicit?
`when we reach the end of the current file and ...`

quickwit-indexing/src/source/logrotate_source.rs
```diff
@@ -0,0 +1,984 @@
...
+/// Try to find the next file to read from.
+/// Returns Ok(Left(filename)) if a filename was found, Ok(Right(true)) if no file was found,
+/// and it's safe to read from latest_log_file, or Ok(Right(false)) when no file was found and
+/// reading from latest_log_file could cause skipping some files.
+async fn find_next_file(
+    dir: &Path,
+    current_file: &str,
+    pattern: &GlobPattern,
+    latest_log_file: &Path,
+) -> Result<Either<String, bool>, io::Error> {
```
fulmicoton, line 279: can we make that Either an explicit enum?

quickwit-indexing/src/source/logrotate_source.rs
```diff
@@ -0,0 +1,984 @@
...
+/// be used. Alternatively we could use \0 but it may mess with some debug print.
+const START_FILE: &str = "        ";
+
+/// Time to sleep when there is nothing to do.
+const SLEEP_DURATION: Duration = Duration::from_secs(5);
+
+#[derive(Default, Clone, Debug, Eq, PartialEq, Serialize)]
+pub struct LogRotatePosition {
+    pub filename: String,
+    pub is_next: bool,
```
fulmicoton, line 57: Can you add some doc to explain what `is_next` semantically means?
I'm not sure the name is good too.

quickwit-indexing/src/source/logrotate_source.rs
```diff
@@ -0,0 +1,984 @@
...
+                // fine as no new data has arrived yet.
+                Ok(false)
+            }
+        }
+    }
+
+    /// If the position has is_next set, try to resolve the name of the file if it's now available.
+    /// Returns Ok(true) if the file transitioned from having an unknown path to a known path.
+    async fn resolve_next(&mut self) -> Result<bool, anyhow::Error> {
+        if !self.is_next {
```
fulmicoton, line 208: I am not sure I understand, but I suspect it is more a `assert` situation here.

quickwit-indexing/src/source/logrotate_source.rs
```diff
@@ -0,0 +1,984 @@
...
+            continue;
+        };
+        if filename == current_file {
+            found_current_file = true;
+        }
+        if pattern.matches(&filename) {
+            found_any_log = true;
+            if filename.as_str() > current_file {
+                // this is a log file which is newer than current file
+                let len = entry.metadata().await?.len();
```
fulmicoton, line 302: If I understand correctly, we identify the rotation by globlisting the directory, and picking the first file (lexicographically) that is above our last processed files?

Was inspecting `creation_time` an option too?

Can we add explanation about that somewhere?

fulmicoton (changes requested):
Great job. We really need to replace the Either thing with a proper enum though.

see comments inline

quickwit-config/src/source_config.rs
trinity-1686a: it should work, reading from `archive_dir` when there are things to read there, and reading `current_file` when there is no longer anything new in `archive_dir`, and going back to `archive_dir` rotation occur. I should probably add a UT to make sure it works as intended. There could be issues however if `archive_dir` is unset, and `name_pattern` would match the filename of `current_file`, which needs to be documented or dealt with

quickwit-indexing/src/source/logrotate_source.rs
trinity-1686a: I'm generally reluctant to use filesystem time, I find it often leads to broken assumptions. As a case that would definitely break (but is not supported yet anyway) when logrotate compresses older log files, it creates a new files, reading `creation_time` could cause an old log file getting compressed to be considered as newer than a recent (uncompressed) log file.
There would also be issue with what the checkpoints should then look like (as it would relax the requirement on lexicographic ordering of file names). Replacing the filename with creation time seems rather fragile

quickwit-config/src/source_config.rs
trinity-1686a: actually that seems to be the opposite, but [`parent`](https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#method.parent) is what I should have used

quickwit-indexing/src/source/logrotate_source.rs
trinity-1686a: the name isn't very good, but I don't know what a good name would be

quickwit-indexing/src/source/logrotate_source.rs
fulmicoton: > reading creation_time could cause an old log file getting compressed to be considered as newer than a recent (uncompressed) log file.

That's not what I meant.
My point is that we could us `creation_time` to identify that the current file has been moved to another file.

```
for file in list_dir() {
   if file.creation_time == current_tail_file_creation_time {
      // we rotated
   }
}
```

I agree all solutions are muddy though. Let's stick to yours!

quickwit-indexing/src/source/logrotate_source.rs
fulmicoton: Naming is hard.
A method I like to find decent names for stuff that are semantically difficult like this one is to force yourself to write a doc about it.

Sometimes it even helps spotting an obvious refactoring that simplify the semantics.

PR #1316, opened 2022-04-20
handle dict preloading for agg

PSeitz:
handle dict preloading for agg
handle cardinality for fields


quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -147,14 +147,7 @@ impl DefaultDocMapperBuilder {
         let field_entries = root_field_mapping.compute_field_entries();
         let required_fast_fields: Vec<FieldPath> = field_entries
             .into_iter()
-            .filter(|(_, field_type)| match field_type {
-                FieldType::U64(options)
-                | FieldType::I64(options)
-                | FieldType::F64(options)
-                | FieldType::Date(options) => options.is_fast(),
-                FieldType::Bytes(option) => option.is_fast(),
-                _ => false,
```
fulmicoton, line 156: you mentionned there was a bug in tantiv? Is it solved?

quickwit-search/src/leaf.rs
```diff
@@ -121,23 +125,83 @@ pub(crate) async fn open_index(
 ///
 /// The downloaded data depends on the query (which term's posting list is required,
 /// are position required too), and the collector.
+///
+/// * `query` - query is used to extract the terms and the fields which will be loaded from the
```
fulmicoton, line 129: in which case do we load a field due to the query? what does  a field mean here? Is it a fast field?

quickwit-search/src/leaf.rs
```diff
@@ -121,23 +125,83 @@ pub(crate) async fn open_index(
...
+                let dict = inverted_index.terms();
+                dict.warm_up_dictionary().await
+            });
+        }
+    }
+    try_join_all(warm_up_futures).await?;
     Ok(())
 }
 
+pub fn get_fastfield_cardinality(field_type: &FieldType) -> Option<Cardinality> {
```
fulmicoton, line 192: does this need to be public?

fulmicoton, line 192: Thanks for thinking about the name. (calling it fastfield_cardinality.)
The notion of cardinality is indeed ambiguous.

A single valued bytes field in quickwit has a multivalued fast field cardinality.

Maybe adding a comment would be a good addition. 
(and make the function private)

quickwit-search/src/leaf.rs
```diff
@@ -158,14 +222,42 @@ async fn warm_up_fastfields(
...
+                    fast_field_name, field_entry
+                )
+            })?;
+
+        fast_fields.push((fast_field, cardinality));
     }
 
-    let mut warm_up_futures = Vec::new();
-    for field in fast_fields {
+    type SendebleFuture = dyn Future<Output = Result<OwnedBytes, AsyncIoError>> + Send;
```
fulmicoton, line 236: Sendable?

fulmicoton (approved):
see comments inline

quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
PSeitz: will be fixed with https://github.com/quickwit-oss/tantivy/pull/1346

quickwit-search/src/leaf.rs
PSeitz: I just commented on the function to document the behaviour. field is a tantivy `Field` here.

fulmicoton (approved):

Issue #1296, opened 2022-04-15
Remove dependency to chrono and use time directly


fulmicoton:
.. if possible?

guilload:
Closed by #1243.

PR #1317, opened 2022-04-20
remove chrono

PSeitz:
remove chrono

#1296


fulmicoton (approved):

Cargo.lock
```diff
@@ -2702,7 +2701,6 @@ dependencies = [
  "anyhow",
  "async-trait",
  "byte-unit",
- "chrono",
```
fulmicoton, line 2705: Is there still a deps on chrono somewhere?
(I was expecting a larger diff in Cargo.lock)

PSeitz: yes, on diesel. They use it for serialization and don't plan to replace it https://github.com/diesel-rs/diesel/issues/2928

fmassot (approved):

fulmicoton (commented):
Thanks for asking the diesel people. They mentioned it was optional. Do we actually use this feature?

PSeitz:
> Thanks for asking the diesel people. They mentioned it was optional. Do we actually use this feature?

We use it in the postgremodel https://github.com/quickwit-oss/quickwit/blob/main/quickwit-metastore/src/postgresql/model.rs#L64

I think we could replace it and use i64 directly

NativeDateTime seems to be converted to timestamp which is i64

https://github.com/diesel-rs/diesel/blob/50c2c85032857a65e4a8d68dc851fa7ab91d65b2/diesel/src/pg/types/date_and_time/chrono.rs#L45

https://github.com/diesel-rs/diesel/blob/50c2c85032857a65e4a8d68dc851fa7ab91d65b2/diesel/src/pg/types/date_and_time/mod.rs#L22

Cargo.lock
fulmicoton: It seems optional though. Do we need that feature?

PSeitz:
I did a test, and simply replacing the field with i64 will not work with an additional migration step.

PR #1336, opened 2022-04-26
single postgre instance

### LINKED ISSUE #954: Cache the postgresql metastores
Similarly to what was done in #926, we probably want to ensure we have a single postgresql metastore instance per postgresql uri.

PSeitz:
fixes #954

### DIFF
--- a/quickwit-metastore/src/metastore_resolver.rs
+++ b/quickwit-metastore/src/metastore_resolver.rs
@@ -79,8 +79,44 @@ pub fn quickwit_metastore_uri_resolver() -> &'static MetastoreUriResolver {
                 .register("postgresql", PostgresqlMetastoreFactory::default());
         }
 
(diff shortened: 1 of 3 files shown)


quickwit-metastore/src/metastore_resolver.rs
```diff
@@ -79,10 +79,46 @@ pub fn quickwit_metastore_uri_resolver() -> &'static MetastoreUriResolver {
...
+                    UnsuportedMetastore {
+                        message: "postgres unsupported, quickwit was compiled without the \
+                                  'postgres' feature flag"
+                            .to_string(),
+                    },
+                )
+                .register(
+                    "postgresql",
+                    UnsuportedMetastore {
+                        message: "postgresql unsupported, quickwit was compiled without the \
```
fulmicoton, line 96: neat idea.

fulmicoton (approved):

PR #1345, opened 2022-04-28
Added a `doc_to_json` in the DocMapper.

### LINKED ISSUE #1293: doc_to_json ignores Cardinality and makes everything an array.
Right now we rely on tantivy to generate the json representation of the documents we fetch.

This is not great, because tantivy does not have a clear notion of cardinality.
As a result, all fields end up being wrapped in arrays in this JSON.

`{  "body": ["hello"] }` instead of `{  "body": "hello" }`

We should add  a `json_to_doc` method to the `DocMapper`.

### LINKED ISSUE #1325: Fixed dynamic field in get_doc json output.

fulmicoton:
Formatting returned documents.
    
    Before this PR, documents are just json serialized
    by tantivy.
    
    This is bad because:
    - object fields are not constructed,
    - cardinality is ignored, and all fields are wrapped in arrays,
    - the dynamically mapped fields are added in `_dynamic`,
    
    This PR adds a method to the `DocMapper` in
    order to format JSON Object for documents.
    
    This formatting happens at the root level, in order
    to use the same DocMapper (output format) for all hits.
    
    In addition, the _source field used to be stored in
    text field. After this change, the _source field is stored in a
    JSON field, and returned as a JSON Object (it used to be a JSON
    serialized string within the json document).
    
    Closes #1325
    Closes #1293

### DIFF
--- a/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
+++ b/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
@@ -18,6 +18,7 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::any::type_name;
+use std::collections::BTreeMap;
 
 use anyhow::bail;
 use serde_json::Value as JsonValue;
@@ -35,6 +36,16 @@ use crate::default_doc_mapper::field_mapping_entry::{
 use crate::default_doc_mapper::{FieldMappingType, QuickwitJsonOptions};
 use crate::{DocParsingError, FieldMappingEntry, ModeType};
 
+#[derive(Copy, Clone, PartialEq, Eq)]
+enum JsonType {
+    Null,
+    Bool,
+    Array,
+    Number,
+    String,
+    Object,
+}
+
 #[derive(Clone)]
 pub enum LeafType {
     Text(QuickwitTextOptions),
@@ -47,6 +58,16 @@ pub enum LeafType {
 }
 
 impl LeafType {
+    fn json_type(&self) -> JsonType {
+        match self {
+            LeafType::Text(_) => JsonType::String,
+            LeafType::I64(_) | LeafType::U64(_) | LeafType::F64(_) => JsonType::Number,
+            LeafType::Date(_) => JsonType::String,
+            LeafType::Bytes(_) => JsonType::String,
+            LeafType::Json(_) => JsonType::Object,
+        }
+    }
+
     pub fn is_fast_field(&self) -> bool {
         match self {
             LeafType::Text(_opt) => false, // TODO fixme once we have text fast field
@@ -152,6 +173,19 @@ impl MappingLeaf {
         Ok(())
     }
 
+    fn populate_json<'a>(
+        &'a self,
+        named_doc: &mut BTreeMap<String, Vec<JsonValue>>,
+        field_path: &[&'a str],
+        doc_json: &mut serde_json::Map<String, serde_json::Value>,
+    ) {
+        let json_type = self.typ.json_type();
+        if let Some(json_val) = extract_json_val(json_type, named_doc, field_path, self.cardinality)
+        {
+            insert_json_val(field_path, json_val, doc_json);
+        }
+    }
+
     pub fn field(&self) -> Field {
         self.field
     }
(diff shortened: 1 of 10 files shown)


quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -161,6 +186,74 @@ impl MappingLeaf {
     }
 }
 
+fn convert_json_val(json_type: JsonType, json_val: JsonValue) -> Option<JsonValue> {
```
PSeitz, line 189: it's not really a conversion

the match can be simplified with

```
JsonValue::Number(_) if json_type == JsonType::Number => Some(json_val),
```

quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -161,6 +186,74 @@ impl MappingLeaf {
...
+    }
+}
+
+fn insert_json_val(
+    field_path: &[&str],
+    json_val: JsonValue,
+    mut doc_json: &mut serde_json::Map<String, serde_json::Value>,
+) {
+    assert!(!field_path.is_empty());
+    for field_name in &field_path[..field_path.len() - 1] {
```
PSeitz, line 237: [split_last](https://doc.rust-lang.org/std/primitive.slice.html#method.split_last) should make the intent clearer

quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -161,6 +186,74 @@ impl MappingLeaf {
...
+    }
+    doc_json.insert(field_path[field_path.len() - 1].to_string(), json_val);
+}
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+enum JsonType {
+    Number,
+    Object,
+    String,
+}
```
PSeitz, line 255: I'd move this up next to LeafType

quickwit-doc-mapper/src/doc_mapper.rs
```diff
@@ -43,6 +44,21 @@ pub trait DocMapper: Send + Sync + Debug + DynClone + 'static {
...
+    ///
+    /// Tantivy does not have any notion of cardinality nor object.
+    /// It is therefore up to the `DocMapper` to pick a tantivy named document
+    /// and convert it into a final quickwit document.
+    ///
+    /// Because this operation is dependent on the `DocMapper`, this
+    /// method is meant to be called on the root node using the most recent
+    /// `DocMapper`. This ensures that the different hits are formatted according
+    /// to the same schema.
+    fn doc_to_json(
```
PSeitz, line 57: We could consider custom payloads in the tantivy docstore. This could be done by converting `Document` to a trait, that e.g. implements doc_to_json or similar

quickwit-search/src/lib.rs
```diff
@@ -146,6 +149,35 @@ async fn list_relevant_splits(
...
+/// to multivalued, we do want the documents emitted from old splits to
+/// also serialize the fields values as a JsonArray.
+///
+/// The `convert_leaf_hit` is critical and needs to be tested against
+/// allowed DocMapper changes.
+fn convert_leaf_hit(
+    leaf_hit: quickwit_proto::LeafHit,
+    doc_mapper: &dyn DocMapper,
+) -> crate::Result<quickwit_proto::Hit> {
+    let hit_json: BTreeMap<String, Vec<JsonValue>> = serde_json::from_str(&leaf_hit.leaf_json)
```
PSeitz, line 171: Why BTreeMap and not a HashMap?

quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
fulmicoton: Nice catch. Yes it is not a conversion.

quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
fulmicoton: agreed!

quickwit-search/src/lib.rs
fulmicoton: We can make it a hashmap.

fulmicoton: ah.. it's part of the trait. I'd rather keep it BTreeMap then.

For the reason, I made it a BTreeMap because:
- NamedFieldDocument is ordered. It does not matter much though.
- I first wanted to handle more complex changes of schema, such as changing a field from 
a JsonObject to a more structured object type.

That would probably involve search different prefix of a field_path in decreasing order...

quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
fulmicoton: done

quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
fulmicoton: Actually I did not know split_last. very nice!

quickwit-doc-mapper/src/doc_mapper.rs
PSeitz: I added something here https://github.com/quickwit-oss/tantivy/issues/1352

fulmicoton: thank you!

PR #1355, opened 2022-05-02
document aggregation

### LINKED ISSUE #1198: Update docs for the Search REST API
The search REST API docs are not up to date with the new aggregation endpoint.

To do after #1197

### LINKED ISSUE #1350: Document aggregation

PSeitz:
closes #1350
closes #1198

### Description
Add new document for aggregations, reference document in rest-api.md

### DIFF
--- a/docs/reference/aggregation.md
+++ b/docs/reference/aggregation.md
@@ -286,7 +286,7 @@ The from value is inclusive in the range.
 The to value is not inclusive in the range.
 
 The first bucket can omit the `from` value, and the last bucket the `to` value.
-Note that this aggregation includes the from value and excludes the to value for each range. Extra buckets will be created until the first to, and last from, if necessary.
+Note that this aggregation includes the `from` value and excludes the `to` value for each range. Extra buckets will be created until the first `to`, and last `from`, if necessary.
 
 
 ### Terms
--- a/docs/reference/aggregation.md
+++ b/docs/reference/aggregation.md
@@ -136,7 +136,7 @@ Example request, histogram with stats in each bucket:
 
 #### Limitations/Compatibility
 
-Currently aggregations work only on single value fast fields of type u64, f64 and i64.
+Currently aggregations work only on single value fast fields of type `u64`, `f64`, `i64` and `text`.
 
 Elasticsearch `keyed` parameter is not yet supported.
 
@@ -213,7 +213,7 @@ The minimum number of documents in a bucket to be returned. Defaults to 0.
 
 Limits the data range to [min, max] closed interval.
 This can be used to filter values if they are not in the data range.
-hard_bounds only limits the buckets, to force a range set both extended_bounds and hard_bounds to the same range.
+hard_bounds only limits the buckets, to force a range set both `extended_bounds` and `hard_bounds` to the same range.
 
 ```json
 {
@@ -237,7 +237,7 @@ hard_bounds only limits the buckets, to force a range set both extended_bounds a
 ###### **extended_bounds**
 
 Can be set to extend your bounds. The range of the buckets is by default defined by the data range of the values of the documents. As the name suggests, this can only be used to extend the value range. If the bounds for min or max are not extending the range, the value has no effect on the returned buckets.
-Cannot be set in conjunction with min_doc_count > 0, since the empty buckets from extended bounds would not be returned.
+Cannot be set in conjunction with `min_doc_count` > 0, since the empty buckets from extended bounds would not be returned.
 
 ```json
 {
@@ -287,7 +287,7 @@ Note that this aggregation includes the from value and excludes the to value for
 
 Overlapping ranges are not yet supported.
 
-The keyed parameter (elasticsearch) is not yet supported.
+Elasticsearch `keyed` parameter is not yet supported.
 
 
 #### Parameters
--- a/docs/reference/aggregation.md
+++ b/docs/reference/aggregation.md
@@ -106,7 +106,7 @@ Currently aggregations work only on single value fast fields of type u64, f64, i
 
 ## Bucket Aggregations
 
-BucketAggregations create buckets of documents. Each bucket is associated with a rule which determines whether or not a document in the falls into it. 
+BucketAggregations create buckets of documents. Each bucket is associated with a rule which determines whether or not a document falls into it. 
 In other words, the buckets effectively define document sets. Buckets are not necessarily disjunct, therefore a document can fall into multiple buckets. 
 In addition to the buckets themselves, the bucket aggregations also compute and return the number of documents for each bucket. 
 Bucket aggregations, as opposed to metric aggregations, can hold sub-aggregations. 
--- a/docs/reference/aggregation.md
+++ b/docs/reference/aggregation.md
@@ -281,7 +281,7 @@ The field to aggregate on.
 
 ###### **ranges**
 
-The list of buckets, with from and to values. 
+The list of buckets, with `from` and `to` values. 
 The from value is inclusive in the range.
 The to value is not inclusive in the range.
 
--- a/docs/reference/aggregation.md
+++ b/docs/reference/aggregation.md
@@ -41,24 +41,22 @@ In some examples below is not the full request shown, but only the payload for `
 Request
 ```json
 {
-  "query": "*",
-  "max_hits": 0,
-  "aggregations": {
-    "sites_and_aqi": {
-      "terms": {
-        "field": "County",
-        "size": 2,
-        "order": {"average_aqi": "asc"}
-      },
-      "aggs": {
-        "average_aqi": {
-          "avg": {
-            "field": "AQI"
-          }
+    "query": "*",
+    "max_hits": 0,
+    "aggs": {
+        "sites_and_aqi": {
+            "terms": {
+                "field": "County",
+                "size": 2,
+                "order": {"average_aqi": "asc"}
+            },
+            "aggs": {
+                "average_aqi": {
+                    "avg": { "field": "AQI" }
+                }
+            }
         }
-      }
     }
-  }
 }
 ```
 
@@ -66,7 +64,7 @@ Request
 Response
 ```json
 ...
-"aggregations": {
+"aggs": {
     "sites_and_aqi": {
       "buckets": [
         {
@@ -118,23 +116,21 @@ Some define a single bucket, some define fixed number of multiple buckets, and o
 Example request, histogram with stats in each bucket:
 ```json
 {
-  "query": "*",
-  "max_hits": 1,
-  "aggregations": {
-    "stats_per_day": {
-      "histogram": {
-        "field": "timestamp",
-        "interval": 86400
-      },
-      "aggs": {
-        "timestamp_stats": {
-          "stats": {
-            "field": "timestamp"
-          }
+    "query": "*",
+    "max_hits": 0,
+    "aggs": {
+        "stats_per_day": {
+            "histogram": {
+                "field": "timestamp",
+                "interval": 86400
+            },
+            "aggs": {
+                "timestamp_stats": {
+                    "stats": { "field": "timestamp" }
+                }
+            }
         }
-      }
     }
-  }
 }
 ```
 
@@ -162,10 +158,14 @@ The value range of the buckets can bet extended via extended_bounds or limit the
 
 ```json
 {
-    "prices": {
-        "histogram": {
-            "field": "price",
-            "interval": 10
+    "query": "*",
+    "max_hits": 0,
+    "aggs": {
+        "prices": {
+            "histogram": {
+                "field": "price",
+                "interval": 10
+            }
         }
     }
 }
@@ -190,11 +190,15 @@ As an example, if there are two documents with value 8 and 12 and interval 10.0,
 
 ```json
 {
-    "prices": {
-        "histogram": {
-            "field": "price",
-            "interval": 10,
-            "offset": 2.5
+    "query": "*",
+    "max_hits": 0,
+    "aggs": {
+        "prices": {
+            "histogram": {
+                "field": "price",
+                "interval": 10,
+                "offset": 2.5
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+
+- What is the average price of all sold articles?
+- How many errors with status code 500 do we have per day?
+- What is the average listing price of cars grouped by color?
+
+There are two categories: [Metrics](#metric) and [Buckets](#bucket-aggregations).
+
+#### Prerequisite
+
+To be able to use aggregations on a field, the field needs have a fast field index created. A fast field index is a columnar storage,
```
fmassot, line 18:
```suggestion
To be able to use aggregations on a field, the field needs to have a fast field index created. A fast field index is a columnar storage, 
```

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+- What is the average listing price of cars grouped by color?
+
+There are two categories: [Metrics](#metric) and [Buckets](#bucket-aggregations).
+
+#### Prerequisite
+
+To be able to use aggregations on a field, the field needs have a fast field index created. A fast field index is a columnar storage, 
+where documents values are extracted and stored to.
+
+Example to create a fast field on text for term aggregations.
```
fmassot, line 21: Maybe mention the fact that the text field can be a fast field only if there is not tokenization

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+```
+
+See the [index config](./index-config.md) for more details and examples.
+
+#### Format
+
+The aggregation request and result de/serialize into elasticsearch compatible JSON. 
+If not documented otherwise you should be able to drop in your elasticsearch aggregation queries.
+
+In some examples below is not the full request shown, but only the payload for `aggregations`.
```
fmassot, line 37: I would put always the complete query. By the way, can't we put `"max_hits": 0` in the request?

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+          "doc_count": 28675,
+          "key": "花蓮縣"
+        }
+      ],
+      "sum_other_doc_count": 1872055
+    }
+```
+#### Limitations
+
+Currently aggregations work only on single value fast fields of type u64, f64, i64 and on string fields.
```
fmassot, line 92: it's `text` field with `raw` tokenizer only, we don't have yet a "string" field but that would be nice to have one (like a shortcut to define a text field with tokenizer raw).

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+    - [Range](#range)
+    - [Terms](#terms)
+- Metric
+    - [Average](#average)
+    - [Stats](#stats)
+
+
+## Bucket Aggregations
+
+BucketAggregations create buckets of documents. Each bucket is associated with a rule which determines whether or not a document in the falls into it.
```
fmassot, line 109:
```suggestion
BucketAggregations create buckets of documents. Each bucket is associated with a rule which determines whether or not a document falls into it. 
```

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+#### Limitations/Compatibility
+
+Currently aggregations work only on single value fast fields of type u64, f64 and i64.
```
fmassot, line 142: and `text` with a `raw` tokenizer.

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+
+The keyed parameter (elasticsearch) is not yet supported.
```
fmassot, line 144:
```suggestion
Elasticsearch `keyed` parameter is not yet supported.
```

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+
+### Histogram
+
+Histogram is a bucket aggregation, where buckets are created dynamically for given interval. Each document value is rounded down to its bucket.
```
fmassot, line 148:
```suggestion
Histogram is a bucket aggregation, where buckets are created dynamically for the given interval. Each document value is rounded down to its bucket.
```

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+
+E.g. if we have a price 18 and an interval of 5, the document will fall into the bucket with the key 15. The formula used for this is: ((val - offset) / interval).floor() * interval + offset
```
fmassot, line 150:
```suggestion
E.g. if we have a price 18 and an interval of 5, the document will fall into the bucket with the key 15. The formula used for this is: ((val - offset) / interval).floor() * interval + offset.
```

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+
+#### Parameters
+
+###### **field**
+
+The field to aggregate on.
+
+###### **interval**
+
+The interval to chunk your data range. Each bucket spans a value range of [0..interval). Must be a positive value.
```
fmassot, line 181: Does an interval can be equal to 0?

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+The field to aggregate on.
+
+###### **ranges**
+
+The list of buckets, with from and to values. 
+The from value is inclusive in the range.
+The to value is not inclusive in the range.
+
+The first bucket can omit the `from` value, and the last bucket the `to` value.
+Note that this aggregation includes the from value and excludes the to value for each range. Extra buckets will be created until the first to, and last from, if necessary.
```
fmassot, line 289:
```suggestion
Note that this aggregation includes the `from` value and excludes the `to` value for each range. Extra buckets will be created until the first `to`, and last `from`, if necessary.
```

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+
+#### Parameters
+
+###### **field**
+
+The field to aggregate on.
+
+###### **ranges**
+
+The list of buckets, with from and to values.
```
fmassot, line 284:
```suggestion
The list of buckets, with `from` and `to` values. 
```

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+    "genres": {
+        "terms":{ "field": "genre" }
+    }
+}
+```
+
+
+#### Terminology
+Shard parameters are supposed to be equivalent to elasticsearch shard parameter.
+Since they are
```
fmassot, line 307: Unfinished sentence :)

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+
+#### Document count error
+To improve performance, results from one segment are cut off at `segment_size`. On a index with
```
fmassot, line 310: I think it's better to talk about split and not segment.

fmassot: I guess `segment_size` is from elasticsearch API, so maybe explain that a quickwit split is equivalent to an elasticsearch segment

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+The field to aggregate on.
+
+###### **size**
+
+By default, the top 10 terms with the most documents are returned. Larger values for size are more expensive.
+
+###### **segment_size**
+
+
+The get more accurate results, we fetch more than size from each segment.
```
fmassot, line 341: "The get more accurate results, we fetch more than the size from each segment."

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+Increasing this value is will increase the cost for more accuracy.
```
fmassot, line 342: "Increasing this value is will increase the accuracy but also the CPU/memory usage"?

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+
+Defaults to 10 * size.
+
+###### **show_term_doc_count_error**
+
+If you set the show_term_doc_count_error parameter to true, the terms aggregation will include doc_count_error_upper_bound, which is an upper bound to the error on the doc_count returned by each shard.
```
fmassot, line 348: A shard for quickwit is equivalent to a split, right? Maybe we should say that somewhere here because we never use shard across the documentation.

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+Increasing this value is will increase the cost for more accuracy.
+
+Defaults to 10 * size.
+
+###### **show_term_doc_count_error**
+
+If you set the show_term_doc_count_error parameter to true, the terms aggregation will include doc_count_error_upper_bound, which is an upper bound to the error on the doc_count returned by each shard. 
+It’s the sum of the size of the largest bucket on each segment that didn’t fit into shard_size.
+
+Defaults to true when ordering by counts desc.
```
fmassot, line 351:
```suggestion
Defaults to true when ordering by count desc.
```

docs/reference/aggregation.md
```diff
@@ -0,0 +1,488 @@
...
+
+
+###### **min_doc_count**
+
+Filter all terms than are lower min_doc_count. Defaults to 1.
```
fmassot, line 356:
```suggestion
Filter all terms that are lower than `min_doc_count`. Defaults to 1.
```

docs/reference/aggregation.md
PSeitz: It's mentioned later, text field supports also tokenization with fast field but is not recommended

docs/reference/aggregation.md
fulmicoton: I like the idea of a tag field.

tag = indexed + fast + raw tokenizer.

docs/reference/aggregation.md
PSeitz: No, 0 will return an error. I'll update the wording

docs/reference/aggregation.md
PSeitz: segment_size is a tantivy specific variable. This is where we deviate a little from elasticsearch since we don't have a global termmapping

PSeitz: We also have `shard_size`, which is not used in quickwit currently. There is actually something missing here.
I added an alias for shard_size to split_size (https://github.com/quickwit-oss/tantivy/pull/1358) and this split_size would be set by quickwit to set the shard_size, and we only document the split_size parameter.

docs/reference/aggregation.md
```diff
@@ -0,0 +1,513 @@
...
+
+Filter all terms that are lower than `min_doc_count`. Defaults to 1.
+
+_Expensive_ : When set to 0, this will return all terms in the field.
+
+
+###### **order**
+
+Set the order. String is here a target, which is either “_count”, “_key”, or the name of a metric sub_aggregation.
+Single value metrics like average can be adressed by its name. Multi value metrics like stats are required to adress their field by name e.g. “stats.avg”
```
fmassot, line 381:
```suggestion
Single value metrics like average can be adressed by its name. Multi value metrics like stats are required to adress their field by name e.g. “stats.avg”.
```

docs/reference/rest-api.md
```diff
@@ -55,6 +55,7 @@ Search for documents matching a query in the given index `<index id>`.
 | **max_hits**              | `Integer`            | Maximum number of hits to return (by default 20)                                                           | `20`                                                                                            |
 | **search_field**          | `[String]`           | Fields to search on if no field name is specified in the query. Comma-separated list, e.g. "field1,field2" | index_config.search_settings.default_search_fields                                              |
 | **format**                | `Enum`               | The output format. Allowed values are "json" or "prettyjson"                                               | `prettyjson`                                                                                    |
+| **aggregations**          | `JSON`               | The aggregations request. See the [aggregations doc](aggregation.md) for supported aggregations. |
```
fmassot, line 58: `aggs` is also possible right?

docs/reference/aggregation.md
```diff
@@ -0,0 +1,513 @@
...
+
+**Response**
+```json
+{
+  "num_hits": 9582098,
+  "hits": [],
+  "elapsed_time_micros": 101942,
+  "errors": [],
+  "aggs": {
+    "average_in_range": {
```
fmassot, line 461: `average_price`?

docs/reference/aggregation.md
```diff
@@ -0,0 +1,513 @@
...
+Order by sub_aggregation
+
+```json
+{
+    "query": "*",
+    "max_hits": 0,
+    "aggs": {
+        "articles_by_price": {
+            "terms":{ "field": "article_name" },
+            "order":{ “average_price”: “asc” },
```
fmassot, line 408:
```suggestion
            "order":{ "average_price": "asc" },
```

docs/reference/aggregation.md
```diff
@@ -0,0 +1,513 @@
...
+    "aggs": {
+        "genres": {
+            "terms":{ "field": "genre" }
+        }
+    }
+}
+```
+
+#### Document count error
+In quickwit we have one segment per split.
```
fmassot, line 326:
```suggestion
In Quickwit, we have one segment per split.
```

docs/reference/aggregation.md
PSeitz: I updated it to always include the complete query

docs/reference/aggregation.md
fmassot: you can just add `text`.

docs/reference/aggregation.md
```diff
@@ -0,0 +1,507 @@
...
+                }
+            }
+        }
+    }
+}
+```
+
+#### Limitations/Compatibility
+
+Currently aggregations work only on single value fast fields of type u64, f64 and i64.
```
fmassot, line 139:
```suggestion
Currently aggregations work only on single value fast fields of type `u64`, `f64`, `i64` and `text`.
```

docs/reference/aggregation.md
```diff
@@ -0,0 +1,507 @@
...
+            }
+        }
+    }
+}
+```
+
+###### **extended_bounds**
+
+Can be set to extend your bounds. The range of the buckets is by default defined by the data range of the values of the documents. As the name suggests, this can only be used to extend the value range. If the bounds for min or max are not extending the range, the value has no effect on the returned buckets.
+Cannot be set in conjunction with min_doc_count > 0, since the empty buckets from extended bounds would not be returned.
```
fmassot, line 240:
```suggestion
Cannot be set in conjunction with `min_doc_count` > 0, since the empty buckets from extended bounds would not be returned.
```

docs/reference/aggregation.md
```diff
@@ -0,0 +1,507 @@
...
+        }
+    }
+}
+```
+
+#### Limitations/Compatibility
+
+Overlapping ranges are not yet supported.
+
+The keyed parameter (elasticsearch) is not yet supported.
```
fmassot, line 290:
```suggestion
Elasticsearch `keyed` parameter is not yet supported.
```

PR #1358, opened 2022-05-02
Set default config location to `./config/quickwit.yaml`

### LINKED ISSUE #1314: Log a warning if the cluster_id default value is not overridden

### LINKED ISSUE #1339: Default value for config filepath
The quickwit config filepath should default to `./config/quickwit.yaml`.

In other words, the value should be populated from the first of the following option:
- config CLI value
- ENV variable
- ./config/quickwit.yaml

guilload:
### Description
- Set default config location to `./config/quickwit.yaml`; fixes #1339
- Warn user when default cluster ID is used; fixes #1314

### How was this PR tested?
- `make test-all`
- ran a bunch of commands with `--config` at various positions.

### DIFF
--- a/quickwit-config/src/config.rs
+++ b/quickwit-config/src/config.rs
@@ -31,7 +31,9 @@ use quickwit_common::uri::{Extension, Uri};
 use serde::{Deserialize, Serialize};
 use tracing::{info, warn};
 
-static DEFAULT_DATA_DIR_PATH: &str = "./qwdata";
+pub const DEFAULT_QW_CONFIG_PATH: &str = "./config/quickwit.yaml";
+
+const DEFAULT_DATA_DIR_PATH: &str = "./qwdata";
 
 fn default_data_dir_path() -> PathBuf {
     PathBuf::from(DEFAULT_DATA_DIR_PATH)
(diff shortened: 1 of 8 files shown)


quickwit-config/src/config.rs
```diff
@@ -31,7 +31,11 @@ use quickwit_common::uri::{Extension, Uri};
 use serde::{Deserialize, Serialize};
 use tracing::{info, warn};
 
-static DEFAULT_DATA_DIR_PATH: &str = "./qwdata";
+pub const DEFAULT_QW_CONFIG_PATH: &str = "./config/quickwit.yaml";
+
+const DEFAULT_DATA_DIR_PATH: &str = "./qwdata";
+
+const DEFAULT_CLUSTER_ID: &str = "quickwit-test-cluster";
```
evanxg852000, line 38: Nitpik: `quickwit-default-cluster` sounds better to me.

evanxg852000 (approved):
Looks good, just  a nitpick you can suggestion

quickwit-config/src/config.rs
guilload: Yes, I kept the original name, but I prefer `quickwit-default-cluster` too.

PR #1360, opened 2022-05-03
Validate config URIs

### LINKED ISSUE #1061: empty string is not a valid Metastore URI
and should therefore be rejected.

Same for `default_index_root_uri`

guilload:
### Description
- Validate config URIs (metastore URI, index URI, index root URI), fixes #1061

### How was this PR tested?
- Added unit tests
- Ran `make test-all`

### DIFF
--- a/quickwit-cli/src/index.rs
+++ b/quickwit-cli/src/index.rs
@@ -513,14 +513,12 @@ pub async fn list_index_cli(args: ListIndexesArgs) -> anyhow::Result<()> {
     debug!(args = ?args, "list");
     let metastore_uri_resolver = quickwit_metastore_uri_resolver();
     let quickwit_config = load_quickwit_config(&args.config_uri, None).await?;
-    let metastore_uri = if let Some(uri) = args.metastore_uri {
-        uri.to_string()
-    } else {
-        quickwit_config.metastore_uri()
-    };
+    let metastore_uri = args
+        .metastore_uri
+        .unwrap_or_else(|| quickwit_config.metastore_uri());
     let metastore = metastore_uri_resolver.resolve(&metastore_uri).await?;
-    let res = metastore.list_indexes_metadatas().await?;
-    let index_table = make_list_indexes_table(res);
+    let indexes = metastore.list_indexes_metadatas().await?;
+    let index_table = make_list_indexes_table(indexes);
 
     println!();
     println!("{}", index_table);
--- a/quickwit-common/Cargo.toml
+++ b/quickwit-common/Cargo.toml
@@ -13,15 +13,17 @@ documentation = "https://quickwit.io/docs/"
 anyhow = "1"
 colored = "2"
 env_logger = "0.9"
+futures = "0.3"
+home = "0.5.3"
 once_cell = "1"
 prometheus = { version = "0.13", features = ["process"] }
 rand = "0.8"
 regex = "1"
+serde = "1.0"
 tokio = { version = "1", features = ["fs", "macros", "rt"] }
-warp = '0.3'
 tracing = "0.1.29"
-home = "0.5.3"
-futures = "0.3"
+warp = '0.3'
 
 [dev-dependencies]
+serde_json = "1.0"
 tempfile = "3"
--- a/quickwit-common/src/uri.rs
+++ b/quickwit-common/src/uri.rs
@@ -23,9 +23,14 @@ use std::fmt::Display;
 use std::path::{Component, Path, PathBuf};
 
 use anyhow::{bail, Context};
+use serde::{Serialize, Serializer};
 
 /// Default file protocol `file://`
-const FILE_PROTOCOL: &str = "file";
+pub const FILE_PROTOCOL: &str = "file";
+
+const POSTGRES_PROTOCOL: &str = "postgres";
+
+const POSTGRESQL_PROTOCOL: &str = "postgresql";
 
 const PROTOCOL_SEPARATOR: &str = "://";
 
@@ -57,23 +62,25 @@ pub struct Uri {
 }
 
 impl Uri {
-    /// Tries to to construct a Uri from the raw string.
+    /// Attempts to construct a [`Uri`] from a raw string slice.
     /// A `file://` protocol is assumed if not specified.
-    /// File URIs are resolved (normalised) relative to the current working directory
+    /// File URIs are resolved (normalized) relative to the current working directory
     /// unless an absolute path is specified.
-    /// Handles special characters like (~, ., ..)
+    /// Handles special characters like `~`, `.`, `..`.
     pub fn try_new(uri: &str) -> anyhow::Result<Self> {
+        if uri.is_empty() {
+            bail!("URI is empty.");
+        }
         let (protocol, mut path) = match uri.split_once(PROTOCOL_SEPARATOR) {
             None => (FILE_PROTOCOL, uri.to_string()),
             Some((protocol, path)) => (protocol, path.to_string()),
         };
-
         if protocol == FILE_PROTOCOL {
             if path.starts_with('~') {
                 // We only accept `~` (alias to the home directory) and `~/path/to/something`.
                 // If there is something following the `~` that is not `/`, we bail out.
                 if path.len() > 1 && !path.starts_with("~/") {
-                    bail!("This path syntax `{}` is not supported.", uri);
+                    bail!("Path syntax `{}` is not supported.", uri);
                 }
 
                 let home_dir_path = home::home_dir()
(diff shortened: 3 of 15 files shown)


quickwit-common/Cargo.toml
```diff
@@ -13,15 +13,17 @@ documentation = "https://quickwit.io/docs/"
 anyhow = "1"
 colored = "2"
 env_logger = "0.9"
+futures = "0.3"
```
guilload, line 16: Sorted the deps and added `serde`.

quickwit-config/src/config.rs
```diff
@@ -387,6 +391,16 @@ impl std::fmt::Debug for QuickwitConfig {
...
+where D: Deserializer<'de> {
+    let uri_opt: Option<String> = Deserialize::deserialize(deserializer)?;
+    uri_opt
+        .map(|uri| Uri::try_new(&uri))
+        .transpose()
+        .map_err(D::Error::custom)
+}
+
 #[cfg(test)]
 mod tests {
```
guilload, line 405: Todo: add an extra unit test here for config with invalid URI.

quickwit-core/src/index.rs
```diff
@@ -83,21 +84,23 @@ impl IndexService {
...
             );
-            default_index_uri
+            index_uri
         };
-
         let index_metadata = IndexMetadata {
-            index_id: index_config.index_id.clone(),
-            index_uri,
+            index_id,
+            index_uri: index_uri.into_string(),
```
guilload, line 103: `IndexMetadata.index_uri` should be a `Uri` eventually.

quickwit-metastore/src/metastore_resolver.rs
```diff
@@ -126,11 +126,14 @@ impl MetastoreUriResolver {
     }
 
     /// Resolves the given URI.
-    pub async fn resolve(&self, uri: &str) -> Result<Arc<dyn Metastore>, MetastoreResolverError> {
-        let protocol = uri.split("://").next().ok_or_else(|| {
+    pub async fn resolve<S: AsRef<str>>(
```
guilload, line 129: Adding `S: AsRef<str>` allowed me to not change `.resolve(&metastore_uri)` to `.resolve(metastore_uri.as_ref())` all over the place. We can get rid of it once we change the signature of `resolve(&str)` to `resolve(&Uri)`

quickwit-config/src/config.rs
guilload: Done.

quickwit-common/src/uri.rs
```diff
@@ -83,48 +90,94 @@ impl Uri {
...
+                self.uri,
+                path
+            );
+        }
+        let joined = match self.protocol() {
+            FILE_PROTOCOL => Path::new(&self.uri)
+                .join(path)
+                .to_string_lossy()
+                .to_string(),
+            RAM_PROTOCOL | S3_PROTOCOL => format!(
```
evanxg852000, line 162: I would just do this for any other protocol and specially treat `file://` like you have above.
The semantic being that when you join a path, you need it with a separator. we also have `postgres` `postgresql` and many more.

quickwit-common/src/uri.rs
```diff
@@ -83,48 +90,94 @@ impl Uri {
...
+                path
+            ),
+            other => bail!("Protocol `{}` does not support `Uri::join`.", other),
+        };
+        Ok(Self {
+            uri: joined,
+            protocol_idx: self.protocol_idx,
+        })
+    }
+}
```
evanxg852000, line 175: I also found it handy to add 
```rust
pub fn to_str(&self) -> &str {
    self.as_ref()
}
```

quickwit-common/src/uri.rs
guilload: I agree, adding `as_str`.

`join` does not really make sense for `postgres` URIs that is why I wanted those to fall in the catchall category and raise. However, the way I've implemented this function prevents the use of protocols that we have not anticipated, which is bound to break at some point. So I'm going to go with your recommendation for that reason :)

quickwit-cli/src/index.rs
```diff
@@ -513,14 +513,12 @@ pub async fn list_index_cli(args: ListIndexesArgs) -> anyhow::Result<()> {
     debug!(args = ?args, "list");
     let metastore_uri_resolver = quickwit_metastore_uri_resolver();
     let quickwit_config = load_quickwit_config(&args.config_uri, None).await?;
-    let metastore_uri = if let Some(uri) = args.metastore_uri {
-        uri.to_string()
-    } else {
-        quickwit_config.metastore_uri()
-    };
+    let metastore_uri = args
```
fmassot, line 516: Nice, the previous code was "ugly".

quickwit-common/Cargo.toml
fmassot: Is there a tool to sort the deps?

quickwit-common/src/uri.rs
```diff
@@ -83,48 +90,103 @@ impl Uri {
...
         }
     }
+
+    /// Consumes the [`Uri`] struct and returns the normalized URI as a string.
+    pub fn into_string(self) -> String {
+        self.uri
+    }
+
+    /// Creates a new [`Uri`] with `path` adjoined to `self`.
+    /// Fails if `path` is absolute, it replaces the current path.
```
fmassot, line 153: Is `/// Fails if `path` is absolute.` not sufficient for the docs? Do you want to justify why it fails? In this case I would write `Fails if `path` is absolute as we don't want to replace the current path`?

quickwit-common/src/uri.rs
```diff
@@ -83,48 +90,103 @@ impl Uri {
...
         Ok(Self {
             uri: format!("{}{}{}", protocol, PROTOCOL_SEPARATOR, path),
             protocol_idx: protocol.len(),
         })
     }
 
-    /// Returns the URI extension.
+    /// Constructs a [`Uri`] from a properly formatted string `<protocol>://<path>` where `path` is
+    /// normalized. Use this method exclusively for trusted input.
+    pub fn new(uri: String) -> Self {
```
fmassot, line 112: It is strange to make this function public.

quickwit-config/src/config.rs
```diff
@@ -252,6 +255,11 @@ impl QuickwitConfig {
         if self.peer_seeds.is_empty() {
             warn!("Seed list is empty.");
         }
+        let data_dir_uri = Uri::try_new(&self.data_dir_path.to_string_lossy())?;
+
+        if data_dir_uri.protocol() != "file" {
```
fmassot, line 260: Use `FILE_PROTOCOL` instead of `"file"`

guilload: Good catch.

quickwit-common/src/uri.rs
guilload: Yes, I don't know what happened here, seems like an unfortunate copy and paste :D
Will fix.

quickwit-core/src/index.rs
```diff
@@ -83,21 +84,23 @@ impl IndexService {
...
-        let index_uri = if let Some(index_uri) = index_config.index_uri.as_ref() {
-            index_uri.to_string()
+        let index_id = index_config.index_id.clone();
+        let index_uri = if let Some(index_uri) = &index_config.index_uri {
+            index_uri.clone()
         } else {
-            let default_index_uri =
-                format!("{}/{}", self.default_index_root_uri, index_config.index_id);
+            let index_uri = self.default_index_root_uri.join(&index_id).expect(
+                "Failed to create default index URI. This should never happen. Please, report.",
```
fmassot, line 92: Don't know if it's a good idea to put the github issue link.

quickwit-common/src/uri.rs
guilload: Yes. I don't like that it opens the door for shooting yourself in the foot. At the same time, sometimes you know the input is trustworthy (deserialization from metastore), and that's really what you want to use. It's also very convenient for writing less verbose unit tests.

quickwit-core/src/index.rs
guilload: Might as well.

quickwit-core/src/index.rs
fmassot: Indeed, we could open an issue to track that.

quickwit-common/src/uri.rs
fmassot: ok, so at least there is one useful use case. 
For unit tests, we can make it available only for that.

fmassot (approved):
Nice job as always, log messages are way better than before.

quickwit-common/src/uri.rs
guilload: I don't know if we can do it across crates of the same workspace. Last time I tried I failed.

quickwit-common/Cargo.toml
guilload: The cargo team is still working on it.

PR #1362, opened 2022-05-04
Fixing how we handle "."

### LINKED ISSUE #1334: Sort out what should be done for fields with a "."
Right now we accepts fields with a dot in its name, but quickwti's behavior is then ill-defined.
In particular it interfers with the concept of object fields, json fields, and dynamic fields.

... We need some spec!

### LINKED ISSUE #1338: Dynamic field, "." in field names and query parsing.
As spotted by @fmassot we have a bug when the query has a dot.

```rust
#[test]
    fn test_doc_mapper_accept_sub_field_query_on_non_json_field() {
        let mut doc_mapper_builder = DefaultDocMapperBuilder::default();
        doc_mapper_builder.field_mappings.push(FieldMappingEntry {
            name: "json_field".to_string(),
            mapping_type: FieldMappingType::Json(
                QuickwitJsonOptions::default(),
                Cardinality::SingleValue,
            ),
        });
        doc_mapper_builder.field_mappings.push(FieldMappingEntry {
            name: "body".to_string(),
            mapping_type: FieldMappingType::Text(
                QuickwitTextOptions::default(),
                Cardinality::SingleValue,
            ),
        });
        
        let doc_mapper = doc_mapper_builder.build().unwrap();
        let schema = doc_mapper.schema();
            index_id: "quickwit-index".to_string(),
            query: "body.wrong_field:hello".to_string(),
            search_fields: vec![],
            start_timestamp: None,
            end_timestamp: None,
            max_hits: 10,
            start_offset: 0,
            sort_order: None,
            sort_by_field: None,
            aggregation_request: None,
        };
        let query = doc_mapper.query(schema, &search_request).unwrap();
        assert_eq!(
            format!("{:?}", query),
            r#"TermQuery(Term(type=Json, field=1, vtype=Str, "hello"))"#
        );
    }
```

in fact it is 
  left: `"TermQuery(Term(type=Str, field=1, \"hello\"))"`,

but anyway it sucks.

More importantly we need to make sure (as in write a unit test) to check that the query parser  works (it probably doesn't) correctly if we have a dynamic field with an object with a partial schema.

For instance,
```
field_mappings:
   - name: params
     type: object
     field_mappings:
         - name: color
           type: text
mode: dynamic
```

if we have json object `{"params": {"size": "L"}}`, how do we query for params.size?
params.size:L should work and params.color:red should work as well.

fulmicoton:
```
    Fix how quickwit handles "." in field names.
    
    After this PR we still accept field names with a ".".
    On the query side this dot then needs to be escaped with a backslash
    to disambiguate it from the "." of Json fields, and the "." of
    object fields.
    
    Under the hood, we still rely on tantivy's query parser.
    The hack consists in escaping the tantivy field name itself.
    
    This PR updates tantivy in order to include
    https://github.com/quickwit-oss/tantivy/pull/1356.
    
    Closes #1338
    Closes #1334
```

### DIFF
--- a/quickwit-metastore/src/metastore/index_metadata.rs
+++ b/quickwit-metastore/src/metastore/index_metadata.rs
@@ -131,8 +131,8 @@ impl IndexMetadata {
         let search_settings = SearchSettings {
             default_search_fields: vec![
                 "body".to_string(),
-                "attributes.server".to_string(),
-                "attributes.server.status".to_string(),
+                r#"attributes.server"#.to_string(),
+                r#"attributes.server\.status"#.to_string(),
             ],
         };
         let now_timestamp = utc_now_timestamp();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -972,7 +972,7 @@ checksum = "25c7df09945d65ea8d70b3321547ed414bbc540aad5bac6883d021b970f35b04"
 [[package]]
 name = "fastfield_codecs"
 version = "0.1.0"
-source = "git+https://github.com/quickwit-oss/tantivy?rev=8c1e1cf#8c1e1cf1adaa4c84551ce4ffa797e0684408474a"
+source = "git+https://github.com/quickwit-oss/tantivy?rev=ed26552#ed26552296e4e380c7a286bfe52d994f8211d04e"
 dependencies = [
  "tantivy-bitpacker",
  "tantivy-common",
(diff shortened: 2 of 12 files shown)


quickwit-metastore/src/metastore/index_metadata.rs
```diff
@@ -131,8 +131,8 @@ impl IndexMetadata {
         let search_settings = SearchSettings {
             default_search_fields: vec![
                 "body".to_string(),
-                "attributes.server".to_string(),
-                "attributes.server.status".to_string(),
+                r#"attributes.server"#.to_string(),
+                r#"attributes.server\.status"#.to_string(),
```
PSeitz, line 135: not `r#"attributes\.server\.status"#.to_string()` ?

fulmicoton: no. the first "." is because of an object type. the second one is a quickwit field name with a "." in it.

PSeitz (approved):

Issue #1255, opened 2022-04-05
Expose available services via scuttlebutt


fulmicoton:
Nodes can expose indexer, searcher or both.

We want to expose the available services via scuttlebutt and use this information to populate the searcher pool accurately.

fulmicoton:
@evanxg852000 isn't this closed?

PR #1386, opened 2022-05-06
Available services through chitchat

evanxg852000:
Exposing available services via chitchat 
refer to : https://github.com/quickwit-oss/quickwit/issues/1255

### How was this PR tested?
 - unittest


quickwit-cluster/src/cluster.rs
```diff
@@ -95,13 +98,13 @@ impl From<Member> for NodeId {
...
 
-/// This is an implementation of a cluster using the SWIM protocol.
+/// This is an implementation of a cluster using Chitchat.
 pub struct Cluster {
     pub node_id: String,
     /// A socket address that represents itself.
     pub listen_addr: SocketAddr,
 
-    /// The actual cluster that implement the SWIM protocol.
+    /// The actual cluster that implement Chitchat.
```
guilload, line 107:
```suggestion
    /// The actual cluster that implements Chitchat.
```

quickwit-cluster/src/cluster.rs
```diff
@@ -132,7 +136,17 @@ impl Cluster {
...
             cluster_id,
-            vec![(GRPC_ADDRESS_KEY, grpc_addr)],
+            vec![
+                (GRPC_ADDRESS_KEY, grpc_addr.to_string()),
+                (
+                    AVAILABLE_SERVICES_KEY,
+                    services
+                        .iter()
+                        .map(|service| service.to_string())
+                        .collect::<Vec<_>>()
```
guilload, line 146: You can avoid allocating into a vec and joining directly using itertools.

quickwit-cluster/src/cluster.rs
```diff
@@ -132,7 +136,17 @@ impl Cluster {
...
             listen_addr.to_string(),
             cluster_id,
-            vec![(GRPC_ADDRESS_KEY, grpc_addr)],
+            vec![
+                (GRPC_ADDRESS_KEY, grpc_addr.to_string()),
+                (
+                    AVAILABLE_SERVICES_KEY,
+                    services
+                        .iter()
+                        .map(|service| service.to_string())
```
guilload, line 145: With itertools, we can also get away with `service.as_str()`.

quickwit-cluster/src/cluster.rs
```diff
@@ -193,10 +207,13 @@ impl Cluster {
         self.members.borrow().clone()
     }
 
-    /// Return the grpc addresses corresponding to the list of members.
-    pub async fn members_grpc_addresses(
+    /// Return the grpc addresses corresponding to the list of members
```
guilload, line 210: Returns the gRPC addresses of the members providing the specified service. Returns the addresses of all the members if no service is specified.

quickwit-cluster/src/cluster.rs
```diff
@@ -206,6 +223,22 @@ impl Cluster {
                 .node_state(&NodeId::from(member.clone()))
                 .unwrap();
 
+            let available_services = node_state
+                .get(AVAILABLE_SERVICES_KEY)
+                .ok_or_else(|| {
```
guilload, line 228: In which cases is this key missing? This would be a bug, right?

quickwit-cluster/src/cluster.rs
```diff
@@ -326,9 +368,16 @@ fn create_failure_detector_config_for_test() -> FailureDetectorConfig {
...
-pub fn create_cluster_for_test(seeds: &[String]) -> anyhow::Result<Cluster> {
+pub fn create_cluster_for_test(seeds: &[String], services: &[&str]) -> anyhow::Result<Cluster> {
     let peer_uuid = Uuid::new_v4().to_string();
-    let cluster = create_cluster_for_test_with_id(peer_uuid, "test-cluster".to_string(), seeds)?;
+    let services = services
+        .iter()
+        .map(|service_str| QuickwitService::try_from(*service_str))
+        .collect::<Result<Vec<_>, _>>()?
+        .into_iter()
+        .collect::<HashSet<_>>();
```
guilload, line 378: You can skip the `.collect::<Result<Vec<_>, _>>()?` above and directly collect into `.collect::<Result<HashSet<_>, _>>()`

quickwit-cluster/src/cluster.rs
evanxg852000: Yes, it has to be there

quickwit-cluster/src/cluster.rs
evanxg852000: Unless my rust-analyser is fooling me again. This was not working. I will check.

quickwit-cluster/src/cluster.rs
fulmicoton: It is not necessarily a bug.
If a user just joined, we might have been sending partial updates.
In that case, the member might be there, but we might not know its services.

We can try to frown really hard and try to find reason for it to never happen, but I think the right mental model here is that it might happen.

Returning an error here is not a good a idea.

quickwit-cluster/src/cluster.rs
```diff
@@ -206,6 +219,22 @@ impl Cluster {
...
+                        AVAILABLE_SERVICES_KEY,
+                        member.internal_id()
+                    )
+                })
+                .map(|services_str| {
+                    services_str
+                        .split(',')
+                        .map(QuickwitService::try_from)
+                        .collect::<Result<Vec<_>, _>>()
+                })??;
```
fulmicoton, line 236: The logic here is needlessly complicated.
Returning an error if the key is absent is an error... but assuming this is what you wanted, this is a very convoluted way to code this.

```
 let available_services_val = node_state
        .get(AVAILABLE_SERVICES_KEY)
        .ok_or_else(|| ...)?;
let services = parse_available_services_val(available_services_val)?;
```

is nicer for the reader.

quickwit-cluster/src/cluster.rs
```diff
@@ -216,6 +245,13 @@ impl Cluster {
                     )
                 })
                 .map(|addr_str| get_socket_addr(&addr_str))??;
+
+            if let Some(service) = &service_opt {
```
fulmicoton, line 249: I think I'd keep only one site to add the grpc_address to the list.
```
if let Some(service) = &service_opt {
    if !available_services.contains(service) {
        continue;
    }
}
```

quickwit-cluster/src/cluster.rs
guilload: There's also the case of rolling updates. During a rolling update, version X and version Y need to coexist. So if version Y adds a new service (maybe metastore or coordinator), version X will throw some errors when it probably should ignore the new service.

fulmicoton: Yes or an adversarial payload, a faulty dev server, etc. 
I think the rule of thumb is that whenever possible, we need to treat anything stuff coming from the network very cautiously.

In this case, a simple way to handle this is to assume that if the key is not there yet then we assume no service is exposed yet.

quickwit-cluster/src/cluster.rs
```diff
@@ -358,13 +401,64 @@ mod tests {
...
+        indexers.sort();
+        assert_eq!(indexers.len(), 2);
+        assert_eq!(indexers, expected_indexers);
+
+        let mut searchers = cluster1
+            .members_grpc_addresses_by_service(&members, Some(QuickwitService::Searcher))
+            .await?;
+        searchers.sort();
+        assert_eq!(searchers.len(), 2);
+        assert_eq!(searchers, expected_searchers);
```
fulmicoton, line 450: Let's make that 3 different test. the 3 cluster in a test is not a nice read.

quickwit-cluster/src/cluster.rs
```diff
@@ -193,10 +204,12 @@ impl Cluster {
         self.members.borrow().clone()
     }
 
-    /// Return the grpc addresses corresponding to the list of members.
-    pub async fn members_grpc_addresses(
+    /// Returns the gRPC addresses of the members providing the specified service.
+    /// Returns the addresses of all the members if no service is specified.
+    pub async fn members_grpc_addresses_by_service(
         &self,
         members: &[Member],
```
fulmicoton, line 211: I am not fond of the prototype of this API.

What is the point of the `members` arg? Do we call it for something else than the member of the cluster?

If possible I would rather have:

I'd rather have 
```
pub async fn members_grpc_addresses() -> Vec<SocketAddr>;
pub async fn members_grpc_addresses_for_service(service: QuickwitService) -> Vec<SocketAddr>;
```

evanxg852000: Unless we change this `members_grpc_addresses(&self) -> Vec<SocketAddr>;` to 
`members_grpc_addresses(&self) -> Vec<(SocketAddr, Vec<QuickwitService>)>; reusing it will be weird as we still need members to get the services, there is also locks involved this is why I preferred a single API

fulmicoton (approved):

fulmicoton (changes requested):
Can you break the test_cluster_available_service unit test?

evanxg852000:
Sorry, @fulmicoton  I don't understand your new change request. I replace the previous `test_cluster_available_service` with two tests `test_cluster_available_searcher` and `test_cluster_available_indexer`. Did you meant one of those?

fulmicoton (approved):

fulmicoton:
@evanxg852000 ah yes my bad.

Issue #1392, opened 2022-05-09
Revert default tokenizer for JSON field.s


fulmicoton:
I made "raw", the default tokenizer for JSON Fields.

While this is the most likely best default for a lot of users, it is also a nice foot gun.
We should use default (or the log tokenizer once we have one).

Issue #1370, opened 2022-05-05
Truncate tokens when using the raw tokenizer.


fulmicoton:
It can be easy for someone to shoot themselves in the foot with the raw tokenzier today.

PR #1393, opened 2022-05-09
skip long tokens in raw tokenizer

PSeitz:
update tantivy

skip tokens longer than 100 for raw tokenizer, set tokenizer when opening Index and in query_parser
change default tokenizer for JSON to the 'default' tokenizer

Fixes https://github.com/quickwit-oss/quickwit/issues/1392
Fixes https://github.com/quickwit-oss/quickwit/issues/1370

### DIFF
--- a/quickwit-doc-mapper/src/tokenizers.rs
+++ b/quickwit-doc-mapper/src/tokenizers.rs
@@ -1,0 +1,48 @@
+// Copyright (C) 2022 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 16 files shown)


quickwit-doc-mapper/src/tokenizers.rs
```diff
@@ -0,0 +1,48 @@
...
+    let my_haiku = r#"
+        white sandy beach
+        a strong wind is coming 
+        sand in my face
+        "#;
+    let my_long_text = "a text, that is just too long, no one will type it, no one will like it, \
+                        no one shall find it. I just need some more chars, now you may not pass.";
+
+    let tokenizer = get_quickwit_tokenizer_manager().get("raw").unwrap();
+    assert!(tokenizer.token_stream(my_haiku).advance());
```
fulmicoton, line 46: can you check that there is only one token for my haiku?

fulmicoton (approved):
Approving, but it would be nicer if you can edit your unit test.

PR #1398, opened 2022-05-10
update architecture page for sources

PSeitz:
Also Closes https://github.com/quickwit-oss/quickwit/issues/1277


docs/design/architecture.md
```diff
@@ -22,7 +22,12 @@ Quickwit gathers index metadata into a metastore to make them available across t
 
 In a clustered deployment, the metastore is typically a traditional RDBMS like PostgreSQL which we only support today. In a single-server deployment, it’s also possible to rely on a local file or on Amazon S3.
 
-### Distributed message queues
+### Source
```
PSeitz, line 25: Should we mention push api here?

fulmicoton: You can mention that it is one special type of source that does not need to be manually configured.

PR #1401, opened 2022-05-10
Doc on avoiding `.` in field names

evanxg852000:
Update docs to recommend avoiding `.` in field names

### DIFF
--- a/docs/reference/index-config.md
+++ b/docs/reference/index-config.md
@@ -361,6 +361,10 @@ In plain language:
 - it should only contain latin letter `[a-zA-Z]` digits `[0-9]` or (`.`, `-`, `_`).
 - the first character needs to be a letter.
 
+:::caution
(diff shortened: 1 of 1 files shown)


docs/reference/index-config.md
```diff
@@ -361,6 +361,10 @@ In plain language:
 - it should only contain latin letter `[a-zA-Z]` digits `[0-9]` or (`.`, `-`, `_`).
 - the first character needs to be a letter.
 
+:::caution
+For field names containing the `.` character, you will need to escape it when referencing them. Otherwise the `.` character will be interpreted as a JSON object property access. Because of this, it is recommended to avoid using field names containing the `.` character.
```
fulmicoton, line 365:
```suggestion
For field names containing the `.` character, you will need to escape it when referencing them in queries. Otherwise the `.` character will be interpreted as a JSON object property access. Because of this, it is recommended to avoid using field names containing the `.` character.
```

fulmicoton (approved):

Issue #1219, opened 2022-03-25
Binaries packaging broken


evanxg852000:
Nightly builds broken on musl build and need to be fixed.
Most likely a docker image update with Rockdbs dependencies (LLVM)

https://github.com/quickwit-oss/quickwit/actions/workflows/publish_nightly_packages.yml

fulmicoton:
Closing: same discussion in  #1221

evanxg852000:
@fulmicoton Re-opening this as we need to fix the binaries build for Rcksdb inclusion libs

PR #1425, opened 2022-05-11
Fix binaries build

evanxg852000:
Fixes builds:
- x86_64-unknown-linux-gnu
- aarch64-unknown-linux-gnu

Commented both `musl` builds: 
There seem to be conflicting statements about RocksDB compilation with musl (possible/impossible). 
An investigation is being carried out in the [branch](https://github.com/quickwit-oss/quickwit/tree/fix-1219-musl). The build step is ok but the linking step fails. I suspect this can take a bit longer so let's have the other build and our release workflow at least working.

Part of https://github.com/quickwit-oss/quickwit/issues/1219


build/cross-images/x86_64-unknown-linux-musl.dockerfile
```diff
@@ -11,3 +11,5 @@ RUN echo "Upgrading CMake" && \
 ENV CC=musl-gcc \
     CFLAGS=-I/usr/local/musl/include \
     LIB_LDFLAGS=-L/usr/lib/x86_64-linux-gnu
+
+RUN sudo chown -R rust:rust /opt/rust/
```
guilload, line 15: Just curious, why is this needed? Also, do you need `sudo`, I believe docker containers are built and run using the user `root`.

build/cross-images/aarch64-unknown-linux-gnu.dockerfile
```diff
@@ -4,7 +4,8 @@ COPY libpq.sh /
 RUN bash /libpq.sh aarch64-linux-gnu
 
 RUN apt-get update -y && \
-    apt-get install -y libpq-dev && \
+    apt-get install -y libpq-dev libclang-3.9-dev clang-3.9 \
```
guilload, line 7: nit: can we have one dependency per line and sort the dependencies

build/cross-images/x86_64-unknown-linux-gnu.dockerfile
```diff
@@ -2,5 +2,6 @@ FROM quickwit/cross-base:x86_64-unknown-linux-gnu
 # See https://github.com/quickwit-inc/cross
 
 RUN apt-get update && \
-    apt-get install -y zlib1g-dev libssl-dev libpq-dev && \
+    apt-get install -y zlib1g-dev libssl-dev \
```
guilload, line 5: ditto

guilload (approved):
So if I understood correctly, what was missing to build RocksDB is clang?

build/cross-images/x86_64-unknown-linux-musl.dockerfile
guilload: No worries :+1:

evanxg852000: On the other hand for the dependencies yes, RocksDB need LLVM and Clang to work properly

Issue #1424, opened 2022-05-11
Search with sort by on a fast field different from the sort by field defined at the index level is not working


fmassot:
I think @kstaken encountered this issue, here is the reported error:

```
curl 'http://127.0.0.1:7280/api/v1/noaa3/search?query=station_id:01112099999&sort_by_field=temperature_c'

{
  "InternalError": "(Internal error: `An IO error occurred: 'Unsupported operation. StorageDirectory only supports async reads: \"76d65a1a4065404b83d8900cf52cf38e.fast\"'`., split_id: 01G2T0S8H072A2PNFHKYR7Y87K)"
}
```

This is due to the fact that Quickwit does not prefetch fast field values on the requested sort_by field.

Here is the rust function that extracts the fast field names for the warmup:

```rust
fn extract_fast_field_names(doc_mapper: &dyn DocMapper) -> HashSet<String> {
    let mut fast_fields = HashSet::new();
    if let Some(timestamp_field) = doc_mapper.timestamp_field_name() {
        fast_fields.insert(timestamp_field);
    }
    if let SortBy::FastField { field_name, .. } = doc_mapper.sort_by() {
        fast_fields.insert(field_name);
    }
    fast_fields
}
```

PR #1434, opened 2022-05-12
Fix query sortBy on other fast fields

evanxg852000:
Included query sort field in the list of fast_fields to prefetch

Closes https://github.com/quickwit-oss/quickwit/issues/1424

### DIFF
--- a/quickwit-search/src/lib.rs
+++ b/quickwit-search/src/lib.rs
@@ -273,7 +273,7 @@ mod tests {
     use assert_json_diff::assert_json_include;
     use quickwit_doc_mapper::DefaultDocMapper;
     use quickwit_indexing::TestSandbox;
-    use quickwit_proto::LeafHit;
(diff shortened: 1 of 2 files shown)


fmassot (approved):

quickwit-search/src/lib.rs
```diff
@@ -494,6 +494,62 @@ mod tests {
...
+            sort_by_field: Some("temperature".to_string()),
+            ..Default::default()
+        };
+        let single_node_response = single_node_search(
+            &search_request,
+            &*test_sandbox.metastore(),
+            test_sandbox.storage_uri_resolver(),
+        )
+        .await?;
+        assert_eq!(single_node_response.num_hits, 30);
```
PSeitz, line 548: can you add an assert that the results are sorted?

evanxg852000: So @PSeitz, great suggestion as I found another bug when using `Asc` sorting

PR #1444, opened 2022-05-13
PushAPI body Limit

evanxg852000:
Added a limit on push API body allowed size.
Closes https://github.com/quickwit-oss/quickwit/issues/1446


PSeitz (approved):
20MB could be a little small for scenarios where the documents are very large

evanxg852000:
@PSeitz  10M is the value suggested https://github.com/quickwit-oss/quickwit/issues/1446

fulmicoton (approved):

Issue #1440, opened 2022-05-13
Document Push API


PSeitz:

evanxg852000:
@fmassot should this be part of the `reference/rest-api.md` page or should we have another page for it?

PR #1453, opened 2022-05-15
Docs for ingest API 

evanxg852000:
Closes https://github.com/quickwit-oss/quickwit/issues/1440

### DIFF
--- a/docs/reference/rest-api.md
+++ b/docs/reference/rest-api.md
@@ -146,7 +146,7 @@ The response is a JSON object, and the content type is `application/json; charse
 | **num_ingested_docs**   | Total number of documents ingested | `number`   |
 
 
-### Ingest data with elastic compatible API
+### Ingest data with Elasticsearch compatible API
 
 ```
 POST api/v1/_bulk -d \
--- a/docs/reference/rest-api.md
+++ b/docs/reference/rest-api.md
@@ -158,7 +158,7 @@ POST api/v1/_bulk -d \
 {"url":"https://en.wikipedia.org/wiki?id=3","title":"baz","body":"baz"}'
 ```
 
-Ingest a batch of documents to make them searchable using the [Elastic Search](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) bulk API. This endpoint provides compatibility with tools or systems that already send data to Elastic Search for indexing. Currently, only the `create` action of the elastic bulk API is supported, all other actions such as `delete` or `update` are ignored.
+Ingest a batch of documents to make them searchable using the [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) bulk API. This endpoint provides compatibility with tools or systems that already send data to Elasticsearch for indexing. Currently, only the `create` action of the bulk API is supported, all other actions such as `delete` or `update` are ignored.
 
 :::info
 The payload size is limited to 10MB as this endpoint is intended to receive documents in batch.
--- a/docs/reference/rest-api.md
+++ b/docs/reference/rest-api.md
@@ -60,7 +60,7 @@ Search for documents matching a query in the given index `<index id>`.
 
 #### Response
 
-The response for the is a JSON object, and the content type is `application/json; charset=UTF-8.`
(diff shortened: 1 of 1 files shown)


docs/reference/rest-api.md
```diff
@@ -114,3 +114,60 @@ The formatting is based on the specified output format.
 
 On error, an "X-Stream-Error" header will be sent via the trailers channel with information about the error, and the stream will be closed via [`sender.abort()`](https://docs.rs/hyper/0.14.16/hyper/body/struct.Sender.html#method.abort).
 Depending on the client, the trailer header with error details may not be shown. The error will also be logged in quickwit ("Error when streaming search results").
+
+
+### Ingest data in an index
+
+```
+POST api/v1/<index id>/ingest -d \
```
fmassot, line 122: Does it work with 'curl -d'? Or should we put `--data-binary`?

evanxg852000: Yes it's curl flag, this was a tip from @PSeitz `-d` does not load the whole file in memory.
Maybe we can covert all the verbs in curl?

docs/reference/rest-api.md
```diff
@@ -114,3 +114,60 @@ The formatting is based on the specified output format.
...
+POST api/v1/_bulk -d \
+'{ "create" : { "_index" : "wikipedia", "_id" : "1" } }
+{"url":"https://en.wikipedia.org/wiki?id=1","title":"foo","body":"foo"}
+{ "create" : { "_index" : "wikipedia", "_id" : "2" } }
+{"url":"https://en.wikipedia.org/wiki?id=2","title":"bar","body":"bar"}
+{ "create" : { "_index" : "wikipedia", "_id" : "3" } }
+{"url":"https://en.wikipedia.org/wiki?id=3","title":"baz","body":"baz"}'
+```
+
+Ingest a batch of documents to make them searchable using the [Elastic Search](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) bulk API. This endpoint provides compatibility with tools or systems that already send data to Elastic Search for indexing. Currently, only the `create` action of the elastic bulk API is supported, all other actions such as `delete` or `update` are ignored.
```
fmassot, line 161:
```suggestion
Ingest a batch of documents to make them searchable using the [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) bulk API. This endpoint provides compatibility with tools or systems that already send data to Elasticsearch for indexing. Currently, only the `create` action of the bulk API is supported, all other actions such as `delete` or `update` are ignored.
```

docs/reference/rest-api.md
```diff
@@ -114,3 +114,60 @@ The formatting is based on the specified output format.
...
+#### Response
+
+The response is a JSON object, and the content type is `application/json; charset=UTF-8.`
+
+| Field                   | Description                        | Type       |
+| --------------------    | ---------------------------------- | :--------: |
+| **num_ingested_docs**   | Total number of documents ingested | `number`   |
+
+
+### Ingest data with elastic compatible API
```
fmassot, line 149:
```suggestion
### Ingest data with Elasticsearch compatible API
```

docs/reference/rest-api.md
```diff
@@ -114,3 +114,60 @@ The formatting is based on the specified output format.
 
 On error, an "X-Stream-Error" header will be sent via the trailers channel with information about the error, and the stream will be closed via [`sender.abort()`](https://docs.rs/hyper/0.14.16/hyper/body/struct.Sender.html#method.abort).
 Depending on the client, the trailer header with error details may not be shown. The error will also be logged in quickwit ("Error when streaming search results").
+
+
+### Ingest data in an index
```
fmassot, line 119: Can we add something like "This endpoint is only available if you query on a node that is running an indexer service is running."?

Can you add such a comment also on the search endpoint? For the search though, it will work as long as you have one searcher in the cluster.

fmassot (approved):
Almost good! 

Final ask: can you fix the search URL? 
The `GET api/v1/indexes/<index id>/search?query=searchterm` is wrong and should be replaced by 'GET api/v1/<index id>/search?query=searchterm'

Issue #1449, opened 2022-05-13
Performance falls rapidly as max_hits increases in size


kstaken:
Was poking around trying to isolate some weird sorting behavior and noticed that query times increase rapidly as the size of max_hits goes up. This is running a build from `main`, single node with local file system storage.

Index info
```
1. General infos
===============================================================================
Index id:                           noaa-1m
Index uri:                          file:///Users/kstaken/projects/opensource/quickwit-test/qwdata/indexes/noaa-1m
Number of published splits:         2
Number of published documents:      1634467
Size of published splits:           84 MB
```

```
$ curl -sS 'http://127.0.0.1:7280/api/v1/noaa-1m/search?query=name:NO&max_hits=10' | jq .elapsed_time_micros
46021
$ curl -sS 'http://127.0.0.1:7280/api/v1/noaa-1m/search?query=name:NO&max_hits=100' | jq .elapsed_time_micros
68718
$ curl -sS 'http://127.0.0.1:7280/api/v1/noaa-1m/search?query=name:NO&max_hits=1000' | jq .elapsed_time_micros
252134
$ curl -sS 'http://127.0.0.1:7280/api/v1/noaa-1m/search?query=name:NO&max_hits=10000' | jq .elapsed_time_micros
2832835
$ curl -sS 'http://127.0.0.1:7280/api/v1/noaa-1m/search?query=name:NO&max_hits=100000' | jq .elapsed_time_micros
null
```

For the 100,000 record case the server panics as it's out of file descriptors.

```
2022-05-13T18:36:10.966Z ERROR fetch_docs: quickwit_search::fetch_docs: Error when fetching docs in splits. split_ids=["01G2Z5RC4880BTWNSKX3P6YM9N", "01G2Z4B7QHBRE31G3203BWVV6N"] error=searcher-doc-async

Caused by:
    0: An IO error occurred: 'Too many open files (os error 24)'
    1: Too many open files (os error 24)
2022-05-13T18:36:13.817Z ERROR fetch_docs: quickwit_search::fetch_docs: Error when fetching docs in splits. split_ids=["01G2Z5RC4880BTWNSKX3P6YM9N", "01G2Z4B7QHBRE31G3203BWVV6N"] error=searcher-doc-async

Caused by:
    0: An IO error occurred: 'Too many open files (os error 24)'
    1: Too many open files (os error 24)
```

PSeitz:
Thanks for the bug report.
Do you have a use case to retrieve that many results?

[...]

PR #1455, opened 2022-05-16
limit concurrency of fetching docs

PSeitz:
Limits the concurrency of fetching docs to 10

Addresses #1449

### DIFF
--- a/quickwit-search/src/fetch_docs.rs
+++ b/quickwit-search/src/fetch_docs.rs
@@ -21,6 +21,7 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use anyhow::Context;
+use futures::stream::{StreamExt, TryStreamExt};
 use itertools::Itertools;
 use quickwit_proto::{FetchDocsResponse, PartialHit, SplitIdAndFooterOffsets};
 use quickwit_storage::Storage;
@@ -33,11 +34,11 @@ use crate::GlobalDocAddress;
 /// Given a list of global doc address, fetches all the documents and
 /// returns them as a hashmap.
 #[allow(clippy::needless_lifetimes)]
-async fn fetch_docs_to_map<'a>(
-    mut global_doc_addrs: Vec<GlobalDocAddress<'a>>,
+async fn fetch_docs_to_map(
+    mut global_doc_addrs: Vec<GlobalDocAddress>,
     index_storage: Arc<dyn Storage>,
     splits: &[SplitIdAndFooterOffsets],
-) -> anyhow::Result<HashMap<GlobalDocAddress<'a>, String>> {
+) -> anyhow::Result<HashMap<GlobalDocAddress, String>> {
     let mut split_fetch_docs_futures = Vec::new();
 
     let split_offsets_map: HashMap<&str, &SplitIdAndFooterOffsets> = splits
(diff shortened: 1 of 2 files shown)


quickwit-search/src/fetch_docs.rs
```diff
@@ -46,16 +47,16 @@ async fn fetch_docs_to_map<'a>(
         .collect();
 
     // We sort global hit addrs in order to allow for the grouby.
-    global_doc_addrs.sort_by_key(|global_doc_addr| global_doc_addr.split);
```
fulmicoton, line 49: I'm lost. Why is it better than sort_by_key?

PSeitz:
Removed reference from GlobalDocAddress, since I got a weird compiler error (compiler bug?) otherwise

```
error: implementation of `FnOnce` is not general enough
   --> quickwit-search/src/service.rs:151:43
    |
151 |       ) -> crate::Result<FetchDocsResponse> {
    |  ___________________________________________^
152 | |         let storage = self
153 | |             .storage_uri_resolver
154 | |             .resolve(&fetch_docs_request.index_uri)?;
...   |
163 | |         Ok(fetch_docs_response)
164 | |     }
    | |_____^ implementation of `FnOnce` is not general enough
    |
    = note: closure with signature `fn(GlobalDocAddress<'0>) -> impl futures::Future<Output = [async output]>` must implement `FnOnce<(GlobalDocAddress<'1>,)>`, for any two lifetimes `'0` and `'1`...
    = note: ...but it actually implements `FnOnce<(GlobalDocAddress<'_>,)>`
```

fulmicoton:
@PSeitz Can you confirm having tested it on a large number of docs?

PSeitz:
> @PSeitz Can you confirm having tested it on a large number of docs?

Yes, I can confirm it works

fulmicoton:
@PSeitz can you confirm that at no point we debounce read calls to docs from a same docstore block, and open an issue accordingly? This could have massive performance gain for @kstaken use case described in #1449

fulmicoton (approved):

fulmicoton:
@PSeitz 

> With this change we still may load the same block multiple times, but it's more unlikely. The current state is
> 10 concurrent doc requests (no optimization on same blocks), a cache of 100 and e.g. 10_000 docs.

Are you talking about tantivy's docstore cache?
Is this something you actually observed? I think we end up loading the same block most of the time.

I don't think there is any debouncing logic anywhere.
Let's assume we are loading 10 docs from the same block.
In that case, we do the fetch request all at the same time, which result in a cache miss and all request are emitted.


Let's assume we do 10k requests in random block, with big overlaps.
In that case the cache, the cache is not big enough to deal with this and we experience mostly cache misses.
We probably want to debounce requests at the storage layer.

PSeitz:
> Are you talking about tantivy's docstore cache?
Is this something you actually observed? I think we end up loading the same block most of the time.

Yes tantivy's docstore cache. I did not observe it.

> I don't think there is any debouncing logic anywhere.
Let's assume we are loading 10 docs from the same block.
In that case, we do the fetch request all at the same time, which result in a cache miss and all request are emitted.

For TOP 10 docs it will have no effect, since they are executed concurrently.

> Let's assume we do 10k requests in random block, with big overlaps.
In that case the cache, the cache is not big enough to deal with this and we experience mostly cache misses.
We probably want to debounce requests at the storage layer.

Yes, but also depends on how many docs are there per block.

I was thinking about collecting the block ranges from the docs and then load them only once. 
On the storage layer it would also be possible with a cache or some timeout collector on the same byte range. This could have some synergies, but we could since have similar issues with cache invalidation.

PR #1456, opened 2022-05-16
Default metastore is polling the filesystem.

### LINKED ISSUE #1432: The default metastore should be polling
By default, the metastore used is a file backed metastore that has its root in `qwdata/indexes`.

We should set polling by default for that default metastore.
This default is probably the cause for #1430.

fulmicoton:
Closes #1432

### DIFF
--- a/config/quickwit.yaml
+++ b/config/quickwit.yaml
@@ -29,16 +29,16 @@ version: 0
 #
 #data_dir: /path/to/data
 #
-# Metastore URI. Default to `data_dir/indexes#polling_interval=30s`
+# Metastore URI. Defaults to `data_dir/indexes#polling_interval=30s`
 # which is a file backed metastore and mostly convenient for testing. A cluster would
 # require a metastore backed by S3 or postgreqsl.
 #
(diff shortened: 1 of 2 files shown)


config/quickwit.yaml
```diff
@@ -29,12 +29,18 @@ version: 0
...
+# Metastore URI. Default to `data_dir/indexes#polling_interval=30s`
+# which is a file backed metastore and mostly convenient for testing. A cluster would
+# require a metastore backed by S3 or postgreqsl.
 #
-#metastore_uri: s3://your-bucket/indexes
-#metastore_uri: postgres://username:password@host:port/db
+# metastore_uri: s3://your-bucket/indexes
+# metastore_uri: postgres://username:password@host:port/db
+#
+# When using a file backed metastore, the state of the metastore will be cached forever
```
fmassot, line 39: # When using a file-backed metastore, the state of the metastore will be cached forever.
# If you are indexing and search on a different process, it is possible to periodically
# refresh the state of the metastore on the searcher using a `polling_interval` hashtag.
#

fmassot (approved):

PR #1472, opened 2022-05-17
check aggregation request on root node

### LINKED ISSUE #1464: Handle errors on aggregation payload deserialization
Currently, we deserialize the aggregation payload in `make_collector_for_split` and use `unwrap`.
https://github.com/quickwit-oss/quickwit/blob/main/quickwit-search/src/collector.rs#L450

We should:
- handle the deserialization error so that the user can get a proper response
- validate the aggregation payload at the root node,  `make_collector_for_split` is called at the leaf level.

Error trace on the server side:

```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Error("data did not match any variant of untagged enum Aggregation", line: 1, column: 127)', quickwit-search/src/collector.rs:444:48
```

The server returns this response:
```
{
  "InternalError": "transport error"
}
```

PSeitz:
fixes #1464


fulmicoton:
@PSeitz can you add a unit test to make sure the error message will make sense for users?

fulmicoton (approved):

PSeitz:
> @PSeitz can you add a unit test to make sure the error message will make sense for users?

I added a test and created an error. Error messages suck currently, because we we use the `#[serde(untagged)]`, since flatten is not supported on enums.

```
#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
#[serde(untagged)]
pub enum Aggregation {
    /// Bucket aggregation, see [BucketAggregation] for details.
    Bucket(BucketAggregation),
    /// Metric aggregation, see [MetricAggregation] for details.
    Metric(MetricAggregation),
}
```

PR #1491, opened 2022-05-19
Add a search service to address REST requests on indexer only nodes.

### LINKED ISSUE #1482: Investigate: Search REST request on an indexer
A node running indexing only might receive search request via the interface.

See if it is possible to easily address those by redirecting it to a search server.

fulmicoton:
We add a search service to address REST requests in the UI of index-only
nodes as quickfix for #1482.
It will only address the root part of the requests.

That search service is not mounted as a gRPC service and hence will not address
leaf requests.

We also return an explicit error if there are no search nodes in the cluster.
(before we were panicking).

Closes #1482

### DIFF
--- a/quickwit-serve/src/grpc.rs
+++ b/quickwit-serve/src/grpc.rs
@@ -42,7 +42,7 @@ pub(crate) async fn start_grpc_server(
     let grpc_cluster_service = GrpcClusterAdapter::from(quickwit_services.cluster_service.clone());
     let mut server_router = server.add_service(ClusterServiceServer::new(grpc_cluster_service));
 
-    // We only mount the gRPC servcice if the searcher is enabled on this node.
+    // We only mount the gRPC service if the searcher is enabled on this node.
     if quickwit_services
         .services
         .contains(&QuickwitService::Searcher)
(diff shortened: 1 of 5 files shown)


quickwit-serve/src/grpc.rs
```diff
@@ -41,7 +42,12 @@ pub(crate) async fn start_grpc_server(
     let grpc_cluster_service = GrpcClusterAdapter::from(quickwit_services.cluster_service.clone());
     let mut server_router = server.add_service(ClusterServiceServer::new(grpc_cluster_service));
 
-    if let Some(search_service) = quickwit_services.search_service.clone() {
+    // We only mount the gRPC servcice if the searcher is enabled on this node.
```
fmassot, line 45:
```suggestion
    // We only mount the gRPC service if the searcher is enabled on this node.
```

fmassot (approved):
LGTM. 

Would be nice to have a test with 1 indexer + 1 searcher + the case where there is no searcher at all. I can open an issue on that. I will test it anyway with the UI.

fulmicoton:
@fmassot i opened a separate ticket for th unit test #1494. It might be trickier than the PR itself.

PR #1501, opened 2022-05-20
Redirects / to your ui/search.

### LINKED ISSUE #1383: "/" should serve the admin
Right now "/" is not a valid route on our HTTP server.

One has to type `HOSTNAME:PORT/ui/` to get an admin UI.

/ should give back index.html.
all of the static content should only be accessible under ui/ or static/ (or whatever)

fulmicoton:
@fmassot maybe a redirect will be a decent quick and dirty solution?

Closes #1383


fmassot:
Actually, I think this is the right approach for us.

fmassot (approved):

PR #1506, opened 2022-05-22
Docker image build 

evanxg852000:
Adds docker image build to ci 
Part of https://github.com/quickwit-oss/quickwit/issues/1219

@fmassot  I think what we are missing is having the binaries build for `.github/workflows/publish_release_packages.yml`
Should I add it here?

### DIFF
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -1,0 +1,47 @@
+name: 'Build Quickwit docker image'
+description: 'Build & push docker image'
+inputs:
+  version:
+    description: 'Version'
+    required: false
+runs:
+  using: "composite"
+  steps:
+    - run: echo "IMAGE_ID=quickwit/quickwit-${{ inputs.version }}" >> $GITHUB_ENV
+      shell: bash
+    - run: echo "IMAGE_DATE=$(date '+%Y-%m-%d')" >> $GITHUB_ENV
+      shell: bash
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 16
+        cache: 'yarn'
+        cache-dependency-path: quickwit-ui/yarn.lock
+    - run: make build-ui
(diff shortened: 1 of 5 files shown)


fmassot:
we can first merge the PR to have the nightly docker image and have another PR for the release, as you like

.github/actions/build-docker/action.yml
```diff
@@ -0,0 +1,53 @@
+name: 'Build Quickwit docker image'
+description: 'Build & push docker image'
+inputs:
+  version:
+    description: 'Version'
+    required: false
```
fmassot, line 6: Not sure why version is not required

.github/actions/build-docker/action.yml
```diff
@@ -0,0 +1,53 @@
...
+      run: echo "IMAGE_ID=quickwit/quickwit" >> $GITHUB_ENV
+      shell: bash
+    - run: echo "IMAGE_DATE=$(date '+%Y-%m-%d')" >> $GITHUB_ENV
+      shell: bash
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 16
+        cache: 'yarn'
+        cache-dependency-path: quickwit-ui/yarn.lock
+    - run: make build-ui
```
fmassot, line 25: For this build, I guess it would be nice to include the UI build inside the Dockerfile as we are doing everything there.

fmassot (approved):

.github/actions/build-docker/action.yml
evanxg852000: ok

.github/actions/build-docker/action.yml
evanxg852000: For production build we don't specify `version` in order to have an image `quickwit/quickwit` as opposed to `quickwit/quickwit-nightly`

PR #1507, opened 2022-05-23
Garbage Collect IngestAPI queues

evanxg852000:
Implements the ingest API queue garbage collector

### How was this PR tested?
- unittest

Closes https://github.com/quickwit-oss/quickwit/issues/1229

### DIFF
--- a/quickwit-indexing/src/actors/indexing_service.rs
+++ b/quickwit-indexing/src/actors/indexing_service.rs
@@ -38,15 +38,15 @@ use thiserror::Error;
 use tracing::{error, info};
 
 use crate::models::{
-    DetachPipeline, IndexingPipelineId, Observe, ObservePipeline, SpawnMergePipeline,
-    SpawnPipeline, SpawnPipelinesForIndex,
+    DetachPipeline, IndexingPipelineId, Observe, ObservePipeline, ShutdownPipeline,
+    SpawnMergePipeline, SpawnPipeline, SpawnPipelinesForIndex,
 };
 use crate::{IndexingPipeline, IndexingPipelineParams, IndexingStatistics};
 
 pub const INDEXING: &str = "indexing";
 
 /// Reserved source ID used for the ingest API.
-const INGEST_API_SOURCE_ID: &str = ".ingest-api";
+pub const INGEST_API_SOURCE_ID: &str = ".ingest-api";
 
 #[derive(Error, Debug)]
 pub enum IndexingServiceError {
@@ -293,6 +293,23 @@ impl IndexingService {
         let index_metadata = self.metastore.index_metadata(index_id).await?;
         Ok(index_metadata)
     }
+
+    async fn shutdown_pipeline(
+        &mut self,
+        index_id: String,
+        source_id: String,
+        _ctx: &ActorContext<Self>,
+    ) -> Result<(), IndexingServiceError> {
+        let pipeline_id = IndexingPipelineId {
+            index_id,
+            source_id,
+        };
+        let pipeline_handle_opt = self.pipeline_handles.remove(&pipeline_id);
+        if let Some(pipeline_handle) = pipeline_handle_opt {
+            pipeline_handle.quit().await;
+        }
+        Ok(())
+    }
 }
 
 #[async_trait]
@@ -424,8 +441,22 @@ impl Handler<SpawnPipelinesForIndex> for IndexingService {
         &mut self,
         message: SpawnPipelinesForIndex,
         ctx: &ActorContext<Self>,
-    ) -> Result<Result<Vec<IndexingPipelineId>, IndexingServiceError>, ActorExitStatus> {
+    ) -> Result<Self::Reply, ActorExitStatus> {
         Ok(self.spawn_pipelines(message.index_id, ctx).await)
+    }
+}
+
+#[async_trait]
+impl Handler<ShutdownPipeline> for IndexingService {
+    type Reply = Result<(), IndexingServiceError>;
+    async fn handle(
+        &mut self,
+        message: ShutdownPipeline,
+        ctx: &ActorContext<Self>,
+    ) -> Result<Self::Reply, ActorExitStatus> {
+        Ok(self
+            .shutdown_pipeline(message.index_id, message.source_id, ctx)
+            .await)
     }
 }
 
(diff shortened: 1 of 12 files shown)


quickwit-ingest-api/src/queue.rs
```diff
@@ -264,6 +264,12 @@ impl Queues {
             doc_batch: Some(doc_batch),
         })
     }
+
+    pub fn list_queues(&self) -> crate::Result<ListQueuesResponse> {
+        Ok(ListQueuesResponse {
```
fulmicoton, line 269: Why result then?

quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -293,6 +293,23 @@ impl IndexingService {
...
+        &mut self,
+        index_id: String,
+        source_id: String,
+        _ctx: &ActorContext<Self>,
+    ) -> Result<(), IndexingServiceError> {
+        let pipeline_id = IndexingPipelineId {
+            index_id,
+            source_id,
+        };
+        let pipeline_handle_opt = self.pipeline_handles.remove(&pipeline_id);
```
fulmicoton, line 307: Considering how little it does, I'd put that logic directly in the handler.

In particular, the passage from Shutdown `message -> (index_id, source_id) positional argument -> IndexingPipelineId` is offering 2 spots where a bug could be hidden from the compiler and the reader view (positional arguments inverted).

To mitigate, you could also have a
`struct Shutdown(IndexingPipelineId);`

quickwit-indexing/src/actors/ingest_api_garbage_collector.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    async fn delete_queue(&self, queue_id: &str) -> Result<(), ActorExitStatus> {
+        // shutdown the pipeline if any
+        self.indexing_service
+            .ask_for_res(ShutdownPipeline {
+                index_id: queue_id.to_string(),
+                source_id: INGEST_API_SOURCE_ID.to_string(),
+            })
+            .await
+            .map_err(|error| anyhow::anyhow!(error))?;
```
fulmicoton, line 78: This actor is unsupervized. It means that if an error is emitted in a handler, the actor will terminated and will not be restarted again.

We may want to be careful about the meaning and the effect of returning an error here.

I suggest we just log an error if that happens.

quickwit-indexing/src/actors/ingest_api_garbage_collector.rs
```diff
@@ -0,0 +1,146 @@
...
+    ) -> Self {
+        Self {
+            metastore,
+            ingest_api_service,
+            indexing_service,
+            counters: IngestApiGarbageCollectorCounters::default(),
+        }
+    }
+
+    async fn delete_queue(&self, queue_id: &str) -> Result<(), ActorExitStatus> {
```
fulmicoton, line 70: Return anyhow::Result here

quickwit-indexing/src/actors/ingest_api_garbage_collector.rs
```diff
@@ -0,0 +1,146 @@
...
+        // shutdown the pipeline if any
+        self.indexing_service
+            .ask_for_res(ShutdownPipeline {
+                index_id: queue_id.to_string(),
+                source_id: INGEST_API_SOURCE_ID.to_string(),
+            })
+            .await
```
fulmicoton, line 77: could be simply `.await`? if we return anyhow::Result.

quickwit-indexing/src/actors/ingest_api_garbage_collector.rs
```diff
@@ -0,0 +1,146 @@
...
+
+#[async_trait]
+impl Handler<Loop> for IngestApiGarbageCollector {
+    type Reply = ();
+
+    async fn handle(&mut self, _: Loop, ctx: &ActorContext<Self>) -> Result<(), ActorExitStatus> {
+        info!("ingest-api-garbage-collect-operation");
+        self.counters.num_passes += 1;
+
+        let queues: HashSet<_> = self
```
fulmicoton, line 120: The type of the element in the HashSet would be helpful here.

quickwit-indexing/src/actors/ingest_api_garbage_collector.rs
```diff
@@ -0,0 +1,146 @@
...
+            .ingest_api_service
+            .ask_for_res(ListQueuesRequest {})
+            .await
+            .map_err(|error| anyhow::anyhow!(error))?
+            .queues
+            .into_iter()
+            .collect();
+
+        let indexes: HashSet<_> = self
```
fulmicoton, line 129: The type of the element in Hashset would be helful

quickwit-indexing/src/actors/ingest_api_garbage_collector.rs
```diff
@@ -0,0 +1,146 @@
...
+            .metastore
+            .list_indexes_metadatas()
+            .await
+            .map_err(|error| anyhow::anyhow!(error))?
+            .into_iter()
+            .map(|index_metadata| index_metadata.index_id)
+            .collect();
+
+        for queue_id in queues.difference(&indexes) {
+            self.delete_queue(queue_id).await?;
```
fulmicoton, line 139: catch the error and log but do no return an error.

evanxg852000:
@fmassot @fulmicoton  Another thing is RocksDB has a default queue that cannot be deleted. if an index is named `default`, the queue will be there all the time. we should guard against this.

fmassot:
As discussed, a restart is always needed on deletion with the CLI with the file-backed metastore.

To avoid the clash with the `default` queue, I would always use `queue_{index-id}` as the name of the queue, but I really don't know if it can cause some issues. If it's too risky, we can just forbid the index name `default` for the 0.3 and remove that constraint later.

quickwit-indexing/src/actors/ingest_api_garbage_collector.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    async fn handle(&mut self, _: Loop, ctx: &ActorContext<Self>) -> Result<(), ActorExitStatus> {
+        info!("ingest-api-garbage-collect-operation");
+        self.counters.num_passes += 1;
+
+        let queues: HashSet<String> = self
+            .ingest_api_service
+            .ask_for_res(ListQueuesRequest {})
+            .await
+            .map_err(|error| anyhow::anyhow!(error))?
```
fulmicoton, line 126: If an error is returned here, the actor will stop running for good today.

evanxg852000:
I think it's safe to internally prefix our queues so I went with `.queue_{index_id}`

fulmicoton (approved):

PR #1539, opened 2022-05-24
set max value for start_offset

PSeitz:
```
➜  quickwit-indices curl -sS 'http://localhost:7280/api/v1/hdfs/search?query=*&max_hits=0&start_offset=40000000'
{
  "InvalidArgument": "max value for start_offset is 10_000, but got 40000000"
}⏎
```


fulmicoton (approved):

fulmicoton:
Can we add a unit test, or create a ticket to add a unit test?

Issue #1525, opened 2022-05-23
Quickwit can't support Elasticsearch _bulk `create` semantics


kstaken:
I saw in the documentation for the Elasticsearch _bulk ingest API that the only method that is supported is `create` however I believe Quickwit can't actually implement the semantics required for `create`. The action it should be supporting is `index`. 

`create` is used when you want to store a document only if it doesn't already exist in the index. It's a really handy operation but without the ability to quickly tell what docs already exist it can't be implemented correctly.

`index` simply indexes the document. Note: if `_id` is provided to the request this will also be semantically incorrect as that tells ES to reindex and replace the document but in Quickwit it will result in a duplicate document being indexed.

fulmicoton:
confirmed

evanxg852000:
@fmassot  should switch to `index` and document?

PR #1547, opened 2022-05-25
Elasticsearch create semantic note

evanxg852000:
I amended the docs following the discussion on discord.
Closes https://github.com/quickwit-oss/quickwit/issues/1525
We can still switch to `index` action instead of `create` in this PR @fulmicoton

### DIFF
--- a/docs/reference/rest-api.md
+++ b/docs/reference/rest-api.md
@@ -159,7 +159,8 @@ POST api/v1/_bulk -d \
 {"url":"https://en.wikipedia.org/wiki?id=3","title":"baz","body":"baz"}'
 ```
 
-Ingest a batch of documents to make them searchable using the [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) bulk API. This endpoint provides compatibility with tools or systems that already send data to Elasticsearch for indexing. Currently, only the `create` action of the bulk API is supported, all other actions such as `delete` or `update` are ignored. This endpoint is only available on a node that is running an indexer service.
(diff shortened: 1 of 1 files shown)


docs/reference/rest-api.md
```diff
@@ -159,7 +159,8 @@ POST api/v1/_bulk -d \
...
 ```
 
-Ingest a batch of documents to make them searchable using the [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) bulk API. This endpoint provides compatibility with tools or systems that already send data to Elasticsearch for indexing. Currently, only the `create` action of the bulk API is supported, all other actions such as `delete` or `update` are ignored. This endpoint is only available on a node that is running an indexer service.
+Ingest a batch of documents to make them searchable using the [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) bulk API. This endpoint provides compatibility with tools or systems that already send data to Elasticsearch for indexing. Currently, only the `create` action of the bulk API is supported, all other actions such as `delete` or `update` are ignored. 
+We also don't support the Elasticsearch `create` action semantics as of now, `create` only means the documents are safely stored regardless of `_id ` and will be indexed. This endpoint is only available on a node that is running an indexer service.
```
fulmicoton, line 163:
```suggestion
:::caution
In Elasticsearch, the `create` action has a specific behavior when the ingest documents contain an identifier (the `_id` field). It only inserts such a document if it was not inserted before. This is extremely handy to achieve At-Most-Once indexing.

Quickwit does not have any notion of document id and does not support this feature. 
```

fulmicoton (approved):

fulmicoton:
I tried to rephrase a bit and added a :::caution

Issue #1526, opened 2022-05-24
Query on no fields does not return an error for dynamic mapping


PSeitz:
Query on dynamic mapping without default search fields and the query itself also defining no search fields returns an empty result instead of an an error

```
mode: dynamic

➜  aggregation_doc git:(aggregation_doc) curl -sS 'http://localhost:7280/api/v1/air_quality_no_ff_s/search?query=a'
{
  "num_hits": 0,
  "hits": [],
  "elapsed_time_micros": 1906,
  "errors": []
}⏎

mode: lenient
➜  aggregation_doc git:(aggregation_doc) curl -sS 'http://localhost:7280/api/v1/air_quality_no_ff/search?query=a'
{
  "InvalidQuery": "No default field declared and no field specified in query"
}⏎
```

evanxg852000:
@fulmicoton 
I don't think it's good to use the `_dynamic` field in search query 
https://github.com/quickwit-oss/quickwit/blob/main/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs#L483-L485

at least from our first design, even base on the comment here
https://github.com/quickwit-oss/quickwit/blob/main/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs#L95-L97
Did we change our approach?

fulmicoton:
I'm not sure what you mean exactly?

fulmicoton:
So if you are in the `_dynamic` mode, the fields that are not mapped end up in the `_dynamic`  field.

We add it as a default field, so that you can search them naturally.

Did you observe an incorrect behavior?

PR #1550, opened 2022-05-25
Fix default search field on dynamic mapping

evanxg852000:
Checks if default search is not specified and query has no search field declared for dynamic mapping.

Closes https://github.com/quickwit-oss/quickwit/issues/1526

### DIFF
--- a/quickwit-doc-mapper/src/query_builder.rs
+++ b/quickwit-doc-mapper/src/query_builder.rs
@@ -20,10 +20,10 @@
 use quickwit_proto::SearchRequest;
 use tantivy::query::{Query, QueryParser, QueryParserError as TantivyQueryParserError};
 use tantivy::schema::{Field, Schema};
-use tantivy_query_grammar::{UserInputAst, UserInputLeaf};
+use tantivy_query_grammar::{UserInputAst, UserInputLeaf, UserInputLiteral};
 
 use crate::sort_by::validate_sort_by_field_name;
-use crate::{QueryParserError, QUICKWIT_TOKENIZER_MANAGER};
+use crate::{QueryParserError, DYNAMIC_FIELD_NAME, QUICKWIT_TOKENIZER_MANAGER};
 
 /// Build a `Query` with field resolution & forbidding range clauses.
 pub(crate) fn build_query(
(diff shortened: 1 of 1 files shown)


quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -65,8 +74,27 @@ fn has_range_clause(user_input_ast: UserInputAst) -> bool {
...
             false
         }
-        UserInputAst::Boost(ast, _) => has_range_clause(*ast),
-        UserInputAst::Leaf(leaf) => matches!(*leaf, UserInputLeaf::Range { .. }),
+        UserInputAst::Boost(ast, _) => has_range_clause(ast),
+        UserInputAst::Leaf(leaf) => matches!(**leaf, UserInputLeaf::Range { .. }),
+    }
+}
+
+fn has_search_field(user_input_ast: &UserInputAst) -> bool {
```
fulmicoton, line 82: We should reject queries with the same error message if a query contains a single literal with no field.

In this PR, the query `hello` is rejected.
But the query `hello AND my_field:toto` is not rejected.

We want to reject both with this error message.

To fix it, you just need to change the code for `Clause` below, and make the logic stronger.  

```
 for (_, sub_ast) in sub_queries {
                if !has_search_field(sub_ast) {
                    return false;
                }
            }
            true
```

fulmicoton: And probably rename the function `has_search_field_everywhere` or a better name you can think of

fulmicoton: a unit test for this case would be good to :)

fulmicoton (changes requested):
added comment/

quickwit-doc-mapper/src/query_builder.rs
evanxg852000: No sure what you meant by "make the logic stronger" but I refactored and renamed from `has_search_field` to `needs_default_search_field`

fulmicoton: Here by "stronger", I meant it as make "the predicate stricter".

quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -65,8 +74,29 @@ fn has_range_clause(user_input_ast: UserInputAst) -> bool {
...
-        UserInputAst::Boost(ast, _) => has_range_clause(*ast),
-        UserInputAst::Leaf(leaf) => matches!(*leaf, UserInputLeaf::Range { .. }),
+        UserInputAst::Boost(ast, _) => has_range_clause(ast),
+        UserInputAst::Leaf(leaf) => matches!(**leaf, UserInputLeaf::Range { .. }),
+    }
+}
+
+/// Tells if the query has a Term or Range node which does not
+/// specify a search field.
+fn needs_default_search_field(user_input_ast: &UserInputAst) -> bool {
```
fulmicoton, line 84: good name

fulmicoton (approved):

PR #1554, opened 2022-05-26
rename num_docs_ingested

PSeitz:
num_docs_ingested is misleading, since the docs may be just staged for processing.


evanxg852000 (approved):

guilload:
`num_docs_in_flight`?

PSeitz:
> `num_docs_in_flight`?

I'd go with `num_docs_for_processing` here, since the nuance with `num_docs_in_flight` may be slightly less accurate. The docs may already have been processed when returning, for small batches.

Issue #1551, opened 2022-05-25
Rest API Ingest triggers infinite loop if there is a document error.


kstaken:
The Rest API ingest endpoint seems to trigger an infinite loop with error logging if any document in the request has any kind of issue. This is on main.

This request is missing fields in the document.

```
curl 'localhost:7280/api/v1/noaa-1m/ingest' -d '{"station_id":"01088099999","name":"VADSO, NO","temperature_c":-7.2}'
{
  "num_ingested_docs": 1
}
```

The server logs this error endlessly and will continue logging it even if you restart the process. Clearing the queues is required to make it stop.

```
2022-05-25T20:10:22.236Z  WARN {actor=quickwit_indexing::actors::indexing_service::IndexingService}:{msg_id=1}::{index=noaa-1m gen=0}:{actor=Indexer}:{msg_id=127451}: quickwit_indexing::actors::indexer: err=RequiredFastField("wind_direction_deg")
2022-05-25T20:10:22.236Z  WARN {actor=quickwit_indexing::actors::indexing_service::IndexingService}:{msg_id=1}::{index=noaa-1m gen=0}:{actor=Indexer}:{msg_id=127452}: quickwit_indexing::actors::indexer: err=RequiredFastField("wind_direction_deg")
2022-05-25T20:10:22.237Z  WARN {actor=quickwit_indexing::actors::indexing_service::IndexingService}:{msg_id=1}::{index=noaa-1m gen=0}:{actor=Indexer}:{msg_id=127453}: quickwit_indexing::actors::indexer: err=RequiredFastField("wind_direction_deg")
2022-05-25T20:10:22.237Z  WARN {actor=quickwit_indexing::actors::indexing_service::IndexingService}:{msg_id=1}::{index=noaa-1m gen=0}:{actor=Indexer}:{msg_id=127454}: quickwit_indexing::actors::indexer: err=RequiredFastField("wind_direction_deg")
2022-05-25T20:10:22.237Z  WARN {actor=quickwit_indexing::actors::indexing_service::IndexingService}:{msg_id=1}::{index=noaa-1m gen=0}:{actor=Indexer}:{msg_id=127455}: quickwit_indexing::actors::indexer: err=RequiredFastField("wind_direction_deg")
[...]

PR #1556, opened 2022-05-26
Fixed infinite loop on ingest source error

evanxg852000:
Changed `current_offset` type from u64 to Option<u64> to signify none value at the start. 
Closes https://github.com/quickwit-oss/quickwit/issues/1551


fulmicoton (approved):
LGTM but to be honest, I lack context to understand what was the bug originally.

evanxg852000:
I commented on the issue https://github.com/quickwit-oss/quickwit/issues/1551#issuecomment-1140791049

Issue #1555, opened 2022-05-26
Indexer creates empty split


evanxg852000:
When all documents sent to the indexer are invalid, an `indexed_split`  is still sent to the packager despite that split being physically absent from the file system.

**Expected behavior**
The `indexed_split` should be ignored

Indexing pipeline logs
``` 

INFO {actor=quickwit_indexing::actors::indexing_service::IndexingService}:{msg_id=1}::{index=wikipedia gen=0}:{actor=Publisher}:{msg_id=301}: quickwit_metastore::checkpoint: delta=∆(wikipedia:(00000000000000000000..00000000000000000000]) checkpoint=Ckpt()
2022-05-26T11:29:17.347Z ERROR {actor=quickwit_indexing::actors::indexing_service::IndexingService}:{msg_id=1}::{index=wikipedia gen=0}:{actor=Publisher}: quickwit_actors::runner: actor-failure cause=Failed to publish splits.

Caused by:
    Splits `["01G401ZN6J475TR6M5SP8YDXFN"]` do not exist. exit_status=Failure(Failed to publish splits.

Caused by:
    Splits `["01G401ZN6J475TR6M5SP8YDXFN"]` do not exist.)
2022-05-26T11:29:17.348Z  INFO {actor=quickwit_indexing::actors::indexing_service::IndexingService}:{msg_id=1}::{index=wikipedia gen=0}:{actor=MergePlanner}: quickwit_actors::runner: actor-exit actor_id=MergePlanner-hidden-JNe3 exit_status=Success
2022-05-26T11:29:17.348Z  INFO {actor=quickwit_indexing::actors::indexing_service::IndexingService}:{msg_id=1}::{index=wikipedia gen=0}:{actor=Publisher}: quickwit_actors::runner: actor-exit actor_id=Publisher-lingering-6LdT exit_status=Failure(cause=Failed to publish splits.

Caused by:
    Splits `["01G401ZN6J475TR6M5SP8YDXFN"]` do not exist.)
2022-05-26T11:29:17.348Z ERROR {actor=quickwit_indexing::actors::indexing_service::IndexingService}:{msg_id=1}::{index=wikipedia gen=0}:{actor=Publisher}: quickwit_actors::actor: actor-failure actor_name="Publisher-lingering-6LdT" actor_exit_status=Failure(Failed to publish splits.

Caused by:
    Splits `["01G401ZN6J475TR6M5SP8YDXFN"]` do not exist.)
[...]

PR #1557, opened 2022-05-26
Avoid sending empty split

evanxg852000:
Add check to see if indexed_split does not contain a document.
Closes https://github.com/quickwit-oss/quickwit/issues/1555

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2586,6 +2586,8 @@ dependencies = [
  "futures",
  "rand",
  "rusoto_core",
+ "rusoto_kinesis",
+ "rusoto_s3",
  "tokio",
  "tracing",
 ]
--- a/quickwit-aws/Cargo.toml
+++ b/quickwit-aws/Cargo.toml
@@ -14,5 +14,10 @@ documentation = "https://quickwit.io/docs/"
 futures = "0.3"
 rand = "0.8"
 rusoto_core = {version = "0.47", default-features = false, features = ["rustls"]}
+rusoto_kinesis = { version = "0.47", default-features = false, features = ["rustls"], optional = true }
+rusoto_s3 = { version = "0.47", default-features = false, features = ["rustls"] }
 tokio = "1.18"
 tracing = "0.1"
+
+[features]
+kinesis = ["rusoto_kinesis"]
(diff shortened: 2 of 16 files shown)


guilload:
Hold on! What if `num_docs == 0` but `num_parsing_errors > 0`?

Imagine the indexer receives malformed docs for several minutes. Empty splits are not great, but at least the checkpoint is updated, and some progress is made in the consumption of the source. If the pipeline crashes down the line and has not committed, on restart, it will start processing all those bad records again.

However, I don't know if we have a metastore API just to update the checkpoint.

It seems like we need to put more thought into this.

evanxg852000:
@guilload I thought of sending a checkpoint before returning Ok. But my reasoning is that if the entire dataset is malformed then we probably should not be advancing it, the user should correct it (he probably wants to process it after cleaning). With a good dataset, we will keep dropping the wrong docs until we get into a range of good docs at one point, that will generate a commit delta. 
I might be missing some details, will keep thinking further.

guilload:
I don't get why we should wait for a valid doc to register some form of progress? Rejecting malformed docs is also progress and should be taken into account. As long as we keep track of the number of errors and error rate and this information is correctly surfaced to users, they can take the appropriate measures:
- for files, clean them and re-ingest
- for streams (Ingest API,  Kafka, Kinesis) fix whatever is sending malformed data and re-enqueue

For immutable streams, if we don't skip the bad data, users will have to edit checkpoints manually or apply retention surgically.

quickwit-indexing/src/actors/indexer.rs
```diff
@@ -397,12 +404,26 @@ impl Indexer {
...
         ctx: &ActorContext<Self>,
-    ) -> Result<(), SendError> {
+    ) -> anyhow::Result<()> {
         let indexed_split = if let Some(indexed_split) = self.current_split_opt.take() {
             indexed_split
         } else {
             return Ok(());
         };
+
+        // Avoid sending empty split.
```
fulmicoton, line 414:
```suggestion
        // Avoid producing empty split, but still update the checkpoint to avoid
        // reprocessing the same faulty documents.
```

fulmicoton:
@evanxg852000 I modified your PR a little bit to avoid introducing a new API call. Instead I replaced apply_delta by calling publish_splits with an empty array in place of the split ids. Can you review?

fulmicoton (approved):

evanxg852000 (commented):
Nice, that's very clever to reduce the amount of code change.

PR #1566, opened 2022-05-30
return correct error status to http client

### LINKED ISSUE #1564: Payload too large on REST API ingest should provide a better error message
When you submit an ingest request that is too large you get an internal server error and the server logs a more specific error. The end user should also receive an error that tells them what is wrong with the request. Ideally the message would also include the value of the limit.

```
$ curl -XPOST 'localhost:7280/api/v1/data1/ingest' --data-binary @grabber/data/2022-05-26T06:12:41+00:00-2022-05-26T07:12:41+00:00.ldjson
{
  "error": "Internal server error."
}
```

The server logs

```
2022-05-28T00:09:59.824Z ERROR quickwit_serve::rest: REST server error: Rejection(PayloadTooLarge)
```

PSeitz:
fixes #1564

### How was this PR tested?
```
➜  quickwit-indices curl -X POST http://127.0.0.1:7280/api/v1/hdfs/ingest --data-binary @./wiki-articles_number_small.json
{
  "error": "Internal server error."
}⏎
➜  quickwit-indices curl -X POST http://127.0.0.1:7280/api/v1/hdfs/ingest --data-binary @./wiki-articles_number_small.json
{
  "error": "The request payload is too large"
}⏎
```

### DIFF
--- a/quickwit-serve/src/error.rs
+++ b/quickwit-serve/src/error.rs
@@ -37,6 +37,8 @@ use warp::http;
 pub enum ServiceErrorCode {
     NotFound,
     Internal,
+    MethodNotAllowed,
+    UnsupportedMediaType,
     BadRequest,
     PermissionDenied,
 }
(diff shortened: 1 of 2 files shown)


fulmicoton:
@PSeitz Can you rebase so we can ship that in 0.3?

PR #1589, opened 2022-06-01
Display cluster ID in logs and UI

### LINKED ISSUE #1315: Display the cluster_id in the quickwit UI and in logs.

guilload:
### Description
Display cluster ID in logs and UI. Fixes #1315.

#### Logs
Before:
```
2022-06-01T21:35:04.479Z  INFO quickwit_cluster::cluster: Create new cluster. member=Member { node_unique_id: "node-green-PJji", generation: 1654119304, gossip_public_address: 127.0.0.1:7280, is_self: true } listen_addr=127.0.0.1:7280
2022-06-01T21:35:04.480Z
```

After:
```
2022-06-01T22:03:25.197Z  INFO quickwit_cluster::cluster: Joining cluster. cluster_id=quickwit-default-cluster node_id=node-nameless-HM3k listen_address=127.0.0.1:7280 gossip_address=127.0.0.1:7280
```

#### UI
Screenshot_20220601_173646

### DIFF
--- a/quickwit-proto/proto/cluster.proto
+++ b/quickwit-proto/proto/cluster.proto
@@ -53,6 +53,9 @@ message ListMembersRequest {
 }
 
 message ListMembersResponse {
+  /// Cluster ID
+  string cluster_id = 2;
+  /// All nodes belonging to the cluster.
   repeated Member members = 1;
 }
 
--- a/quickwit-cluster/src/cluster.rs
+++ b/quickwit-cluster/src/cluster.rs
@@ -104,6 +104,9 @@ impl From<Member> for NodeId {
 
 /// This is an implementation of a cluster using Chitchat.
 pub struct Cluster {
+    /// ID of the cluster joined.
+    pub cluster_id: String,
+    /// Node ID.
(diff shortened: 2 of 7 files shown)


quickwit-proto/proto/cluster.proto
```diff
@@ -53,6 +53,9 @@ message ListMembersRequest {
 }
 
 message ListMembersResponse {
```
fmassot, line 55: Should we keep `ListMembersResponse` or go for `ClusterState` as mentioned in a JS file?
I'm ok to not change the name now, there is another issue that suggest dumping all keys/values of each members so that we have the full view of the cluster from a given node.

fmassot (approved):

quickwit-proto/proto/cluster.proto
guilload: Yeah, the naming isn't great. I'd like to see how the use of this endpoint evolves before we start renaming things.

fmassot: yes exactly, so we are aligned :)

PR #1612, opened 2022-06-07
force http1

### LINKED ISSUE #1584: Support Google cloud storage.
We already support specifying an a non-AWS endpoint. In theory everything should work just fine, but let's check that by indexing a few splits and deleting an index.

fulmicoton:
Disables http2 to fix bug with gcloud
    
Google cloud storage allows http2, but hyper's http2 client
does not seem to support it properly, leading to an error.
    
This PR manually disables http2.
    
Closes #1584

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -756,15 +756,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ct-logs"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c1a816186fa68d9e426e3cb4ae4dff1fcd8e4a2c34b781bf7a822574a0d0aac8"
-dependencies = [
- "sct 0.6.1",
-]
-
-[[package]]
 name = "ctor"
 version = "0.1.22"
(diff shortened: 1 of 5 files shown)


saroh (approved):
LGTM! 
I can re raise the issue w/ rusoto and probably hyper-rustls (I haven't validated the presence of the host header tbh)  WDYT ?
 give hyper-tls is ok w/ http2 
https://github.com/rusoto/rusoto/blob/master/rusoto/core/src/lib.rs#L22-L25
https://github.com/rusoto/rusoto/blob/master/rusoto/core/src/request.rs#L220-L260 
https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.3
https://github.com/hyperium/h2/issues/548#issuecomment-969887965

fmassot:
@saroh that would really nice to open an issue but it would be better to check that it comes from some header.

fulmicoton:
@saroh yes please! that would be awesome to get that fixed in rusoto. (that's more a rusoto problem than an hyper problem I think)

fulmicoton:
Merging as a termporary fix, but it would be really awesome to get http2 to work eventually.

PR #1616, opened 2022-06-07
Create dependency.yml

PSeitz:
check if dependencies contain vulnerabilities

### DIFF
--- a/.github/workflows/dependency.yml
+++ b/.github/workflows/dependency.yml
@@ -1,0 +1,14 @@
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency.yml
```diff
@@ -0,0 +1,14 @@
...
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v3
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-
```
fulmicoton, line 14: is the trailing dash correct?

PSeitz: copy paste error, just fixed it

fulmicoton (approved):

PR #1635, opened 2022-06-14
Introducing a sequencer actor.

fulmicoton:
File uploads are done concurrently.
They may end in a different order than they started.

Before this PR, maintaining the ordering was the job of the publisher.
After this PR, this sequencing logic is done by a dedicated actor
called Sequencer.

The purpose of this PR is to pave the way to make the Publisher a
singleton Service/Actor (as opposed to one actor per index).

We do not want to observe orders for all indices.
One slow upload on one index does not need to delay the publication of
split for all indices. It was therefore necessary to keep the sequencing
within the per-index indexing pipeline.

Besides, it separate the concern, and makes the unit test & logic much
cleaner.

### Description

### How was this PR tested?

### DIFF
--- a/quickwit-indexing/src/actors/sequencer.rs
+++ b/quickwit-indexing/src/actors/sequencer.rs
@@ -1,0 +1,120 @@
+// Copyright (C) 2021 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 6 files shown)


quickwit-indexing/src/actors/sequencer.rs
```diff
@@ -0,0 +1,120 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use anyhow::Context;
+use async_trait::async_trait;
+use quickwit_actors::{Actor, ActorContext, ActorExitStatus, Handler, Mailbox};
+use tokio::sync::oneshot;
+
+pub struct Sequencer<A: Actor> {
```
guilload, line 25: Maybe briefly describe the goal of this actor in a comment?

quickwit-indexing/src/actors/uploader.rs
```diff
@@ -116,12 +117,12 @@ impl Handler<PackagedSplitBatch> for Uploader {
         ctx: &ActorContext<Self>,
     ) -> Result<(), ActorExitStatus> {
         fail_point!("uploader:before");
-        let (split_uploaded_tx, split_uploaded_rx) = tokio::sync::oneshot::channel();
+        let (split_uploaded_tx, split_uploaded_rx) = oneshot::channel::<PublisherMessage>();
 
         // We send the future to the publisher right away.
```
guilload, line 122: This comment and a few other comments are now outdated.

guilload (approved):

quickwit-indexing/src/actors/sequencer.rs
fulmicoton: yes

PR #1639, opened 2022-06-15
Added to semaphore to help with the too many file description erro

fulmicoton:


guilload:
I pushed and merged something similar last week (#2017). We no longer create multiple scratch directories, so this is no longer an issue, but I still limited how many indexing pipelines we spawn concurrently. Seemed to help my issues when spawning 50+ Kafka consumers.

Closing.

PR #1645, opened 2022-06-15
Fix index create overwrite

fmassot:
…te event if no index has been created.

### How it was tested
Add one test.


PSeitz (approved):

quickwit-cli/src/index.rs
```diff
@@ -784,8 +788,19 @@ pub async fn create_index_cli(args: CreateIndexArgs) -> anyhow::Result<()> {
         quickwit_storage_uri_resolver().clone(),
         quickwit_config.default_index_root_uri(),
     );
+
     if args.overwrite {
-        index_service.delete_index(&index_id, false).await?;
+        match index_service.delete_index(&index_id, false).await {
```
guilload, line 793: I wish this logic was folded in the `IndexService::delete_index` method.

guilload (approved):

quickwit-cli/src/index.rs
fmassot: yes, me too :D

PR #1647, opened 2022-06-17
add LocalFileStorage delay to simulate s3

PSeitz:


quickwit-cli/Cargo.toml
```diff
@@ -72,5 +72,6 @@ serial_test = "0.7.0"
 ci-test = []
 tokio-console = ["console-subscriber"]
 openssl-support = ["openssl-probe"]
+simulates3 = ["quickwit-metastore/simulates3"]
```
fmassot, line 75: `simulate-s3` seems more readable

fulmicoton: that will show up with `--all-features`.

fulmicoton:
Good idea, but making it a feature is not a good idea.
- `--all-features` will enable it and no one expects that. 
- also it requires recompiling things way too much.

In many places, rust treats feature flags as additive semantically.
It is not supposed to change the behavior (we make that mistake in tantivy I think too).

For instnace, if deps A and deps B depends on C with feature flag [feaflag1] and [featflag2] and compatible versions,
rustc will compile a single version of C with [featflag1, featflag2].

fulmicoton (changes requested):
Good idea, but making it a feature is not a good idea.
- `--all-features` will enable it and no one expects that. 
- also it requires recompiling things way too much.

In many places, rust treats feature flags as additive semantically.
It is not supposed to change the behavior (we make that mistake in tantivy I think too).

For instnace, if deps A and deps B depends on C with feature flag [feaflag1] and [featflag2] and compatible versions,
rustc will compile a single version of C with [featflag1, featflag2].

As an alternative could we just add 
- a wrapper over a storage to add latency / throughput. There used to be one maybe it is still there actually.
- a storage called simulate_s3 or something like that, that wraps the file_storage with the wrapper descrbied above.

PR #1650, opened 2022-06-17
Add indexer and searcher statefulsets along with discovery service

guilload:
### Description
Add Kubernetes example manifest files

### How was this PR tested?
Tested on Minikube cluster

### DIFF
--- a/distribution/kubernetes/discovery-service.yaml
+++ b/distribution/kubernetes/discovery-service.yaml
@@ -1,0 +1,16 @@
+# Headless service discovery
+apiVersion: v1
+kind: Service
+metadata:
+  name: discovery
+  labels:
+    app: discovery
+spec:
+  clusterIP: None
+  publishNotReadyAddresses: true
+  ports:
+    - port: 7282
+      protocol: UDP
+      name: discovery
+  selector:
+    app: quickwit
--- a/distribution/kubernetes/searcher-statefulset.yaml
+++ b/distribution/kubernetes/searcher-statefulset.yaml
@@ -83,14 +83,14 @@ spec:
               value: "discovery.$(NAMESPACE).svc.cluster.local"
             - name: QW_ADVERTISE_ADDRESS
               value: "$(POD_IP)"
-          readinessProbe:
-            httpGet:
-              path: /healthz/searcher/readyz
-              port: rest
-          livenessProbe:
-            httpGet:
-              path: /healthz/searcher/livez
-              port: rest
+          # readinessProbe:
+          #   httpGet:
+          #     path: /healthz/searcher/readyz
+          #     port: rest
+          # livenessProbe:
+          #   httpGet:
+          #     path: /healthz/searcher/livez
+          #     port: rest
       volumes:
         - name: config-volume
           configMap:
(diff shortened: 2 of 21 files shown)


distribution/kubernetes/searcher-statefulset.yaml
```diff
@@ -0,0 +1,94 @@
+apiVersion: apps/v1
+kind: StatefulSet
```
fmassot, line 2: Do we really need a statefulset for searchers?

guilload: Yes, I believe our logic for node/split affinity via rendez-vous hashing relies on stable node IDs.

fmassot: Legit.

mkhpalm: I'm not sure I understand the sts here either. Does the node/split affinity depend on the presence of previously existing cache on disk? Meaning, if you delete/restart a sts pod the name will stay the same but the emptyDir for cache will be gone. So it would need to repopulate any split cache it had previously regardless?

distribution/kubernetes/discovery-service.yaml
```diff
@@ -0,0 +1,15 @@
+# Headless service discovery
+apiVersion: v1
+kind: Service
+metadata:
+  name: discovery
+  labels:
+    app: discovery
+spec:
+  clusterIP: None
```
mkhpalm, line 9: Would it make sense to add `publishNotReadyAddresses: true` and then add readynessProbes to the searcher-statefulset? This way discovery udp:7282 will resolve all endpoints all the time but you can limit searcher-service to only the functioning searcher nodes. To avoid bad requests.

distribution/kubernetes/searcher-statefulset.yaml
guilload: You're right; in this contrived example, stable node IDs between restarts are useless because the `emptyDir` is trashed on restart.

These manifests are intended to serve as a basis for customization. I hope to commit a helm chart to make this easier in the next few weeks.

distribution/kubernetes/discovery-service.yaml
guilload: Yes, I wasn't aware of `publishNotReadyAddresses`, but that makes total sense. I'm going to add liveness and readiness problems in a follow-up PR. I need to implement them in our services first :)

Also, would you mind looking at the checklist I wrote in #1658? I may have missed a few things.

guilload:
Thanks for your comments @mkhpalm!

distribution/kubernetes/searcher-statefulset.yaml
mkhpalm: We have already created a helm chart using deployments if you're interested. If I knew where you wanted it I can submit it to you via a PR or just put it somewhere for you guys to get. Its not perfect or completely flushed out but might help.

guilload: That would be a great start for us. I've just created a [repo](https://github.com/quickwit-oss/helm-charts) for our helm charts and invited you to the Quickwit dev team on GitHub.

How do you handle stable node IDs with deployments?

distribution/kubernetes/discovery-service.yaml
mkhpalm: Checklist looks pretty good to me, I did a pr for an initial helm chart we've been using. 

https://github.com/quickwit-oss/helm-charts/pull/1

Its kind of just a start, or what we've been using to test quickwit with. I'm not 100% on the direction of everything and made a few assumptions about the future. Hopefully thats of some help with that checklist

fulmicoton (approved):

PR #1664, opened 2022-06-24
small fix setting filename in "quickstart.md "

higumachan:
### Description
When performing "Create your first index" on quickstart, I noticed that the config file name specified in the CLI command was different from the config file name displayed.


guilload (approved):

fmassot:
thanks @higumachan!

PR #1680, opened 2022-06-28
Limiting the number of concurrent split search.

### LINKED ISSUE #1679: OOMs after repeated queries on larger amounts of data.
**Describe the bug**

We've now loaded Quickwit with 16.4 Billion records and have started to trigger some out of memory (OOM) failures. In addition we've also seen clustering issues so to isolate the OOMs we scaled down to a single search node. 

The test query matches 67 million records but there's no sorting, timestamps or anything complicated on the query, just a single criteria i.e.:`field:value` and max_hits=1. 

On a single searcher node this query will run successfully in about 38 seconds on the first run, 25 seconds on the second run and then consistently OOM on the third. Queries are not concurrent and no other queries are submitted between subsequent runs.

In this case it's the kernel killing Quickwit since it's exceeding the memory limit allocated. The searcher is running in Kubernetes with 32 GB of RAM allocated. 

**Configuration:**

This index currently has 1,340 splits, with 10M doc target per split.

```
# quickwit --version
Quickwit 0.3.0 (commit-hash: 6d07599)
```

Memory and cache settings are the defaults. 
```
searcher:
  fast_field_cache_capacity: 10G
  split_footer_cache_capacity:  1G
  max_num_concurrent_split_streams: 100
```

fulmicoton:
Closes #1679


quickwit-config/resources/tests/config/quickwit.json
```diff
@@ -18,7 +18,8 @@
     "searcher": {
         "fast_field_cache_capacity": "10G",
         "split_footer_cache_capacity": "1G",
-        "max_num_concurrent_split_streams": 120
+        "max_num_concurrent_split_streams": 120,
+        "max_num_concurrent_split_searchs": 150
```
guilload, line 22: `max_num_concurrent_split_searches`?

guilload (approved):

quickwit-config/resources/tests/config/quickwit.json
fulmicoton: ok renamed.

PR #1688, opened 2022-06-29
Make advertise IP configurable and sniff it when necessary

### LINKED ISSUE #1661: Make public IP configurable in node config.
Currently only available via the `QW_PUBLIC_IP` env var. Assignee, don't forget to update the docs.

guilload:
### Description
- Add public IP to node config
- Infer public IP from network interfaces if necessary.

Fixes #1661

This is what [memberlist](https://github.com/hashicorp/memberlist/blob/696ff46201c1b64d31b60d57971d07d8cf15df7a/net_transport.go#L130) does:

1. Use public IP from env var or config if provided
2. Use any private IP from network interfaces if listen address is unspecified (`0.0.0.0`)
3. Use `listen_address`.

### How was this PR tested?
- Added unit tests
- Tested on my machine

### DIFF
--- a/config/quickwit.yaml
+++ b/config/quickwit.yaml
@@ -20,12 +20,25 @@ version: 0
 # - for its gRPC service (TCP)
 # - for its Gossip cluster membership service (UDP)
 #
-# All three services are bound to the same host and different port.
+# All three services are bound to the same host and a different port. The host can be an IP address or a hostname.
 #
 # Default HTTP server host is `127.0.0.1` and default HTTP port is 7280.
+# The default host value was chosen to avoid exposing the node to the open-world without users' explicit consent.
+# This allows for testing Quickwit in single-node mode or with multiple nodes running on the same host and listening
+# on different ports. However, in cluster mode, using this value is never appropriate because it causes the node to
+# ignore incoming traffic.
+# There are two options to set up a node in cluster mode:
+#   1. specify the node's hostname or IP
+#   2. pass `0.0.0.0` and let Quickwit do its best to discover the node's IP (see `advertise_address`)
 #
 # listen_address: 127.0.0.1
 # rest_listen_port: 7280
+#
+# IP address advertised by the node, i.e. the IP address that peer nodes should use to connect to the node for RPCs.
+# The environment variable `QW_ADVERTISE_ADDRESS` can also be used to configure this value.
+# The default advertise address is `listen_address`. If `listen_address` is unspecified (`0.0.0.0`),
+# Quickwit attempts to sniff the node's IP by scanning the available network interfaces.
+# advertise_address: 192.168.0.42
 #
 # In order to join a cluster, one needs to specify a list of
 # seeds to connect to. If no port is specified, Quickwit will assume
@@ -33,7 +46,7 @@ version: 0
 # By default, the peer seed list is empty.
 #
 # peer_seeds:
-#   - quickwit-searcher-0.local:10000
+#   - quickwit-searcher-0.local
 #   - quickwit-searcher-1.local:10000
 #
 # Path to directory where temporary data (caches, intermediate indexing data structures)
--- a/quickwit-cluster/src/lib.rs
+++ b/quickwit-cluster/src/lib.rs
@@ -76,7 +76,7 @@ pub async fn start_cluster_service(
     let member = Member::new(
         quickwit_config.node_id.clone(),
         unix_timestamp(),
-        quickwit_config.gossip_public_addr().await?,
+        quickwit_config.gossip_advertise_addr().await?,
     );
 
     let cluster = Cluster::join(
@@ -84,7 +84,7 @@ pub async fn start_cluster_service(
         services,
         quickwit_config.gossip_listen_addr().await?,
         quickwit_config.cluster_id.clone(),
-        quickwit_config.grpc_public_addr().await?,
+        quickwit_config.grpc_advertise_addr().await?,
         quickwit_config.peer_seed_addrs().await?,
         FailureDetectorConfig::default(),
         &UdpTransport,
--- a/quickwit-common/src/net.rs
+++ b/quickwit-common/src/net.rs
@@ -18,16 +18,50 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::fmt::Display;
-use std::net::{IpAddr, SocketAddr, TcpListener};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, TcpListener};
+use std::str::FromStr;
 
 use anyhow::{bail, Context};
+use itertools::Itertools;
+use once_cell::sync::OnceCell;
+use pnet::datalink::{self, NetworkInterface};
+use pnet::ipnetwork::IpNetwork;
+use serde::{Deserialize, Serialize, Serializer};
 use tokio::net::{lookup_host, ToSocketAddrs};
 
 /// Represents a host, i.e. an IP address (`127.0.0.1`) or a hostname (`localhost`).
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum Host {
     Hostname(String),
     IpAddr(IpAddr),
+}
+
+impl Host {
+    /// Returns [`true`] for the "unspecified" address (all bits set to zero).
+    pub fn is_unspecified(&self) -> bool {
+        match &self {
+            Host::Hostname(_) => false,
+            Host::IpAddr(ip_addr) => ip_addr.is_unspecified(),
+        }
+    }
+
+    /// Appends `port` to `self` and returns a [`HostAddr`].
+    pub fn with_port(&self, port: u16) -> HostAddr {
+        HostAddr {
+            host: self.clone(),
+            port,
+        }
+    }
+
+    /// Resolves the host if necessary and returns an [`IpAddr`].
+    pub async fn resolve(&self) -> anyhow::Result<IpAddr> {
+        match &self {
+            Host::Hostname(hostname) => get_socket_addr(&(hostname.as_str(), 0))
+                .await
+                .map(|socket_addr| socket_addr.ip()),
+            Host::IpAddr(ip_addr) => Ok(*ip_addr),
+        }
+    }
 }
 
 impl Display for Host {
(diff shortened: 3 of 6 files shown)


quickwit-cluster/src/lib.rs
```diff
@@ -78,15 +78,15 @@ pub async fn start_cluster_service(
     let member = Member::new(
         quickwit_config.node_id.clone(),
         unix_timestamp(),
-        quickwit_config.gossip_public_addr().await?,
+        quickwit_config.gossip_advertise_addr().await?,
```
fulmicoton, line 81: I personally preferred public actually

quickwit-config/src/config.rs
```diff
@@ -306,16 +310,39 @@ impl QuickwitConfig {
...
-        get_socket_addr(&(self.listen_address.as_str(), self.grpc_listen_port())).await
+        self.listen_address
+            .with_port(self.grpc_listen_port())
+            .to_socket_addr()
+            .await
+    }
+
+    /// Returns the advertise
+    pub fn advertise_addr(&self) -> anyhow::Result<Host> {
+        if let Ok(advertise_address) = env::var("QW_ADVERTISE_ADDRESS") {
```
fulmicoton, line 321: Is it common to have override in that order
ENV > CLI?

I think I usually expect CLI > ENV

quickwit-common/src/net.rs
```diff
@@ -116,6 +193,10 @@ pub fn find_available_tcp_port() -> anyhow::Result<u16> {
     Ok(port)
 }
 
+pub fn find_private_ip() -> anyhow::Result<IpAddr> {
+    unimplemented!()
```
fulmicoton, line 197: Let's make sure this is implemented before merging (currently PR is marked as WIP so no pb)

guilload: I just added the implementation. Can you take a look while I finish this PR (adding comments and unit tests)?

quickwit-cluster/src/lib.rs
fmassot: Cassandra uses [broadcast_address](https://docs.datastax.com/en/cassandra-oss/3.x/cassandra/configuration/configCassandra_yaml.html#configCassandra_yaml__broadcast_address) 

Docs description is clear.

I'm both ok with advertise / public, I have a slight preference for advertise though.

quickwit-config/src/config.rs
guilload: This case is ENV > Config file.

quickwit-cluster/src/lib.rs
guilload: I'm going to stick with `advertise_address`, which is also used for the Kube API server. My grief with `public_ip` is that the address usually used for inter-node communication is actually private (`192.168.`, `172.16.`).

fulmicoton: ok

guilload:
This is ready for a final review.

Sniffing the private IP should work in Docker compose:
```
guilload@x13 ~/W/r/quickwit (guilload--public-ip)> docker exec -it 996b05d0674a /bin/bash
root@996b05d0674a:/# ip address
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
19: eth0@if20: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:ac:10:07:04 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 172.16.7.4/24 brd 172.16.7.255 scope global eth0
       valid_lft forever preferred_lft forever
root@996b05d0674a:/#
```
and Kube:
```
guilload@x13 ~/W/r/quickwit (guilload--public-ip)> k exec -n quickwit --stdin --tty indexer-0 -- /bin/bash
root@indexer-0:/quickwit# ip address
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
11: eth0@if12: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:ac:11:00:06 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 172.17.0.6/16 brd 172.17.255.255 scope global eth0
       valid_lft forever preferred_lft forever
root@indexer-0:/quickwit#
```

(This is minikube though)

quickwit-config/src/config.rs
```diff
@@ -306,16 +310,49 @@ impl QuickwitConfig {
...
     pub async fn grpc_listen_addr(&self) -> anyhow::Result<SocketAddr> {
-        get_socket_addr(&(self.listen_address.as_str(), self.grpc_listen_port())).await
+        self.listen_address
+            .with_port(self.grpc_listen_port())
+            .to_socket_addr()
+            .await
+    }
+
+    /// Returns the advertise
+    pub fn advertise_addr(&self) -> anyhow::Result<Host> {
```
guilload, line 320: This is actually called twice, which also displays the log line twice. Any neat way to avoid this?

config/quickwit.yaml
```diff
@@ -20,20 +20,23 @@ version: 0
 # - for its gRPC service (TCP)
 # - for its Gossip cluster membership service (UDP)
 #
-# All three services are bound to the same host and different port.
+# All three services are bound to the same host and a different port. The host can be an IP address or a hostname.
 #
 # Default HTTP server host is `127.0.0.1` and default HTTP port is 7280.
```
fulmicoton, line 25: For good reasons, our default is not a sensible value for most use case.

Can you add a comment explaining that `127.0.0.1` cannot work when working on more than one node,
and document the behavior of `0.0.0.0`

config/quickwit.yaml
```diff
@@ -20,20 +20,23 @@ version: 0
...
 #
 # listen_address: 127.0.0.1
 # rest_listen_port: 7280
 #
+# IP address advertised by the node, i.e. the IP address that peer nodes should use to connect to the node for gossip and gRPC.
+# advertise_address: 192.168.0.42
```
fulmicoton, line 31: Can you describe what the default value is here?

quickwit-common/src/net.rs
```diff
@@ -116,6 +200,42 @@ pub fn find_available_tcp_port() -> anyhow::Result<u16> {
     Ok(port)
 }
 
+/// Attempts to find the private IP of the host. Returns the matching interface name along with it.
+pub fn find_private_ip() -> Option<(String, IpAddr)> {
+    _find_private_ip(datalink::interfaces())
+}
+
+fn _find_private_ip(interfaces: Vec<NetworkInterface>) -> Option<(String, IpAddr)> {
```
fulmicoton, line 208: nitpick: does not matter here, but really does in some other place to the point where it should be a clippy lint imho.

```suggestion
fn _find_private_ip(interfaces: &[NetworkInterface]) -> Option<(String, IpAddr)> {
```

fulmicoton: I did not fix it on the caller sites.

config/quickwit.yaml
guilload: Will do!

config/quickwit.yaml
guilload: Yes, I can elaborate.

quickwit-common/src/net.rs
guilload: Yeah, makes sense, I was lazy.

config/quickwit.yaml
fulmicoton: and that it can be overriden by `QW_ADVERTISE_ADDRESS`

quickwit-config/src/config.rs
```diff
@@ -306,16 +310,49 @@ impl QuickwitConfig {
...
+        if let Ok(advertise_address) = env::var("QW_ADVERTISE_ADDRESS") {
```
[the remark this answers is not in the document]
guilload: Would we be ok with a static variable here?

quickwit-config/src/config.rs
```diff
@@ -306,16 +310,49 @@ impl QuickwitConfig {
...
     }
 
     /// Returns the gRPC public address of the node, i.e. the socket address to connect to in order
     /// to send gRPC requests to the node.
-    pub async fn grpc_public_addr(&self) -> anyhow::Result<SocketAddr> {
-        match env::var("QW_PUBLIC_IP") {
-            Ok(ip_addr) => get_socket_addr(&(ip_addr, self.grpc_listen_port())).await,
-            Err(_) => self.grpc_listen_addr().await,
-        }
+    pub async fn grpc_advertise_addr(&self) -> anyhow::Result<SocketAddr> {
```
fulmicoton, line 351: This PR makes it apparent that we have a problem with our approach to configuration...

Our current logic takes values from
- env variables (in different ways: overrides, via env variables, and templating also in the future)
- conf
- defaults values from serde
in that order...

but also some more complicated accessor logic.

The issue with this approach, is that it will make it difficult for people to debug.

For instance, we just added somethign in the UI to serialize the conf here. However the advertise_addr  after serialization might be wrong. (overridden by an env variable for instance)

I suggest we rename this QuickwitConfig object QuickwitConfigbuilder or something 
and add another `QuickwitConfig` that is a simple struct with no impl, and contains the results of all of the acccessors of QuickwitConfigBuilder.

That would be the thing we serve in the API.

Ideally it could also embed some debug info about all of the overrides that happened.
If we don't have any magical idea on how to do that it coudl just be a disgusting 
`Vec<String>` in which `QuickwitConfigBuilder::build` could append all worth mentioning mutations as if it was a log.

fulmicoton (approved):
Approved but please look at comments inline.

quickwit-config/src/config.rs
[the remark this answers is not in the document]
fulmicoton: yes you can have a static once_cell in the method body itself

Issue #1667, opened 2022-06-24
Split sizes in the metastore appear to be incorrect


kstaken:
**Describe the bug**
The split sizes stored in the metastore appear to be incorrect and result in over reporting on the size of the data in the UI and other places where the size is surfaced. 

Split as reported by Quickwit

```
  {
    "split_state": "Published",
    "update_timestamp": 1656028266,
    "version": "1",
    "split_id": "01G69FGNK6G5STF596PAV60RCR",
    "num_docs": 11616042,
    "size_in_bytes": 9520864939,
    "time_range": {
      "start": 1655915049,
      "end": 1655939291
    },
```

Object size in Object storage.

```
2022-06-23 23:50 3276823007   s3://quickwit/indexes/rc_test_v1/01G69FGNK6G5STF596PAV60RCR.split
```

fulmicoton:
This size is the original size of your documents payload, not the size of the splits.
We did a proper rename in the code but not in the serialization.

It would also be nicer to have both.

PR #1692, opened 2022-06-30
Rename SplitMetadata.size_in_bytes

evanxg852000:
Renamed `SplitMetadata.size_in_bytes` to `original_size_in_bytes` to avoid confusion

Closes https://github.com/quickwit-oss/quickwit/issues/1667


guilload:
I find that `original_size_in_bytes` is merely an improvement. What about something like `uncompressed_payload_size_in_bytes`?

evanxg852000:
Yes @guilload  I still found it confusing but I wanted to conform to a previous renaming thinking this was what we wanted after all. I had these names in mind  too `ingested_docs_size_in_bytes` `original_docs_size_in_bytes`. so let's decide here.

PSeitz:
I'm in favor of `original_docs_size_in_bytes` or `uncompressed_docs_size_in_bytes`

evanxg852000:
@guilload  any take on this? I am leaning towards `original_docs_size_in_bytes`

guilload:
Good suggestion, @PSeitz. I'm leaning towards `uncompressed_docs_size_in_bytes` or `raw_docs_size_in_bytes`. Original makes perfect sense for us, Quickwit devs, but I'm afraid that might not be explicit enough for newcomers.

evanxg852000:
@guilload @PSeitz  I finally settled with `raw_docs_size_in_bytes`. It's shorter IMO

PR #1696, opened 2022-06-30
Improved CLI docs and add prompt to confirm index create overwrite.

fmassot:
### Description

Improve CLI docs and add prompt to confirm overwrite. By default, the answer is no.
A flag "-y --assume-yes" is here  

### Note

I like having a better XP on the CLI. Unfortunately, the current code I introduce is lame as it's very specific to one command.

I think other risky operations should behave the same way. 

### prompt

Capture d’écran 2022-06-30 à 23 07 18

### yes
Capture d’écran 2022-06-30 à 23 08 19

### no

Capture d’écran 2022-06-30 à 23 08 01

### How was this PR tested?

With the current tests.

### DIFF
--- a/quickwit-cli/src/index.rs
+++ b/quickwit-cli/src/index.rs
@@ -335,7 +335,7 @@ impl IndexCliCommand {
             .expect("`index-config` is a required arg.")?;
         let data_dir = matches.value_of("data-dir").map(PathBuf::from);
         let overwrite = matches.is_present("overwrite");
-        let assume_yes = matches.is_present("assume-yes");
+        let assume_yes = matches.is_present("yes");
 
         Ok(Self::Create(CreateIndexArgs {
             config_uri,
--- a/quickwit-cli/src/index.rs
+++ b/quickwit-cli/src/index.rs
@@ -76,7 +76,9 @@ pub fn build_index_command<'a>() -> Command<'a> {
                     arg!(--"data-dir" <DATA_DIR> "Where data is persisted. Override data-dir defined in config file, default is `./qwdata`.")
                         .env("QW_DATA_DIR")
                         .required(false),
-                    arg!(--overwrite "Overwrites pre-existing index.")
+                    arg!(--overwrite "Overwrites pre-existing index. This will delete all existing data stored at `index-uri` before creating a new index.")
+                        .required(false),
+                    arg!(-y --"yes" "Assume \"yes\" as an answer to all prompts and run non-interactively.")
                         .required(false),
                 ])
             )
@@ -200,6 +202,7 @@ pub struct CreateIndexArgs {
     pub index_config_uri: Uri,
     pub data_dir: Option<PathBuf>,
     pub overwrite: bool,
+    pub assume_yes: bool,
 }
 
 #[derive(Debug, Eq, PartialEq)]
@@ -332,12 +335,14 @@ impl IndexCliCommand {
             .expect("`index-config` is a required arg.")?;
         let data_dir = matches.value_of("data-dir").map(PathBuf::from);
         let overwrite = matches.is_present("overwrite");
+        let assume_yes = matches.is_present("assume-yes");
 
         Ok(Self::Create(CreateIndexArgs {
             config_uri,
             data_dir,
             index_config_uri,
             overwrite,
+            assume_yes,
         }))
     }
 
@@ -363,7 +368,6 @@ impl IndexCliCommand {
             .value_of("config")
             .map(Uri::try_new)
             .expect("`config` is a required arg.")?;
-
         let metastore_uri = matches
             .value_of("metastore-uri")
             .map(Uri::try_new)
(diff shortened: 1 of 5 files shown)


quickwit-cli/src/index.rs
```diff
@@ -784,6 +789,18 @@ pub async fn create_index_cli(args: CreateIndexArgs) -> anyhow::Result<()> {
         quickwit_storage_uri_resolver().clone(),
         quickwit_config.default_index_root_uri(),
     );
+
+    // On overwrite and if assume yes if false, ask the user to confirm the destructive operation.
+    if args.overwrite && !args.assume_yes {
+        // Stop if user answers no.
+        if !prompt_confirmation(
+            "This operation will delete all your index data. Is this ok?",
```
guilload, line 797:
```suggestion
            "This operation will delete all the pre-existing index data. Continue? (yes/no)",
```

quickwit-cli/src/index.rs
```diff
@@ -784,6 +789,18 @@ pub async fn create_index_cli(args: CreateIndexArgs) -> anyhow::Result<()> {
         quickwit_storage_uri_resolver().clone(),
         quickwit_config.default_index_root_uri(),
     );
+
+    // On overwrite and if assume yes if false, ask the user to confirm the destructive operation.
+    if args.overwrite && !args.assume_yes {
```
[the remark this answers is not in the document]
fmassot: Yes indeed. Now I feel like we need to go back to check the existence of the index in the CLI and not in the index service.

The main reason for putting the logic on the CLI is that the CLI will soon use the REST API to create an index and in this case I'm not a big fan of a POST/PUT REQUEST that will make a delete. WDYT?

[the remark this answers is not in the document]
guilload: I think that's fine. I don't think the API of the CLI has to map perfectly the Index API.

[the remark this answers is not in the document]
fmassot: > I don't think the API of the CLI has to map perfectly the Index API.

Yes, I agree with that too. I was just sharing my feeling that now I just want to put the logic of checking if an index exists + delete the index in the CLI  and remove the overwrite `parameter` from the `IndexService.create_index` method.

Anyway, right now, I'm ok to leave it like this and review the logic once we are calling the REST API from the CLI.

quickwit-metastore/src/metastore/mod.rs
```diff
@@ -69,10 +69,13 @@ pub trait Metastore: Send + Sync + 'static {
     /// Checks whether the metastore is available.
     async fn check_connectivity(&self) -> anyhow::Result<()>;
 
-    /// Checks whether the given index is in this metastore.
-    async fn check_index_available(&self, index_id: &str) -> anyhow::Result<()> {
```
fmassot, line 73: This `check_index_available` can make you believe that an existing index is not present when there is a network error.

quickwit-core/src/index.rs
```diff
@@ -353,17 +353,15 @@ pub async fn remove_indexing_directory(data_dir_path: &Path, index_id: String) -
 
 /// Resolve storage endpoints to validate.
 pub async fn validate_storage_uri(
-    metastore_uri_resolver: &MetastoreUriResolver,
```
fmassot, line 356: Here we validate a storage URI with a `metastore_uri_resolver`.... am I missing something?

quickwit-metastore/src/metastore/mod.rs
```diff
@@ -69,10 +69,13 @@ pub trait Metastore: Send + Sync + 'static {
...
-        self.index_metadata(index_id).await?;
-        Ok(())
+    /// Returns true if index is present in the metastore.
```
guilload, line 72:
```suggestion
    /// Returns whether an index exists in the metastore.
```

quickwit-metastore/src/metastore/mod.rs
```diff
@@ -69,10 +69,13 @@ pub trait Metastore: Send + Sync + 'static {
...
+    async fn is_index_present(&self, index_id: &str) -> MetastoreResult<bool> {
```
guilload, line 73: `index_exists`?

quickwit-cli/src/index.rs
```diff
@@ -594,7 +598,25 @@ pub async fn create_index_cli(args: CreateIndexArgs) -> anyhow::Result<()> {
...
+        &quickwit_config,
+        &index_config,
+    )
+    .await?;
+
+    // On overwrite and index present and `assume_yes` if false, ask the user to confirm the
+    // destructive operation.
+    let is_index_present = metastore.is_index_present(&index_id).await?;
+    if args.overwrite && is_index_present && !args.assume_yes {
+        // Stop if user answers no.
```
guilload, line 612:
```rust
 let prompt = format!(
            "This operation will overwrite the index `{}` and delete all its data. Do you want to proceed?",
            args.index_id
        );
        if !prompt_confirmation(&prompt, false) {
            return Ok(());
        }
```

quickwit-cli/src/index.rs
```diff
@@ -77,7 +77,9 @@ pub fn build_index_command<'a>() -> Command<'a> {
                     arg!(--"data-dir" <DATA_DIR> "Where data is persisted. Override data-dir defined in config file, default is `./qwdata`.")
                         .env("QW_DATA_DIR")
                         .required(false),
-                    arg!(--overwrite "Overwrites pre-existing index.")
+                    arg!(--overwrite "Overwrites pre-existing index. This will delete all existing data stored at `index-uri` before creating a new index.")
+                        .required(false),
+                    arg!(-y --"assume-yes" "Assume \"yes\" as an answer to all prompts and run non-interactively.")
```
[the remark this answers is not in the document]
fmassot: all right :)

quickwit-metastore/src/metastore/mod.rs
fmassot: better indeed

quickwit-cli/src/index.rs
```diff
@@ -332,12 +335,14 @@ impl IndexCliCommand {
             .expect("`index-config` is a required arg.")?;
         let data_dir = matches.value_of("data-dir").map(PathBuf::from);
         let overwrite = matches.is_present("overwrite");
+        let assume_yes = matches.is_present("assume-yes");
```
guilload, line 338:
```suggestion
        let assume_yes = matches.is_present("yes");
```

fmassot: and the test catched it :)

PR #1702, opened 2022-07-01
Fix bad error status & message on search REST requests. Close #1632.

### LINKED ISSUE #1632: HTTP method not allowed for wrong json
Any additional fields on the json request returns "HTTP method not allowed"
```
➜  quickwit-indices curl -X POST -H "Content-Type: application/json" \
                            -d '{"query":"body:cliid", "does not exist": true}' \
                            http://127.0.0.1:7280/api/v1/hdfs/search
{
  "error": "HTTP method not allowed"
}⏎
```

Ignoring the field or returning an error with the unrecognized field would be better.

fmassot:
### Description
Close #1632

### How was this PR tested?
Tests on search API were updated with a handler that combines all search handlers. By doing this, we can detect the errors seen in #1632.


fmassot:
My bad, I still have to fix the POST response.

sunisdown:
LGTM

fulmicoton (approved):

Issue #1697, opened 2022-07-01
Show quickwit config in the UI


fulmicoton:
We need to have a place in the API showing the Quickwit COnfig as deserialized / reserialized by Quickwit.

It would help showing defaults and would a great place for people to check if the config was read correctly.
It is also something people could copy paste in their bug report.

fmassot:
3 remarks: 
- when serializing, we should not leak the metastore password.
- quickwit config is a node config. We could show the current node config by default and let the user choose another node if he wants to
- showing quickwit version on the UI will be very helpful too. But again, quickwit version can be different on each node... This makes me think that we could add the version in the node state.

fmassot:
Here are the endpoints that we can expose under `/cluster` path:

- `/api/v1/cluster/state` returns the cluster state (serialization of chichat cluster state) 

- `/api/v1/config` returns the node config
```json
{
  "metastore_uri": "postgresql://*******"
  "other config fields": "and config values"
}
```

- `/api/v1/info` returns general info about the running binary: quickwit version, commit hash
```json
{
  "version": "0.3.1-nightly",
  "build_target": "aarch64-apple-darwin",
  "commit_short_hash": "993a25bf1",
  "commit _date": "2022-06-29",
}
```

fmassot:
I'm not sure what you mean by "showing the Quickwit COnfig as deserialized / reserialized by Quickwit". What I understand is that we need to show QuickwitConfig values in use, we need to be careful to show the default values if some variables were set in the yaml file instead of showing "null".

What do you mean by "deserialized/reserialized"? I'm just going to serialize the config with the right values, did you have anything else in mind?


Example of quickwit config in json
```
{
  "version":0,
  "cluster_id":"quickwit-default-cluster",
  "node_id":"node-silent-T4WS",
  "listen_address":"127.0.0.1",
  "rest_listen_port":7280,
  "gossip_listen_port":7280,
  "grpc_listen_port":"127.0.0.1",
  "grpc_public_addr": "..",
  "peer_seeds":[],
  "metastore_uri":"...",
  "default_index_root_uri":"....",
  "data_dir":"./qwdata",
  "indexer":{"split_store_max_num_bytes":100000000000,"split_store_max_num_splits":1000},"searcher":{"fast_field_cache_capacity":1000000000,"split_footer_cache_capacity":500000000,"max_num_concurrent_split_searches":100,"max_num_concurrent_split_streams":100},
 "storage":null
}
```

[...]

PR #1708, opened 2022-07-01
Remove cluster proto and add a REST endpoint to expose the cluster st…

fmassot:
…ate.

### Description
- [x] Remove cluster proto code
- [x] Add `/cluster/state` REST endpoint that return chitchat cluster state.
- [x] Update the UI to show the cluster state

I will handle #1697 in another PR.

### How was this PR tested?
Tests.

### Screenshot
Capture d’écran 2022-07-02 à 01 16 07


quickwit-serve/src/cluster_api/rest_handler.rs
```diff
@@ -20,51 +20,48 @@
...
     #[serde(default)]
     pub format: Format,
 }
 
-fn list_members_filter(
-) -> impl Filter<Extract = (ListMembersRequestQueryString,), Error = Rejection> + Clone {
-    warp::path!("cluster" / "members")
+fn cluster_state_filter(
+) -> impl Filter<Extract = (ClusterStateQueryString,), Error = Rejection> + Clone {
+    warp::path!("cluster" / "state")
```
fulmicoton, line 50: so there is only one route? Why not cluster/

fulmicoton (approved):

quickwit-serve/src/cluster_api/rest_handler.rs
fmassot: True. Let's put cluster and see if we need other routes later.

PR #1710, opened 2022-07-03
Add REST endpoint for config and build info. Fix #1697.

### LINKED ISSUE #1697: Show quickwit config in the UI

fmassot:
### Description

This PR is a follow up of PR #1708 (look at the diff from the last commit only).

- [x] Add REST endpoint for config and build info. Fix #1697.
- [x] Add UI page to show config.
- [x] Add UI page to show build info.
- [x] Show cluster id in the top bar

### Important note 1

As suggested by @guilload I tried to add a dropdown to be able to select a node ID and display its config/build. But I faced CORS requests as I wanted to make directly HTTP requests on the node of interest. I'm not sure if this is what we want. So for now, we can only display the build/config for the current node. We can probably review that once the control plane will be ready as we may be able to get all the info from it for example.

### Important note 2

I move the build.rs file from quickwit-cli to quickwit-server:
- pros: we have all the build info in quickwit-server, and this is the place where we want to have it
- cons: it's a bit weird to set this kind for things outside the built binary.

### How was this PR tested?

- [x] Add one test on new REST endpoints.
- [x] Update UI tests

### Screenshots

Capture d’écran 2022-07-06 à 16 54 22


quickwit-serve/src/node_info_handler.rs
```diff
@@ -55,5 +56,50 @@ fn node_config_handler(
 }
 
 async fn get_config(config: Arc<QuickwitConfig>) -> Result<impl warp::Reply, Rejection> {
-    Ok(warp::reply::json(&*config))
+    // We need to hide sensitive information from metastore URI.
+    let mut config_to_serialize = (*config).clone();
```
fmassot, line 60: This is lame... I decided to return the raw quickwit config and hide sensitive data from the metastore uri.
It's interesting to get that info but it does not give you the real config in use (some values if absent are then deduced from another value).

quickwit-core/src/lib.rs
```diff
@@ -33,6 +33,41 @@ pub use index::{
...
+pub struct QuickwitBuildInfo {
+    commit_version_tag: String,
+    cargo_pkg_version: String,
+    cargo_build_target: String,
+    commit_short_hash: String,
+    commit_date: String,
+    pub version: String,
+}
+
+impl QuickwitBuildInfo {
```
fulmicoton, line 48: That's one of the rare thing we don't care to test in unit test.
I'd just make it const.

```
pub struct QuickwitBuildInfo {
    commit_version_tag: &'static str,
    cargo_pkg_version: &'static str,
    cargo_build_target: &'static str,
    commit_short_hash: &'static str,
    commit_date: &'static str,
    pub version: &'static str,
}

const QUICKWIT_BUILD_INFO: QuickwitBuildInfo = build_quickwit_build_info()

const fn build_quickwit_build_info() {
    // ...
}
```

The main benefit is that you can get rid of the extra argument in quickwit-serve.
A second benefit is that the code express/enforce what it is. A bunch of information specific to the build.

quickwit-serve/src/node_info_handler.rs
fulmicoton: I think your comment is obsolete. You now clone the config and edit the value manually (which is nicer)

fulmicoton (approved):

quickwit-core/src/lib.rs
fmassot: good point. I moved the build function in quickwit-serve.

fmassot:
screenshot updated with new UI.

fulmicoton:
Can we merge? @fmassot

PR #1717, opened 2022-07-04
optimize fetch_docs

### LINKED ISSUE #1690: Rework docs fetching
#### One Searcher per Document
- Potentially excessive amount of searchers
- We never hit the docstore cache 
- We cache all uncompressed data from fetching docs (e.g. 10_000 hits all in different blocks -> 1MB blocks -> 10GB)

#### Decompression running on tokio thread:
- 1MB Block -> Decomp 2000MB/s
less than 0.1ms: should be fine

####  Unlimited Cache, when fetching documents.
- All compressed data is cached (e.g. 10_000 hits all in different blocks -> 0,2MB blocks -> 2GB)

## Design
Ideally we decompress a block only once, and limit the amount of caching.
- Use a single searcher
- Remove unlimited cache on `open_index`
- Debounce inflight requests
    - Order fetched docs by docid
    - Set docstore cache to number of parallel requests, so that requests on the the same block always hit the cache

### Alternative Design
Instead of de-bouncing inflight requests, we could expose the doc store checkpoint API. But the docstore reader lives on the segment level which is not exposed usually.

PSeitz:
### Description
deduplicate concurrent requests to the same source
use a single shared searcher for doc fetching
set docstore cache size to NUM_CONCURRENT_REQUESTS(10)
remove unlimited cache for fetching docs

closes #1690

### How was this PR tested?

Fetch 10_000 docs on hdfs data set
Executed 2 times
```
curl -X POST -H "Content-Type: application/json" \
                                             -d '{"query":"body:cliid","max_hits":10000}' \
                                             http://127.0.0.1:7280/api/v1/hdfs/search
```

#### Execution Time
~10x faster
```
Pre       "elapsed_time_micros": 7371078
Post      "elapsed_time_micros": 787165
```

#### Memory Usage
Drop from 10GB to 143MB (yes GB and MB :)

fetch_docs_pre
fetch_docs_post


fulmicoton:
have you  tested it?

quickwit-search/src/fetch_docs.rs
```diff
@@ -150,9 +150,11 @@ async fn fetch_docs_in_split(
     index_storage: Arc<dyn Storage>,
     split: &SplitIdAndFooterOffsets,
 ) -> anyhow::Result<Vec<(GlobalDocAddress, String)>> {
-    let index_reader = get_searcher_for_split(global_doc_addrs.len(), index_storage, split).await?;
+    let num_concurrent_requests = 10;
+    let index_reader = get_searcher_for_split_without_cache(1, index_storage, split).await?;
+    let searcher = Arc::new(index_reader.searcher());
     let doc_futures = global_doc_addrs.into_iter().map(|global_doc_addr| {
-        let searcher = index_reader.searcher();
+        let searcher = searcher.clone();
```
fulmicoton, line 157: oh? So I think we hadn't notice but we accidentally removed the mutability required on the search (I think you introduced a mutex in the search cache a long time ago or something like that).

Doesn't it suggests we can ditch the searcher pool and make the searcher cloneable?

quickwit-search/src/fetch_docs.rs
```diff
@@ -150,9 +150,11 @@ async fn fetch_docs_in_split(
     index_storage: Arc<dyn Storage>,
     split: &SplitIdAndFooterOffsets,
 ) -> anyhow::Result<Vec<(GlobalDocAddress, String)>> {
-    let index_reader = get_searcher_for_split(global_doc_addrs.len(), index_storage, split).await?;
+    let num_concurrent_requests = 10;
```
fulmicoton, line 153: nitpick: Make it const?

quickwit-search/src/fetch_docs.rs
```diff
@@ -150,9 +150,11 @@ async fn fetch_docs_in_split(
...
+    let index_reader = get_searcher_for_split_without_cache(1, index_storage, split).await?;
+    let searcher = Arc::new(index_reader.searcher());
     let doc_futures = global_doc_addrs.into_iter().map(|global_doc_addr| {
```
fulmicoton, line 156: Nothing in this function contract tells me whether the global_doc_addrs
Maybe the caller sorts them I am not sure.

Should we sort them in this function maybe?
Or document that it should be sorted and add an assert that.

fulmicoton (changes requested):
See comments inline

quickwit-search/src/fetch_docs.rs
PSeitz: good point, we should sort it here

PSeitz:
> have you tested it?

Update the original commit message

quickwit-search/src/fetch_docs.rs
```diff
@@ -146,13 +150,16 @@ async fn get_searcher_for_split(
 #[tracing::instrument(skip(global_doc_addrs, index_storage, split))]
 #[allow(clippy::needless_lifetimes)]
 async fn fetch_docs_in_split(
-    global_doc_addrs: Vec<GlobalDocAddress>,
+    mut global_doc_addrs: Vec<GlobalDocAddress>,
     index_storage: Arc<dyn Storage>,
     split: &SplitIdAndFooterOffsets,
 ) -> anyhow::Result<Vec<(GlobalDocAddress, String)>> {
-    let index_reader = get_searcher_for_split(global_doc_addrs.len(), index_storage, split).await?;
+    global_doc_addrs.sort_by(|a, b| a.doc_addr.cmp(&b.doc_addr));
```
fulmicoton, line 157: sort_by_key does not work here?

quickwit-storage/src/debouncer.rs
```diff
@@ -0,0 +1,307 @@
...
+///
+/// Since most Futures return an Result<V, Error>, this also encompasses the error.
+pub struct AsyncDebouncer<K, V: Clone> {
+    cache: Mutex<FnvHashMap<K, Shared<BoxFuture<'static, V>>>>,
+}
+
+impl<K, V: Clone> Default for AsyncDebouncer<K, V> {
+    fn default() -> Self {
+        Self {
+            cache: Default::default(),
```
fulmicoton, line 47: Nitpick: derive default should work here, no?

quickwit-storage/src/debouncer.rs
```diff
@@ -0,0 +1,307 @@
...
+
+/// The AsyncDebouncer debounces inflight Futures, so that concurrent async request to the same data
+/// source can be deduplicated.
+///
+/// Since we pass the Future potentially to multiple consumer, everything needs to be cloneable. The
+/// data and the future. This is reflected on the generic type bounds for the value V: Clone.
+///
+/// Since most Futures return an Result<V, Error>, this also encompasses the error.
+pub struct AsyncDebouncer<K, V: Clone> {
+    cache: Mutex<FnvHashMap<K, Shared<BoxFuture<'static, V>>>>,
```
fulmicoton, line 41: i love it!

quickwit-storage/src/debouncer.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// Instead of the future directly, a constructor to build the future is passed.
+    /// In case there is already an existing Future for the passed key, the constructor is not
+    /// used.
+    pub async fn get_or_create<T, F>(&self, key: K, build_a_future: T) -> V
+    where
+        T: FnOnce() -> F,
+        F: Future<Output = V> + Send + 'static,
+    {
+        // explicit scope to drop the lock
+        let fut_opt = { self.cache.lock().unwrap().get(&key).cloned() };
```
fulmicoton, line 62: loving the scope too.

quickwit-storage/src/debouncer.rs
```diff
@@ -0,0 +1,307 @@
...
+        if let Some(future) = fut_opt {
+            return future.await;
+        }
+
+        let fut = Box::pin(build_a_future()) as BoxFuture<'static, V>;
+        let fut = fut.shared();
```
fulmicoton, line 68: shared future don't poll themselves.
Let's just make sure we do not use this for the long term cache...

quickwit-storage/src/debouncer.rs
```diff
@@ -0,0 +1,307 @@
...
+        self.cache.lock().unwrap().insert(key.clone(), fut.clone());
+        let res = fut.await;
+
+        self.cache.lock().unwrap().remove(&key);
```
fulmicoton, line 72: I am a bit scared of the possibity for a consumer to call 
`get_or_create` and the dropping the resulting future.

The unbouncing map would be populated, but the future would then not be polled for a while. I then another consume comes in an asks for the future, it could get an error, not because the value is wrong but because of a server timeout for instance.

fulmicoton: (Note that I am not sure what we should do here)

fulmicoton (changes requested):
Great. I have one comment remaining... I don't have any solutions however... can you have a look ?

quickwit-storage/src/debouncer.rs
```diff
@@ -0,0 +1,307 @@
...
+        let key = (path.to_owned(), range);
+        debouncer
+            .get_or_create(key.clone(), || async move {
+                underlying.get_slice(&key.0, key.1).await
+            })
+            .await
+    }
+
+    async fn delete(&self, path: &Path) -> StorageResult<()> {
+        self.underlying.delete(path).await
```
fulmicoton, line 125: I think there is a very complicated race conditions.

It makes i possible to have

- start get #1
- start delete #2
- end delete #2
- start get #3
- end get #2
- end get #3

Where get #3 returns a key that added by delete.

quickwit-storage/src/debouncer.rs
PSeitz: But we explicitly call `.await` on line 70

quickwit-storage/src/debouncer.rs
PSeitz: I think the future always gets polled to completion since we clone and  await the future in `get_or_create`

quickwit-storage/src/debouncer.rs
PSeitz: it's less flexible, because it requires Default on `V`

quickwit-search/src/fetch_docs.rs
PSeitz: yes, that's better

quickwit-storage/src/debouncer.rs
fulmicoton: My example does not consider delete to be debounced.

Just Get being debounced with the current code.

quickwit-storage/src/debouncer.rs
fulmicoton: ah yes I keep forgetting it is a bit dumb.

quickwit-storage/src/debouncer.rs
```diff
@@ -0,0 +1,404 @@
...
+impl<K: Hash + Eq + Clone, V: Clone> AsyncDebouncer<K, V> {
+    /// Returns the number of inflight futures.
+    pub fn len(&self) -> usize {
+        self.cache.lock().unwrap().len()
+    }
+
+    /// Cleanup
+    /// In case there is already an existing Future for the passed key, the constructor is not
+    /// used.
+    pub fn cleanup(&self) {
```
fulmicoton, line 61:
```suggestion
    fn cleanup(&self) {
```

fulmicoton (approved):

PR #1720, opened 2022-07-05
Removed num_threads indexing resource setting.

fulmicoton:
It was not supposed to be configurable.
We want each pipeline to operate on a single tantivy
thread and produce one segment in most case.


guilload (approved):
TIL `IgnoredAny`

PR #1732, opened 2022-07-06
Tweaking cache eviction policy to help with super large queries.

### LINKED ISSUE #1714: A better fastfield / split footer cache policy
The current cache is using an LRU policy.

For very large queries involving more data than can fit in the cache this is a catastrophy, as it hits the LRU cache worst case sscenario and the cache hit rate will be 0%.

fulmicoton:
The new policy is like LRU, except we avoid admitting
a new item if it would mean evicting an item that has been accessed
in the last 60s.

The goal is to behave better on scan access patterns, without being as aggressive as
using a MRU strategy.

Closes #1714

### DIFF
--- a/quickwit-storage/Cargo.toml
+++ b/quickwit-storage/Cargo.toml
@@ -48,6 +48,7 @@ tracing = "0.1.29"
 [dev-dependencies]
 mockall = "0.11"
 tracing-subscriber = "0.3"
+tokio = { version = "1", features = ["full", "test-util"] }
 
 [features]
 testsuite = ["mockall"]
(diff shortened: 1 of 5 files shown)


fmassot (approved):
Sounds legit.

Do you have some cache miss ratio numbers on this?

We are using this cache policy for 3 pieces of data:
- global hotcache cache
- global fast field cache
- directory cache per query per split

On hotcache/fastfield I guess we would want some kind of mix between most frequently accessed/most recent hotcache.
I need more time to think about what should be the behavior of the caching for a directory. 

60 seconds seems like a good enough magic number but I have no idea of what a good value should be. To dig more into, I saw recent work on ClickHouse talking about ARC https://dbs.uni-leipzig.de/file/ARC.pdf. It's interesting.

fulmicoton:
image

That's the cache hit rate for a query where around half of the fast field fit in RAM.
Before the patch the cache hit rate is 0 for obvious reasons.

Issue #1536, opened 2022-05-24
More efficient on multi-indexes.


fulmicoton:
Currently, each index pipelines is spawning a lot of threads:
Blocking actors are executed in their own single threaded tokio runtime.

In addition, tantivy's IndexWriter allocate a large buffer to build its index.
By running several pipeline in parallel, we end up with an overall memory
footprint proportional to the number of indexes. (at least as a first approximation).

Finally, the number of file descriptor is also an issue.

For multi-indexes, we want to keep the individual index pipelines, as it proved
to be very simple to reason with, but we want to play on the orchestration to 
remove all of the above negative side-effects.

Packager & Merger are special. They have to run in a 

The solution will require:

# Allowing actors to run on a specific tokio runtime

The difference between blocking / async actors will be even lighter
than it was before. It should make it possible to simplify the actor framework code
even further. For instance, we will lose the need for a JoinHandle.

The ingest queues actor is blocking (as it relies on rocksdb). We need to have it 
run on its own single threaded runtime.

# Indexer scheduling

While the packager and the merger naturally should yield in between message, the 
indexer is different.

We want the indexer to yield once the IndexWriter is sent to the packager.
Also, we do not want to always wait for the scheduled commit.
If there are no more messages associated to an index, we want to stop indexing.

A robust and simple scheduling logic could be as follows:
- Start indexing record time t.
- The indexer should NOT yield in between messages.
- If no more messages are available or if `COMMIT_TIMEOUT` is passed, close the split and send it to the packager, yield and "go to sleep".
- Attempt to resume indexing at `t + COMMIT_TIMEOUT`

-----

- [x] Making the tokio runtime the actors run on configurable.
- [x] Change the scheduling logic of indexers

guilload:
Closed via #2810. The feature is behind the configuration flag `enable_cooperative_indexing`. 

Also, let's not forget to address #2671 later this year.

PR #1736, opened 2022-07-07
Make actor's tokio runtime configurable.

fulmicoton:
This PR is part of the effort of #1536,
handling of multiple indexes.
    
It changes the actor API to allow them to target
a specific tokio runtime.
    
After this PR, blocking actors are executed
in a shared tokio runtime, drastically reducing the
number of threads, and relies on cooperative
task yielding instead of multithreading for
concurrency.


quickwit-actors/src/spawn_builder.rs
```diff
@@ -83,15 +91,149 @@ impl<A: Actor> SpawnBuilder<A> {
...
+        let span = actor.span(&ctx);
+        let loop_async_actor_future =
+            async move { actor_loop(actor, inbox, ctx, state_tx).await }.instrument(span);
+        let join_handle = runtime_handle.spawn(loop_async_actor_future);
+        let actor_handle = ActorHandle::new(state_rx, join_handle, ctx_clone);
         (mailbox, actor_handle)
     }
 }
+
+fn process_command<A: Actor>(
```
fulmicoton, line 109: that's actually more or less a copy paste from the runner.rs file.

quickwit-indexing/src/actors/merge_planner.rs
```diff
@@ -53,10 +51,6 @@ impl Actor for MergePlanner {
         QueueCapacity::Bounded(0)
     }
 
-    fn runner(&self) -> ActorRunner {
-        ActorRunner::DedicatedThread
```
fulmicoton, line 57: no clue why we run that as a blocking actor.

quickwit-indexing/src/actors/packager.rs
```diff
@@ -102,15 +102,15 @@ impl Actor for Packager {
     }
 
     fn queue_capacity(&self) -> QueueCapacity {
-        QueueCapacity::Bounded(0)
+        QueueCapacity::Bounded(1)
```
fulmicoton, line 105: This had an impact on indexing throughput.

quickwit-actors/src/actor_handle.rs
```diff
@@ -188,7 +188,13 @@ impl<A: Actor> ActorHandle<A> {
 
     /// Waits until the actor exits by itself. This is the equivalent of `Thread::join`.
     pub async fn join(self) -> (ActorExitStatus, A::ObservableState) {
-        let exit_status = self.join_handle.join().await;
+        let exit_status = self.join_handle.await.unwrap_or_else(|join_err| {
+            if join_err.is_panic() {
+                ActorExitStatus::Panicked
+            } else {
+                ActorExitStatus::Killed
```
fmassot, line 195: just for your info, in tokio, the join error is either panic or a cancellation. Killed seems a bit harsh :)

quickwit-cli/src/main.rs
```diff
@@ -116,6 +117,34 @@ async fn jemalloc_metrics_loop() -> tikv_jemalloc_ctl::Result<()> {
     }
 }
 
+/// If a bunch of tokio runtimes need to be started for actors,
+/// return the right configuration.
+///
+/// TODO making it configurable could be useful in the
```
fmassot, line 123: `/// TODO making it configurable could be useful in the` future?

fmassot: ok repetition at the end, so you can ditch "in the".

quickwit-cli/src/main.rs
```diff
@@ -116,6 +117,34 @@ async fn jemalloc_metrics_loop() -> tikv_jemalloc_ctl::Result<()> {
...
+        }
+        CliCommand::Index(_) => Some(RuntimesConfiguration::default()),
+        CliCommand::Split(_) | CliCommand::Source(_) => None,
+    }
+}
+
+fn start_actor_runtimes(cli_command: &CliCommand) -> anyhow::Result<()> {
+    if let Some(runtime_configuration) = runtime_configuration_for_cmd(cli_command) {
+        quickwit_common::runtimes::initialize_runtimes(runtime_configuration)
+            .context("Failed to start runtimes")?;
```
fmassot, line 143: I like the dot at the end.

quickwit-common/src/runtimes.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn light_for_tests() -> RuntimesConfiguration {
+        RuntimesConfiguration {
+            num_threads_blocking: 1,
+            num_threads_non_blocking: 1,
+        }
+    }
+
+    pub fn with_num_cpus(num_cpus: usize) -> Self {
+        let num_threads_non_blocking = if num_cpus > 6 { 2 } else { 1 };
+        let num_threads_blocking = (num_cpus - num_threads_non_blocking).max(1);
```
fmassot, line 70: The rationale here is that we want to give the maximum number of threads to the CPU-consuming actors: Indexer, Merger, Packager.
You can add a comment maybe to justify the default values?

quickwit-indexing/src/actors/packager.rs
fmassot: On the contrary, we will consume more RAM.

fulmicoton: Yes I saw a real difference in throughput. I don't quite understand why to be honest. 
Probably some race conditions due to the limitation of the thread pool.

(Indexer just needs to pass this value to the packager to yield while Packager is waiting for someone to yield.)

quickwit-actors/src/actor_handle.rs
fulmicoton: Killed is correct.

Names are for the client of our API. 
We want a name that matches quickwit's naming here, not tokio's naming.

This task being cancelled can only happen if an actor got killed.

quickwit-cli/src/main.rs
fulmicoton: fixed

fulmicoton:
> LGTM. I'm only worried about the memory consumption with the change in the packager queue capacity. But I don't have any figures to justify my feeling.

If we have a single pipeline the usage is higher indeed.
If we have several the usage is already much better than before.

PR #1743, opened 2022-07-08
Actor can sleep

fulmicoton:


quickwit-common/src/lib.rs
```diff
@@ -47,11 +47,7 @@ pub fn into_u64_range(range: Range<usize>) -> Range<u64> {
 }
 
 pub fn setup_logging_for_tests() {
-    use std::sync::Once;
-    static INIT: Once = Once::new();
-    INIT.call_once(|| {
-        env_logger::builder().format_timestamp(None).init();
-    });
+    let _ = env_logger::builder().format_timestamp(None).try_init();
```
fulmicoton, line 50: It seems to work better. Probably has to do with the fact that all unit test create their own tokio runtime,

quickwit-actors/src/actor.rs
```diff
@@ -308,6 +318,24 @@ impl<A: Actor> ActorContext<A> {
         self.actor_state.pause();
     }
 
+    /// Puts the actor to sleep for the given duration.
+    ///
+    /// For the period of the sleep, the actor is like on pause:
+    /// It stops processing message, in a way that does not
+    /// consume CPU, but it still reacts to commands.
+    ///
+    /// If the actor will wake up before the `sleep_duration` is elapsed
```
fmassot, line 327:
```suggestion
    /// The actor will wake up before the `sleep_duration` is elapsed
```

Issue #1744, opened 2022-07-08
Avoid fetching timestamp fast field when there is no sorting on this field nor filtering


fmassot:

kstaken:
As an addendum to this, the system should also only load fast fields when absolutely needed. I believe currently they are fetched in a `warmer` process before the data selection occurs but on large scale queries, where many splits will produce no results, that will result in fetching a large amount of unneeded data and kill the effectiveness of the fast field cache.

The same also applies if a split only returns a single result, even if the query asks for sorting, a single result is already inherently sorted and fast fields aren't needed for that split unless required by other operations.

fmassot:
> I believe currently they are fetched in a warmer process before the data selection occurs but on large scale queries

True. When #1745 will be merged, we will prefetch the timestamp fast field only if we need to filter values. NB: we are already pruning splits that have no overlap between the requested time range and the split time range, in this case, no prefetch is done.


> The same also applies if a split only returns a single result, even if the query asks for sorting, a single result is already inherently sorted and fast fields aren't needed for that split unless required by other operations.

This may be hard to do as we don't know in advanced the result. We plan to work on fast fields codecs as we know to be more efficient (for sparse fast fields for example), looking at Parquet format can be interesting as we can apply filtering on blocks and thus reduce the number of bytes to fetch. This is still in discussion and we don't know yet when we will work on that.

fmassot:
Partly closed by #1749 

Issue #1748 is tracking another optimization that we can make.

Closing.

PR #1745, opened 2022-07-08
Don't fetch timestamp fast field when there is no timestamp filtering…

fmassot:
… in the query.

### Description

Partial fix of #1744: it only avoids the prefetch of the timestamp fast field when there is no timestamp filtering in the search request. This could be improved by not prefetching on a filter that has no overlap with the split time range.

### How was this PR tested?

Soon.

### DIFF
--- a/quickwit-search/src/collector.rs
+++ b/quickwit-search/src/collector.rs
@@ -280,13 +280,20 @@ impl Collector for QuickwitCollector {
         // starting from 0 for every leaves.
         let leaf_max_hits = self.max_hits + self.start_offset;
 
+        // Timestamp filter is needed when:
+        //   - there is a timestamp field.
+        //   - AND there is a start OR end timestamp.
         let timestamp_filter_opt = if let Some(timestamp_field) = self.timestamp_field_opt {
-            TimestampFilter::new(
-                timestamp_field,
-                self.start_timestamp_opt,
-                self.end_timestamp_opt,
-                segment_reader,
-            )?
+            if self.start_timestamp_opt.is_some() || self.end_timestamp_opt.is_some() {
+                TimestampFilter::new(
+                    timestamp_field,
+                    self.start_timestamp_opt,
+                    self.end_timestamp_opt,
+                    segment_reader,
+                )?
+            } else {
+                None
+            }
         } else {
             None
         };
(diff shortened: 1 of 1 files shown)


quickwit-search/src/collector.rs
```diff
@@ -280,16 +280,18 @@ impl Collector for QuickwitCollector {
         // starting from 0 for every leaves.
         let leaf_max_hits = self.max_hits + self.start_offset;
 
-        let timestamp_filter_opt = if let Some(timestamp_field) = self.timestamp_field_opt {
-            TimestampFilter::new(
+        // Create a timestamp filter only if necessary as, once created,
+        // we will read some bytes in the fast field file that are not in the hotcache.
+        let timestamp_filter_opt = self.timestamp_field_opt
+            .filter(|_| self.start_timestamp_opt.is_some() || self.end_timestamp_opt.is_some())
+            .map(|timestamp_field| TimestampFilter::new(
```
fmassot, line 287: We should use split metadata to create a timestamp filter only if there is an overlap between the query filter and the split start / end timestamp.

fulmicoton: I don't understand...
Split pruning already happened at this point?

quickwit-search/src/collector.rs
```diff
@@ -280,16 +280,18 @@ impl Collector for QuickwitCollector {
...
                 timestamp_field,
                 self.start_timestamp_opt,
                 self.end_timestamp_opt,
                 segment_reader,
-            )?
-        } else {
-            None
-        };
+            ))
+            .transpose()?
```
fulmicoton, line 293: The `transpose()?.flatten()` is way less readable.

fulmicoton (approved):

quickwit-search/src/collector.rs
fmassot: I need to think a bit more about this because this is not very readable.
But currently, what I want to avoid is to create a `TimestampFilter` when there is no timestamp pruning because instanciating a `TimestampFilter` will lead a read of fast field min and max.

quickwit-search/src/collector.rs
fmassot: Yes it's horrible. I finally came back to 2 good old `if else` conditions, I generally tried to use as few as possible nested if conditions, but in this case, I think this is ok.

quickwit-search/src/collector.rs
fulmicoton: > We should use split metadata to create a timestamp filter only if there is an overlap between the query filter and the split start / end timestamp.

I think I understand what you meant but as far as I can tell, your formulation is wrong.

- No overlap: the leaf split search will not be emitted anyway, there is no problem today
- Split timestamps entirely included in the search timestamp range: We do not need ot warmup the timestamp field and we do not need the filter. Right now we forget the timestamp filter. I opened #1748 accordingly.
- Other (= some overlap but not included)

fulmicoton:
I completed your PR in this:
https://github.com/quickwit-oss/quickwit/pull/1752/files

Issue #1749, opened 2022-07-11
Avoid warmup of timestamp field if timestamp filtering is not required in searhc stream as well


fulmicoton:
That's the same as #1744 and #1748 but for search stream

PR #1752, opened 2022-07-11
Improve fast field warm up logic

### LINKED ISSUE #1751: Avoid warming up both the default sort and the request sort
Right now both the default sort field  and the request specified sort field are warmed up.

fulmicoton:
Closes #1751
Related to #1749 (does not address the case where the entire split is
within the timestamp bounds)

### DIFF
--- a/quickwit-search/src/collector.rs
+++ b/quickwit-search/src/collector.rs
@@ -30,10 +30,10 @@ use tantivy::aggregation::intermediate_agg_result::IntermediateAggregationResult
 use tantivy::aggregation::AggregationSegmentCollector;
 use tantivy::collector::{Collector, SegmentCollector};
 use tantivy::fastfield::{DynamicFastFieldReader, FastFieldReader};
-use tantivy::schema::{Field, Schema};
+use tantivy::schema::Schema;
 use tantivy::{DocId, Score, SegmentOrdinal, SegmentReader};
 
-use crate::filters::TimestampFilter;
+use crate::filters::{TimestampFilter, TimestampFilterBuilder};
 use crate::partial_hit_sorting_key;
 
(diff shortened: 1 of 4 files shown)


quickwit-search/src/collector.rs
```diff
@@ -240,19 +240,25 @@ pub struct QuickwitCollector {
...
+        match &self.sort_by {
+            SortBy::DocId => {}
+            SortBy::FastField { field_name, .. } => {
+                fast_field_names.insert(field_name.clone());
+            }
+        }
         if let Some(aggregate) = self.aggregation.as_ref() {
             fast_field_names.extend(get_fast_field_names(aggregate));
         }
+        if let Some(timestamp_filter_buider) = &self.timestamp_filter_buider_opt {
```
fmassot, line 259:
```suggestion
        if let Some(timestamp_filter_builder) = &self.timestamp_filter_builder_opt {
```

quickwit-search/src/collector.rs
```diff
@@ -240,19 +240,25 @@ pub struct QuickwitCollector {
     pub start_offset: usize,
     pub max_hits: usize,
     pub sort_by: SortBy,
-    pub fast_field_names: HashSet<String>,
-    pub timestamp_field_opt: Option<Field>,
-    pub start_timestamp_opt: Option<i64>,
-    pub end_timestamp_opt: Option<i64>,
+    timestamp_filter_buider_opt: Option<TimestampFilterBuilder>,
```
fmassot, line 243: typo `timestamp_filter_builder_opt`

quickwit-search/src/collector.rs
```diff
@@ -283,20 +289,12 @@ impl Collector for QuickwitCollector {
         // Timestamp filter is needed when:
```
fmassot, line 289: I would remove the comment here.

PR #1767, opened 2022-07-13
Ensure replaced splits are in published state

### LINKED ISSUE #1764: Replace splits should ensure that all the replaced splits are in published state
Currently, a race condition exists where the data carried by a split can magically reappear if a split is deleted while being merged.

guilload:
### Description
Ensure replaced splits are in published state. Fix #1764. 

### How was this PR tested?
Added unit test


guilload:
This is ready for review.

fulmicoton (approved):

PR #1786, opened 2022-07-22
WIP: Add disk based cache.

### LINKED ISSUE #1761: Add an SSD cache
Right now, Quickwit only long term caches are:
- a cache use to store split footer (which contains the so-called hotcache) 
- a cache to cache fast field data (a column-oriented representation of a field)

Both are extremely good candidate for caching as the same piece of data is often accessed for different queries.
Right now those caches are in RAM only.

For fast field this is quite a problem, as they can be bulky.
We would like to be able to cache fast fields (and possibly more pieces of data in the future) on an SSD.

# More info

- The keys of this cache have the form `PathBuf, Range<usize>` where the range represents a range of bytes.
R- The data is immutable.

# Pitfalls

- The cache should be resilient to corruption. Validating a checksum on read might be wise.
- Killing the process in the middle of a download should not result in an invalid state (e.g. truncated value in cache) or a leak (some files needs to be removed manually)
- The number of file descriptor should be limited.
- Restarting should fast enough  (tm).
- It is ok to assume only one process is writing into the cache (but a write lock should be added, right now there are none).
- A configured capacity in Bytes should be strictly respected.
- Ideally eviction/admission logic should be decoupled and stackable with the RAM cache.
- We do not trust actime/mtime/ctime

# External pointers

There are a bunch of caches in qucikwit. The one discussed here is 
- https://github.com/quickwit-oss/quickwit/blob/03d51183c8dc26b8b17afc3e082fd985917223e1/quickwit-storage/src/cache/quickwit_cache.rs

Right now it only applies to .fast files, and it should remain this way for this ticket.
It is relatively easy to add a new SSD based implmenetaiton of `https://github.com/quickwit-oss/quickwit/blob/03d51183c8dc26b8b17afc3e082fd985917223e1/quickwit-storage/src/cache/mod.rs#L58-L66`.

# Pointers

- https://github.com/pkhuong/kismet-cache/tree/main/src
- https://github.com/mozilla/sccache/tree/main/src/lru_disk_cache

ChillFish8:
### Description

Closes #1761 

This PR adds the start of a disk-based LRU cache. And has the following structure and behaviour:

- The core disk handling is detached from the main LRU logic, via the `Store<D>` struct.
- The LRU logic is largely copied over from the memory implementation but with some adjustments made to the eviction behaviour ([notably the LRU state is reset to its original state if it turns out we don't want to evict any files. ](https://github.com/ChillFish8/quickwit/blob/qw-1761-add-ssd-lru-cache/quickwit-storage/src/cache/disk/cache.rs#L304-L316))
- Files are checked when the cache is opened and their checksums checked.
- The maximum number of file descriptors can be adjusted, the higher the number the more open files will be kept in an LRU cache rather than having to open and close different files when a request comes in.

#### Durability
Durability is a best-effort type system, focussing more on a reactive system than the pre-active system. I.e. 
The store will attempt to write all the data and flush it to the disk. However, it does not employ something like a write-ahead log to guarantee that the file will be saved correctly to the disk in the event of a sudden shutdown/power outage. Instead, the following solution is employed to prevent a corrupted state:

- File access (read, write) is logged to a file using fixed-length rows as part of a background writer thread, so as to not negatively affect performance too much.
- The log will batch sudden influxes of events up, write the changes to disk, and then flush afterwards. Waiting for the next event(s).
- The log itself is effectively an array of 'slots' where each slot contains the metadata for a single file `(slot_checksum, file_key, file_checksum, file_length)`.
- These slots by themselves are small enough to where the OS *should* be able to write the data atomically to disk anyway, however, in the case a partial write does occur due to a power outage or the likes. The system walks through the log, checking each 'slots' checksum. If the checksums do not align, the slot is marked as 'free' so that it can be overwritten in future. Only the files that have correct checksums *and are not marked as empty* are added to the list of valid files. The cache will then purge any files which are not marked as valid or have invalid checksums, thus preventing any leaking of files or invalid states from being left over.

#### Current Limitations
- Unix only support.
- LRU wrapper is not currently tested within quickwit itself.
- FileKeys are currently only calculated as the crc32 hash of the file path -> this could lead to conflicts which we don't want, and this probably wants to change. (In prototyping I use a metro or city hash rather than crc32)
- Writing slices to a given file can potentially cause some less than ideal behaviour if you then attempt to read a slice of a file which hasn't necessarily been written to entirely as the OS will pad the file with intermediate `0`s should the write range lie outside of the file length. TL;DR:
  *  Writing files using write_all, or writing incrementally without exceeding the current file_length 👍
  *  Writing files using write_range starting from a random point of the file and then reading a range from the start 👎 

### How was this PR tested?

Due to the WIP nature of this PR testing is limited to unit testing. Which in itself is slightly WIP still.

Testing covers:

- FileBackedDirectory handling of data.
- Access log standard functionality, durability and corruption recovery.


quickwit-storage/src/cache/disk/cache.rs
```diff
@@ -0,0 +1,330 @@
...
+
+impl Capacity {
+    fn exceeds_capacity(&self, num_bytes: usize) -> bool {
+        match *self {
+            Capacity::Unlimited => false,
+            Capacity::InBytes(capacity_in_bytes) => num_bytes > capacity_in_bytes,
+        }
+    }
+}
+
```
ChillFish8, line 37: Realistically this can be moved out into a shared file so that we're not duplicating code across the memory and disk implementation.

quickwit-storage/src/cache/disk/cache.rs
```diff
@@ -0,0 +1,330 @@
...
+                    warn!(
+                        error = ?e,
+                        "IO Error: Failed to remove file from disk cache. A retry will happen on the next write.",
+                    );
+                    failed_files.push(key);
+                };
+            }
+
+            failed_files
+        }).await;
```
ChillFish8, line 231: It seemed that we don't want to necessarily abort writing a file to the cache if another file can't be removed for some reason. Instead, the system just marks the file as something to attempt to remove on the next write. Extending this would probably add a backoff system or some method of tracking how many times a file has attempted to be removed.

quickwit-storage/src/cache/disk/file.rs
```diff
@@ -0,0 +1,106 @@
...
+        key: 0,
+        file_checksum: 0,
+        last_accessed: 0,
+        file_size: 0,
+    };
+    pub(crate) const RAW_SIZE: usize = std::mem::size_of::<FileKey>()
+        + std::mem::size_of::<u32>()
+        + std::mem::size_of::<u64>()
+        + std::mem::size_of::<u64>()
+        + std::mem::size_of::<u32>(); // Row checksum
```
ChillFish8, line 34: We *could* use `std::mem::size_of::<FileEntry>()` but this could potentially change, so I think it's better to leave it as is for now so you can explicitly see the structure of the slot.

quickwit-storage/src/cache/disk/store.rs
```diff
@@ -0,0 +1,910 @@
...
+    File::options().create(true).open(lock_file)?;
+
+    Ok(())
+}
+
+/// Attempts to delete the directory lock file.
+fn try_remove_lock(base_path: &Path) -> io::Result<()> {
+    let lock_file = base_path.join(".cache-lock");
+    std::fs::remove_file(lock_file)
+}
```
ChillFish8, line 65: The directory lock system is very naive and could really do with being improved, even if it just adds something like an exponential backoff for retrying to acquire the lock.

quickwit-storage/src/cache/disk/file.rs
fulmicoton: This is an interesting comment.

Alternatively you could use `mem::size_of::<FileEntry>`  and add a unit test that will break when the size changes.

```
/// This size is used in ... . Changing it is a dangerous operation as 
/// it will break ... .
assert_eq!(RAW_SIZE, 24);
```

quickwit-storage/src/cache/disk/file.rs
```diff
@@ -0,0 +1,106 @@
...
+fn read_be_u64(input: &mut &[u8]) -> Result<u64, TryFromSliceError> {
+    let (int_bytes, rest) = input.split_at(std::mem::size_of::<u64>());
+    *input = rest;
+
+    let converted = int_bytes.try_into()?;
+
+    Ok(u64::from_be_bytes(converted))
+}
+
+fn read_be_u32(input: &mut &[u8]) -> Result<u32, TryFromSliceError> {
```
fulmicoton, line 99: why be and not le?

quickwit-storage/src/cache/disk/file.rs
```diff
@@ -0,0 +1,106 @@
...
+        Ok(Self {
+            key,
+            file_checksum: checksum,
+            last_accessed,
+            file_size,
+        })
+    }
+}
+
+fn read_be_u64(input: &mut &[u8]) -> Result<u64, TryFromSliceError> {
```
fulmicoton, line 90: why be and not le?

quickwit-storage/src/cache/disk/file.rs
```diff
@@ -0,0 +1,106 @@
...
+        buff[16..24].copy_from_slice(&self.last_accessed.to_be_bytes());
+        buff[24..32].copy_from_slice(&self.file_size.to_be_bytes());
+
+        let row_checksum = crc32fast::hash(&buff[4..]);
+        buff[0..4].copy_from_slice(&row_checksum.to_be_bytes());
+
+        buff
+    }
+
+    pub(crate) fn from_bytes(mut buff: &[u8]) -> Result<Self, Corrupted> {
```
fulmicoton, line 68: We'll really need a unit test here.

quickwit-storage/src/cache/disk/access_log.rs
```diff
@@ -0,0 +1,570 @@
...
+    pub(crate) writer: AccessLogWriter,
+    pub(crate) metadata: Metadata,
+}
+
+fn open_log_file(base_path: &Path) -> io::Result<File> {
+    let file = File::options()
+        .create(true)
+        .read(true)
+        .write(true)
+        .open(base_path.join("files").with_extension("meta"))?;
```
fulmicoton, line 43: is there a benefit to using with_extensions here?

quickwit-storage/src/cache/disk/access_log.rs
```diff
@@ -0,0 +1,570 @@
...
+
+    for (key, index) in data.key_mapping.iter() {
+        let entry = data.entries[*index];
+        metadata.insert(*key, entry);
+    }
+
+    metadata
+}
+
+pub(crate) struct LoadedData {
```
fulmicoton, line 59: I love this.

quickwit-storage/src/cache/disk/access_log.rs
```diff
@@ -0,0 +1,570 @@
...
+#[derive(Clone)]
+pub(crate) struct AccessLogWriter(Sender<Event>);
+
+impl AccessLogWriter {
+    fn send_event(&self, event: Event) {
+        // TODO: Should this be handled?
+        let _ = self.0.send(event);
+    }
+
+    pub(crate) fn register_read(&self, file_key: u64, accessed_at: Duration) {
```
fulmicoton, line 125: The prototype is unclear here. How come is access_at a `Duration`?

quickwit-storage/src/cache/disk/access_log.rs
```diff
@@ -0,0 +1,570 @@
...
+                pending_changes: changes,
+                free_slots: data.free_slots,
+                entries: data.entries,
+                key_mapping: data.key_mapping,
+                writer,
+            };
+
+            instance.start()
+        })
+        .expect("spawn background thread"); // TODO: Handle error.
```
fulmicoton, line 161: Yeah ... it is fine to not handle this specific error.

quickwit-storage/src/cache/disk/access_log.rs
```diff
@@ -0,0 +1,570 @@
...
+    fn drop(&mut self) {
+        let _ = self.0.send(Event::Shutdown);
+    }
+}
+
+fn start_background_task(changes: Receiver<Event>, data: LoadedData, writer: File) {
+    std::thread::Builder::new()
+        .name("disk-cache-background-worker".to_string())
+        .spawn(move || {
+            let instance = FileAccessTask {
```
fulmicoton, line 151: it seems like a job for a tokio task or an actor here.

quickwit-storage/src/cache/disk/access_log.rs
```diff
@@ -0,0 +1,570 @@
...
+
+                if let Err(e) = self.handle_event(event) {
+                    error!(error = ?e, "IO Cache Error: Failed to handle event due to the the given IO error.");
+                }
+            }
+
+            self.try_flush_data();
+
+            // Wait until the next set of events come through.
+            match self.pending_changes.recv() {
```
fulmicoton, line 201: is there a reason why you did not put this match statement at the top of the loop?

quickwit-storage/src/cache/disk/access_log.rs
```diff
@@ -0,0 +1,570 @@
...
+    /// A lookup dictionary to map a given file key to their relevant index where
+    /// the file entry lies within `entries`.
+    key_mapping: HashMap<FileKey, usize>,
+
+    writer: File,
+}
+
+impl FileAccessTask {
+    fn start(mut self) {
+        loop {
```
fulmicoton, line 184: Ok... So the loop tries to avoid writing after every single event... but I don't think it does that great of a job?

quickwit-storage/src/cache/disk/access_log.rs
```diff
@@ -0,0 +1,570 @@
...
+                        error!(error = ?e, "IO Cache Error: Failed to handle event due to the the given IO error.");
+                    }
+                }
+                Err(_) => return,
+            }
+        }
+    }
+
+    fn try_flush_data(&mut self) {
+        if let Err(e) = self.writer.sync_data() {
```
fulmicoton, line 219: I think you need to sync metas too, because recording a writing change the size of the file.

quickwit-storage/src/cache/disk/store.rs
```diff
@@ -0,0 +1,910 @@
...
+        cursor += n as u64;
+    }
+
+    Ok(FileInfo {
+        checksum: hasher.finalize(),
+        file_size,
+    })
+}
+
+fn read_to_fill_buff(file: &File, mut buf: &mut [u8], mut offset: u64) -> io::Result<()> {
```
fulmicoton, line 584: I don't like the name/the implem/the contract here, but I am not sure which whay it should be fixed.

Let me explain.

The implem is needlessly complex, and I have literally no clue how many times seek is called here.
I would be tempted to replace it to seek + read_exact...

However this function has a different behavior if the file finishes earler...
I suspect this is voluntary... Yet the name of the function does not help knowing this important detail and the prototype suggest otherwise.

For instance it does not return how much was read.
Maybe the caller saves the day by checking the integrity of the data in the buf, but this is not exactly pretty.

quickwit-storage/src/cache/disk/store.rs
```diff
@@ -0,0 +1,910 @@
...
+    ///   OS will infill missing parts of the file with `0`s if the writer's slice
+    ///   range lies beyond the end of the file.
+    ///
+    /// TODO: Come up with a smarter solution to storing slices.
+    fn read_all(&self, key: FileKey) -> io::Result<Vec<u8>> {
+        let metadata = self.try_get_file_metadata(key)?;
+
+        let file = self.get_open_file(key)?;
+
+        let mut buffer = vec![0; metadata.file_size as usize];
```
fulmicoton, line 468: fs::read_all()?

quickwit-storage/src/cache/disk/store.rs
```diff
@@ -0,0 +1,910 @@
...
+
+fn calculate_file_info(file: &File) -> io::Result<FileInfo> {
+    let mut hasher = crc32fast::Hasher::new();
+
+    let file_size = file.metadata()?.len();
+    let mut cursor = 0;
+
+    let mut buffer = [0; 32 << 10];
+    while cursor < file_size {
+        let n = file.read_at(&mut buffer, cursor)?;
```
fulmicoton, line 568: why this use of read_at so much !? File implements io::Read.

quickwit-storage/src/cache/disk/mod.rs
```diff
@@ -0,0 +1,12 @@
+mod access_log;
+mod cache;
+mod file;
+mod store;
+
+pub(crate) use cache::DiskBackedLRUCache;
+
+fn time_now() -> std::time::Duration {
```
fulmicoton, line 8: nitpick:

```
#[derive(Clone, Copy)]
struct UnixTimestamp(i64);
```

or sticking to SystemTime would have been nicer

quickwit-storage/src/cache/disk/cache.rs
```diff
@@ -0,0 +1,330 @@
...
+        let data = convert_and_log_error(result, "get range").flatten();
+
+        self.handle_cache_stats(&data);
+
+        data
+    }
+
+    async fn get_all(&self, path: &Path) -> Option<OwnedBytes> {
+        let result = self.try_get_all(path).await;
+        let data = convert_and_log_error(result, "get all").flatten();
```
fulmicoton, line 265: convert_and_log_error returning an option + using flatten is too complicated.

quickwit-storage/src/cache/disk/cache.rs
```diff
@@ -0,0 +1,330 @@
...
+        let result = self.try_get(path, byte_range).await;
+        let data = convert_and_log_error(result, "get range").flatten();
+
+        self.handle_cache_stats(&data);
+
+        data
+    }
+
+    async fn get_all(&self, path: &Path) -> Option<OwnedBytes> {
+        let result = self.try_get_all(path).await;
```
fulmicoton, line 264: The type of result would be helpful here.

quickwit-storage/src/cache/disk/cache.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    while let Some((key, _)) = existing_cache.pop_lru() {
+        new_lru.put(key, ());
+    }
+
+    *existing_cache = new_lru;
+}
+
+fn convert_and_log_error<T>(res: anyhow::Result<T>, op: &str) -> Option<T> {
```
fulmicoton, line 318: I suspect this function is not very helpful

quickwit-storage/src/cache/disk/store.rs
```diff
@@ -0,0 +1,910 @@
...
+
+        if !entry.path().is_file() {
+            continue;
+        }
+
+        let file_key = entry
+            .file_name()
+            .to_string_lossy()
+            .strip_suffix(".data")
+            .and_then(|name| name.parse::<FileKey>().ok().and_then(|v| metadata.get(&v)));
```
fulmicoton, line 82: no crazy functional piping style please.

It's hard to read.

quickwit-storage/src/cache/disk/store.rs
```diff
@@ -0,0 +1,910 @@
...
+
+        let file_entry = match file_key {
+            Some(entry) => entry,
+            None => {
+                // We don't want to prevent cleaning up other files.
```
fulmicoton, line 87: very good point...

Logging somethign would be nice though

quickwit-storage/src/cache/disk/cache.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    async fn try_get(
+        &self,
+        path: &Path,
+        byte_range: Range<usize>,
+    ) -> anyhow::Result<Option<OwnedBytes>> {
+        let path = path.to_path_buf();
+        let store = self.store.clone();
+        let range =
+            tokio::task::spawn_blocking(move || store.get_range(&path, byte_range)).await??;
```
fulmicoton, line 104: that should be async.

quickwit-storage/src/cache/disk/store.rs
```diff
@@ -0,0 +1,910 @@
...
+    ///   exist, this can occur when a file is made up purely of slices where the
+    ///   OS will infill missing parts of the file with `0`s if the writer's slice
+    ///   range lies beyond the end of the file.
+    ///
+    /// TODO: Come up with a smarter solution to storing slices.
+    fn read_range(&self, key: FileKey, range: Range<usize>) -> io::Result<Vec<u8>> {
+        let metadata = self.try_get_file_metadata(key)?;
+
+        // Short circuit if the range is beyond bounds.
+        if range.start as u64 >= metadata.file_size {
```
fulmicoton, line 437: io Error might make more sense.

fulmicoton:
Cool work! I like the idea of the access log with fixed-size entries and a freelist. It is not-so-common solution and I think this is a good fit here.

A markdown document in the directory describing the overall logic would be really useful.
Right now I had to look at the implementation to understand the logic.

Something describing roughly
- the read path
- the write path
- file cleanup
- what happens if there is a collision. 

would have been helpful.

PR #1788, opened 2022-07-22
[WIP] Add generic way to implement probes for services

guilload:
### Description
Per title

### How was this PR tested?
Pending


guilload:
I will go with a more straightforward approach in my next probe PR.

PR #1792, opened 2022-07-25
Adds a `on_drained_messages` event to actors.

fulmicoton:


fmassot (approved):

quickwit-actors/src/actor.rs
```diff
@@ -166,6 +166,21 @@ pub trait Actor: Send + Sync + Sized + 'static {
         Ok(())
     }
 
+    /// This function is called after a series of one or message have been processed and
```
fmassot, line 169:
```suggestion
    /// This function is called after a series of one, or several messages have been processed and
```

PR #1798, opened 2022-07-26
WIP: Distributed Kafka consumer based on group subscription

### LINKED ISSUE #1794: Add support for kafka indexing with consumer groups and  exactly-once semantics
As suggested by @kstaken in #1625 

```
omething else that could be considered here is that it should be possible to use the existing Kafka Consumer group logic without storing the offsets in Kafka. This will allow the usage of the group and rebalancing logic that already exists but you simply never call commit on the consumer it self.

For this to work and maintain exactly-once semantics you have to hook the rebalance events on the consumer and correctly handle the dynamic partition assignments when a rebalance occurs. I believe the rust rdkafka lib exposes everything required for this.

The approach is something like this.

Create a consumer in a group for the source with auto commit and offset storage disabled.
Add hooks for the rebalance events
Subscribe to the topic
Rebalance events will provide partition assignments
use the metastore offset to seek assigned partitions to the latest offset.
Start reading using the standard Kafka APIs
At each checkpoint store the offsets in the metastore
If a rebalance occurs while split creation is in progress and partition assignments change -> the in progress split will have to be invalidated and restarted on the latest assignments.
In a correctly functioning system this should be very rare
This should allow running N ingest nodes with exactly-once semantics, HA and dynamic scalability without having to implement the clustering manually.

In the metastore it will require a separate table for the offsets with 1 row per partition and the following fields.

source
topic
partition
offset
````

Also `https://github.com/kstaken/qw-kafka-test`

### PR DESCRIPTION
**Pull request by kstaken**
### Description

This converts kafka_source to be based on Kafka consumer group membership but with offset storage still maintained by Quickwit. Once completed this will maintain exactly-once processing semantics on postgres backed metastores while allowing multiple nodes to participate in data ingest.

When starting on this I got a little over zealous and had converted the metastore checkpoint storage into a separate table in Postgres. After discussing with @guilload we concluded that wasn't truly needed at this time so I cherry picked those commits into a new branch and removed that change from this PR. The JSON mutation approach for storing these offsets still makes me a little uncomfortable but in the short term it's still transactional so should be OK.

#### Status

* Existing kafka tests are passing and behavior matches existing implementation.
   * I do not think this behavior is correct though as it depends on the actor exiting when the end of the topic is reached. In streaming scenarios this is not what you want since low volume streams will hit this scenario continually. 
   * One possible solution for this is to add an option to KafkaSourceParams that determines whether the source acts as a read till end of topic source or as a streaming source.
* Current implementation still needs to handle split invalidation when a rebalance occurs and it looks like more changes are required to enable that without also triggering a Kafka rebalance. 
   * There's a hook in place to handle this just need the API to call.
   * As it stands this implementation will achieve at-least-once processing semantics.
* This still needs more automated tests to verify consumer behavior around rebalance events.
* check_connectivity still needs to be implemented as you don't want that to use the same consumer. 
   * How is this functionality used in the system? Should be trivial to implement. 

### How was this PR tested?

Kafka integration tests currently pass. I've done some basic manual testing with multiple Quickwit instances and confirmed rebalances and offset restoration appear correct. More advanced testing on verifying results is pending completion of exactly once processing behavior.

@guilload let me know if you want to take this over otherwise I'm happy to continue it. 

fixes #1794

### CODE REVIEW

File: quickwit-metastore/src/checkpoint.rs [hunk 1]
```diff
@@ -46,6 +46,20 @@ impl From<&str> for PartitionId {
     }
 }
 
+impl From<u32> for PartitionId {
+    fn from(partition_id: u32) -> Self {
+        let partition_id_str = format!("{:0>20}", partition_id);
```
fulmicoton, line 51: Let's not have this.
People can cast.

Also if such converters were needed, they should all call into the same converter.

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -68,29 +64,118 @@ impl TypedSourceFactory for KafkaSourceFactory {
...
+            let result = futures::executor::block_on(async {
+                // This is intended to invalidate the in progress split. Since partitions may be
+                // moving away, any in progress state will be invalid and a new split would need
+                // to be started from the most recently committed offsets.
+                // TODO: this can throw an error
+                let _result = self.rebalance_events.send(RebalanceEvent::Starting).await;
+
+                self.ctx
+                    .metastore
+                    .index_metadata(&self.ctx.index_id)
```
fulmicoton, line 108: If we fetch the checkpoint in the prebalance handler, I think we cannot really prevent pre-rebalance splits traveling in the pipeline from being published after the rebalance.

Right now the most likely thing that will happen is this split being correctly published, and the postrebaalance pipeline failing / restarting once all of the time.

The failure will in turn trigger a rebalance and the system may or may not stabilize.
 
Is there a plan here?

quickwit-metastore/src/checkpoint.rs
kstaken: Ah, yes of course, showing my rust newbie-ness.

quickwit-indexing/src/source/kafka_source.rs
kstaken: Yeah, anything that kills in flight splits needs to do it without taking down the kafka connection, otherwise it runs the risk of a rebalance cycle which isn't viable.  

If that's not possible then the serial number approach I described in #1794 should be sufficient to guard this but at the expense of having to complete all in flight splits before they can be invalidated at publication.

kstaken:
Closing, new PR will be coming based on the local feature branch.

Issue #1742, opened 2022-07-08
Separate parsing, validation, and transformation logic of config objects


guilload:
- We want to manipulate fully resolved and validated configuration objects without accessors.
- We want to know the source of the values (config file, CLI, env var)

It might be worth looking into good crates that could help us here.

See this [comment](https://github.com/quickwit-oss/quickwit/pull/1688/files#r916543980) for additional context.

guilload:
Closed via #1802

PR #1802, opened 2022-07-27
Decouple configuration parsing logic from validation and transformation

### LINKED ISSUE #1011: Specify and implement a generic way of overriding config parameters

boraarslan:
Closes #1011, #1742 

This PR separates the `QuickwitConfig` into two structs: `QuickwitConfig` and `QuickwitConfigObject`, and adds templating support to config files.

- `QuickwitConfig` is now only responsible for deserializing data. All other logic (value transformations, checking validities, accessing values) are transferred to `QuickwitConfigObject`

- `QuickwitConfigObject` does not have any accessor functions (one exception is `peer_seed_addrs` for DNS resolution purposes); instead, all of the required config values are generated when the config is initialized.

- All of the `QuickwitConfig` accesses are changed to `QuickwitConfigObject`.

- `StorageConfig` is removed because it is currently not used.

- Added templating support with docker-like syntax. For example:

> quickwit.yaml
> ```yaml
> rest_listen_port: ${REST_LISTEN_PORT:-4444}
> ```
replaces the value with `REST_LISTEN_PORT` or with `4444` if the given ENV_VAR couldn't be accessed.

We decided to drop some of the override requirements (_like half of them:_ overriding with environment vars and cli args, storing the source of the config value) since it would add too much complexity to the code.
The main reason for added complexity comes from accessing struct fields as strings and parsing them back. It is certainly possible, but we couldn't come up with anything simple and generic (all suggestions are welcome :) )

### DIFF
--- a/quickwit-cli/src/template.rs
+++ b/quickwit-cli/src/template.rs
@@ -1,0 +1,155 @@
+// Copyright (C) 2022 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use std::collections::HashMap;
+
+use anyhow::{bail, Context, Result};
+use new_string_template::template::Template;
+use once_cell::sync::Lazy;
+use quickwit_common::uri::Uri;
+use regex::Regex;
+use tracing::debug;
+
+// Matches ${value} if value is in format of:
+// ENV_VAR or ENV_VAR:DEFAULT
+// Ignores whitespaces in curly braces
+static TEMPLATE_ENV_VAR_CAPTURE: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(r"\$\{\s*([A-Za-z0-9_]+)(?:(?::\-)([\S]+))?\s*}").expect("Failed to compile regular expression. This should never happen! Please, report on https://github.com/quickwit-oss/quickwit/issues.")
+});
+
+pub fn render_config_file(contents: &[u8], uri: &Uri) -> Result<String> {
+    let contents_as_string = String::from_utf8(contents.to_vec())
+        .with_context(|| format!("Config file `{uri}` contains invalid UTF-8 characters."))?;
+    let template = Template::new(&contents_as_string).with_regex(&TEMPLATE_ENV_VAR_CAPTURE);
+    let mut data = HashMap::new();
+
+    for captured in TEMPLATE_ENV_VAR_CAPTURE.captures_iter(&contents_as_string) {
+        let env_var_name = captured.get(1).unwrap().as_str(); // Captures always have at least one match
+        let subst_val = {
+            if let Ok(env_var_value) = std::env::var(env_var_name) {
+                debug!(
+                    env_var_name,
+                    env_var_value, "Found ENV_VAR: {} with value: {}", env_var_name, env_var_value
+                );
+                env_var_value
+            } else if let Some(default_val) = captured.get(2) {
+                let default_val = default_val.as_str();
+                debug!(
+                    env_var_name,
+                    default_val,
+                    "Unable to get ENV_VAR specified: {}, Using default value instead: {}",
+                    env_var_name,
+                    default_val
+                );
+                default_val.to_string()
+            } else {
+                bail!(
+                    "Couldn't find ENV_VAR: {env_var_name} and the default value for the given \
+                     template"
+                );
+            }
+        };
+        data.insert(env_var_name, subst_val);
+    }
+
+    let rendered = template
+        .render(&data)
+        .with_context(|| format!("Failed to render templated config file `{uri}`."))?;
+    Ok(rendered)
+}
+
+#[cfg(test)]
+mod test {
+    use std::env;
+
+    use once_cell::sync::Lazy;
+    use quickwit_common::uri::Uri;
+
+    use super::render_config_file;
+
+    static TEST_URI: Lazy<Uri> = Lazy::new(|| Uri::new("file://config.yaml".to_string()));
+
+    #[test]
+    fn test_template_render() {
+        let mock_config = b"metastore_uri: ${TEST_TEMPLATE_RENDER_ENV_VAR_PLEASE_DONT_NOTICE}";
+        env::set_var(
+            "TEST_TEMPLATE_RENDER_ENV_VAR_PLEASE_DONT_NOTICE",
+            "s3://test-bucket/metastore",
+        );
+        let rendered = render_config_file(mock_config, &TEST_URI).unwrap();
+        std::env::remove_var("TEST_TEMPLATE_RENDER_ENV_VAR_PLEASE_DONT_NOTICE");
+        assert_eq!(rendered, "metastore_uri: s3://test-bucket/metastore");
+    }
+
+    #[test]
+    fn test_template_render_whitespaces() {
+        env::set_var(
+            "TEST_TEMPLATE_RENDER_WHITESPACE_QW_TEST",
+            "s3://test-bucket/metastore",
+        );
+
+        {
+            let mock_config = b"metastore_uri: ${TEST_TEMPLATE_RENDER_WHITESPACE_QW_TEST}";
+            let rendered = render_config_file(mock_config, &TEST_URI).unwrap();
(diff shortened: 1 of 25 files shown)


quickwit-cli/src/template.rs
```diff
@@ -0,0 +1,163 @@
...
+            "s3://test-bucket/metastore",
+        );
+        let rendered = render_config_file(mock_config).unwrap();
+        std::env::remove_var("TEST_TEMPLATE_RENDER_ENV_VAR_PLEASE_DONT_NOTICE");
+        assert_eq!(rendered, "metastore_uri: s3://test-bucket/metastore");
+    }
+
+    #[test]
+    fn test_template_render_whitespaces() {
+        let mock_config = OwnedBytes::new(
```
guilload, line 104: just write the tests one by one:
```rust
{
  let config = b"..."
  let rendered_config = ...
  assert_eq!(...)
}
{
 ...
}
```

quickwit-cluster/src/lib.rs
```diff
@@ -26,7 +26,7 @@ use std::sync::Arc;
 use anyhow::bail;
 use chitchat::transport::UdpTransport;
 use chitchat::FailureDetectorConfig;
-use quickwit_config::QuickwitConfig;
+use quickwit_config::QuickwitConfigObject;
```
guilload, line 29: Since `QuickwitConfigObject` is the object that developpers are mostly going to manipulate`, I'd rather name it `QuickwitConfig` and call the intermediate object `QuickwitConfigBuilder` or something like that.

quickwit-config/src/config.rs
```diff
@@ -69,8 +70,8 @@ fn default_node_id() -> String {
     new_coolid("node")
 }
 
-fn default_listen_address() -> Host {
-    Host::from(Ipv4Addr::LOCALHOST)
+fn default_listen_address() -> String {
+    "127.0.0.1".to_string()
```
guilload, line 74:
```suggestion
    Ipv4Addr::LOCALHOST.to_string()
```

quickwit-config/src/config.rs
```diff
@@ -56,7 +57,7 @@ fn default_metastore_uri(data_dir_path: &Path) -> Uri {
 }
 
 // See comment above.
-fn default_index_root_uri(data_dir_path: &Path) -> Uri {
+fn _default_index_root_uri(data_dir_path: &Path) -> Uri {
```
guilload, line 60: Remove then?

quickwit-config/src/config.rs
```diff
@@ -208,8 +193,88 @@ pub struct QuickwitConfig {
     #[serde(rename = "searcher")]
     #[serde(default)]
     pub searcher_config: SearcherConfig,
-    #[serde(rename = "storage")]
-    pub storage_config: Option<StorageConfig>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
guilload, line 198: This is never meant to be deserialized, so why do we derive  `Deserialize` and some attributes have some serde annotations?

quickwit-config/src/config.rs
```diff
@@ -208,8 +193,88 @@ pub struct QuickwitConfig {
...
+        validate_identifier("Node ID", &self.node_id)?;
+
+        if self.cluster_id == DEFAULT_CLUSTER_ID {
+            warn!(
+                cluster_id = DEFAULT_CLUSTER_ID,
+                "Cluster ID is not set, falling back to default value."
+            );
+        }
+        if self.peer_seeds.is_empty() {
+            warn!("Seed list is empty.");
```
guilload, line 235:
```suggestion
            warn!("Peer seed list is empty.");
```

quickwit-config/src/config.rs
```diff
@@ -208,8 +193,88 @@ pub struct QuickwitConfig {
...
+        }
+        let data_dir_uri = Uri::try_new(&self.data_dir_path.to_string_lossy())?;
+
+        if !data_dir_uri.protocol().is_file() {
+            bail!("Data dir must be located on local file system")
```
guilload, line 240:
```suggestion
            bail!("Data dir must be located on local file system. Current location: `{data_dir_uri}`.")
```

quickwit-config/src/config.rs
```diff
@@ -208,8 +193,88 @@ pub struct QuickwitConfig {
...
+                "Data dir `{}` does not exist.",
+                self.data_dir_path.display()
+            );
+        }
+        Ok(())
+    }
+
+    /// Returns the list of peer seed addresses. The addresses MUST NOT be resolved. Otherwise, the
+    /// DNS-based discovery mechanism implemented in Chitchat will not work correctly.
+    pub async fn peer_seed_addrs(&self) -> anyhow::Result<Vec<String>> {
```
guilload, line 253: Why have you kept those behind an accessor?

quickwit-config/src/config.rs
```diff
@@ -332,9 +395,18 @@ impl QuickwitConfig {
...
+                addr
+            }).with_context(|| {
+                format!(
+                    "Failed to parse advertise address `{advertise_addr}` read from \
+                     config file."
+                )
+            });
         }
-        if self.listen_address.is_unspecified() {
+        // This is safe, check the declaration.
```
guilload, line 408: What do you mean?

quickwit-config/src/config.rs
```diff
@@ -478,13 +549,19 @@ impl std::fmt::Debug for QuickwitConfig {
...
+/// Deserializes a valid [`Uri`].
+pub(super) fn deser_valid_uri<'de, D>(deserializer: D) -> Result<Uri, D::Error>
+where D: Deserializer<'de> {
+    let uri_opt: String = Deserialize::deserialize(deserializer)?;
+    Uri::try_new(&uri_opt).map_err(D::Error::custom)
+}
+
 /// Deserializes and validates a [`Uri`].
-pub(super) fn deser_valid_uri<'de, D>(deserializer: D) -> Result<Option<Uri>, D::Error>
+pub(super) fn deser_and_valid_uri<'de, D>(deserializer: D) -> Result<Option<Uri>, D::Error>
```
guilload, line 564:
```suggestion
pub(super) fn deser_and_validate_uri<'de, D>(deserializer: D) -> Result<Option<Uri>, D::Error>
```

quickwit-config/src/config.rs
boraarslan: This is still being used in `QuickwitConfig`. I changed the name to differentiate the two. I can change it to something else if you have any recommendations.

quickwit-config/src/config.rs
boraarslan: It was not working properly when I tried to generate addresses on config load time. It couldn't resolve the addresses of `peer_seeds` (probably due to DNS stuff but I couldn't understand why).

quickwit-config/src/config.rs
boraarslan: Artifact comment from my previous attempt to do something stupid. Not relevant anymore. Removed

quickwit-config/src/config.rs
```diff
@@ -478,13 +550,19 @@ impl std::fmt::Debug for QuickwitConfig {
...
     }
 }
 
+/// Deserializes a valid [`Uri`].
+pub(super) fn _deser_valid_uri<'de, D>(deserializer: D) -> Result<Uri, D::Error>
+where D: Deserializer<'de> {
+    let uri_opt: String = Deserialize::deserialize(deserializer)?;
+    Uri::try_new(&uri_opt).map_err(D::Error::custom)
+}
+
```
boraarslan, line 563: This is not being used. Should we remove it?

quickwit-cli/src/template.rs
```diff
@@ -0,0 +1,163 @@
...
+use std::collections::HashMap;
+
+use anyhow::{bail, Context, Result};
+use new_string_template::template::Template;
+use once_cell::sync::Lazy;
+use quickwit_storage::OwnedBytes;
+use regex::Regex;
+use tracing::{debug, warn};
+
+// Matches ${value} if value is in format of:
```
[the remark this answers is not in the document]
boraarslan: I am unsure if we should support `:+` since it would probably raise an error to deserialize empty strings. 

We could do something like removing the field from config if `VAR` is not set to a value. That way, it would fall back to its default value.

quickwit-config/src/config.rs
guilload: Let's investigate why this didn't work.

quickwit-config/src/config.rs
guilload: Yes.

quickwit-cli/src/template.rs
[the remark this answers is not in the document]
guilload: Ok, let's not support `:+`.

guilload (approved):

PR #1807, opened 2022-07-28
[Kafka source] Use `consumer.recv()` directly rather than `consumer.stream()`

guilload:
### Description
Simplify tokio select loop using  `consumer.recv()` directly rather than `consumer.stream()`

### How was this PR tested?
Ran Kafka source on an empty topic.

### DIFF
--- a/quickwit-indexing/src/source/kafka_source.rs
+++ b/quickwit-indexing/src/source/kafka_source.rs
@@ -197,6 +197,38 @@ impl ConsumerContext for RdKafkaContext {
     }
 }
 
+#[derive(Debug)]
+pub enum NumActivePartitions {
+    Initializing,
+    Some(usize),
+}
+
+impl NumActivePartitions {
+    fn dec(&mut self) {
+        match self {
+            Self::Initializing => panic!("We should not decrement the number of active partitions while initializing the source. This should never happen! Please, report on https://github.com/quickwit-oss/quickwit/issues."),
+            Self::Some(ref mut counter) => *counter -= 1,
+        }
+    }
+
+    fn is_zero(&self) -> bool {
+        matches!(self, Self::Some(0))
+    }
+
+    fn json(&self) -> serde_json::Value {
+        match self {
+            Self::Initializing => json!(null),
+            Self::Some(ref counter) => json!(counter),
+        }
+    }
+}
+
+impl Default for NumActivePartitions {
+    fn default() -> Self {
+        Self::Initializing
+    }
+}
+
 type RdKafkaConsumer = StreamConsumer<RdKafkaContext>;
 
 #[derive(Default)]
@@ -206,7 +238,7 @@ pub struct KafkaSourceState {
     /// Offset for each partition of the last message received.
     pub current_positions: HashMap<i32, Position>,
     /// Number of active partitions, i.e., that have not reached EOF.
-    pub num_active_partitions: usize,
+    pub num_active_partitions: NumActivePartitions,
     /// Number of bytes processed by the source.
     pub num_bytes_processed: u64,
     /// Number of messages processed by the source (including invalid messages).
@@ -256,8 +288,6 @@ impl KafkaSource {
             .context("Failed to resume from checkpoint.")?;
 
         let state = KafkaSourceState {
-            assigned_partition_ids: HashMap::new(),
-            num_active_partitions: 0,
             ..Default::default()
         };
         Ok(KafkaSource {
@@ -296,7 +326,7 @@ impl Source for KafkaSource {
                                 },
                                 RebalanceEvent::Assignments(assignments) => {
                                     debug!("Received partition assignments {:?}", assignments);
-                                    self.state.num_active_partitions = assignments.len();
+                                    self.state.num_active_partitions = NumActivePartitions::Some(assignments.len());
                                     self.state.assigned_partition_ids = assignments;
                                 }
                             }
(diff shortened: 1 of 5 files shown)


quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -89,6 +89,7 @@ impl ClientContext for RdKafkaContext {}
 
 impl ConsumerContext for RdKafkaContext {
     fn pre_rebalance(&self, rebalance: &Rebalance) {
+        info!("Pre rebalance {:?}", rebalance);
```
guilload, line 92: Temporarily adding those to understand what happens when adding/removing consumers/partitions.

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -268,13 +277,12 @@ impl Source for KafkaSource {
         batch_sink: &Mailbox<Indexer>,
         ctx: &SourceContext,
     ) -> Result<Duration, ActorExitStatus> {
+        let mut batch_num_bytes = 0;
         let mut docs = Vec::new();
         let mut checkpoint_delta = CheckpointDelta::default();
 
-        let deadline = tokio::time::sleep(quickwit_actors::HEARTBEAT / 2);
-        let mut message_stream = Box::pin(self.consumer.stream().take_until(deadline));
```
guilload, line 275: Removing the stream since we can directly poll the deadline future with `tokio::select!`.

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -296,86 +304,78 @@ impl Source for KafkaSource {
...
-                            }
-                            // FIXME: This is assuming that Kafka errors are not recoverable, it may not be the
-                            // case.
-                            Err(err) => return Err(ActorExitStatus::from(anyhow::anyhow!(err))),
-                        };
-                        if let Some(doc) = parse_message_payload(&message) {
-                            docs.push(doc);
-                        } else {
-                            self.state.num_invalid_messages += 1;
+                message_res = self.consumer.recv() => {
```
guilload, line 307: We get the message directly instead of an `Option<Message>` and remove one indentation level.

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -296,86 +304,78 @@ impl Source for KafkaSource {
...
-                        }
-                        ctx.record_progress();
+                        // FIXME: This is assuming that Kafka errors are not recoverable, it may not be the
+                        // case.
+                        Err(err) => return Err(ActorExitStatus::from(anyhow::anyhow!(err))),
+                    };
+                    if let Some(doc) = parse_message_payload(&message) {
+                        docs.push(doc);
                     } else {
-                        // TODO: this is probably not what should happen here.
```
guilload, line 354: Removing the comment since a source actor behaves precisely as suggested. The source actor is a spinning actor: when we exit `emit_batches`, the actor performs internal bookkeeping and then calls the method again.

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -296,86 +304,78 @@ impl Source for KafkaSource {
...
         }
-
-        // TODO: not entirely sure why this is here. Seems like this would prevent stream processing
-        if self.state.num_active_partitions == 0 {
-            info!(topic = %self.topic, "Reached end of topic.");
-            ctx.send_exit_with_success(batch_sink).await?;
-            return Err(ActorExitStatus::Success);
-        }
-
+        // TODO: Find a better way to stop the source for testing, this not longer works with consumer groups
```
guilload, line 373: We can always stop an actor and performs some assert, but that means that we need to first sleep long enough for the actor to do his work before sending the stop command. The beauty of stopping when partitions are no longer active is that we can just wait for the actor to exit gracefully and the test takes the minimum amount of time required.

kstaken: I was looking at adding a `streaming` bool parameter to `KafkaSourceParams`. This would default to `true` and then if set to `false` would set `enable.partition.eof:true` and then enable this block of code to exit once all partitions are caught up. Would also require a check to wait for the assignments from the initial rebalance before becoming active and then the tests should be reliable (I think). 

This would also provide the ability to run workloads to simply load a topic and then exit the source without continuing to poll it. We do this fairly often in various scenarios.

quickwit-indexing/src/source/kafka_source.rs
kstaken: Yeah, I agree. It was the partition.eof behavior that was making me question this.

quickwit-indexing/src/source/kafka_source.rs
kstaken: Behind the scenes there's actually only one event in the kafka protocol, the Rust lib then adds these additional callbacks at convenient points in its handler.

quickwit-indexing/src/source/kafka_source.rs
guilload: Agreed. Something similar already exists for the Kinesis source. I'll push something tomorrow.

quickwit-indexing/src/source/kafka_source.rs
guilload: I see. Good to know. Thanks for the explanation. One thing that I observed is that sometimes the same partitions are first revoked and then added again. Makes me question whether we should handle `Revoke` events.

kstaken: I was wondering the same but I'm pretty sure now that simple rebalances like this won't need it. 

There are more advanced rebalancing strategies that try to avoid stopping the entire consumer group while partitions move but those are only really needed when you're trying to minimize latency and I don't think the Rust lib has good support for those yet. This use case also won't really need that since there's inherent latency in the ingest process no matter what.

kstaken (approved):
This looks good.

PR #1811, opened 2022-07-28
Improve the reliability of the localstack container startup for tests

kstaken:
### Description

I've been running the tests on an M1 Mac and Docker may not have the best performance in this scenario. The tests pass for me about 20% of the time with various failures due to latency issues. This PR fixes one particular issue that I was seeing in the startup of the localstack container. The container runs the script that creates the bucket in the background which meant the container would be considered up and the tests would start running before the bucket was actually created. There is a surprising amount of latency in the bucket creation on my machine. Once the bucket gets created successfully the problem tests will pass but stopping the docker services will bring the issue back.

### How was this PR tested?

```
make test-all
make docker-compose-down
make test-all
```

### DIFF
--- a/.localstack/init.sh
+++ b/.localstack/init.sh
@@ -2,6 +2,9 @@
 
 set -eu
 
-awslocal kinesis create-stream --stream-name quickwit-dev-stream --shard-count 3
 awslocal s3 mb s3://quickwit-dev
 awslocal s3 mb s3://quickwit-integration-tests && awslocal s3 rm --recursive s3://quickwit-integration-tests
+
+if ! awslocal kinesis list-streams | grep quickwit-dev-stream ; then
+    awslocal kinesis create-stream --stream-name quickwit-dev-stream --shard-count 3
+fi
--- a/.localstack/init.sh
+++ b/.localstack/init.sh
@@ -5,6 +5,6 @@ set -eu
 awslocal s3 mb s3://quickwit-dev
 awslocal s3 mb s3://quickwit-integration-tests && awslocal s3 rm --recursive s3://quickwit-integration-tests
 
-if ! awslocal kinesis list-streams | grep quickwit-dev-stream ; then
+if ! awslocal kinesis list-streams | grep -q quickwit-dev-stream ; then
     awslocal kinesis create-stream --stream-name quickwit-dev-stream --shard-count 3
 fi
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,6 +23,13 @@ services:
       - "14250:14250"
       - "14268:14268"
       - "16686:16686"
+    # This dependency is to insure that localstack is fully initialized and the test bucket
+    # created before docker-compose considers the stack fully up. Without this the localstack
+    # image will be labeled ok with indeterminate latency in bucket creation because the
+    # init script to create the bucket runs in the background.
+    depends_on:
+      localstack:
+        condition: service_healthy
 
   localstack:
     image: localstack/localstack:latest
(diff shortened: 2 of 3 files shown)


kstaken:
Just found a scenario where this still fails. Closing until it's more reliable.

kstaken:
Reopened as I'm an idiot and forgot I had switched branches which lost these changes.

.localstack/init.sh
```diff
@@ -2,6 +2,9 @@
 
 set -eu
 
-awslocal kinesis create-stream --stream-name quickwit-dev-stream --shard-count 3
 awslocal s3 mb s3://quickwit-dev
 awslocal s3 mb s3://quickwit-integration-tests && awslocal s3 rm --recursive s3://quickwit-integration-tests
+
+if ! awslocal kinesis list-streams | grep quickwit-dev-stream ; then
```
guilload, line 8:
```suggestion
if ! awslocal kinesis list-streams | grep -q quickwit-dev-stream ; then
```

docker-compose.yml
```diff
@@ -23,6 +23,13 @@ services:
       - "14250:14250"
       - "14268:14268"
       - "16686:16686"
+    # This dependency is to insure that localstack is fully initialized and the test bucket
```
guilload, line 26:
```suggestion
    # This dependency is to ensure that localstack is fully initialized and the test bucket
```

docker-compose.yml
```diff
@@ -23,6 +23,13 @@ services:
...
+    # created before docker-compose considers the stack fully up. Without this the localstack
```
guilload, line 27:
```suggestion
    # created before docker-compose considers the stack fully up. Without this, the localstack
```

docker-compose.yml
```diff
@@ -23,6 +23,13 @@ services:
...
+    # image will be labeled ok with indeterminate latency in bucket creation because the
```
guilload, line 28: I had forgotten the nuance between `indeterminate` and `undetermined`. Thanks :)

docker-compose.yml
```diff
@@ -23,6 +23,13 @@ services:
...
+    # init script to create the bucket runs in the background.
```
guilload, line 29: Thanks for the great comment.

docker-compose.yml
```diff
@@ -23,6 +23,13 @@ services:
...
+    depends_on:
```
guilload, line 30: Docker Compose introduced a `--wait` option that waits for all the services to be healthy. I've just tested it local and it seems to work. We can remove this `depends_on` and the option in Makefile for the `docker-compose-up` command.

guilload (approved):
Pre-approved :) Thanks for the fix.

docker-compose.yml
kstaken: Yeah, that's a much better solution but of course things are never simple. It looks like docker-compose is now a CLI plugin which is only available as `docker compose` by default on a Mac. The `docker-compose` I had was V1 and I didn't realize it wasn't being updated by Docker Desktop. To get a wrapper to enable `docker-compose` to be based on V2 you need to brew install it separately. 

Is `docker compose` normally available on Linux? If so it may make sense to disambiguate and use that in the Makefile along with `--wait` since that's the future trajectory for this command.

guilload: That's just me obviously but `docker compose` seems to work on my machine (Manjaro):
```bash
guilload@x13 ~> docker compose version
Docker Compose version 2.6.1
```

kstaken: OK, changed to use `docker compose` and `--wait`. There is also significant latency on the startup of some of the other services before their health checks pass, especially Kafka.

Issue #1796, opened 2022-07-25
Index partitioning


fulmicoton:
We want the indexer to be able to produce more splits, with a better selectivity for tag pruning.
That routing logic would be configurable using a simple DSL

partiton_field: <partition_expr>

```
partition_expr: fieldname / HASH(field_name, 10) / partition_expr,partition_expr
```

Issue #1795, opened 2022-07-25
Merge siloes and filtering


fulmicoton:
This ticket is actually about two similar problems:

# Merge filtering.

Right now a MergePlanner is instantiated for each index pipeline.
It is initialized withthe list of existing splits as given by the metastore and updates itself with the newly created splits.

If several pipeline targetting the same index exist on the same node, or if on several nodes, merge planner will 
spawn conflicting merge operations, ending up as a failure on  publish, respawning the entire indexing pipeline.

In the future the plan is to handle this problem by centralizing the merge planning on a node responsible for the control plane.

However, we will probably have a need for a workaroudn to allow some clients to distribute their indexing using #1794.

# Merge siloes

We will soon introduce index partitioning, a functionality in which indexers are producing many splits that attempt to isolates tags.
Merging splits from different tags would defeat the purpose.

We need a way to tell the merge policy to consider the different partition independently. 

----

# Solution

We want to isolate the extract the merge pipeline out of the regular pipeline.
After the extraction, we should have:
- one merge planer per index.
- one supervized merge pipeline per index (downloader, executor, packager, uploader, publisher)
- one indexing pipeline per source

If there are several source on the same node indexing the same indexer, their respective publisher will route published splits to the same merge pipeline (associated to the index).

We can assume that the merge planner is simple enough to never die, and does not require supervision.

The rest of the pipeline however is tricky. 
The simplest is probably to have a supervised pipeline containing merge executor, packager, uploader, publisher. (no sequencer needed)

We need a mechanism for the merge planner to know about failed merge.
The merge pipeline supervisor could for instance send the merge planner a message to trigger the reload of the information from the SQL database (or we could have some Drop Guard travel with the splits... The Drop guard would lock/unlock the splits).

Once this is done, we only need to add a node_id to SplitMetadatas and have the merge planner only consider splits with the right node id. (Source_id is not necessary)

Finally we want, we want the merge planner to siloe the splits per partition_id (as added in #1821) and only consider merging split with the same partition_id.

fulmicoton:
Assigning to @guilload as I'm going in holiday and this might be blocking.

fmassot:
Merge siloes and filtering is done in #1865

But we can track other parts in new issues:
> The simplest is probably to have a supervised pipeline containing merge executor, packager, uploader, publisher. (no sequencer needed)

> We need a mechanism for the merge planner to know about failed merge.
[...]

PR #1821, opened 2022-08-01
Added partition_id to split metadata data.

fulmicoton:
Added partition_id to split metadata data.
    
    The `partition_id` is required for the merge planner to avoid
    merging  splits with different `partition_id`.
    
    Related to #1796 and #1795
    
    This PR also changes the hasher for the SipHasher, since nothing
    ensures that the implementation of the DefaultHasher won't change
    in the future.
    
    Finally, it seeds the hash using the expression itself, to avoid
    a class of collisions that could happpen with a change in the partition
    logic.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -64,9 +64,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.61"
+version = "1.0.62"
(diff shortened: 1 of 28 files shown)


guilload (approved):
I'd keep the routing expression around in the split metadata for debugging/investigation purposes.

PR #1824, opened 2022-08-01
Distributed indexing with Kafka based on consumer groups

### LINKED ISSUE #1794: Add support for kafka indexing with consumer groups and  exactly-once semantics

kstaken:
### Description

This converts kafka_source to be based on Kafka consumer group membership but with offset storage still maintained by Quickwit. Once completed this will maintain exactly-once processing semantics on postgres backed metastores while allowing multiple nodes to participate in data ingest.

Pending
- [x] Invalidate in flight splits during rebalance on the local node
- [ ] Guard against stale split publication from remote nodes 
- [ ] Optimize rebalancing so split invalidation only occurs if partition assignments change
- [x] Ensure no orphaned splits are left in `Staged` status by the exactly once guards. This currently happens with the checkpoint delta guard.
- [x] Merge planner fails when there are parallel indexers #1795 

This replaces PR #1798 to incorporate collaboration with @guilload. 

### How was this PR tested?

Kafka integration tests currently pass. Some basic manual testing with multiple Quickwit instances to confirm rebalances and offset restoration appear correct. More advanced testing to verify results is pending completion of exactly once processing behavior.

fixes #1794


fulmicoton:
@kstaken The PR is in mergeable state I believe. Can you test it on your data and see if it works for you?

fulmicoton (approved):

kstaken:
This looks good so far. Tested a few scenarios with 2 nodes sharing indexing responsibility and indexed counts have come out correct across rebalances and partition movements between nodes while active indexing is in progress. 

We'll update our larger K8s deployment and see how far we can push the indexing throughput now that we can utilize more than one node.

guilload:
Thanks for letting us know, @kstaken. I have yet to write an integration test with multiple consumers aggressively joining and leaving the group.

Issue #1208, opened 2022-03-21
Distributed indexing / control plane [design]


fulmicoton:
Currently Quickwit's indexing is not distributed. 

Our plan for distributed indexing is to rely on dispatching indexing over a set of indexing nodes, using 
a `(subset-of-partitions of a source)` as the insecable unit of job.

Let me redescribe these concepts here.
Indexing works by starting a so-called indexing pipeline, that consists of a managed set of actors.

The first actor is called a source, and it is in charge of defining a (partially more on this in a second) ordered stream of documents, and pushing them to the indexer.

At the end of the pipeline, the publisher publishes splits. That consists in uploading the split file to an object storage and updating 
the metastore. During that update, the split and its metadata are added AND the position up to which the indexing is done is updated. That position is expressed as a "Checkpoint".

As a first approximation, it is ok to consider the source as a [log](https://engineering.linkedin.com/distributed-systems/log-what-every-software-engineer-should-know-about-real-time-datas-unifying),
where each doc is an entry with a growing sequence number.

In reality, for scale purpose sources are partitionned. Each partition is its own independent log, and message are only ordered within a given log.

In order to make indexing distributed, we will need a way to decided which node should take care of the indexing of which `(index, set of source partitions)`, react to the loss of nodes, etc.

Ideally the design should fit both in an on-premise environment, AND on k8s.

sunisdown:
This can be divided into two parts

1. form an indexer cluster 
This can be easily obtained based on gossip. However, it is important to distinguish between indexer and searcher.

2. distribute the logs to the nodes in the indexer cluster
can partition the logs based on hash

fmassot:
We are working on this, we will share more details about the design soon :)

On the cluster formation, we already use [chitchat](https://github.com/quickwit-oss/chitchat) (based on anti-entropy gossip algorithm scuttlebutt). The good thing is that the gossip gives the metadata to get the indexer node states.

On the distribution of indexing, our goal is to have a control plane that distributes the indexing tasks (index/source/partitions set) on the indexers. Note that the distribution of the tasks will be oversimplified at the start,  the user will probably be able to configure it manually, also we won't handle the control plane's high availability in this first version.

sunisdown:
@fmassot  glad to know it. thanks for your reply. 
> On the distribution of indexing

Are you going to store the routing information in meta store and let the indexing task read the information from meta store?

fmassot:
[...]

PR #1825, opened 2022-08-01
 Add control plane first stone: a gRPC metastore service.

fmassot:
### Description

Add a metastore service with 2 implementations:
- One backed by a `Metastore` instance.
- One backed by a gRPC client.

### Important note

The PR still needs a lot of cleaning and testing but I think the API is good so ready to have some feedback about the API.


quickwit-control-plane/src/index.rs
```diff
@@ -0,0 +1,240 @@
+// Copyright (C) 2022 Quickwit, Inc.
```
fmassot, line 1: This is essentially a copy/paste from the functions present in `quickwit-core`

Cargo.toml
```diff
@@ -7,9 +7,9 @@ members = [
   "quickwit-cluster",
   "quickwit-common",
   "quickwit-config",
-  "quickwit-core",
```
fmassot, line 10: `quickwit-core` did not has really functionnal meaning so removing it is a good thing as now we have a place where we can put all stuff relative to index operations (control plane).

quickwit-cli/src/index.rs
```diff
@@ -34,12 +34,15 @@ use quickwit_common::GREEN_COLOR;
 use quickwit_config::{
```
fmassot, line 34: Essentially modifications due to the move of code from quickwit-core to quickwit-control-plane.

kstaken:
Is there a ticket that covers the design and reason for introducing the control plane service type and the plans for what it will solve? It looks like it's mirroring the metastore APIs, is it intended that nodes will talk to this service instead of talking to the metastore directly? Is it intended to be stateless? Will it be expected that you run more than one?

I know there's talk about having the merge planner run in this plane, what else is envisioned to live here?

fmassot:
@kstaken some short answers so that you know a bit what's going on there:

> Is there a ticket that covers the design and reason for introducing the control plane service type and the plans for what it will solve?

yes I will document that very soon (hopefully by next week)

> It looks like it's mirroring the metastore APIs, is it intended that nodes will talk to this service instead of talking to the metastore directly? Is it intended to be stateless? Will it be expected that you run more than one?

Yes, this is correct. For the moment the plan is to have indexers making calls to the control plane that reads/writes to the metastore.

> I know there's talk about having the merge planner run in this plane, what else is envisioned to live here? 

We will run the merge planner there indeed. We will also handle indexing scheduling to decide which node should index a given (source_id, set of partition_ids). The solution should obviously work well with indexing with Kafka consumers. 

There is an open issue #1208 but I need to add there the current plan.

fmassot:
Closing this old PR which is replaced by https://github.com/quickwit-oss/quickwit/pull/1853

PR #1826, opened 2022-08-02
Handle rebalance events in source rather than consumer context

guilload:
### Description
I moved the event handling logic into the main actor using sync channels in the consumer context. This simplifies the code, removes the need for messing with the tokio runtime, and makes testing easier.

### How was this PR tested?
Pending

### DIFF
--- a/config/quickwit.yaml
+++ b/config/quickwit.yaml
@@ -13,7 +13,7 @@ version: 0
 #
 # Node ID. Must be unique within a cluster. If not set, a random node ID is generated on each startup.
 #
-# node_id: node-1
+node_id: ${QW_NODE_ID:-node-1}
 #
 # Quickwit opens three sockets.
 # - for its HTTP server, hosting the UI and the REST API (TCP)
@@ -32,12 +32,13 @@ version: 0
 #   2. pass `0.0.0.0` and let Quickwit do its best to discover the node's IP (see `advertise_address`)
 #
 # listen_address: 127.0.0.1
-# rest_listen_port: 7280
+rest_listen_port: ${QW_REST_LISTEN_PORT:-7280}
 #
 # IP address advertised by the node, i.e. the IP address that peer nodes should use to connect to the node for RPCs.
 # The environment variable `QW_ADVERTISE_ADDRESS` can also be used to override this value.
 # The default advertise address is `listen_address`. If `listen_address` is unspecified (`0.0.0.0`),
 # Quickwit attempts to sniff the node's IP by scanning the available network interfaces.
+#
 # advertise_address: 192.168.0.42
 #
 # In order to join a cluster, one needs to specify a list of
@@ -52,13 +53,13 @@ version: 0
 # Path to directory where temporary data (caches, intermediate indexing data structures)
 # is stored. Defaults to `./qwdata`.
 #
-# data_dir: /path/to/data/dir
+data_dir: ${QW_DATA_DIR:-./qwdata}
 #
 # Metastore URI. Defaults to `data_dir/indexes#polling_interval=30s`,
 # which is a file-backed metastore and mostly convenient for testing. A cluster would
 # require a metastore backed by Amzon S3 or PostgreSQL.
 #
-# metastore_uri: s3://your-bucket/indexes
+metastore_uri: ${QW_METASTORE_URI:-./qwdata/indexes#polling_interval=5s}
 # metastore_uri: postgres://username:password@host:port/db
 #
 # When using a file-backed metastore, the state of the metastore will be cached forever.
--- a/quickwit-actors/src/envelope.rs
+++ b/quickwit-actors/src/envelope.rs
@@ -44,6 +44,14 @@ impl<A: Actor> Envelope<A> {
         self.0.message()
     }
 
+    pub fn message_typed<M: 'static>(&mut self) -> Option<M> {
+        if let Ok(boxed_msg) = self.0.message().downcast::<M>() {
+            Some(*boxed_msg)
+        } else {
+            None
+        }
+    }
+
     /// Execute the captured handle function.
     pub async fn handle_message(
         &mut self,
--- a/quickwit-indexing/src/actors/indexer.rs
+++ b/quickwit-indexing/src/actors/indexer.rs
@@ -168,9 +168,9 @@ impl IndexerState {
     ///
     /// This function will then create it, and can hence return an Error.
     async fn get_or_create_workbench<'a>(
-        &self,
+        &'a self,
         indexing_workbench_opt: &'a mut Option<IndexingWorkbench>,
-        ctx: &ActorContext<Indexer>,
+        ctx: &'a ActorContext<Indexer>,
     ) -> anyhow::Result<&'a mut IndexingWorkbench> {
         if indexing_workbench_opt.is_none() {
             let indexing_workbench = self.create_workbench()?;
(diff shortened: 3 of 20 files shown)


kstaken:
This isn't working for me. I see the pre_rebalance event but never see the post_rebalance event, it looks like it's blocking waiting for ack_rx  in pre_rebalance.

kstaken:
It looks like RebalanceEvent::Starting never makes it to emit_batches so the whole process blocks on the first rebalance.

guilload:
Yes, there's a deadlock t_t

guilload:
I fixed the deadlock. Since we're executing sync code in the rebalance hooks, calling `consumer.recv()` is actually sync and blocking. Running the consumer poll loop in a blocking tokio task solves the issue.

guilload:
Tests are not fully passing because we ignore the passed checkpoint. Therefore, they need to be rewritten.

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -446,6 +476,49 @@ impl Source for KafkaSource {
...
+        }
+    }
+}
+
+fn spawn_consumer_poll_loop(
+    consumer: Arc<RdKafkaConsumer>,
+) -> (JoinHandle<()>, flume::Receiver<KafkaMessage>) {
+    let (messages_tx, messages_rx) = flume::bounded(2);
+    let join_handle = spawn_blocking(move || {
+        while let Some(message) = consumer.poll(None) {
```
kstaken, line 513: So in trying to test this I'm now seeing the code block on this line and not process data, "most of the time". I have seen it process some data a couple times but I'm not sure what the magic incantation was for that to work. 

This indirection and message passing between threads seems kinda complex and problematic to me. I believe it is the result from trying to remove the `futures::executor::block_on` calls but all those were doing is blocking the rebalance thread until the async operations complete which is the behavior you want in that instance anyway. Obviously I'm not an expert on Rust or tokio so maybe there's some side effect in the simpler implementation that I'm not aware of, but looking at how this is behaving now it's hard to see what advantage it has and it seems to be kinda falling down a rabbit hole of complexity.

guilload: I have the same issue. The code does not block. The consumer is simply waiting for new messages... Indeed, moving the reassignment logic from the pre-rebalance hook to the post-rebalance one does not work. Mutating the rebalance object in this listener is no-op; see this [code](https://github.com/fede1024/rust-rdkafka/blob/master/src/consumer/mod.rs#L96). With the current code, the consumer resumes at the offset dictated by the reset policy (`auto.offset.reset`) after the rebalance. The default is `latest`.

For our use case, the Kafka Definitive Guide recommends seeking after the rebalance. So I tried that, and I hit two bugs in librdkafka. The current recommendation of the maintainers librdkafka is to simply not seek with consumer groups... The Java client seems to work fine, though. So I will open an issue and make a case for our cause. Otherwise, we will have to go with another strategy in the pre-rebalance where we lock each partition before consumption and receive a "publish token", which you already suggested in the past.

Regarding consuming messages in a tokio block task, I'm not trying to avoid `futures::executor::block_on`. Actually, the new calls to `send` and `recv` that I introduced in the pre and post-rebalance hooks are also blocking.

I'm trying to achieve two things:
1. Avoid blocking the tokio main loop
While `consumer.recv()` returns a future and seems non-blocking, this is not always the case. Sometimes, when calling `consumer.recv()`, the Kafka group coordinator informs the consumer that a rebalance is required, and the consumer calls the rebalance code right away in the same future. Unfortunately, the rebalance code is sync and while it is executing, the tokio main loop does not make any progress; hence the lock I observed in my first attempt. There are ways to write some "smart blocking code" that does not block for too long, but that's a risky endeavor.

2. Avoid sharing the state of the consumer with the consumer context
On pre or post-rebalance, we will eventually want to update the state of the consumer (increment the indexing generation, clear the current in-flight batch of docs, update some stats.), which means we will have to share the state of the consumer with the context via an `Arc<Mutex<>>` or `Arc<RwLock<>>`. Dealing with locks is hard. I find the message passing way much cleaner: first, the context sends a message to the consumer and waits for an ack; second, the consumer handles the message with unrestricted access to its state thanks to its `&mut self` methods, and finally acks. Some additional benefits are:
- it's easy to reason about the order of the events
- the rebalance logic exists in one place instead of being spread across two places
- the rebalance logic is easier to test

tl;dr 
Bear with me while I'm trying to make the seek offsets after rebalancing strategy work.

guilload: Also, I measured the message passing overhead on my machine:
- `main` branch: 250-400k messages/s
- this branch: 100k messages/s (channel size: 2)
- tweaking the size of the channel between the consumer and the actor: up to 1.2M messages/s with a channel size of 1,000.

quickwit-actors/src/envelope.rs
```diff
@@ -44,6 +44,14 @@ impl<A: Actor> Envelope<A> {
         self.0.message()
     }
 
+    pub fn message_typed<M: 'static>(&mut self) -> Option<M> {
```
guilload, line 47: Sorry for the noise, I started writing a test that needed `message_typed`, and ended up removing it, but I left this method.

config/quickwit.yaml
```diff
@@ -52,13 +53,13 @@ version: 0
...
 #
-# data_dir: /path/to/data/dir
+data_dir: ${QW_DATA_DIR:-./qwdata}
 #
 # Metastore URI. Defaults to `data_dir/indexes#polling_interval=30s`,
 # which is a file-backed metastore and mostly convenient for testing. A cluster would
 # require a metastore backed by Amzon S3 or PostgreSQL.
 #
-# metastore_uri: s3://your-bucket/indexes
+metastore_uri: ${QW_METASTORE_URI:-./qwdata/indexes#polling_interval=5s}
```
fulmicoton, line 62: can you revert this file? The default logic is acually more refined than `./qwdata/indexes#polling_interval=5s`

quickwit-indexing/src/actors/indexer.rs
```diff
@@ -168,9 +168,9 @@ impl IndexerState {
     ///
     /// This function will then create it, and can hence return an Error.
     async fn get_or_create_workbench<'a>(
-        &self,
+        &'a self,
```
fulmicoton, line 171: Wasn't it better before?

guilload: I was getting random compile errors here when some other parts of the code were not also not compiling. @kstaken did too. I never got them after adding this lifetime constraint on `&self`.

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -78,169 +77,125 @@ impl TypedSourceFactory for KafkaSourceFactory {
...
-
-    fn is_zero(&self) -> bool {
-        matches!(self, Self::Some(0))
-    }
-
-    fn json(&self) -> serde_json::Value {
-        match self {
-            Self::Initializing => json!(null),
-            Self::Some(ref counter) => json!(counter),
+        if let Rebalance::Revoke(tpl) = rebalance {
```
fulmicoton, line 164: I would move the Revoke part above... (it happens before the assign)

quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -279,173 +235,294 @@ impl KafkaSource {
...
+            topic=%self.topic,
+            partition=%partition,
+            num_inactive_partitions=?self.state.num_inactive_partitions,
+            "Reached end of partition."
+        );
+    }
+
+    fn should_exit(&self) -> bool {
+        self.backfill_mode_enabled
+            && self.state.num_inactive_partitions > 0
```
fulmicoton, line 375: comment

fulmicoton (approved):

PR #1833, opened 2022-08-05
Discard splits on revoke partitions

guilload:
### Description
This PR introduces the notion of "indexing generation". An indexing generation covers the period between two rebalances. Each time a rebalance occurs, the source increments the indexing generation. The current batches of messages and splits in flight become invalid.

The various actors involved along the pipeline check whether the indexing generation of the batches or the splits are current before performing any heavy operation (commit, upload) and discard them if necessary.

On publish, the indexing generation is locked. The lock is kept for the duration of the publish operation to avoid a race condition between publishers of generation N and consumers of generation N + 1 resuming after a rebalance.

### How was this PR tested?
Pending


quickwit-indexing/src/actors/sequencer.rs
```diff
@@ -50,8 +52,14 @@ impl<A: Actor> Actor for Sequencer<A> {
     fn observable_state(&self) {}
 }
 
+#[derive(Debug)]
+pub enum SequencerCommand<T: Debug> {
+    Cancel,
```
fulmicoton, line 57: need a comment on what `Cancel` means...

quickwit-indexing/src/actors/indexer.rs
```diff
@@ -369,9 +379,8 @@ impl Handler<CommitTimeout> for Indexer {
         commit_timeout: CommitTimeout,
         ctx: &ActorContext<Self>,
     ) -> Result<(), ActorExitStatus> {
-        if let Some(indexing_workbench) = self.indexing_workbench_opt.as_ref() {
-            // This is a timeout for a different split.
-            // We can ignore it.
+        if let Some(indexing_workbench) = &self.indexing_workbench_opt {
+            // If this is a timeout for a different workbench, we can ignore it.
```
fulmicoton, line 383:
```suggestion
            // If this is a timeout for a different workbench, we have to ignore it.
```

quickwit-indexing/src/models/indexing_sync.rs
```diff
@@ -0,0 +1,90 @@
...
+/// must:
+///
+/// 1. Acquire the publish lock.
+/// 2. Increment the indexing generation.
+/// 3. Release the publish lock.
+/// 4. Propagate to the downstream consumers the new current indexing generation object.
+#[derive(Debug, Default)]
+pub struct IndexingGenerationLeader(Arc<AtomicUsize>);
+
+impl IndexingGenerationLeader {
```
fulmicoton, line 39: Can we merge the lock and the generation stuff into one single object in order to enforce the contract described in the comment (both on the source / writer side and the actor side).

quickwit-indexing/src/actors/publisher.rs
```diff
@@ -136,17 +140,27 @@ impl Handler<SplitUpdate> for Publisher {
...
+                // TODO: Remove the junk right away?
+                info!(
+                    split_ids=?split_ids,
+                    splits_gen=%indexing_generation.this_gen(),
+                    current_gen=%indexing_generation.current_gen(),
+                    "Splits' indexing generation is no longer current."
+                );
+                return Ok(());
+            }
+            ctx.protect_future(self.metastore.publish_splits(
```
fulmicoton, line 155: Add a timeout

fulmicoton (changes requested):
See comments inline

PR #1840, opened 2022-08-09
Add support for BM25

### LINKED ISSUE #1712: Add support for bm25
- probably rename the sort_by_field option
- find a spec to describe sort by_field, but also sort by doc id and bm25 / desc and asc
- implement and plug bm25. (it requires adding fieldnorms to the warmup when it is enabled)

boraarslan:
Closes #1712 

This PR adds support for BM25 scoring in Quickwit.
To calculate the BM25 score, you need to opt in by passing `_score` to `sort_by_field` option.
For CLI, a new flag option called `--sort-by-score` is added.


docs/reference/rest-api.md
```diff
@@ -159,15 +153,14 @@ POST api/v1/_bulk -d \
...
+Ingest a batch of documents to make them searchable using the [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) bulk API. This endpoint provides compatibility with tools or systems that already send data to Elasticsearch for indexing. Currently, only the `create` action of the bulk API is supported, all other actions such as `delete` or `update` are ignored.
 :::caution
 The quickwit API will not report errors, you need to check the server logs.
 
 In Elasticsearch, the `create` action has a specific behavior when the ingest documents contain an identifier (the `_id` field). It only inserts such a document if it was not inserted before. This is extremely handy to achieve At-Most-Once indexing.
-Quickwit does not have any notion of document id and does not support this feature. 
+Quickwit does not have any notion of document id and does not support this feature.
```
boraarslan, line 161: Turns out I accidentally formatted some irrelevant stuff, but it shouldn't cause any problems.

quickwit-doc-mapper/src/sort_by.rs
```diff
@@ -90,9 +90,17 @@ pub enum SortBy {
...
+    Score {
+        /// Order to sort by. A usual top-k search implies a descending order.
+        order: SortOrder,
+    },
 }
 
 pub(crate) fn validate_sort_by_field_name(field_name: &str, schema: &Schema) -> anyhow::Result<()> {
+    if check_reserved_field_names(field_name).is_ok() {
+        return Ok(());
+    }
```
boraarslan, line 103: This is a bit unintuitive and is definitely open for suggestions.

quickwit-search/src/search_stream/leaf.rs
```diff
@@ -165,11 +165,17 @@ async fn leaf_search_stream_single_split(
         search_request.end_timestamp,
     );
 
+    let requires_scoring = match &search_request.sort_by_field {
+        Some(field_name) if field_name == "_score" => true,
+        _ => false,
+    };
+
```
boraarslan, line 172: As far as I understand, there is no way that stream requests have any sort order set, but I still added this in case I'm wrong or something changes in the future.

boraarslan: Still relevant. Thanks to Clippy, it is uglier now.

docs/concepts/querying.md
```diff
@@ -41,3 +41,7 @@ Quickwit does caching in many places to deliver a highly performing query engine
 
 - Hotcache caching: A static cache that holds information about a split file internal representation. It helps speed up the opening of a split file. Its size can be defined via the `split_footer_cache_capacity` configuration parameter.
 - Fast field caching: Fast fields tend to be accessed very frequently by users especially for stream requests. They are cached in a RAM whose size can be limited by the `fast_field_cache_capacity` configuration value.
+
+### Scoring
+
+Quickwit supports sorting docs by their BM25 scores. By default BM25 scoring is disabled to improve query times but it can be opt-in by setting `sort_by_field` option to `_score` in queries.
```
guilload, line 47: Did you end up fixing the fieldnorms option not being correctly propagated from the index config to tantivy?

guilload, line 47: This needs an example.

quickwit-search/src/collector.rs
```diff
@@ -39,19 +40,22 @@ use crate::partial_hit_sorting_key;
 /// The `SortingFieldComputer` can be seen as the specialization of `SortBy` applied to a specific
 /// `SegmentReader`. Its role is to compute the sorting field given a `DocId`.
 enum SortingFieldComputer {
-    SortByFastField {
+    FastField {
```
guilload, line 43: yes :)

quickwit-search/src/collector.rs
```diff
@@ -64,7 +68,12 @@ impl SortingFieldComputer {
                     SortOrder::Asc => u64::MAX - field_val,
                 }
             }
-            SortingFieldComputer::SortByDocId => 0u64,
+            SortingFieldComputer::DocId => 0u64,
+            SortingFieldComputer::Score { order } => match order {
+                // We lose some precision in order to properly cast the f32 to u64
+                SortOrder::Desc => score.mul(1000.0) as u64,
```
guilload, line 74: That works, but just for fun:
https://lemire.me/blog/2020/12/14/converting-floating-point-numbers-to-integers-while-preserving-order/

quickwit-search/src/leaf.rs
```diff
@@ -314,6 +319,27 @@ async fn warm_up_terms(searcher: &Searcher, query: &dyn Query) -> anyhow::Result
     Ok(())
 }
 
+async fn warm_up_fieldnorms(searcher: &Searcher, requires_scoring: bool) -> anyhow::Result<()> {
```
guilload, line 322: @PSeitz, is there some optimization we can do here knowing the terms in the query?

quickwit-search/src/tests.rs
```diff
@@ -377,6 +377,70 @@ async fn test_single_node_sorting_with_query() -> anyhow::Result<()> {
     Ok(())
 }
 
+// Sorry Evance for stealing your test
```
guilload, line 380: You know what they say about copy-pasting code, right? :)
Now, you've earned the right to write a helper function to run those two tests.

guilload (changes requested):

guilload:
So to order asc, one has to pass (REST) `&sort_by_field=_score` or `&sort_by_field=+_score`? and for desc?  `&sort_by_field=-_score`?

quickwit-directories/Cargo.toml
```diff
@@ -19,7 +19,7 @@ quickwit-storage = { version = "0.3.1", path = "../quickwit-storage" }
 serde = "1"
 serde_cbor = "0.11"
 serde_json = "1"
-tantivy = { git = "https://github.com/quickwit-oss/tantivy/", rev = "d24f31f", default-features = false, features = [
+tantivy = { git = "https://github.com/quickwit-oss/tantivy/", rev = "f4d7621", default-features = false, features = [
```
PSeitz, line 22: Try to make a separate PR to update tantivy in the future, so it's clear which changes are required by the tantivy upgrade

quickwit-search/src/leaf.rs
PSeitz: fieldnorms cost 1 byte per document per field. For a mature split with 5million documents that's 5MB per field. I would consider loading the whole file, since this will probably minimize the costs (we pay per S3 requests).

We have 70ms latency to first byte and around 70Mb/s.
In 210ms total, we can download 2 field norm fields.

Also we can improve the compression for fast fields in tantivy to less than 1byte per doc

guilload: Thanks, @PSeitz. Thorough answers are always appreciated.

quickwit-search/src/collector.rs
```diff
@@ -64,11 +67,30 @@ impl SortingFieldComputer {
...
+                }
+            }
         }
     }
 }
 
+/// Takes a u32 representation of a f32 and negates all bits if the most
+/// significant bit is set.
+fn sign_flip(value: u32) -> u64 {
+    let mut mask = (value as i64 >> 63) as u64;
```
guilload, line 89: The conversion from `u32` to `i64` will screw the order up. There are two ways to go about this:
```rust
/// Converts a float to an unsigned integer while preserving order.
/// See https://lemire.me/blog/2020/12/14/converting-floating-point-numbers-to-integers-while-preserving-order/
fn f32_to_u64(value: f32) -> u64 {
    let value_u32 = u32::from_le_bytes(value.to_le_bytes());
    let mut mask = (value_u32 as i32 >> 31) as u32;
    mask |= 0x80000000;
    (value_u32 ^ mask) as u64
}
```
OR
```rust
fn f32_to_u64_bora(value: f32) -> u64 {
    let value_u64 = u64::from_le_bytes((value as f64).to_le_bytes());
    let mut mask = (value_u64 as i64 >> 63) as u64;
    mask |= 0x8000000000000000;
    value_u64 ^ mask
}
```
According to Godbolt, the former generates more efficient machine code, so let's go with that along with a bunch of unit tests such as:
```rust
assert!(f32_to_u64(-1.0); < f32_to_u64(-0.9999999));
```
This is a great use case for property testing.

quickwit-search/src/leaf.rs
```diff
@@ -314,6 +319,27 @@ async fn warm_up_terms(searcher: &Searcher, query: &dyn Query) -> anyhow::Result
...
+    if !requires_scoring {
+        return Ok(());
+    }
+    let mut warm_up_futures = Vec::new();
+    for field in searcher.schema().fields() {
+        for segment_reader in searcher.segment_readers() {
+            let fieldnorm_readers = segment_reader.fieldnorms_readers();
+            let file_handle = fieldnorm_readers.get_inner_file().open_read(field.0);
```
guilload, line 330:
```suggestion
            let file_handle_opt = fieldnorm_readers.get_inner_file().open_read(field.0);
            if let Some(file_handle) = file_handle_opt {
                warm_up_futures.push(async move { file_handle.read_bytes_async().await })
            }
```

quickwit-doc-mapper/src/sort_by.rs
```diff
@@ -17,7 +17,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
-use anyhow::{bail, Context};
+use anyhow::{bail, Context, Ok};
```
guilload, line 20: Do we need this import?

quickwit-search/Cargo.toml
```diff
@@ -62,3 +62,4 @@ quickwit-metastore = { version = "0.3.1", path = "../quickwit-metastore", featur
 ] }
 serde_json = "1"
 tempfile = "3.3"
+proptest = "1.0"
```
guilload, line 65: <sub>nitpick: keep deps sorted</sub>

quickwit-search/src/collector.rs
```diff
@@ -64,11 +67,27 @@ impl SortingFieldComputer {
...
     }
 }
 
+/// Converts a float to an unsigned integer while preserving order.
+/// See `<https://lemire.me/blog/2020/12/14/converting-floating-point-numbers-to-integers-while-preserving-order/>`
+fn f32_to_u64(value: f32) -> u64 {
+    let value_u32 = u32::from_le_bytes(value.to_le_bytes());
+    let mut mask = (value_u32 as i32 >> 31) as u32;
+    mask |= 0x80000000;
+    (value_u32 ^ mask).into()
```
guilload, line 88:
```suggestion
    (value_u32 ^ mask) as u64
```

quickwit-search/src/collector.rs
```diff
@@ -525,4 +549,31 @@ mod tests {
...
+        // zero.
+        fn any_f32_without_negative_zero()(val in any::<f32>().prop_filter("Value can't be negative zero", |val| *val != -0.0)) -> f32 {
+            val
+        }
+    }
+
+    proptest! {
+        #![proptest_config(ProptestConfig::with_cases(10000))]
+        #[test]
+        fn test_proptest_f32_to_u64_compare_positive(a in 0.0..f32::MAX, b in 0.0..f32::MAX) {
```
guilload, line 565: Why do we need the first two tests? The last one seems to cover the same search space.

guilload (approved):

quickwit-search/src/collector.rs
boraarslan: You are right. I added the third one last and did not remove the first two. It is removed now.

quickwit-doc-mapper/src/sort_by.rs
boraarslan: Turns out we don't. I removed it.

Issue #1789, opened 2022-07-25
Add support for snippetting.


fulmicoton:
Tantivy offers a way to return a sample of the documents that contains some passage containiing the matched keywords.

It would be nice to add this feature to quickwit's API.

evanxg852000:
I would go with 
- Rest: `https://quickwit-host/api/v1/hdfs/search?query=body:foo&snippet_field=body`
- Cli: .quickwit index search --query "body:foo"  --snippet_field body

The following checks need to happen during search: 
- does snippet_field appear in the query or part of the default search fields?
- is snippet_field of field type that supports snippeting (text,..)
- Serialise the snippet

PR #1842, opened 2022-08-10
Support for snippet highlight

evanxg852000:
Implement support for snippet highlight in tantivy

Closes https://github.com/quickwit-oss/quickwit/issues/1789


quickwit-proto/proto/search_api.proto
```diff
@@ -168,13 +170,17 @@ message LeafHit {
   string leaf_json = 1;
   // The partial hit (ie: the sorting field + the document address)
   PartialHit partial_hit = 2;
+  // The highlighted part of the matching content 
+  optional string leaf_highlight_json = 3;
```
guilload, line 174: Why mix the `snippet` and `highlight` naming?

quickwit-proto/proto/search_api.proto
```diff
@@ -237,6 +243,13 @@ message FetchDocsRequest {
   // Index URI. The index URI defines the location of the storage that contains the
   // split files.
   string index_uri = 4;
+
+  // Search request. This is a perfect copy of the original search request,
+  // that was sent to root apart from the start_offset & max_hits params.
+  SearchRequest search_request = 5;
+
+  // `DocMapper` as json serialized trait.
+  string doc_mapper = 6;
```
guilload, line 252: Optional?

quickwit-proto/proto/search_api.proto
```diff
@@ -280,6 +293,9 @@ message SearchStreamRequest {
 
   // The field by which we want to partition
   optional string partition_by_field = 9;
+
+  // Fields to extract snippet on
```
guilload, line 297:
```suggestion
  // Fields to extract snippets on.
```

quickwit-search/src/fetch_docs.rs
```diff
@@ -144,30 +163,140 @@ async fn get_searcher_for_split_without_cache(
...
+    for (term, _) in terms {
+        if term.field() != field {
+            continue;
+        }
+        let term_str = if let Some(term_str) = term.as_str() {
+            term_str
+        } else {
+            continue;
+        };
+        let doc_freq = searcher.doc_freq_async(&term).await?;
```
guilload, line 270: Should we return the doc frequencies in partial hits to avoid another fetch of the term dictionary during the fetch docs phase?

quickwit-search/src/fetch_docs.rs
```diff
@@ -144,30 +163,140 @@ async fn get_searcher_for_split_without_cache(
...
+                        highlights_json: None,
+                    },
+                ));
+            }
+
+            let mut highlight = HashMap::new();
+            for (field, field_values) in doc.get_sorted_field_values() {
+                let field_name = searcher.schema().get_field_name(field);
+                if let Some(snippet_generator) = moved_snippet_generators.get(field_name) {
+                    let values = snippets_from_field_values(snippet_generator, field_values);
```
guilload, line 212: I wonder if we can avoid collecting to a vector here since we're serializing to JSON right after this step.

quickwit-search/src/search_response_rest.rs
```diff
@@ -48,12 +49,26 @@ impl TryFrom<quickwit_proto::SearchResponse> for SearchResponseRest {
...
-                    SearchError::InternalError(format!(
-                        "Failed to serialize document `{}` to JSON: `{}`.",
-                        hit.json, err
-                    ))
-                })
+                let mut hit_value = Map::with_capacity(2);
+                let document: serde_json::Value =
+                    serde_json::from_str(&hit.json).map_err(|err| {
+                        SearchError::InternalError(format!(
+                            "Failed to serialize document `{}` to JSON: `{}`.",
```
guilload, line 56: I don't think displaying the whole doc is a good idea. Maybe the first n chars?

quickwit-search/src/service.rs
```diff
@@ -150,6 +150,10 @@ impl SearchService for SearchServiceImpl {
         &self,
         fetch_docs_request: FetchDocsRequest,
     ) -> crate::Result<FetchDocsResponse> {
+        let search_request = fetch_docs_request
+            .search_request
+            .ok_or_else(|| SearchError::InternalError("No search request.".to_string()))?;
```
guilload, line 155: Would be nice to require the search request and the doc mapper only when snippeting is on.

quickwit-serve/src/search_api/rest_handler.rs
```diff
@@ -90,6 +90,11 @@ pub struct SearchRequestQueryString {
     #[serde(rename(deserialize = "search_field"))]
     #[serde(deserialize_with = "from_simple_list")]
     pub search_fields: Option<Vec<String>>,
+    /// Fields to extract snippet on
```
guilload, line 93:
```suggestion
    /// Fields to extract snippets on.
```

quickwit-serve/src/search_api/rest_handler.rs
```diff
@@ -90,6 +90,11 @@ pub struct SearchRequestQueryString {
...
+    #[serde(default)]
+    #[serde(rename(deserialize = "snippet_fields"))]
```
guilload, line 95: Remove the rename?

quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -82,41 +84,97 @@ fn has_range_clause(user_input_ast: &UserInputAst) -> bool {
...
+// Collects the fields nodes on term or range ast nodes
+fn collect_field_nodes(user_input_ast: &UserInputAst) -> Vec<Option<String>> {
     match user_input_ast {
         UserInputAst::Clause(sub_queries) => {
+            let mut fields = vec![];
             for (_, sub_ast) in sub_queries {
-                if needs_default_search_field(sub_ast) {
-                    return true;
-                }
+                fields.extend(collect_field_nodes(sub_ast));
```
guilload, line 109: There's probably to avoid those allocations.

quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -82,41 +84,97 @@ fn has_range_clause(user_input_ast: &UserInputAst) -> bool {
...
+        let field = schema
+            .get_field(field_name)
+            .ok_or_else(|| TantivyQueryParserError::FieldDoesNotExist(field_name.clone()))?;
+        fields.push(field);
+    }
+    Ok(fields)
+}
+
+// Collects the fields nodes on term or range ast nodes
+fn collect_field_nodes(user_input_ast: &UserInputAst) -> Vec<Option<String>> {
```
fmassot, line 104: the name `collect_field_nodes` is confusing, does it collect field nodes or field names?

quickwit-search/src/fetch_docs.rs
```diff
@@ -144,30 +163,140 @@ async fn get_searcher_for_split_without_cache(
     Ok(reader)
 }
 
+/// A struct for holding a fetched document's content and highlights.
+struct DocBundle {
```
fmassot, line 167: Not fan `DocBundle` name. Maybe `Document` with `content_json` and `snippets_json`?

quickwit-search/src/fetch_docs.rs
```diff
@@ -144,30 +163,140 @@ async fn get_searcher_for_split_without_cache(
...
 
     let index_reader = get_searcher_for_split_without_cache(index_storage, split).await?;
-    let searcher = Arc::new(index_reader.searcher());
+    let searcher = index_reader.searcher();
+    let snippet_generators =
+        Arc::new(create_snippet_generators(&searcher, doc_mapper, search_request).await?);
+
     let doc_futures = global_doc_addrs.into_iter().map(|global_doc_addr| {
         let searcher = searcher.clone();
+        let moved_snippet_generators = snippet_generators.clone();
```
fmassot, line 191: Generally we use `_clone` when we clone. `moved` in the name suggest it's already moved but it's not the case at this line :)

quickwit-search/src/fetch_docs.rs
```diff
@@ -144,30 +163,140 @@ async fn get_searcher_for_split_without_cache(
...
-) -> anyhow::Result<Vec<(GlobalDocAddress, String)>> {
+    doc_mapper: Arc<dyn DocMapper>,
+    search_request: &SearchRequest,
+) -> anyhow::Result<Vec<(GlobalDocAddress, DocBundle)>> {
     global_doc_addrs.sort_by_key(|doc| doc.doc_addr);
 
     let index_reader = get_searcher_for_split_without_cache(index_storage, split).await?;
-    let searcher = Arc::new(index_reader.searcher());
+    let searcher = index_reader.searcher();
+    let snippet_generators =
```
fmassot, line 186: Maybe we can have a struct like `MultiFieldSnippetGenerator`

fmassot: Apart from putting code related to snippet generation inside, one pros is that we will have a variable `Option< FieldsSnippetGenerator>` that show clearly that we use the generator only when there are snippet fields.

quickwit-search/src/fetch_docs.rs
fmassot: Not easy to do :/
We have a frequency per term and this would mean adding a list of (term, freq) for each partial hit. On possibility would be to retrieve from each leaf request the terms freq and put it then in the fetch_docs query but it seems too complex.

fmassot: Here what we can do is warming the freq before calling tantivy snippet. I believe this will avoid the copy paste from tantivy code.

quickwit-search/src/fetch_docs.rs
evanxg852000: I thought of this, but it involves implementing the serialization for the iterator and this seems to be discouraged by serde since serializing an iterator twice could yield different results.

quickwit-search/src/service.rs
evanxg852000: I found it makes the logic convoluted, but it has the advantage on send less data over the network when not needed.

quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -82,41 +84,97 @@ fn has_range_clause(user_input_ast: &UserInputAst) -> bool {
...
+        let field = schema
+            .get_field(field_name)
+            .ok_or_else(|| TantivyQueryParserError::FieldDoesNotExist(field_name.clone()))?;
+        fields.push(field);
+    }
+    Ok(fields)
+}
+
+// Collects the fields names on term and range ast nodes.
+fn collect_field_names(user_input_ast: &UserInputAst) -> Vec<Option<String>> {
```
fulmicoton, line 104: What is the result type about? why not `Vec<String>` ?

fulmicoton: The prototype and the comment are however not helping.

I suggest you write the functions

```rust
fn leaves(user_input_ast: &UserInputAst) -> Vec<&UserInputLeaf>;
fn extract_field_name(leaf: &UserInputLeaf) -> Option<&str>;
// calls leaves and extract_field_name
fn needs_default_search_field(ast: &UserInputLeaf) -> bool;
// calls leaves and extract_field_name
fn field_names(ast: &UserInputAst) -> HashSet<&str>;
```

fulmicoton: If you think this is overkill, you can also add a comment to the function.

evanxg852000: I am reusing it elsewhere to test if some ast nodes don't have field name specified. 
The naming is not great, I initially called it `collect_field_nodes`

fulmicoton:
@evanxg852000 Can you edit the doc too? I am not sure I understand the spec of the `snippet_fields` parameter

quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -64,59 +68,115 @@ pub(crate) fn build_query(
...
-        let field = schema
+/// Collects all the fields names on the query ast nodes.
+fn field_names(user_input_ast: &UserInputAst) -> HashSet<&str> {
+    collect_leaves(user_input_ast)
+        .into_iter()
+        .filter_map(extract_field_name)
+        .collect()
+}
+
+#[allow(clippy::needless_collect)]
```
fulmicoton, line 128:
```suggestion
```

quickwit-search/src/fetch_docs.rs
fulmicoton: yeah... let's just eventually get better at caching.

quickwit-search/src/search_response_rest.rs
fulmicoton: agreed... Be careful there. @guilload mentioned chars because you cannot truncate bytes as it might end up being an invalid string. `.chars().take().collect()`, or if you want to avoid the alloc, write a `truncate_str` helper function relying on `str.is_char_boundary()` and add it to quickwit-common.

fulmicoton (approved):
Approved to make the process smooth, but please address the truncation discussion.

PR #1850, opened 2022-08-13
Fix fieldnorms option is not propagated to tantivy

### LINKED ISSUE #1844: Text fieldnorms option is not propagated to tantivy
The option is basically ignored and fieldnorms are always generated. With a log management and analytics use case in mind, the default should be false.

boraarslan:
Fixes #1844


quickwit-search/src/lib.rs
```diff
@@ -205,6 +205,10 @@ pub async fn single_node_search(
     .map_err(|err| {
         SearchError::InternalError(format!("Failed to build doc mapper. Cause: {}", err))
     })?;
+
+    // Try to build query against current schema
+    let _query = doc_mapper.query(doc_mapper.schema(), search_request)?;
+
```
boraarslan, line 211: `single_node_search` now builds the query, so in cases where this function is used to search, if the query is invalid, it returns an error instead of returning a result with 0 hits. If this is something we don't want, I can revert the changes.

quickwit-search/src/tests.rs
```diff
@@ -310,34 +310,51 @@ async fn test_single_node_filtering() -> anyhow::Result<()> {
...
+                    {}
+                  - name: ts
+                    type: i64
+                    fast: true
+                  - name: temperature
+                    type: i64
+                    fast: true
+            "#,
+        fieldnorms_str
+    );
```
boraarslan, line 349: This is atrocious, I know. I will fix this before merging.

quickwit-search/src/lib.rs
```diff
@@ -205,6 +205,10 @@ pub async fn single_node_search(
     .map_err(|err| {
         SearchError::InternalError(format!("Failed to build doc mapper. Cause: {}", err))
     })?;
+
+    // Try to build query against current schema
```
guilload, line 209:
```suggestion
    // Validates the query by effectively building it against the current schema.
```

quickwit-search/src/tests.rs
```diff
@@ -310,34 +310,51 @@ async fn test_single_node_filtering() -> anyhow::Result<()> {
...
     )
-    .await?;
-    assert_eq!(single_node_response.num_hits, 0);
-    assert_eq!(single_node_response.hits.len(), 0);
-
+    .await;
+    assert!(single_node_response.is_err());
+    assert_eq!(
+        single_node_response.err().map(|err| err.to_string()),
+        Some("Invalid query: Field does not exists: 'tag'".to_string())
```
guilload, line 317: Can you fix the code that raises this error:
```Invalid query: field `tag` does not exist.```

guilload (approved):
LGTM except for the ugly test :)

quickwit-search/src/lib.rs
boraarslan: Yes, `single_node_search` is only used in CLI. I added the request validation as well.

quickwit-search/src/tests.rs
boraarslan: I believe the error message comes from tantivy's `QueryParserError::FieldDoesNotExist`.

PR #1852, opened 2022-08-15
Fix nightly build

evanxg852000:
- updated cross docker  images 
- installed up-to-date protobuf-compiler

### DIFF
--- a/build/cross-images/x86_64-unknown-linux-gnu.dockerfile
+++ b/build/cross-images/x86_64-unknown-linux-gnu.dockerfile
@@ -1,15 +1,14 @@
-FROM quickwit/cross-base:x86_64-unknown-linux-gnu
-# See https://github.com/quickwit-inc/cross
+FROM ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main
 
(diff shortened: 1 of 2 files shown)


build/cross-images/x86_64-unknown-linux-gnu.dockerfile
```diff
@@ -1,9 +1,15 @@
 FROM quickwit/cross-base:x86_64-unknown-linux-gnu
```
guilload, line 1: Two questions:
- What about installing `protoc` in the base image?
- What about installing `protoc` with the package manager: `apt-get install protobuf-compiler`?

evanxg852000: The `protobuf-compiler` package installs protoc version `3.6.1` this does not seem to support the experimental flag
`Error: Custom { kind: Other, error: "protoc failed: Unknown flag: --experimental_allow_proto3_optional\n" }`
We want a more recent version of protoc

Also, I am now switching to cross base image as they have updated to recent ubuntu versions. I think this can even help solve the rockdb-musl issue.

guilload (approved):
Sweet

PR #1855, opened 2022-08-16
Improve naming/comments related to index reader and caching.

fmassot:


quickwit-search/src/leaf.rs
```diff
@@ -118,11 +119,12 @@ pub(crate) async fn open_index(
...
 /// Opens a `tantivy::Index` for the given split.
 ///
-/// The resulting index uses a dynamic and a static cache.
+/// The resulting index uses a dynamic cache (deleted when the index is dropped)
+/// and a static cache global to Quickwit.
 pub(crate) async fn open_index_with_cache(
     index_storage: Arc<dyn Storage>,
     split_and_footer_offsets: &SplitIdAndFooterOffsets,
-    unlimited_cache: bool,
+    dynamic_cache: bool,
```
PSeitz, line 127: I think unlimited_cache is a better name here

fmassot: When `unlimited_cache` is set to false, we don't have at all this "dynamic cache". By the way, I don't really like "dynamic cache". From our usage, I would rather say "cache by query" as it's dropped once we return the result.

Currently, I believe that we have 2 cache layers:
- a cache by query which is unlimited
- a searcher cache or static cache used only for fast fields.

I would prefer to distinguish cache layers with this knowledge rather than using "unlimited_cache" that does not help to understand why we have a cache layer.

There is also another problème with `unlimited_cache`. By setting the variable to false, you can believe that you have then "limited_cache" but in fact there is no cache at all.

guilload: Maybe `query_cache` or `ephemeral_cache`. Then, the docs could state: `An ephemeral cache whose lifetime is tied to the query`.

PSeitz: The cache is not bound to the lifetime of the query, but the lifetime of the returned Index, so `ephemeral_cache` is better imo

fmassot: I added `An ephemeral unbounded cache directory whose lifetime is tied to the returned `Index``

guilload (approved):
Crystal clear.

PR #1864, opened 2022-08-21
Fix code coverage by starting only one indexer accross all tests.

fmassot:


quickwit-serve/src/test_utils/cluster_sandbox.rs
```diff
@@ -51,6 +51,10 @@ pub struct NodeConfig {
 ///
 /// The goal is to start several nodes and use the gRPC or REST clients to
 /// test it.
+/// 
+/// WARNING: currently we cannot start an indexer in different test as it will
```
guilload, line 55:
```suggestion
/// WARNING: Currently, we cannot start an indexer in a different test as it will
```

guilload (approved):

quickwit-serve/src/test_utils/cluster_sandbox.rs
fmassot: you are incredible :D

PR #1865, opened 2022-08-21
Add support for parallel indexing pipelines

guilload:
### Description
- Add `num_pipelines` parameter to source config
- Spawn `num_pipelines` in indexing service, each pipeline is assigned a `pipeline_ord` between 0 and `num_pipelines` - 1
- Add `source_id`, `node_id` and `pipeline_ord` to split metadata
- Fence split merges by `partition_id` and `pipeline_id`
- Disable garbage collector actors with `pipeline_ord` > 0

### How was this PR tested?
Add unit tests

### DIFF
--- a/quickwit-config/src/source_config.rs
+++ b/quickwit-config/src/source_config.rs
@@ -30,9 +30,25 @@ use crate::{is_false, validate_identifier};
 /// Reserved source ID for the `quickwit index ingest` CLI command.
 pub const CLI_INGEST_SOURCE_ID: &str = ".cli-ingest-source";
 
+fn default_num_pipelines() -> usize {
+    1
+}
+
+fn is_one(num: &usize) -> bool {
+    *num == 1
+}
+
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub struct SourceConfig {
     pub source_id: String,
+
+    #[doc(hidden)]
+    #[serde(default = "default_num_pipelines", skip_serializing_if = "is_one")]
+    /// Number of indexing pipelines spawned for the source on each indexer.
+    /// Therefore, if there exists `n` indexers in the cluster, there will be `n` * `num_pipelines`
+    /// indexing pipelines running for the source.
+    pub num_pipelines: usize,
+
     #[serde(flatten)]
     pub source_params: SourceParams,
 }
@@ -134,6 +150,13 @@ impl SourceConfig {
             SourceParams::IngestApi(params) => serde_json::to_value(params),
         }
         .unwrap()
+    }
+
+    pub fn num_pipelines(&self) -> Option<usize> {
+        match &self.source_params {
+            SourceParams::Kafka(_) | SourceParams::Void(_) => Some(self.num_pipelines),
+            _ => None,
+        }
     }
 }
 
@@ -269,7 +292,7 @@ impl TryFrom<KinesisSourceParamsInner> for KinesisSourceParams {
 #[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct VecSourceParams {
-    pub items: Vec<String>,
+    pub docs: Vec<String>,
     pub batch_num_docs: usize,
     #[serde(default)]
     pub partition: String,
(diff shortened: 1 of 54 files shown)


quickwit-config/src/source_config.rs
```diff
@@ -30,9 +30,25 @@ use crate::{is_false, validate_identifier};
...
+}
+
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub struct SourceConfig {
     pub source_id: String,
+
+    #[doc(hidden)]
+    #[serde(default = "default_num_pipelines", skip_serializing_if = "is_one")]
+    /// Number of indexing pipelines spawned for the source on each indexer.
+    /// Therefore, there will exist `num_pipelines` * `num_indexers` indexing pipelines for the
```
fmassot, line 48: (very small detail)

It makes you think that there is a variable `num_indexers` used somewhere.

Therefore, if you start `n` indexers, there will exist `num_pipelines` * `n` indexing pipelines for the source.

quickwit-indexing/src/actors/merge_planner.rs
```diff
@@ -66,46 +70,88 @@ impl Handler<NewSplits> for MergePlanner {
...
-        }
-        if has_new_young_split {
-            self.send_operations(ctx).await?;
+        let mut target_partition_ids = Vec::new();
+
+        let partitioned_new_young_splits = message
+            .new_splits
+            .into_iter()
+            .filter(|split| {
+                !self.merge_policy.is_mature(split) || {
```
fmassot, line 79: I see the problem you are solving, but I'm not a big fan of the usage `||`

quickwit-indexing/src/actors/merge_planner.rs
```diff
@@ -66,46 +70,88 @@ impl Handler<NewSplits> for MergePlanner {
...
     ) -> MergePlanner {
+        let mut partitioned_young_splits = HashMap::new();
+        for (partition_id, young_splits) in &published_splits
+            .into_iter()
+            .filter(|split| {
+                belongs_to_pipeline(&pipeline_id, split) && !merge_policy.is_mature(split)
+            })
+            .group_by(|split| split.partition_id)
+        {
+            partitioned_young_splits.insert(partition_id, young_splits.collect());
```
fmassot, line 119: A for loop is sometimes more readable than chaining, I think it's ok here.

but fyi, we can do a collect directly here.

```rust
    let data = vec![1, 3, 2, 2, 1, 0, 1, 2, 1, 2, 3, 4];
    let groups: HashMap<usize, Vec<usize>>  = data.iter()
        .group_by(|element| **element)
        .into_iter()
        .map(|(key, group)| (key, group.cloned().collect()))
        .collect();
```

quickwit-indexing/src/models/indexing_pipeline_id.rs
```diff
@@ -0,0 +1,26 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
+pub struct IndexingPipelineId {
```
fmassot, line 21: Why not `IndexingPipeline`?

one cons that we lose: adding the `id` enforces the idea of uniqueness
one pros: `IndexingPipelineId` is not an ID.

quickwit-config/src/source_config.rs
guilload: Good point.

quickwit-indexing/src/actors/merge_planner.rs
guilload: Yes, I will use an "explanatory" variable instead.

quickwit-indexing/src/actors/merge_planner.rs
guilload: I'm trying to extend the already existing vectors of young splits without allocating a temporary vector for each group. `Vec::extends` accepts an iterator, and I'm taking advantage of this API here.

quickwit-indexing/src/models/indexing_pipeline_id.rs
guilload: Yeah, I was using `IndexingPipelinePKey` at some point. `PKey` for primary key.

fmassot (approved):
Nice PR (and cleanup by the same occasion).

I thought the problem was more complex and was surprised by the simplicity of the code. Nice job.

PR #1866, opened 2022-08-21
Add a metastore grpc client

fmassot:


quickwit-cli/src/service.rs
```diff
@@ -70,15 +69,23 @@ impl RunCliCommand {
...
-            } else {
-                [QuickwitService::Indexer, QuickwitService::Searcher]
-                    .into_iter()
-                    .collect()
-            };
+        let services = matches
+            .value_of("service")
+            .map(|services_str| {
+                let services: Result<HashSet<_>, _> = services_str
+                    .split(',')
```
fulmicoton, line 76: why not several flags?

quickwit-metastore/src/metastore/grpc_metastore/mod.rs
```diff
@@ -0,0 +1,754 @@
...
+        })
+        .map(|member| member.grpc_advertise_addr)
+        .collect()
+}
+
+/// Updates channel endpoints by:
+/// - Sending `Change::Insert` grpc addresses not already present in `grpc_addresses_in_use`.
+/// - Sending `Change::Remove` event on grpc addresses present in `grpc_addresses_in_use` but not in
+///   `updated_members`.
+async fn update_channel_endpoints(
```
fulmicoton, line 431: is there code we could put in common with search?

quickwit-metastore/src/metastore/grpc_metastore/mod.rs
```diff
@@ -0,0 +1,754 @@
...
+                async move {
+                    if let Some(client) = client {
+                        Ok(client)
+                    } else {
+                        Err(std::io::Error::new(
+                            std::io::ErrorKind::Other,
+                            "Client already taken",
+                        ))
+                    }
+                }
```
fulmicoton, line 122:
```suggestion

                    client.take()
                        .ok_or_else(|| {
                               std::io::Error::new(
                            std::io::ErrorKind::Other,
                            "Client already taken",
                        )
                        })
                        
```

guilload: "Client already in use."

quickwit-cli/src/service.rs
fmassot: true, better to used `multiple_occurence` from clap.

quickwit-metastore/src/tests.rs
```diff
@@ -208,6 +208,7 @@ pub mod test_suite {
             .delete_index("non-existent-index")
             .await
             .unwrap_err();
+        println!("error {:?}", result);
```
fulmicoton, line 211:
```suggestion
```

quickwit-actors/src/lib.rs
```diff
@@ -77,3 +78,13 @@ pub enum AskError<E: fmt::Debug> {
     #[error("The handler returned an error: `{0:?}`.")]
     ErrorReply(#[from] E),
 }
+
+impl<E: fmt::Debug + ServiceError> ServiceError for AskError<E> {
+    fn status_code(&self) -> ServiceErrorCode {
+        match self {
+            AskError::MessageNotDelivered => ServiceErrorCode::Internal,
```
fulmicoton, line 85: poulet

quickwit-metastore/src/metastore/grpc_metastore/mod.rs
fmassot: The code is really different in the search pool. But it would be possible to put in common a diff function that would return something like a `Difference` struct with `added` and `removed` fields. I don't think it's worth it right now.

fulmicoton (approved):

PR #1868, opened 2022-08-22
Show BM25 scores when they are sorted by their score 

### LINKED ISSUE #1851: Show BM25 scores of docs on search response when they are sorted by their score
We want to see calculated BM25 results when the docs are sorted by their score. It possibly requires the internal gRPC API to be changed.

boraarslan:
Closes #1851


fulmicoton:
#1851 is a "nice to have". Changing all scores to use f32 without knowing if this a good idea or not seems overkill to me. Let's just not address the issue for the moment.

PR #1898, opened 2022-08-30
Rocks db free

fulmicoton:
This is a WIP.

Flushing my brain here so it can be resumed later.
Here is what is missing:

With the rocksdb solution was based on the idea that position were local to a given queue. My recordlog on the other hand uses a single global position.
As a complicated side effect, it did not implement queue creation/existence. Queues are created on the fly and just either have
records or are empty. We do not need to track the last position for each individual queues... just the last position.

I have not dug into whether queue existence is a requirement for quickwit, but I suspect it is.
The code will require some modificcation in that case.

I added a rather artificial create_queue remove_queue operation, but those are bogus as they are not durable. (they only update the memqueue). The problem is trickier than just adding a Create operation to the log, as it needs to survive truncation.

- the position returned by the queue are not necessarily contiguous, so the fetch record calls response does not make any sense.

- the batch fetch (one hashmap lookup) & batch ingest (one hashmap lookup + one flush) are not optimized at all.

Finally, we need more tests.

If we end up deciding that we truely need queue existence & local positions, we can probably add global truncation operation
that lists existing queues, and then truncates... and have both local and global id on each record.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3209,6 +3209,7 @@ version = "0.3.1"
 dependencies = [
  "anyhow",
  "async-trait",
+ "crc32fast",
  "flume",
  "futures",
  "once_cell",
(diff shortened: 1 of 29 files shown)


guilload:
Superseded by #2189, closing.

PR #1902, opened 2022-08-31
Adding a `max_num_partitions` parameter.

fulmicoton:
The `max_num_partitions` is used to apply
a modulo to partition_ids.

### DIFF
--- a/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
+++ b/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
@@ -18,7 +18,7 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::collections::{BTreeMap, BTreeSet};
-use std::str::FromStr;
+use std::num::NonZeroU64;
 
 use anyhow::{bail, Context};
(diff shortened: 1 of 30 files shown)


evanxg852000 (approved):

quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -137,6 +137,11 @@ impl DefaultDocMapper {
         }
         Ok(())
     }
+
+    /// Default maximum number of partitions.
+    pub fn default_max_num_partitions() -> NonZeroU64 {
```
guilload, line 142: I assume we'll advise users to always go with a small number of partitions to prevent indexing issues down the line. In that case, affirming that Quickwit supports partitioning looks pretty misleading. I'd be more comfortable if we called this feature "bucketing" and ship true partitioning later.

guilload: Not that Quickwit is comparable to those engines, but for reference, Hive and Spark start choking when the number of partitions is in the thousands, and the size of partitions is reaching < 1 GB.

Issue #1803, opened 2022-07-27
Refactor garbage collection with the new janitor service coming


fmassot:
## Current situation
Garbage collection is runned on a indexer for each pipeline. The garbage collection of splits is intertwined with the `IndexingSplitStore`. This store is keeping splits locally for merge purposes but currently the removal of these splits is done by the garbage collector.

## Issues
This is bad for several reasons:
- we should do it by index as mentioned in #1572.
- this does not work if we have several indexers.
- `IndexingSplitStore` should delete local splits as soon as splits are published and mature and should not rely on the garbage collector.

## Targeted solution

We want to run garbage collection on a janitor service (#1774, #1773). This implies separating the garbage collection logic from the `IndexingSplitStore` logic.

TODO:
- Implement a garbage collection logic at the indexer level for the `IndexingSplitStore` so that we don't keep useless splits locally.
- Refactor GC to operate on multiple indexes and move to janitor service.
 
IndexingSplitStore refactoring should resolve https://github.com/quickwit-oss/quickwit/issues/855

evanxg852000:
## On refactoring the SplitStore @guilload @fmassot 
We want a global SplitStore per indexing node. Currently, we have a split-store per source. 
We kind of need to decide the indexing directory layout. My idea is that we create a top level SplitStore that will manage all indexes SplitStore. So we get rid of the split-store per source separation, all indexing pipelines for a given index work in the same `cache`  and `scratch`directory.
- Is there any constrain/benefit from what we are doing currently by separating split-store per source? 
- Is it always ok to merge splits from different sources?

```
qwdata/indexing
├── index_a
│   ├── caches
│   │   └── splits
│   └── scratch
└── index_b
    ├── cache
    │   └── splits
    └── scratch
```

guilload:
Let's not merge splits from different sources. After the refactor, the merge key should be `(node_id, index_id, source_id)`.

fmassot:
> Is it always ok to merge splits from different sources (source_id)?

I'm not sure of that right now but this question seems unrelated to the fundamental problem we want to solve: having a global split cache.

> Is there any constrain/benefit from what we are doing currently by separating split-store per source?
The benefit is isolation between indexing pipelines, it's always nice to have different working directories, also for debugging purpose it's convenient.

My personal opinion on this: let's focus on the problem of having a global cache split directory so that we can control its size.
[...]

PR #1903, opened 2022-08-31
Refactor GC into Janitor service

evanxg852000:
Moved Gc into janitor service

Part of https://github.com/quickwit-oss/quickwit/issues/1803

### DIFF
--- a/quickwit-indexing/failpoints/mod.rs
+++ b/quickwit-indexing/failpoints/mod.rs
@@ -47,7 +47,7 @@ use quickwit_indexing::actors::MergeExecutor;
 use quickwit_indexing::merge_policy::MergeOperation;
 use quickwit_indexing::models::{IndexingPipelineId, MergeScratch, ScratchDirectory};
 use quickwit_indexing::{get_tantivy_directory_from_split_bundle, new_split_id, TestSandbox};
-use quickwit_metastore::{SplitMetadata, SplitState};
+use quickwit_metastore::{Split, SplitMetadata, SplitState};
 use tantivy::Directory;
 
 #[tokio::test]
@@ -254,8 +254,8 @@ async fn test_merge_executor_controlled_directory_kill_switch() -> anyhow::Resul
     }
 
     let metastore = test_index_builder.metastore();
-    let splits = metastore.list_all_splits(index_id).await?;
-    let split_metadatas: Vec<SplitMetadata> = split_attrs
+    let splits: Vec<Split> = metastore.list_all_splits(index_id).await?;
+    let split_metadatas: Vec<SplitMetadata> = splits
         .into_iter()
         .map(|split| split.split_metadata)
         .collect();
@@ -265,7 +265,7 @@ async fn test_merge_executor_controlled_directory_kill_switch() -> anyhow::Resul
         merge_scratch_directory.named_temp_child("downloaded-splits-")?;
     let storage = test_index_builder.storage();
     let mut tantivy_dirs: Vec<Box<dyn Directory>> = vec![];
-    for split in &splits {
+    for split in &split_metadatas {
         let split_filename = split_file(split.split_id());
         let dest_filepath = downloaded_splits_directory.path().join(&split_filename);
         storage
@@ -278,7 +278,7 @@ async fn test_merge_executor_controlled_directory_kill_switch() -> anyhow::Resul
     let merge_scratch = MergeScratch {
         merge_operation: MergeOperation {
             merge_split_id: new_split_id(),
-            splits,
+            splits: split_metadatas,
         },
         merge_scratch_directory,
         downloaded_splits_directory,
--- a/quickwit-search/src/search_stream/mod.rs
+++ b/quickwit-search/src/search_stream/mod.rs
@@ -33,11 +33,33 @@ use tantivy::fastfield::FastValue;
 
 use self::collector::PartitionValues;
 
+pub trait ToLittleEndian {
+    fn to_le_bytes(&self) -> [u8; 8];
+}
+
+impl ToLittleEndian for u64 {
+    fn to_le_bytes(&self) -> [u8; 8] {
+        u64::to_le_bytes(*self)
+    }
+}
+
+impl ToLittleEndian for i64 {
+    fn to_le_bytes(&self) -> [u8; 8] {
+        i64::to_le_bytes(*self)
+    }
+}
+
+impl ToLittleEndian for f64 {
+    fn to_le_bytes(&self) -> [u8; 8] {
+        f64::to_le_bytes(*self)
+    }
+}
+
 /// Serialize the values into the `buffer` as bytes.
 ///
 /// Please note that the `buffer` is always cleared.
-pub fn serialize<TFastValue: FastValue + Display>(
-    values: &[TFastValue],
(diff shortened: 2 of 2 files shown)


quickwit-janitor/src/actors/janitor_service.rs
```diff
@@ -17,48 +17,84 @@
...
 }
 
 pub struct JanitorService {
-    _data_dir_path: PathBuf,
-    _metastore: Arc<dyn Metastore>,
-    _storage_resolver: StorageUriResolver,
+    node_id: String,
+    metastore: Arc<dyn Metastore>,
+    storage_resolver: StorageUriResolver,
+    garbage_collector_handle_opt: Option<ActorHandle<GarbageCollector>>,
```
fmassot, line 64: why not starting the GC right from the beginning and avoid this option?

quickwit-janitor/src/actors/janitor_service.rs
```diff
@@ -17,48 +17,84 @@
...
-            _storage_resolver: storage_resolver,
+            node_id,
+            metastore,
+            storage_resolver,
+            garbage_collector_handle_opt: None,
             counters: JanitorServiceCounters::default(),
         }
     }
+
+    async fn spawn_garbage_collector(
```
fmassot, line 83: I would do that right in the new. I don't feel the need for a `SpawnGarbageCollector` message, we can do everything in the new method and we're done.

quickwit-janitor/src/actors/janitor_service.rs
```diff
@@ -73,32 +109,89 @@ impl Actor for JanitorService {
...
+#[async_trait]
+impl Handler<Supervise> for JanitorService {
+    type Reply = ();
+
+    async fn handle(
+        &mut self,
+        _message: Supervise,
+        ctx: &ActorContext<Self>,
+    ) -> Result<(), ActorExitStatus> {
+        if let Some(gc_handle) = &self.garbage_collector_handle_opt {
```
fmassot, line 127: What should we do if the GC actor is unhealthy?
In which cases the GC is unhealthy?

fmassot: As a first PR, I would even remove the supervise stuff and see what we will do the retention policy actors.

quickwit-janitor/src/actors/garbage_collector.rs
```diff
@@ -115,30 +110,71 @@ impl Handler<Loop> for GarbageCollector {
...
-                .map(|deleted_entry| deleted_entry.file_name.as_str())
-                .collect();
-            info!(deleted_files=?deleted_files, "gc-delete");
-
-            self.counters.num_deleted_files += deleted_file_entries.len();
-            self.counters.num_deleted_bytes += deleted_file_entries
-                .iter()
-                .map(|entry| entry.file_size_in_bytes as usize)
-                .sum::<usize>();
+        let index_metadatas = self
```
fmassot, line 113: We could add a method on the metastore to directly get the splits to delete. I would not do it in this PR but rather open an issue to avoid looping all indexes.

fmassot: actually, we don't really want that as we need to support the file backed metastore :/

quickwit-janitor/src/actors/janitor_service.rs
evanxg852000: My reasoning is that when we fail to GC an index this actor will be in the wrong state. 
But we could just log the error and let the service keep running in which case we don't need this supervision.

quickwit-janitor/src/actors/garbage_collector.rs
evanxg852000: There is still the need to fetch the index_metadatas since we need to resolve the storage for each index.

fmassot: Indeed, I always forget about it. So we could directly have a method on the metastore to get the list of splits to delete whatever the index.

quickwit-janitor/src/janitor_service.rs
```diff
@@ -0,0 +1,48 @@
...
+use quickwit_actors::ActorHandle;
+use quickwit_metastore::MetastoreError;
+use quickwit_storage::StorageResolverError;
+use thiserror::Error;
+use tracing::error;
+
+use crate::actors::GarbageCollector;
+
+#[derive(Error, Debug)]
+pub enum JanitorServiceError {
```
fmassot, line 29: Having to redefine these error enum everywhere is a bit annoying, especially when there is the same type of errors. I have no magic solution right now but I put it here just to share my thoughts :)

quickwit-janitor/src/lib.rs
```diff
@@ -19,28 +19,35 @@
...
+) -> anyhow::Result<JanitorService> {
     info!("Starting janitor service.");
-    let janitor_service = JanitorService::new(
-        config.data_dir_path.to_path_buf(),
-        metastore,
-        storage_uri_resolver,
-    );
-    let (janitor_service_mailbox, _) = universe.spawn_actor(janitor_service).spawn();
+    let garbage_collector = GarbageCollector::new(metastore, storage_uri_resolver);
+    let (_, garbage_collector_handle) = universe.spawn_actor(garbage_collector).spawn();
```
fmassot, line 47: I would remove the `node_id` here. We have for the indexing pipeline as we need to put that info in the split metadata.

fmassot (approved):
LGTM.

The retention policy and delete tasks will soon land, and we will need to do some supervision Sorry if my comments made you delete that part but I was not sure about how to handle that t with all the different actors. Let's think about that in the next PRs.

quickwit-janitor/src/janitor_service.rs
guilload: My rule of thumb is: don't create custom errors if you're not going to use them. I'd remove them for now.

quickwit-janitor/src/actors/garbage_collector.rs
```diff
@@ -115,30 +110,71 @@ impl Handler<Loop> for GarbageCollector {
...
+            .metastore
+            .list_indexes_metadatas()
+            .await
+            .context("Failed to list indexes.")?;
```
guilload, line 117: Since the GC actor is running unsupervised, this will crash and stop the GC on the janitor service.  Let's properly log the error and return early.

The GC actor is a glorified for-loop, so I agree with the unsupervised strategy but let's be more resilient to errors.

quickwit-janitor/src/actors/garbage_collector.rs
```diff
@@ -115,30 +110,71 @@ impl Handler<Loop> for GarbageCollector {
...
+                    .context(format!(
+                        "Failed to resolve the index storage Uri: `{0}`.",
+                        index_metadata.index_uri
+                    ))?;
+                Result::<(String, Arc<dyn Storage>), anyhow::Error>::Ok((
+                    index_metadata.index_id,
+                    storage,
+                ))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
guilload, line 135: Ditto. Let's skip this index and try again later.

guilload (commented):
Please, address my comments in a follow-up PR.

PR #1907, opened 2022-09-01
Factorizing common split metadata into a `SplitInfo` type.

fulmicoton:


guilload (approved):
My personal preference is `SplitAttrs`, but LGTM.

fulmicoton:
@guilload Ok for SplitAttrs. By the way this is in preparation to a following PR where all these attributes will be duplicated again.

PR #1916, opened 2022-09-04
First stone of the delete query feature.

fmassot:
### Description

This PR is adding a delete query feature.

This is only a first step as it needs more testing.

There is 2 parts in this PR.

1. New metastore API for delete with some tests, I did not think too much about tests here, I will probably add a few more there.
2. Delete actors and pipeline
    - Update `MergeExectutor` to process deletes with one test to prove it runs.
    - New `DeleteTaskPlanner` with one test to prove it runs.
    - New `DeleteTaskService` with one test to prove it runs.
    - New `DeleteTaskPipeline` with one test to prove it runs.

I can split the PR in 2 parts if it's too long.

## Important notes
1. The `DeleteTaskPipeline` is not resilient to faulty actors. The goal in a next PR would be to restart faulty actors when they are failing.

2. Not major but still worth mentioning it. The delete task planner does not know about the ongoing delete operations, this can be annoying for observability reasons. This could be solved by using a weak reference on the merge operation throughout the pipelines. It seems worth investigating that solution.


quickwit-indexing/src/actors/merge_split_downloader.rs
```diff
@@ -25,18 +25,17 @@ use quickwit_metastore::SplitMetadata;
 use tantivy::Directory;
 use tracing::{info, info_span, warn, Span};
 
-use crate::actors::MergeExecutor;
 use crate::merge_policy::MergeOperation;
 use crate::models::{MergeScratch, ScratchDirectory};
 use crate::split_store::IndexingSplitStore;
 
-pub struct MergeSplitDownloader {
+pub struct MergeSplitDownloader<A: Actor> {
```
fulmicoton, line 32: what is the generics about?

quickwit-indexing/src/models/indexed_split.rs
```diff
@@ -86,6 +86,7 @@ impl IndexedSplit {
             uncompressed_docs_size_in_bytes: 0,
             time_range: None,
             replaced_split_ids: Vec::new(),
+            delete_opstamp: 0, // TODO: fetch the last delete task opstamp to set it.
```
fulmicoton, line 89: is it something we can either fix before merge or open an issue?

quickwit-indexing/src/split_store/indexing_split_store.rs
```diff
@@ -197,26 +197,29 @@ impl IndexingSplitStore {
     /// The output_path is expected to be a directory path.
     pub async fn fetch_split(
         &self,
-        split_id: &str,
+        split: &SplitMetadata,
```
fulmicoton, line 200: I don't see any other use than taking the split_id.

quickwit-proto/proto/metastore_api.proto
```diff
@@ -23,44 +23,62 @@ syntax = "proto3";
...
+  rpc last_delete_opstamps_by_index_id(LastDeleteOpstampsByIndexIdRequest) returns (LastDeleteOpstampsByIndexIdResponse);
+
+  // Creates a delete task.
+  rpc create_delete_task(DeleteQueryRequest) returns (DeleteTaskResponse);
+
+  // Deletes all delete tasks for a given `index_id`.
+  rpc delete_delete_tasks(DeleteDeleteTasksRequest) returns (DeleteTasksResponse);
+
+  // Updates splits `delete_opstamp`.
+  rpc update_splits_delete_opstamp(UpdateSplitsDeleteOpstampRequest) returns (UpdateSplitsDeleteOpstampResponse);
```
fulmicoton, line 75: can you confirm that this is used when there are no deleted docs, and that the other  path works using deletes.

quickwit-janitor/src/actors/delete_task_planner.rs
```diff
@@ -0,0 +1,545 @@
...
+            .iter()
+            .position(|operation| operation.merge_split_id == merged_split.split_id)
+            .ok_or_else(|| {
+                anyhow!(
+                    "The published split after the delete operation must have its corresponding \
+                     operation in `ongoing_delete_operations`."
+                )
+            })?;
+        self.ongoing_delete_operations
+            .remove(operation_idx_to_remove);
```
fulmicoton, line 386:
```suggestion
            .swap_remove(operation_idx_to_remove);
```

quickwit-janitor/src/actors/delete_task_planner.rs
```diff
@@ -0,0 +1,545 @@
...
+        Ok(())
+    }
+}
+
+#[derive(Debug)]
+struct PlanDeleteOperations;
```
fulmicoton, line 392: Can we pool loop in the message name?

quickwit-janitor/src/actors/delete_task_planner.rs
```diff
@@ -0,0 +1,545 @@
...
+#[async_trait]
+impl Handler<PlanDeleteOperations> for DeleteTaskPlanner {
+    type Reply = ();
+
+    async fn handle(
+        &mut self,
+        _: PlanDeleteOperations,
+        ctx: &ActorContext<Self>,
+    ) -> Result<(), ActorExitStatus> {
+        let index_id = self.pipeline_id.index_id.clone();
```
fulmicoton, line 403: useless clone?

quickwit-janitor/src/actors/delete_task_planner.rs
```diff
@@ -0,0 +1,545 @@
...
+
+    /// Send delete operations for a given `index_id`.
+    async fn send_delete_operations(
+        &mut self,
+        index_id: &str,
+        ctx: &ActorContext<Self>,
+    ) -> anyhow::Result<()> {
+        if self.ongoing_delete_operations.len() > NUM_STALE_SPLITS_TO_FETCH {
+            info!(
+                "Abort sending delete operations, there is already `{}` ongoing delete operations.",
```
fulmicoton, line 135:
```suggestion
                "Abort sending delete operations, there are already `{}` ongoing delete operations.",
```

quickwit-indexing/src/split_store/indexing_split_store.rs
fmassot: Mmmm, sorry for this noise :/
Seems I don't know why I did that...

quickwit-proto/proto/metastore_api.proto
fmassot: Do you mean confirm by some tests?

quickwit-indexing/src/actors/merge_split_downloader.rs
fmassot: Good point. 

I changed the name "A" for `ExecutorActor` as what I want is to reuse the `MergeSplitDownloader` but send the download split to the `DeleteTaskExecutor`.

quickwit-indexing/src/models/indexed_split.rs
fmassot: We can yes, this is not a hard pb, I'm just annoyed by the fact that I will probably need to add the metastore to the `IndexerState` in order to fetch the task.

quickwit-indexing/src/test_utils.rs
```diff
@@ -65,6 +65,7 @@ impl TestSandbox {
         doc_mapping_yaml: &str,
         indexing_settings_yaml: &str,
         search_fields: &[&str],
+        metastore_uri: Option<&str>,
```
fmassot, line 68: This is useful as a shared metastore URI can break some tests.

quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index.rs
```diff
@@ -387,8 +400,93 @@ impl FileBackedIndex {
         Ok(true)
     }
 
-    /// Resets the checkpoint of a source. Returns whether a mutation occurred.
```
fmassot, line 390: This is the only function with the docs, either we put that everywhere or we don't.

guilload: lol, put it everywhere then, rather than removing this line. How is removing some relevant docs a good thing?

fmassot: Yes.... my comment seems strange but... the doc is already in the `mod.rs` file. We can choose to repeat the docs everywhere. I don't know what's the good practice when you are just implementing a trait.

quickwit-indexing/src/actors/indexer.rs
```diff
@@ -172,19 +175,25 @@ impl IndexerState {
...
                 Ok(vacant_entry.insert(indexed_split))
             }
         }
     }
 
-    fn create_workbench(&self) -> anyhow::Result<IndexingWorkbench> {
+    async fn create_workbench(&self) -> anyhow::Result<IndexingWorkbench> {
+        let last_delete_opstamp = self
+            .metastore
+            .last_delete_opstamp(&self.pipeline_id.index_id)
```
fulmicoton, line 195: That's one extra call every 30s per index.
Can you open an issue tracking the fact that we might want to remove this at one point?

quickwit-indexing/src/actors/merge_executor.rs
```diff
@@ -135,7 +155,7 @@ fn open_split_directories(
 }
 
 /// Creates a directory with a single `meta.json` file describe in `index_meta`
-fn create_shadowing_meta_json_directory(index_meta: IndexMeta) -> anyhow::Result<RamDirectory> {
+pub fn create_shadowing_meta_json_directory(index_meta: IndexMeta) -> anyhow::Result<RamDirectory> {
```
fulmicoton, line 158:
```suggestion
fn create_shadowing_meta_json_directory(index_meta: IndexMeta) -> anyhow::Result<RamDirectory> {
```

quickwit-indexing/src/actors/mod.rs
```diff
@@ -40,11 +41,13 @@ pub use self::indexer::{Indexer, IndexerCounters};
 pub use self::ingest_api_garbage_collector::{
     IngestApiGarbageCollector, IngestApiGarbageCollectorCounters,
 };
-pub use self::merge_executor::{combine_partition_ids, MergeExecutor};
+pub use self::merge_executor::{
```
fulmicoton, line 44: no need to expose this function.

quickwit-indexing/src/actors/merge_executor.rs
```diff
@@ -231,10 +251,75 @@ fn merge_split_directories(
...
 }
 
+/// Applies delete tasks on the directory stack, commit, and merge to rebuild the split
+/// without the deleted documents.
+/// The directory stack is made of:
+/// - The output directory that is a `MmapDirectory` on `output_path`. This directory receives all
+///   writes.
+/// - The current split directory which is expected to be a `BundleDirectory` as the split has just
+///   been downloaded. This `BundleDirectory` is read-only.
+fn apply_deletes_and_merge_split(
```
fulmicoton, line 261: Could this be the same function for merges? (With an empty vec for delete_tasks)

quickwit-janitor/src/actors/delete_task_service.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub num_running_pipelines: usize,
+}
+
+pub struct DeleteTaskService {
+    metastore: Arc<dyn Metastore>,
+    search_client_pool: SearchClientPool,
+    storage_resolver: StorageUriResolver,
+    data_dir_path: PathBuf,
+    pipeline_handles_by_index_id: HashMap<String, ActorHandle<DeleteTaskPipeline>>,
+    state: DeleteTaskServiceState,
```
fulmicoton, line 47:
```suggestion
```

Please remove the state from `DeleteTaskService` and compute it on the fly on calls to
`observalbe_state`.

Right now the invariant `num_running_pipelines == self.pipeline_handles_by_index_Id` is needlessly difficult to proofread.

quickwit-janitor/src/actors/delete_task_service.rs
```diff
@@ -0,0 +1,230 @@
...
+            info!(
+                deleted_index_id = deleted_index_id,
+                "Remove deleted index from delete task pipelines."
+            );
+            let pipeline_handle = self
+                .pipeline_handles_by_index_id
+                .remove(deleted_index_id)
+                .expect("Handle must be present.");
+            pipeline_handle.quit().await;
+            self.state.num_running_pipelines -= 1;
```
fulmicoton, line 113:
```suggestion
```

quickwit-janitor/src/actors/delete_task_service.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        // Start new pipelines and add them to the handles hashmap.
+        for index_id in index_ids.difference(&pipeline_index_ids) {
+            let index_metadata = index_metadata_by_index_id
+                .remove(index_id)
+                .expect("Index metadata must be present.");
+            if self.spawn_pipeline(index_metadata, ctx).await.is_ok() {
+                self.state.num_running_pipelines += 1;
+            }
```
fulmicoton, line 123:
```suggestion
            if self.spawn_pipeline(index_metadata, ctx).await.is_err() {
                warn!("Failed to spawn delete pipeline for {index_id}");
            }
```

quickwit-metastore/src/metastore/delete_task.rs
```diff
@@ -0,0 +1,95 @@
...
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use quickwit_proto::metastore_api::{DeleteQueryRequest, DeleteTaskResponse};
+use serde::{Deserialize, Serialize};
+
+/// A delete query.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct DeleteQuery {
+    /// Index ID.
+    pub index_id: String,
```
fulmicoton, line 27: If the types are really the same (this is the case here), you are better off using the protobuf directly.

quickwit-janitor/src/lib.rs
```diff
@@ -34,17 +36,33 @@ pub use janitor_service::JanitorService;
...
+    config: &QuickwitConfig, // kept it for retention policy
     metastore: Arc<dyn Metastore>,
+    cluster: Arc<Cluster>,
     storage_uri_resolver: StorageUriResolver,
 ) -> anyhow::Result<JanitorService> {
     info!("Starting janitor service.");
-    let garbage_collector = GarbageCollector::new(metastore, storage_uri_resolver);
+    let garbage_collector = GarbageCollector::new(metastore.clone(), storage_uri_resolver.clone());
     let (_, garbage_collector_handle) = universe.spawn_actor(garbage_collector).spawn();
+    let search_client_pool = SearchClientPool::create_and_keep_updated(
```
fulmicoton, line 51: We need to use the same pool as search etc.
Can you open an issue to make that search pool into some kind of singleton?

(a service perhaps)

quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index.rs
```diff
@@ -387,8 +400,93 @@ impl FileBackedIndex {
...
+        self.delete_tasks
+            .iter()
+            .map(|delete_task| delete_task.opstamp)
+            .max()
+            .unwrap_or(0)
+    }
+
+    pub(crate) fn delete_delete_tasks(&mut self) -> MetastoreResult<bool> {
+        self.delete_tasks.clear();
+        Ok(true)
```
fulmicoton, line 431: why return true?

quickwit-metastore/src/metastore/mod.rs
```diff
@@ -188,4 +211,38 @@ pub trait Metastore: Send + Sync + 'static {
...
     fn uri(&self) -> &Uri;
+
+    /// Gets the last delete opstamp for a given `index_id`.
+    async fn last_delete_opstamp(&self, index_id: &str) -> MetastoreResult<u64>;
+
+    /// Creates a [`DeleteTask`] from a [`DeleteQuery`].
+    async fn create_delete_task(&self, delete_query: DeleteQuery) -> MetastoreResult<DeleteTask>;
+
+    /// Deletes all [`DeleteTask`] for a given `index_id`.
+    async fn delete_delete_tasks(&self, index_id: &str) -> MetastoreResult<()>;
```
fulmicoton, line 222: why do we have this call? It is not used as far as I can tell.

I don't think this is a good idea.

fulmicoton: (also the name is silly)

quickwit-metastore/migrations/postgresql/4_create-delete_tasks.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS delete_tasks (
+    create_timestamp TIMESTAMP NOT NULL DEFAULT (CURRENT_TIMESTAMP AT TIME ZONE 'UTC'),
+    opstamp BIGSERIAL PRIMARY KEY,
+    index_id VARCHAR(50) NOT NULL,
+    delete_query_json TEXT NOT NULL,
+
+    FOREIGN KEY(index_id) REFERENCES indexes(index_id)
```
fulmicoton, line 7: I think you need to fix `delete_index`.... and possibly have a look at other calls of the metastore I might have not think of.

Right now in presence of delete ops the foreign key will error.
You probably want to delete all delete task in delete_index. 

A unit test would be awesome.

quickwit-metastore/src/metastore/mod.rs
fmassot: Agreed, I used it only for testing so it's useless.

quickwit-metastore/migrations/postgresql/4_create-delete_tasks.up.sql
fmassot: Yes indeed.

quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index.rs
```diff
@@ -387,8 +400,93 @@ impl FileBackedIndex {
...
+            .collect_vec();
+        Ok(delete_tasks)
+    }
+}
+
+/// Stamper provides Opstamps, which is just an auto-increment id to label
+/// a delete operation.
+/// Copy-pasted from tantivy with an `AtomicUsize` to have a broader support.
+#[derive(Clone, Default)]
+pub struct Stamper(Arc<AtomicUsize>);
```
fulmicoton, line 466: Interior mutability wiht an `Arc<AtomicUsize>` looks like a good idea but it is not here.

That stamper is owned by FileBackedIndex which is modified using a strange pattern to mimick the concept of transaction.
We acqurie a lock, mutate it, and if everything goes right, we store the modfiied copy.

By using this Arc<AtomicUsize> you are idea the fact that the FileBackedIndex ends up not being modified if a transaction fails. Just an usize would have been nicer.

quickwit-metastore/src/metastore/delete_task.rs
fmassot: I was hesitating on this actually. Protobuf is always putting an option so it's a bit annoying but the extra code for just removing that option is too big :/

quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index.rs
```diff
@@ -387,8 +400,93 @@ impl FileBackedIndex {
...
+    pub fn stamp(&self) -> usize {
+        (self.0.fetch_add(1usize, Ordering::SeqCst) + 1) as usize
+    }
+
+    /// Reverts the stamper to a given `Opstamp` value and returns it
+    pub fn revert(&self, to_opstamp: usize) -> usize {
+        self.0.store(to_opstamp, Ordering::SeqCst);
+        to_opstamp
+    }
+}
```
fulmicoton, line 484:
```suggestion
```

This function is not used in the code (and is error prone)

quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index.rs
fulmicoton, line 466:
```suggestion
pub struct Stamper(usize);
```

fulmicoton, line 466:
```suggestion
struct Stamper(Arc<AtomicUsize>);
```

fmassot: oh yes! it crossed my mind a few days ago and totally forget it... actually I feel my PR is bit too big for my brain :/

quickwit-proto/proto/metastore_api.proto
fulmicoton: ah no I just meant explaining me... but I think I saw that in the code.

quickwit-janitor/src/actors/delete_task_service.rs
fmassot: yes indeed, I should have done it :/

quickwit-metastore/src/tests.rs
```diff
@@ -2170,6 +2172,397 @@ pub mod test_suite {
...
+                split_metadata_2.delete_opstamp
+            );
+
+            let splits = metastore.list_stale_splits(index_id, 10, 2).await.unwrap();
+            assert!(splits.is_empty());
+            cleanup_index(&metastore, index_id).await;
+        }
+    }
+
+    pub async fn test_metastore_update_splits_delete_opstamp<
```
fulmicoton, line 2469: please add a unit test to see what happens when  we delete an index with delete tasks.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -732,6 +734,187 @@ impl Metastore for PostgresqlMetastore {
...
+                create_timestamp: create_timestamp.assume_utc().unix_timestamp(),
+                opstamp: opstamp as u64,
+                delete_query,
+            })
+        })
+    }
+
+    /// Deletes delete tasks of a given index.
+    #[instrument(skip(self))]
+    async fn delete_delete_tasks(&self, index_id: &str) -> MetastoreResult<()> {
```
fulmicoton, line 797: i recommend removing this call.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -732,6 +734,187 @@ impl Metastore for PostgresqlMetastore {
...
+        }
+        run_with_tx!(self.connection_pool, tx, {
+            let updated_split_ids: Vec<String> = sqlx::query(
+                r#"
+                UPDATE splits
+                SET delete_opstamp = $1
+                WHERE
+                    index_id = $2
+                    AND split_id = ANY($3)
+                    AND split_state = $4
```
fulmicoton, line 826:
```suggestion
                    AND split_state = $4
```

Why should we check for the state?

quickwit-metastore/migrations/postgresql/4_create-delete_tasks.up.sql
fmassot: I will use `ON DELETE CASCADE`.

fmassot: We could use `ON DELETE CASCADE`.

quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -732,6 +734,187 @@ impl Metastore for PostgresqlMetastore {
...
+            .bind(delete_opstamp as i64)
+            .bind(index_id)
+            .bind(split_ids)
+            .bind(SplitState::Published.as_str())
+            .map(|row| row.get(0))
+            .fetch_all(&mut *tx)
+            .await?;
+            // If no splits were returned, maybe the index itself does not exist
+            // in the first place.
+            if updated_split_ids.is_empty() && index_opt(tx, index_id).await?.is_none() {
```
fulmicoton, line 839: nitpick....

Using RETURNING and fetch the split_ids was overkill here.
You only use their count.

The following function gives you the number of modified rows

`sqlx_result.rows_affected()`
https://docs.rs/sqlx/latest/sqlx/postgres/struct.PgQueryResult.html#method.rows_affected

quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index.rs
```diff
@@ -387,8 +400,93 @@ impl FileBackedIndex {
...
+pub struct Stamper(Arc<AtomicUsize>);
+
+impl Stamper {
+    /// Creates a [`Stamper`].
+    pub fn new(first_opstamp: usize) -> Self {
+        Self(Arc::new(AtomicUsize::new(first_opstamp)))
+    }
+
+    /// Increments the stamper by 1 and returns the incremented value.
+    pub fn stamp(&self) -> usize {
```
fulmicoton, line 475: nitpick: 
It is actually that a delete opstamp cannot be zero.
(indeed a brand new split gets 0 by default).
Maybe working with a NonZeroUsize here, and in some other places WHERE it is not annoying.

quickwit-metastore/src/metastore/postgresql_metastore.rs
fmassot: You're right yes, I should have removed this condition, when I first coded it I wanted to make sure to do that only on published splits because that's what we should do. But the responsibility of this is on the caller anyway. Ideally we would want to only update these stamps only on mature splits but this condition can only be check outside the metastore.

quickwit-indexing/src/actors/indexer.rs
fmassot: I did not think about it when I coded that but... we could fetch the last_delete_opstamp in the metastore directly. 
Pros are that we don't make extra call, the code is simpler on the indexer side.
Cons is that we don't have the flexibility on choosing when to fetch the last delete opstamp.

I like the solution to put that on the metastore.

quickwit-indexing/src/actors/merge_executor.rs
fmassot: Indeed, done.

quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index.rs
fmassot: I'm opening an issue on this.

quickwit-metastore/src/metastore/postgresql_metastore.rs
fmassot: Done.

quickwit-metastore/src/metastore/postgresql_metastore.rs
fmassot: Done.

quickwit-metastore/src/tests.rs
fmassot: Done: `test_metastore_delete_index_with_tasks`.

quickwit-janitor/src/lib.rs
fmassot: Actually this is very simple to do, I have just fixed it in the PR.

quickwit-metastore/src/metastore/delete_task.rs
fmassot: I switched to protobuf.

quickwit-metastore/migrations/postgresql/4_create-delete_tasks.up.sql
fmassot: And there is a test.

quickwit-metastore/migrations/postgresql/4_create-delete_tasks.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS delete_tasks (
+    create_timestamp TIMESTAMP NOT NULL DEFAULT (CURRENT_TIMESTAMP AT TIME ZONE 'UTC'),
+    opstamp BIGSERIAL PRIMARY KEY,
+    index_id VARCHAR(50) NOT NULL,
+    delete_query_json TEXT NOT NULL,
+
+    FOREIGN KEY(index_id) REFERENCES indexes(index_id) ON DELETE CASCADE
```
fulmicoton, line 7: ON DELETE CASCADE is nifty..

fulmicoton (approved):

PR #1918, opened 2022-09-05
Refactor Quickwit directory and CI

boraarslan:
Changes:
- Refactor root directory
   1. Moved rust code into the `quickwit` folder. (As a side effect, I've created another Makefile and moved cargo commands into that one. You can still use cargo-related Makefile commands in the root directory, but it calls the inner Makefile)
   2. Moved rust-related config into the `quickwit` folder. (Cross, nextest, rustfmt etc...)
   3. Moved scripts into the `quickwit` folder.

- Refactor CI
   1. Seperated cargo commands into multiple jobs.
   4. Moved UI CI to a different file.
   5. Set `mold` as the default linker.
   6. Set `cargo nextest` for tests.
   7. Only cache compiling cargo commands
   8. Only trigger rust CI when rust code changes.
   9. Only trigger UI CI when UI code changes.
 
- Rust Code changes
   Jemalloc is now a feature, only included on release binaries.
      *-sys crates can't be cached by sccache or rust-cache (probably because they use custom build scripts), so most of the compile time comes from compiling `jemalloc` and `rocksdb`. By making jemalloc a feature, we entirely skip compiling it on CI (unless you pass `--all-features` to cargo), which, together with caching, nearly cuts CI time in half.


guilload:
Nice work in progress.

.github/workflows/ui-ci.yml
```diff
@@ -0,0 +1,72 @@
...
+          - name: Lint
+            command: yarn --cwd quickwit-ui lint
+          - name: Unit Test
+            command: yarn --cwd quickwit-ui test
+          - name: Cypress run
+            command: |
+              CI=false yarn --cwd quickwit-ui build
+              cargo build --features=postgres
+              mkdir qwdata
+              target/debug/quickwit run --service searcher --service metastore --config ../config/quickwit.yaml &
```
guilload, line 31:
```
cargo run --features=postgres -- run --service searcher --service metastore --config ../config/quickwit.yaml
```

Makefile
```diff
@@ -1,5 +1,7 @@
 DOCKER_SERVICES ?= all
 
+CODE_DIR = quickwit
```
guilload, line 3:
```suggestion
QUICKWIT_SRC = quickwit
```

guilload (approved):
Approved! Awesome work!

Issue #1265, opened 2022-04-07
Add retention policy 


fulmicoton:
We want to let users define a retention policy associated to an index.

The retention policy should periodically delete splits that go out of retention (as part of the GC actor or new specific one).

Related #403

guilload:
A retention policy defines a period of time after which splits are deleted.

## Configuration

### Examples of retention in various systems

Setting a retention policy can be fairly simple or complex depending on the system. For _Amazon CloudWatch_ or _InfluxDB_, it just boils down to defining a retention period with one line of configuration. More complex backends fold retention policies into a more generic rule system that configures actions or transitions that are conditionally applied depending on the properties of the managed objects. Those actions can be moving data between storage tiers, downsampling (rollup), or deleting data. _OpenSearch_ even features a rule engine that is able to send notifications when a rule is applied!

#### Amazon CloudWatch (terraform)
```terraform
resource "aws_cloudwatch_log_group" "yada" {
  name = "Yada"
  retention_in_days = 14
}
```
[Reference](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_log_group)

#### Amazon S3 (terraform) 
```terraform
resource "aws_s3control_bucket_lifecycle_configuration" "example" {
  bucket = aws_s3control_bucket.example.arn

  rule {
    expiration {
      days = 365
    }

    filter {
      prefix = "logs/"
    }

    id = "logs"
  }

  rule {
    expiration {
      days = 7
    }

    filter {
      prefix = "temp/"
    }

    id = "temp"
  }
}
```
[Reference
](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3control_bucket_lifecycle_configuration)

#### InfluxDB
```sql
CREATE RETENTION POLICY "a_year" ON "food_data" DURATION 52w REPLICATION 1;
```
[Reference](https://docs.influxdata.com/influxdb/v1.7/guides/downsampling_and_retention/)

### OpenSearch
```json
{
  "policy": {
    "description": "hot warm delete workflow",
    "default_state": "hot",
    "schema_version": 1,
    "states": [
      {
        "name": "hot",
        "actions": [
          {
            "rollover": {
              "min_index_age": "1d"
            }
          }
        ],
        "transitions": [
          {
            "state_name": "warm"
          }
        ]
      },
      {
        "name": "warm",
        "actions": [
          {
            "replica_count": {
              "number_of_replicas": 5
            }
          }
[...]

PR #1919, opened 2022-09-05
Added publish_timestamp field to split

evanxg852000:
Added `publish_timestamp`  field to split model for retention policy logic.

Part of https://github.com/quickwit-oss/quickwit/issues/1265


quickwit-metastore/migrations/postgresql/3_add-split-publish-timestamp-field.up.sql
```diff
@@ -0,0 +1,20 @@
+ALTER TABLE splits
+    ADD COLUMN publish_timestamp TIMESTAMP DEFAULT NULL;
+
+-- We want to update the split `publish_timestamp` field whenever the split
+-- being is published.
+CREATE OR REPLACE FUNCTION set_split_publish_timestamp_for_split() RETURNS trigger AS $$
+BEGIN
+    IF (TG_OP = 'UPDATE') AND (NEW.split_state = 'Published') THEN
```
guilload, line 8: What happens if the split is already published? Shall we add `AND (OLD.split_state != 'Staged')`?

quickwit-metastore/src/metastore/postgresql_model.rs
```diff
@@ -80,6 +80,8 @@ pub struct Split {
     pub create_timestamp: sqlx::types::time::PrimitiveDateTime,
     /// Timestamp for tracking when the split was last updated.
     pub update_timestamp: sqlx::types::time::PrimitiveDateTime,
+    /// Timestamp for tracking when the split was last updated.
```
guilload, line 83:
```suggestion
    /// Timestamp for tracking when the split was published.
```

quickwit-metastore/src/tests.rs
```diff
@@ -1996,6 +1996,7 @@ pub mod test_suite {
             .unwrap();
```
guilload, line 1996: Let's add a test that checks that the publish timestamp is not updated when a split is re-published.

guilload (approved):

Issue #1915, opened 2022-09-03
Add `index_id` in split metadata


fmassot:
The `Split` and `SplitMetadata` structs do not contain the `index_id`.
It is a bit strange as we store the `index_id` in postgresql splits table.

But we did not feel need for it as the `index_id` is already present in actors via the `IndexPipelineId`. This works as all our actors operate on a given index or pipeline.

But this will not be the case for the delete pipeline where we can have one pipeline working on several indexes.

fulmicoton:
@fmassot +1 for adding index_id to the split metadata

PR #1925, opened 2022-09-06
Add index_id field to SplitMetadata

evanxg852000:
Add `index_id` field to SplitMetadata

Closes https://github.com/quickwit-oss/quickwit/issues/1915


quickwit-metastore/src/split_metadata_version.rs
```diff
@@ -96,6 +97,10 @@ pub(crate) struct SplitMetadataV1 {
     /// in the storage URI resolver: for instance, the Amazon S3 region.
     pub split_id: String,
 
+    /// Id of the index this split belongs to.
+    #[serde(default)]
+    pub index_id: String,
```
fmassot, line 102: 1. We could put "unknown" instead.
2. All the old split metadata deserialized from postgresql or the filebacked metastore will have "unknown" index id. This is a big issue.

One solution could be to override this value when we load the index in the file backed index and when we build the split in the postgresql metastore.

guilload: > One solution could be to override this value when we load the index in the file backed index and when we build the split in the postgresql metastore.

Yes.

quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index.rs
```diff
@@ -102,7 +102,17 @@ impl From<FileBackedIndex> for FileBackedIndexV0 {
 
 impl From<FileBackedIndexV0> for FileBackedIndex {
     fn from(index: FileBackedIndexV0) -> Self {
-        Self::new(index.metadata, index.splits)
+        // Override split index_id to support old SplitMetadata format.
+        let index_id = index.metadata.index_id.clone();
```
guilload, line 106: Can we do this in place to avoid some allocations? Can we do this only when `index_id` is empty?

quickwit-metastore/src/split_metadata_version.rs
```diff
@@ -96,6 +97,10 @@ pub(crate) struct SplitMetadataV1 {
     /// in the storage URI resolver: for instance, the Amazon S3 region.
     pub split_id: String,
 
+    /// Id of the index this split belongs to.
+    #[serde(default)]
+    pub index_id: Option<String>,
```
guilload, line 102:
```suggestion
    pub index_id: String,
```

quickwit-metastore/src/split_metadata_version.rs
```diff
@@ -138,6 +143,7 @@ pub(crate) struct SplitMetadataV1 {
 impl From<SplitMetadataV1> for SplitMetadata {
     fn from(v1: SplitMetadataV1) -> Self {
         let source_id = v1.source_id.unwrap_or_else(|| "unknown".to_string());
```
guilload, line 145: This should have been `_unknown` for the `source_id`.

quickwit-metastore/src/split_metadata_version.rs
```diff
@@ -138,6 +143,7 @@ pub(crate) struct SplitMetadataV1 {
...
+        let index_id = v1.index_id.unwrap_or_else(|| "unknown".to_string());
```
guilload, line 146: Remove this line and use `index_id: v1.index_id` below. That way, we can test for emptiness when overriding the index ID.

quickwit-metastore/src/split_metadata_version.rs
evanxg852000: Maybe `_quickwit__unknown` on both node_id ?

guilload: Yes, sounds good.

guilload (approved):

PR #1929, opened 2022-09-07
Added logging to Gc actor.

evanxg852000:
Added logging to gc actor in janitor service since this actor not be supervised


fmassot (approved):
Sounds good. I would add test on a failing resolver/GC to check if everything is working well.

PR #1930, opened 2022-09-07
Fix Gc reschedule bug

evanxg852000:
Fixed garbage collector reschedule bug when there is an error on listing indexes.

### DIFF
--- a/quickwit-janitor/src/actors/garbage_collector.rs
+++ b/quickwit-janitor/src/actors/garbage_collector.rs
@@ -119,6 +119,7 @@ impl Handler<Loop> for GarbageCollector {
             Ok(metadatas) => metadatas,
             Err(error) => {
                 error!(error=?error, "Failed to list indexes from the metastore.");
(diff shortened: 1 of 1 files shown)


guilload (approved):

guilload:
The following pattern would guarantee that the actor will never crash or stop:

```rust
/// Applies the GC logic. This method must not return an error to prevent the actor from crashing.
async fn handle_inner(&mut self, ctx: &ActorContext<Self>) {
  
}

async fn handle(
        &mut self,
        _: Loop,
        ctx: &ActorContext<Self>,
    ) -> Result<(), quickwit_actors::ActorExitStatus> {
    self.handle_inner(ctx).await;
    ctx.schedule_self_msg(RUN_INTERVAL, Loop).await;
    Ok(())
}
```

evanxg852000:
@guilload Indeed this pattern is far better :+1: , I had already activated auto-merge will apply this to both gc and retention in next PR

PR #1931, opened 2022-09-08
Retention policy 

evanxg852000:
Implements the retention policy logic

Closes https://github.com/quickwit-oss/quickwit/issues/1265


quickwit-config/src/index_config.rs
```diff
@@ -320,6 +321,18 @@ impl RetentionPolicy {
         })
     }
 
+    pub fn duration_till_next_evaluation(&self) -> anyhow::Result<Duration> {
```
guilload, line 324:
```suggestion
    pub fn duration_until_next_evaluation(&self) -> anyhow::Result<Duration> {
```

quickwit-config/src/index_config.rs
```diff
@@ -320,6 +321,18 @@ impl RetentionPolicy {
...
+        let schedule = self.evaluation_schedule()?;
+        let future_date = schedule
+            .upcoming(Utc)
+            .next()
+            .context("Failed to obtain next evaluation date.")?;
```
guilload, line 329: Having to deal with an option is an artifact of the iterator interface. Can a cron expression be "finite"? Shouldn't this be an `expect`?

quickwit-config/src/index_config.rs
guilload, line 324: Please add a unit test for this function.

quickwit-janitor/src/actors/retention_policy_evaluator.rs
```diff
@@ -0,0 +1,460 @@
...
+            metastore,
+            index_metadatas: HashMap::new(),
+            counters: RetentionPolicyEvaluatorCounters::default(),
+        }
+    }
+
+    /// Refresh Loop handler logic.
+    /// Should not return an error to prevent the actor from crashing.
+    async fn handle_refresh_loop(&mut self, ctx: &ActorContext<Self>) {
+        info!("retention-policy-refresh-indexes-operation");
```
guilload, line 73:
```suggestion
        debug!("retention-policy-refresh-indexes-operation");
```

guilload: Use `info` sparingly, our logs tend to be very verbose and noisy.

quickwit-janitor/src/actors/retention_policy_evaluator.rs
```diff
@@ -0,0 +1,460 @@
...
+                *value = index_metadata;
+                continue;
+            }
+
+            let message = Evaluate {
+                index_id: index_metadata.index_id.clone(),
+            };
+            let next_interval = retention_policy
+                .duration_till_next_evaluation()
+                .expect("Expected a valid duration from retention policy schedule.");
```
guilload, line 119: This is going to panic and crash the unsupervised actor. Log the error and skip the index.

quickwit-janitor/src/actors/retention_policy_evaluator.rs
```diff
@@ -0,0 +1,460 @@
...
+#[derive(Debug)]
+struct Loop;
+
+#[derive(Debug)]
+struct Evaluate {
+    index_id: String,
+}
+
+/// An actor for collecting garbage periodically from an index.
+pub struct RetentionPolicyEvaluator {
```
guilload, line 55: I don't think `Evaluator` conveys the fact that the actor also performs a side-effect, i.e. mark some splits for deletion. I'd go with a more explicit name such `RetentionPolicyEnforcer` or `RetentionPolicyExecutor`.

quickwit-janitor/src/actors/retention_policy_evaluator.rs
```diff
@@ -0,0 +1,460 @@
...
+            self.index_metadatas.keys(),
+            index_metadatas.iter().map(|metadata| &metadata.index_id),
+        );
+        for index_id in &deleted_indexes {
+            self.index_metadatas.remove(index_id);
+        }
+        info!(index_ids=%deleted_indexes.iter().join(", "), "Deleted indexes from cache.");
+
+        for index_metadata in index_metadatas.into_iter() {
+            ctx.record_progress();
```
guilload, line 95: Feels like this `record_progress` is useless. Everything in the loop happens in memory without long compute-intensive ops or IOs.

quickwit-janitor/src/retention_policy_evaluation.rs
```diff
@@ -0,0 +1,98 @@
...
+/// * `ctx_opt` - A context for reporting progress (only useful within quickwit actor).
+pub async fn run_evaluate_retention_policy(
+    index_id: &str,
+    metastore: Arc<dyn Metastore>,
+    retention_policy: &RetentionPolicy,
+    ctx_opt: Option<&ActorContext<RetentionPolicyEvaluator>>,
+) -> anyhow::Result<Vec<SplitMetadata>> {
+    // Select published splits and filter for eligibility.
+    let current_date_time = OffsetDateTime::now_utc();
+    let discardable_splits: Vec<SplitMetadata> = metastore
```
guilload, line 44: I believe `expired` is better in this context.

quickwit-janitor/src/retention_policy_evaluation.rs
```diff
@@ -0,0 +1,98 @@
...
+/// Checks to see if a split can be discarded based on retention policy.
+fn is_discardable(
+    current_date_time: OffsetDateTime,
+    split: &Split,
+    retention_policy: &RetentionPolicy,
+) -> anyhow::Result<bool> {
+    let retention_period = retention_policy.retention_period()?;
+    let publish_timestamp = split
+        .publish_timestamp
+        .context("Expected a split with a valid published timestamp.")?;
```
guilload, line 81: That's not backward compatible. Use the `update_timestamp` instead for splits produced pre-Quickwit 0.4. Comment the code abundantly.

quickwit-janitor/src/retention_policy_evaluation.rs
```diff
@@ -0,0 +1,98 @@
...
+        RetentionPolicyCutoffReference::PublishTimestamp => {
+            (current_date_time - publish_date_time) >= retention_period
+        }
+        RetentionPolicyCutoffReference::SplitTimestampField => {
+            match &split.split_metadata.time_range {
+                Some(time_range) => {
+                    let time_range_end = OffsetDateTime::from_unix_timestamp(*time_range.end())?;
+                    (current_date_time - time_range_end) >= retention_period
+                }
+                None => (current_date_time - publish_date_time) >= retention_period,
```
guilload, line 93: Sneaky. Let's log a warning and do nothing if the split has no timestamp field and the retention policy demands one. EAFP does not apply when it comes to deleting data.

quickwit-janitor/src/retention_policy_evaluation.rs
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+    };
+    Ok(is_discardable)
+}
```
guilload, line 98: You could test this function unitarily here, and in the actor, just test that the function is correctly called.

quickwit-janitor/src/actors/retention_policy_evaluator.rs
evanxg852000: Yes related to the above comment. in principle we should always get a valid  duration_till_next_interval

quickwit-janitor/src/actors/retention_policy_evaluator.rs
fmassot: I like executor :)

evanxg852000: Ok then 👍 , curiously I started with Executor then switched to Evaluator.

guilload (approved):

PR #1959, opened 2022-09-15
Refactor IndexingSplitStore to be shared among pipelines

evanxg852000:
- Make indexing pipeline on same `(indiex_id, source_id)` share the same `IndexingSplitStore`  and `IndexingDirectory`


quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -401,6 +420,10 @@ impl Handler<SuperviseLoop> for IndexingService {
                     }
                 },
             );
+
+        let active_pipeline_ids = self.pipeline_handles.keys();
```
guilload, line 424:
```suggestion
        let running_pipeline_ids = self.pipeline_handles.keys();
```

quickwit/quickwit-indexing/src/split_store/indexing_split_store.rs
```diff
@@ -244,6 +244,10 @@ impl IndexingSplitStore {
         Ok(())
     }
 
+    pub fn get_merge_policy(&self) -> Arc<dyn MergePolicy> {
```
guilload, line 247: We shouldn't have to go through the split store to get to the merge policy. See my comment above.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -123,6 +127,7 @@ impl IndexingService {
             pipeline_handles: Default::default(),
             state: Default::default(),
             enable_ingest_api,
+            shared_resource_manager: IndexingSharedResourceManager::new(),
```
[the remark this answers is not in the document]
evanxg852000: ok 👍 . I started with indexing_directory, and split_store like here but realized we will do that for the merge pipeline too.  I was feeling like it's going to get in the way of the main actor's logic so to me it would be simpler isolating it..

quickwit/quickwit-indexing/src/models/shared_resource.rs
```diff
@@ -0,0 +1,223 @@
...
+        index_metadata: &IndexMetadata,
+        max_num_bytes: usize,
+        max_num_splits: usize,
+    ) -> Result<IndexingSharedResource, IndexingServiceError> {
+        let key = self.resource_key(pipeline_id);
+        if let Some(resource) = self.shared_resources.get(&key) {
+            return Ok(resource.clone());
+        }
+
+        let stable_multitenant_merge_policy = StableMultitenantWithTimestampMergePolicy {
```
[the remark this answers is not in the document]
evanxg852000: I don't think there will be a need to create a function pointer `is_mature_fn` then, split_store can still have a cloned copy of the merge policy. Just that we change where we are instanciating it as you mentioned. 
The plan here was to only care about the split_store and then refactor it if necessary to work with the new Merge-pipeline.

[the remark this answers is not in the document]
guilload: I agree, the pointer function is just nice to have: I never understood why the split store depends on the entire merge policy.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -332,6 +360,120 @@ impl IndexingService {
...
+            .map(|pipeline_id| pipeline_id.index_id.clone())
+            .collect();
+        let current_keys: HashSet<String> = self.merge_policies.keys().cloned().collect();
+        let deleted_keys = &current_keys - &running_keys;
+        for key in deleted_keys {
+            self.merge_policies.remove(&key);
+        }
+    }
+
+    fn retain_indexing_directories(&mut self, running_pipeline_ids: &[IndexingPipelineId]) {
```
evanxg852000, line 450: @guilload  `retain_indexing_directories` and `retain_split_stores` could be combined into one function but I kept them separate for simplicity. let me know if you feel it's ok to combine.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -96,6 +105,12 @@ pub struct IndexingService {
     pipeline_handles: HashMap<IndexingPipelineId, ActorHandle<IndexingPipeline>>,
     state: IndexingServiceState,
     enable_ingest_api: bool,
+    // Hashmap of of index_id to merge policy.
+    merge_policies: HashMap<String, Arc<dyn MergePolicy>>,
+    // Hashmap of (index_id, source_id) to indexing directory.
+    indexing_directories: HashMap<String, IndexingDirectory>,
```
evanxg852000, line 111: @guilload I preferred a combination a  String to avoid 2x allocations, besides constructs like `(pipeline_id.index_id.clone(), pipeline.source_id.clone())` would have been creeping.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -332,6 +360,120 @@ impl IndexingService {
...
+            storage.clone(),
+            cache_directory,
+            IndexingSplitStoreParams {
+                max_num_bytes: self.split_store_max_num_bytes,
+                max_num_splits: self.split_store_max_num_splits,
+            },
+            merge_policy,
+        )?;
+
+        self.split_stores.insert(key.clone(), split_store);
```
guilload, line 434: Ditto. Insert cloned object + return object = no ugly return with lookup + unwrap + clone;

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -332,6 +360,120 @@ impl IndexingService {
...
+        Ok(self.split_stores.get(&key).unwrap().clone())
+    }
+
+    fn retain_merge_policies(&mut self, running_pipeline_ids: &[IndexingPipelineId]) {
```
guilload, line 438: Or...
```
        let running_merge_policies: HashSet<&String> = running_pipeline_ids
            .iter()
            .map(|pipeline_id| &pipeline_id.index_id)
            .collect();
        self.merge_policies.retain(|index_id, _| running_merge_policies.contains(index_id));
```

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -332,6 +360,120 @@ impl IndexingService {
...
+            .collect();
+        let current_keys: HashSet<String> = self.merge_policies.keys().cloned().collect();
+        let deleted_keys = &current_keys - &running_keys;
+        for key in deleted_keys {
+            self.merge_policies.remove(&key);
+        }
+    }
+
+    fn retain_indexing_directories(&mut self, running_pipeline_ids: &[IndexingPipelineId]) {
+        let running_keys: HashSet<String> = running_pipeline_ids
```
guilload, line 451: Ditto.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
guilload: That's fine as is.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -84,6 +90,9 @@ pub struct IndexingServiceState {
     pub num_running_pipelines: usize,
     pub num_successful_pipelines: usize,
     pub num_failed_pipelines: usize,
+    pub num_merge_policies: usize,
```
guilload, line 93: Seems convenient for unit testing but that's not really needed otherwise.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -332,6 +360,120 @@ impl IndexingService {
...
+            .map(|pipeline_id| self.index_source_key(pipeline_id))
+            .collect();
+        let current_keys: HashSet<String> = self.indexing_directories.keys().cloned().collect();
+        let deleted_keys = &current_keys - &running_keys;
+        for key in deleted_keys {
+            self.indexing_directories.remove(&key);
+        }
+    }
+
+    fn retain_split_stores(&mut self, running_pipeline_ids: &[IndexingPipelineId]) {
```
guilload, line 462: Tritto.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -572,10 +723,13 @@ mod tests {
         );
         let (indexing_server_mailbox, indexing_server_handle) =
             universe.spawn_actor(indexing_server).spawn();
-        let observation = indexing_server_handle.observe().await;
```
guilload, line 575: On my machine, this test takes 6-7s to run on `main` vs. 9-10s. Why is that, it's not like we are doing more CPU-heavy work. I think that is why you had to switch from `observe` to `process_pending_and_observe`.

evanxg852000: I changed the way to test, but having to `simulate_time_shift` and wait for the supervise message to be handled seems to be the cause. reducing the assertions reduces the time it takes.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -539,6 +674,36 @@ mod tests {
 
     use super::*;
 
+    #[derive(Debug)]
+    struct AssertState(usize, usize, usize);
```
guilload, line 678: Cool.

guilload:
It just came to my mind that we can hold weak references to the indexing directory, merge policy, and split store objects so that we don't have to write that retain logic for those. Let's do that, and then we should be good to go.

guilload (approved):

PR #1965, opened 2022-09-16
Supervisor

fulmicoton:


quickwit/quickwit-actors/src/kill_switch.rs
```diff
@@ -18,36 +18,82 @@
...
+            children: Mutex::default(),
         }
     }
 }
 
-impl KillSwitch {
-    pub fn kill(&self) {
-        debug!("kill-switch-activated");
-        self.alive.store(false, Ordering::Relaxed);
+fn gc(children: &mut Vec<Weak<Inner>>) {
```
fmassot, line 44: nitpick: use `garbage_collect` as we tend to use longer and explicit names in the project.

quickwit/quickwit-actors/src/supervisor.rs
```diff
@@ -0,0 +1,383 @@
...
+        Ok(())
+    }
+}
+
+impl<A: Actor> Supervisor<A> {
+    async fn supervise(
+        &mut self,
+        ctx: &ActorContext<Supervisor<A>>,
+    ) -> Result<(), ActorExitStatus> {
+        match self.handle_opt.as_ref().unwrap().health() {
```
fmassot, line 102: It's possible to create a supervisor without a handle so unwrap.
We could either force the instantiation of a supervisor with an ActorHandle directly (and not Option<ActorHandle>) or avoid the unwrap here.

quickwit/quickwit-actors/src/supervisor.rs
```diff
@@ -0,0 +1,383 @@
...
+            }
+            ActorExitStatus::Panicked => {
+                self.state.num_panics += 1;
+            }
+        }
+        info!("respawning-actor");
+        let (_, actor_handle) = ctx
+            .spawn_actor()
+            .set_mailboxes(self.mailbox.clone(), self.inbox.clone())
+            .set_kill_switch(KillSwitch::default())
```
fmassot, line 146: We probably want to use a supervisor child kill switch

fulmicoton: oups yes absolutely.

quickwit/quickwit-actors/src/supervisor.rs
fulmicoton: The Option  is not for the instantiation, It's because joininig the Handle consume the Handle

fulmicoton: (in finalize)

quickwit/quickwit-actors/src/actor_handle.rs
```diff
@@ -158,13 +158,13 @@ impl<A: Actor> ActorHandle<A> {
     /// as the finalize logic may behave differently depending on the exit status.
     pub async fn kill(self) -> (ActorExitStatus, A::ObservableState) {
         self.actor_context.kill_switch().kill();
-        let _ = self
```
fmassot, line 161: Ok, I think you're right in removing this send kill message. It seems that it was not necessary to send this kill message as in the actor loop we always check if the kill switch is activated. Plus in the in actor handle, the implementation can check if the kill switch too. So in the end, activating the kill switch is always more powerful.

quickwit/quickwit-actors/src/supervisor.rs
fmassot: yes yes we need to keep the option ofc. I was just unsatisfied by the unwrap.

fmassot: (and we consume also the handle in the supervise method)

fmassot (approved):
Great PR, can't wait to use the supervisor.

quickwit/quickwit-actors/src/supervisor.rs
fmassot: To be complete, this is what I meant by enforcing the user to give a handle

```rust
impl<A: Actor> Supervisor<A> {
  pub fn new(
    actor_name: String,
    actor_factory: Box<dyn Fn() -> A + Sync + Send>,
    inbox: Inbox<A>,
    mailbox: Mailbox<A>,
    handle: ActorHandle<A>,
  ) -> {
    Self {
        actor_name,
        actor_factory,
        inbox,
        mailbox,
        handle_opt: Some(handle),
    }
 }
}
```

fulmicoton: ok I added a constructor.

PR #1971, opened 2022-09-19
Remove sequencer from merge pipelines.

### LINKED ISSUE #1940: Simplify the merge pipeline
As discussed in #1795, the merge pipeline can be simplified:
- we can remove the sequencer which is needed only by the indexing pipeline
- we don't need to restart the whole pipeline if an actor is failing, we can just respawn the failing actor. However this would require to track ongoing merges or failing merges with a drop guard. 

Ref: 
> The merge pipeline supervisor could for instance send the merge planner a message to trigger the reload of the information from the SQL database (or we could have some Drop Guard travel with the splits... The Drop guard would lock/unlock the splits).

fmassot:
Fix #1940 

This PR updates the `Uploader` actor to bypass the `Sequencer` when it's not needed (that is for merge & delete pipelines).

It's still a draft as I need to add a test. Even if the code is not "beautiful" as I'm just using enums to handle the case where we use the sequencer and the case where we use the publisher, I think it's worth it as it removes the sequencer from 2 pipelines.

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/uploader.rs
+++ b/quickwit/quickwit-indexing/src/actors/uploader.rs
@@ -649,6 +649,7 @@ mod tests {
                         replaced_split_ids: Vec::new(),
                         split_id: "test-split".to_string(),
                         delete_opstamp: 10,
+                        num_merge_ops: 0,
                     },
                     split_scratch_directory,
                     tags: Default::default(),
--- a/quickwit/quickwit-indexing/src/actors/uploader.rs
+++ b/quickwit/quickwit-indexing/src/actors/uploader.rs
@@ -34,12 +34,13 @@ use quickwit_metastore::checkpoint::IndexCheckpointDelta;
 use quickwit_metastore::{Metastore, SplitMetadata};
 use quickwit_storage::SplitPayloadBuilder;
 use time::OffsetDateTime;
+use tokio::sync::oneshot::Sender;
 use tokio::sync::{oneshot, Semaphore, SemaphorePermit};
 use tracing::{info, info_span, warn, Instrument, Span};
 
 use crate::actors::sequencer::{Sequencer, SequencerCommand};
 use crate::actors::Publisher;
-use crate::models::{PackagedSplit, PackagedSplitBatch, PublishLock, SplitUpdate};
+use crate::models::{PackagedSplit, PackagedSplitBatch, PublishLock, SplitsUpdate};
 use crate::split_store::IndexingSplitStore;
 
 pub const MAX_CONCURRENT_SPLIT_UPLOAD: usize = 4;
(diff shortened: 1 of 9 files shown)


guilload:
Might be easier to pass a no-op Sequencer.

quickwit/quickwit-indexing/src/actors/uploader.rs
```diff
@@ -132,15 +144,22 @@ impl Handler<PackagedSplitBatch> for Uploader {
...
-            oneshot::channel::<SequencerCommand<SplitUpdate>>();
-
-        // We send the future to the sequencer right away.
-
-        // The sequencer will then resolve the future in their arrival order and ensure that the
-        // publisher publishes splits in order.
-        ctx.send_message(&self.sequencer_mailbox, split_uploaded_rx)
-            .await?;
+        let uploader_sender = match &self.split_update_mailbox {
+            SplitUpdateMailbox::Sequencer(sequencer_mailbox) => {
```
fulmicoton, line 148: this is not the right place to put this code. This should be in the `impl SplitUpdateMailbox`

quickwit/quickwit-indexing/src/actors/uploader.rs
```diff
@@ -174,10 +194,12 @@ impl Handler<PackagedSplitBatch> for Uploader {
                             split_ids=?split_ids,
                             "Splits' publish lock is dead."
                         );
-                        if split_uploaded_tx.send(SequencerCommand::Discard).is_err() {
-                            bail!("Failed to send cancel command to sequencer. The sequencer is probably dead.");
+                        if let SplitUpdateSender::Sequencer(split_uploader_tx) = uploader_sender {
```
fulmicoton, line 197: This should be part of the `impl SplitUpdateSender`

quickwit/quickwit-indexing/src/actors/uploader.rs
```diff
@@ -193,12 +215,23 @@ impl Handler<PackagedSplitBatch> for Uploader {
...
                     packaged_splits_and_metadatas.push((split, upload_result.unwrap()));
                 }
-                let publisher_message = make_publish_operation(index_id, batch.publish_lock, packaged_splits_and_metadatas, batch.checkpoint_delta_opt, batch.date_of_birth);
-                if let Err(publisher_message) = split_uploaded_tx.send(publisher_message) {
-                    bail!(
-                        "Failed to send upload split `{:?}`. The publisher is probably dead.",
-                        &publisher_message
-                    );
+                let split_update = make_publish_operation(index_id, batch.publish_lock, packaged_splits_and_metadatas, batch.checkpoint_delta_opt, batch.date_of_birth);
+                match uploader_sender {
```
fulmicoton, line 219: It should be part of the `impl SplitUpdateSender`

PR #1973, opened 2022-09-19
Implement weak indexing directory and split store

guilload:
### Description
Introduce a `WeakIndexingDirectory` and `WeakIndexingSplitStore` to make the code a little nicer.

### How was this PR tested?
```bash
cargo test --manifest-path quickwit/Cargo.toml -p quickwit-indexing -- test_indexing_service
```


fulmicoton:
I am on the fence with that kind of change. The intent is to hide an implement detail and reduce "noise"...
But the fact it is an Arc actually provides a lot of very important information:
- you can clone that, and it is cheap
- expect this thing to have copies, and Drop code to run only when those are all dropped.
- don't create an Arc of this thing.. It is already in an Arc.
- etc.

fulmicoton (approved):

PR #1974, opened 2022-09-19
[Kube] Add metastore service deployment

guilload:
### Description
Per title.

### How was this PR tested?
Deployed in local minikube

### DIFF
--- a/distribution/kubernetes/metastore-deployment.yaml
+++ b/distribution/kubernetes/metastore-deployment.yaml
@@ -1,25 +1,23 @@
 apiVersion: apps/v1
-kind: StatefulSet
+kind: Deployment
 metadata:
-  name: indexer
+  name: metastore
 spec:
   replicas: 1
   selector:
     matchLabels:
       app: quickwit
-      service: indexer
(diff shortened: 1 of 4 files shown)


distribution/kubernetes/metastore-deployment.yaml
```diff
@@ -0,0 +1,101 @@
...
+      containers:
+        - name: quickwit
+          image: quickwit/quickwit:edge
+          args: ["run", "--service", "metastore"]
+          ports:
+            - containerPort: 7280
+              name: rest
+            - containerPort: 7281
+              name: grpc
+            - containerPort: 7282
```
fulmicoton, line 26: why we not the default ports?

fulmicoton (approved):

distribution/kubernetes/metastore-deployment.yaml
guilload: Kube cannot do TCP and UDP on the same port, so I set the gossip port on `7282`. Because of that, I chose to set the other ports explicitly.

PR #1977, opened 2022-09-20
Failpoint test fix.

fulmicoton:
After removing the use of the Kill message,
some actors could be stuck awaiting for a message, and never checking for possible message.

Relying on a Kill message is not an option, due to the recycling of mailbox introduced in the supervisor PR.

The fix consists in adding a Nudge message that
is No-op.

### DIFF
--- a/quickwit/quickwit-actors/src/command.rs
+++ b/quickwit/quickwit-actors/src/command.rs
@@ -79,16 +79,27 @@ pub enum Command {
     /// This is the equivalent of sending SIGINT/Ctrl-C to a process.
     Quit,
 
-    /// Kill the actor. The behavior is the same as if an actor detected that its kill switch
-    /// was pushed.
+    /// Nudging is a No-op message.
     ///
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-actors/src/command.rs
```diff
@@ -79,16 +79,27 @@ pub enum Command {
     /// This is the equivalent of sending SIGINT/Ctrl-C to a process.
     Quit,
 
-    /// Kill the actor. The behavior is the same as if an actor detected that its kill switch
-    /// was pushed.
+    /// Nudging is a No-op message.
```
fmassot, line 82: Great doc.

fmassot (approved):
LGTM.

Let's wait the result of the code coverage [here](https://github.com/quickwit-oss/quickwit/actions/runs/3088100093)
and merge if successful.

fmassot:
Coverage is fixed.

PR #1983, opened 2022-09-20
Fix and clean cluster members.

### LINKED ISSUE #1982: Watching clusters members does not return an updated `me` member

fmassot:
Fix #1982.

And I also changed a bit the way we use the member watcher.

Note:  the current watcher is broken if we want to use up-to-date properties of cluster members. Example: we start a node with the server `searcher` and restart afterward with another server. Chitchat will probably not detect a change in readiness and the member watcher of other nodes will not receive member changes. And the other nodes will have their search pool with this no-more searcher node.

PTAL.

### DIFF
--- a/quickwit/quickwit-cluster/src/cluster.rs
+++ b/quickwit/quickwit-cluster/src/cluster.rs
@@ -62,8 +62,6 @@ pub struct ClusterMember {
     pub generation: u64,
     /// Gossip advertise address.
     pub gossip_advertise_addr: SocketAddr,
-    /// If true, it means self.
-    pub is_self: bool,
     /// Available services.
     pub available_services: HashSet<QuickwitService>,
     /// gRPC advertise address.
@@ -82,7 +80,6 @@ impl ClusterMember {
             node_unique_id,
             generation,
             gossip_advertise_addr,
-            is_self: true,
             available_services,
             grpc_advertise_addr,
         }
@@ -104,7 +101,7 @@ pub struct Cluster {
     /// ID of the cluster joined.
     pub cluster_id: String,
     /// Node ID.
-    pub node_id: String,
+    pub node_id: NodeId,
     /// A socket address that represents itself.
     pub gossip_listen_addr: SocketAddr,
 
@@ -112,8 +109,8 @@ pub struct Cluster {
     /// If it is dropped, the chitchat server will stop.
     chitchat_handle: ChitchatHandle,
 
-    /// A receiver (channel) for exchanging information about the nodes in the cluster.
-    members: watch::Receiver<Vec<ClusterMember>>,
+    /// A watcher that receives ready members changes from chitchat.
+    members_rx: watch::Receiver<Vec<ClusterMember>>,
 
     /// A stop flag of cluster monitoring task.
     /// Once the cluster is created, a task to monitor cluster events will be started.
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -246,22 +232,45 @@ impl Cluster {
...
     }
 
-    /// Return `members` list
-    pub fn members(&self) -> Vec<ClusterMember> {
-        self.members.borrow().clone()
+    /// Returns ready [`ClusterMember`]s built directly from the current state of the cluster.
+    pub async fn ready_members_from_cluster_state(&self) -> Vec<ClusterMember> {
+        let cluster_state = self.state().await;
+        let mut ready_nodes = cluster_state.ready_nodes.clone();
+        let is_self_node_ready = cluster_state
```
fmassot, line 256: ok I admit, this is ugly. Adding a method on `ClusterState` would make it nicer but for now `ClusterState` is just used for the API so I did not want to "enrich" it right now.

quickwit/quickwit-search/src/search_client_pool.rs
```diff
@@ -167,41 +176,15 @@ impl SearchClientPool {
     /// When a client pool is created, the thread that monitors cluster members
     /// will be started at the same time.
     pub async fn create_and_keep_updated(
-        current_members: &[ClusterMember],
```
fmassot, line 170: Passing `current_members` and a member `WatchStream` is, IMHO, a wrong usage of a watch stream.
The watch stream always start with the last seen values. And we always send at least the "self" member, so we always have something. At least self.

quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -209,32 +210,17 @@ impl Cluster {
...
-                        }
-                        member_res.ok()
-                    })
-                    .collect_vec();
-                members.push(me.clone());
-
+            while let Some(mut members_set) = node_change_receiver.next().await {
+                let cluster_state_snapshot = chitchat.lock().await.state_snapshot();
+                // `members_set` does not contain `self`.
+                members_set.insert(self_node_id.clone());
```
fmassot, line 216: This fixed the issue.

fmassot: But I broke the `is_self`.

quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -62,8 +62,6 @@ pub struct ClusterMember {
     pub generation: u64,
     /// Gossip advertise address.
     pub gossip_advertise_addr: SocketAddr,
-    /// If true, it means self.
```
fmassot, line 65: Not used currently.

quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -510,35 +539,39 @@ mod tests {
...
             .iter()
             .map(|member| member.gossip_advertise_addr)
             .collect();
         let expected_members = vec![cluster.gossip_listen_addr];
         assert_eq!(members, expected_members);
         assert!(!cluster.is_self_node_ready().await);
-        let cluster_state = cluster.state().await;
-        let self_node_state_not_ready = cluster_state
-            .state
+        assert_eq!(cluster.ready_members_from_chitchat_state().await.len(), 0);
```
fulmicoton, line 550: I suspect clippy will annoy you 
and tell you to make it

assert!(cluster.ready_members_from_chitchat_state().await.is_empty());

fulmicoton (approved):
Great job! The code is much cleaner

PR #1995, opened 2022-09-22
Added a simple quickwit actor benchmark

fulmicoton:
We recently added an extra heavy actor in the pipeline (the doc processor).
Before the document processing work was done in a dedicated thread consuming tantivy documents.

I thought it might be a good time to measure the overhead added by the actor framework vs
just using regular tokio tasks and a channel.

Before the PR, actor process messages at a pace of around 1M messages per second.
This is roughly 10x slower than using a stock channel and a tokio task.

The main cause for the slowdown is the fact that we yield after each message.
The reason is that tokio does not really preempt all tasks on `.await`.

Instead, tokio has a counter that gets incremented after each call to .await, and if a 
task does not get stopped 128 times in a row because it received a `Pending` on a poll, tokio will preempt the task.
Some of our actors however are long running and SHOULD definitely yield manually (e.g. the packager). 

By removing this yield, we reach a pace of  `2 M  messages / s`, and possibly other hidden benefits.

This PR adds a parameter to make this yield after each message behavior optional, and adds an actor benchmark.


quickwit/quickwit-actors/src/spawn_builder.rs
```diff
@@ -204,10 +204,16 @@ impl<A: Actor> ActorExecutionEnv<A> {
     }
 
     async fn yield_and_check_if_killed(&self) -> Result<(), ActorExitStatus> {
-        self.ctx.protect_future(tokio::task::yield_now()).await;
         if self.ctx.kill_switch().is_dead() {
             return Err(ActorExitStatus::Killed);
         }
+        if self.actor.yield_after_each_message() {
+            self.ctx.protect_future(tokio::task::yield_now()).await;
+            if self.ctx.kill_switch().is_dead() {
```
fulmicoton, line 212: The yield can actually be rather long.
We also add a killswitch check after.

quickwit/quickwit-indexing/src/source/kinesis/shard_consumer.rs
```diff
@@ -157,6 +157,10 @@ impl Actor for ShardConsumer {
         Ok(())
     }
 
+    fn yield_after_each_message(self) -> bool {
```
fmassot, line 160:
```suggestion
    fn yield_after_each_message(&self) -> bool {
```

quickwit/quickwit-actors/benches/bench.rs
```diff
@@ -0,0 +1,174 @@
...
+            sum += 1;
+        }
+        sum
+    });
+    drop(tx);
+    let total = join.await.unwrap();
+    assert_eq!(total, num_messages as u64);
+}
+
+async fn highprior_bench_code(num_messages: usize) {
```
fmassot, line 88: We are calling `send_low_priority` in this bench
```suggestion
async fn low_priority_bench_code(num_messages: usize) {
```

quickwit/quickwit-indexing/src/source/kinesis/shard_consumer.rs
fulmicoton: thx

fmassot (approved):
Sounds good.

fmassot:
This topic is really tricky.

While the benchmark result is very clear, I expect that the hidden benefits are very very small (or I may have missed something).

Yield or not yield is not obvious, we should probably add some docs in the Actor trait that explains a bit how to choose.

Basically, we should yield at every message on long-running task.

fmassot:
(I can do a PR on this)

PR #1998, opened 2022-09-22
Fix nightly release CI

boraarslan:


.github/actions/build-binary/action.yml
```diff
@@ -19,22 +19,24 @@ runs:
...
     - run: make build-ui
       shell: bash
     - uses: actions-rs/toolchain@v1
       with:
         toolchain: stable
         target: ${{ inputs.target }}
         override: true
-    - uses: actions-rs/cargo@v1
-      with:
-        use-cross: true
```
guilload, line 32: This doesn't work?

boraarslan: Unfortunately, GitHub actions do not let you change directories with `working-directory`, and cargo action do not expose any args to specify the rust project directory.

guilload (approved):

Issue #2004, opened 2022-09-22
Kafka SourceActor timeout when starting indexers on k8s


fmassot:
It happens only when starting the indexer, the pipelines are working well after generation 2 or 3.

I attached the log but there is not much to say as we only see the actor timeout.

[logs.txt](https://github.com/quickwit-oss/quickwit/files/9630030/logs.txt)

guilload:
I'm able to reproduce source actor timeout. When I set `yield_after_each_message` to `true`, this behavior seems to disappear.

PR #2009, opened 2022-09-23
Fix source actor timeout

guilload:
### Description
Attempt to fix source actor timeout reported in #2004. Recording progress after each pass through the `emit_batches` method seems to fix the issue, but I feel like this problem should be addressed in the actor framework, so I attempted to reproduce the bug in a unit test in the actors' crate. Without success, alas.

### How was this PR tested?
Successfully spawned 50 indexing pipelines consuming an empty topic on my local branch. I was not able to do so on `main`.

### DIFF
--- a/quickwit/quickwit-actors/src/tests.rs
+++ b/quickwit/quickwit-actors/src/tests.rs
@@ -341,6 +341,10 @@ impl Actor for LoopingActor {
         self.clone()
     }
 
+    fn yield_after_each_message(&self) -> bool {
+        false
+    }
+
     async fn initialize(&mut self, ctx: &ActorContext<Self>) -> Result<(), ActorExitStatus> {
         self.handle(Loop, ctx).await
     }
@@ -374,7 +378,7 @@ impl Handler<SingleShot> for LoopingActor {
     }
 }
 
-#[tokio::test]
+#[tokio::test(flavor = "multi_thread")]
 async fn test_looping() -> anyhow::Result<()> {
     let universe = Universe::new();
     let looping_actor = LoopingActor::default();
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-indexing/src/source/mod.rs
```diff
@@ -266,6 +266,7 @@ impl Handler<Loop> for SourceActor {
             .source
             .emit_batches(&self.doc_processor_mailbox, ctx)
             .await?;
+        ctx.record_progress();
```
guilload, line 269: This seems to fix the issue.

quickwit/quickwit-actors/src/tests.rs
```diff
@@ -374,7 +378,7 @@ impl Handler<SingleShot> for LoopingActor {
     }
 }
 
-#[tokio::test]
+#[tokio::test(flavor = "multi_thread")]
```
guilload, line 381: The test times out with a single-threaded runtime, which makes sense.

quickwit/quickwit-actors/src/tests.rs
```diff
@@ -374,7 +378,7 @@ impl Handler<SingleShot> for LoopingActor {
...
 async fn test_looping() -> anyhow::Result<()> {
```
guilload, line 382: I time-shifted the universe by `HEARTBEAT * 2`, hoping to reproduce the issue, but that did not work. Same when sleeping for the same amount of time.

fulmicoton (approved):

fulmicoton:
I suggest we merge right away considering the number of people playing with Kafka these days.

PR #2016, opened 2022-09-26
Fix consumer group name collision

guilload:
### Description
Per title.

### How was this PR tested?
```
cargo test --manifest-path quickwit/Cargo.toml --all-features -p quickwit-indexing
```


fmassot (approved):
LGTM

Issue #1638, opened 2022-06-15
Starting quickwit with 100 index fails with "Failed to empty scratch directory" (too many file descriptors)


fulmicoton:

guilload:
Mitigated by #2017. Systems with a low limit of file descriptors will still suffer from this issue, and the only fix is to increase the limit of FDs.

PR #2017, opened 2022-09-26
Limit the number of indexing pipelines that can be spawned concurrently.

guilload:
### Description
Limit the number of indexing pipelines that can be spawned concurrently.

### How was this PR tested?
`make test-all`

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexing_pipeline.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexing_pipeline.rs
@@ -30,6 +30,7 @@ use quickwit_doc_mapper::DocMapper;
 use quickwit_metastore::{IndexMetadata, Metastore, MetastoreError, SplitState};
 use quickwit_storage::Storage;
 use tokio::join;
+use tokio::sync::Semaphore;
 use tracing::{debug, error, info, info_span, instrument, Span};
 
 use crate::actors::doc_processor::DocProcessor;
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -44,6 +45,11 @@ use crate::split_store::IndexingSplitStore;
 
 const MAX_RETRY_DELAY: Duration = Duration::from_secs(600); // 10 min.
 
+/// Spawning an indexing pipeline puts a lot of pressure on the file system, metastore, etc. so
```
fmassot, line 48: I'm not sure if putting a semaphore like this one is a good idea.

I expect to see a user starts an indexer and open an issue saying that it's not working.

Do you think we can put a limit right from the start of an indexer? So that we will tell the user that only the first 10 pipelines will run? One drawback of this approach is that if some pipelines can end, we end up by running less than 10 pipelines.

guilload: This change does not limit the total number of indexing pipelines that can run on the node. This limits the number of indexing pipelines spawned concurrently on the node. The semaphore is only held while spawning the pipeline.

fmassot: Mmmm interesting, that's indeed a good idea, sorry for the confusion.

fmassot: Nitpick: the "See also #1638" is a bit frustrating as we want go click on it :)

fmassot (approved):
LGTM

PR #2027, opened 2022-09-28
use tantivy delete_query for document deletion

### LINKED ISSUE #2014: Plug the new tantivy delete code into quickwit
See in the title :)

We have a bunch of actors in the quickwit-indexing/src/actors that are running deletes in a silly way today. Adopt the new tantivy code in them.

trinity-1686a:
### Description

fix #2014
Use `delete_query` to delete documents.

### How was this PR tested?

I ingested a bunch of documents, requested the deletion of some of them 
```
curl "http://127.0.0.1:7280/api/v1/wikipedia/delete-tasks" \
  -H 'Content-type: application/json' \
  -d '{"query":"barack+AND+obama", "search_fields":["title", "body"]}'
```
and verified it did delete the matching documents.
/!\ it did delete marching documents in mature splits, but not in immature ones, so some matching documents may not get deleted. My understanding is it's a known limitation


fmassot:
> /!\ it did delete marching documents in mature splits, but not in immature ones, so some matching documents may not get deleted. My understanding is it's a known limitation

Indeed, this is sufficient right now, but this limitation is not very convenient for the user. My approach on this would be to document the limitation and tell the user to check that his splits should not stay immature. If they are, the user needs to fix it by reconfiguring its merge policy for example. And this is too much pain for the user, we will improve the feature. We could even add a health check endpoint by index so that it's not too painful to check if an index is healthy or not.

Let's open an issue on this anyway.

fmassot (approved):
LGTM

PR #2032, opened 2022-09-28
Fix misleading UI numbers

### LINKED ISSUE #2006: num docs as displayed in the UI is misleading
As reported in #1987.

The UI reports the overall number of docs in splits. (including splits marked as deleted).

https://github.com/quickwit-oss/quickwit/blob/f32f31694b0815da2b0eb63f70db122a23421bee/quickwit/quickwit-ui/src/components/IndexSummary.tsx#L43-L49

This is quite misleading for users.

We probably want 

- num docs (defined as published splits only)
- num published splits

Size is fishy too. I don't know how it is computed. We need to edit the lable to tell if this is the split size of the raw document size, and make sure this is just published stuff if it is the latter.

And possibly:
- num staged splits
- num splits marked as deleted

boraarslan:
Fixes #2006 

I also changed the style a little bit. (Thanks @Fursum)

It currently looks like this:
image

### DIFF
--- a/quickwit/quickwit-ui/src/components/IndexSummary.tsx
+++ b/quickwit/quickwit-ui/src/components/IndexSummary.tsx
@@ -21,6 +21,7 @@ import styled from "@emotion/styled";
 import { Paper } from "@mui/material";
 import dayjs from "dayjs";
 import utc from "dayjs/plugin/utc"
+import { FC, ReactNode } from "react";
 import NumberFormat from "react-number-format";
 import { Index } from "../utils/models";
 dayjs.extend(utc);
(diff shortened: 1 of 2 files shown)


fmassot (approved):
LGTM! Nice job!

quickwit/quickwit-ui/src/components/IndexSummary.tsx
```diff
@@ -34,57 +35,69 @@ const Row = styled.div`
...
+        <IndexRow title="Size of published splits:">
+          <NumberFormat
+            value={total_num_bytes / 1000000}
+            displayType={"text"}
+            thousandSeparator={true}
+            suffix=" MB"
+            decimalScale={2}
+          />
+        </IndexRow>
+        <IndexRow title="Number of documents (Published):">
```
guilload, line 91:
```suggestion
        <IndexRow title="Number of published documents:">
```

guilload: ?

quickwit/quickwit-ui/src/utils/models.ts
```diff
@@ -253,4 +252,4 @@ export type NodeId = {
   id: string,
   grpc_address: string,
   self: boolean,
-}
```
guilload, line 256: :D

boraarslan: LOL

PR #2043, opened 2022-09-30
[fix] Refactoring

ppodolsky:
@fulmicoton Would you mind to take a look? I do not insist on this particular refactoring, it is just a suggestion.
The main idea behind is to decouple `quickwit_directories` from tokio and some other dependencies for making it WASM-compatible.

For this purpose I:
- Splitted quickwit_storage into quickwit_storage and quickwit_cache
- Added feature flags to quickwit_common to make tokio-less build possible
- Moved metrics server from quickwit_common to the place where it is actually used. It allowed to eliminate dependency from `wrap`
- Refactored some methods to accept `Cache/StorageMetrics` as arguments and eliminated global metric object usage there. Though, I didn't manage to do it to the very end, the global object is still here. It allowed to do decoupling further


fulmicoton:
@ppodolsky We don't have any use case for quickwit to compile in wasm. This PR would add extra complexity without beneficial to the project.

I understand you would like to depend on a quickwit internal crate for your project. The best option here is probably for you to fork (and respect the AGPL license of course).

PR #2056, opened 2022-10-03
Fixes tracing in indexing

### LINKED ISSUE #2055: A trace has a large number of spans
image

fulmicoton:
Closes #2055 

Before this PR actors would be executed on a root span that was
associated to their execution loop.

That span would be gigantic and would never be closed.

In addition, the message spans did not have any associated name by
default. This solution made logs that looked good in stdout but were
terribly broken when consumed as traces in a tool like jaeger.

The following PR fixes the issue by removing the crazy root span.
It also does something a tiny bit more ambitious by creating a root span
per indexing batch. The batch follows the split, so that it is easier
to inspect how a trip travels through the indexing pipeline.

It is also possible to see the associated context, timings etc.
for each split creation or merge operation.

Closes https://github.com/quickwit-oss/quickwit/issues/2055

screenshot2

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexer.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexer.rs
@@ -20,7 +20,6 @@
 use std::collections::hash_map::Entry;
 use std::ops::RangeInclusive;
 use std::sync::Arc;
-use std::time::Instant;
 
 use anyhow::Context;
 use async_trait::async_trait;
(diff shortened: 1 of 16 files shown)


quickwit/quickwit-indexing/src/actors/indexer.rs
```diff
@@ -225,14 +231,11 @@ impl IndexerState {
...
+    span: Span,
+    _indexing_span: Span,
     indexed_splits: FnvHashMap<u64, IndexedSplitBuilder>,
     checkpoint_delta: IndexCheckpointDelta,
     publish_lock: PublishLock,
-    // TODO create this Instant on the source side to be more accurate.
-    // Right now this instant is used to compute time-to-search, but this
-    // does not include the amount of time a document could have been
-    // staying in the indexer queue or in the push api queue.
-    date_of_birth: Instant,
```
fulmicoton, line 235: We don't need `date_of_birth` since the top span of a indexing batch does that job.

guilload (approved):

Issue #2053, opened 2022-09-30
Depend on release version of Azure SDK


guilload:
We currently depend on a git rev.

PR #2057, opened 2022-10-03
Upgraded azure dependencies

evanxg852000:
### Description

Upgraded azure blob storage dependencies

Closes https://github.com/quickwit-oss/quickwit/issues/2053

### DIFF
--- a/quickwit/quickwit-storage/src/object_storage/azure_compatible_storage.rs
+++ b/quickwit/quickwit-storage/src/object_storage/azure_compatible_storage.rs
@@ -26,7 +26,7 @@ use std::{fmt, io};
 use async_trait::async_trait;
 use azure_core::error::ErrorKind;
 use azure_core::{Pageable, StatusCode};
-use azure_storage::core::prelude::*;
+use azure_storage::prelude::*;
 use azure_storage::Error as AzureError;
 use azure_storage_blobs::blob::operations::GetBlobResponse;
 use azure_storage_blobs::prelude::*;
@@ -86,8 +86,9 @@ impl fmt::Debug for AzureBlobStorage {
 impl AzureBlobStorage {
     /// Creates an object storage.
     pub fn new(account: &str, access_key: &str, uri: Uri, container: &str) -> Self {
+        let storage_credentials = StorageCredentials::access_key(account, access_key);
         let container_client =
-            StorageClient::new_access_key(account, access_key).container_client(container);
+            BlobServiceClient::new(account, storage_credentials).container_client(container);
         Self {
             container_client,
             uri,
@@ -116,7 +117,7 @@ impl AzureBlobStorage {
     /// Creates an emulated storage for testing.
     #[cfg(feature = "testsuite")]
     pub fn new_emulated(container: &str) -> Self {
-        let container_client = StorageClient::new_emulator_default().container_client(container);
+        let container_client = ClientBuilder::emulator().container_client(container);
         Self {
             container_client,
             uri: Uri::new(format!("azure://tester/{}", container)),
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-storage/src/object_storage/azure_compatible_storage.rs
```diff
@@ -302,9 +303,14 @@ impl Storage for AzureBlobStorage {
         let mut output_stream = self.container_client.blob_client(name).get().into_stream();
 
         let mut dest_file = File::create(output_path).await?;
-        while let Some(result) = output_stream.next().await {
-            let chunk = result.map_err(AzureErrorWrapper::from)?.data;
-            dest_file.write_all(&chunk).await?;
+        while let Some(chunk_result) = output_stream.next().await {
+            let chunk_response = chunk_result.map_err(AzureErrorWrapper::from)?;
+            let chuck_data = chunk_response
```
guilload, line 308: Do we have to collect here? Can we put a buf writer in front of the file and stream the data into the file directly? Or at least reuse the same vec between chunks?

quickwit/quickwit-storage/src/object_storage/azure_compatible_storage.rs
```diff
@@ -436,8 +442,8 @@ async fn download_all(
         .object_storage_download_num_bytes
         .clone();
     while let Some(chunk_result) = chunk_stream.next().await {
-        let chunk_response = chunk_result.map_err(AzureErrorWrapper::from)?;
-        let chuck_data = chunk_response.data;
+        let chunk_response = chunk_result?;
```
guilload, line 445: Same here.

guilload: Why do we do `let chunk_response = chunk_result.map_err(AzureErrorWrapper::from)?;` in one case and `let chunk_response = chunk_result?;` in another?

quickwit/quickwit-storage/src/object_storage/azure_compatible_storage.rs
```diff
@@ -436,11 +446,17 @@ async fn download_all(
...
-        object_storage_download_num_bytes.inc_by(chuck_data.len() as u64);
-        output.extend(chuck_data.as_ref());
+        let chunk_response = chunk_result?;
+        let stream_reader = chunk_response
+            .data
+            .map_err(|err| FutureError::new(FutureErrorKind::Other, err))
+            .into_async_read()
+            .compat();
+        let mut body_reader = BufReader::new(stream_reader);
+        let num_byte_copied = tokio::io::copy_buf(&mut body_reader, output).await?;
```
guilload, line 456:
```suggestion
        let num_bytes_copied = tokio::io::copy_buf(&mut body_reader, output).await?;
```

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -457,25 +457,23 @@ impl S3CompatibleObjectStorage {
...
     let object_storage_download_num_bytes = crate::STORAGE_METRICS
         .object_storage_download_num_bytes
         .clone();
-    while let Some(chunk_res) = byte_stream.next().await {
-        let chunk = chunk_res?;
-        object_storage_download_num_bytes.inc_by(chunk.len() as u64);
-        output.extend(chunk.as_ref());
-    }
+    let mut body_reader = BufReader::new(byte_stream.into_async_read());
+    let num_byte_copied = tokio::io::copy(&mut body_reader, output).await?;
```
guilload, line 475:
```suggestion
    let num_bytes_copied = tokio::io::copy(&mut body_reader, output).await?;
```

quickwit/quickwit-storage/src/object_storage/azure_compatible_storage.rs
evanxg852000: - In this case `let chunk_response = chunk_result.map_err(AzureErrorWrapper::from)?;` we need to propagate `StorageError`, and we cannot jump from `azure::Error` to `StrorageError` we need this map. 

In this case `chunk_response = chunk_result?;` we just need to propagate `AzureErrorWrapper`  

do you think removing it by providing an implementation from `azure::Error` to `StrorageError` makes it better?

guilload (approved):

Issue #2042, opened 2022-09-30
Can not create a text field with "index" is false while "stored" is true


geek-frio:
**Describe the bug**
Can not create a text field with "index" is false while "stored" is true.

**Steps to reproduce (if applicable)**

1. When I want to create index like this:
```
doc_mapping:
  field_mappings:
    - name: data_binary
      type: text 
      indexed: false
 ...
```
I only want to use text field to store a value without tokenizer while not index this field. 
2. Quickwit return an error, says: "doc_mapping.field_mappings: Error while parsing field `data_binary`: `record`, `tokenizer`, and `fieldnorms` parameters are allowed only if indexed is true. at line 11 column 5"

After I has reviewed quickwit's code, I find the problem may occured in the serde process,the QuickwitTextOptions's serde config is like this:

```rust
pub struct QuickwitTextOptions {
    #[serde(default)]
    #[serde(skip_serializing_if = "Option::is_none")]
    pub description: Option<String>,
    #[serde(default = "default_as_true")]
    pub indexed: bool,
    #[serde(default)]
    #[serde(skip_serializing_if = "Option::is_none")]
    pub tokenizer: Option<QuickwitTextTokenizer>,
    #[serde(default)]
    pub record: IndexRecordOption,
    #[serde(default = "default")]
    pub fieldnorms: bool,
    #[serde(default = "default_as_true")]
    pub stored: bool,
    #[serde(default)]
    pub fast: bool,
}
```

If I don't config record or fieldnorms, there default values are IndexRecordOption::Basic and false, so this logic forbids me to create a not indexed but can stored field.

**Expected behavior**
Can create a text not indexed field.

**Configuration:**

1. Output of `quickwit --version`

0.3.1

2. The index_config.yaml

```yaml
#
# Index config file for hdfs-logs dataset.
#

version: 0

index_id: segment_new

doc_mapping:
  field_mappings:
    - name: data_binary
      type: text 
      indexed: false
    - name: end_time
      type: i64
      indexed: true
      fast: true
    - name: endpoint_id
      type: text
      tokenizer: raw
      indexed: true
      fast: false
    - name: endpoint_name_match
      type: text
      indexed: true
      tokenizer: default
      fast: false
    - name: endpoint_name
      type: text
      indexed: true
      tokenizer: raw
      fast: false
    - name: is_error
      type: i64
      indexed: true
      #tokenizer: raw
      fast: false
    - name: latency
      type: i64
      indexed: true
[...]

PR #2075, opened 2022-10-06
Fix non-indexed text fields

nigel-andrews:
### Description

Changed the IndexRecordOption to an option that is now set to None from an IndexRecordOption::Basic

### How was this PR tested?

Cargo test to check for regression and successfully created the index defined in issue #2042 . Added some unit tests to ensure the correct behaviour


quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -156,7 +156,9 @@ fn validate_tag_fields(tag_fields: &[String], schema: &Schema) -> anyhow::Result
                     .get_indexing_options()
                     .map(|text_options| text_options.tokenizer());
 
-                if tokenizer_opt != Some(QuickwitTextTokenizer::Raw.get_name()) {
+                if tokenizer_opt != None
```
guilload, line 159:
```suggestion
                if tokenizer_opt.is_some()
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
```diff
@@ -258,8 +258,8 @@ fn deserialize_mapping_type(
             #[allow(clippy::collapsible_if)]
             if !text_options.indexed {
                 if text_options.tokenizer.is_some()
-                    || text_options.record == IndexRecordOption::Basic
-                    || !text_options.fieldnorms
+                    || text_options.record != IndexRecordOption::Basic
```
guilload, line 261: With this check, `indexed: false, record: basic` will not fail. It should. `text_options.record` should probably be an option.

guilload (commented):
This PR lacks unit testing.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
nigel-andrews: By default, indexed is set to true. If the IndexRecordOption is an option, what should the default value be since it can't be None ?

nigel-andrews: To be more accurate, at first I had already changed to IndexRecordOption to None, this caused issues since there was the possibility of having a text that could be indexed but there were no RecordOptions.

nigel-andrews: > `None`. If `indexed` is `false` and `record` is `Some(...)`, then that's an error. If `indexed` is true and `record` is `None`, then we use the default value `IndexRecordOption::Basic`.

I will update the code then.

fmassot: I have no preference on this so let's follow @guilload advice :)

As a side note, if we put an Option on record, why not put an option on fieldnorm and tokenizer too? (we don't need to do that now).

guilload: Yes, `tokenizer` and `record` should be options. `fieldnorms` is a boolean, so not really :)

nigel-andrews: `record` is now an option, `tokenizer` was already one so no changes needed there.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -156,7 +156,9 @@ fn validate_tag_fields(tag_fields: &[String], schema: &Schema) -> anyhow::Result
                     .get_indexing_options()
                     .map(|text_options| text_options.tokenizer());
 
-                if tokenizer_opt != Some(QuickwitTextTokenizer::Raw.get_name()) {
+                if tokenizer_opt.is_some()
```
guilload, line 159: Actually, isn't that a bug? We want the raw tokenizer in the presence of a tag field, but this seems to allow `None`.

guilload (commented):
Let's take the opportunity of this PR to add the same check for `QuickwitJsonOptions`.

nigel-andrews:
> Let's take the opportunity of this PR to add the same check for `QuickwitJsonOptions`.

Should `record` in `QuickwitJsonOptions` also be an Option then ?

fulmicoton:
Can you explain what was the problem exactly?

fulmicoton:
@nigel-andrews probably.
1- Add a similar unit test and tell us if it is necessary 
2- if it is not broken today, dig into it and explain why it is different... do we have a check missing etc.

fmassot (approved):
LGTM

PR #2079, opened 2022-10-06
Skip index metadata update when no mutation occurs

guilload:
### Description
Skip index metadata update when no mutation occurs
For fun and barely any profit.

### How was this PR tested?
`make test-all`


quickwit/quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
```diff
@@ -144,12 +144,12 @@ impl FileBackedMetastore {
     async fn mutate(
         &self,
         index_id: &str,
-        mutation: impl FnOnce(&mut FileBackedIndex) -> crate::MetastoreResult<bool>,
+        mutate_fn: impl FnOnce(&mut FileBackedIndex) -> crate::MetastoreResult<bool>,
     ) -> MetastoreResult<()> {
```
evanxg852000, line 148: I think we still need to propagate whether there was a mutation or not to the callers. `-> MetastoreResult<bool> `

evanxg852000 (commented):
I had already made these changes except for `checkpoint.rs` when applying your comments on https://github.com/quickwit-oss/quickwit/pull/2068

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
guilload: I'm not entirely sure what you intend to do with it, so I'll let you do it in your PR.

PR #2086, opened 2022-10-08
Fix #2085: supervisor gracefully exits when the last supervised actor…

### LINKED ISSUE #2085: Gracefull exit of supervisor when the supervised actor gracefully exit
An non-supervised actor will gracefully exit once all mailboxes are dropped (except for the the mailbox in the ActorContext).

But this mechanism does not work anymore if the actor is supervised as... its mailbox is also in its supervisor.

It could be nice to keep the same behavior for non-supervised and supervised actors.

### 2. Use a weak reference

I like 1.

fmassot:
… mailbox is dropped (except for the actor context).

Test added.


fulmicoton (approved):
As far as i know that should work but I am a bit scared tbh :)

This "last mailbox" feature has proved to be the largest source of bugs.

PR #2088, opened 2022-10-10
Issue/2022     Simplifies and fixes the local split cache.

### LINKED ISSUE #2022: Fix the split store
Is the split store broken?

fulmicoton:
It was tested via unit testing and by indexing wikipedia, and looking at the evolution of the datadir/cache directory.
`--keep-cache` was tested too.

After the fix, there is only one split cache shared
by all of the indexes.
    
Splits are all stored without any specific hierarchy in the
`data_dir/cache` directory.
    
The split cache evicts oldest splits first, and introduces
an extra constraint on the creation date to make sure
dangling splits eventually get removed even when the split is not under
capacity pressure.
    
---
    
    The Local Split Store is a local cache used to improve the performance of indexing nodes.
    Its purpose is simple: when a new split is freshly created, it is usely merged
    very rapidly after.
    
    In order to prevent this merge to force its download, we store it in the
    `LocalSplitStore`. This store is just a cache: a cache miss is acceptable and
    just means that the split will be redownloaded.
    
    The local split store eviction policy however, is rather uncommon.
    On our happy path, a split is stored into the cache, and is then used only once
    to undergo a merge.
    
    For this reason, we simply offer a way to `move splits into the cache`,
    and `move splits out of the cache`. A split is removed from the split store
    after its first access.
    
    Of course a failed merge could require accessing a given split more than once. In that
    case the split will be downloaded again.
    
    The cache size is limited by 3 things:
    - a maximum number of splits as defined in the `SplitStoreQuota`.
    - a maximum number of bytes as defined in the `SplitStoreQuota`.
    - finally, we evince older splits to make sure that newest split and the oldest
    split only differ by at most `SPLIT_MAX_AGE`.
    
    The point of this final rule invariant is to make sure that the disk space will be
    if the cache is NOT under pressure but some splits are actually useless.
    
    When adding a new split into the cache, if adding the split would break one of the following
    limit, we simply remove split one by one starting by the oldest first, until the split
    can be added.
    
    Closes #2022

### DIFF
--- a/quickwit/quickwit-cli/tests/cli.rs
+++ b/quickwit/quickwit-cli/tests/cli.rs
@@ -28,10 +28,9 @@ use anyhow::Result;
 use helpers::{TestEnv, TestStorageType};
 use predicates::prelude::*;
 use quickwit_cli::index::{create_index_cli, search_index, CreateIndexArgs, SearchIndexArgs};
+use quickwit_common::fs::get_cache_directory_path;
 use quickwit_common::rand::append_random_suffix;
 use quickwit_common::uri::Uri;
-use quickwit_config::CLI_INGEST_SOURCE_ID;
-use quickwit_core::get_cache_directory_path;
 use quickwit_indexing::actors::INDEXING_DIR_NAME;
 use quickwit_metastore::{quickwit_metastore_uri_resolver, Metastore};
 use serde_json::{json, Number, Value};
@@ -220,13 +219,8 @@ fn test_cmd_ingest_keep_cache() -> Result<()> {
         &test_env,
         "--keep-cache",
     );
-
     // Ensure cache directory is not empty.
-    let cache_directory_path = get_cache_directory_path(
-        &test_env.data_dir_path,
-        &test_env.index_id,
-        CLI_INGEST_SOURCE_ID,
-    );
+    let cache_directory_path = get_cache_directory_path(&test_env.data_dir_path);
     assert!(cache_directory_path.read_dir()?.next().is_some());
     Ok(())
 }
@@ -265,11 +259,8 @@ fn test_cmd_ingest_simple() -> Result<()> {
     .stdout(predicate::str::contains("Now, you can query the index"));
 
     // Ensure cache directory is empty.
-    let cache_directory_path = get_cache_directory_path(
-        &test_env.data_dir_path,
-        &test_env.index_id,
-        CLI_INGEST_SOURCE_ID,
-    );
+    let cache_directory_path = get_cache_directory_path(&test_env.data_dir_path);
+
     assert!(cache_directory_path.read_dir()?.next().is_none());
 
     Ok(())
@@ -823,6 +814,7 @@ async fn test_cmd_dry_run_delete_on_s3_localstack() -> Result<()> {
 #[tokio::test]
 #[serial]
 async fn test_all_local_index() -> Result<()> {
+    quickwit_common::setup_logging_for_tests();
     let index_id = append_random_suffix("test-all");
     let test_env = create_test_env(index_id, TestStorageType::LocalFileSystem)?;
     make_command(
@@ -841,7 +833,7 @@ async fn test_all_local_index() -> Result<()> {
         .exists(&Path::new(&test_env.index_id).join("metastore.json"))
         .await
         .unwrap();
-    assert_eq!(metadata_file_exists, true);
+    assert!(metadata_file_exists);
 
     ingest_docs(test_env.resource_files["logs"].as_path(), &test_env);
 
--- a/quickwit/quickwit-core/src/index.rs
+++ b/quickwit/quickwit-core/src/index.rs
@@ -18,15 +18,14 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::io;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
 
-use quickwit_common::fs::empty_dir;
+use quickwit_common::fs::{empty_dir, get_cache_directory_path};
 use quickwit_common::uri::Uri;
 use quickwit_config::{IndexConfig, QuickwitConfig};
 use quickwit_indexing::actors::INDEXING_DIR_NAME;
-use quickwit_indexing::models::CACHE;
 use quickwit_janitor::{
     delete_splits_with_files, run_garbage_collect, FileEntry, SplitDeletionError,
 };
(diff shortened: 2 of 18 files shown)


quickwit/quickwit-cli/tests/cli.rs
```diff
@@ -841,60 +833,61 @@ async fn test_all_local_index() -> Result<()> {
...
-        .as_str(),
-    )
-    .assert()
-    .success();
-    let metadata_file_exists = test_env
-        .storage
-        .exists(&Path::new(&test_env.index_id).join("quickwit.json"))
-        .await?;
-    assert_eq!(metadata_file_exists, false);
+    // let mut server_process = spawn_command(
```
fulmicoton, line 842: Need fix.

quickwit/quickwit-core/src/index.rs
```diff
@@ -302,27 +301,12 @@ impl IndexService {
     }
 }
 
-/// Helper function to get the cache path.
-pub fn get_cache_directory_path(data_dir_path: &Path, index_id: &str, source_id: &str) -> PathBuf {
```
fulmicoton, line 306: I moved this to quickwit_common

quickwit/quickwit-indexing/src/models/indexing_directory.rs
```diff
@@ -55,7 +53,6 @@ pub struct IndexingDirectory {
 
 struct InnerIndexingDirectory {
     root: Root,
-    pub cache_directory: PathBuf,
```
guilload, line 58: We can remove the indexing directory indirection in a follow-up PR since it no longer holds two directories but one.

quickwit/quickwit-indexing/src/split_store/indexing_split_store.rs
```diff
@@ -201,34 +88,30 @@ impl IndexingSplitStore {
...
         };
-        Ok(Self {
+        Self {
             inner: Arc::new(inner),
-        })
+        }
     }
 
+    /// Helper function to create a indexing split store for tests.
+    /// The resulting store does not have any local cache.
```
guilload, line 107: `:#[cfg(any(test, feature = "testsuite"))]`?

quickwit/quickwit-indexing/src/split_store/indexing_split_store.rs
```diff
@@ -281,92 +165,64 @@ impl IndexingSplitStore {
...
+        } else {
+            tracing::Span::current().record("cache_hit", false);
         }
-        let start_time = Instant::now();
         let dest_filepath = output_dir_path.join(&path);
-        info!(split_id = split_id, "fetch-split-from-remote-storage-start");
         self.inner
             .remote_storage
             .copy_to_file(&path, &dest_filepath)
+            .instrument(info_span!("fetch_split_from_remote_rstorage", path=?path))
```
guilload, line 221:
```suggestion
            .instrument(info_span!("fetch_split_from_remote_storage", path=?path))
```

quickwit/quickwit-indexing/src/split_store/local_split_store.rs
```diff
@@ -17,21 +17,60 @@
...
+// and `move splits out of the cache`. A split is removed from the split store
+// after its first access.
+//
+// Of course a failed merge could require accessing a given split more than once. In that
+// case the split will be downloaded again.
+//
+// The cache size is limited by 3 things:
+// - a maximum number of splits as defined in the `SplitStoreQuota`.
+// - a maximum number of bytes as defined in the `SplitStoreQuota`.
+// - finally, we evince older splits to make sure that newest split and the oldest
```
guilload, line 42:
```suggestion
// - finally, we evict older splits to make sure that the newest split and the oldest
```

quickwit/quickwit-indexing/src/split_store/split_store_quota.rs
```diff
@@ -0,0 +1,127 @@
...
+
+use byte_unit::Byte;
+use quickwit_config::IndexerConfig;
+use tracing::warn;
+
+/// A struct for keeping in check multiple SplitStore.
+#[derive(Debug)]
+pub struct SplitStoreQuota {
+    /// Current number of splits in the cache.
+    num_splits_in_cache: usize,
```
guilload, line 28: I was reading some code in tokio and they consistently use the word `capacity` to track use:
```
    num_splits_capacity: usize,
    num_bytes_capacity: Byte,
    num_splits_capacity: usize,
    max_num_bytes_capacity: Byte
```

guilload (approved):

quickwit/quickwit-indexing/src/models/indexing_directory.rs
fulmicoton: Created #2128 accordingly

quickwit/quickwit-indexing/src/split_store/split_store_quota.rs
fulmicoton: That's very poor naming in my opinion.
Capacity is a very good name that has a different meaning: how much stuff you can store without extra effort.

That's how it is used in real life and in the standard library.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -119,15 +115,22 @@ impl IndexingService {
...
         storage_resolver: StorageUriResolver,
         enable_ingest_api: bool,
-    ) -> IndexingService {
-        Self {
+    ) -> anyhow::Result<IndexingService> {
+        let split_store_space_quota = SplitStoreQuota::new(
+            indexer_config.split_store_max_num_splits,
+            indexer_config.split_store_max_num_bytes,
+        );
+        let split_cache_dir_path = get_cache_directory_path(&data_dir_path);
```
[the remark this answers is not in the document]
fulmicoton: I changed it to `<data dir>/cache/splits`.
That's probably safer indeed.

quickwit/quickwit-indexing/src/split_store/indexing_split_store.rs
fulmicoton: we can't really do this because it is also used in the janitor.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -256,14 +255,7 @@ impl IndexingService {
...
             crate::merge_policy::merge_policy_from_settings(&index_metadata.indexing_settings);
-        let split_store = self
-            .get_or_create_split_store(
-                &pipeline_id,
-                indexing_directory.cache_directory(),
-                storage.clone(),
-                merge_policy,
-            )
-            .await?;
+        let split_store = self.create_split_store(storage.clone(), merge_policy);
```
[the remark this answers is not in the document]
fulmicoton: done.

quickwit/quickwit-indexing/src/split_store/indexing_split_store.rs
guilload: So let's change the comment t_t

PR #2096, opened 2022-10-12
Removing binary from quickwit O_O

fulmicoton:


guilload:
Oh, that's why I was prompted to install `git-lfs` :D

PR #2097, opened 2022-10-12
Fix CLI inconsistencies #1320

### LINKED ISSUE #1320: [Papercut] Fix some CLI inconsistencies
- [ ] Remove `data-dir` option when not necessary
- [ ] Fix the display order of some commands (1. `config`, 2. `index`, 3. `source`/`split`, 4. other options)
- [ ] Make required options such as `config` or `index` actually required
- [ ] Set default values with `default_value`
- [ ] Add aliases (`ls`, `rm`, ...) whenever possible

kalil-pelissier:
### Description

Fix CLI inconsistencies:

- [ ] Remove data-dir option when not necessary
- [x] Fix the display order of some commands (1. config, 2. index, 3. source/split, 4. other options)
- [x] Make required options such as config or index actually required
- [ ] Set default values with default_value
- [x] Add aliases (ls, rm, ...) whenever possible

### How was this PR tested?

Try local CLI + run tests

### Content

- Fix the display order of some commands (1. config, 2. index, 3. source/split, 4. other options) 
- Add aliases (`ls` for list, `del` for delete, `desc` for describe)
- Remove data-dir when not necessary

close #1320

### DIFF
--- a/quickwit/quickwit-cli/src/cli.rs
+++ b/quickwit/quickwit-cli/src/cli.rs
@@ -35,7 +35,8 @@ pub fn build_cli<'a>() -> Command<'a> {
                 .help("Config file location")
                 .env("QW_CONFIG")
                 .default_value(DEFAULT_QW_CONFIG_PATH)
-                .global(true),
+                .global(true)
+                .display_order(1),
         )
         .subcommand(build_run_command().display_order(1))
         .subcommand(build_index_command().display_order(2))
--- a/quickwit/quickwit-cli/src/index.rs
+++ b/quickwit/quickwit-cli/src/index.rs
@@ -686,9 +686,7 @@ pub async fn list_index_cli(args: ListIndexesArgs) -> anyhow::Result<()> {
 }
 
 fn make_list_indexes_table<I>(indexes: I) -> Table
-where
-    I: IntoIterator<Item = IndexMetadata>,
-{
+where I: IntoIterator<Item = IndexMetadata> {
     let rows = indexes
         .into_iter()
         .map(|index| IndexRow {
--- a/quickwit/quickwit-cli/src/index.rs
+++ b/quickwit/quickwit-cli/src/index.rs
@@ -90,7 +90,8 @@ pub fn build_index_command<'a>() -> Command<'a> {
             Command::new("ingest")
                 .about("Indexes JSON documents read from a file or streamed from stdin.")
                 .args(&[
-                    arg!(--index <INDEX> "ID of the target index"),
+                    arg!(--index <INDEX> "ID of the target index")
+                        .display_order(1),
                     arg!(--"data-dir" <DATA_DIR> "Where data is persisted. Override data-dir defined in config file, default is `./qwdata`.")
                         .env("QW_DATA_DIR")
                         .required(false),
@@ -106,7 +107,8 @@ pub fn build_index_command<'a>() -> Command<'a> {
             Command::new("ingest-api")
                 .about("Enables/disables the ingest API of an index.")
                 .args(&[
-                    arg!(--index <INDEX> "ID of the target index"),
+                    arg!(--index <INDEX> "ID of the target index")
+                        .display_order(1),
                     arg!(--enable "Enables the ingest API.")
                         .required(true)
                         .conflicts_with("disable")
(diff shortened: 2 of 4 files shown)


quickwit/quickwit-cli/src/cli.rs
```diff
@@ -37,10 +37,10 @@ pub fn build_cli<'a>() -> Command<'a> {
...
                 .global(true),
         )
-        .subcommand(build_run_command().display_order(1))
-        .subcommand(build_index_command().display_order(2))
-        .subcommand(build_source_command().display_order(3))
-        .subcommand(build_split_command().display_order(4))
+        .subcommand(build_index_command().display_order(1))
+        .subcommand(build_source_command().display_order(2))
+        .subcommand(build_split_command().display_order(2))
+        .subcommand(build_run_command().display_order(3))
```
fulmicoton, line 43: `run`, `index`, `source`, `split` makes more sense.

fulmicoton:
@kalil-pelissier thank you Kalil.

I made the PR "draft" since it seems like it is a work in progress.
If you want to have it merged independently, that's even better. Just switch it to "Ready for Review", and edit the PR description with the actual content of the PR.

We require contributors to sign a CLA electronically. The process is rather fast.
You work is still under the AGPL license, but it gives Quickwit the right to also emit commercial licenses.

quickwit/quickwit-cli/src/cli.rs
kalil-pelissier: Ok, it was already this display order before my update. Maybe I didn't understand the task:

- `Fix the display order of some commands (1. config, 2. index, 3. source/split, 4. other options)` 

Is there any modifications to make for this task ?

fulmicoton: Ah well.... I guess I disagree with @guilload.

I'm ok with that order then.

fulmicoton: (apologies for the noise)

guilload: You changed `1, 2, 3, 4` to `1, 2, 2, 3`.

guilload: Regarding the order of the commands, I did not express myself correctly. I mean the order of the options in the commands: `1. config, 2. index, 3. checkpoint/source/split, 4. other options/flags`.

guilload:
`data_dir` is only necessary for commands such as `index ingest, index merge, index  search` that write intermediates data structure (index files, caches) `.

Let's no worry about setting default values for now.

kalil-pelissier:
@guilload I think I'm missing something. I search where `data_dir` is not used but each time this option is given to the `load_quickwit_config` function to generate a config:

```rust
    let quickwit_config = load_quickwit_config(&args.config_uri, args.data_dir).await?;
```

I don't think that I should remove the option in the subcommand and let a default value be set?

guilload:
`load_quickwit_config(&args.config_uri, None)` is fine. We already do that for a bunch of commands.

kalil-pelissier:
Ok so for example the `index create` command don't need `data-dir` option?

quickwit/quickwit-cli/src/index.rs
```diff
@@ -89,7 +89,9 @@ pub fn build_index_command<'a>() -> Command<'a> {
             Command::new("ingest")
                 .about("Indexes JSON documents read from a file or streamed from stdin.")
                 .args(&[
-                    arg!(--index <INDEX> "ID of the target index"),
+                    arg!(--index <INDEX> "ID of the target index")
+                        .display_order(1)
+                        .required(true),
```
evanxg852000, line 94: By default, arg is required so the pattern is more like use `required(false)` to make it optional. The exception I know is when you use `.conflicts_with()`

kalil-pelissier: Ok, I had doubts that args is required by default. 

So that means there is nothing to do for this task: `Make required options such as config or index actually required`.

I will remove the commit to get it back as it was.

quickwit/quickwit-cli/src/cli.rs
```diff
@@ -35,7 +35,8 @@ pub fn build_cli<'a>() -> Command<'a> {
                 .help("Config file location")
                 .env("QW_CONFIG")
                 .default_value(DEFAULT_QW_CONFIG_PATH)
-                .global(true),
+                .global(true)
+                .display_order(0),
```
guilload, line 39: +1

guilload (approved):

quickwit/quickwit-cli/src/cli.rs
kalil-pelissier: @guilload I just fixed the `display_order` but there is another trouble with the tests.
The option `data_dir` that I have deleted is required for tests so I removed my last commit.

guilload:
Merged, congrats on your first contribution @kalil-pelissier 🎉

PR #2104, opened 2022-10-14
Missing protect futures

fulmicoton:


fmassot (approved):

guilload (approved):
Idea that came to my mind: add a label to each protected zone and display a warning when an actor spends too much time in one.

fulmicoton:
[We can also rely on this. [****](https://doc.rust-lang.org/nightly/core/panic/struct.Location.html#method.caller)](https://doc.rust-lang.org/core/panic/struct.Location.html)

PR #2113, opened 2022-10-14
Add upload bytes to object storage metric.

fmassot:


quickwit/quickwit-storage/src/object_storage/azure_compatible_storage.rs
```diff
@@ -226,6 +230,10 @@ impl AzureBlobStorage {
             .map(|(num, range)| {
                 let moved_blob_client = blob_client.clone();
                 let moved_payload = payload.clone();
+                crate::STORAGE_METRICS.object_storage_put_parts.inc();
+                crate::STORAGE_METRICS
+                    .object_storage_upload_num_bytes
+                    .inc_by(range.end - range.start);
```
evanxg852000, line 236: Maybe shorter I think and less likely to have a one-off error.
```
crate::STORAGE_METRICS
                    .object_storage_upload_num_bytes
                    .inc_by(range.len());
```
However, I think it might be better to increment in the while loop further down since this is where the actual upload take place. increment when a chunk is uploaded.

evanxg852000 (approved):
Left a comment

quickwit/quickwit-storage/src/object_storage/azure_compatible_storage.rs
fmassot: Unfortunately, we don't have `range.len()` here. It would be nice to change that though. I saw that we were using `range.count()` which seems weird, the code is like this: `let mut buf: Vec<u8> = Vec::with_capacity(range.count());`

PR #2116, opened 2022-10-15
Fix timestamp reader in delete executor.

fmassot:
Test updated to check it's fixed.


quickwit/quickwit-common/Cargo.toml
```diff
@@ -22,6 +23,7 @@ prometheus = { workspace = true }
 rand = { workspace = true }
 regex = { workspace = true }
 serde = { workspace = true }
+tantivy = { workspace = true }
```
guilload, line 26: I'm really bummed that `fastfield_codecs` and `tantivy` were added as `quickwit-common` dependencies. Now, sub-crates that have nothing to do with `tantivy` pull those dependencies transitively. It really sucks when you do a `cargo build/test -p quickwit-<...>`, which I do a lot when developing to avoid compiling the whole project.

fmassot: Yes, I was almost sure you will be annoyed by that :/
I'm opening an issue to move that, maybe in the doc mapper.

PR #2117, opened 2022-10-15
Fix merge_num_ops set on a split after a delete operation.

fmassot:
Test updated to check if num_merge_ops is unchanged.

Two `record_progress` were added too as I got some actor timeout error during my tests.

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/merge_executor.rs
+++ b/quickwit/quickwit-indexing/src/actors/merge_executor.rs
@@ -651,8 +651,8 @@ mod tests {
         )
         .await?;
         let docs = vec![
-            serde_json::json!({"body": "info", "ts": 0 }),
-            serde_json::json!({"body": "delete", "ts": 0 }),
+            serde_json::json!({"body": "info", "ts": 1624928208 }),
+            serde_json::json!({"body": "delete", "ts": 1634928208 }),
         ];
         test_sandbox.add_documents(docs).await?;
         let metastore = test_sandbox.metastore();
--- a/quickwit/quickwit-indexing/src/actors/merge_executor.rs
+++ b/quickwit/quickwit-indexing/src/actors/merge_executor.rs
@@ -99,9 +99,16 @@ impl Handler<MergeScratch> for MergeExecutor {
                 .await?,
             ),
             MergeOperationType::DeleteAndMerge => {
+                assert_eq!(
+                    merge_op.splits.len(),
+                    1,
+                    "Delete tasks can be applied only on one split."
+                );
+                assert_eq!(merge_scratch.tantivy_dirs.len(), 1);
+                let split_with_docs_to_delete = merge_op.splits.into_iter().next().unwrap();
                 self.process_delete_and_merge(
                     merge_op.merge_split_id.clone(),
-                    merge_op.splits.clone(),
+                    split_with_docs_to_delete,
                     merge_scratch.tantivy_dirs,
                     merge_scratch.merge_scratch_directory,
                     ctx,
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-indexing/src/actors/merge_executor.rs
```diff
@@ -376,8 +376,9 @@ impl MergeExecutor {
             "Delete tasks can be applied only on one split."
         );
         assert_eq!(tantivy_dirs.len(), 1);
-        let split = splits.pop().unwrap();
+        let split = &splits[0];
         let index_metadata = self.metastore.index_metadata(&split.index_id).await?;
+        ctx.record_progress();
```
guilload, line 381: It seems you want to protect futures here instead. `protect_future` should call `record_progress` after the future returns.

fmassot: oh nice, I did not know that.

guilload: I'm not saying it does call `record_progress` but it probably should.

fmassot: Ok. For a simple call like `index_metadata`, I'm not fond of using `protect_future` here, we generally reserve it for bigger requests.

The issue is always the same with the actor timeout... we are in the dark and it's not always easy to guess which piece of code is responsible for the timeout.

guilload: if you think you don't need `protect_future` then you don't need `record_progress` :D

fmassot: I don't know... I need to rerun some deletes. Let's remove them for now and see what's happening.

fmassot: I've got 2 calls to the metastore, that's probably too much. Having one record progress between the two seems fair.

fmassot: I removed them. Let's wait and see. At least now these actors are supervised.

quickwit/quickwit-indexing/src/actors/merge_executor.rs
```diff
@@ -397,6 +398,7 @@ impl MergeExecutor {
             .metastore
             .list_delete_tasks(&split.index_id, split.delete_opstamp)
             .await?;
+        ctx.record_progress();
```
fmassot, line 401: this is not a `record_progress`.

quickwit/quickwit-indexing/src/actors/merge_executor.rs
```diff
@@ -386,8 +386,9 @@ impl MergeExecutor {
             "Delete tasks can be applied only on one split."
         );
         assert_eq!(tantivy_dirs.len(), 1);
-        let split = splits.pop().unwrap();
+        let split = &splits[0];
         let index_metadata = self.metastore.index_metadata(&split.index_id).await?;
+        ctx.record_progress();
```
fmassot, line 391: *mind trick*

there is no `record_progress` here.

quickwit/quickwit-indexing/src/actors/merge_executor.rs
```diff
@@ -375,7 +375,7 @@ impl MergeExecutor {
     async fn process_delete_and_merge(
         &mut self,
         merge_split_id: String,
-        mut splits: Vec<SplitMetadata>,
+        splits: Vec<SplitMetadata>,
```
fulmicoton, line 378: If we only accept splits.len() == 1, maybe we would be better off calling this `process_delete_and_compact` and passing  `split: SplitMetadata` as argument.

fulmicoton: (this bug would have been avoided.)

fulmicoton (approved):

quickwit/quickwit-indexing/src/actors/merge_executor.rs
fulmicoton: you probably want to protect_future the call above instead.

PR #2120, opened 2022-10-15
Improve fetch of stale splits.

fmassot:
Currently, it's very easy to not get any stale splits if there are more than 100 immature splits... which can happen very easily.

This PR tries to mitigate the issue by:
- fetch 1000 splits
- order splits also by publishing date (it was already ordered by delete_opstamp) to get old split first

I put a longer polling loop on the delete task planner to avoid sending the same operations twice or even more.

### DIFF
--- a/quickwit/quickwit-metastore/src/tests.rs
+++ b/quickwit/quickwit-metastore/src/tests.rs
@@ -2443,8 +2443,8 @@ pub mod test_suite {
                 .publish_splits(index_id, &[split_id_4], &[], None)
                 .await
                 .unwrap();
-            // Sleep 1 millisecond to have different publish timestamp.
-            tokio::time::sleep(Duration::from_millis(1000)).await;
+            // Sleep for 1 second to have different publish timestamps.
+            tokio::time::sleep(Duration::from_secs(1)).await;
             metastore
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-metastore/src/tests.rs
```diff
@@ -2417,32 +2427,42 @@ pub mod test_suite {
...
                 .unwrap();
+            metastore
+                .stage_split(index_id, split_metadata_4.clone())
+                .await
+                .unwrap();
+            metastore
+                .publish_splits(index_id, &[split_id_4], &[], None)
+                .await
+                .unwrap();
+            // Sleep 1 millisecond to have different publish timestamp.
```
guilload, line 2446:
```suggestion
            // Sleep for 1 second to have different publish timestamps.
```

quickwit/quickwit-metastore/src/tests.rs
```diff
@@ -2417,32 +2427,42 @@ pub mod test_suite {
...
+            tokio::time::sleep(Duration::from_millis(1000)).await;
```
guilload, line 2447:
```suggestion
            tokio::time::sleep(Duration::from_secs(1)).await;
```

PR #2125, opened 2022-10-16
Add metrics to measure backpressure in indexing pipeline.

fmassot:
Just a draft to play with metrics and quickly have info about the time spent to send a message to an actor mailbox.


quickwit/quickwit-indexing/src/lib.rs
```diff
@@ -98,3 +104,43 @@ pub async fn start_indexing_service(
...
+            Poll::Ready(output) => output,
+        };
+        if let Some(histogram_timer) = this.histogram_timer.take() {
+            histogram_timer.observe_duration();
+        };
+        Poll::Ready(output)
+    }
+}
+
+impl<T: Sized> InstrumentMetric for T {}
```
fulmicoton, line 146: If it compiles, it might be worth making it 

```suggestion
impl<T: Future + Sized> InstrumentMetric for T {}
```

Hopefully it might prevent silly autocomplete on all objects.

quickwit/quickwit-indexing/src/lib.rs
```diff
@@ -98,3 +104,43 @@ pub async fn start_indexing_service(
...
+        #[pin]
+        histogram_timer: Option<HistogramTimer>,
+    }
+}
+
+impl<T: Future> Future for InstrumentedMetric<T> {
+    type Output = T::Output;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut this = self.project();
```
fulmicoton, line 134: gg. 

I always struggle on doing stuff like that!

fulmicoton (changes requested):
There is bit of schizophrenia in this PR.

1) The instrumenting code, makes it possible to instrument any future.
2) It target a metrics that is documented as "time to process a message" which is an incorrect description everywhere (I think)
3) The single metric it targets has a single label key, yet we pass it an array of label values.

I'm not sure on which side it should lean, so I could not suggest the change.

For instance, "waiting time for an async operation", label_key=operation, etc.

fmassot:
Yes, this was just an experimentation... I updated the code to pass a metric to the function... but I don't really like that :/

Currently, I think we want to monitor:
- the time spent waiting to send a message to an actor mailbox (this is the backpressure) 
- the time spent processing a message. Ideally we would need a proc macro but it seems quite hard to implement that on a async function so I would go for a manual implementation for the main function we want to monitor closely.

fmassot:
Let's close this for now as these metrics are not super useful.

PR #2141, opened 2022-10-19
Bugfix: Replacing .progress() where .record_progress() was meant

fulmicoton:
.progress() is quite confusing. It just returns a `Progress` object. I added a `must_use` attribute to prevent misuse in the future.


fmassot (approved):
WT… LGTM

PR #2153, opened 2022-10-20
QW-2145: Improve error message when an IO error occurs when fetching metadata.

ChillFish8:
This improves the error message/handling when we are going through the list of splits in the cache directory and the system cannot read the file's metadata due to an IO error.

Originally this was implicitly handled by `is_dir()` using `unwrap_or(false)` which caused some confusing errors in the event an IO error did occur.

The change effectively just returns the IO error if there is one so that it can be correctly populated and displayed.


quickwit/quickwit-indexing/src/split_store/local_split_store.rs
```diff
@@ -134,12 +134,19 @@ impl SplitFolder {
     }
 }
 
-fn split_id_from_split_folder(dir_path: &Path) -> Option<&str> {
-    if !dir_path.is_dir() {
-        return None;
+fn split_id_from_split_folder(dir_path: &Path) -> io::Result<Option<&str>> {
+    let metadata = dir_path.metadata()?;
+
+    if !metadata.is_dir() {
```
guilload, line 140: Now, we're going to throw an error about the file name when the issue is that we're dealing with a file. I'd deal with the path not found or not a dir outside of this function:
```rust
for dir_entry in ... {
  let dir_path: PathBuf = dir_entry.path();
  let metadata = tokio::fs::metadata(&dir_path).await?;
  
  if metadata.is_file() {
    warn!("Unexpected file found in split cache directory: `{}`.", dir_path.display());
    continue;
  }
  let split_id = split_id_from_split_folder(&dir_path).ok_or_else(|| ...)?;
}

File: quickwit/quickwit-indexing/src/split_store/local_split_store.rs [hunk 2]
```diff
@@ -307,7 +314,7 @@ impl LocalSplitStore {
         for dir_entry_result in fs::read_dir(&split_store_folder)? {
```
guilload, line 314: Can we switch to the async version of `read_dir` while we're at it?

quickwit/quickwit-indexing/src/split_store/local_split_store.rs
ChillFish8: Sounds good to me, have moved it out

quickwit/quickwit-indexing/src/split_store/local_split_store.rs
```diff
@@ -135,11 +135,7 @@ impl SplitFolder {
 }
 
 fn split_id_from_split_folder(dir_path: &Path) -> Option<&str> {
-    if !dir_path.is_dir() {
-        return None;
-    }
-    let split_id: &str = dir_path.file_name()?.to_str()?.strip_suffix(".split")?;
-    Some(split_id)
+    dir_path.file_name()?.to_str()?.strip_suffix(".split")
 }
```
ChillFish8, line 139: Im tempted to move this out of the function now, but it equally doesn't do any harm to be a bit more verbose I guess.

guilload (approved):

Issue #2126, opened 2022-10-16
Improve GC performance


guilload:
Splits sometimes need to be deleted:
- because they were replaced after a merge.
- because they were uploaded, but some failure prevented their publication. (In that case they are in the Staged state)
- because they only contain documents out of the retention period.

We do never delete splits right away, to avoid interferring with in-flight queries. Instead we change their state to MarkForDelete. The GC then periodically delete the splits that have been in the MarkForDelete splits for more than a given grace period.

Delete here means "delete them from the storage", then "delete them from the metastore".
The current GC is pretty much stateless. Periodically it wakes up, and queries the metastore. 

On a large volume (50 indexes) with partitioning, our current approach to GC has proved to be too inefficient.

There is a lot of room for improvement for our GC:
- push down the predicates that allow computing the list of splits candidate for deletion, add indexes, `SELECT split_id` etc.
- `LIMIT 100`?
- incremental GC
- batch delete request at the storage level

Also it would be very helpful to improve our observability.

fulmicoton:
> Adding a limit and trying to incrementally run the deletes presents a difficult problem around how you actually catch up to the rate of deletes being produced, situations where you can catch up tend to not have the original performance issue in the first place although it probably wouldn't hurt to have, it doesn't solve the main issue which is the effects on other services when trying to keep up with a large number of deletes being produced at high load.

Right now the GC loop look like 

```
loop {
  gc()
  wait(duration)
}
```

it could be something like 
```
loop {
     let splits_to_gc_possibly_remaining = gc_with_limit();
     if !splits_to_gc_possibly_remaining {
         wait(duration)
     }
  }
}
```

guilload:
A few things regarding optimizing our schema/indexes/triggers:
- there is a trigger that updates the index `update_timestamp` each time we update/delete a split; that might be the culprit.
- PostgreSQL supports partial indexes, we can add partial index on `split_state` for the `Staged` and `MarkForDeletion` states
- the `splits` table has no index on `index_id`; that's probably a must-have for deployments with many indexes.

ChillFish8:
I've attempted both but neither seems to convince the query planner that it's worth using when working with index (maybe we can convince it if there's a high number? Will try that next) 😞 

```sql
delete from splits where split_state = 'MarkedForDeletion';
```

produces the following explanation, seemingly with and without the index being present:
[...]

PR #2157, opened 2022-10-21
QW-2126: Relax trigger condition from ROW to STATEMENT

ChillFish8:
Changes the trigger condition from running for every row to running once per statement.

This has roughly the same effect as removing the trigger and manually implementing the update logic. Still, IMO this is more maintainable as it prevents the update call from being missed if the code which interacts with the splits table changes.

Related to #2126


ChillFish8:
This is a breaking change to the metastore because we're modifying the existing migration. I'm under the impression that we can make this breaking change as the next version is breaking.  Otherwise we'll need to create another migration.

PR #2158, opened 2022-10-21
Refactor `test_cmd_ingest_on_non_existing_file`

boraarslan:
It did not make sense to me that we exit early from the process when the running checklist failed. (It was also the only place we exit from the process directly outside the main function.) To properly (well, kind of) check the return type, I added a new error.


quickwit/quickwit-common/src/checklist.rs
```diff
@@ -84,12 +88,47 @@ pub fn print_checklist(check_list_results: &[(&str, anyhow::Result<()>)]) {
...
+
+    Ok(())
+}
+
+#[derive(Error, Debug)]
+pub struct ChecklistError {
+    pub errors: Vec<(String, anyhow::Result<()>)>,
+}
+
+impl From<Vec<(&str, anyhow::Result<()>)>> for ChecklistError {
```
guilload, line 108: I don't think we need that `From`. Let's use a proper constructor in `ChecklistError`. Maybe `ChecklistError::from_errors`?

quickwit/quickwit-common/src/checklist.rs
```diff
@@ -84,12 +88,47 @@ pub fn print_checklist(check_list_results: &[(&str, anyhow::Result<()>)]) {
...
+            .into_iter()
+            .filter(|(_, check_res)| check_res.is_err())
+            .map(|(check_elem, check_res)| (check_elem.to_string(), check_res))
+            .collect();
+        ChecklistError { errors }
+    }
+}
+
+impl Display for ChecklistError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
guilload, line 120:
```suggestion
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```

quickwit/quickwit-common/src/checklist.rs
```diff
@@ -84,12 +88,47 @@ pub fn print_checklist(check_list_results: &[(&str, anyhow::Result<()>)]) {
...
+        let err_string = self
+            .errors
+            .iter()
+            .map(|(check_item, check_item_err)| {
+                format!(
+                    "\n{}: {}",
+                    check_item,
+                    check_item_err.as_ref().err().unwrap()
```
guilload, line 128: Use `expect` and tell me why it should work :)

guilload (approved):
Slowly but surely making this test suite much better. Thank you!

PR #2159, opened 2022-10-21
QW-2126: Remove metastore trigger and replace with helper function.

ChillFish8:
As discussed in #2126 the Postgres trigger seems to have a significant performance impact when deleting rows from the splits table due to it running another `UPDATE` operation on the `indexes` table for every row that gets deleted.

This removes the trigger and instead has a helper function that gets ran when ever we call a method that modifies the splits table (this is not done implicitly though)
another potential solution was changing from `EACH ROW` to `EACH STATEMENT` but unfortunately, we lose the context of the query itself so cannot use this to update the index efficient and implicitly.


quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -172,10 +172,32 @@ async fn mark_splits_as_published_helper(
     .map(|row| row.get(0))
     .fetch_all(tx)
     .await?;
+
```
guilload, line 175: That's just noise.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -172,10 +172,32 @@ async fn mark_splits_as_published_helper(
...
     Ok(published_split_ids)
 }
 
-/// Marks mutiple splits for deletion.
+#[instrument(skip(tx))]
+/// Updates the given index's `update_timestamp` to the current
+/// timestamp.
+async fn update_index_last_update_ts(
```
guilload, line 182: `touch_index`?

guilload (approved):

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
ChillFish8: I think `touch_index` is a bit vague about what it actually does, but I can't think of a better name for it :/

guilload: Then, `update_index_update_timestamp` or `update_index_update_ts` is the most accurate.

guilload (approved):

PR #2161, opened 2022-10-24
The index serializer should set its progress/killswitch on the contro…

fulmicoton:
The index serializer should set its progress/killswitch on the controlled directory
    
    This logic existed on the Packager, and should have been moved
    when we moved the serialization to the IndexSerializer.


evanxg852000 (commented):
Some changes from the previous commit in the main branch crept here. Maybe a git reset & rebase can clear those changes from your commit.

PR #2163, opened 2022-10-24
Option IO throttling of the merger.

fulmicoton:


quickwit/quickwit-indexing/src/actors/merge_executor.rs
```diff
@@ -233,6 +234,7 @@ async fn merge_split_directories(
         Box::new(MmapDirectory::open(output_path)?),
         ctx.progress().clone(),
         ctx.kill_switch().clone(),
+        Some(Byte::from_bytes(20_000_000)),
```
fulmicoton, line 237: need to fix

quickwit/quickwit-indexing/src/throttle.rs
```diff
@@ -0,0 +1,242 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::{Arc, Weak};
+use std::time::{Duration, Instant};
+
+use quickwit_common::metrics::IntCounter;
+pub struct Throttle {
```
guilload, line 25: We might not need a custom implementation: the `governor` rate limiter wraps an `AtomicU64`, so cloneable with an `Arc` and has sync (`check*`) and async `until*` APIs.

fulmicoton: One thing I am scared with governor is that 
a) the docs only talks about requests / operations etc. and does not talk about using it to throttle a throughput
b) the logic that throttles is a loop that have tasks wait for the capacity to be available. I am not fond of that:
I don't know what limits the number of times a thread will have to wake up, I am not sure if there can be some wait of time here.
c) I don't understand how it works so I am not able to reason with it:
For instance, if I have two writers, sharing a rate limiter, and being throttled. One is writing 1kB at a time, when the other is doing it 10kb at a time, what throughput do they achieve?

fulmicoton: created an issue on github: https://github.com/antifuchs/governor/issues/154

fulmicoton: created an issue on github: https://github.com/antifuchs/governor/issues/154

fulmicoton: https://github.com/tikv/async-speed-limit seems to do exactly what we need and more (can wrap async read/write directly).

fulmicoton:
Switching to draft, will replace with async-speed-limit

fulmicoton:
Closed for #2205

PR #2166, opened 2022-10-24
Refactor `test_cmd_search`

boraarslan:


quickwit/quickwit-cli/tests/helpers.rs
```diff
@@ -101,16 +101,6 @@ pub fn make_command(arguments: &str) -> Command {
     cmd
 }
 
-pub fn make_command_with_list_of_args(arguments: &[&str]) -> Command {
```
guilload, line 104: We still want to disable telemetry while running the tests.

boraarslan: I think telemetry is only enabled when you run the CLI command (since the telemetry loop is only started on main()). When we don't execute the CLI and just call the underlying fn, I don't think it sends any telemetry data.

quickwit/quickwit-cli/tests/cli.rs
```diff
@@ -404,67 +403,72 @@ async fn test_cmd_search_with_snippets() -> Result<()> {
     Ok(())
 }
 
-#[test]
-fn test_cmd_search() -> Result<()> {
+#[tokio::test]
+async fn test_cmd_search() {
```
guilload, line 407:
```suggestion
async fn test_search_index_cli() {
```

quickwit/quickwit-cli/tests/cli.rs
```diff
@@ -404,67 +403,72 @@ async fn test_cmd_search_with_snippets() -> Result<()> {
...
-        .as_str(),
-    )
-    .assert()
-    .success()
-    .stdout(predicate::function(|output: &[u8]| {
-        println!("{}", from_utf8(output).unwrap());
-        let result: Value = serde_json::from_slice(output).unwrap();
-        result["num_hits"] == Value::Number(Number::from(2i64))
-    }));
+    let args = SearchIndexArgs {
```
guilload, line 414: Looks like `SearchIndexArgs` could use a `Default` Impl.

guilload (approved):

PR #2167, opened 2022-10-24
Refactor `test_cmd_dry_run_delete_on_s3_localstack`

boraarslan:


guilload:
There's already a test on dry run delete and delete for an index on file storage, right?

boraarslan:
Yes there are

guilload:
We can remove this test then.

guilload (approved):

PR #2189, opened 2022-10-26
remove dependancy on RocksDB

### LINKED ISSUE #2073: Remove RocksDB dependency
A Quickwit node is meant to support thousands of indexes, while maintaining a decent throughput, and small memory footprint.

Currently the Ingest API works by using RocksDB as our storage.
Each queue gets its own column family.

When we publish a split, we truncate it. RocksDB allows efficient truncation. If a SSTable is entirely deleted, the truncation will just delete the file instead of using tombstone & compaction. In my experience it is however unable to truncate the WAL, so the current solution ends up creation 1000+ files.

Also the rocksdb dependency is making building portable binaries very difficult, and the build is much slower because of it.

The following project aims at replacing the current RocksDB-based solution.
https://github.com/quickwit-oss/mrecordlog/tree/main/src

#1898 shows what a naive replacement could look like.

trinity-1686a:
fix #2073

some of the features we used to have can't be reproduced as is without some more changes on mrecordlog side


quickwit/quickwit-ingest-api/src/errors.rs
```diff
@@ -85,10 +81,66 @@ impl From<IngestApiError> for tonic::Status {
...
+                IngestApiError::InvalidPosition("appending record in the future".to_owned())
+            }
+        }
+    }
+}
+
+impl From<DeleteQueueError> for IngestApiError {
+    fn from(e: DeleteQueueError) -> IngestApiError {
+        match e {
+            DeleteQueueError::IoError(ioe) => ioe.into(),
```
fulmicoton, line 121:
```suggestion
            DeleteQueueError::IoError(io_err) => io_err.into(),
```

quickwit/quickwit-ingest-api/src/errors.rs
```diff
@@ -85,10 +81,66 @@ impl From<IngestApiError> for tonic::Status {
...
+            ReadRecordError::IoError(ioe) => ioe.into(),
+            ReadRecordError::Corruption => IngestApiError::Corruption {
+                msg: "failed to read record".to_owned(),
+            },
+        }
+    }
+}
+
+impl From<AppendError> for IngestApiError {
+    fn from(e: AppendError) -> IngestApiError {
```
fulmicoton, line 103:
```suggestion
    fn from(err: AppendError) -> IngestApiError {
```

quickwit/quickwit-ingest-api/src/ingest_api_service.rs
```diff
@@ -96,7 +93,9 @@ impl IngestApiService {
             // TODO better error handling.
             // If there is an error, we probably want a transactional behavior.
             let records_it = iter_doc_payloads(doc_batch);
-            self.queues.append_batch(&doc_batch.index_id, records_it)?;
+            self.queues
```
fulmicoton, line 96: This implementation is properly asynchronous.
We can now have that actor run on the regular runtime, and use the `ctx.protect_future` mechanism to guard us from waiting for the runtime, or io wait.

fulmicoton (changes requested):
See comment inline.

quickwit/quickwit-ingest-api/src/ingest_api_service.rs
fulmicoton: @trinity you need protect_future on this

fulmicoton (approved):

PR #2193, opened 2022-10-27
Fix coverage

boraarslan:


guilload:
Let's just install `libsasl2-dev` and `libsasl2-2` for now and then switch to the builder image in another PR.

boraarslan:
Coverage runs but tests are broken

PR #2196, opened 2022-10-27
Publish docker image on tag happy-plazza.

fmassot:


.github/workflows/publish_docker_images.yml
```diff
@@ -31,6 +32,7 @@ jobs:
             type=edge,branch=main,suffix=-bullseye
             type=semver,pattern={{version}}
             type=semver,pattern={{version}},suffix=-bullseye
+            type=ref,event=tag
```
guilload, line 35: Just be aware that if you add a `happy-plazza` git tag that is not at the tip of `main`, we're (temporarily) moving the `edge` Docker tag back.

fmassot: Mmm, yes I'm discovering the metadata action. We can avoid that with

```
enable=${{ !startsWith(github.ref, 'refs/tags') }}
```

`edge` tag should be triggered only on a commit push on main.

fmassot: I'm quite new on this, what do you think?

guilload: I'm rereading the doc of the metadata action, and it might be smart enough to push an edge tag only for the last commit. Let's try this as is.

guilload (approved):

PR #2198, opened 2022-10-30
Add metrics to monitor merge and indexer written bytes.

fmassot:


quickwit/quickwit-indexing/src/controlled_directory.rs
```diff
@@ -131,21 +139,6 @@ impl io::Write for ControlledWrite {
         let _guard = self.check_if_alive();
         self.underlying_wrt.flush()
     }
-
-    fn write_vectored(&mut self, bufs: &[io::IoSlice<'_>]) -> io::Result<usize> {
```
fmassot, line 135: Seems that we don't need that. `write_vectored`, `write_all` and `write_fmt` will end up calling `write`

fulmicoton:
I add these metrics in #2205.

fmassot:
ok perfect, closing now.

PR #2225, opened 2022-11-02
Remove `#[serial]` from CLI tests

boraarslan:


boraarslan:
CI passes, but I'm not convinced this will always work. `find_available_tcp_port()` locks the socket temporarily to get another port from OS. So potentially, REST servers in the tests might try to bind to address and fail while `find_available_tcp_port()` is running.

quickwit/quickwit-cli/tests/helpers.rs
```diff
@@ -212,14 +219,16 @@ pub fn create_test_env(index_id: String, storage_type: TestStorageType) -> anyho
             .replace("index_uri: #index_uri\n", ""),
     )?;
     let quickwit_config_path = resources_dir_path.join("config.yaml");
-    let rest_listen_port = find_available_tcp_port()?;
+    let rest_listen_port = get_random_port();
```
guilload, line 222: Why not call `find_available_tcp_port` twice?

guilload (approved):

PR #2237, opened 2022-11-03
QW-2209: Bulk delete splits in chunks.

### LINKED ISSUE #2209: Use bulk delete in GC
The GC currently deletes files one by one. With the introduction of bulk deletes, we can speed up this process.

ChillFish8:
Closes #2209 

This moves from the use of the `Storage::delete` method to the `Stroage::bulk_delete` method and batches a set of splits to be deleted, into chunks, this should significantly reduce the number of delete requests being sent to S3 improving performance.


quickwit/quickwit-janitor/src/garbage_collection.rs
```diff
@@ -21,25 +21,29 @@ use std::path::Path;
...
 use time::OffsetDateTime;
+use tokio::sync::{mpsc, Semaphore};
 use tracing::error;
 
 use crate::actors::GarbageCollector;
 
 const MAX_CONCURRENT_STORAGE_REQUESTS: usize = if cfg!(test) { 2 } else { 10 };
 
+/// The number of split delete requests to combine and issue as a single bulk request.
+const BULK_DELETE_CHUNK_SIZE: usize = 10;
```
ChillFish8, line 39: Currently, the chunk size is set to `10` which is pretty arbitrary, in S3 the maximum number per request is `1000` but this may be different for each blob store and is also probably overkill. I would say some number between 10 - 100 is sufficient but it will want some load testing to really know how much of a difference it's actively making.

Another option is we make this user configurable, but that adds an additional set of hoops for us to jump through and is probably not critical enough to be worth exposing to the user.

quickwit/quickwit-janitor/src/garbage_collection.rs
```diff
@@ -186,31 +190,16 @@ pub async fn delete_splits_with_files(
...
-                let split_filename = quickwit_common::split_file(split.split_id());
-                let split_path = Path::new(&split_filename);
-                let delete_result = moved_storage.delete(split_path).await;
-                if let Some(ctx) = ctx_opt {
-                    ctx.record_progress();
-                }
-                (split.split_id().to_string(), file_entry, delete_result)
-            }
-        })
-        .buffer_unordered(MAX_CONCURRENT_STORAGE_REQUESTS);
```
ChillFish8, line 203: Unfortunately, I had to replace this with a more basic task + channel approach as I was unable to convince the compiler about higher order lifetimes which was causing a very hard-to-debug error. I'm not sure why it works with the channel approach and not tokio_stream.

quickwit/quickwit-janitor/src/garbage_collection.rs
```diff
@@ -186,30 +178,60 @@ pub async fn delete_splits_with_files(
...
+            for (split_id, entry) in paths_to_splits.into_values() {
+                deleted_split_ids.push(split_id);
+                deleted_file_entries.push(entry);
+            }
+        }
+        Err(state) => {
+            let split_ids = state
+                .failures
+                .keys()
+                .chain(state.unattempted.iter())
```
guilload, line 218: Truncate to a few splits to avoid flooding the logs and collect into a string directly `Itertools::join`. Since we're truncating, display how many splits could not be deleted.

ChillFish8: I think truncating and joining a string could potentially be misleading or at the very least not be the most helpful way of displaying that information if we want to use the tracing values. I'd probably opt for just listing how many splits failed to be deleted by itself or we keep all of them displayed. 

If we do still want to display only the first few split ids, I would probably opt for displaying them as such:

```rust
            error!(
                error = ?bulk_delete_error.error, 
                index_id = ?index_id,
                num_split_ids = num_split_ids, 
                "Failed to {:?} and {} other splits.",
                truncated_splits_list, 
                num_split_ids
            );
```

The reason is that I don't think having a string value of `split_ids_sample = "phiu4h23423,uygeuyfg234,2g23uy4g2fgd"` is any more helpful or useful than displaying just a ` Vec<Path>` (which will also play nicer if we ever wanted to use tracing's ability to display logs in a JSON format or any of the other formatter they support)

guilload: Most systems will truncate or reject oversized attribute values or spans. For instance, OTEL has limits for the number of attributes and length of values. New Relic limits value size to 4095 chars.

Logging potentially long values **here and in the rest of the codebase** is wrong.

ChillFish8: In that case, then I think we should go with the aforementioned `error!` log then.

guilload (approved):

PR #2240, opened 2022-11-03
Add some gc tests

guilload:
### Description
Verifies that the GC drops staged and marked splits after the corresponding grace periods.

### How was this PR tested?
:)

### DIFF
--- a/quickwit/quickwit-janitor/src/garbage_collection.rs
+++ b/quickwit/quickwit-janitor/src/garbage_collection.rs
@@ -108,7 +108,7 @@ pub async fn run_garbage_collect(
     .await?
     .into_iter()
     // TODO: Update metastore API and push this filter down.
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-janitor/src/garbage_collection.rs
```diff
@@ -108,7 +108,7 @@ pub async fn run_garbage_collect(
     .await?
     .into_iter()
     // TODO: Update metastore API and push this filter down.
-    .filter(|meta| meta.update_timestamp < grace_period_timestamp)
+    .filter(|meta| meta.update_timestamp <= grace_period_timestamp)
```
guilload, line 111: This makes no difference in practice but allows for running the tests 1 second faster ;)

fmassot (approved):

PR #2246, opened 2022-11-04
fix life time compile error in rust 1.65

Liyixin95:
### Description

This function can not be compiled on 1.65.Here is the error:

```
error[E0106]: missing lifetime specifier
  --> quickwit-storage/src/local_file_storage.rs:90:69
   |
90 | fn delete_all_dirs<'a>(root: &'a Path, path: &'a Path) -> BoxFuture<'_, std::io::Result<()>> {
   |                              --------        --------               ^^ expected named lifetime parameter
   |
   = help: this function's return type contains a borrowed value with an elided lifetime, but the lifetime cannot be derived from the arguments
help: consider using the `'a` lifetime
   |
90 | fn delete_all_dirs<'a>(root: &'a Path, path: &'a Path) -> BoxFuture<'a, std::io::Result<()>> {
   |                                                                     ~~

For more information about this error, try `rustc --explain E0106`.
error: could not compile `quickwit-storage` due to previous error
```

By the way, I am curious about why the boxed future are used here, why not just use async fn?


guilload:
Hi @Liyixin95, thank you for this fix. The future is boxed because the function is recursive.

fulmicoton:
Hello @Liyixin95 . Can you electronically sign the CLA using the CLA Assistant interface? This is a requirement for us to accept the PR. See the CLA Assistant box above.

Liyixin95:
> Hello @Liyixin95 . Can you electronically sign the CLA using the CLA Assistant interface? This is a requirement for us to accept the PR. See the CLA Assistant box above.

I have signed the CLA before but using a wrong email. Now it's fixed.

fulmicoton:
Thank you @Liyixin95 !

guilload (approved):

PR #2251, opened 2022-11-05
Implement safe int counter vec as `IntCounterVec<const N:usize>`

guilload:
### Description
Per title.

### How was this PR tested?
The compiler is doing all the work :)

### DIFF
--- a/quickwit/quickwit-common/src/metrics.rs
+++ b/quickwit/quickwit-common/src/metrics.rs
@@ -19,8 +19,39 @@
 
 use prometheus::{Encoder, HistogramOpts, Opts, TextEncoder};
 pub use prometheus::{
-    Histogram, HistogramTimer, HistogramVec, IntCounter, IntCounterVec, IntGauge, IntGaugeVec,
+    Histogram, HistogramTimer, HistogramVec as PrometheusHistogramVec, IntCounter,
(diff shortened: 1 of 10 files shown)


fmassot (approved):
Nice :)

quickwit/quickwit-common/src/metrics.rs
```diff
@@ -19,29 +19,35 @@
...
+    IntGauge, IntGaugeVec,
 };
 
+pub struct IntCounterVec<const N: usize> {
+    underlying: PrometheusIntCounterVec,
+}
+
+impl<const N: usize> IntCounterVec<N> {
+    pub fn new(name: &str, description: &str, namespace: &str, labels: [&str; N]) -> Self {
+        let counter_opts = Opts::new(name, description).namespace(namespace);
```
fulmicoton, line 32: The constructor registers the Counter. Despite the similar naming, that's a very different behavior from the prom counters. 

Can we put the code for this constructor in a `new_counter_vec` function.

```
pub fn new_counter_vec<const N: usize>(...) -> IntCounterVec<N> {
...
}
```

fulmicoton (approved):

fulmicoton:
Ah cool. This will save us time and trouble.

quickwit/quickwit-common/src/metrics.rs
guilload: Ok.

PR #2256, opened 2022-11-07
Shut down delete pipeline actors on pipeline finalization.

### LINKED ISSUE #2252: Shut down delete pipeline supervisors when shutting down the delete pipeline
When we detect that an index has been deleted, we shut down the corresponding pipeline. But supervisors are not shut down.
We need to shut down each supervisor on the delete pipeline finalization.

fmassot:
Fix #2252


quickwit/quickwit-actors/src/registry.rs
```diff
@@ -102,9 +102,9 @@ impl ActorRegistryForSpecificType {
 
 #[derive(Serialize, Debug)]
 pub struct ActorObservation {
-    type_name: &'static str,
-    instance_id: String,
-    obs: Option<serde_json::Value>,
+    pub type_name: &'static str,
```
fmassot, line 105: Made it `pub` for tests.

quickwit/quickwit-janitor/src/actors/delete_task_pipeline.rs
```diff
@@ -143,10 +158,7 @@ impl DeleteTaskPipeline {
             SplitsUpdateMailbox::Publisher(publisher_mailbox),
             self.max_concurrent_split_uploads,
         );
-        let (uploader_mailbox, uploader_supervisor_handler) = ctx
-            .spawn_actor()
-            .set_kill_switch(KillSwitch::default())
```
fmassot, line 148: The `supervise` function is doing the right thing on this, it calls `kill_switch.child()` and pass the child to the supervised actor.

With this modification, we can now kill the pipeline and expects that all children will be killed too.

fulmicoton (approved):

Issue #2257, opened 2022-11-07
Fix DateTime output on search


evanxg852000:
Currently Quickwit DateTime field output a string representation of the date during the search. This is inherited from Tantivy where the value is serialized as ` time::serde::rfc3339::serialize(&date.into_utc(), serializer)`.

We want users to specify their preferred output format using the available options in `Quickwit::DateTimeFormat`.

https://github.com/quickwit-oss/quickwit/issues/2220

evanxg852000:
@guilload I think we need to decide what tradeoff is best for us here. 
- Adding an `output_format` option in Quickwit and keeping Tantivy's output will require us to parse again if the `output_format` does not match the default tantivy `output_fromat` . The advantage is that we only change Quickwit. But we might need to parse datetime from RFC3339 to other formats we support (ISO8601, RFC2822, Timestamp).
- Change the default DateTime output in tantivy as microseconds `JsonValue::Number(i64)`  this will help us not parse the data again but this makes Tantivy DateTime less usable since user want to see datetime formated.
- A drastic change could be: to add an output format option (`timestamp`, `rfc3339`) to Tantivy::Date field. Then make sure we have access to the field option during serialization and use that option to output the desired format. Update the implementation of `Tantivy::Schema::to_json` to use the FieldEntry.

evanxg852000:
@fulmicoton Maybe I don't understand where this is done. But if we are doing it in Quicwit then at some point we have to serialize `NamedFieldDocument` to in order to send as `LeafHit` across nodes.
I think it's the serialization to JSON that we need to avoid so that we can get the original `BTreeMap<String, Vec<TanrivyValue>>,` which is just the inner value of  `NamedFieldDocument` we already use TanrivyValue.

PR #2259, opened 2022-11-07
Added DateField output_format

evanxg852000:
- Added `output_format`  to  QuickwitDateTimeOptions
- This PR contains changes to the format of `LeafHit` document sent to root search. In order to avoid potentially parsing a JSON document for displaying things in the correct format, we now format the output document at the leaf level. When the root receives a LeafHit, the JSON document within is ready to be shipped into a user search result with the user-requested format. 
Closes https://github.com/quickwit-oss/quickwit/issues/2257


docs/configuration/index-config.md
```diff
@@ -217,10 +218,11 @@ precision: milliseconds
 | Variable      | Description   | Default value |
 | ------------- | ------------- | ------------- |
 | `input_formats` | Formats used to parse input dates | [`rfc3339`, `unix_timestamp`] |
+| `output_format` | Formats used to display dates in search results | `rfc3339` |
 | `stored`        | Whether the field values are stored in the document store | `true` |
 | `indexed`       | Whether the field values are indexed | `true` |
 | `fast`          | Whether the field values are stored in a fast field | `false` |
-| `precision`     | The precision (`seconds`, `milliseconds`, or `microseconds`) used to store the fast values | `seconds` |
+| `precision`     | The precision (`seconds`, `milliseconds`, or `microseconds`) used to store the fast values.It's also used to display date when output_format is set to `unix_timestamp`  | `seconds` |
```
guilload, line 225:
```suggestion
| `precision`     | The precision (`seconds`, `milliseconds`, or `microseconds`) used to store the fast values. It's also used to display date when output_format is set to `unix_timestamp`  | `seconds` |
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -214,34 +183,53 @@ impl MappingLeaf {
...
+    if let (Value::Date(date_time), LeafType::DateTime(date_time_options)) =
+        (&matched_value, leaf_type)
+    {
+        let json_value = date_time_options
+            .format_to_json(*date_time)
+            .expect("Invalid datetime is not allowed.");
+        return Some(json_value);
+    }
+
+    let json_value =
```
fmassot, line 228: Question: for bytes, we need to encode into base64 as we want to return that to the user. Where is the encoding part done?

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -214,34 +183,53 @@ impl MappingLeaf {
...
         .into_iter()
-        .filter(|json_val| json_type_from_json_value(json_val) == json_type);
+        .filter_map(|value| value_to_json(value, leaf_type));
     match cardinality {
         Cardinality::SingleValue => vals_with_correct_type_it.next(),
         Cardinality::MultiValues => Some(JsonValue::Array(vals_with_correct_type_it.collect())),
     }
 }
 
+fn value_to_json(value: Value, leaf_type: &LeafType) -> Option<JsonValue> {
```
fmassot, line 203: I would write this function a bit differently

```rust
fn value_to_json(value: Value, leaf_type: &LeafType) -> Option<JsonValue> {
    // Make sure the requested type and the value type are consistent.
    match (&value, leaf_type) {
        (Value::Str(_), LeafType::Text(_))
        | (Value::I64(_), LeafType::I64(_))
        | (Value::U64(_), LeafType::U64(_))
        | (Value::F64(_), LeafType::F64(_))
        | (Value::Bool(_), LeafType::Bool(_))
        | (Value::IpAddr(_), LeafType::IpAddr(_))
        | (Value::Date(_), LeafType::DateTime(_))
        | (Value::Bytes(_), LeafType::Bytes(_))
        | (Value::JsonObject(_), LeafType::Json(_)) => Some(
            serde_json::to_value(&matched_value).expect("Json serialization should never fail."),
        ),
        (Value::Date(date_time), LeafType::DateTime(date_time_options)) => {
            let json_value = date_time_options
                .format_to_json(*date_time)
                .expect("Invalid datetime is not allowed.");
            Some(json_value)
        }
        _ => return None, // TODO: log a warning.
    }
}
```

quickwit/quickwit-search/src/lib.rs
```diff
@@ -139,34 +138,23 @@ async fn list_relevant_splits(
         .collect::<Vec<_>>())
 }
 
-/// Converts a `LeafHit` into a `Hit`.
-///
-/// Splits may have been created with different DocMappers.
-/// For this reason, leaves are returning a document that is -as much
-/// as possible-, `DocMapper` agnostic.
+/// Converts a leaf document into a JSON String before sending
```
fmassot, line 141: what is a leaf document? I would use `NamedFieldDocument` and optionally add the fact that it is coming from tantivy, tantivy `NamedFieldDocument` for example.

fmassot: It's important to mention that the conversion maps the tantivy `NamedFieldDocument` to the targeted schema defined by the `DocMapper`.

I did not really understand the comments. Here is what I would write:

`This conversion is performed at the leaf level only, it avoids having another intermeidate json format between the root and the leaves.`

quickwit/quickwit-search/src/lib.rs
```diff
@@ -139,34 +138,23 @@ async fn list_relevant_splits(
...
 ///
-/// For instance, if the cardinality of a field changed from single-valued
-/// to multivalued, we do want the documents emitted from old splits to
-/// also serialize the fields values as a JsonArray.
-///
-/// The `convert_leaf_hit` is critical and needs to be tested against
-/// allowed DocMapper changes.
-fn convert_leaf_hit(
-    leaf_hit: quickwit_proto::LeafHit,
+/// Important! The doc_mapper should be the most recent doc_mapper.
```
fmassot, line 148: It's better to explain why.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
evanxg852000: TantivyValue::Bytes already serializes it to base64

quickwit/quickwit-search/src/lib.rs
```diff
@@ -139,34 +138,23 @@ async fn list_relevant_splits(
...
+fn convert_leaf_doc_to_json(
```
fmassot, line 149: nitpick: `convert_document_to_json_string`

quickwit/quickwit-search/src/lib.rs
```diff
@@ -222,16 +205,20 @@ pub async fn single_node_search(
         leaf_search_response.partial_hits,
         index_storage,
         &split_metadata,
-        doc_mapper_opt,
+        doc_mapper.clone(),
```
fmassot, line 208: unecessary clone

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
evanxg852000: This is more compact  :+1:, I wanted more detailed steps in code rather than commenting.

docs/configuration/index-config.md
```diff
@@ -217,10 +218,11 @@ precision: milliseconds
 | Variable      | Description   | Default value |
 | ------------- | ------------- | ------------- |
 | `input_formats` | Formats used to parse input dates | [`rfc3339`, `unix_timestamp`] |
+| `output_format` | Format used to display dates in search results | `rfc3339` |
 | `stored`        | Whether the field values are stored in the document store | `true` |
 | `indexed`       | Whether the field values are indexed | `true` |
 | `fast`          | Whether the field values are stored in a fast field | `false` |
-| `precision`     | The precision (`seconds`, `milliseconds`, or `microseconds`) used to store the fast values | `seconds` |
+| `precision`     | The precision (`seconds`, `milliseconds`, or `microseconds`) used to store the fast values. It's also used to display date when output_format is set to `unix_timestamp`  | `seconds` |
```
fmassot, line 225: what does "It's also used to display date when output_format is set to `unix_timestamp`" mean?

Do we return milliseconds if precision is milliseconds?

evanxg852000: Yes

quickwit/quickwit-search/src/lib.rs
evanxg852000: Yes indeed last use

docs/configuration/index-config.md
fmassot, line 225: It would be nice to explain in the docs that:
- values are stored in the term dictionary at seconds precision
- values are stored in fast field storage  at the given precision
- values are stored in the doc store as is.

fmassot:
One remark: I don't really like binding the output format `unix_timestamp` to the precision that applies only to the fast field.

I would rather choose between these 2 options:
- return microseconds
- fix the `output_format`. `unix_timestamp` is a broken format as we don't know if we have seconds, milliseconds, microseconds. I would prefer to have `unix_timestamp_sec`, `unix_timestamp_ms`, ``unix_timestamp_micros` for example. `unix_timestamp` is nice only for ingestion as we can guess the format.

evanxg852000:
@fmassot I think what I can do in this PR at least to avoid confusion is to add `output_precision`  with default like other and have them separated from the fast field storage precision.

docs/configuration/index-config.md
evanxg852000: So my concern about adding this in the docs is that they are implementation details. I wanted to just limit to what precision is used for. The rest as long as we return correctly what use put in.

fmassot: Yes, you're right. You can add maybe that in the `docs/internals` :)

docs/internals/date-time.md
```diff
@@ -0,0 +1,6 @@
+# Split format
+
+Quickwit's DateTime is a wrapper around Tantivy's provided DateTime type which is internally represented as an `i64` microseconds value. For optimization reasons, Tantivy stores the value differently at the following locations:
+- DocStore: Dates are stored as they are received from the input document.
+- TermDic: Dates are stored with `seconds` precision.
```
guilload, line 5:
```suggestion
- TermDict: Dates are stored with `seconds` precision.
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -25,7 +25,7 @@ use quickwit_proto::SearchRequest;
 use serde::{Deserialize, Serialize};
 use serde_json::{self, Value as JsonValue};
 use tantivy::query::Query;
-use tantivy::schema::{Cardinality, Field, FieldType, Schema, STORED};
+use tantivy::schema::{Cardinality, Field, FieldType, Schema, Value, STORED};
```
guilload, line 28:
```suggestion
use tantivy::schema::{Cardinality, Field, FieldType, Schema, Value as TantivyValue, STORED};
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -186,21 +164,13 @@ impl MappingLeaf {
 
     fn populate_json<'a>(
         &'a self,
-        named_doc: &mut BTreeMap<String, Vec<JsonValue>>,
+        named_doc: &mut BTreeMap<String, Vec<Value>>,
```
guilload, line 167:
```suggestion
        named_doc: &mut BTreeMap<String, Vec<TantivyValue>>,
```

guilload (commented):
Let's consistently use `TantivyValue` for `tantivy::schema::Value`.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
evanxg852000: @guilload 
My initial understanding of this when I worked on the search result snippet structure was that we wanted to make the distinction for JsonValue when we have conflicting Value types in a file. 
Now I have the feeling that we want to always have the distinction between JsonValue and TantivyValue clear. In the later case, I think a separate PR should be better since we use these value types in many places.
I have fixed both files I touched on here but let me know if we should do the refactor across the whole project. I will create an issue.

guilload: Yes, I'd rather always distinguish them. You don't always get to see the imports when reviewing on GitHub, and as the size of the codebase increase, I have a hard time keeping track of what's what.

fmassot:
@evanxg852000: I just talked to @guilload and @fulmicoton : it would be super nice to have an `output_format` with `unix_timestamp_secs`, `unix_timestamp_millis`, `unix_timestamp_micros`, can you handle that?

quickwit/quickwit-doc-mapper/src/default_doc_mapper/date_time_format.rs
```diff
@@ -97,55 +102,132 @@ impl std::hash::Hash for StrptimeParser {
 
 /// Specifies the datetime and unix timestamp formats to use when parsing date strings.
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
-pub enum DateTimeFormat {
+pub enum InputDateTimeFormat {
```
guilload, line 105:
```suggestion
pub enum DateTimeInputFormat {
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/date_time_format.rs
```diff
@@ -97,55 +102,132 @@ impl std::hash::Hash for StrptimeParser {
...
+    where D: Deserializer<'de> {
+        let date_time_format_str: String = Deserialize::deserialize(deserializer)?;
+        let date_time_format = date_time_format_str.parse().map_err(D::Error::custom)?;
+        Ok(date_time_format)
+    }
+}
+
+/// Specifies the datetime format to use when displaying datetime values.
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub enum OutputDateTimeFormat {
```
guilload, line 170:
```suggestion
pub enum DateTimeOutputFormat {
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -214,38 +184,61 @@ impl MappingLeaf {
...
+    named_doc: &mut BTreeMap<String, Vec<TantivyValue>>,
     field_path: &[&str],
     cardinality: Cardinality,
 ) -> Option<JsonValue> {
     let full_path = field_path.join(".");
     let vals = named_doc.remove(&full_path)?;
     let mut vals_with_correct_type_it = vals
         .into_iter()
-        .filter(|json_val| json_type_from_json_value(json_val) == json_type);
+        .filter_map(|value| value_to_json(value, leaf_type));
```
guilload, line 197:
```suggestion
        .flat_map(|value| value_to_json(value, leaf_type));
```

guilload: `flat_map` is semantically better when flattening sequences:

```rust
let all_chars: Vec <_> = ["foo", "bar"].into_iter().flat_map(|word| word.chars()).collect();
```

`filter_map`

```rust
let even_octals: Vec <_> = [1, 2].into_iter().filter_map(|i| if i % 2 == 0 { Some(format!("{i:o}")) } else { None }).collect();
```

guilload (approved):

PR #2260, opened 2022-11-07
Add JSON property support for default search fields

### LINKED ISSUE #1503: Can't use a key as default search field for JSON field
Occurred as I was writing the Kafka tutorial. On main branch.

```yaml
...
doc_mapping:
  field_mappings:
    - name: org
      type: json
      tokenizer: default
...
search_settings:
  default_search_fields: [org.name]
```

```bash
quickwit index create --index-config ...
Command failed: Unknown default search field: `org.name`
````

### PR DESCRIPTION
**Pull request by boraarslan**
Closes #1503

### DIFF
--- a/quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
+++ b/quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
@@ -275,16 +275,8 @@ impl TryFrom<DefaultDocMapperBuilder> for DefaultDocMapper {
         validate_tag_fields(&builder.tag_fields, &schema)?;
 
         // Resolve default search fields
-        let mut default_search_field_names = Vec::new();
-        for field_name in &builder.default_search_fields {
-            if default_search_field_names.contains(field_name) {
-                bail!("Duplicated default search field: `{}`", field_name)
-            }
-            schema
-                .get_field(field_name)
-                .with_context(|| format!("Unknown default search field: `{}`", field_name))?;
-            default_search_field_names.push(field_name.clone());
-        }
+        let default_search_field_names =
+            resolve_default_search_fields(&builder.default_search_fields, &schema)?;
 
         resolve_timestamp_field(builder.timestamp_field.as_ref(), &schema)?;
         let sort_by = resolve_sort_field(builder.sort_by, &schema)?;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs [hunk 1]
```diff
@@ -320,6 +312,52 @@ impl TryFrom<DefaultDocMapperBuilder> for DefaultDocMapper {
...
+    default_search_fields: &[String],
+    schema: &Schema,
+) -> Result<Vec<String>, anyhow::Error> {
+    let mut default_search_field_names = Vec::new();
+    for field_name in default_search_fields {
+        if default_search_field_names.contains(field_name) {
+            bail!("Duplicated default search field: `{}`", field_name)
+        }
+        if field_name.contains('.') {
+            let potential_json_field_name = field_name
```
fulmicoton, line 325: Quickwit has a weird feature where you can start schemaless and add some schema specification for some stuff later.

This works by having a JSON field called `product_details` for instance, and a field called `product_details.price` (For instance to specify it as a fast field).

On the search side, if you want to identify the field that is targetted, you need to pick the longest field name.
so product_details.price targets the u64 field in my example.

Your implementation should replicate this logic.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -320,6 +312,52 @@ impl TryFrom<DefaultDocMapperBuilder> for DefaultDocMapper {
...
+                .split('.')
+                .next()
+                .expect("Field name should contain `.`");
```
fulmicoton, line 328: I don't think this expect is needed is it?

```rust
if let Some(potential_json_field_name) = field_name.split('.').next() {
   //...
}  
``` 
(or use split_once)

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -320,6 +312,52 @@ impl TryFrom<DefaultDocMapperBuilder> for DefaultDocMapper {
...
+
+            if !potential_json_field_name.ends_with('\\')
+                && schema.get_field(potential_json_field_name).is_some()
```
fulmicoton, line 331: same as above. 
is_some() + unwrap is unnecessary.

fulmicoton: `if let Some() =` is cleaner

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
fulmicoton: No you do no need to espace the . here.
I am talking of the object type
https://quickwit.io/docs/configuration/index-config#object

Let me give a concrete example:
I am using dynamic mode. I'd like to index everything I am being given. 
All of the text should be indexed with the raw tokenizer, except the  `product.description` part that should be tokenized.

I can do that today with the following doc mapper

```yaml
mode: dynamic
field_mappings:
    - name: product
      type: object
      field_mappings:
          name: description
          type: text
          tokenizer: default
```

Given a product 
```json
{"author": "user2312", "product": {"description": "cool umbrella", "id": "2222131"}}
```

In that case
```
author: is mapped in the dynamic field. the raw tokenizer is used by default
product.id is mapping in the dynamic field. the raw tokenizer is used by default
product.descirpiton is mapped in the static adhoc field and tokenizer
```

In the query parser
`product.description:red` works without escaping the `.`

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -320,6 +312,55 @@ impl TryFrom<DefaultDocMapperBuilder> for DefaultDocMapper {
...
+    schema: &Schema,
+) -> Result<Vec<String>, anyhow::Error> {
+    let mut default_search_field_names = Vec::new();
+    'outer: for field_name in default_search_fields {
+        if default_search_field_names.contains(field_name) {
+            bail!("Duplicated default search field: `{}`", field_name)
+        }
+
+        // Block to handle potential json fields
+        'inner: {
```
fulmicoton, line 326: can we remove the tagged loops by having an extra function?

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -320,6 +312,55 @@ impl TryFrom<DefaultDocMapperBuilder> for DefaultDocMapper {
...
+            let Some((potential_json_field_name, _)) = field_name.split_once('.') else {
```
fulmicoton, line 327: a json field could be nested an object

PR #2263, opened 2022-11-07
Remove publisher finalize and shut down property a merge pipeline wit…

### LINKED ISSUE #2262: Merge pipeline stops if an actor in the indexing pipeline fails.
This is due to the fact that the publisher will send an exit command to the merge planner.

fmassot:
Fix #2262 

+ kill a merge pipeline when there is no more indexing pipeline registered on a given (index_id, source_id).

I think we should also kill an indexing pipeline when we shut it down (but will see that in another PR).

2 tests added.

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexing_service.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexing_service.rs
@@ -90,12 +90,13 @@ pub struct IndexingServiceState {
     pub num_running_pipelines: usize,
     pub num_successful_pipelines: usize,
     pub num_failed_pipelines: usize,
+    pub num_running_merge_pipelines: usize,
 }
 
 type IndexId = String;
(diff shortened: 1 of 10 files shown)


quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -385,19 +386,40 @@ impl IndexingService {
                     }
                 },
             );
-        // Evict merge pipelines that are not needed or failing.
+        // Evict and and killing merge pipelines that are not needed.
```
fulmicoton, line 389:
```suggestion
        // Evict and kill merge pipelines that are not needed.
```

fulmicoton (approved):

fulmicoton (approved):

fulmicoton (approved):

PR #2276, opened 2022-11-09
Implement connectivity check for gRPC metastore client

guilload:
### Description
Per title. Pass number of metastore servers in the pool via watch channel between pool task and metastore.

### How was this PR tested?
Added asserts in current test suite


guilload:
I'm using this PR as a warm-up for my work on probes for k8s.

fmassot (approved):

PR #2282, opened 2022-11-10
Reduce attack surface on the FileStorage.

### LINKED ISSUE #53: Investigate security considerations on the local file storage

fulmicoton:
We prevent people from using relative path including a `..` in the FileStorage.

Closes #53

### DIFF
--- a/quickwit/quickwit-storage/src/local_file_storage.rs
+++ b/quickwit/quickwit-storage/src/local_file_storage.rs
@@ -21,7 +21,7 @@ use std::collections::{BTreeSet, HashMap};
 use std::fmt;
 use std::io::{ErrorKind, SeekFrom};
 use std::ops::Range;
-use std::path::{Path, PathBuf};
+use std::path::{Component, Path, PathBuf};
 use std::sync::Arc;
 
 use async_trait::async_trait;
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-storage/src/local_file_storage.rs
```diff
@@ -71,23 +76,55 @@ impl LocalFileStorage {
...
+        ignore_error_kind!(ErrorKind::NotFound, fs::remove_file(full_path).await)?;
+        Ok(())
+    }
+}
+
+/// Ensure that the path given does not include any ".." for security reasons.
+///
+/// In order to reduce the attack surface, we want to make sure the `FileStorage`
+/// only access/delete files that are children of its root_directory.
+fn ensure_no_parent_dir_double_dot(path: &Path) -> StorageResult<()> {
```
ChillFish8, line 103: I think this could be called something like `ensure_valid_path` or similar as this does technically do more than just ensure the parent isn't a `..` for example, `/todo` will also get rejected by this routine.

fulmicoton: If you just have "nitpicks" comment, please mark the CR as approved. This is a practise that is a bit specific to quickwit: we are all in different timezone, so it is critical that we keep the process asynchronous when possible.

fulmicoton: (I did the rename and will abuse my privilege and merge that  PR  :) )

ChillFish8: Ah apologies 😅

fulmicoton: No problem! you could not know! :)

PR #2284, opened 2022-11-10
Delete term set

### LINKED ISSUE #2217: Make it possible to delete a set of terms. (using the TermSetQuery on tantivy side)
(discuss spec with Francois)

trinity-1686a:
fix #2217

Add support for deleting using a set of terms (and searching with a term of search).


quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -33,38 +34,103 @@ pub(crate) fn build_query(
...
 
-    if needs_default_search_field(&user_input_ast)
-        && request.search_fields.is_empty()
-        && (default_field_names.is_empty() || default_field_names == [DYNAMIC_FIELD_NAME])
-    {
-        return Err(
-            anyhow::anyhow!("No default field declared and no field specified in query.").into(),
-        );
-    }
+            if needs_default_search_field(&user_input_ast)
```
fulmicoton, line 46: what is the diff here?

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -33,38 +34,103 @@ pub(crate) fn build_query(
...
-    } else {
-        resolve_fields(&schema, &request.search_fields)?
-    };
+            let mut query_parser =
+                QueryParser::new(schema, search_fields, QUICKWIT_TOKENIZER_MANAGER.clone());
+            query_parser.set_conjunction_by_default();
+            query_parser.parse_query(query)?
+        }
+        Some(SearchQuery::SetQuery(set_query)) => {
+            let _ = (&set_query.terms, &set_query.field_name, &set_query.tags);
```
fulmicoton, line 79: What is this line about?

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -33,38 +34,103 @@ pub(crate) fn build_query(
...
+            let field_name = &set_query.field_name;
+            let field = schema
+                .get_field(field_name)
+                .ok_or_else(|| TantivyQueryParserError::FieldDoesNotExist(field_name.clone()))?;
+
+            let field_type = schema.get_field_entry(field).field_type();
+            if !field_type.is_indexed() {
+                return Err(anyhow::anyhow!("Attempted to search on not indexed field.").into());
```
fulmicoton, line 87:
```suggestion
                return Err(anyhow::anyhow!("Attempted to search on a field that is not indexed {}.", field_name).into());
```

quickwit/quickwit-doc-mapper/src/query_builder.rs
trinity-1686a: something I forgot to cleanup

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -33,38 +34,103 @@ pub(crate) fn build_query(
...
+                    anyhow::anyhow!("Attempted to search on unsuported field type.").into(),
+                );
+            }
+
+            // TODO error on any parse error instead??
+            let terms: Vec<_> = set_query
+                .terms
+                .iter()
+                // TODO not correct
+                .flat_map(|term| field_type.value_from_json(term.clone().into()).ok())
```
fulmicoton, line 102: is `.ok()` needed here?

```suggestion
                .flat_map(|term| field_type.value_from_json(term.clone().into()))
```

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -33,38 +34,103 @@ pub(crate) fn build_query(
...
-    }
+            // TODO maybe Facet could be allowed?
+            if matches!(field_type.value_type(), Type::Json | Type::Facet) {
+                return Err(
+                    anyhow::anyhow!("Attempted to search on unsuported field type.").into(),
+                );
+            }
+
+            // TODO error on any parse error instead??
+            let terms: Vec<_> = set_query
```
fulmicoton, line 98: Yes let's return an explicit error with the first failing term.

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -33,38 +34,103 @@ pub(crate) fn build_query(
...
+                .terms
+                .iter()
+                // TODO not correct
+                .flat_map(|term| field_type.value_from_json(term.clone().into()).ok())
+                .map(|term| match term {
+                    Value::Str(text) => Term::from_field_text(field, &text),
+                    Value::PreTokStr(pre_tokenized_string) => {
+                        Term::from_field_text(field, &pre_tokenized_string.text)
+                    }
+                    Value::U64(int) => Term::from_field_u64(field, int),
```
fulmicoton, line 108:
```suggestion
                    Value::U64(int_val) => Term::from_field_u64(field, int_val),
```

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -33,38 +34,103 @@ pub(crate) fn build_query(
...
+                    Value::I64(int) => Term::from_field_i64(field, int),
+                    Value::F64(int) => Term::from_field_f64(field, int),
+                    Value::Bool(b) => Term::from_field_bool(field, b),
```
fulmicoton, line 111:
```suggestion
                    Value::Bool(bool_val) => Term::from_field_bool(field, bool_val),
```

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -33,38 +34,103 @@ pub(crate) fn build_query(
...
+                    Value::Date(date) => Term::from_field_date(field, date),
+                    Value::Bytes(buf) => Term::from_field_bytes(field, &buf),
+                    Value::IpAddr(ip) => Term::from_field_ip_addr(field, ip),
+                    Value::JsonObject(_) | Value::Facet(_) => unreachable!(),
```
fulmicoton, line 115: I agree this is unreachable, but I'd rather return an error here too.

The point is to not make anyone think when thinking can this function panic.
```suggestion
                    Value::JsonObject(_) | Value::Facet(_) => unreachable!(),
```

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -33,38 +34,103 @@ pub(crate) fn build_query(
...
+            .into())
+        }
+    };
 
-    let mut query_parser =
-        QueryParser::new(schema, search_fields, QUICKWIT_TOKENIZER_MANAGER.clone());
-    query_parser.set_conjunction_by_default();
-    let query = query_parser.parse_query(&request.query)?;
     Ok(query)
 }
```
fulmicoton, line 135: Can we add unit tests testing what happens if terms are wrong, or are correct?

quickwit/quickwit-indexing/src/actors/merge_executor.rs
```diff
@@ -459,7 +459,7 @@ impl MergeExecutor {
                     .expect("A delete task must have a delete query.");
                 let search_request = SearchRequest {
                     index_id: delete_query.index_id,
-                    query: delete_query.query,
+                    query: delete_query.query.map(Into::into),
```
fulmicoton, line 462: Blabla::from would be more readable I think. Here I don't know what the type is.

quickwit/quickwit-janitor/src/actors/delete_task_planner.rs
```diff
@@ -236,8 +237,28 @@ impl DeleteTaskPlanner {
                         delete_query.end_timestamp,
                     );
                     // TODO: validate the query at the beginning and return an appropriate error.
-                    let tags_filter = extract_tags_from_query(&delete_query.query)
-                        .expect("Delete query must have been validated upfront.");
+                    let tags_filter = match &delete_query.query {
+                        Some(Query::Text(query_string)) => extract_tags_from_query(query_string)
+                            .expect("Delete query must have been validated upfront."),
```
fulmicoton, line 242: The TODO is right let's return an error. I am not 100% sure what upfront means here.

quickwit/quickwit-janitor/src/actors/delete_task_planner.rs
```diff
@@ -236,8 +237,28 @@ impl DeleteTaskPlanner {
...
+                                        .map(|tag| TagFilterAst::Tag {
+                                            is_present: true,
+                                            tag: tag.clone(),
+                                        })
+                                        .collect(),
+                                ))
+                            }
+                        }
+                        None => return false, // invalid query: ignore. TODO add log?
+                    };
```
fulmicoton, line 260: Can we make this a function?

quickwit/quickwit-janitor/src/actors/delete_task_planner.rs
```diff
@@ -292,7 +313,7 @@ impl DeleteTaskPlanner {
                 .expect("Delete task must have a delete query.");
             let search_request = SearchRequest {
                 index_id: delete_query.index_id.clone(),
-                query: delete_query.query.clone(),
+                query: delete_query.query.clone().map(|q| q.into()),
```
fulmicoton, line 316:
```suggestion
                query: delete_query.query.clone().map(|query| Blabla::from(query)),
```

quickwit/quickwit-proto/protos/quickwit/common.proto
```diff
@@ -0,0 +1,42 @@
...
+syntax = "proto3";
+
+package quickwit_common;
+
+message SetQuery {
+  // List of terms to search.
+  repeated Term terms = 1;
+  // Field to search in.
+  string field_name = 2;
+  // Union of tags a split must have ot be searched.
```
fulmicoton, line 29:
```suggestion
  // Union of tags a split must have to be searched.
```

quickwit/quickwit-proto/protos/quickwit/common.proto
```diff
@@ -0,0 +1,42 @@
...
+  repeated string tags = 3;
+}
+
+message Term {
```
fulmicoton, line 33: overkill?
Down the line users are using text in the query parser. We probably want to have them enter text here too, and parse that text the way we parse it in the query parser.

quickwit/quickwit-proto/protos/quickwit/metastore_api.proto
```diff
@@ -185,8 +186,10 @@ message DeleteQuery {
   optional int64 start_timestamp = 2;
   /// If set, restrict search to documents with a `timestamp < end_timestamp``.
   optional int64 end_timestamp = 3;
-  /// Query text. The query language is that of tantivy.
-  string query = 4;
+  oneof query {
+    string text = 4;
```
fulmicoton, line 190:
```suggestion
    string query_text = 4;
```

quickwit/quickwit-proto/src/lib.rs
```diff
@@ -198,3 +202,118 @@ impl fmt::Display for SplitSearchError {
...
+
+pub(crate) mod serde_helpers {
+    use serde::de::{Deserialize, Deserializer};
+    use serde::Serialize;
+    use crate::metastore_api::delete_query::Query;
+    use crate::quickwit_common::SetQuery;
+
+    pub fn required_option<'de, D, T: Deserialize<'de> + std::fmt::Debug>(deserializer: D) -> Result<Option<T>, D::Error>
+    where D: Deserializer<'de> {
+            dbg!(T::deserialize(deserializer).map(Some))
```
fulmicoton, line 290: what is this ?

quickwit/quickwit-search/src/fetch_docs.rs
```diff
@@ -308,6 +308,7 @@ async fn create_snippet_generator(
     field: Field,
 ) -> anyhow::Result<SnippetGenerator> {
     let mut terms: Vec<&Term> = Vec::new();
+    // TODO ok with termset?
```
fulmicoton, line 311: what does the TODO mean?

fulmicoton: Ah is that working with TermSets?

well we should probably make sure the query_terms query is implemented and truncate the list
when there are too many terms.

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -33,38 +34,103 @@ pub(crate) fn build_query(
...
+            let field_type = schema.get_field_entry(field).field_type();
+            if !field_type.is_indexed() {
+                return Err(anyhow::anyhow!("Attempted to search on not indexed field.").into());
+            }
 
-    if let Some(sort_by_field) = request.sort_by_field.as_ref() {
-        validate_sort_by_field_name(sort_by_field, &schema, Some(&search_fields))?;
-    }
+            // TODO maybe Facet could be allowed?
+            if matches!(field_type.value_type(), Type::Json | Type::Facet) {
```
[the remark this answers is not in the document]
trinity-1686a: I'm not sure how that would work (or how Json indexing works for that matter), so I'll open a separate issue when this PR get merged

quickwit/quickwit-doc-mapper/src/query_builder.rs
trinity-1686a: hum, that should be a filter_map actually, but that's related to the TODO just above (which you answered), so it'll go away

quickwit/quickwit-doc-mapper/src/query_builder.rs
trinity-1686a: I haven't added much tests yet, but yes, will do

quickwit/quickwit-janitor/src/actors/delete_task_planner.rs
trinity-1686a: that bit just got moved around, not sure what upfront means either

quickwit/quickwit-search/src/fetch_docs.rs
trinity-1686a: I don't know if that part of the code works properly with TermSetQuery, I haven't tested it yet. I don't know if it's important either, or can be considered a limitation, and queries with termset and snippet be rejected

quickwit/quickwit-proto/protos/quickwit/common.proto
trinity-1686a: things here don't go through the query parser, they get converted [here](https://github.com/quickwit-oss/quickwit/blob/e4268d41cf4657a4c326e11047bb9c27f0c2eb94/quickwit/quickwit-doc-mapper/src/query_builder.rs#L102-L116) once we have a schema to know what type these terms are supposed to be

quickwit/quickwit-doc-mapper/src/doc_mapper.rs
```diff
@@ -134,6 +134,19 @@ pub struct NamedField {
...
+/// running the query.
+#[derive(Debug, Default, Clone)]
+#[non_exhaustive]
+pub struct WarmupInfo {
+    /// Name of fields from the term dictionnary which needs to be entirely
+    /// loaded
+    pub term_dict_names: HashSet<String>,
+    /// Name of fields from the posting lists which needs to be entirely
+    /// loaded
+    pub posting_names: HashSet<String>,
```
fulmicoton, line 147: I like this approach. Could/Should we include individual terms as well?

quickwit/quickwit-doc-mapper/src/doc_mapper.rs
```diff
@@ -134,6 +134,19 @@ pub struct NamedField {
...
 clone_trait_object!(DocMapper);
 
+/// Informations about what a DocMapper think should be warmed up before
+/// running the query.
+#[derive(Debug, Default, Clone)]
+#[non_exhaustive]
+pub struct WarmupInfo {
+    /// Name of fields from the term dictionnary which needs to be entirely
+    /// loaded
+    pub term_dict_names: HashSet<String>,
```
fulmicoton, line 144:
```suggestion
    pub term_dict_field_names: HashSet<String>,
```

quickwit/quickwit-doc-mapper/src/doc_mapper.rs
fulmicoton, line 147:
```suggestion
    pub posting_field_names: HashSet<String>,
```

quickwit/quickwit-doc-mapper/src/doc_mapper.rs
```diff
@@ -236,7 +249,7 @@ mod tests {
             sort_by_field: None,
             aggregation_request: None,
         };
-        let query = doc_mapper.query(schema, &search_request).unwrap();
+        let (query, _) = doc_mapper.query(schema, &search_request).unwrap();
```
fulmicoton, line 252: do we unit test the Warmup structure somewhere?
This file seems like a good place to do that.

quickwit/quickwit-doc-mapper/src/doc_mapper.rs
```diff
@@ -134,6 +134,19 @@ pub struct NamedField {
 
 clone_trait_object!(DocMapper);
 
+/// Informations about what a DocMapper think should be warmed up before
+/// running the query.
+#[derive(Debug, Default, Clone)]
+#[non_exhaustive]
```
fulmicoton, line 140: TIL non_exhaustive made sense for Struct too.

quickwit/quickwit-doc-mapper/src/doc_mapper.rs
trinity-1686a: latest commit add a simple test case [in `query_builder.rs`](https://github.com/quickwit-oss/quickwit/pull/2284/commits/8e248ca698c8fdfb55a8e4f0026fed588b8dd98b#diff-c7130467f6b55c09737ded5297aa3d1b083d5e4913fc97874d2bc95af18b8baaR502). Maybe `doc_mapper.rs` is a better place

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -485,4 +499,49 @@ mod test {
             "The snippet field `server.running` must be of type `Str`, got `Bool`."
         );
     }
+
+    #[test]
+    fn test_build_queryè_warmup_info() -> anyhow::Result<()> {
```
fulmicoton, line 504:
```suggestion
    fn test_build_query_warmup_info() -> anyhow::Result<()> {
```

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -0,0 +1,359 @@
...
+        key.range_start = final_range.start;
+        let value = CacheValue {
+            range_end: final_range.end,
+            bytes: final_bytes,
+        };
+        self.cache.insert(key, value);
+        self.record_item(final_range.end - final_range.start);
+    }
+
+    // Return a block that contain everything between query.range_start and range_end
```
fulmicoton, line 204: great comment.

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -0,0 +1,359 @@
...
+    fn get_block<'a, 'b: 'a>(
+        &'a self,
+        query: &CacheKey<'b, T>,
+        range_end: usize,
```
fulmicoton, line 208: I could be wrong... because it requires a bit of code wrangling to check that but....

wouldn't the code be clearer by changing this function into...

```
fn get_block<'a>(&'a self, tag: &'b T, byte_range: Range<usize>) -> Option<(Range<usize>, &CacheValue)> {}
```

the goal being to restrict the number of place where we manipulate `CachedKey`.

fulmicoton: ... 
i'm not sure the refactoring I suggest is a good idea, due to the write usage.

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -0,0 +1,359 @@
...
+
+        if first_matching_block.is_some() && first_matching_block == last_matching_block {
+            // same start and end: all the range is already covered
+            return;
+        }
+
+        let first_matching_block = first_matching_block.unwrap_or(&start_key);
+        let last_matching_block = last_matching_block.unwrap_or(&end_key);
+
+        let overlapping = self
```
fulmicoton, line 126:
```suggestion
        let overlapping: Vec<usize> = self
```

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -0,0 +1,359 @@
...
+                byte_range.start
+            } else {
+                // if no first, can_drop_first is false
+                overlapping.first().unwrap().start
+            };
+
+            let end = if can_drop_last {
+                byte_range.end
+            } else {
+                // if no last, can_drop_last is false
```
fulmicoton, line 155:
```suggestion
                // if no last, can_drop_last is true
```

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    fn ops_strategy() -> impl Strategy<Value = Vec<Operation>> {
+        prop::collection::vec(op_strategy(), 1..100)
+    }
+
+    proptest::proptest! {
+        #[test]
+        fn test_proptest_byte_range_cache(ops in ops_strategy()) {
+            let mut state = HashMap::new();
```
fulmicoton, line 324:
```suggestion
            let mut state: HashMap<&'static str, Vec<bool>> = HashMap::new();
```

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -0,0 +1,359 @@
...
+            tag: &'static str,
+        },
+    }
+
+    fn tag_strategy() -> impl Strategy<Value = &'static str> {
+        prop_oneof![Just("path1"), Just("path2"),]
+    }
+
+    fn range_strategy() -> impl Strategy<Value = Range<usize>> {
+        (0usize..4095usize).prop_perturb(|start, mut rng| start..rng.gen_range(start..4096))
```
fulmicoton, line 305: TIL prob_perturb

fulmicoton, line 305: This 4096 is way too high and prevents you from covering much ground.

```suggestion
        (0usize..11usize).prop_perturb(|start, mut rng| start..rng.gen_range(start..12usize))
```

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -0,0 +1,359 @@
...
+                    } => {
+                        let slice = cache.get_slice(Path::new(tag), range.clone());
+                        if state[tag][range.clone()].iter().all(|t| *t) {
+                            let slice = slice.unwrap();
+                            let bytes = range.clone().map(|i| (i%256) as u8).collect::<Vec<_>>();
+                            assert_eq!(slice[..], bytes[..]);
+
+                        } else {
+                            assert!(slice.is_none());
+                        }
```
fulmicoton, line 353: it mights be worth adding an assert on the counter. They are somewhat critical in that a bug could cause a panic (if we check underflow)

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -0,0 +1,359 @@
...
+        }
+    }
+}
+
+struct CacheValue {
+    range_end: usize,
+    bytes: OwnedBytes,
+}
+
+// T is a tag, usually a file path, R is the range bound type
```
fulmicoton, line 59:
```suggestion
/// `NeedMutByteRangeCache` is a cache without any eviction strategy.
/// It assumes an infinite capacity.  
///
/// This cache is used in the contraption that makes it possible for Quickwit
/// to use tantivy while doing asynchronous io.
/// Quickwit manually populates this cache in an asynchronous "warmup" phase.
/// tantivy then gets its data from this cache without performing any IO.
///
/// T is a tag, usually a file path.
```

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -0,0 +1,359 @@
...
+            .sub(self.num_bytes as i64);
+    }
+}
+
+/// Cache for ranges of bytes in files.
+/// Contrary to `MemorySizedCache`, it's able to answer subset of known ranges.
+///
+/// This cache assume immutable data: if you put a new slice and it overlap with
+/// cached data, the changes may or may not get recorded.
+// At the moment this is hardly a cache as it features no eviction policy.
```
fulmicoton, line 247:
```suggest
/// Cache for ranges of bytes in files.
///
/// This cache is used in the contraption that makes it possible for Quickwit
/// to use tantivy while doing asynchronous io.
/// Quickwit manually populates this cache in an asynchronous "warmup" phase.
/// tantivy then gets its data from this cache without performing any IO.
///
/// Contrary to `MemorySizedCache`, it's able to answer subset of known ranges,
/// does not have any eviction, and assumes an infinite capacity. 
///
/// This cache assume immutable data: if you put a new slice and it overlap with
/// cached data, the changes may or may not get recorded.
///
/// At the moment this is hardly a cache as it features no eviction policy.
```

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -0,0 +1,359 @@
...
+        self.cache_counters.hits_num_items.inc();
+        self.cache_counters
+            .hits_num_bytes
+            .inc_by((end - start) as u64);
+
+        Some(v.bytes.slice(start..end))
+    }
+
+    fn put_slice(&mut self, tag: T::Owned, byte_range: Range<usize>, bytes: OwnedBytes) {
+        let len = byte_range.end - byte_range.start;
```
fulmicoton, line 103: I suspect this code could have been mildly simpler... but it is well encapsulated and well tested, so actually I think we don't care.

PR #2152, opened 2022-10-20
Add `ListSplitsQuery` builder for filtering splits metadata.

ChillFish8:
A WIP implementation of moving the way we query and filter splits from the metastore from a parameter-based system to a builder pattern.

The idea is that this simplifies the API, allows for finer grain filtering and future-proofs us if we want to extend what splits can be filtered.

**Example usage**
```rust
let filter = SplitFilter::for_index(index_id)
            .with_opstamp_older_than(delete_opstamp)
            .with_limit(num_splits)
            .with_offset(500)
            .with_time_range_from(100)
            .with_time_range_to(400)
            .with_opstamp_older_than(delete_opstamp)
            .with_split_state(SplitState::Staged)
            .updated_after(time::OffsetDateTime::UNIX_EPOCH.unix_timestamp())   // `updated_after_or_at` variant available for inclusive bound.
            .updated_before(time::OffsetDateTime::now_utc().unix_timestamp())   // `updated_before_or_at` variant available for inclusive bound.
            .with_tags_filter(TagFilterAst::Tag { is_present: true, tag: "sample".to_string() });

let splits = store.list_splits(filter).await
```

These changes are needed to allow for some nicer optimisations on #2126

PR #2287, opened 2022-11-10
QW-2281: Fix timestamp filters for PSQL

### LINKED ISSUE #2281: Failed to run garbage collection on index caused by a query error "operator does not exist"
Very recently, the garbage collector is not working on my index, the logs are like this:

```
2022-11-10T08:32:12.119Z ERROR quickwit_janitor::actors::garbage_collector: Failed to run garbage collection on index. index_id=quickwit-fts-staging error=Database error: `error returned from database: operator does not exist: timestamp without time zone <= integer`.
2022-11-10T08:33:12.122Z ERROR quickwit_janitor::actors::garbage_collector: Failed to run garbage collection on index. index_id=quickwit-fts-staging error=Database error: `error returned from database: operator does not exist: timestamp without time zone <= integer`.
2022-11-10T08:33:12.171Z ERROR quickwit_janitor::actors::garbage_collector: Failed to run garbage collection on index. index_id=ec2-quickwit-fts-staging error=Database error: `error returned from database: operator does not exist: timestamp without time zone <= integer`.
```

It may be related to my database schema.

ChillFish8:
Fixes #2281 

This fixes the bug introduced by #2152 and adds unit tests to prevent this error from re-occurring.

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
+++ b/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
@@ -259,23 +259,24 @@ fn write_sql_filter<V: Display>(
     sql: &mut String,
     field_name: impl Display,
     filter_range: &FilterRange<V>,
+    value_formatter: impl Fn(&V) -> String,
 ) {
     match &filter_range.start {
         Bound::Included(value) => {
-            let _ = write!(sql, " AND {} >= {}", field_name, value);
+            let _ = write!(sql, " AND {} >= {}", field_name, (value_formatter)(value));
         }
         Bound::Excluded(value) => {
-            let _ = write!(sql, " AND {} > {}", field_name, value);
+            let _ = write!(sql, " AND {} > {}", field_name, (value_formatter)(value));
         }
         Bound::Unbounded => {}
     };
 
     match &filter_range.end {
         Bound::Included(value) => {
-            let _ = write!(sql, " AND {} <= {}", field_name, value);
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -259,23 +259,24 @@ fn write_sql_filter<V: Display>(
     sql: &mut String,
     field_name: impl Display,
     filter_range: &FilterRange<V>,
+    value_formatter: impl Fn(&V) -> String,
```
fmassot, line 262: Nitpick: Using an enum is a bit better to avoid a typo next time we handle a filter that needs a conversion.

(if you are not aware of nitpick meaning, see: https://github.com/quickwit-oss/quickwit/blob/main/CODE_STYLE.md)

quickwit/quickwit-metastore/src/tests.rs
```diff
@@ -2064,6 +2070,61 @@ pub mod test_suite {
                     "list-splits-six",
                 ])
             );
+
+            let query =
+                ListSplitsQuery::for_index(index_id).with_update_timestamp_gte(current_timestamp);
```
fmassot, line 2075: Here we could run the test with `with_update_timestamp_gte(split_6_create_timestamp)` (just create the variable above when create the split 6), so we can check the filtering too.

fmassot (approved):
LGTM. Could be nice to improve the test on `update_timestamp_gte`.

PR #2292, opened 2022-11-11
Remove all `splits` table triggers from the Postgres metastore. 

ChillFish8:
While testing and profiling queries on a large database, I notice we still have a few triggers on the splits table that add some additional overhead to each update we perform on the table, although the overhead is not extreme, it is not minimal either.

This PR removes the triggers as we can perform all of the trigger's functions as part of the queries invoking the trigger, some of the logic with the `CASE` statements are probably unnecessary depending on the desired behaviour we want, but I thought it's best not to change the behaviour in this PR.

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
+++ b/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
@@ -173,25 +173,25 @@ async fn mark_splits_as_published_helper(
     if split_ids.is_empty() {
         return Ok(Vec::new());
     }
-    let publishable_states = [SplitState::Staged.as_str(), SplitState::Published.as_str()];
     let published_split_ids: Vec<String> = sqlx::query(
         r#"
         UPDATE splits
         SET
-            split_state = $1,
+            split_state = 'Published',
             update_timestamp = (CURRENT_TIMESTAMP AT TIME ZONE 'UTC'),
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -173,22 +173,29 @@ async fn mark_splits_as_published_helper(
...
+                WHEN split_state = 'Staged' THEN (CURRENT_TIMESTAMP AT TIME ZONE 'UTC')
+                ELSE publish_timestamp
+            END
         WHERE
-                index_id = $2
-            AND split_id = ANY($3)
-            AND split_state = ANY($4)
+                index_id = $1
+            AND split_id = ANY($2)
+            AND split_state IN ('Published', 'Staged')
```
ChillFish8, line 193: I removed the placeholders for the split states here because we hard code the `Staged` split state within the triggers anyway, and need also hard code it in the `CASE` statement, so to avoid confusion placeholders were removed.

guilload:
Can you quantify the performance improvement more precisely using Paul's branch?

IMO, the PR is still appealing even in the presence of a minor performance improvement or because it removes triggers, which are always painful to manage.

guilload (approved):

ChillFish8:
> Can you quantify the performance improvement more precisely using Paul's branch?

I shall have a go.

PR #2294, opened 2022-11-11
QW-2273: Add the ability to disable ANSI terminal codes in logging.

ChillFish8:
This adds the ability to pass a `--no-ansi-logging` flag or `QW_NO_ANSI_LOGGING` environment variable in order to disable tracing_subscriber from injecting ANSI terminal codes, reducing the noise when logging to things like files which do not support rendering the codes.

### DIFF
--- a/quickwit/quickwit-cli/src/cli.rs
+++ b/quickwit/quickwit-cli/src/cli.rs
@@ -40,9 +40,12 @@ pub fn build_cli<'a>() -> Command<'a> {
         )
         .arg(
             Arg::new("no-ansi")
-                .long("no-ansi-logging")
-                .help("Disable ANSI terminal codes being injected into the logging output")
-                .env("QW_NO_ANSI_LOGGING")
+                .long("no-color")
+                .help(
+                    "Disable ANSI terminal codes (colors, etc...) being injected into the logging \
+                     output",
+                )
+                .env("NO_COLOR")
                 .global(true)
                 .display_order(2)
                 .takes_value(false),
--- a/quickwit/quickwit-cli/src/cli.rs
+++ b/quickwit/quickwit-cli/src/cli.rs
@@ -41,10 +41,7 @@ pub fn build_cli<'a>() -> Command<'a> {
         .arg(
             Arg::new("no-ansi")
                 .long("no-ansi-logging")
-                .help(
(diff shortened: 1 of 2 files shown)


fmassot:
@ChillFish8 I think we can follow this standard, it's pretty standard:  https://no-color.org/

ChillFish8:
> @ChillFish8 I think we can follow this standard, it's pretty standard:  https://no-color.org/
> 
> 

That's true, I was unsure whether it's a good idea to call it `no color` or not as it affects all ANSI codes rather than just the colours, which could be miss leading.

But if everyone is using the same naming then it's probably worth changing.

fmassot:
mm, good point, so maybe `NO_ANSI` is better here.

ChillFish8:
I think we should go with  `NO_COLOR` for familiarity, I think if people want to disable ANSI codes it's probably because of the colours and the noise they bring. Probably better to be consistent with what others are doing.

fmassot:
I like the `NO_COLOR` too, let's go for this one :)

quickwit/quickwit-cli/src/cli.rs
```diff
@@ -38,6 +38,18 @@ pub fn build_cli<'a>() -> Command<'a> {
                 .global(true)
                 .display_order(1),
         )
+        .arg(
+            Arg::new("no-ansi")
```
fmassot, line 42:
```suggestion
            Arg::new("no-color")
```

quickwit/quickwit-cli/src/main.rs
```diff
@@ -130,7 +135,11 @@ async fn main() -> anyhow::Result<()> {
     #[cfg(feature = "jemalloc")]
     start_jemalloc_metrics_loop();
 
-    setup_logging_and_tracing(command.default_log_level(), &build_info)?;
+    setup_logging_and_tracing(
+        command.default_log_level(),
+        !matches.is_present("no-ansi"),
```
fmassot, line 140:
```suggestion
        !matches.is_present("no-clor"),
```

ChillFish8: Have changed this, thought it would make sense to keep it as ansi but it ended up looking more confusing 😅

fmassot (approved):

PR #2297, opened 2022-11-12
Remove transaction from `stage_split`

guilload:
### Description
- Remove transaction from `stage_split`
- Fix `update_index_update_timestamp`

### How was this PR tested?
Replayed gRPC requests and observed a 20-15s second improvement (70-75s vs. 90-100s)

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
+++ b/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
@@ -112,20 +112,25 @@ impl PostgresqlMetastore {
         })
     }
 
+    /// This function attempts to update an index update timestamp. Since we call this method after
+    /// a successful update of splits or delete tasks, we never return an error to not mislead the
+    /// clients into believing that the update failed. We log the error instead.
     #[instrument(skip(self))]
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -599,44 +611,51 @@ impl Metastore for PostgresqlMetastore {
...
             .bind(SplitState::Staged.as_str())
             .bind(time_range_start)
             .bind(time_range_end)
             .bind(tags)
             .bind(split_metadata_json)
             .bind(index_id)
-            .bind(metadata.delete_opstamp as i64)
-            .execute(&mut *tx)
+            .bind(split_metadata.delete_opstamp as i64)
+            .execute(&mut conn)
```
ChillFish8, line 648: I believe you can just pass the pool here as an immutable reference https://docs.rs/sqlx/latest/sqlx/pool/#using-a-connection-pool

guilload: That's nice!

fmassot (approved):

PR #2302, opened 2022-11-13
Implement grpc metatastore uri method.

fmassot:
Two main things here:
- I was forced to change the metastore uri  signature `uri() -> &Uri` into `uri() -> Uri` to implement it for the gRPC metastore. This gRPC metastore holds a `watch::Receiver<HashSet<QuickwitUri>>` to keep the current metastore URIs and returns the first element `uri()` is called. And we cannot return a value referencing a temporary value created in the `uri` method.
- I decided to add a `grpc` protocol variant to use it in`Uri`. I hesitated with an `http` variant which seems too broad when you look at the existing variants: `postgresql`, `s3`. I found grpc a bit more consistent.

### DIFF
--- a/quickwit/quickwit-common/src/uri.rs
+++ b/quickwit/quickwit-common/src/uri.rs
@@ -38,6 +38,7 @@ pub enum Protocol {
     PostgreSQL,
     Ram,
     S3,
+    Grpc,
 }
 
 impl Protocol {
@@ -48,6 +49,7 @@ impl Protocol {
             Protocol::PostgreSQL => "postgresql",
             Protocol::Ram => "ram",
             Protocol::S3 => "s3",
+            Protocol::Grpc => "grpc",
         }
     }
 
@@ -82,6 +84,10 @@ impl Protocol {
     pub fn is_database(&self) -> bool {
         matches!(&self, Protocol::PostgreSQL)
     }
+
+    pub fn is_grpc(&self) -> bool {
+        matches!(&self, Protocol::Grpc)
+    }
 }
 
 impl Display for Protocol {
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-common/src/uri.rs
```diff
@@ -38,6 +38,7 @@ pub enum Protocol {
     PostgreSQL,
     Ram,
     S3,
+    Http,
```
guilload, line 41:
```suggestion
    Http { secure: bool },
```

guilload: Shall we handle `https` as well?

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
```diff
@@ -497,8 +497,8 @@ impl Metastore for FileBackedMetastore {
         .await
     }
 
-    fn uri(&self) -> &Uri {
-        self.storage.uri()
+    fn uri(&self) -> Uri {
```
guilload, line 500: Why not let the caller decide whether they want to clone or not instead of changing the signature?

quickwit/quickwit-metastore/src/metastore/grpc_metastore/mod.rs
```diff
@@ -73,7 +73,7 @@ const CLIENT_TIMEOUT_DURATION: Duration = if cfg!(test) {
 #[derive(Clone)]
 pub struct MetastoreGrpcClient {
     underlying: MetastoreApiServiceClient<Timeout<Channel>>,
-    pool_size_rx: watch::Receiver<usize>,
+    metastore_uris_rx: watch::Receiver<Vec<QuickwitUri>>,
```
guilload, line 76: HashSet?

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
fmassot: Yes, I was going to add a comment on this because I was forced to do it :/

I did not manage to make it work with a dynamic Uri like the one in the grpc metastore

quickwit/quickwit-common/src/uri.rs
fmassot: I thought also about using `Grpc` protocol. For example, s3 is using HTTP requests so http seems a bit misleading here.

PR #2303, opened 2022-11-13
Wait 2 heartbeats before asserts.

fmassot:
Ok, this PR is lame... 

The alternative would be to have an assert with a timeout. Maybe we could add this as a util function for our tests. I'm not very fond of it though :/


guilload:
Yeah, this is super lame. Slow tests are painful.

guilload (approved):

fulmicoton:
Maybe we could do something that just accelerate time to the next scheduled event in unit test whenever no actor is in the  "processing" state?

fulmicoton (approved):

PR #2324, opened 2022-11-15
Fix `test_metastore_list_splits` flaky test

ChillFish8:
This PR adjusts the metastore test so that we're more forgiving towards the potential of network latency and similar affecting the GRPC metastore tests.

The flaky-ness of this test is due to the fact that we're comparing timestamps in a relatively short time frame while we're only accurate to 1 second.
This means under the right amount of network latency and time, it's possible for the sixth split to be within the range we're filtering out (which it shouldn't do).
The solution to this is to extend the time we wait for by another second (making it `3` seconds) and then filtering out splits which are older than `2` seconds ago so that there is no potential overlap between the sixth split's timestamp and the selection timestamp.

I've ran this several times now in loops and haven't observed this behaviour anymore after adjusting this.

### DIFF
--- a/quickwit/quickwit-metastore/src/tests.rs
+++ b/quickwit/quickwit-metastore/src/tests.rs
@@ -2821,9 +2821,7 @@ macro_rules! metastore_test_suite {
             #[tokio::test]
             async fn test_metastore_list_splits() {
                 let _ = tracing_subscriber::fmt::try_init();
-                for _ in 0..100 {
-                    crate::tests::test_suite::test_metastore_list_splits::<$metastore_type>().await;
-                }
+                crate::tests::test_suite::test_metastore_list_splits::<$metastore_type>().await;
             }
 
             #[tokio::test]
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-metastore/src/tests.rs
```diff
@@ -2104,8 +2104,8 @@ pub mod test_suite {
                 .collect();
             assert_eq!(split_ids, to_hash_set(&["list-splits-six"]));
 
-            let query =
-                ListSplitsQuery::for_index(index_id).with_create_timestamp_lte(select_timestamp);
+            let query = ListSplitsQuery::for_index(index_id)
+                .with_create_timestamp_lte(select_timestamp - 1);
```
guilload, line 2108: which is equivalent to `lt`.

guilload (approved):

ChillFish8:
> .with_update_timestamp_gte(split_metadata_6.create_timestamp); (first assert)
> .with_create_timestamp_lt(split_metadata_6.create_timestamp); (second assert)

I considered this but ran into issues with the Postgres implementation, which overwrites the timestamps we pass as part of the metadata, which caused problems in itself with flaky tests.

guilload:
Interesting, I'm not running into this issue. Tests are passing for the three implementations on my machine.

ChillFish8:
I'll have another look and see if I can replicate the behaviour again, if not I'll go with the aforementioned approach.

ChillFish8:
Re-running the tests repeatedly I'm no longer able to re-produce this so I'll go with your suggested approach :)

PR #2326, opened 2022-11-15
Implement liveness probes for indexer and janitor service

guilload:
### Description
Implement basic liveness probes for indexer and janitor service

### How was this PR tested?
Ran quickwit and killed the metastore while  `curl`ing the livness endpoint.

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexing_service.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexing_service.rs
@@ -23,7 +23,7 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use quickwit_actors::{
-    Actor, ActorContext, ActorExitStatus, ActorHandle, ActorState, Handler, Health, Mailbox,
+    Actor, ActorContext, ActorExitStatus, ActorHandle, ActorState, Handler, Healthz, Mailbox,
     Observation,
 };
 use quickwit_common::fs::get_cache_directory_path;
@@ -86,7 +86,7 @@ impl ServiceError for IndexingServiceError {
 }
 
 #[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
-pub struct IndexingServiceState {
+pub struct IndexingServiceCounters {
     pub num_running_pipelines: usize,
     pub num_successful_pipelines: usize,
     pub num_failed_pipelines: usize,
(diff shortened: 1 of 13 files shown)


guilload:
The probe endpoints being behind the `api/v1` prefix is a bit awkward. Shall I fix that?

fmassot:
> The probe endpoints being behind the `api/v1` prefix is a bit awkward. Shall I fix that?

Yes, good point, it was stupid to put them here.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -156,7 +151,7 @@ impl IndexingService {
             storage_resolver,
             local_split_store: Arc::new(local_split_store),
             indexing_pipeline_handles: Default::default(),
-            state: Default::default(),
+            counters: Default::default(),
```
fmassot, line 154: You can't resist, it's too strong :)

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -86,7 +86,7 @@ async fn run_postgres_migrations(pool: &Pool<Postgres>) -> MetastoreResult<()> {
         tx.rollback().await?;
         error!(err=?migration_err, "Database migrations failed");
         return Err(MetastoreError::InternalError {
-            message: "Failed to run migrator on Postgresql database.".to_string(),
+            message: "Failed to run migration on Postgresql database.".to_string(),
```
fmassot, line 89: nice catch :)

quickwit/quickwit-serve/src/rest.rs
```diff
@@ -37,15 +39,19 @@ use crate::search_api::{search_get_handler, search_post_handler, search_stream_h
...
     rest_listen_addr: SocketAddr,
+    universe: &Universe,
     quickwit_services: &QuickwitServices,
 ) -> anyhow::Result<()> {
     info!(rest_listen_addr = %rest_listen_addr, "Starting REST server.");
     let request_counter = warp::log::custom(|_| {
         crate::SERVE_METRICS.http_requests_total.inc();
     });
+
+    let delete_task_service_opt: Option<Mailbox<DeleteTaskService>> = universe.get_one();
```
fmassot, line 53: it's a bit weird to have this here alone, but I see the rationale: `delete_task_service_opt` is not a service as important as the others. So you did not include it in the health checks right? Thus this `get_one`.

fmassot: and now we also return a janitor service so it's harder to get this delete service. okokok

Issue #2199, opened 2022-10-30
Retention policy can potentially never delete some splits 


fmassot:
If we take the case where:
- we continuously make delete operations on some splits
- or we make some merges during a very long time on some splits

The retention policy with its criteria based on the publish timestamp will delete such splits in a very long time.

By setting a retention of a period of X days, a user would expect that X days after the publication of the first split, it would be deleted, without taking into account merges or deletes.

This issue came up to my mind when thinking about persisting maturity in the metastore. Right now, I don't see how we can avoid storing the timestamp at which the documents come in quickwit. Maybe `indexing_timestamp` could be the name.

evanxg852000:
@fmassot  @guilload 
After trying to see if we can do without adding new field:
1. I am thinking that the `publish_timestamp`  can play the same role as the new field we want to introduce `first_published_timestamp` , (potentially just rename it)
- During merge we just need keep the max `pusblish_timestamp`  
- During delete we keep the same  `pusblish_timestamp` 

The issue right now is that our metastore API does not expose a way to change the `publish_timestamp` . So this might be a lot of change we might not want.

2. Add the new field in SplitMetadata `first_published_timestamp: Option<i64>`.  
The issue is still how to update this value on the first publication of the Split since SplitMetadata is an opaque JSON.
My proposed solution is to avoid updating it when the split is first published. But derive it from `Split::pusblish_timestamp` if it's not set yet. This implies that for a split that has never gone through merge or delete, it's `first_published_timestamp` will be none so we can use the `Split::pusblish_timestamp` until it goes through merge or delete at which point a new Split is created with the  `first_published_timestamp`
In merge policy for instance that will result in:
```rust
let default_publish_timestamp = split.publish_timestamp;
let publish_timestamp = split.metadata.first_published_timestamp.unwrap_or(default_publish_timestamp);
retention_policy.is_eligible(publish_timestamp)
```

fmassot:
With the solution of letting the `first_published_timestamp` empty if no merge/delete means that, in some cases, we will have splits with no `first_published_timestamp` which is weird. Handling this missing values in the code will be confusing too.

I see 3 possible solutions to this:

1. Let the metastore handle the `publish_timestamp` like today and also `first_publish_timestamp`. This means that `first_publish_timestamp` is set at the same time `publish_timestamp` is there is no replaced splits. If there is some replaced splits, then take the max of their publish timestamp. 

[...]

PR #2327, opened 2022-11-15
Fix retention policy bug

evanxg852000:
Added `indexing_end_timestamp` field to SplitMetadata

I refrained from adding a new SplitMetadata version for the newly added field. Since we scrapped previous versions I thought this is acceptable until the release right?

Closes https://github.com/quickwit-oss/quickwit/issues/2199


quickwit/quickwit-metastore/migrations/postgresql/8_add-split-indexing-end-timestamp-field.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE splits
+    ADD COLUMN indexing_end_timestamp TIMESTAMP NOT NULL DEFAULT (CURRENT_TIMESTAMP AT TIME ZONE 'UTC');
```
evanxg852000, line 2: We could avoid all those Postgres conversions `INTERVAL '{} seconds'` or `to_timestamp(time_range_end)` by using `BIGINT` type. We then use  `extract(epoch from now())` to extract the current timestamp for comparison. Should we do that?

quickwit/quickwit-metastore/src/metastore/mod.rs
```diff
@@ -304,6 +305,9 @@ pub struct ListSplitsQuery<'a> {
 
     /// The create timestamp range to filter by.
     pub create_timestamp: FilterRange<i64>,
+
+    /// The age to filter by.
+    pub age_filter_opt: Option<AgeFilter>,
```
fmassot, line 310: Why not putting `pub indexing_end_timestamp: FilterRange<i64>`?

fmassot: No need to add a new type of filter.

quickwit/quickwit-metastore/src/metastore/mod.rs
```diff
@@ -464,6 +469,20 @@ impl<'a> ListSplitsQuery<'a> {
         self.create_timestamp.start = Bound::Excluded(v);
         self
     }
+
+    /// Set the age filter to match splits with `indexing_end_timestamp`
+    /// *older than or equal to* the provided duration.
+    pub fn with_age_on_indexing_end_timestamp(mut self, v: Duration) -> Self {
```
fmassot, line 475: I would just use the `FilterRange`.

quickwit/quickwit-metastore/migrations/postgresql/8_add-split-indexing-end-timestamp-field.up.sql
fmassot: I would not make this change now and remain consistent with other timestamps.

fmassot:
Ok @evanxg852000 I think we should stay on `FilterRange` stuff, we don't really want to add some new 'semantics' in the split filter struct. Just plain and simple filter.

quickwit/quickwit-metastore/migrations/postgresql/8_add-split-indexing-end-timestamp-field.up.sql
evanxg852000: Ok, I was just talking in terms of simplifying.

quickwit/quickwit-metastore/src/tests.rs
```diff
@@ -2142,6 +2142,114 @@ pub mod test_suite {
         }
     }
 
+    pub async fn test_metastore_list_splits_with_indexing_timestamp_and_time_range<
```
fmassot, line 2145: Nitpiick: I would prefer to do the test on published splits as this is the query that will be sent to the metastore.

docs/configuration/index-config.md
```diff
@@ -580,5 +580,5 @@ retention:
   - `years`, `year`, `y` -- a year is defined as `365.25 days`
 
 (2) `cutoff_reference` possible values:
-  - `publish_timestamp` will evaluate based on the timestamp the split was published at.
+  - `indexing_timestamp` will evaluate based on the timestamp the split was finalized at.
```
fmassot, line 583: we could be a bit clearer here. Something like 

"The retention policy will remove all splits with `formula`."

One thing is not clear too: we don't know what "split finalization" means. We should probably say something like "indexing_timestamp" corresopnds to the timestamp at which we stopped indexing document in the split.

quickwit/quickwit-indexing/src/models/split_attrs.rs
```diff
@@ -60,6 +60,9 @@ pub struct SplitAttrs {
 
     // Number of merge operation the split has been through so far.
     pub num_merge_ops: usize,
+
+    /// Timestamp for tracking when the split was finalized.
```
fmassot, line 64: Nitpick: if you find a better wording without using finalize as mentionned on the docs, it's nice.

```
/// Timestamp for tracking when indexing ends for the split.
```

quickwit/quickwit-janitor/src/retention_policy_execution.rs
```diff
@@ -41,21 +41,31 @@ pub async fn run_execute_retention_policy(
...
     ctx: &ActorContext<RetentionPolicyExecutor>,
 ) -> anyhow::Result<Vec<SplitMetadata>> {
-    // Select published splits and filter for expiration.
-    let current_date_time = OffsetDateTime::now_utc();
-
-    let query = ListSplitsQuery::for_index(index_id).with_split_state(SplitState::Published);
+    // Select splits that are published and older than the retention period.
+    let retention_period = retention_policy.retention_period()?;
+    let current_timestamp = OffsetDateTime::now_utc().unix_timestamp();
+    let retention_timestamp = current_timestamp - retention_period.as_secs() as i64;
```
fmassot, line 47: Nitpick

```suggestion
    let max_retention_timestamp = current_timestamp - retention_period.as_secs() as i64;
```

quickwit/quickwit-janitor/src/retention_policy_execution.rs
```diff
@@ -41,21 +41,31 @@ pub async fn run_execute_retention_policy(
...
+        RetentionPolicyCutoffReference::SplitTimestampField => {
+            base_query.with_time_range_end_lte(retention_timestamp)
+        }
+    };
 
-    let expired_splits: Vec<SplitMetadata> = ctx
+    let candidate_splits: Vec<SplitMetadata> = ctx
         .protect_future(metastore.list_splits(query))
         .await?
         .into_iter()
```
fmassot, line 61: Nitpick: I would have kept the `is_split_expired` filter.

quickwit/quickwit-janitor/src/retention_policy_execution.rs
```diff
@@ -70,145 +80,36 @@ pub async fn run_execute_retention_policy(
...
+        }
     }
 
-    #[test]
-    fn test_is_split_expired_with_time_range() {
-        let one_hour_retention_policy = RetentionPolicy::new(
-            "1 hour".to_string(),
-            RetentionPolicyCutoffReference::SplitTimestampField,
-            "hourly".to_string(),
+    if !not_applicable_splits.is_empty() {
```
fmassot, line 107: Nitpick: I think a warning on each split is ok and that the usage of a filter will make the code shorter. Now we need two vectors, one for expired splits, and one for non applicable splits, as a reader, I would have prefered a shorter code on this.

quickwit/quickwit-janitor/src/retention_policy_execution.rs
```diff
@@ -70,145 +80,36 @@ pub async fn run_execute_retention_policy(
...
-                "Retention policy execution expected a `time_range` on the split, but none exist.");
-                false
-            }
-        }
-    };
-    Ok(is_expired)
-}
-
-#[cfg(test)]
-mod tests {
```
fmassot, line 105: No need for a test?

evanxg852000: The tests here were doing what we now do in the metastore. I added a test on metastore to compensate.

quickwit/quickwit-metastore/src/tests.rs
evanxg852000: Ok, but really the state could even be removed as that is tested elsewhere. Here we are just using stage split because that is what creates the splits in the metastore.

quickwit/quickwit-janitor/src/retention_policy_execution.rs
evanxg852000: I thought it would too verbose to output every split as a new warning message and given that we are trying to avoid using so iterator pattern this made me choose to route.

fmassot:
> I refrained from adding a new SplitMetadata version for the newly added field. Since we scrapped previous versions I thought this is acceptable until the release right?

Agreed.

evanxg852000:
Moved changes to https://github.com/quickwit-oss/quickwit/pull/2378

PR #2329, opened 2022-11-16
Update tantivy to fix json path search escaping '.'

### LINKED ISSUE #2312: Searching in a json field  `path1.path2.path3:value` should return a result
Currently this does not work as tantivy is doing something slightly different when writing and when reading.

Let's take a document to index that is very common in the OpenTelemetry world:

```json
{"resource": {"k8s.container.name": "prometheus"}}
```

The doc_mapping is as follows:
```yaml
doc_mapping:
   - name: resource
      type: json

```

When writing the index, tantivy will interpret `k8s.container.name` as the field name and will store that as is along with the value (and the type).
When searching document matching the query `resource.k8s.container.name:prometheus`, quickwit will remove the `resource` part and give tantivy this term to match `k8s.container.name:prometheus`. The issue is that tantivy will interpret the dots as a separator and will build a term with an internal separator and this won't match what was written previously.

I suggest modifying a bit how tantivy writes the JSON terms by using the dots in field names to define segment path.

fulmicoton:
Also updated documentation, to explain how nested structure can be searched.

Closes #2312

Tested on quickiwt by indexing a single document 
`{"resource": {"k8s.container.name": "prometheus"}}`
and searched it with quickwit.

test_escape

### DIFF
--- a/docs/reference/query-language.md
+++ b/docs/reference/query-language.md
@@ -26,7 +26,7 @@ Note that a query like `title:barack obama` will find only `barack` in the title
 ### Searching structures nested in documents.
 
 Quickwit is designed to index structured data.
-If you search into some object nested into your document, whether it is an `object`, a `json` object, or a whether it was caught through the `dynamic` mode, the query language is the same. You simply need to chain the different steps to reach your value from the root of the document.
+If you search into some object nested into your document, whether it is an `object`, a `json` object, or whether it was caught through the `dynamic` mode, the query language is the same. You simply need to chain the different steps to reach your value from the root of the document.
 
 For instance, the document `{"product": {"attributes": {color": "red"}}}` is returned if you query `product.attributes.color:red`.
 If a dot `.` exists in one of the key of your object, you need to escape it.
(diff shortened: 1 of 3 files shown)


fmassot (approved):

docs/reference/query-language.md
```diff
@@ -23,6 +23,15 @@ title:"barack obama" AND president
 
 Note that a query like `title:barack obama` will find only `barack` in the title and `obama` in the default fields. If no default field has been set on the index, this will result in an error.
 
+### Searching structures nested in documents.
+
+Quickwit is designed to index structured data.
+If you search into some object nested into your document, whether it is an `object`, a `json` object, or a whether it was caught through the `dynamic` mode, the query language is the same. You simply need to chain the different steps to reach your value from the root of the document.
```
fmassot, line 29:
```suggestion
If you search into some object nested into your document, whether it is an `object`, a `json` object, or whether it was caught through the `dynamic` mode, the query language is the same. You simply need to chain the different steps to reach your value from the root of the document.
```

PR #2332, opened 2022-11-16
Add `mark_splits_for_deletion_by_query` method to simplify process.

ChillFish8:
### Description
This adds an additional method allowing us to mark splits as deleted where splits match a given `ListSplitsQuery`.

This has a few advantages over what we're currently doing in a couple places:
- Only requires a single metastore call rather than first listing all splits with a given query, and then a second query to actually issue the updates.
- Removes any additional network and disk pressure of reading a potentially very large set of rows from the metastore into memory.
- Allows us to optimise the operation a bit more on a per-metastore basis.

Currently I have only changed this to be used in the GC. But another place of interest is in the `run_execute_retention_policy` method which currently issues a `list_splits` request on *all* published splits, and then manually filters out the splits it wants, and then marks them for deletion. This could be completely replaced by this new method, but requires extending the query builder a bit we can do so.

### How was this PR tested?

- The GC tests have been refactored to account for the new operation and the queries provided checked.
- The method itself has a new unit test as part of the metastore test suite, most os this system shares tests with the `ListSplitsQuery` tests.


ChillFish8:
Closing PR as it's become stale, we can re-visit this issue later if we need to.

PR #2341, opened 2022-11-17
Added a backpressure counter within actors.

fulmicoton:


quickwit/quickwit-actors/src/mailbox.rs
```diff
@@ -157,13 +159,49 @@ impl<A: Actor> Mailbox<A> {
...
+    ) -> Result<oneshot::Receiver<A::Reply>, SendError>
     where
         A: Handler<M>,
         M: 'static + Send + Sync + fmt::Debug,
     {
         let (envelope, response_rx) = wrap_in_envelope(message);
-        self.inner.tx.send_low_priority(envelope).await?;
-        Ok(response_rx)
+        if let Some(backpressure_microsecs) = backpressure_microsecs_opt {
+            match self.inner.tx.try_send_low_priority(envelope) {
```
fulmicoton, line 186: A couple of notes here.

1- We use try_send to make the cost is minimal if we do not have any backpressure. 
2- We do not use `is_full()` because, `is_full() is true` even without backpressure, if the queue capacity is 0.

quickwit/quickwit-actors/src/mailbox.rs
```diff
@@ -157,13 +159,45 @@ impl<A: Actor> Mailbox<A> {
...
+
+    /// Sends a message to the actor owning the associated inbox.
+    ///
+    /// If the actor experiences some backpressure, then
+    /// `backpressure_microsecs` will be increased by the amount of
+    /// microseconds of backpressure experienced.
+    pub async fn send_message_with_backpressure_counter<M>(
+        &self,
+        message: M,
+        backpressure_microsecs_opt: Option<&IntCounter>,
```
fmassot, line 178: Nitpick: I would use the name `backpressure_microsecs_counter_opt`.

quickwit/quickwit-actors/src/mailbox.rs
```diff
@@ -320,4 +381,115 @@ mod tests {
...
+            .unwrap();
+        assert!(backpressure_microsecs_counter.get() > 1_000u64);
+    }
+
+    #[tokio::test]
+    async fn test_mailbox_waiting_for_processing_does_not_counter_as_backpressure() {
+        let universe = Universe::new();
+        let back_pressure_actor = BackPressureActor;
+        let (mailbox, _handle) = universe.spawn_builder().spawn(back_pressure_actor);
+        mailbox
```
fmassot, line 480: question: why do you need to send do `.ask_with_backpressure_counter(Duration::default(), None)` before executing the real test?

fmassot (approved):
LGTM. Can you remove the proto files changes before merging?

PR #2350, opened 2022-11-18
Add tracing context in grpc metastore and refactor.

fmassot:
Tested with jaeger.

Capture d’écran 2022-11-18 à 10 45 07


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -345,8 +345,8 @@ async fn leaf_search_single_split(
         quickwit_collector.requires_scoring(),
     )
     .await?;
+    let span = info_span!( "tantivy_search", split_id = %split.split_id);
```
fmassot, line 348: We lose the link to the parent span if the span is in placed in `run_cpu_intensive`.

evanxg852000 (approved):

PR #2351, opened 2022-11-18
QW-2320: Fix UI run search behavour.

### LINKED ISSUE #2320: UI when no search index is selected
If no index is selected in the combo box (none is selected by default, at least on my browser), if
I run a query, I get the following error

`Error: { "error": "Route not found" }`

The UI does emit a query, but the queyr is sent to:
,http://localhost:8280/search?query=*&max_hits=20

ChillFish8:
Closes #2320 

This fixes the issue described in #2320 and prevents confusing behaviour where you could clear the selected index, but it would still use the previously set value.
To correct this I've adjusted the check so that you can't run a search without an index id being set both in the `Run` button and the trigger from the editor itself. We also explicitly disable the `clear` button now on the index drop-down since we do not support the ability to run a search without a specific index id in the first place.


fmassot:
Ok you fixed the issue but you introduced a new one :)

If you have an existing index `otel-log-v0` and you want to open the page http://localhost:7280/ui/search?query=%0A%0A&index_id=otel-log-v0&max_hits=10&sort_by_field=-timestamp

you won't see the index name in the dropdown as you can see in the screenshot.

Capture d’écran 2022-11-21 à 09 26 45

ChillFish8:
Always nice to see Javascript being consistent. 

I've fixed the new bug, but intern I've had to remove the change on the auto complete so that it prevents people clearing the index but it still using the old value as I was unable to get the input to use the preset value when an existing index was provided.

fmassot (approved):
Sounds good!!!!

PR #2364, opened 2022-11-21
Remove i64 as Index timestamp field

evanxg852000:
Forcing the index timestamp field to be of type DateTime type.


fmassot:
@evanxg852000 we had a misunderstanding about this, we really want to merge this before the release. Can you do finish your PR today?

fmassot (approved):
LGTM, Thanks @evanxg852000 !!

PR #2366, opened 2022-11-21
Fix tokio runtime error on consumer drop

guilload:
### Description
Move the Kafka consumer entirely in the poll loop to ensure that `rd_kafka_consumer_close` is always executed in a blocking task.

### How was this PR tested?
I was not able to reproduce the bug we observed in production :/

### DIFF
--- a/quickwit/quickwit-indexing/src/source/kafka_source.rs
+++ b/quickwit/quickwit-indexing/src/source/kafka_source.rs
@@ -33,7 +33,7 @@ use quickwit_metastore::checkpoint::{
 };
 use rdkafka::config::{ClientConfig, RDKafkaLogLevel};
 use rdkafka::consumer::{
-    BaseConsumer, Consumer, ConsumerContext, DefaultConsumerContext, Rebalance, RebalanceProtocol,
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -545,11 +534,6 @@ impl Source for KafkaSource {
         _ctx: &SourceContext,
     ) -> anyhow::Result<()> {
         self.poll_loop_jh.abort();
-
-        let consumer = self.consumer.clone();
-        spawn_blocking(move || {
-            consumer.unsubscribe();
```
fulmicoton, line 551: so unsubscribe is not required? Kafka does it on drop?

fulmicoton (approved):

quickwit/quickwit-indexing/src/source/kafka_source.rs
guilload: I don't know, and the documentation is unclear. `close` is called on drop. `unsubscribe` removes the consumer from the group while leaving the connection opened, whereas `close` drops the connection.

We still call `unsubscribe` in the poll loop.

PR #2369, opened 2022-11-21
Fixes the merge planner.

### LINKED ISSUE #2348: Improve MergePlanner
Series of Long merges can stall the merge pipeline for a long time.

A lot of small splits are then needlessly there, impacting search.

Index partitioning amplify the problem considerably, aswe can very normally end up  piping up 100GB worth of merges.
(3 merge, commit every minute, 20MB/s ingestion).

# Solution

Run small merge with  a higher priority. This means two things.
We need to sort merge opportunity depending on their num_merge_ops, AND
avoid queueing all of them.

The latter part my seem very simple, but one issue is that we still want
the merge planner to queue merge opportunities, regardless of whether or not new splits are received.

fulmicoton:
Fixes the merge planner.
    
This PR tweaks the way the merge planner schedules its merge operation
in a subtle way to avoid adding a lot of large merge operations at the
same time... Hence stalling smaller, with a larger bang-for-the-buck
merge operations.
    
This is done by stopping queue merge operations as soon as push back is
experienced.
    

Closes #2348

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/merge_planner.rs
+++ b/quickwit/quickwit-indexing/src/actors/merge_planner.rs
@@ -132,6 +132,10 @@ impl MergePlanner {
         merge_policy: Arc<dyn MergePolicy>,
         merge_split_downloader_mailbox: Mailbox<MergeSplitDownloader>,
     ) -> MergePlanner {
+        let published_splits: Vec<SplitMetadata> = published_splits
+            .into_iter()
+            .filter(|split_metadata| belongs_to_pipeline(&pipeline_id, split_metadata))
+            .collect();
         let mut merge_planner = MergePlanner {
             pipeline_id,
             partitioned_young_splits: Default::default(),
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-indexing/src/actors/merge_planner.rs
```diff
@@ -144,47 +132,122 @@ impl MergePlanner {
...
+        //
+        // The reason is that in presence of partitioning, it is very possible
+        // to receive a set of splits opening the opportunity to run a lot "large" merge
+        // operations at the same time.
+        //
+        // These large merge operation will in turn prevent small merge operations
+        // from being executed, when in fact small merge operations should be executed
+        // in priority.
+        //
+        // As an alternative approach, this function push merge operations until it starts
```
guilload, line 208:
```suggestion
        // As an alternative approach, this function pushes merge operations until it starts
```

quickwit/quickwit-indexing/src/actors/merge_planner.rs
```diff
@@ -144,47 +132,122 @@ impl MergePlanner {
...
+        // experience some push back, and then just "loops".
```
guilload, line 209:
```suggestion
        // experiencing some push back, and then just "loops".
```

guilload (approved):
Very nice!

PR #2378, opened 2022-11-22
Fix retention policy bug

evanxg852000:
Removed  `RetentionPolicy.cuttoff_reference` 

Closes https://github.com/quickwit-oss/quickwit/issues/2374
Related to https://github.com/quickwit-oss/quickwit/issues/2199

### DIFF
--- a/quickwit/quickwit-janitor/src/retention_policy_execution.rs
+++ b/quickwit/quickwit-janitor/src/retention_policy_execution.rs
@@ -20,8 +20,8 @@
 use std::sync::Arc;
 
 use quickwit_actors::ActorContext;
-use quickwit_config::{RetentionPolicy, RetentionPolicyCutoffReference};
-use quickwit_metastore::{ListSplitsQuery, Metastore, Split, SplitMetadata, SplitState};
+use quickwit_config::RetentionPolicy;
+use quickwit_metastore::{ListSplitsQuery, Metastore, SplitMetadata, SplitState};
 use time::OffsetDateTime;
 use tracing::{info, warn};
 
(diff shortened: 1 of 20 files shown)


quickwit/quickwit-janitor/src/retention_policy_execution.rs
```diff
@@ -41,22 +41,21 @@ pub async fn run_execute_retention_policy(
...
-    let current_date_time = OffsetDateTime::now_utc();
-
-    let query = ListSplitsQuery::for_index(index_id).with_split_state(SplitState::Published);
+    // Select splits that are published and older than the retention period.
+    let retention_period = retention_policy.retention_period()?;
+    let current_timestamp = OffsetDateTime::now_utc().unix_timestamp();
+    let max_retention_timestamp = current_timestamp - retention_period.as_secs() as i64;
+    let query = ListSplitsQuery::for_index(index_id)
+        .with_split_state(SplitState::Published)
+        .with_time_range_end_lte(max_retention_timestamp);
```
guilload, line 50: Nice side effect of the new spec.

quickwit/quickwit-janitor/src/retention_policy_execution.rs
```diff
@@ -41,22 +41,21 @@ pub async fn run_execute_retention_policy(
...
         .protect_future(metastore.list_splits(query))
         .await?
         .into_iter()
-        // TODO: possibly move this in DB query
-        .filter(|split| {
-            is_split_expired(current_date_time, split, retention_policy).unwrap_or(false)
-        })
-        .map(|meta| meta.split_metadata)
+        .map(|split| split.split_metadata)
+        .filter(is_split_expired)
```
guilload, line 57: Can we filter out splits without time range separately and emit a single warning?

guilload (approved):
LGTM. Thanks for fixing the typos along the way.

quickwit/quickwit-janitor/src/retention_policy_execution.rs
evanxg852000: We had that but given that it needed a separate vector, @fmassot and I found this way better and concluded it was ok to warn on each split.

PR #2385, opened 2022-11-22
delete segment when last document get deleted

### LINKED ISSUE #1960: Handle the case where delete query remove all documents in a split
In this case, we just want to remove the split.

trinity-1686a:
### Description

fix #1960 

### How was this PR tested?

I created an index with a single document repeated thousands of times and one other document. On delete, the split containing only repeated document get marked for deletion, the split with one document left is... left with one document.

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/merge_executor.rs
+++ b/quickwit/quickwit-indexing/src/actors/merge_executor.rs
@@ -366,11 +366,20 @@ impl MergeExecutor {
         ctx.record_progress();
 
         // Compute merged split attributes.
-        let merged_segment = merged_index
-            .searchable_segments()?
-            .into_iter()
-            .next()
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-indexing/src/actors/merge_executor.rs
```diff
@@ -366,11 +366,20 @@ impl MergeExecutor {
...
-            .searchable_segments()?
-            .into_iter()
-            .next()
-            .ok_or_else(|| anyhow!("Delete operation should output one segment."))?;
+        let merged_segment =
+            if let Some(segment) = merged_index.searchable_segments()?.into_iter().next() {
+                segment
+            } else {
+                info!(
+                    "All document from split `{}` were deleted.",
```
fmassot, line 374:
```suggestion
                    "All documents from split `{}` were deleted.",
```

PR #2386, opened 2022-11-22
update documentation and fix incorrect help message

trinity-1686a:


guilload (approved):

docs/reference/query-language.md
```diff
@@ -58,6 +58,10 @@ Quickwit also supports phrase queries with a slop parameter using the slop opera
 Slop queries can only be used on field indexed with the [record option](./../configuration/index-config.md#text-type) set to `position` value.
 :::
 
+### Set Operator
+
+Quickwit supports `IN [value1 value2 ...]` as a set appartenance operator. This is more cpu efficient than the equivalent `OR`ing of many terms, but may download more of the split than `OR`ing, especially when only a few terms are searched. You must specify a field being searched for Set queries.
```
ChillFish8, line 63: Unfortunately `appartenance` doesn't really have any great translation from french to English, it looks like it translates to `membership` but that might not reflect what you're trying to describe here.

trinity-1686a: Wikipedia is saying "set membership" is actually what I meant (https://en.wikipedia.org/wiki/Element_(mathematics)). Thanks for spotting the frenchism

PR #2390, opened 2022-11-22
Add Kubernetes deployment docs

guilload:


docs/deployment/kubernetes.md
```diff
@@ -1,6 +1,33 @@
...
+
+2. Add the Quickwit Helm chart repository to Helm:
+
+    helm repo add quickwit https://helm.quickwit.io
+
+3. Update the repository
+
+    helm repo update
+
+4. Create and customize your configuration file `values.yaml`
```
fmassot, line 22: What seems to be useful here is that you can use `helm show values [CHART] [flags]` to get the default values file. At least, I found it useful when install the kube-prometheus and otel charts.

fmassot (approved):

PR #2392, opened 2022-11-22
QW-2308: Adjust indexing service to let CLI detach merge pipeline.

### LINKED ISSUE #2308: The `ingest` CLI command does not merge splits
When creating a set of splits with the following index config (index_uri omitted) we see no merges being attempted by Quickwit while indexing documents via the `ingest` command.

We can observe the merge pipeline being started in the logs:
```
2022-11-14T15:09:57.681Z  INFO spawn_pipeline{^[[3mindex^[[0m^[[2m=^[[0mtest-index ^[[3mgen^[[0m^[[2m=^[[0m0}: quickwit_indexing::actors::indexing_pipeline: Spawning indexing pipeline. index_id=test-index source_id=_cli-ingest-source pipeline_ord=0 root_dir=/home/chillf>
```

And then see it very shortly after being killed with the following log:
```
2022-11-14T15:10:00.602Z  INFO quickwit_indexing::actors::indexing_service: No more indexing pipeline on this index and source, killing merge pipeline. index_id=test-index source_id=_cli-ingest-source
```

This does not get restarted and leaves us with hundreds of thousands of small splits that need to be merged.

#### Index Config
```yaml
version: 0

index_id: "test-index"

doc_mapping:
  mode: lenient
  field_mappings:
    - name: user
      type: text
      tokenizer: raw
    - name: item
      type: text
      tokenizer: raw
    - name: rating
      type: f64
      fast: true
    - name: timestamp
      type: i64
      fast: true

  partition_key: user
  max_num_partitions: 100

indexing_settings:
  timestamp_field: timestamp
  commit_timeout_secs: 1

  resources:
    num_threads: 1


search_settings:
  default_search_fields: [user, item]
```

ChillFish8:
Closes #2308 

### Description
This adds a new handler for detaching a given merge pipeline, which lets us detach both the indexing and merge pipelines to avoid the indexing service from killing the merge pipeline.

This also inadvertently fixed a small bug where merge pipeline counters were not set until the actor tries to clean them up.

### How was this PR tested?
Existing tests have been adjusted to use the new way of spawning merge pipelines, and the new handler has got a test.

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexing_service.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexing_service.rs
@@ -41,9 +41,9 @@ use tracing::{error, info};
 use super::merge_pipeline::{MergePipeline, MergePipelineParams};
 use super::MergePlanner;
 use crate::models::{
-    DetachPipeline, IndexingDirectory, IndexingPipelineId, Observe, ObservePipeline,
-    ShutdownPipeline, ShutdownPipelines, SpawnMergePipeline, SpawnPipeline, SpawnPipelines,
-    WeakIndexingDirectory,
+    DetachIndexingPipeline, DetachMergePipeline, IndexingDirectory, IndexingPipelineId, Observe,
+    ObservePipeline, ShutdownPipeline, ShutdownPipelines, SpawnMergePipeline, SpawnPipeline,
+    SpawnPipelines, WeakIndexingDirectory,
 };
 use crate::split_store::{LocalSplitStore, SplitStoreQuota};
 use crate::{IndexingPipeline, IndexingPipelineParams, IndexingSplitStore, IndexingStatistics};
@@ -96,8 +96,8 @@ pub struct IndexingServiceCounters {
 type IndexId = String;
 type SourceId = String;
 
-#[derive(Clone, Hash, Eq, PartialEq)]
-struct MergePipelineId {
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+pub struct MergePipelineId {
     index_id: String,
     source_id: String,
 }
@@ -171,6 +171,21 @@ impl IndexingService {
             })?;
         self.counters.num_running_pipelines -= 1;
         Ok(pipeline_handle)
+    }
+
+    async fn detach_merge_pipeline(
+        &mut self,
+        pipeline_id: &MergePipelineId,
+    ) -> Result<ActorHandle<MergePipeline>, IndexingServiceError> {
+        let pipeline_handle = self
+            .merge_pipeline_handles
+            .remove(pipeline_id)
+            .ok_or_else(|| IndexingServiceError::MissingPipeline {
+                index_id: pipeline_id.index_id.clone(),
+                source_id: pipeline_id.source_id.clone(),
+            })?;
+        self.counters.num_running_merge_pipelines = 1;
+        Ok(pipeline_handle.handle)
     }
 
     async fn observe_pipeline(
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-indexing/src/models/indexing_service_message.rs
```diff
@@ -52,7 +52,15 @@ pub struct ShutdownPipeline {
...
+pub struct DetachIndexingPipeline {
+    pub pipeline_id: IndexingPipelineId,
+}
+
+#[derive(Debug)]
+/// Detaches a merge pipeline from the indexing service. The pipeline is no longer managed by the
+/// server. This is mostly useful for preventing the server killing an existing merge pipeline
+/// if a indexing pipeline is detached.
+pub struct DetachMergePipeline {
     pub pipeline_id: IndexingPipelineId,
```
ChillFish8, line 64: This should really be `MergePipelineId` but in other applications like spawning a merge pipeline, we use an `IndexingPipelineId` so for now in this WIP I've left it as an `IndexingPipelineId` which then gets converted to a `MergePipelineId`

fmassot: Ok, this is weird indeed, the spawn merge pipeline seems to be related to the way we used to spawn pipelines: one merge pipeline for one indexing pipeline.

If we look at `spawn_merge_pipeline` function, it is calling `spawn_pipeline_inner`... so really this is not great. Can you just remove the `spawn_merge_pipeline` and `SpawnMergePipeline` which is useless?

fmassot: And then you can really detach a merge pipeline with a `MergePipelineId`.

ChillFish8: Sounds like a plan, I noticed we do something equally weird as part of the merge CLI where we create a merge pipeline, then detach the resultant indexing pipeline, which we should probably cleanup as part of this.

fmassot: Yes the merge CLI is also somewhat broken. A plan for the merge CLI would be:
- start the indexing pipeline
- detach the merge pipeline (I'm not sure we need to detach the indexing pipeline)
- use observe command to get the `MergePlannerState` (it's easy to get a given actor from the universe and send him the observe message). If the `state.ongoing_merge_operations` are not empty, wait a 1 second, observe again, etc.

We can do that in a separated PR.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -955,17 +968,13 @@ mod tests {
...
-                .num_running_merge_pipelines,
-            0
-        );
+
+        // Let the service cleanup the merge pipelines.
+        universe.simulate_time_shift(HEARTBEAT).await;
+
+        let observation = indexing_server_handle.process_pending_and_observe().await;
+        assert_eq!(observation.num_running_pipelines, 0);
+        assert_eq!(observation.num_running_merge_pipelines, 0);
```
ChillFish8, line 977: Interestingly, the only reason why this unit test passed was that we never increment the `running_merge_pipelines` counter, instead, it only gets set when the actor runs through and cleans up any unused mergers. This meant that when it observed the service it was always `0` and therefore passed.

quickwit/quickwit-indexing/src/models/indexing_service_message.rs
ChillFish8: I've gone through and cleaned all this up now 👍 The merge CLI has been slightly adjusted to use `spawn_pipeline` so we can remove spawn_merge_pipeline.

fmassot (approved):

PR #2402, opened 2022-11-23
QW-2289: Add bulk stage splits method to metastores.

### LINKED ISSUE #2289: Statge splits from a batch in an single metastore call
With 100 partitions, we do 100 `stage_split` requests. We could do it in one single call.

ChillFish8:
Closes #2289 

### Description
This PR adds a new method allowing us to stage multiple splits in a bulk operation.

Notes
_The 'bulk' part of this statement is down to the individual metastore implementation. E.g. The file-backed metastore just uses the default implementation of iterating through each split metadata individually as it does not have a more efficient bulk operation._

### How was this PR tested?
Unit tests have been adjusted and the new method has new tests.


fmassot:
I'm craving for this :)

ChillFish8:
This has proven to be way more cursed than I initially thought, as SQLx has very bad support for bulk insert operations, along with not supporting arrays which are more than 1 dimensional, which has made unnesting even more fun :)

quickwit/quickwit-metastore/src/metastore/mod.rs
```diff
@@ -141,6 +141,19 @@ pub trait Metastore: Send + Sync + 'static {
         split_metadata: SplitMetadata,
     ) -> MetastoreResult<()>;
 
+    /// Stages multiple splits.
+    async fn stage_splits(
```
guilload, line 145: This is not atomic. Remove. Let's remove `stage_split` as well.

guilload: After discussing with @fulmicoton, we realized that we'd like this operation to be nilpotent. Can you update the doc string of this method and implement the corresponding behavior?

quickwit/quickwit-metastore/src/error.rs
```diff
@@ -48,6 +48,13 @@ pub enum MetastoreError {
     #[error("Failed to deserialize index metadata: `{message}`")]
     InvalidManifest { message: String },
 
+    /// Something has gone very wrong.
+    /// In theory this error should never be returned as doing so
+    /// indicates that the operation is attempting to perform an action
+    /// on the metastore state that would violates the integrity of the metastore.
+    #[error("Cannot complete operation as it is incorrect: `{message}`")]
+    CorrectnessError { message: String },
```
ChillFish8, line 56: I'm not really happy with the description or name for this error, but I couldn't think of anything more appropriate.

ChillFish8: Extending on this as well, we could potentially return the failed split ids as part of this error but the behaviours between the Postgres and file-backed metastore would be different unless we made the file-backed store eagerly try update the splits and then return the error.

quickwit/quickwit-indexing/src/actors/uploader.rs
```diff
@@ -295,29 +295,63 @@ impl Handler<PackagedSplitBatch> for Uploader {
...
+                    split_metadata_list.push(split_metadata);
+                }
+
+                metastore
+                    .stage_splits(&index_id, split_metadata_list.clone())
+                    .instrument(tracing::info_span!("staging_split"))
+                    .await?;
+                counters.num_staged_splits.fetch_add(split_metadata_list.len() as u64, Ordering::SeqCst);
+
+                let mut packaged_splits_and_metadata = Vec::new();
```
guilload, line 327:
```suggestion
                let mut packaged_splits_and_metadata = Vec::with_capacity(batch.splits.len());
```

quickwit/quickwit-metastore/src/metastore/mod.rs
```diff
@@ -130,15 +130,19 @@ pub trait Metastore: Send + Sync + 'static {
 
     // Split API
 
-    /// Stages a split.
+    /// Stages multiple splits.
+    ///
+    /// If a split already exists and is not in the `Staged` state, a `CorrectnessError`
```
guilload, line 135:
```suggestion
    /// If a split already exists and is not in the `Staged` state, a `SplitsNotStaged` error
```

quickwit/quickwit-indexing/src/actors/uploader.rs
```diff
@@ -295,29 +295,63 @@ impl Handler<PackagedSplitBatch> for Uploader {
...
+                        packaged_split.tags.clone(),
+                        split_streamer.footer_range.start as u64..split_streamer.footer_range.end as u64,
+                    );
+
+                    split_metadata_list.push(split_metadata);
+                }
+
+                metastore
+                    .stage_splits(&index_id, split_metadata_list.clone())
+                    .instrument(tracing::info_span!("stage_splits"))
```
fulmicoton, line 323: I don't think we need this instrument.
It is already in the metastore implementation as an annotation.

```suggestion
```

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index/mod.rs
```diff
@@ -165,20 +165,15 @@ impl FileBackedIndex {
...
-                    "Failed to stage split  `{}`: split already exists.",
-                    split_metadata.split_id()
-                ),
-                cause: "".to_string(),
-            });
+        // If the split exists, we check what state it is in. If it's anything other than `Staged`
+        // something has gone very wrong and we should abort the operation.
+        if let Some(existing_split) = self.splits.get(split_metadata.split_id()) {
+            if existing_split.split_state != SplitState::Staged {
+                return false;
```
fulmicoton, line 175: The contract in the method doc is imprecise.
If anything it does into seem to describe the current implementation.

It suggest the return value is similar to a return bool of `Set::add`.
If the value was already there `return false`, if it was not `return true`.

In other methods, this is also what it means, it is used to identify whether it is necessary or not to write the file.

In your new implem, it is used to mark a failure, which seems like a poor choice.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -571,46 +571,96 @@ impl Metastore for PostgresqlMetastore {
...
+                .map(|range| *range.start());
+            let time_range_end = split_metadata.time_range.map(|range| *range.end());
+            let tags: Vec<String> = split_metadata.tags.into_iter().collect();
+
+            split_ids.push(split_metadata.split_id);
+            time_range_start_list.push(time_range_start);
+            time_range_end_list.push(time_range_end);
+            tags_list.push(sqlx::types::Json(tags));
+            split_metadata_json_list.push(split_metadata_json);
+            delete_opstamps.push(split_metadata.delete_opstamp as i64);
```
fulmicoton, line 606: Can you reorder the lines to make more sense?

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -571,46 +571,96 @@ impl Metastore for PostgresqlMetastore {
...
+                    time_range_start,
+                    time_range_end,
+                    ARRAY(SELECT json_array_elements_text(tags_json::json)) as tags,
+                    split_metadata_json,
+                    delete_opstamp,
+                    $7 as split_state,
+                    $8 as index_id
+                FROM
+                    unnest($1, $2, $3, $4, $5, $6)
+                    as tr(split_id, time_range_start, time_range_end, tags_json, split_metadata_json, delete_opstamp)
```
fulmicoton, line 624: ah I see what you meant in the comments about sqlx not supporting batch insert now T_T

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -571,46 +571,96 @@ impl Metastore for PostgresqlMetastore {
...
+                    split_id,
+                    time_range_start,
+                    time_range_end,
+                    ARRAY(SELECT json_array_elements_text(tags_json::json)) as tags,
+                    split_metadata_json,
+                    delete_opstamp,
+                    $7 as split_state,
+                    $8 as index_id
+                FROM
+                    unnest($1, $2, $3, $4, $5, $6)
```
fulmicoton, line 623:
```suggestion
                    UNNEST($1, $2, $3, $4, $5, $6)
```

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -571,46 +571,96 @@ impl Metastore for PostgresqlMetastore {
...
+                    as tr(split_id, time_range_start, time_range_end, tags_json, split_metadata_json, delete_opstamp)
+                ON CONFLICT(split_id) DO UPDATE
```
fulmicoton, line 625: TIL about `ON CONFLICT`, neato

fulmicoton (changes requested):
One minor comment.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
ChillFish8: I've shuffled them around so that they're grouped rather than separated.

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index/mod.rs
ChillFish8: Switched to returning and matching on a `MetastoreError`, docs have been updated.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
ChillFish8: The bigger issue was dealing with the fact that SQLx only supports 1 dimensional arrays 😅

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -571,46 +571,99 @@ impl Metastore for PostgresqlMetastore {
         Ok(())
     }
 
-    #[instrument(skip(self, split_metadata), fields(index_id=index_id, split_id=split_metadata.split_id))]
```
fulmicoton, line 574: so we will lose the logging of the `split_ids` here. Do we have it somewhere else?

split_ids are VERY useful when debugging.
Can you readd split_ids one way or another or add an issue to do it?

fulmicoton (approved):

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
ChillFish8: Added!

Issue #2338, opened 2022-11-17
Refactor tantivy::Value and serde_json::Value


evanxg852000:
We currently have imports of Value from both `tantivy::Value` and `serde_json::Value`. To facilitate reading the code base, we need to alias these types when importing them:
- `tantivy::Value as TantivyValue`
- `serde_json::Value as JsonValue`

PR #2404, opened 2022-11-24
Differentiate serde_json-value from tantivy-value

evanxg852000:
Make a  clear difference `serde_json::Value` from `tantivy::schema::Value`

Closes https://github.com/quickwit-oss/quickwit/issues/2338


guilload (approved):
Cool.

PR #2406, opened 2022-11-24
Remove transaction from `delete_splits`

guilload:
### Description
Remove transaction from `delete_splits` and update index update timestamp in the same query. 

In addition, I have modified the behavior of the method:
- Attempting to delete staged splits returns an error
- We no longer return an error when some splits are not found (but we log a warning)

### How was this PR tested?
Updated unit test.

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
+++ b/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
@@ -28,6 +28,7 @@ use std::time::Duration;
 use async_trait::async_trait;
 use itertools::Itertools;
 use quickwit_common::uri::Uri;
+use quickwit_common::PrettySample;
 use quickwit_config::SourceConfig;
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -765,37 +766,76 @@ impl Metastore for PostgresqlMetastore {
...
+                    SELECT split_id, split_state
+                    FROM splits
+                    WHERE
+                        index_id = $1
+                        AND split_id = ANY($2)
+                    FOR UPDATE
+                    ) AS splits
+                USING (split_id)
+            ),
+            -- Delete the splits iff all the splits are marked for deletion.
```
fulmicoton, line 785:
```suggestion
            -- Delete the splits if and only if all the splits are marked for deletion.
```

I love iff too, but several developers reading my comments thought it was a spelling mistake.

fulmicoton (approved):

PR #2433, opened 2022-11-27
Remove transaction from marked splits for deletion

guilload:
### Description
- Remove transaction from marked splits for deletion
- Display a warning when some splits are not found rather than returning an error

### How was this PR tested?
- Updated and completed unit test
- Ran `make test-all`


quickwit/quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index/mod.rs
```diff
@@ -199,6 +199,7 @@ impl FileBackedIndex {
         &mut self,
         split_ids: &[&str],
         deletable_states: &[SplitState],
+        return_error_on_splits_not_found: bool,
```
fulmicoton, line 202: why did you make it an argument?

fulmicoton (approved):

PR #2436, opened 2022-11-27
Removes error log when actor exits.

### LINKED ISSUE #2426: Remove error log message upon QUIT
Right now when indexin with the CLI we get the following ERROR log line

```
2022-11-27T11:48:07.977Z  INFO quickwit_actors::spawn_builder: actor-exit actor_id=MergePipeline-morning-ePOA exit_status=Quit
2022-11-27T11:48:07.977Z ERROR quickwit_actors::actor: actor-failure actor_name="MergePipeline-morning-ePOA" actor_exit_status=Quit
```

fulmicoton:
The entire logging is removed, since it was redundant with the logging spawn_builder.

Closes #2426

### DIFF
--- a/quickwit/quickwit-actors/src/actor.rs
+++ b/quickwit/quickwit-actors/src/actor.rs
@@ -424,9 +424,6 @@ impl<A: Actor> ActorContext<A> {
     }
 
     pub(crate) fn exit(&self, exit_status: &ActorExitStatus) {
-        if !exit_status.is_success() {
-            error!(actor_name=self.actor_instance_id(), actor_exit_status=?exit_status, "actor-failure");
-        }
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-actors/src/actor.rs
```diff
@@ -424,9 +424,6 @@ impl<A: Actor> ActorContext<A> {
     }
 
     pub(crate) fn exit(&self, exit_status: &ActorExitStatus) {
-        if !exit_status.is_success() {
-            error!(actor_name=self.actor_instance_id(), actor_exit_status=?exit_status, "actor-failure");
```
fulmicoton, line 428: that logging already happens in a more accurate way in spawn_builder

PR #2451, opened 2022-11-28
Document supported hardware architectures

### LINKED ISSUE #1651: Document supported hardware architectures
Hi!

Could you please put the information about supported architectures to the documentation please? E.e. about supported architectures for different operating systems, some specific requirements to the supported instructions, if you have any (e.g. maybe AVX is required - I do not know). This kind of information is important for the end-users.

The only thing that I have found - info about prebuilt binaries. But prebuilt binary that is not the same as supported architecture. If in your case it it is - please write a note about it somewhere in the documentation.

Thanks in advance!

PSeitz:
Closes #1651

### DIFF
--- a/docs/get-started/installation.md
+++ b/docs/get-started/installation.md
@@ -8,6 +8,13 @@ import TabItem from '@theme/TabItem';
 
 Quickwit compiles to a single binary, we provide different methods to install it.
 
+## Prerequisites
+
+Quickwit is officially only supported for Linux. Freebsd and MacOS are not officially supported, but should work as well.
+
+Supported CPU architectures are x86-64 and ARMv7.
+No special instruction set is required, but on x86-64 SSE3 is recommended.
+
 ## Download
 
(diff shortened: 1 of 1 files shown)


guilload (approved):

docs/get-started/installation.md
```diff
@@ -8,6 +8,13 @@ import TabItem from '@theme/TabItem';
 
 Quickwit compiles to a single binary, we provide different methods to install it.
 
+## Prerequisites
+
+Quickwit is officially only supported for Linux. Freebsd and MacOS are not officially supported, but should work as well.
+
+Supported CPU architectures are x86-64 and ARMv7.
+No special instruction set is required, but on x86-64 SSE3 is recommended.
```
fulmicoton, line 16:
```suggestion
Quickwit is officially only supported for Linux. Freebsd and MacOS are not officially supported, but should work as well.

Quickwit only supplies binaries for x86-64. No special instruction set is required, but on x86-64 SSE3 is recommended. Quickwit compiles on Armv8, but we do not publish binaries for this platform at the moment and its usage is experimental.
```

Issue #2412, opened 2022-11-25
CLI ingest source  checkpoint not cleared


evanxg852000:
When and index is cleared manually or via overwrite in  `cargo run -- index ingest --index wikipedia --overwrite`, the checkpoint entry for the source `_cli-ingest-source` is not removed. This happens because we currently don't track the `_cli-ingest-source` source in the index sources list. 

We could do the same we did for the INGEST_API_SOURCE_ID. That is explicitly add the `_cli-ingest-source` source when we create an index.

PR #2463, opened 2022-11-28
Fix index cleanup

evanxg852000:
Added default cli-ingest-source to index metadata when creating an index.
Closes https://github.com/quickwit-oss/quickwit/issues/2412


guilload:
Unlike the ingest API source, the CLI source is not a first-class citizen and is seldom used. In the future, we want to deprecate it and push the data directly to the ingest API. As such, I don't think this PR is worth it, and I'd rather fix `IndexService::clear_index` directly for the time being:

```rust
for source_id in index_metadata.sources.keys() {
    self.metastore
        .reset_source_checkpoint(index_id, source_id)
        .await?;
}
// The CLI source ...
self.metastore
    .reset_source_checkpoint(index_id, CLI_INGEST_SOURCE_ID)
    .await?;
```

quickwit/quickwit-config/src/source_config/mod.rs
```diff
@@ -102,6 +102,16 @@ impl SourceConfig {
...
         }
     }
+
+    /// Creates the default cli-ingest source config.
+    pub fn cli_ingest_source() -> SourceConfig {
+        SourceConfig {
+            source_id: CLI_INGEST_SOURCE_ID.to_string(),
+            num_pipelines: 1,
+            enabled: true,
+            source_params: SourceParams::void(),
```
guilload, line 112: Might be worth creating a `Cli` source type not to confuse users who run the `source list` command.

quickwit/quickwit-cli/src/source.rs
```diff
@@ -337,7 +339,7 @@ async fn toggle_source_cli(args: ToggleSourceArgs) -> anyhow::Result<()> {
     let metastore = quickwit_metastore_uri_resolver()
         .resolve(&config.metastore_uri)
         .await?;
-    if args.source_id == INGEST_API_SOURCE_ID {
+    if args.source_id == INGEST_API_SOURCE_ID || args.source_id == CLI_INGEST_SOURCE_ID {
```
guilload, line 342: We may want to align the naming of the IDs now that we're going to display both. Right now, we have `_cli-ingest-source` and `_ingest-api`. 

Your choosing ;)

guilload (approved):

quickwit/quickwit-config/src/source_config/mod.rs
evanxg852000: Yes, we could also make `SourceParams::stdin()` point to a `FileSourceParams{ file_path: Some("/dev/stdin") }` instead of the current None value. 
What do you think?

quickwit/quickwit-cli/src/source.rs
evanxg852000: I am willing to go with  change `_ingest-api` -> `_api-ingest-source`

quickwit/quickwit-config/src/source_config/mod.rs
guilload: Sounds good.

quickwit/quickwit-cli/src/source.rs
guilload: Sounds good as well.

quickwit/quickwit-config/src/source_config/mod.rs
evanxg852000: Finally went with your solution. It looked bizarre to see Type `file` for `_ingest-cli-source` in this list for 
```
+--------------------+------------+---------+
|                  Sources                  |
+--------------------+------------+---------+
| ID                 | Type       | Enabled |
+--------------------+------------+---------+
| _ingest-api-source | ingest-api | true    |
+--------------------+------------+---------+
| _ingest-cli-source | ingest-cli | true    |
+--------------------+------------+---------+
```

guilload: Looks clean.

PR #2490, opened 2022-11-29
add docs about required system dependencies

### LINKED ISSUE #1775: Document that make and cmake are required to build
(due to jemalloc)

on ubuntu
`sudo apt install clang make cmake pkg-config libssl-dev`

trinity-1686a:
### Description

fix #1775
I added only packages I actually needed to run quickwit. I got no complain about make/cmake/pkg-config on neither platform so I assume they are probably no longer needed.

### How was this PR tested?

install the dependencies in `ubuntu:latest`, `amazonlinux:latest` and `archlinux:latest` and verify `cargo run --release` print quickwit help message

### DIFF
--- a/docs/get-started/installation.md
+++ b/docs/get-started/installation.md
@@ -64,6 +64,10 @@ pacman -S openssl
 Additionally it requires a few more dependencies to compile it. These dependencies are not required on production system:
 - `clang`: used to compile some dependencies.
 - `protobuf-compiler`: used to compile protobuf definitions.
+- `libssl-dev`: headers for libssl.
+- `pkg-config`: used to locate libssl.
+- `cmake`: used to build librdkafka, for kafka support.
+- `libsasl2-dev`: required for sasl support in kafka.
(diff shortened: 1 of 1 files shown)


docs/get-started/installation.md
```diff
@@ -61,6 +61,44 @@ pacman -S openssl
 
 </Tabs>
 
+Additionally it requires a few more dependencies to compile it. These dependencies are not required on production system:
```
ChillFish8, line 64:
```suggestion
Additionally, it requires a few more dependencies to compile it. These dependencies are not required on production systems running the produced binary:
```

ChillFish8:
Can this just be checked if these are all that is needed when compiling with the `release-feature-set` feature enabled? This enables OpenSSL and the various other tools like Postgres and AWS that are disabled by default.

trinity-1686a:
`pkg-config` (maybe more) is required with that feature. I'll update accordingly

ChillFish8 (approved):
LGTM

PR #2528, opened 2022-12-01
Fix build (UI)

evanxg852000:
Added ui build to mac and fix version env vars


fmassot (approved):
Just checked on my laptop that the UI is correctly built.

Issue #2531, opened 2022-12-01
Unknown build info in binariy


evanxg852000:
When the binary is built, the build info contains some unknown values.
- Tags should be be empty

```json
{
  "build_date": "2022-12-01T08:44:16Z",
  "build_profile": "release",
  "build_target": "x86_64-unknown-linux-gnu",
  "cargo_pkg_version": "0.4.0",
  "commit_date": "unknown",
  "commit_hash": "unknown",
  "commit_short_hash": "unknown",
  "commit_tags": [],
  "version": "0.4.0-nightly"
}
```

fmassot:
Fixed by #2544

PR #2532, opened 2022-12-01
Fixing build info in binaries

evanxg852000:
https://github.com/quickwit-oss/quickwit/issues/2531

### DIFF
--- a/.github/actions/build-binary/action.yml
+++ b/.github/actions/build-binary/action.yml
@@ -30,6 +30,9 @@ runs:
     - name: Install cross
       run: cargo install cross
       shell: bash
+    - name: Help cross find the git repo
+      run: cp -r .git/ ./quickwit/.git
+      shell: bash
     - name: Build Quickwit
       run: cross build --release --features release-feature-vendored-set --target ${{ inputs.target }}
       shell: bash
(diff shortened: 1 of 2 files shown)


.github/actions/build-binary/action.yml
```diff
@@ -30,6 +30,9 @@ runs:
     - name: Install cross
       run: cargo install cross
       shell: bash
+    - name: Help cross find the git repo
+      run: cp -r .git/ ./quickwit/.git
```
guilload, line 34: Use a symlink.

evanxg852000: Yes much better. I just wanted something radical given the time it takes to test it. Also, this ends up mounted in a docker container.

guilload: Can't we use passthrough env variables for `cross`?
https://github.com/cross-rs/cross#passing-environment-variables-into-the-build-environment

evanxg852000: Yes, it was used at some point

guilload: The build script is nop if git is not available in the env. `QuickwitBuildInfo` attempts to read the `QW_COMMIT_*` en vars directly. So as long as they are populated by the build script or whatever process, we're good.

PR #2535, opened 2022-12-01
Use gssapi-vendored features for rdkafka

trinity-1686a:


ChillFish8:
I think we can update the docs again https://quickwit.io/docs/main-branch/get-started/installation#note-on-external-dependencies around `sasl2` since this is vendored now.

quickwit/Cargo.toml
```diff
@@ -205,3 +206,6 @@ tantivy-query-grammar = { git = "https://github.com/quickwit-oss/tantivy/", rev
 # This is actually not used directly the goal is to fix the version
 # used by reqwest. 0.8.30 has an unclear license.
 encoding_rs = "=0.8.29"
+
+[patch.crates-io]
+sasl2-sys = { git = "https://github.com/infinyon/rust-sasl/", branch = "fix-cross-compiling" }
```
fulmicoton, line 211: @trinity-1686a we should never always use rev and never branch, espeically for third party repository.

a) it gives an opportunity for a thirdparty to inject any kind of code into quickwit.
b) it makes the build brittle.

We also need to clone such a repo to depend on it, to make sure this branch does not disappear tomorrow.

PR #2537, opened 2022-12-01
Remove 2 round trips in publish splits transaction

guilload:
### Description
- Remove 2 round trips in publish splits transaction. 
- Remove ability to re-publish a published split (but reapplying a non-empty checkpoint delta is not possible anyway)

### How was this PR tested?
- Updated test suite
- Ran `make test-all`

### DIFF
--- a/quickwit/quickwit-metastore/src/checkpoint.rs
+++ b/quickwit/quickwit-metastore/src/checkpoint.rs
@@ -555,28 +555,36 @@ mod tests {
     }
 
     #[test]
-    fn test_checkpoint_simple() -> anyhow::Result<()> {
+    fn test_checkpoint_simple() {
         let mut checkpoint = SourceCheckpoint::default();
         assert_eq!(format!("{:?}", checkpoint), "Ckpt()");
-        let delta1 = {
+
+        let delta = {
             let mut delta = SourceCheckpointDelta::from_partition_delta(
                 PartitionId::from("a"),
                 Position::from(123u64),
                 Position::from(128u64),
             );
-            delta.record_partition_delta(
-                PartitionId::from("b"),
-                Position::from(60002u64),
-                Position::from(60187u64),
-            )?;
+            delta
+                .record_partition_delta(
+                    PartitionId::from("b"),
+                    Position::from(60002u64),
+                    Position::from(60187u64),
+                )
+                .unwrap();
             delta
         };
-        assert!(checkpoint.try_apply_delta(delta1).is_ok());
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-metastore/src/checkpoint.rs
```diff
@@ -555,28 +555,35 @@ mod tests {
...
+                .unwrap();
             delta
         };
-        assert!(checkpoint.try_apply_delta(delta1).is_ok());
+        checkpoint.try_apply_delta(delta.clone()).unwrap();
+        assert_eq!(
+            format!("{:?}", checkpoint),
+            "Ckpt(a:00000000000000000128 b:00000000000000060187)"
+        );
+        checkpoint.try_apply_delta(delta).unwrap_err();
```
guilload, line 582: Just verifying that `try_apply_delta` is not idempotent.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -617,73 +455,137 @@ impl Metastore for PostgresqlMetastore {
     async fn publish_splits<'a>(
         &self,
         index_id: &str,
-        new_split_ids: &[&'a str],
+        staged_split_ids: &[&'a str],
         replaced_split_ids: &[&'a str],
         checkpoint_delta_opt: Option<IndexCheckpointDelta>,
     ) -> MetastoreResult<()> {
         run_with_tx!(self.connection_pool, tx, {
+            let mut index_metadata = index_metadata(tx, index_id).await?;
```
guilload, line 463: We can remove this transaction when we move index checkpoints to a dedicated table. This is already doable, but the JSON query is awful.

quickwit/quickwit-metastore/src/tests.rs
```diff
@@ -2261,26 +2260,6 @@ pub mod test_suite {
             split_meta.publish_timestamp,
             Some(split_meta.update_timestamp)
         );
-
-        // wait a sec & re-publish and check publish_timestamp has not changed
```
guilload, line 2265: This is no longer possible. Note that this test was utterly artificial because the second publish op used a different delta.

quickwit/quickwit-metastore/src/checkpoint.rs
fulmicoton, line 582: It is a good comment, That should be in the code.

```suggestion
        // Just verifying that try_apply_delta is not idempotent.
        checkpoint.try_apply_delta(delta).unwrap_err();
```

PR #2551, opened 2022-12-05
QW-2128: Remove `IndexingDirectory` and just use `ScratchDirectory`.

### LINKED ISSUE #2128: Remove IndexingDirectory.
(as spotted by @guilload)
We have an `IndexingDirectory` that was introduced because indexing directories where hosting both scratch directories and split caches.
The split cache has been removed in https://github.com/quickwit-oss/quickwit/pull/2088 so the extra indirection is not necessary anymore.

See if it makes sense and simplifies the code to remove the entire `IndexingDirectory` thingy.

ChillFish8:
Closes #2128 

### Description
Does what is said on the tin, remove the extra layer of indirection which is now unnecessary?

### How was this PR tested?
Tests have been adjusted.


quickwit/quickwit-indexing/src/models/indexing_directory.rs
```diff
@@ -1,179 +0,0 @@
...
-
-    #[tokio::test]
-    async fn test_indexing_directory() -> anyhow::Result<()> {
-        let tempdir = tempfile::tempdir()?;
-        let indexing_directory = IndexingDirectory::create_in_dir(tempdir.path()).await?;
-        let indexing_directory_path = indexing_directory.path().to_path_buf();
-        assert_eq!(indexing_directory_path, tempdir.path());
-
-        let scratch_directory_path = indexing_directory_path.join("scratch");
-        assert!(scratch_directory_path.try_exists()?);
```
fulmicoton, line 158: We lost this unit test. Can you port it?

fulmicoton (changes requested):
The change is ok but you lost unit tests in the process.

quickwit/quickwit-indexing/src/models/indexing_directory.rs
ChillFish8: Done!

fulmicoton (approved):

PR #2553, opened 2022-12-06
Modulo partition key

### LINKED ISSUE #2473: Add support for modulo in the partition key DSL

trinity-1686a:
### Description

fix #2473 

### How was this PR tested?
I ingested a few hundred documents with a `partition_key` %3 and verified 3 split are generated. I deleted a split and made sure document deleted had all the same number field mod 3.


quickwit/quickwit-doc-mapper/src/routing_expression/mod.rs
```diff
@@ -125,22 +141,37 @@ impl Display for RoutingExpr {
...
                 ExprType::Composite.hash(hasher);
                 for child in children {
                     child.eval_hash(ctx, hasher);
                 }
             }
+            InnerRoutingExpr::Modulo(field_name, modulo) => {
+                ExprType::Modulo.hash(hasher);
+                ctx.hash_attribute(field_name, hasher, &|value| {
+                    if let Some(unsigned) = value.as_u64() {
+                        unsigned.rem_euclid(*modulo).into()
```
fulmicoton, line 165: I'd apply the modulo after the hash.

quickwit/quickwit-doc-mapper/src/routing_expression/mod.rs
```diff
@@ -26,7 +26,8 @@ use serde_json::Value as JsonValue;
 use siphasher::sip::SipHasher;
 
 pub trait RoutingExprContext {
-    fn hash_attribute<H: Hasher>(&self, attr_name: &str, hasher: &mut H);
+    fn hash_attribute<H: Hasher, F>(&self, attr_name: &str, hasher: &mut H, tranform: &F)
```
fulmicoton, line 29:
```suggestion
    fn hash_attribute<H: Hasher, F>(&self, attr_name: &str, hasher: &mut H)
```

the transform stuff is not needed if you apply the modulo after the hash.

fulmicoton (changes requested):
The point of this functionality is to let people use partitioning on a field that as cardinality of a few thousands of elements.
(if we tweak the logic of "tag", we could even push that to any cardinality)

For such a cardinality, the current "one partition per field value" would create too many splits.

Instead, people would be able to partition by "hash(field) % 100" for instance and get 100 partitions.

I am not sure what the best dsl should be.

`(field%10)` is indeed a bit misleading, especially when field is an integer...
but after all this is a DSL, so maybe this is ok.

Other possibilities could be
- `(hash_modulo(field, 10)`
- `bucket(field, 10)` 
- ... whatever you can think of?

quickwit/quickwit-doc-mapper/src/routing_expression/mod.rs
```diff
@@ -125,11 +139,19 @@ impl Display for RoutingExpr {
...
     Composite(Vec<InnerRoutingExpr>),
-    // TODO Enrich me! Map / Modulo
+    Modulo(Box<InnerRoutingExpr>, u64),
+    // TODO Enrich me! Map / ...
 }
 
 impl InnerRoutingExpr {
-    fn eval_hash<Ctx: RoutingExprContext, H: Hasher>(&self, ctx: &Ctx, hasher: &mut H) {
+    // our hasher is Copy, but restrict ourself to Clone so it's easier to see
+    // when it get duplicated.
```
fulmicoton, line 148: I agree :)

quickwit/quickwit-doc-mapper/src/routing_expression/mod.rs
```diff
@@ -138,9 +160,16 @@ impl InnerRoutingExpr {
...
                 ExprType::Composite.hash(hasher);
                 for child in children {
-                    child.eval_hash(ctx, hasher);
+                    child.eval_hash(ctx, hasher, salted_empty_hasher.clone());
                 }
             }
+            InnerRoutingExpr::Modulo(inner_expr, modulo) => {
+                ExprType::Modulo.hash(hasher);
+
+                let mut sub_hasher = salted_empty_hasher.clone();
```
fulmicoton, line 169: I don't think we care about the salt here.

fulmicoton, line 169:
```suggestion
                let mut sub_hasher = H::default();
```

fulmicoton: Maybe we could get rid of the salt for the sub_hasher, just work off a `H::default()` and call it a day.

quickwit/quickwit-doc-mapper/src/routing_expression/mod.rs
```diff
@@ -209,26 +284,110 @@ impl Display for InnerRoutingExpr {
...
+        Ok(res)
+    }
+
+    // DSL:
+    //
+    // RoutingExpr := RoutingSubExpr [ , RoutingExpr ]
+    // RougingSubExpr := Identifier [ \( Arguments \) ]
+    // Identifier := FieldChar [ Identifier ]
+    // FieldChar := { a..z | A..Z | 0..9 | _ }
+    // Arguments := Argument [ ; Arguments ]
```
fulmicoton, line 327:
```suggestion
    // Arguments := Argument [ ; Arguments ]
```
`;` is a weird separator for arguments. Can we make it ","?

trinity-1686a: `,` would be ambiguous with an argument being a list of fields:
`hash_mod(tennant_id,app_id, 5)`

We could either: 
- keep the `;` as arg delimiter
- use an other delimiter than `,` for list of fields
- don't allow functions to take a list of fields, but be variadic (the example would get parsed to 3 args, annd the function would know the last is supposed to be a number, and fuse all others in a Composite)

fulmicoton: Parenthesis?

`hash_mod((tenant_id,app_id), 5)`

quickwit/quickwit-doc-mapper/src/routing_expression/mod.rs
```diff
@@ -209,26 +277,113 @@ impl Display for InnerRoutingExpr {
...
+    pub(crate) enum Argument {
+        Expression(Vec<ExpressionAst>),
+        Number(u64),
+    }
+
+    pub(crate) fn parse_expression(expr_dsl_str: &str) -> anyhow::Result<Vec<ExpressionAst>> {
+        let (i, res) = routing_expr(expr_dsl_str)
+            .finish()
+            .map_err(|e| anyhow::anyhow!("error parsing routing expression: {e}"))?;
+        t(eof(i))?;
```
fulmicoton, line 309: I did not know about this pattern.
Did it help the compiler?

Can you either:
- remove the pattern to something less ninja. I suspect something like this compiles.
- rename `t` to something like `type_hinting` or somehting like that?

fulmicoton (approved):

quickwit/quickwit-doc-mapper/src/routing_expression/mod.rs
trinity-1686a: The compiler can't find the error type alone. I'm used to having the name very short because it generally appears much more often (dozens of times at least). It's only there once so I can use a turbofish instead

Issue #2495, opened 2022-11-29
Return a 404 in the error code


fulmicoton:
Delete a source or an index that does not exist should return a 404.
Right now it is a 500.

Listing splits should return a 404 if the index does not exists.

PR #2561, opened 2022-12-07
Fixed REST status code

evanxg852000:
Converted error to correct status code

Close https://github.com/quickwit-oss/quickwit/issues/2495


quickwit/quickwit-metastore/src/error.rs
```diff
@@ -118,8 +118,8 @@ impl ServiceError for MetastoreError {
             Self::InvalidManifest { .. } => ServiceErrorCode::Internal,
             Self::Io { .. } => ServiceErrorCode::Internal,
             Self::SourceAlreadyExists { .. } => ServiceErrorCode::BadRequest,
-            Self::SourceDoesNotExist { .. } => ServiceErrorCode::BadRequest,
-            Self::SplitsDoNotExist { .. } => ServiceErrorCode::BadRequest,
+            Self::SourceDoesNotExist { .. } => ServiceErrorCode::NotFound,
+            Self::SplitsDoNotExist { .. } => ServiceErrorCode::NotFound,
```
guilload, line 122: I find that `BadRequest` is better for `SplitsDoNotExist` because it happens when attempting to publish or mark a split for deletion, which is more of an illegal operation than a simple not found error on a CRUD call. This is really the same class of error as `SplitsNotStaged` or `SplitsNotDeletable`.

guilload (approved):

PR #2574, opened 2022-12-14
Updating to tokio 1.23 and updating unit tests.

### LINKED ISSUE #2571: Tokio version 1.23 breaks the tests
Upgrading the Tokio version 1.23 breaks the tests: [CI Run](https://github.com/quickwit-oss/quickwit/actions/runs/3674921766/jobs/6233840586)

Tests pass without a problem on macOS and some Linux distros but fail on other instances. 

My first instinct was this PR being responsible: https://github.com/tokio-rs/tokio/pull/5223 since on the test: 
`test_mailbox_send_with_backpressure_counter_no_backpressure_cleansheet` we measure some backpressure on the zero capacity channels where we expect none:
```bash
        FAIL [   0.002s]                 quickwit-actors mailbox::tests::test_mailbox_send_with_backpressure_counter_no_backpressure_cleansheet

--- STDOUT:                              quickwit-actors mailbox::tests::test_mailbox_send_with_backpressure_counter_no_backpressure_cleansheet ---

running 1 test
test mailbox::tests::test_mailbox_send_with_backpressure_counter_no_backpressure_cleansheet ... FAILED

failures:

failures:
    mailbox::tests::test_mailbox_send_with_backpressure_counter_no_backpressure_cleansheet

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 48 filtered out; finished in 0.00s


--- STDERR:                              quickwit-actors mailbox::tests::test_mailbox_send_with_backpressure_counter_no_backpressure_cleansheet ---
thread 'mailbox::tests::test_mailbox_send_with_backpressure_counter_no_backpressure_cleansheet' panicked at 'assertion failed: `(left == right)`
  left: `15`,
 right: `0`', quickwit-actors/src/mailbox.rs:466:9
```

However, inserting `tokio::time::sleep(Duration::ZERO).await;` before sending the second message fixes the test. I am not sure what causes this, but my guess is context switch causes the task to resume after some time; in the meantime, the `BackPressureActor` processes the message, and the channel becomes empty again.

fulmicoton:
Closes #2571

### DIFF
--- a/quickwit/quickwit-janitor/src/actors/delete_task_planner.rs
+++ b/quickwit/quickwit-janitor/src/actors/delete_task_planner.rs
@@ -423,11 +423,12 @@ mod tests {
                 fast: true
         "#;
         let test_sandbox = TestSandbox::create(index_id, doc_mapping_yaml, "{}", &["body"]).await?;
-        let docs = vec![
+        let docs = [
             serde_json::json!({"body": "info", "ts": 0 }),
             serde_json::json!({"body": "info", "ts": 0 }),
             serde_json::json!({"body": "delete", "ts": 0 }),
         ];
+        // Creates 3 splits
         for doc in docs {
             test_sandbox.add_documents(vec![doc]).await?;
         }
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-janitor/src/actors/delete_task_planner.rs
```diff
@@ -522,7 +523,7 @@ mod tests {
         );
         // Trigger new plan evaluation and check that we don't have new merge operation.
         delete_planner_mailbox
-            .send_message(PlanDeleteOperations)
+            .ask(PlanDeleteOperations)
```
fulmicoton, line 526: send message just queues a message and returns. It does not wait for the message to be processed.

boraarslan (approved):
Great!

PR #2609, opened 2022-12-29
Modify `delete_source` to return empty body

### LINKED ISSUE #2493: Delete source should return an empty object.
(right now it returns `{}`)

boraarslan:
Closes #2493

### DIFF
--- a/quickwit/quickwit-serve/src/index_api/rest_handler.rs
+++ b/quickwit/quickwit-serve/src/index_api/rest_handler.rs
@@ -20,6 +20,7 @@
 use std::sync::Arc;
 
 use bytes::Bytes;
+use hyper::StatusCode;
 use quickwit_config::{ConfigFormat, QuickwitConfig, SourceConfig};
 use quickwit_core::{IndexService, IndexServiceError};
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -183,16 +185,19 @@ fn delete_source_handler(
...
 
 async fn delete_source(
     index_id: String,
     source_id: String,
     index_service: Arc<IndexService>,
-) -> Result<(), IndexServiceError> {
+) -> impl warp::Reply {
     info!(index_id = %index_id, source_id = %source_id, "delete-source");
-    index_service.delete_source(&index_id, &source_id).await
+    let res = index_service.delete_source(&index_id, &source_id).await;
```
fmassot, line 196: It works... but it's really not a good solution, unfortunately. We have other DELETE endpoints that may return just an empty body. And an empty body is not a valid JSON.

Having something like `Format::make_rest_reply_non_serializable_error` specific to DELETE requests.

Issue #2201, opened 2022-10-30
Add confirm promp for CLI destructive commands


fmassot:
It would be nice to have that or some users will lose their indexes. I saw that `quickwit index delete` does not have a prompt.

fatelei:
assign to me

fulmicoton:
Thanks for chiming in @fatelei. It's all yours!

fulmicoton:
Assigned to druznek aka Maurizio.

ph:
I think this issue can be closed, following the work in #2618 and #3147.

I did a small audit of the `quickwit-cli/src` source code and identified the following destructive sub-commands and they have a user confirmation.

- index clear
- index delete
- split mark-for-deletion 
- source delete

guilload:
Closing then. Thanks @ph !

PR #2618, opened 2023-01-04
feature: added safe mode to delete index

druzn3k:
### Description

The CLI interface of quickwit has a subcommand for deleting an index, given it's ID. The subcommand, except for dry-run, has no safety built-in. I thought about putting a `--yes` flag, like other subcommands in the CLI, but I feared that this may be disrupting for existing client's pipelines. This seems a more sensible choice because it doesn't interfere with existing processes in place and it can be used only where is necessary. It solves issue #2201. 

### How was this PR tested?

Added the relevant flags in the existing tests.

### DIFF
--- a/quickwit/quickwit-cli/tests/cli.rs
+++ b/quickwit/quickwit-cli/tests/cli.rs
@@ -43,6 +43,7 @@ use quickwit_common::uri::Uri;
 use quickwit_common::ChecklistError;
 use quickwit_config::service::QuickwitService;
 use quickwit_config::CLI_INGEST_SOURCE_ID;
+use quickwit_indexing::actors::INDEXING_DIR_NAME;
 use quickwit_metastore::{quickwit_metastore_uri_resolver, Metastore, MetastoreError, SplitState};
 use serde_json::{json, Number, Value};
 use tokio::time::{sleep, Duration};
@@ -501,6 +502,43 @@ async fn test_delete_index_cli_dry_run() {
     delete_index_cli(args).await.unwrap();
     let metastore = refresh_metastore(metastore).await.unwrap();
     assert!(!metastore.index_exists(&index_id).await.unwrap());
+}
+
+#[tokio::test]
+async fn test_delete_index_cli() {
+    let index_id = append_random_suffix("test-delete-cmd");
+    let test_env = create_test_env(index_id.clone(), TestStorageType::LocalFileSystem).unwrap();
+    test_env.start_server().await.unwrap();
+    create_logs_index(&test_env).await.unwrap();
+
+    local_ingest_docs(test_env.resource_files["logs"].as_path(), &test_env)
+        .await
+        .unwrap();
+
+    let args = DeleteIndexArgs {
+        cluster_endpoint: test_env.cluster_endpoint.clone(),
+        index_id: index_id.clone(),
+        assume_yes: true,
+        dry_run: false,
+    };
+
+    delete_index_cli(args).await.unwrap();
+
+    assert!(test_env
+        .metastore()
+        .await
+        .unwrap()
+        .index_metadata(&test_env.index_id)
+        .await
+        .is_err());
+
+    assert!(!test_env
+        .data_dir_path
+        .join(INDEXING_DIR_NAME)
+        .join(test_env.index_id)
+        .as_path()
+        .try_exists()
+        .unwrap());
 }
 
 #[tokio::test]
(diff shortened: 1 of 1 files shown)


druzn3k:
I tried running test on my machine following the `CONTTRIBUTING.md` files. It brutally fails at the moment (I think that the issue is related to rust-toolchain but I'm not 100% sure. I will do some more dissection and I will eventually ask for help if necessary.

fulmicoton:
@ChillFish8 can you review?

quickwit/quickwit-cli/src/index.rs
```diff
@@ -192,6 +192,8 @@ pub fn build_index_command<'a>() -> Command<'a> {
                         .display_order(1),
                     arg!(--"dry-run" "Executes the command in dry run mode and only displays the list of splits candidates for deletion.")
                         .required(false),
+                    arg!(--"safe-mode" "Prompt for confirmation before deleting an index.")
```
ChillFish8, line 195: I think `--safe` might be a better name for this flag, IMO it gets the point across more directly.

quickwit/quickwit-cli/src/index.rs
```diff
@@ -1098,6 +1103,16 @@ pub async fn delete_index_cli(args: DeleteIndexArgs) -> anyhow::Result<()> {
...
+    if args.safe_mode {
+        let prompt = format!(
+            "This operation will delete the index `{}`. Do you want to proceed?",
+            args.index_id
+        );
+        if !prompt_confirmation(&prompt, false) {
+            return Ok(());
+        }
+    }
+
```
ChillFish8, line 1115: Could we add a unit test for this? I think you can probably copy what we have done with `--dry-run`

druzn3k: I tryed, but I clashed with [prompt_confirmation](https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-cli/src/lib.rs#L163), because I have no idea how to "bypass" it. I though about putting something that conditionally *NOP*ped the function `prompt_confirmation`, but I didn't know how to do that. Maybe I missed the point where dry-run was tested, that's the most probable explanation.

druzn3k: could you point me to the right code to replicate @ChillFish8, please?

quickwit/quickwit-cli/src/index.rs
druzn3k: Not a problem, I could change the name no problem. Aaaaaaaaand... done!

quickwit/quickwit-cli/src/index.rs
ChillFish8: Ah yes, that is awkward, in that case we may have to skip the unit test although luckily it's simple logic. Thoughts @fmassot ?

fmassot: I think we should remain consistent with other prompt confirmations. This would means that by default, we have a prompt, and we can use `--yes` args to make the tests pass.

druzn3k: Not a problem for me, but wouldn't this change existing behavior? If there are automated pipelines, with this update that seems minor, they would break. Again, not a problem for me, just want to make sure. :) You for sure know your user base better than me!

fmassot: I'm more in favor of consistency on all commands. just pinging @guilload to double check.

guilload: I agree with you, @fmassot.

fmassot: @druzn3k can you update the PR to use a `--yes` flag?

druzn3k: Yes, tomorrow I will 👍🏻

druzn3k: Sorry for the latency, life got in the way (positively, nothing bad luckily) and still I didn't have time to apply the required changes. I promise that I will do it as soon as I can!

quickwit/quickwit-cli/src/index.rs
```diff
@@ -1124,6 +1129,16 @@ pub async fn delete_index_cli(args: DeleteIndexArgs) -> anyhow::Result<()> {
     debug!(args=?args, "delete-index");
     quickwit_telemetry::send_telemetry_event(TelemetryEvent::Delete).await;
 
+    if !args.yes{
```
ChillFish8, line 1132: Small nitpick -> formatting

ChillFish8 (approved):
Looks good now just one really small nitpick which I think is fixed by running cargo fmt

fulmicoton:
@druzn3k You can correctly format the code by running `make fmt` at the root of the repo.

It requires nightly.

fmassot:
@druzn3k this PR will clash with a big refactoring of the CLI. Let me handle the merge.

druzn3k:
Ok, thank you @fmassot

fmassot (approved):

PR #2625, opened 2023-01-06
Large actor scheduler refactoring

fulmicoton:
The scheduler is not a actor in the sense
of the actor framework anymore.

It removes the necessity to create some fake
scheduler mailbox to spawn the scheduler
itself.

The scheduler also now has an improved logic to
simulate time shift.

It only jumps forward when no actors has any work to do. Provided all of the processing is done in actor, the results should be rigorously the same as if someone used time::sleep... Only faster.


quickwit/quickwit-actors/src/actor_context.rs
```diff
@@ -340,16 +331,13 @@ impl<A: Actor> ActorContext<A> {
         M: Sync + Send + std::fmt::Debug + 'static,
     {
         let self_mailbox = self.inner.self_mailbox.clone();
-        let callback = Callback(Box::pin(async move {
+        let callback = async move {
```
fmassot, line 334: Can you add the following docs for this function? This is not obvious that a scheduled message to self is sent into the high-priority queue.  

```
    /// Schedules a message that will be sent to the high-priority
    /// queue of the actor Mailbox.
    /// This function is especially useful for an actor to monitor/observe himself. Using the high-priority queue ensures that the message will be processed after the processing of all queued messages, which will defeat the purpose of monitoring.
```

quickwit/quickwit-actors/src/scheduler.rs
```diff
@@ -58,139 +52,196 @@ impl PartialOrd for TimeoutEvent {
...
+    tx: flume::Sender<SchedulerMessage>,
 }
 
-impl fmt::Debug for ScheduleEvent {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.debug_struct("ScheduleEvent")
-            .field("timeout", &self.timeout)
-            .finish()
+impl SchedulerClient {
+    pub fn disconnected() -> Self {
```
fmassot, line 86:
```
Returns a `SchedulerClient` with no `Scheduler` behind, thus disconnected.
```

quickwit/quickwit-actors/src/scheduler.rs
```diff
@@ -58,139 +52,196 @@ impl PartialOrd for TimeoutEvent {
...
-            num_pending_events: self.future_events.len(),
-            total_num_events: self.event_id_generator,
-        }
+    pub(crate) fn timeout(&self) {
+        let _ = self.inner.tx.send(SchedulerMessage::Timeout);
     }
 
-    fn name(&self) -> String {
-        "Scheduler".to_string()
+    pub fn no_advance_time_guard(&self) -> NoAdvanceTimeGuard {
```
fmassot, line 144:
```rust
    /// Returns a `NoAdvanceTimeGuard` which calls `inc_no_advance_time`
    /// on `NoAdvanceTimeGuard::new` and `dec_no_advance_time` when dropped.
```

quickwit/quickwit-actors/src/spawn_builder.rs
```diff
@@ -306,6 +350,8 @@ async fn actor_loop<A: Actor>(actor: A, inbox: Inbox<A>, ctx: ActorContext<A>) -
         actor_env.process_messages().await
     };
 
+    // TODO the no advance time guard for finalize has a race condition. Ideally we would
```
fmassot, line 353: Good point indeed.
We could have a guard in the returned error by `process_messages`.
One note here, I would have probably kept a guard in `process_all_available_messages` and avoid keeping a guard in each envelope. I may have missed subtleties here though.

quickwit/quickwit-actors/src/actor_context.rs
```diff
@@ -96,6 +89,17 @@ impl<A: Actor> ActorContext<A> {
         }
     }
 
+    pub fn spawn_ctx(&self) -> &SpawnContext {
+        &self.spawn_ctx
+    }
+
+    pub async fn sleep(&self, duration: Duration) {
```
fmassot, line 96: nitpick: the function name `sleep` can be misleading, I would stick to `simulate_sleep`.

quickwit/quickwit-actors/src/scheduler.rs
```diff
@@ -58,139 +52,196 @@ impl PartialOrd for TimeoutEvent {
...
+        }
+    });
+    scheduler_client
+}
+
+struct Scheduler {
+    // We attribute an event_id to all event just to break ties
+    // if two events are scheduled on the same time.
+    event_id_generator: u64,
+    // simulated_time = Instant::now() + simulated_time_shift.
```
fmassot, line 217: nitpick: docs

```rust
    // Simulated time shift which defines the scheduler time reference as `simulated_time = Instant::now() + simulated_time_shift`.
    // By default `simulated_time_shift` is set to 0 but it can be shifted when the scheduler has to process a simulate sleep event`.
```

quickwit/quickwit-actors/src/scheduler.rs
```diff
@@ -199,62 +250,103 @@ impl Scheduler {
...
-        tx: Sender<()>,
-        ctx: &ActorContext<Self>,
-    ) {
-        let now = self.simulated_now();
-        let deadline = match time_shift {
-            TimeShift::ToInstant(instant) => instant,
-            TimeShift::ByDuration(duration) => now + duration,
+    async fn process_simulate_sleep(&mut self, oneshot_tx: oneshot::Sender<()>, timeout: Duration) {
+        let Some(scheduler_client) = self.scheduler_client() else { return; };
+        let accelerate_time_count_guard = AccelerateTimeCountGuard::new(scheduler_client);
```
fmassot, line 259: Note: I found a bit quirky the increment/decrement dance with new/drop with the scheduler client. I might have tried something else: put the `accelerate_time_count_guard` inside the scheduler itself. The downside is that we now have a counter on the scheduler client and one on the scheduler... 
I added a couple of docs to help the reader.

quickwit/quickwit-actors/src/scheduler.rs
```diff
@@ -199,62 +250,103 @@ impl Scheduler {
...
+        Scheduler {
+            event_id_generator: 0u64,
+            simulated_time_shift: Duration::default(),
+            future_events: Default::default(),
+            next_timeout: None,
+            weak_scheduler_client: Arc::downgrade(&scheduler_client.inner),
+        }
+    }
+
+    /// Return the duration we should advance time to.
```
fmassot, line 318: outdated docs.

```rust
/// Adds the necessary `time_shitf` to `self.simulated_time_shift` to place the scheduler a the next event deadline when possible.
/// It does nothing if no one is waiting for Universe::simulate_time_shift or
/// if some message handler, actor initialize, or finalize have not been processed yet.
```

fmassot (approved):
Awesome, I added a couple of docs to help the reader as the role of the tandem (scheduler, scheduler client) is not obvious at first sight.

quickwit/quickwit-actors/src/actor_context.rs
fulmicoton: I prefer to stick to sleep here... but comment.

quickwit/quickwit-actors/src/scheduler.rs
fulmicoton: In the scheduler? Do you mean in the context maybe?

quickwit/quickwit-actors/src/spawn_builder.rs
fulmicoton: That would have been incorrect... We want to avoid jump in time as soon as there are messages scheduled for processing.

fulmicoton:
I integrated your docs/ comments, and removed the SchedulerClient::disconnected (which was leaking I think)

PR #2631, opened 2023-01-08
Build  macos binaries 

### LINKED ISSUE #1928: Support for ARM architecture (Mac M1)
I see that there is no binary available for ARM / Mac M1. If nobody is on this project, I can participate on my time (because I need it, the indexing time is way too high on M1 as I have to mimic the x64 architecture). 

Can I build it and fix any bugs if needed? 
And also try to find solutions to build an ARM compatible Docker.

fmassot:
Fix #1928.

- to bypass the cross-compilation issue on macOS, I added the feature `release-macos-feature-vendored-set` for macOS builds. It deactivates the libsasl support, which is fine for macOS binaries.
- ~~I added an arm64 build for docker images.~~ => too slow

Nighly builds success: https://github.com/quickwit-oss/quickwit/actions/runs/3866977651/jobs/6591462839
Docker images (failing due to network issue but should work): https://github.com/quickwit-oss/quickwit/actions/runs/3867131208/jobs/6591714909


quickwit/quickwit-ui/Makefile
```diff
@@ -4,7 +4,7 @@ build:
 	yarn build
 
 install:
-	yarn install
+	yarn install --network-timeout 1000000
```
fmassot, line 7: I had network issues during `yarn install` on the arm64 docker build. I will remove this before merging.

.github/actions/cargo-build-macos-binary/action.yml
```diff
@@ -0,0 +1,70 @@
...
+      with:
+        node-version: 16
+        cache: "yarn"
+        cache-dependency-path: quickwit/quickwit-ui/yarn.lock
+    - run: make build-ui
+      shell: bash
+    - name: Install protoc
+      run: brew install protobuf
+      shell: bash
+    - uses: dtolnay/rust-toolchain@v1
```
evanxg852000, line 28: Wondering why you switched to this action for rust toolchain instead of the usual `actions-rs/toolchain@v1` .
Also, from the readme we need `dtolnay/rust-toolchain@master` since we are specifying the toolchain explicitly.

quickwit/quickwit-indexing/Cargo.toml
```diff
@@ -55,7 +52,8 @@ quickwit-storage = { workspace = true }
 [features]
 kafka = ["rdkafka", "backoff"]
 kafka-broker-tests = []
-vendored-kafka = ["kafka", "libz-sys/static", "openssl/vendored"]
+vendored-kafka = ["kafka", "libz-sys/static", "openssl/vendored", "rdkafka/gssapi-vendored"]
+vendored-kafka-macos = ["kafka", "libz-sys/static", "openssl/vendored"]
```
evanxg852000, line 56: Nice figuring this out :+1:

evanxg852000 (commented):
LGTM, please see my comment.

.github/actions/cargo-build-macos-binary/action.yml
fmassot: > we need dtolnay/rust-toolchain@master since we are specifying the toolchain explicitly.

What do you mean? Is `dtolnay/rust-toolchain@v1` should not work?

fmassot: Using `dtolnay/rust-toolchain` seems like a good shot if you look at the repository. One is maintained, not the other. 
(example of issues with the unmaintained repos: https://github.com/actions-rs/toolchain/issues/219)

evanxg852000: dtolnay/rust-toolchain@v1 works but from the docs since we are using toolchain parameter, they recommend specifying `@master`. r
Regarding the fact this action is maintained, then we need to update all other places we use action-rs (in a separate PR)

evanxg852000 (approved):

.github/actions/cargo-build-macos-binary/action.yml
fmassot: yes, I will keep using the action-rs/toolchain for now and will open an issue on this to migrate all workflows on this. This is not a priority :)

PR #2633, opened 2023-01-08
Add control plane scheduler.

fmassot:
This PR introduces the control plane scheduler. It brings:

- A scheduler actor that builds the indexing plan and schedules it. (cf [mermaid graph](https://github.com/quickwit-oss/quickwit/tree/main/quickwit/quickwit-control-plane). It sends the plan to each indexer.
- A metastore that wraps a concrete metastore and notifies the control plane of index changes.
- An indexing service, which receives an indexing plan to apply from the control plane (gRPC), and update its chitchat state with the running indexing pipelines.

Last things to check:

- [x] Add a TODO on the indexer pool as currently there is a leak as we don't remove the dead clients.
- [x] Add a test to check that indexing tasks order has no effect.
- [x] Show diff of plans in the control loop. 
- [x] check that an ingest API source will generate one pipeline per indexer.
- [x] tested locally
- [x] tested it on our EKS cluster
- [x] Add a test on an invalid source config
- [x] Trigger scheduling on new indexers
- [ ] why are there those warnings `#[warn(opaque_hidden_inferred_bound)]`???

TODO in the next PRs:
- ~~Decouple the two tasks computing a new plan if necessary and making sure the plan is applied.~~
- ~~Investigate what should be done with file sources.~~
- ~~Make the physical indexing plan stable when sources are added/deleted and when the control plane restart. Currently, adding an index will totally move tasks to different nodes.~~
- ~~Use a metastore version number to detect if there are some index/source change and avoid triggering tasks scheduling.~~
- ~~CLI metastore should send event to the control plane. This can be done if the CLI uses a gRPC call to the metastore instead of directly running locally the index/source creation/deletion.~~
- ~~Use a client pool for indexers clients.~~
- ~~Enhance the actor framework to allow actors to communicate via gRPC.~~
- ~~Forbid a `MetastoreWithControlPlaneTriggers` that wraps a gRPC client metastore.~~
- rename `quickwit-grpc-clients` into `quickwit-grpc-utils`.


quickwit/quickwit-serve/src/cluster_api/rest_handler.rs
```diff
@@ -29,7 +29,7 @@ use crate::Format;
 /// Cluster handler.
 pub fn cluster_handler(
     cluster: Arc<Cluster>,
-) -> impl Filter<Extract = impl warp::Reply, Error = Rejection> + Clone {
+) -> impl Filter<Extract = (impl warp::Reply,), Error = Rejection> + Clone {
```
fmassot, line 32: I had some weird warnings, I need to figure it out. You can ignore all the changes of `Filter`. Sorry for the noise

```
warning: opaque type `impl warp::Filter + std::clone::Clone + warp::filter::FilterBase<Extract = impl warp::Reply, Error = warp::Rejection>` does not satisfy its associated type bounds
  --> quickwit-serve/src/cluster_api/rest_handler.rs:32:18
   |
32 | ) -> impl Filter<Extract = impl warp::Reply, Error = Rejection> + Clone {
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
  ::: /Users/fmassot/.cargo/registry/src/github.com-1ecc6299db9ec823/warp-0.3.3/src/filter/mod.rs:40:19
   |
40 |     type Extract: Tuple; // + Send;
   |                   ----- this associated type bound is unsatisfied for `impl warp::Reply`
   |
   = note: `#[warn(opaque_hidden_inferred_bound)]` on by default
help: add this bound
   |
32 | ) -> impl Filter<Extract = impl warp::Reply + warp::generic::Tuple, Error = Rejection> + Clone {
```

quickwit/quickwit-cli/src/index.rs
```diff
@@ -904,6 +908,11 @@ pub async fn ingest_docs_cli(args: IngestDocsArgs) -> anyhow::Result<()> {
         let index_service = IndexService::from_config(config.clone()).await?;
         index_service.clear_index(&args.index_id).await?;
     }
+    // The indexing service needs to update its cluster chitchat state so that the control plane is
+    // aware of the running tasks. We thus create a fake cluster to instantiate the indexing service
+    // and avoid impacting potential control plane running on the cluster.
+    let transport = ChannelTransport::default();
+    let fake_cluster = create_cluster_for_test(Vec::new(), &["indexer"], &transport, false).await?;
```
fulmicoton, line 915: Is this the only reason for chitchat dependency? 
Could we either re-export the transport symbol in quickwit-cluster or add a method to create a test cluster using a `channel::Transport::default()`

quickwit/quickwit-control-plane/src/grpc_adapter.rs
```diff
@@ -0,0 +1,125 @@
...
+use quickwit_actors::Mailbox;
+use quickwit_proto::control_plane_api::control_plane_service_server::{self as grpc};
+use quickwit_proto::control_plane_api::{NotifyIndexChangeRequest, NotifyIndexChangeResponse};
+use quickwit_proto::tonic;
+
+use crate::scheduler::IndexingScheduler;
+
+#[allow(missing_docs)]
+#[derive(Clone)]
+pub struct GrpcControlPlaneAdapter(Mailbox<IndexingScheduler>);
```
fulmicoton, line 30: It is starting to be a bit lame that our actor framework is not distributed.

quickwit/quickwit-control-plane/src/grpc_adapter.rs
```diff
@@ -0,0 +1,125 @@
...
+#[async_trait]
+impl grpc::ControlPlaneService for GrpcControlPlaneAdapter {
+    async fn notify_index_change(
+        &self,
+        request: tonic::Request<NotifyIndexChangeRequest>,
+    ) -> Result<tonic::Response<NotifyIndexChangeResponse>, tonic::Status> {
+        let index_event_request = request.into_inner();
+        let create_index_reply = self
+            .0
+            .send_message(index_event_request)
```
fulmicoton, line 47: Prefer using .ask(...)

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -0,0 +1,657 @@
...
+/// to identify if the plan is up to date with the metastore.
+#[derive(Debug, PartialEq, Clone, Serialize)]
+pub struct PhysicalIndexingPlan {
+    indexing_tasks_per_node_id: HashMap<String, Vec<IndexingTask>>,
+}
+
+impl PhysicalIndexingPlan {
+    /// Creates a [`PhysicalIndexingPlan`] from a non empty list of node IDs.
+    pub fn new(node_ids: Vec<String>) -> Self {
+        assert!(!node_ids.is_empty());
```
fulmicoton, line 41: Is the assert really safe?

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -0,0 +1,657 @@
...
+            .into_iter()
+            .map(|node_id| (node_id, Vec::new()))
+            .collect();
+        Self {
+            indexing_tasks_per_node_id: indexing_tasks_per_indexer,
+        }
+    }
+
+    /// Adds the indexing task to the list of tasks assigned to the given node ID.
+    pub fn assign_indexing_task(&mut self, node_id: &String, indexing_task: IndexingTask) {
```
fulmicoton, line 52:
```suggestion
    pub fn assign_indexing_task(&mut self, node_id: &str, indexing_task: IndexingTask) {
```

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -0,0 +1,657 @@
...
+        let node_indexing_tasks = self
+            .indexing_tasks_per_node_id
+            .get_mut(node_id)
+            .expect("The node must always be present.");
+        node_indexing_tasks.push(indexing_task);
+    }
+
+    /// Returns the number of indexing tasks for the given node ID.
+    pub fn num_indexing_tasks_for_node(&self, node_id: &String) -> usize {
```
fulmicoton, line 61:
```suggestion
    pub fn num_indexing_tasks_for_node(&self, node_id: &str) -> usize {
```

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -0,0 +1,657 @@
...
+            indexing_tasks_per_node_id: indexing_tasks_per_indexer,
+        }
+    }
+
+    /// Adds the indexing task to the list of tasks assigned to the given node ID.
+    pub fn assign_indexing_task(&mut self, node_id: &String, indexing_task: IndexingTask) {
+        let node_indexing_tasks = self
+            .indexing_tasks_per_node_id
+            .get_mut(node_id)
+            .expect("The node must always be present.");
```
fulmicoton, line 56: this method has a hidden contract and does not need to be public (if node_id is not present it panics)

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -0,0 +1,657 @@
...
+use quickwit_proto::indexing_api::IndexingTask;
+use serde::Serialize;
+
+/// A [`PhysicalIndexingPlan`] defines the list of indexing tasks
+/// each indexer, identified by its node ID, should run.
+/// TODO(fmassot): a metastore version number will be attached to the plan
+/// to identify if the plan is up to date with the metastore.
+#[derive(Debug, PartialEq, Clone, Serialize)]
+pub struct PhysicalIndexingPlan {
+    indexing_tasks_per_node_id: HashMap<String, Vec<IndexingTask>>,
```
fulmicoton, line 35: This function exposes a lot of public methods with quirky contract.

I'd feel safer if there was a builder and the actual plan was just a read only object with a restricted number of methods.

quickwit/quickwit-control-plane/src/indexing_plan.rs
fulmicoton: Looking a bit more in the code, I really think this assert is a foot gun.

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -0,0 +1,657 @@
...
+    }
+
+    /// Returns the hashmap of (node ID, indexing tasks).
+    pub fn indexing_tasks_per_node(&self) -> &HashMap<String, Vec<IndexingTask>> {
+        &self.indexing_tasks_per_node_id
+    }
+
+    /// Returns the mean number of indexing tasks per node.
+    pub fn num_indexing_tasks_mean_per_node(&self) -> f32 {
+        let num_nodes = self.indexing_tasks_per_node_id.len();
```
fulmicoton, line 88: If you ditch the assert, make sure the `indexing_tasks_per_node_id` is non empty.

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -0,0 +1,657 @@
...
+            .values()
+            .map(|tasks| tasks.len())
+            .max()
+            .unwrap_or(0)
+    }
+}
+
+#[derive(Debug)]
+struct NodeScore<'a> {
+    pub node_id: &'a String,
```
fulmicoton, line 124:
```suggestion
    pub node_id: &'a str,
```

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,585 @@
...
+            cluster,
+            metastore,
+            indexing_client_pool,
+            state: IndexingSchedulerState::default(),
+        }
+    }
+
+    async fn schedule_indexing_plan(&mut self) -> anyhow::Result<()> {
+        let indexers = self.get_indexers_from_cluster_state().await;
+        if indexers.is_empty() {
```
fulmicoton, line 136: so that's where the contract was enforced. That's two function calls away. It is a bit scary.

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,585 @@
...
+                    .map(move |(source_id, source_config)| {
+                        (
+                            IndexSourceId {
+                                index_id: index_metadata.index_config.index_id.to_string(),
+                                source_id,
+                            },
+                            source_config,
+                        )
+                    })
+            })
```
fulmicoton, line 156: nitpick

```suggestion
            .flat_map(|index_metadata| index_metadata.sources)
            .map(move |(source_id, source_config)| {
                        (
                            IndexSourceId {
                                index_id: index_metadata.index_config.index_id.to_string(),
                                source_id,
                            },
                            source_config,
                       )
            })
```

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,585 @@
...
+        .collect();
+    let common_node_ids = running_node_ids.intersection(&planned_node_ids);
+    let missing_planned_node_ids = planned_node_ids.difference(&running_node_ids).collect_vec();
+    let unplanned_node_ids = running_node_ids.difference(&planned_node_ids);
+
+    for unplanned_node_id in unplanned_node_ids {
+        plans_are_equal = false;
+        let running_tasks = running_plan
+            .get(*unplanned_node_id)
+            .expect("The node ID must be the HashMap.");
```
fulmicoton, line 339: Really? I am surprsed by this contract. I suspect a foot gun.

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,585 @@
...
+            );
+        }
+    }
+
+    for missing_planned_node_id in missing_planned_node_ids {
+        // An indexer may have left the cluster.
+        plans_are_equal = false;
+        let planned_tasks = last_applied_plan
+            .get(*missing_planned_node_id)
+            .expect("The node ID must be the HashMap.");
```
fulmicoton, line 356: Same as above.

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,585 @@
...
+            "The indexer `{missing_planned_node_id}` is not part of the the running plan, none of \
+             his tasks are running: {planned_tasks:?}."
+        );
+    }
+
+    // Compare indexing tasks for each common node ID.
+    for common_node_id in common_node_ids {
+        let running_tasks: HashSet<_> = running_plan
+            .get(*common_node_id)
+            .expect("The node ID must be the HashMap.")
```
fulmicoton, line 367: Same as above.

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,585 @@
...
+/// Returns true if equal, if not, logs the difference between the plans.
+/// The plans are considered equal if they have the same node IDs and if
+/// each node has the same running tasks.
+fn are_indexing_plans_equal(
+    running_plan: &HashMap<String, Vec<IndexingTask>>,
+    last_applied_plan: &HashMap<String, Vec<IndexingTask>>,
+) -> bool {
+    let mut plans_are_equal = true;
+    // Compare node IDs with at least one task.
+    let running_node_ids: HashSet<&String> =
```
fulmicoton, line 325:
```suggestion
    let running_node_ids: HashSet<&str> =
```

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,585 @@
...
+    let unplanned_node_ids = running_node_ids.difference(&planned_node_ids);
+
+    for unplanned_node_id in unplanned_node_ids {
+        plans_are_equal = false;
+        let running_tasks = running_plan
+            .get(*unplanned_node_id)
+            .expect("The node ID must be the HashMap.");
+        if running_tasks.is_empty() {
+            // An new indexer may have joined the cluster.
+            warn!("The indexer `{unplanned_node_id}` is not part of the last applied plan.")
```
fulmicoton, line 342: why is this a warning? Isn't it normal if this is a new node?

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,585 @@
...
+        // `MIN_DURATION_BETWEEN_SCHEDULING_SECS` value for test and correclty test it. Currently, I
+        // put its value 0 to make test simple.
+        if let Some(last_applied_plan_timestamp) = self.state.last_applied_plan_timestamp.as_ref() {
+            if (OffsetDateTime::now_utc().unix_timestamp() - last_applied_plan_timestamp)
+                < MIN_DURATION_BETWEEN_SCHEDULING_SECS as i64
+            {
+                return Ok(());
+            }
+        }
+        let indexers = self.get_indexers_from_cluster_state().await;
```
fulmicoton, line 187: can you add the type of indexers here? 
```suggestion
        let indexers = self.get_indexers_from_cluster_state().await;
```

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,585 @@
...
+                        .unwrap_or_default(),
+                )
+            })
+            .collect();
+        if !are_indexing_plans_equal(
+            &running_indexing_tasks_by_node_id,
+            last_applied_plan.indexing_tasks_per_node(),
+        ) {
+            info!("Running plan and last applied plan differs: schedule an indexing plan.");
+            self.schedule_indexing_plan().await?;
```
fulmicoton, line 205: I wonder if reschedule is always the right choice here...

If the nodeset is unchanged, this is not the best way to go is it? 
Can we at least add a TODO and create an issue to revisit this logic?

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,585 @@
...
+            .collect_vec()
+    }
+
+    async fn apply_physical_indexing_plan(
+        &mut self,
+        indexers: &[ClusterMember],
+        new_physical_plan: PhysicalIndexingPlan,
+    ) {
+        info!("Apply physical indexing plan: {:?}", new_physical_plan);
+        for (node_id, indexing_tasks) in new_physical_plan.indexing_tasks_per_node().iter() {
```
fulmicoton, line 225: nitpick
```suggestion
        for (node_id, indexing_tasks) in new_physical_plan.indexing_tasks_per_node() {
```

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,585 @@
...
+            .await;
+        Ok(())
+    }
+}
+
+/// Returns true if equal, if not, logs the difference between the plans.
+/// The plans are considered equal if they have the same node IDs and if
+/// each node has the same running tasks.
+fn are_indexing_plans_equal(
+    running_plan: &HashMap<String, Vec<IndexingTask>>,
```
fulmicoton, line 320: Can you detail how we got the running_plan? Is it surfaced by chitchat?

fulmicoton (changes requested):
I am a bit scared of the justification for all of the `.expect()`.

Maybe they are all justified.... but even then at least one of them is not all local to this file, and do constitute a footgun for the future.

For the other, proofreading the justification of the expects/asserts is non-trivial.

quickwit/quickwit-control-plane/src/grpc_adapter.rs
fmassot: Indeed. Maybe a task for @ChillFish8 :)

quickwit/quickwit-control-plane/src/scheduler.rs
fmassot: Yes, it comes from chitchat, just added it in the docs.

quickwit/quickwit-control-plane/src/indexing_plan.rs
fmassot: I removed the hidden contract stuff and foot guns (asserts)...
I did not go for the builder as it seems safer now without asserts and expects.

quickwit/quickwit-control-plane/src/scheduler.rs
fmassot: Indeed, we probably want just to reapply the same plan on the nodes where we don't have the same tasks.

fmassot:
@fulmicoton PR comments taken into account.

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -0,0 +1,655 @@
...
+
+impl PhysicalIndexingPlan {
+    fn new(node_ids: Vec<String>) -> Self {
+        let indexing_tasks_per_indexer = node_ids
+            .into_iter()
+            .map(|node_id| (node_id, Vec::new()))
+            .collect();
+        Self {
+            indexing_tasks_per_node_id: indexing_tasks_per_indexer,
+        }
```
fulmicoton, line 46:
```suggestion
        let indexing_tasks_per_node_id = node_ids
            .into_iter()
            .map(|node_id| (node_id, Vec::new()))
            .collect();
        Self {
            indexing_tasks_per_node_id,
        }
```

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -0,0 +1,655 @@
...
+        self.indexing_tasks_per_node_id
+            .get(node_id)
+            .map(|tasks| tasks.len())
+            .unwrap_or(0)
+    }
+
+    pub fn assign_indexing_task(&mut self, node_id: String, indexing_task: IndexingTask) {
+        self.indexing_tasks_per_node_id
+            .entry(node_id)
+            .or_insert_with(Vec::new)
```
fulmicoton, line 60:
```suggestion
            .or_default()
```

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -0,0 +1,655 @@
...
+        // Score each candidate and select the best node.
+        let best_node_score = candidates
+            .iter()
+            .map(|node_id| NodeScore {
+                node_id,
+                score: -compute_node_score(node_id, &plan),
+            })
+            .sorted()
+            .next()
+            .expect("There is always at least one candidate");
```
fulmicoton, line 206:
```suggestion
        let best_node_score = candidates
            .iter()
            .max_by_key(|node_id|)
            .map(|node_id| compute_node_score(node_id, &plan))
            .expect("There is always at least one candidate");
```

fulmicoton, line 206:
```suggestion
        let Some(best_node_score) = candidates
            .iter()
            .max_by_key(|node_id|)
            .map(|node_id| compute_node_score(node_id, &plan)) else {
        tracing::warn!(indexing_task=?indexing_task, "No indexer candidate available for the indexing task, cannot assign it to an indexer. This should not happen.");
            continue;
       };
```

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -0,0 +1,655 @@
...
+        .iter()
+        .sorted_by(|(left, _), (right, _)| left.cmp(right))
+    {
+        if !source_config.enabled {
+            continue;
+        }
+        if source_config.source_id == CLI_INGEST_SOURCE_ID {
+            continue;
+        }
+        // Ignore file sources as we don't know thee file location.
```
fulmicoton, line 288:
```suggestion
        // Ignore file sources as we don't know the file location.
```

fulmicoton: can you open a ticket to investigate what should be done with file sources?

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,598 @@
...
+                        )
+                    })
+            })
+            .collect();
+        Ok(source_configs)
+    }
+
+    /// Checks if the last applied plan corresponds to the running indexing tasks present in the
+    /// chitchat cluster state. If true, do nothing.
+    /// If false, schedule a new indexing plan.
```
fulmicoton, line 174: Triggering a new schedule plan seems a little bit harsh here.
Can you open a ticket to revisit this logic?

I suspcet it could easily end up in a crazy loop if a single node is faulty, rendering the entire cluster unusablle.
For instance, imagine a node that is missing hard disk.. or a node that cannot connect to kafka. etc.

In fact it is probably worth thinking whetehr an errorr is ndoe speicfic (the node cannot run a task), or source specific (won't work on another node).... Try to infer it... and act differently depending on that.


We could:
a) Send an RPC to try to enforce the plan (also I agree this is strange considering we pass the plan through gRPC already). Afterall. maybe chitchat is just lagging. The node should then tell us if 
it fails to start the pipeline in case we should do b). If we get a success, then we wait.
b)  If the node cannot be accessed, manually marking the faulty node as faulty, putting out of the membership and trigger a replan. This requires refining our membership definition


More generally, my hunch is that we could have decoupled the two following tasks:
- computing a new plan if necessary.
- making sure the plan is applied.

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,598 @@
...
+    type Reply = ();
+
+    async fn handle(
+        &mut self,
+        _message: RefreshPlanLoop,
+        ctx: &ActorContext<Self>,
+    ) -> Result<(), ActorExitStatus> {
+        if let Err(error) = self.schedule_indexing_plan().await {
+            error!("Error when scheduling indexing plan: `{}`.", error);
+        }
```
fulmicoton, line 318: nitpick 

```suggestion
        self.schedule_indexing_plan().await
               .context("forma!!("Error when scheduling indexing plan: `{error}`.")?;
```

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -17,37 +17,598 @@
...
+            info!(
+                "The indexer `{unplanned_node_id}`, which is not part of the last applied plan, \
+                 has some running tasks: {running_tasks:?}."
+            );
+        }
+    }
+
+    for missing_planned_node_id in missing_planned_node_ids {
+        // An indexer may have left the cluster.
+        plans_are_equal = false;
```
fulmicoton, line 367:
```suggestion
```

quickwit/quickwit-metastore/src/metastore/metastore_with_control_plane_triggers.rs
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    async fn check_connectivity(&self) -> anyhow::Result<()> {
+        self.metastore.check_connectivity().await
+    }
+
+    // Source API
+
+    async fn add_source(&self, index_id: &str, source: SourceConfig) -> MetastoreResult<()> {
+        let should_send_event_to_control_plane = source.source_id != CLI_INGEST_SOURCE_ID;
```
fulmicoton, line 172: what about the file source id?

quickwit/quickwit-metastore/src/metastore/metastore_with_control_plane_triggers.rs
```diff
@@ -0,0 +1,382 @@
...
+    ) -> Self {
+        Self {
+            metastore,
+            control_plane_client,
+        }
+    }
+}
+
+// Send an event to the Control Plane. The `action` is just there for logging.
+fn send_event_to_control_plane(mut control_plane_client: ControlPlaneGrpcClient, action: String) {
```
fulmicoton, line 63:
```suggestion
fn send_event_to_control_plane(mut control_plane_client: ControlPlaneGrpcClient, action: &'static str) {
```

quickwit/quickwit-metastore/src/metastore/metastore_with_control_plane_triggers.rs
```diff
@@ -0,0 +1,382 @@
...
+    tokio::spawn(async move {
+        if let Err(error) = control_plane_client.notify_index_change().await {
+            error!(
+                "Failed to send index event to the control plane on `{}`: `{}`.",
+                action, error
```
fulmicoton, line 68:
```suggestion
                "Failed to send index event to the control plane on `{action}`: `{error}`."
```

quickwit/quickwit-serve/src/tests.rs
```diff
@@ -118,6 +120,9 @@ async fn test_multi_nodes_cluster() {
 
     assert!(sandbox.rest_client.is_live().await.unwrap());
 
+    // Wait until indexing pipelines are started.
+    // TODO(fmassot): try to reduce the duration or use a wait until condition.
+    tokio::time::sleep(Duration::from_secs(3)).await;
```
fulmicoton, line 125: have you tries to use the mocking time stuff?

fulmicoton (approved):
see comment inline. I think we need to create a bunch of issues.

quickwit/quickwit-serve/src/tests.rs
fmassot: not yet. Will try it.

PR #2637, opened 2023-01-09
Fixing sleep

fulmicoton:
Removing the concept of jump a given amount of time.
Instead, we just have Universe::sleep and a mode in which
the time is accelerated.

### DIFF
--- a/quickwit/quickwit-actors/src/universe.rs
+++ b/quickwit/quickwit-actors/src/universe.rs
@@ -52,7 +52,7 @@ impl Universe {
     /// Creates a universe were time is accelerated.
     ///
     /// Time is accelerated in a way to exhibit a behavior as close as possible
-    /// the what would have happened with normal time but faster.
+    /// to what would have happened with normal time but faster.
     ///
     /// The time "jumps" only happen when no actor is processing any message,
     /// running initialization or finalize.
--- a/quickwit/quickwit-actors/src/universe.rs
+++ b/quickwit/quickwit-actors/src/universe.rs
@@ -95,6 +95,11 @@ impl Universe {
         self.spawn_ctx.kill_switch.kill();
     }
 
+    /// This function acts as a drop-in replacement of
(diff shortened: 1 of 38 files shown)


quickwit/quickwit-actors/src/universe.rs
```diff
@@ -81,17 +95,8 @@ impl Universe {
...
-    /// This is smarter than just mocking a clock.
-    /// As long as everything is happening in actors handlers, initialize
-    /// or finalize methods, the result should be rigorously the same as
-    /// if we waited.
-    pub async fn simulate_time_shift(&self, duration: Duration) {
-        self.spawn_ctx
-            .scheduler_client
-            .simulate_sleep(duration)
-            .await;
+    pub async fn sleep(&self, duration: Duration) {
```
evanxg852000, line 98: The `simulate_time_shift` seems a better name for this IMHO. We need the bring back the updated docs comment

evanxg852000 (approved):
LGTM :+1: 
See my suggestions and looking forward to your take on some of my comments.

quickwit/quickwit-aws/src/retry.rs
```diff
@@ -162,12 +162,13 @@ mod tests {
     #[async_trait]
     impl super::MockableTime for SchedulerClient {
         async fn sleep(&self, duration: Duration) {
-            self.simulate_sleep(duration).await;
+            SchedulerClient::sleep(&self, duration).await;
```
evanxg852000, line 165:
```suggestion
            SchedulerClient::sleep(self, duration).await;
```

quickwit/quickwit-actors/src/universe.rs
fulmicoton: It is not the same semantics as it used to be. it does not accelerate time in itself anymore.

Instead, its semantics is "just sleep".

But the universe can created with an accelerated time.

PR #2675, opened 2023-01-19
Yield before calling on drained callback

guilload:
### Description
`try_recv` returning `None` means that the channel is empty but not necessarily drained, i.e., an actor can exist at the other end of the channel waiting on `send`. The only way to find out is to yield and check whether the channel is empty when the task is scheduled again.

### How was this PR tested?
- Updated unit tests
- `make test-all`

### DIFF
--- a/quickwit/quickwit-actors/src/mailbox.rs
+++ b/quickwit/quickwit-actors/src/mailbox.rs
@@ -356,6 +356,10 @@ impl<A: Actor> Clone for Inbox<A> {
 }
 
 impl<A: Actor> Inbox<A> {
+    pub(crate) fn is_empty(&self) -> bool {
+        self.rx.is_empty()
+    }
+
     pub(crate) async fn recv(&self) -> Result<Envelope<A>, RecvError> {
         self.rx.recv().await
     }
--- a/quickwit/quickwit-actors/src/mailbox.rs
+++ b/quickwit/quickwit-actors/src/mailbox.rs
@@ -506,7 +506,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_mailbox_send_with_backpressure_counter_no_backpressure_cleansheet() {
+    async fn test_mailbox_send_with_backpressure_counter_low_backpressure() {
         let universe = Universe::with_accelerated_time();
         let back_pressure_actor = BackPressureActor;
         let (mailbox, _handle) = universe.spawn_builder().spawn(back_pressure_actor);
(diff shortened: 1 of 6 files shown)


fulmicoton (approved):

quickwit/quickwit-actors/src/universe.rs
```diff
@@ -99,11 +101,24 @@ impl Universe {
...
     pub async fn sleep(&self, duration: Duration) {
         self.spawn_ctx.scheduler_client.sleep(duration).await;
     }
 
+    /// This function acts as a drop-in replacement of
+    /// `tokio::time::sleep`.
+    ///
+    /// It can however be accelerated when using a time-accelerated
+    /// universe.
+    pub async fn timeout<O>(
```
guilload, line 114: I added this to fix the unit tests for my first attempt but did not use it for the second. I chose to leave it.

quickwit/quickwit-actors/src/spawn_builder.rs
```diff
@@ -268,11 +269,24 @@ impl<A: Actor> ActorExecutionEnv<A> {
 
     async fn process_all_available_messages(&mut self) -> Result<(), ActorExitStatus> {
         self.yield_and_check_if_killed().await?;
-        let envelope = get_envelope(&mut self.inbox, &self.ctx).await;
```
guilload, line 271: I ended up reworking my first fix because it added another pass in `process_one_message` for all the actors to yield, leading to some unit tests failing.

quickwit/quickwit-actors/src/mailbox.rs
```diff
@@ -523,9 +529,9 @@ mod tests {
             )
             .await
             .unwrap();
-        assert_eq!(backpressure_micros_counter.get(), 0u64);
+        assert!(backpressure_micros_counter.get() < 500);
```
guilload, line 532: With this fix, we can't get without adding a bit of latency because of the additional "task" switch. Measured at less than 100 µs on my laptop (debug mode).

guilload (commented):
@fulmicoton, I could use another review because I've made significant changes in the fix since you approved the PR. I think this new fix is better.

quickwit/quickwit-actors/src/universe.rs
```diff
@@ -99,11 +101,24 @@ impl Universe {
...
     ///
     /// It can however be accelerated when using a time-accelerated
-    /// Universe.
+    /// universe.
     pub async fn sleep(&self, duration: Duration) {
         self.spawn_ctx.scheduler_client.sleep(duration).await;
     }
 
+    /// This function acts as a drop-in replacement of
+    /// `tokio::time::sleep`.
```
fulmicoton, line 110: timeout

PR #2676, opened 2023-01-19
Change doc mapper signature

guilload:
### Description
Use `serde_json::Map<String, JsonValue>` instead of `&str` as input of the doc mapper parsing method so we can use the doc mapper directly on docs that already have been parsed as part of the transform process.

Indeed, I'm working on finalizing the VRL PR, and I'm making some changes to the doc processor. The code looks like this:

```rust
let doc_res = if let Some(transform) = self.transform_opt {
    let vrl_value: VrlValue = transform.transform_doc(&json_doc)?;
    let json_value: JsonValue = serde_json::to_value(vrl_doc).map_err(...);
    let JsonValue::Object(json_obj) = json_value else ...
    doc_mapper.doc_from_json_obj(json_obj)
} else {
    doc_mapper.doc_from_json_str(&json_doc)
}
```

Without this change, we have to serialize the `VrlValue` to a JSON string and then back to a `JsonValue`. This is painfully slow.

### How was this PR tested?
`make test-all`


fmassot (approved):
LGTM.

PR #2682, opened 2023-01-20
Basic Generation of Elastic  compatible API 

evanxg852000:
Implement a task to download Elastic API specs and generate warp handlers and types for the endpoints.


quickwit/quickwit-serve/README.md
```diff
@@ -42,3 +42,13 @@ The server also exposes at `/ui` all static files located in `quickwit-ui/build`
 produced by the react app build in `quickwit-ui`.
 During development, the server will serve the local files. When building the binary, these static files will be embedded in it.
 
+## Elastic Compatible API
+
+To help implement Elasticsearch compatible API in Quickwit, we generate endpoints and related types using the [elastic-api-generator](./src/elastic_api_generator.rs) binary.
+
+To update the generated file, you will have to:
```
guilload, line 49: Let's automate this further as part of a `build.rs` script and remove the bin:
1. If the spec file does not exist, download it. Otherwise, issue a HEAD request to compare the length and/or modification date and download if necessary (bonus point for conditional GET)
2. Run code generation if we detect a change.
3. Disable by default but run in CI on release builds
4. Add the ability to opt-in via.. env var maybe?

quickwit/quickwit-serve/src/elastic_api_generator.rs
```diff
@@ -0,0 +1,98 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+/// This file is auto-generated, any change can be overriden.
+"#;
+
+pub static ROOT_DIR: Lazy<path::PathBuf> = Lazy::new(|| {
+    let mf = env::var("CARGO_MANIFEST_DIR").expect("Should be run using 'cargo xtask ...'");
+    path::Path::new(&mf).parent().unwrap().to_owned()
```
guilload, line 53: nit: `to_path_buf` to stay consistent with the code style of the codebase.

quickwit/quickwit-serve/src/elastic_api_generator.rs
```diff
@@ -0,0 +1,98 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+/// This file is auto-generated, any change can be overriden.
+"#;
+
+pub static ROOT_DIR: Lazy<path::PathBuf> = Lazy::new(|| {
+    let mf = env::var("CARGO_MANIFEST_DIR").expect("Should be run using 'cargo xtask ...'");
```
guilload, line 52: I don't get the reference to xtask here. Does this PR require xtask to run codegen?

quickwit/quickwit-serve/src/elastic_api_generator.rs
```diff
@@ -0,0 +1,98 @@
...
+pub const SELECTED_SPEC_FILES: &[&str] = &[
+    "search.json",
+    // Add spec file for endpoints you want to support.
+];
+
+fn main() -> anyhow::Result<()> {
+    let command_opt = Command::new("elastic-api-generator")
+        .about("Quickwit Elasticsearch compatible API generator CLI")
+        .subcommand_required(true)
+        .subcommand(Command::new("download").about("Download the Elasticsearch API spec files."))
```
guilload, line 70:
```suggestion
        .subcommand(Command::new("download").about("Downloads the Elasticsearch API spec files."))
```

quickwit/quickwit-serve/src/elastic_search_api/api_specs.rs
```diff
@@ -0,0 +1,214 @@
...
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+/// This file is auto-generated, any change can be overriden.
+use serde::{Deserialize, Serialize};
+use warp::{Filter, Rejection};
+
+use super::{from_simple_list, to_simple_list, SimpleList, TrackTotalHits};
+#[serde_with::skip_serializing_none]
+#[derive(Default, Debug, Serialize, Deserialize)]
+pub struct SearchQueryParams {
```
guilload, line 27: This is scary :D

quickwit/quickwit-serve/src/elastic_search_api/api_specs.rs
```diff
@@ -0,0 +1,214 @@
...
+#[derive(Debug, PartialEq, Deserialize, Serialize, Clone, Copy)]
+pub enum SuggestMode {
+    #[serde(rename = "missing")]
+    Missing,
+    #[serde(rename = "popular")]
+    Popular,
+    #[serde(rename = "always")]
+    Always,
+}
+#[utoipa::path(get, tag = "Search", path = "/_search")]
```
guilload, line 187: Nice :)

quickwit/quickwit-serve/src/elastic_search_api/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+        .trim_matches(',')
+        .split(',')
+        .map(|item| item.to_owned())
+        .collect::<Vec<_>>();
+    Ok(list)
+}
+
+/// A helper struct to serialize/deserialize a comma separated list.
+#[derive(Debug, Deserialize)]
+pub(crate) struct SimpleList(#[serde(deserialize_with = "from_comma_list")] Vec<String>);
```
guilload, line 61: If you implement `FromStr` on a struct, I believe `serde` can use that to generate a deserializer.

guilload (approved):

guilload:
Sweet. I believe the whole `from_comma_list`, `from_simple_list` stuff can be simplified.

quickwit/quickwit-serve/src/elastic_api_generator.rs
evanxg852000: yes I dropped xtask

quickwit/quickwit-serve/README.md
evanxg852000: I am ok with doing the above if you feel it's still useful now.
However, my take is that this is not a good candidate for automated build.rs at least right now that we are not managing versions. We are currently working based on a released version of the Elastic stack. Meaning these specs files won't change unless we switch versions. 
I would think of this type of automation if we were managing different versions or grabbing the latest from the repository. 
This script is meant to be intentionally run when you want to support more Elasticsearch endpoints in Quickwit. 
I don't see the point of generating it from build.rs step and cargo-formating it if it's not going to change. The only time it changes is when we include a new endpoint spec file which should be intentional. Because this will require writing and registering the handlers in warp server.

evanxg852000:
@guilload  I have removed `from_comma_list` and currently using the implementation of FromStr and ToString to handle ser/deserialsation. Let me know if you have other simplification in  mind. Thanks.

guilload (commented):
Just one nit: Elasticsearch is one word :)

PR #2689, opened 2023-01-21
Instrument Jaeger service

guilload:
### Description
Instrument Jaeger service and fix OTLP service metric labels.

### How was this PR tested?
Ran `quickwit run` and Checked `/metrics`.


fmassot (approved):

quickwit/quickwit-opentelemetry/src/otlp/trace.rs
```diff
@@ -385,7 +385,7 @@ impl OtlpGrpcTraceService {
             .sum::<usize>() as u64;
         let num_bytes = request.encoded_len() as u64;
 
-        let labels = ["trace", "grpc", OTEL_TRACE_INDEX_ID, "protobuf"];
+        let labels = ["trace", OTEL_TRACE_INDEX_ID, "grpc", "protobuf"];
```
fmassot, line 388: I may be wrong because I don't know the code well here, but you have 4 labels here, and you declare only `["operation", "index"]` for this metric.

guilload: This is the metric for the `OTLP` service, not Jaeger. I had wrongly swapped two labels. The compiler checks the number of labels:

```rust
pub fn new_counter_vec<const N: usize>(
    name: &str,
    description: &str,
    namespace: &str,
    label_names: [&str; N],
) -> IntCounterVec<N> {
```

PR #2695, opened 2023-01-21
Using rendez-vous hashing to sort nodes used for scheduling.

### LINKED ISSUE #2661: Improve the stability of the plan scheduling on control plane restart or on new/deteted sources
Currently, the scheduling is dumb, we sort tasks by (index_id, source_id) and place them on nodes.
When sources are added/deleted or when the control plane restart, this could trigger a lot of placement changes.

We have to stabilize this. Two possibilities:
- use a hash (rendezvous hashing like for search)
- use the running plan retrieved from chitchat. The drawback is that, when the control plane restart, we need to wait a bit so that the chitchat state has converged.

fmassot:
Fix #2661

### DIFF
--- a/quickwit/quickwit-common/src/rendezvous_hasher.rs
+++ b/quickwit/quickwit-common/src/rendezvous_hasher.rs
@@ -21,20 +21,20 @@ use std::cmp::Reverse;
 use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
 
-/// Computes the affinity of a node with a given `split_id`.
+/// Computes the affinity of a node with a given `key`.
 ///
 /// We rely on rendez-vous hashing here.
-fn node_affinity<T: Hash>(node: T, split_id: &str) -> u64 {
+fn node_affinity<T: Hash, U: Hash>(node: T, key: &U) -> u64 {
     let mut state = DefaultHasher::new();
-    split_id.hash(&mut state);
+    key.hash(&mut state);
     node.hash(&mut state);
     state.finish()
 }
 
 /// Sorts the list of node base on rendez-vous-hashing.
 /// Nodes are ordered by decreasing order of computed `hash_key`
-pub(crate) fn sort_by_rendez_vous_hash<T: Hash>(nodes: &mut [T], split_id: &str) {
-    nodes.sort_by_cached_key(|node| Reverse(node_affinity(node, split_id)));
+pub fn sort_by_rendez_vous_hash<T: Hash, U: Hash>(nodes: &mut [T], key: U) {
+    nodes.sort_by_cached_key(|node| Reverse(node_affinity(node, &key)));
 }
 
 #[cfg(test)]
--- a/quickwit/quickwit-common/src/rendezvous_hasher.rs
+++ b/quickwit/quickwit-common/src/rendezvous_hasher.rs
@@ -17,24 +17,21 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
-use std::cmp::Reverse;
 use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
 
-/// Computes the affinity of a node with a given `key`.
-///
-/// We rely on rendez-vous hashing here.
-fn node_affinity<T: Hash, U: Hash>(node: T, key: &U) -> u64 {
+/// Computes the hash of the `node` and `key`.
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-common/src/rendezvous_hasher.rs
```diff
@@ -21,20 +21,20 @@ use std::cmp::Reverse;
...
     node.hash(&mut state);
     state.finish()
 }
 
 /// Sorts the list of node base on rendez-vous-hashing.
 /// Nodes are ordered by decreasing order of computed `hash_key`
-pub(crate) fn sort_by_rendez_vous_hash<T: Hash>(nodes: &mut [T], split_id: &str) {
-    nodes.sort_by_cached_key(|node| Reverse(node_affinity(node, split_id)));
+pub fn sort_by_rendez_vous_hash<T: Hash, U: Hash>(nodes: &mut [T], key: U) {
+    nodes.sort_by_cached_key(|node| Reverse(node_affinity(node, &key)));
```
fulmicoton, line 37: I know this is not part of your code, but do we actually care about the Reverse part here?
Do we use the node_affinity function somewhere else? 
```suggestion
    nodes.sort_by_cached_key(|node| node_affinity(node, &key));
```

fmassot: `node_affiinity` is used only here. And actually, I don't know why we put `Reverse` :/

fmassot: Mmmm, it seems that `node_affinity` is a bit misleading.
I would rename `node_affinity` by `rendez_vous_hash` and remove the `Reverse`.

guilload: `Reverse` for highest affinity first:

```rust
for job in jobs {
    sort_by_rendez_vous_hash(&mut nodes, job.split_id());
    // choose one of the the first two nodes based on least loaded
    let chosen_node_index: usize = if nodes.len() >= 2 {
        usize::from(nodes[0].load > nodes[1].load)
    } else {
        0
    };
```

guilload: I'm smoking t_t

fmassot: Also I'm wondering why we did not choose to node weights to deduce a score from hash + weigth.

fulmicoton: > did not choose to use node weights to deduce a score from hash + weigth

I am not entirely sure what this would mean.

fulmicoton: I think `node_affinity` is good. what makes it misleading?

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -192,24 +166,29 @@ pub(crate) fn build_physical_indexing_plan(
...
-            // NodeScore implements Ord and can be used easily for sorting.
-            .map(|node_id| NodeScore {
-                node_id,
-                score: -compute_node_score(node_id, &plan),
-            })
-            .sorted()
+            .map(|&node_id| (node_id, compute_node_score(node_id, &plan)))
+            // Note(fmassot): we can't use max_by as, in case of equality, it returns the last
+            // element and we want to preserve the order given by the rendez-vous hashing.
+            .sorted_by(cmp_node_by_reverse_score)
```
fulmicoton, line 172:
```suggestion
            .rev() //< we use the reverse iterator, because in case of a tie, max picks the first element.
            .map(|&node_id| (node_id, compute_node_score(node_id, &plan)))
            .max_by_key(cmp_node_by_reverse_score)
```

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -192,24 +166,29 @@ pub(crate) fn build_physical_indexing_plan(
         // source.
         let best_node_opt: Option<&str> = candidates
             .iter()
-            // NodeScore implements Ord and can be used easily for sorting.
-            .map(|node_id| NodeScore {
-                node_id,
-                score: -compute_node_score(node_id, &plan),
-            })
-            .sorted()
+            .map(|&node_id| (node_id, compute_node_score(node_id, &plan)))
```
fulmicoton, line 169: I'm not sure exchange the struct with a pair was an improvement.

fulmicoton (approved):

PR #2702, opened 2023-01-24
QW-2584: Add pulsar source

ChillFish8:
Adds the ability to use Pulsar as an ingestion source.

This PR is currently a WIP while I make some changes to the `pulsar-rs` code in order to work with our checkpointing system and add tests.


quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -0,0 +1,681 @@
...
+            "Starting Pulsar source."
+        );
+
+        let pulsar: Pulsar<_> = builder.build().await?;
+
+        let partition_id = PartitionId::from(params.topic.as_str());
+        let seek_to = checkpoint
+            .position_for_partition(&partition_id)
+            .and_then(msg_id_from_position);
+        let pulsar = spawn_message_listener(params.clone(), pulsar, seek_to).await?;
```
ChillFish8, line 154: It's worth noting here that we're using the topic name as the partition ID currently as there are some outstanding questions I'm still looking into around how Pulsar behaves when seeking and acking messages in partitions.

From what I understand it is not really designed to be used in situations where you're using seeking like we are to do the at least once semantics, and because of that, I am still trying to understand how to go about handling partitioned topics without explicitly declaring each indexer's partition and without causing us to accidentally loose messages.

guilload:
Can you add the Docker Compose setup, please?

quickwit/quickwit-config/src/source_config/mod.rs
```diff
@@ -346,6 +350,38 @@ pub struct VecSourceParams {
 #[serde(deny_unknown_fields)]
 pub struct VoidSourceParams;
 
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize, utoipa::ToSchema)]
+#[serde(deny_unknown_fields)]
+pub struct PulsarSourceParams {
+    /// List of the topics that the source consumes.
+    pub topics: Vec<String>,
+    /// The name of the topic subscription.
+    pub subscription: String,
```
fmassot, line 359: I'm not sure what is the purpose of this `subscription` parameter. Maybe rename it `subscription_name`. Could we also have a `default_subscription_name` if it makes sense?

ChillFish8: Honestly this might be safer to not have this be configurable and instead generate this ourselves.

guilload: Yes, for the Kafka source, users ended up using `quickwit` as a group ID for all their sources... so we removed it.

ChillFish8: I have updated this and removed the parameter.

The reason for removal is that this is potentially a foot gun for users when creating distributed indexers where we may silently fail (pulsar-rs will repeatedly try to re-connect) if two subscriptions with the same name try to consume the source.

ChillFish8: I've made this default to `quickwit-indexer` this is left configurable in case someone has some particular use case for it as subscription names have some semantic meaning to how consumers behave.

guilload: That's no bueno. We want a unique subscription name per source. Use `quickwit-{index_id}-{source_id}`. Otherwise, with a unique subscription name such as `quickwit-indexer`, consumers will consume cross-indexes.

guilload: I could be wrong because I'm mapping what I know from Kafka, and I don't know much of Pulsar, but I suspect subscription name is the primary key that determines which consumers are sharing the same set of partitions/topics.

ChillFish8: I double-check in a moment, but from what I understand what you're saying is correct,

ChillFish8: Yes, you're correct, I've changed it to what you suggested.

quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -0,0 +1,1083 @@
...
+                let partition_id = PartitionId::from(partition);
+                let position_opt = checkpoint.position_for_partition(&partition_id).cloned();
+
+                if let Some(position) = position_opt {
+                    previous_positions.insert(partition_id, position);
+                }
+            }
+        }
+
+        let pulsar = spawn_message_listener(
```
fmassot, line 136: We tend to avoid variable shadowing. And you can really call it `pulsar_consumer` here.

quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -0,0 +1,1083 @@
...
+                .with_subscription(subscription_name)
+                .with_subscription_type(SubType::Failover)
+                .build()
+                .await?;
+
+            let consumer_ids = consumer.consumer_id()
+                .into_iter()
+                .map(|id| id.to_string())
+                .collect::<Vec<_>>();
+            info!(positions = ?previous_positions, "Seeking to last checkpoint position.");
```
fmassot, line 367:
```suggestion
            info!(positions = ?previous_positions, "Seeking to last checkpoint positions.");
```

quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -0,0 +1,1083 @@
...
+            index_id=%ctx.index_id,
+            source_id=%ctx.source_config.source_id,
+            topics=?params.topics,
+            subscription_name=%subscription_name,
+            "Create Pulsar source."
+        );
+
+        let pulsar = connect_pulsar(&params).await?;
+
+        let mut previous_positions = BTreeMap::new();
```
fmassot, line 122: Could be nice to know how partition ID are derived from the topic name. Otherwise, we could think that partitions IDs could clash between topics.

fmassot: `"{topic}-partition-{ID}"`

quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -0,0 +1,1083 @@
...
+        if doc.is_empty() {
+            warn!("Message received from queue was empty.");
+            self.state.num_invalid_messages += 1;
+            return Ok(());
+        }
+
+        let partition = PartitionId::from(topic);
+        let num_bytes = doc.as_bytes().len();
+
+        if let Some(previous) = self.previous_positions.get(&partition) {
```
fmassot, line 189:
```suggestion
        if let Some(previous_position) = self.previous_positions.get(&partition) {
```

quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -0,0 +1,1083 @@
...
+            // We skip messages which are older than the previously recorded position.
+            // This is because Pulsar may replay messages which have not yet been acknowledged but
+            // are in the process of being indexed.
```
fmassot, line 192: > but are in the process of being indexed.

I got confused by this last part. Those messages are already indexed and published as the position is inferior to the previous one.

I would say `but have already been published. This can happen for example if an indexer crashes after a publish and before the ack sent to Pulsar.`

quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -0,0 +1,1083 @@
...
+            if &position < previous {
+                self.state.num_invalid_messages += 1;
+                return Ok(());
+            }
+        }
+
+        let previous_position = self
```
fmassot, line 199: I would avoid the shadowing here too, `updated_positions`?

fmassot: I got it, I think it would be better to have `self.current_positions` instead of `self.previous_positions`.

quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -0,0 +1,1083 @@
...
+        handle.join().expect("Join background thread task.")?;
+    }
+
+    Ok(PulsarConsumer {
+        messages: messages_rx,
+        last_ack: last_ack_tx,
+    })
+}
+
+fn msg_id_to_position(msg: &MessageIdData) -> Position {
```
fmassot, line 419: I think we don't want to support [batch index acknowledgment](https://pulsar.apache.org/docs/2.11.x/concepts-messaging/#batching).

In this case, it seems that `batch_index` and `batch_size` are useless. Did you see it in your tests as we don't enable the batch index ack on the consumer.

fmassot: I'm still misunderstanding the batch index feature though.

quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -0,0 +1,1083 @@
...
+                                break 'outer;
+                            }
+                        }
+                    };
+                }
+
+                if last_ack_rx.has_changed().unwrap_or_default() {
+                    let checkpoint = last_ack_rx.borrow();
+
+                    for (partition, position) in checkpoint.iter() {
```
fmassot, line 393: I think we must be a bit more clearer on what is a partition here.
Here a partition is `{topic}-{partitionId}`. So we should extract the partition ID from it. Storing the partition ID inside the position is redundant and strange from a external viewpoint (we have a btree map of partition ID -> Position).

quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -0,0 +1,1083 @@
...
+        Ok(Duration::default())
+    }
+
+    async fn suggest_truncate(
+        &self,
+        checkpoint: SourceCheckpoint,
+        _ctx: &ActorContext<SourceActor>,
+    ) -> anyhow::Result<()> {
+        debug!(ckpt = ?checkpoint, "Truncating message queue.");
+        let _ = self.pulsar.last_ack.send(checkpoint);
```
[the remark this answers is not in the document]
ChillFish8: This could probably just be an unwrap or expect here, since this should be technically infallible.

fmassot (changes requested):
Nice progress on the pulsar source.

Can you have a look at the comments from @guilload and me? The most important point are for the `Send` consumer and the position that we can potentially simplify.

quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -0,0 +1,1083 @@
...
+struct PulsarConsumer {
+    messages: mpsc::Receiver<ConsumerMessage>,
+    last_ack: watch::Sender<SourceCheckpoint>,
+}
+
+#[tracing::instrument(name = "pulsar-consumer", skip(pulsar))]
+/// Spawns a background task listening for incoming messages to
+/// process.
+///
+/// The exists because the `Consumer` is not `Sync` or `Send` in places which means we cannot
```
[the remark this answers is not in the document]
ChillFish8: I should re-word this comment/doc, the `Consumer` type itself is `Send` but seeking and calling the seeking method is *not* `Send` nor `Sync` which when investigating looked to be due to the fact that `pulsar-rs` makes use of some dynamic dispatch and async streams which do not seem to play nicely with the seek method.

I did attempt to use a Mutex (specifically a tokio::sync::Mutex) but this did not satisfy the compiler, other attempts also include writing a separate actor for the consumer using a tokio task, but this limitation requires a local set to guarantee the future stays on the same thread.

For replicating the issue you specifically need to call `seek` as that is what creates the `!Send` future from what I can tell, which under the hood is re-constructing the clients and replacing itself after seeking which I imagine is where the issue lies.

quickwit/quickwit-indexing/Cargo.toml
```diff
@@ -73,6 +76,8 @@ proptest = { workspace = true }
 rand = { workspace = true }
 tempfile = { workspace = true }
 
+tracing-subscriber = { workspace = true }
```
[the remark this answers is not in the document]
ChillFish8: We do not, left over from debugging

quickwit/quickwit-indexing/Cargo.toml
```diff
@@ -59,6 +61,7 @@ quickwit-storage = { workspace = true }
 [features]
 kafka = ["rdkafka", "backoff"]
 kafka-broker-tests = []
+pulsar-broker-tests = []
```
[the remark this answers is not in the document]
ChillFish8: Can do, I was just mirroring what we have done with kafka, should I do this for kafka while I'm at it?

quickwit/quickwit-indexing/src/source/pulsar_source.rs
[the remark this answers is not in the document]
guilload: Instead of running the whole consumer on a dedicated thread, let's run the `!Send` future in a dedicated runtime pinned to a thread and use a local set. At least we can share that runtime between multiple Pulsar source instances. Otherwise, we require one thread per source instance, which is not great (and an issue for the Kafka source).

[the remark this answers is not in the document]
ChillFish8: > Instead of running the whole consumer on a dedicated thread, let's run the !Send future in a dedicated runtime pinned to a thread and use a local set. At least we can share that runtime between multiple Pulsar source instances. Otherwise, we require one thread per source instance, which is not great (and an issue for the Kafka source).

I also tried this but was unsuccessful in getting this approach to work. 

The mid-ground would be we use tokio-utils'  local set pool which will let use share this runtime.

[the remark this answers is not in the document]
guilload: For future reference, [here](https://gist.github.com/guilload/55fab910d11b7eb23aef4d3af227279f) is how to do it.

... or https://github.com/streamnative/pulsar-rs/pull/255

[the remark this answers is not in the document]
ChillFish8: > https://github.com/streamnative/pulsar-rs/pull/255

Looks like that's the way to go then if we can get that PR merged.

quickwit/quickwit-indexing/src/source/pulsar_source.rs
ChillFish8: we still require `partition` and I also believe `batch_size` if we keep `batch_index`.

fmassot: but `partition` is already in the name of the topic right? And I don't understand why `batch_size` is necessary, can you check that to be sure it's needed?

ChillFish8: `partition` in the message is the specific partition ID, not the partition name (which is what we have as our `Partition`) although technically, yes you could parse the topic name to get the ID, I'm not sure how much I trust that for the sake of stripping it out of the position payload. 

> And I don't understand why batch_size is necessary, can you check that to be sure it's needed?

Yes shall do, it's not something I've really looked into.

fmassot: topic name format: https://github.com/streamnative/pulsar-rs/blob/56adb96a6bbf101af3540102faf27da30758a036/src/service_discovery.rs#L260-L265

fmassot:
It would be super cool to add a test that starts with 2 sources (thus 2 consumers) and then stop 1 source and see the failover in action. Could be added in a follow-up PR.

quickwit/quickwit-indexing/src/source/pulsar_source.rs
ChillFish8: I think we discussed this last week, but ultimately it's better that we do not try and connect these two things, it's better for them to be separate to avoid confusion and in order to be correct.

ChillFish8:
@fmassot @guilload Should be ready for a re-review

quickwit/quickwit-indexing/Cargo.toml
```diff
@@ -25,9 +25,11 @@ libz-sys = { workspace = true, optional = true }
 once_cell = { workspace = true }
 oneshot = { workspace = true }
 openssl = { workspace = true, optional = true }
+pulsar = { workspace = true, optional = true }
 rdkafka = { workspace = true, optional = true }
 rusoto_core = { workspace = true, optional = true }
 rusoto_kinesis = { workspace = true, optional = true }
+reqwest = { workspace = true }
```
fmassot, line 32: Seems to be used in test only.

fmassot (approved):
LGTM.

Can you open issues to not forget what we are currently missing in the tests?

- checking that the ack is working
- checking that starting from a given checkpoint (not position 0) is working 
- checking that the rebalancing is working (starts with 2 partitions, 2 sources, and see what happens if we kill one of the source).

PR #2712, opened 2023-01-25
Update doc sections describing indexing behavior

### LINKED ISSUE #2711: Docs: Sections describing behavior with unknown and missing fields are outdated
**Describe the bug**
The section describing behavior with fields not defined in the config is obsolete and no longer needed considering that this topic is discussed in details in the previous section on the mapping mode. The section describing behavior with missing fields doesn't mention that missing fast fields (except text fast fields) will cause the entire record to be rejected. 

1. Create an index with numeric fast field
2. Try to ingest a document with such field missing.

**Expected behavior**
It seems that rejected record is expected behavior for now, but this is not what documentation says.

**Configuration:**
N/A

imotov:
### Description

This commit removes the section describing behavior with fields not defined in the config since it is obsolete and the new behavior is already covered in details in the previous section on the mapping mode.

It also clarifies the section describing behavior with missing fields, which currently doesn't mention that missing fast fields will cause the entire record to be rejected.

### How was this PR tested?

Manually verified that the newly documented behavior is indeed what happens.

Fixes #2711

### DIFF
--- a/docs/configuration/index-config.md
+++ b/docs/configuration/index-config.md
@@ -454,13 +454,9 @@ In plain language:
 For field names containing the `.` character, you will need to escape it when referencing them. Otherwise the `.` character will be interpreted as a JSON object property access. Because of this, it is recommended to avoid using field names containing the `.` character.
 :::
 
-### Behavior with fields not defined in the config
-
-Fields in your JSON document that are not defined in the `index config` will be ignored.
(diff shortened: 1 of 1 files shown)


guilload (approved):

guilload:
Thanks @imotov!

PR #2725, opened 2023-01-27
Replace wikipedia with stackoverflow in quickstart

### LINKED ISSUE #1618: Use an index with a timestamp field in the quickstart
Following #1614 opened by @mr-karan, it would be nicer to have a quickstart with an index with a timestamp field in the quickstart. Let's do it!

imotov:
### Description

Replaces wikipedia dataset in the quickstart guide with the stackoverflow dataset. This new dataset not only has a timestamp but also contains some interesting elements such as tags and record types to make getting started more fun.

Besides replacing the dataset it also adds instructions on how to use docker image on Apple silicon based macOS systems and adds --rm flag to all docker instructions. Without the rm flag docker creates and keeps a new container for every execution and we start a new container at least 6 times during this tutorial.

Closes #1618

### How was this PR tested?

I visually inspected the markdown. I am not really sure how to render it into HTML the same way as it is done on the web site. 

I manually ran both CLI and docker commands on x86_64 Ubuntu 22.04.1 to make sure that the flow describe in the guide really works. I also tested docker commands on arm64 macOS 13.1 machine (the CLI path fails on this machine due to #1928)

### DIFF
--- a/config/tutorials/stackoverflow/index-config.yaml
+++ b/config/tutorials/stackoverflow/index-config.yaml
@@ -1,0 +1,47 @@
+#
+# Index config file for stackoverflow dataset.
+#
+version: 0.4
+
+index_id: stackoverflow
+
+doc_mapping:
+  field_mappings:
+    - name: user
+      type: text
+      fast: true
+      tokenizer: raw
+    - name: tags
+      type: array<text>
+      fast: true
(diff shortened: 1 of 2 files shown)


fmassot (approved):
I tested the docker commands and check the rendering with docusaurus.

PR #2728, opened 2023-01-27
Add backpressure counter to doc processor

guilload:


quickwit/quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -270,23 +270,23 @@ impl IndexingPipeline {
...
         );
-        let (publisher_mailbox, publisher_handler) = ctx
+        let (publisher_mailbox, publisher_handle) = ctx
             .spawn_actor()
             .set_kill_switch(self.kill_switch.clone())
             .set_backpressure_micros_counter(
                 crate::metrics::INDEXER_METRICS
                     .backpressure_micros
-                    .with_label_values([index_id, "Publisher"]),
+                    .with_label_values([index_id, "publisher"]),
```
guilload, line 279: Prometheus labels are case-sensitive, `publisher` vs. `Publisher` will create two buckets, which have to be merged at query time using lower or uppercasing functions. This makes using different cases in labels error-prone, and the best practice is to stick to lowercase labels.

fmassot (approved):

PR #2717, opened 2023-01-25
CLI 0.4.1

fmassot:
Fix #2663 

## TODO for this PR:

- [x] Add ingest API CLI
- [x] Use splits endpoint with filters endpoint
- [x] Use `mark_splits_for_deletion` endpoint
- [x] Use `reset_source_checkpoint`  endpoint
- [x] Use toggle source  endpoint
- [x] Use clear index endpoint
- [x] use dry run delete index endpoint
- [x] use create index with overwrite  endpoint

## TODO later:

- Add `quickwit-integration` to isolate these tests (next PR before release 0.4.1)
- Add tests on CLI commands `mark_splits_for_deletion`, `toggle-source`, `reset source checkpoint`, `clear-index` (next PR before release 0.4.1)
- Update CLI docs (next PR before release 0.4.1)
- Add tags to splits filter endpoint: remove the tags args for the 0.4.1 and open an issue to add that later, not a high priority
- Display files sizes in describe split command, not a high priority
- the REST client depends on quickwit-server because we use `ListSplitsQueryParams` and `SearchRequestQueryString`. We should either move these struct in `quickwit-common` or having separated struct for the client.
- On a delete index, check that all nodes are doing a cleanup of the index directory.

## CLI preview

```bash
USAGE:
    quickwit [OPTIONS] <SUBCOMMAND>

OPTIONS:
    -h, --help        Print help information
        --no-color    Disable ANSI terminal codes (colors, etc...) being injected into the logging
                      output [env: NO_COLOR=]
    -V, --version     Print version information
    -y, --yes         Assume "yes" as an answer to all prompts and run non-interactively.

SUBCOMMANDS:
    run       Starts Quickwit server with all services by default (`indexer`, `searcher`...).
    index     Manages indexes: creates, deletes, ingests, searches, describes...
    source    Manages sources: creates, updates, deletes sources...
    split     Manages splits: lists, describes, marks for deletion...
    tool      Performs utility operations. Requires a node config.
```

Issue #2491, opened 2022-11-29
Splits rest api returns a JSON Array


fulmicoton:
It would probably be more future-friendly and standard friendly to return a json object (although really i am not sure about that!?)

fmassot:
Indeed, I'm in favor of a JSON object. It is practical for pagination, we could, for example, add the total num hits, the offset, and the limit to ease client usage.

PR #2739, opened 2023-01-29
Add missing REST endpoints.

### LINKED ISSUE #2493: Delete source should return an empty object.

fmassot:
To have a CLI that uses the REST API (required for #2717) , we need to add new endpoints:

- [x] List splits with filtering endpoint.
- [x] Add mark_splits_for_deletion.
- [x] Add source reset_checkpoint endpoint.
- [x] Add toggle source endpoint.
- [x] Add clear index endpoint.
- [x] Adding overwrite for creating the endpoint. I suggest dropping that.
- [x] Adding dry_run to the delete index endpoint. It's minor, and it will be added to this PR.

I also removed the usage of `IndexService` when it was not useful and used the metastore directly to call the right methods.

We need to improve the code of the REST API (organization and harmonization of structs, for example to fix #2493 and #2491). I suggest doing that in another PR.

### DIFF
--- a/quickwit/quickwit-serve/src/index_api/rest_handler.rs
+++ b/quickwit/quickwit-serve/src/index_api/rest_handler.rs
@@ -24,23 +24,32 @@ use hyper::header::CONTENT_TYPE;
 use quickwit_config::{ConfigFormat, QuickwitConfig, SourceConfig};
 use quickwit_core::{IndexService, IndexServiceError};
 use quickwit_janitor::FileEntry;
-use quickwit_metastore::{IndexMetadata, Split};
-use serde::Serialize;
+use quickwit_metastore::{
+    IndexMetadata, ListSplitsQuery, Metastore, MetastoreError, Split, SplitState,
+};
+use quickwit_proto::ServiceError;
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tracing::info;
 use warp::{Filter, Rejection, Reply};
 
+use crate::elastic_search_api::from_simple_list;
 use crate::format::Format;
 use crate::with_arg;
 
 #[derive(utoipa::OpenApi)]
 #[openapi(paths(
-    get_all_splits,
-    get_indexes_metadatas,
     create_index,
+    clear_index,
     delete_index,
+    get_indexes_metadatas,
+    list_splits,
+    mark_splits_for_deletion,
     create_source,
-    delete_source
+    reset_source_checkpoint,
+    toggle_source,
+    delete_source,
 ))]
 pub struct IndexApi;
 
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -47,75 +51,160 @@ pub fn index_management_handlers(
...
-        .then(get_all_splits)
+        .and(serde_qs::warp::query(serde_qs::Config::default()))
+        .and(with_arg(metastore))
+        .then(get_splits)
+        .map(format_response)
+}
+
+#[derive(Deserialize)]
+#[serde(deny_unknown_fields)]
+struct SplitsForDeletion {
```
evanxg852000, line 167: I guess using `SplitsForDeletion` instead of  `Vec<String>` is for consistency right?

quickwit/quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -47,75 +51,160 @@ pub fn index_management_handlers(
...
+        .collect();
+    metastore
+        .mark_splits_for_deletion(&index_id, &split_ids)
+        .await
+}
+
+fn mark_splits_for_deletion_handler(
+    metastore: Arc<dyn Metastore>,
+) -> impl Filter<Extract = (impl warp::Reply,), Error = Rejection> + Clone {
+    warp::path!("indexes" / String / "splits" / "mark-splits-for-deletion")
```
evanxg852000, line 203: Should we switch to DELETE instead of POST?
Also making me think that the endpoint could be named `indexes/{index_id}/slipts/delete` because the `mark-splits-for-deletion` is an internal implementation detail. I would be explicit about this if there were a way to recover (undo) the marked split.

quickwit/quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -183,6 +269,36 @@ async fn create_index(
     index_service.create_index(index_config, false).await
 }
 
+fn clear_index_handler(
+    index_service: Arc<IndexService>,
+) -> impl Filter<Extract = (impl warp::Reply,), Error = Rejection> + Clone {
+    warp::path!("indexes" / String / "clear")
+        .and(warp::post())
```
evanxg852000, line 276: `warp::delete()` ?

evanxg852000 (approved):
LGTM :+1: Just left some minor comments to clarify

quickwit/quickwit-serve/src/index_api/rest_handler.rs
fmassot: Good point; I can shorten it to `mark-for-deletion`. I like to remain explicit, and we only change the state for now (if the GC is not running or has some delays), this will be visible anyway. 
Also, consequently, I would keep the `POST` method instead of `DELETE`.

quickwit/quickwit-serve/src/index_api/rest_handler.rs
fmassot: we don't delete the index, `PUT` is more appropriate I guess.

quickwit/quickwit-serve/src/index_api/rest_handler.rs
fmassot: We can use `PUT`instead of `POST`.

quickwit/quickwit-serve/src/index_api/rest_handler.rs
fmassot: The rationale behind is that I want to have a body that is not directly an array, it's more a convention that anything else but I like to have explicit keys and values and I decode a body into json.

Issue #2754, opened 2023-02-02
Violation in times(...) expectations of metastore mocks might not lead to a test failure.


imotov:
**Describe the bug**
When a mock metastore is passed in Arc to an actor, it is not getting finalized until the actor shutdown. The shutdown operation is performed inside tokio task and as a result panic that mock metastore generates if expectations are violated is only printed on screen but doesn't lead to the test failure.

**Steps to reproduce (if applicable)**

1. Replace the following line https://github.com/quickwit-oss/quickwit/blob/eb989e2df2c841dcac275c07e8877b2569b36175/quickwit/quickwit-indexing/src/actors/publisher.rs#L280 with `.times(10)`
2. Run the test
3. Test prints out the [following message](https://gist.github.com/imotov/5df6e837c1a61a6f5cfab97dd9a21770) but passes.

**Expected behavior**
Test should fail.

**Configuration:**
I ran it on 7494137c54cbe08e73ce73556f1b991e166b01c4

**Possible Solution.**
It seems that explicitly quitting the actor that we passed the metastore into and checking its exit status can helps us to identify this situation. Adding the following line at the end of the test, makes the test to fail if mock expectations fail, as expected:
```
assert!(matches!(publisher_handle.quit().await.0, ActorExitStatus::Quit));
```

guilload:
Something like `Universe.join()` could also be useful.

imotov:
So what is the main intent of this method? Is it just join on all spawned actors or is it more like quit, where we shut them down first? And what should I do with the exit statuses? Return them to the caller as a vector? Check for panics and panic?

imotov:
@guilload I have a draft implementation of it. I think the biggest struggle for me is what to do with ActorHandle's quit, kill and join methods. When I call universe.quit() it quits all actors, but since ActorHandles might still exist somebody can call any of these methods on them even though underlying JoinHandler is now gone. At the moment I am preserving and returning ActorExitStatus that was generated during universe.quit(), but I wonder if I should panic instead if somebody tries to call ActorHandle's method after universe.quit();

imotov:
I opened #2766 to demonstrate one of the approaches more clearly.

imotov:
I think this can be closed since #2766 now ensures that at least in case of universe where most of these mocks are used.

fulmicoton:
Thank you @imotov !

PR #2766, opened 2023-02-08
Catch violations in times(...) expectations

imotov:
### Description

Adds a new quit() method to universe that causes all spawned actors to quit. It also collects actors exit codes that can be used to verify that all actors quit gracefully, which in turn allows to fail the test if mock objects expectations are not met.

Relates to #2754

### How was this PR tested?

Added several tests to universe.rs. I also had to modify several `times(...)` parameters to make tests pass. Somebody who is more familiar with these tests should probably check if these changes make sense or if I inadvertently found and buried actual bugs.


imotov:
I opened #2768 to fix the license header checker.

quickwit/quickwit-indexing/src/actors/indexer.rs
```diff
@@ -878,6 +876,11 @@ mod tests {
                 .delete_opstamp,
             last_delete_opstamp
         );
+        assert!(!universe
+            .quit()
+            .await
+            .into_iter()
+            .any(|s| matches!(s, ActorExitStatus::Panicked)));
```
fulmicoton, line 883: no one letter variable please... It is not that simple to get what this one is about.

quickwit/quickwit-indexing/src/actors/indexer.rs
```diff
@@ -950,6 +949,11 @@ mod tests {
         assert_eq!(output_messages.len(), 1);
         assert_eq!(output_messages[0].commit_trigger, CommitTrigger::NoMoreDocs);
         assert_eq!(output_messages[0].splits[0].split_attrs.num_docs, 1);
+        assert!(!universe
+            .quit()
+            .await
+            .into_iter()
+            .any(|s| matches!(s, ActorExitStatus::Panicked)));
```
fulmicoton, line 956: same as above

quickwit/quickwit-janitor/src/actors/garbage_collector.rs
```diff
@@ -666,5 +691,10 @@ mod tests {
         assert_eq!(counters.num_failed_storage_resolution, 0);
         assert_eq!(counters.num_failed_gc_run_on_index, 0);
         assert_eq!(counters.num_failed_splits, 2);
+        assert!(!universe
+            .quit()
+            .await
+            .into_iter()
+            .any(|s| matches!(s, ActorExitStatus::Panicked)));
```
fulmicoton, line 698: You seem to be copying this snippet in many places...

```
        assert!(!universe
            .quit()
            .await
            .into_iter()
            .any(|s| matches!(s, ActorExitStatus::Panicked)));
```

Do we want it as a method in Universe? Do we want it on Drop?

imotov: I really wish I could add it to Drop, but there is no support for async drop in rust as far as I understand. I experimented with it a bit more by adding a check to Drop that ensure that universe.quit() was really called. That identified a whole bunch of tests where it was missing, so I added it there, but now some ingest tests are hanging from time to time. I want to figure out what's going on before I can push the changes.

quickwit/quickwit-actors/src/actor_handle.rs
```diff
@@ -208,13 +208,10 @@ impl<A: Actor> ActorHandle<A> {
...
-        let exit_status = self.join_handle.await.unwrap_or_else(|join_err| {
-            if join_err.is_panic() {
-                ActorExitStatus::Panicked
-            } else {
-                ActorExitStatus::Killed
-            }
-        });
+        let exit_status =
+            self.join_handle.join().await.expect(
+                "Attempting to call join after universe.join() or universe.quit() was called",
```
fulmicoton, line 213: Two problems here:
- `.join()` can be called by someone else than `universe.join()` or `universe.quit()` so the error message is wrong.
- We don't want to bubble up panics, here. We want panics happening in actors to be intercepted and processed properly.
That's what the code was doing before.

quickwit/quickwit-actors/src/registry.rs
```diff
@@ -64,6 +69,17 @@ impl<A: Actor> JsonObservable for TypedJsonObservable<A> {
         let state: <A as Actor>::ObservableState = oneshot_rx.await.ok()?;
         serde_json::to_value(&state).ok()
     }
+
+    async fn quit(&self) -> Option<ActorExitStatus> {
+        let _ = self.weak_mailbox.upgrade().map(|mailbox| {
+            let _ = mailbox.send_message_with_high_priority(Command::Quit);
+        });
```
fulmicoton, line 76: We tend to avoid using map with side-effects.

```suggestion
        if let Some(mailbox) = self.weak_mailbox.upgrade() {
 let _ = mailbox.send_message_with_high_priority(Command::Quit);
        }
```

quickwit/quickwit-indexing/src/source/ingest_api_source.rs
```diff
@@ -307,6 +307,8 @@ mod tests {
         let doc_batches: Vec<RawDocBatch> = doc_processor_inbox.drain_for_test_typed();
         assert_eq!(doc_batches.len(), 2);
         assert!(doc_batches[1].docs[0].starts_with("038462"));
+        ingest_api_source_handle.quit().await;
```
fulmicoton, line 310: was it necessary to quit ingest_api_source independently?
If justified, a comment would be helpful

quickwit/quickwit-indexing/src/source/ingest_api_source.rs
```diff
@@ -405,6 +408,8 @@ mod tests {
             doc_batches[0].checkpoint_delta.partitions().next().unwrap(),
             &partition_id
         );
+        ingest_api_source_handle.quit().await;
```
fulmicoton, line 411: same as above

quickwit/quickwit-indexing/src/source/ingest_api_source.rs
```diff
@@ -456,6 +461,8 @@ mod tests {
         let doc_batches: Vec<RawDocBatch> = doc_processor_inbox.drain_for_test_typed();
         assert_eq!(doc_batches.len(), 1);
         assert!(doc_batches[0].docs[0].starts_with("000000"));
+        ingest_api_source_handle.quit().await;
```
fulmicoton, line 464: same as above

quickwit/quickwit-indexing/src/test_utils.rs
```diff
@@ -198,6 +198,18 @@ impl TestSandbox {
...
+    #[cfg(any(test, feature = "testsuite"))]
+    pub async fn assert_quit(self) {
+        use quickwit_actors::ActorExitStatus;
+
+        assert!(!self
+            .universe
+            .quit()
+            .await
+            .into_iter()
+            .any(|status| matches!(status, ActorExitStatus::Panicked)));
```
fulmicoton, line 211: can we ever end up with `Panicked` after your changes?

quickwit/quickwit-actors/src/actor_handle.rs
imotov: Not sure I understand the first point. We move self into join, so someone can only call join once. So, the only situation where someone calls the join and doesn't find join handle there and this error is thrown is if universe.join() was called before hand. 

Regarding the second point, if panic had occurred, we have already bubbled it up in universe.join().

quickwit/quickwit-janitor/src/actors/delete_task_pipeline.rs
```diff
@@ -406,6 +406,8 @@ mod tests {
             .find(|split| split.split_state == SplitState::Published)
             .unwrap();
         assert_eq!(published_split.split_metadata.delete_opstamp, 1);
+        universe.assert_quit().await;
```
fulmicoton, line 409: We should probably have an accessor to the Universe in `test_sandbox` and avoid having two universes in the same unit test.

(And eventually rework TestSandbox, I am very unhappy with its current state)

quickwit/quickwit-indexing/src/source/ingest_api_source.rs
imotov: It was deadlocking occasionally in my tests. I still need to figure out where the deadlock was occurring but, indeed, I should've added a comment.

quickwit/quickwit-janitor/src/actors/garbage_collector.rs
fulmicoton: Good point. The `assert_quit` solution is good.

quickwit/quickwit-actors/src/actor_handle.rs
fulmicoton: I think I was missing the fact that you swapped the tokio taask JoinHandle by your a new struct. Let me dig into that further.

fulmicoton: (you can hit the mark as resolved button, i am not doing to maximize the chance you read this :) )

imotov: Yeah, it is a bit tricky structure because join handle consumes self and we need to share it between two places - actor handle and universe but can only be joined once.

quickwit/quickwit-actors/src/registry.rs
```diff
@@ -181,6 +223,39 @@ fn get_iter<A: Actor>(
...
+                    ActorExitStatus::Panicked
+                } else {
+                    ActorExitStatus::Killed
+                }
+            });
+            Some(exit_status)
+        } else {
+            None
+        }
+    }
```
fulmicoton, line 256: nitpick...

Whenever you hold a lock, you want to make it trivial for the reader that what you are doing is simple.

Make the code running while holding the lock as short and as simple as possible. Ideally not run any io or call external code.

Here I suggest we do it as follows

```suggestion
        // Lock is held only on the following line.
        let mut join_handle_opt = self.holder.lock().await.take();
        if let Some(join_handle) = join_handle {
            let exit_status = join_handle.await.unwrap_or_else(|join_err| {
                if join_err.is_panic() {
                    ActorExitStatus::Panicked
                } else {
                    ActorExitStatus::Killed
                }
            });
            Some(exit_status)
        } else {
            None
        }
    }
```

fulmicoton: Or maybe this is a bad idea actually...

With the following solution, a second person calling `.join()` concurrently will return `None` instantly which might not be what you wanted.

fulmicoton: Maybe we can do something cleaner using https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.shared ?

fulmicoton (changes requested):
LGTM... 

See comments inline.
Can you investigate the `Shared<JoinHandle>` solution too?
If it is a dead-end no pb.

quickwit/quickwit-actors/src/registry.rs
imotov: That's a decision leftover from one of the iterations where I was storing enum with an exit code here instead of Option and returning it to the second caller instead of panicking. In that solution it was important to wait for join to finish. I am not sure if it is important in the current setup. It "feels" like it is safer to block the second caller until join actually happens. But I don't really have any solid arguments at the moment for either approach.

fulmicoton: I suspected something like that. Can you have a look at the Shared<JoinHandle> solution?
It will simplify the code a little bit, and return the exit code in subsequent calls.

imotov:
One thing that still annoys me with the current implementation is that when a test that uses universe fails for some unrelated reason the logic in universe drop is still triggered. So, you see two panics - first is your test's panic, which is useful, but then you also get the universe's drop panic with irrelevant message "Did you call universe.assert_quit()?". I just implemented it and it confuses me sometimes, I imagine it might also confuse others. Not really sure how to detect it though.

fulmicoton:
@imotov Yes, if it is too confusing we may have to remove this Drop impl. Anyway, LGTM

fulmicoton (approved):

fulmicoton (approved):

PR #2782, opened 2023-02-09
list terms without encoding them as json

trinity-1686a:
### Description

don't encode terms in ListTermsQuery as json string: allows supporting terms of any kind

### How was this PR tested?

ingesting some traces with tracegen, and verifying service names are fine on jaeger (+ unit tests)


quickwit/quickwit-search/src/lib.rs
```diff
@@ -266,3 +266,19 @@ pub async fn start_searcher_service(
     ));
     Ok(search_service)
 }
+
+#[doc(hidden)]
```
guilload, line 270: Can we put this macro behind a feature flag? `#[cfg(any(test, feature = "testsuite"))]`

guilload (approved):

PR #2783, opened 2023-02-09
Switch to custom collector for aggregating trace IDs

guilload:
### Description
Switch to a custom collector for aggregating trace IDs. This PR focuses on the plumbing part. I will follow up shortly with a second PR that implements an efficient algorithm.

### How was this PR tested?
- Added unit tests
- Tested with Jaeger
- Ran `make test-all`

### DIFF
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -24,6 +24,7 @@ use std::sync::Arc;
 use itertools::Itertools;
 use quickwit_doc_mapper::{DocMapper, WarmupInfo};
 use quickwit_proto::{LeafSearchResponse, PartialHit, SearchRequest, SortOrder};
+use serde::Deserialize;
 use tantivy::aggregation::agg_req::{
     get_fast_field_names, get_term_dict_field_names, Aggregations,
 };
@@ -35,6 +36,7 @@ use tantivy::schema::Schema;
 use tantivy::{DocId, Score, SegmentOrdinal, SegmentReader};
 
 use crate::filters::{TimestampFilter, TimestampFilterBuilder};
+use crate::jaeger_collector::{FindTraceIdsCollector, FindTraceIdsSegmentCollector};
 use crate::partial_hit_sorting_key;
 
 #[derive(Clone, Debug)]
@@ -161,6 +163,11 @@ impl PartialEq for PartialHitHeapItem {
 
 impl Eq for PartialHitHeapItem {}
 
+enum AggregationSegmentCollectors {
+    FindTraceIdsSegmentCollector(FindTraceIdsSegmentCollector),
+    TantivyAggregationSegmentCollector(AggregationSegmentCollector),
+}
+
 /// Quickwit collector working at the scale of the segment.
 pub struct QuickwitSegmentCollector {
     num_hits: u64,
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-search/src/jaeger_collector.rs
```diff
@@ -0,0 +1,226 @@
...
+/// same trace can be found in the document set. As a result, this problem is akin to finding the
+/// top k elements with duplicates
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct FindTraceIdsCollector {
+    num_traces: usize,
+    trace_id_field_name: String,
+    span_timestamp_field_name: String,
+}
+
+impl FindTraceIdsCollector {
```
guilload, line 68: This PR implements a naive approach and focuses on plugging in the custom collector. I'll shortly follow up with a PR dedicated to improving the collector's performance.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -170,7 +177,7 @@ pub struct QuickwitSegmentCollector {
     max_hits: usize,
     segment_ord: u32,
     timestamp_filter_opt: Option<TimestampFilter>,
-    aggregation: Option<AggregationSegmentCollector>,
+    aggregation: Option<AggregationSegmentCollectors>,
```
fulmicoton, line 180: Instead of the enum approach, have you considered adding a field:
```
aggregation: Option<AggregationSegmentCollector>,
find_trace_ids: Option<FindTraceids>,
```


```
if let Some() = self.aggregation.as_mut() {
   // ...
}
if let Some() = self.find_trace_ids.as_mut() {
   // ...
}
```

It would have the benefit of not mixing the results.

quickwit/quickwit-search/src/jaeger_collector.rs
```diff
@@ -0,0 +1,215 @@
...
+
+impl FindTraceIdsSegmentCollector {
+    fn trace_id_term_ord(&self, doc: DocId) -> TraceIdTermOrd {
+        self.trace_id_ff_reader
+            .get_first_val(doc)
+            .expect("There should be exactly one trace ID per span.")
+    }
+
+    fn trace_id(&self, trace_id_term_ord: TraceIdTermOrd) -> TraceId {
+        let mut trace_id = vec![0u8; 16];
```
fulmicoton, line 152: You probably wanted to do 

```rust
        let mut trace_id = Vec::with_capacity(16);
```

but I wonder if this is the right idea here?
If you are certain that the trace id is 16 bytes long, then the best is probably to work off the stack 
and remove allocations.

fulmicoton (approved):

quickwit/quickwit-search/src/collector.rs
guilload: I did not. When I first started, I was hoping that tantivy aggregations also implement `Collector/SegmentCollect`. Unfortunately, I ended up in this ugly situation when I realized too late that it was not the case. 

Since we're about to introduce a new query AST that also carries collectors, I will leave it as is for now.

quickwit/quickwit-search/src/jaeger_collector.rs
guilload: > You probably wanted to do

Courtesy of Copilot :D

> but I wonder if this is the right idea here?

Good point. Will fix.

PR #2792, opened 2023-02-11
Each indexing tasks group (index ID, source ID) is stored in a dedica…

fmassot:
…ted key in chitchat state.

### DIFF
--- a/quickwit/quickwit-cluster/Cargo.toml
+++ b/quickwit/quickwit-cluster/Cargo.toml
@@ -14,7 +14,6 @@ anyhow = { workspace = true }
 async-trait = { workspace = true }
 chitchat = { workspace = true }
 itertools = { workspace = true }
-rand = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
@@ -33,4 +32,5 @@ testsuite = []
 [dev-dependencies]
 quickwit-common = { workspace = true, features = ["testsuite"] }
 
+rand = { workspace = true }
 tempfile = { workspace = true }
--- a/quickwit/quickwit-cluster/Cargo.toml
+++ b/quickwit/quickwit-cluster/Cargo.toml
@@ -30,4 +30,6 @@ quickwit-proto = { workspace = true }
 testsuite = []
 
 [dev-dependencies]
+quickwit-common = { workspace = true, features = ["testsuite"] }
+
 tempfile = { workspace = true }
--- a/quickwit/quickwit-cluster/Cargo.toml
+++ b/quickwit/quickwit-cluster/Cargo.toml
@@ -14,6 +14,7 @@ anyhow = { workspace = true }
 async-trait = { workspace = true }
 chitchat = { workspace = true }
 itertools = { workspace = true }
+rand = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
--- a/quickwit/quickwit-cluster/src/cluster.rs
+++ b/quickwit/quickwit-cluster/src/cluster.rs
@@ -626,7 +626,7 @@ mod tests {
             .unwrap(),
         );
         let mut random_generator = rand::thread_rng();
-        let indexing_tasks = (0..100_000)
+        let indexing_tasks = (0..10_000)
             .map(|_| {
                 let index_id = random_generator.gen_range(0..=10_000);
                 let source_id = random_generator.gen_range(0..=100);
--- a/quickwit/quickwit-cluster/src/cluster.rs
+++ b/quickwit/quickwit-cluster/src/cluster.rs
@@ -677,7 +677,7 @@ mod tests {
             .unwrap();
         }
 
-        // Readd tasks.
+        // Re-add tasks.
         cluster1
             .update_self_node_indexing_tasks(&indexing_tasks)
             .await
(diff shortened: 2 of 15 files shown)


fulmicoton:
@guilload can you review?

quickwit/quickwit-cluster/src/member.rs
```diff
@@ -20,18 +20,21 @@
...
 use crate::QuickwitService;
 
 // Keys used to store member's data in chitchat state.
 pub(crate) const GRPC_ADVERTISE_ADDR_KEY: &str = "grpc_advertise_addr";
 pub(crate) const ENABLED_SERVICES_KEY: &str = "enabled_services";
-pub(crate) const RUNNING_INDEXING_PLAN: &str = "running_indexing_plan";
+// An indexing task key is formatted as
+// `{INDEXING_TASK_PREFIX}{INDEXING_TASK_SEPARATOR}{index_id}{INDEXING_TASK_SEPARATOR}{source_id}`.
+pub(crate) const INDEXING_TASK_PREFIX: &str = "indexing_task";
+pub(crate) const INDEXING_TASK_SEPARATOR: &str = ":";
```
fmassot, line 37: I put ":" at first but we may relax again the field name validation rules... So we may want to choose something else.

fmassot:
@guilload, it's ready for review. One test is failing, I just need to avoid setting 100_000 tasks in chitchat state.

quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -559,11 +594,160 @@ mod tests {
...
+                vec![cluster1.gossip_listen_addr.to_string()],
+                &["indexer", "metastore"],
+                &transport,
+                true,
+            )
+            .await
+            .unwrap(),
+        );
+        let mut random_generator = rand::thread_rng();
+        let indexing_tasks = (0..100_000)
```
fmassot, line 629: The test is failing here, I will either reduce the number of tasks or increase the timeout.

quickwit/quickwit-cluster/Cargo.toml
```diff
@@ -14,6 +14,7 @@ anyhow = { workspace = true }
 async-trait = { workspace = true }
 chitchat = { workspace = true }
 itertools = { workspace = true }
+rand = { workspace = true }
```
guilload, line 17: Seems like a dev dependency.

quickwit/quickwit-cluster/src/member.rs
```diff
@@ -20,18 +20,21 @@
...
 use crate::QuickwitService;
 
 // Keys used to store member's data in chitchat state.
 pub(crate) const GRPC_ADVERTISE_ADDR_KEY: &str = "grpc_advertise_addr";
 pub(crate) const ENABLED_SERVICES_KEY: &str = "enabled_services";
-pub(crate) const RUNNING_INDEXING_PLAN: &str = "running_indexing_plan";
+// An indexing task key is formatted as
+// `{INDEXING_TASK_PREFIX}{INDEXING_TASK_SEPARATOR}{index_id}{INDEXING_TASK_SEPARATOR}{source_id}`.
+pub(crate) const INDEXING_TASK_PREFIX: &str = "indexing_task";
+pub(crate) const INDEXING_TASK_SEPARATOR: char = '\u{0000}';
```
guilload, line 37: Let's use a more obvious separator since many characters are not allowed in an index or source ID.

quickwit/quickwit-cluster/src/member.rs
```diff
@@ -167,14 +158,61 @@ pub(crate) fn build_cluster_member(
         enabled_services,
         chitchat_node.gossip_public_address,
         grpc_advertise_addr,
-        running_indexing_plan,
+        indexing_tasks,
     ))
 }
 
+/// Parses indexing tasks serialized in keys formatted as `INDEXING_TASK_PREFIX:index_id:source_id`.
+/// Malformatted keys and values are ignored, just warnings are emitted.
```
guilload, line 166:
```suggestion
/// Malformed keys and values are ignored, just warnings are emitted.
```

quickwit/quickwit-cluster/src/member.rs
```diff
@@ -167,14 +158,61 @@ pub(crate) fn build_cluster_member(
...
+pub(crate) fn parse_indexing_tasks(node_state: &NodeState, node_id: &str) -> Vec<IndexingTask> {
+    node_state
+        .iter_key_values(|key, _| key.starts_with(INDEXING_TASK_PREFIX))
+        .map(|(key, versioned_value)| {
+            let (_prefix, key_without_prefix) =
+                key.split_once(INDEXING_TASK_SEPARATOR).ok_or_else(|| {
```
guilload, line 172: There is zero chance this happens, I'd go an `expect`, which will simplify the rest of the code. There will be an issue only if we change the separator.

guilload (approved):

quickwit/quickwit-cluster/src/member.rs
fulmicoton: Maybe I am misunderstanding something, but the guarantee that this is correct is very loose. We should not panic on that.

fmassot: I would avoid the expect... I'm always suspicious of what we can have in chitchat state....

Let me try a small refactoring to make the code looks nicer.

quickwit/quickwit-cluster/src/member.rs
fmassot: I was clearly hesitating with a character like `:` and something obscure... I was a bit worried that we may authorize more characters for index IDs. 
Does `:` work for you?

PR #2810, opened 2023-02-13
Add cooperative indexing

guilload:
### Description
- Limit the number of concurrent active workbenches to `num_cpus`
- Indexing pipelines yield when their mailbox is drained and sleep until the next "indexing interval"

### How was this PR tested?
- [ ] added unit test
- [x] ran 100+ concurrent indexing pipelines on my machine


quickwit/quickwit-indexing/src/actors/indexer.rs
```diff
@@ -49,6 +52,9 @@ use crate::models::{
     NewPublishLock, PreparedDoc, PreparedDocBatch, PublishLock, ScratchDirectory,
 };
 
+/// Limits the number of concurrent active workbenches to `num_cpus`.
+static INDEXING_PERMITS: Lazy<Semaphore> = Lazy::new(|| Semaphore::new(num_cpus::get()));
```
fulmicoton, line 56: num_cpus is not great here. We want the same logic as what is done in the `Runtimes`.

fulmicoton: you can find that in RuntimesType

fulmicoton (approved):

fulmicoton:
Yes we can merge!

fulmicoton (approved):

fulmicoton (approved):

fulmicoton (approved):

PR #2812, opened 2023-02-13
Generate service trait from proto service definition

guilload:
### Description
Per title.

### How was this PR tested?
- Inspected generated code
- Added unit test

### DIFF
--- a/quickwit/Cargo.lock
+++ b/quickwit/Cargo.lock
@@ -4466,6 +4466,27 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


fulmicoton (approved):
A README would be useful.

fulmicoton (commented):
I think you might need to change something in the license check configuration file?

guilload:
> A README would be useful.

Will do!

> I think you might need to change something in the license check configuration file?

Yes, for some reason, this is not causing an issue in this PR, but it will in other instances.

PR #2818, opened 2023-02-15
Fix attempt to `try_send` before `send`

guilload:
### Description
The code path was different depending on whether a backpressure counter was set or not. However, I believe we always want to `try_send` before `send`:

Before:
```rust
if let Some(backpressure_micros_counter) = backpressure_micros_counter_opt {
    match self.inner.tx.try_send_low_priority(envelope) {
        Ok(()) => Ok(response_rx),
        Err(TrySendError::Full(msg)) => {
            let now = Instant::now();
            self.inner.tx.send_low_priority(msg).await?;
            let elapsed = now.elapsed();
            backpressure_micros_counter.inc_by(elapsed.as_micros() as u64);
            Ok(response_rx)
        }
        Err(TrySendError::Disconnected) => Err(SendError::Disconnected),
    }
} else {
    self.inner.tx.send_low_priority(envelope).await?;
    Ok(response_rx)
}
```

After:
```rust
match self.inner.tx.try_send_low_priority(envelope) {
    Ok(()) => Ok(response_rx),
    Err(TrySendError::Full(envelope)) => {
        if let Some(backpressure_micros_counter) = backpressure_micros_counter_opt {
            let now = Instant::now();
            self.inner.tx.send_low_priority(envelope).await?;
            let elapsed = now.elapsed();
            backpressure_micros_counter.inc_by(elapsed.as_micros() as u64);
        } else {
            self.inner.tx.send_low_priority(envelope).await?;
        }
        Ok(response_rx)
    }
    Err(TrySendError::Disconnected) => Err(SendError::Disconnected),
}
```

### How was this PR tested?
- Ran test suite

### DIFF
--- a/quickwit/quickwit-actors/src/mailbox.rs
+++ b/quickwit/quickwit-actors/src/mailbox.rs
@@ -235,21 +235,20 @@ impl<A: Actor> Mailbox<A> {
         M: 'static + Send + Sync + fmt::Debug,
     {
         let (envelope, response_rx) = self.wrap_in_envelope(message);
-        if let Some(backpressure_micros_counter) = backpressure_micros_counter_opt {
-            match self.inner.tx.try_send_low_priority(envelope) {
-                Ok(()) => Ok(response_rx),
-                Err(TrySendError::Full(msg)) => {
+        match self.inner.tx.try_send_low_priority(envelope) {
+            Ok(()) => Ok(response_rx),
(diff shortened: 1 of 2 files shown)


fulmicoton:
> However, I believe we always want to try_send before send

Can you detail?

fulmicoton (approved):

PR #2826, opened 2023-02-16
Refactor Indexer to talk directly to Publisher

### LINKED ISSUE #2825: If all records fail suggest_truncate is never sent to the source
**Describe the bug**
In my testing for #2699 implementation I noticed that when all documents fail suggest_truncate is never sent to the SourceActor, as a result queue is never notified that it can clean older records. In my experience this situation happens more often than one could expect. For example, a new version of software is installed and it starts generating logs in a slightly different format or mappings are fat-fingered making all login messages suddenly fail. Without proper monitoring this situation can remain unnoticed for quite a while, sometimes days until users wonder why they cannot see new log messages in the system.

1. Observe SourceActor's suggest_truncate behavior while indexing all failed messages.

**Expected behavior**
The  suggest_truncate should be sent periodically even if all documents happened to fail.

imotov:
### Description

Ensures that publisher is still notified even if all documents fail.

### How was this PR tested?

Added unit tests, updated existing unit tests.

Fixes #2825


imotov:
Spoke to @guilload about this, sounds like it needs some rework. So switching it to WIP for now.

imotov:
@guilload could you see if this matches the approach that we have discussed?

quickwit/quickwit-indexing/src/actors/indexing_pipeline.rs
guilload, line 1: Weirdly, I don't see where we no longer passe a metastore instance when we call `Indexer::new` in this file.

quickwit/quickwit-indexing/src/actors/packager.rs
```diff
@@ -164,6 +165,26 @@ impl Handler<IndexedSplitBatch> for Packager {
     }
 }
 
+#[async_trait]
+impl Handler<EmptySplit> for Packager {
+    type Reply = ();
+
+    #[instrument(
+        name="packager_empty_batch"
```
guilload, line 173:
```suggestion
        name="package_empty_batch"
```

guilload (approved):
LGTM. Thanks for the very clean implementation and testing your change extensively.

quickwit/quickwit-indexing/src/actors/indexing_pipeline.rs
imotov: It is still needed in 
https://github.com/quickwit-oss/quickwit/blob/6664abad06908416336804dd9ab2b6bc6222075f/quickwit/quickwit-indexing/src/actors/indexer.rs#L156-L165

guilload: Thanks.

PR #2827, opened 2023-02-16
Add codegen for actor mailbox

guilload:
### Description
Generate the boiler code necessary to use an actor's mailbox as a Tower service. That means that we can implement an endpoint as `fn hello<T>(message: HelloRequest) where T: tower::Service<HelloRequest>` and provide whichever Tower service implementation.

Under the hood, the call can be routed to a local implementation (actor or not) or a remote implementation via gRPC.

### How was this PR tested?
- Updated unit test

### DIFF
--- a/quickwit/quickwit-codegen-example/src/hello.rs
+++ b/quickwit/quickwit-codegen-example/src/hello.rs
@@ -61,6 +61,67 @@ impl tower::Service<HelloRequest> for HelloClient {
         Box::pin(fut)
     }
 }
+#[derive(Debug)]
+struct MailboxAdapter<A: quickwit_actors::Actor, E> {
+    inner: quickwit_actors::Mailbox<A>,
+    phantom: std::marker::PhantomData<E>,
+}
+impl<A, E> std::ops::Deref for MailboxAdapter<A, E>
+where A: quickwit_actors::Actor
+{
+    type Target = quickwit_actors::Mailbox<A>;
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-codegen-example/src/hello.rs
```diff
@@ -61,6 +68,60 @@ impl tower::Service<HelloRequest> for HelloClient {
...
+    crate::HelloError: From<quickwit_actors::AskError<E>>,
+{
+    type Response = T;
+    type Error = crate::HelloError;
+    type Future = BoxFuture<Self::Response, Self::Error>;
+    fn poll_ready(
+        &mut self,
+        _cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        //! This no longer works with parallel actors. The fix is to acquire a permit from the mailbox in `poll_ready` and consumer it in `call`.
```
fmassot, line 114: I don't understand the "parallel actors" reference. Are you referring to a specific implementation (possibly on an old branch)?

fmassot (approved):

quickwit/quickwit-codegen-example/src/hello.rs
guilload: I updated the comment. I was trying to say that this approach does not work with "balance" middleware such as `tower::balance::pool::Pool`.

PR #2831, opened 2023-02-16
Remove ingest API GC and let the indexing service handle the GC.

### LINKED ISSUE #2605: Review the ingest api garbage collector once the control plane is working
Currently, the GC deletes the queue of deleted indexes.

To do that, the actor fetches all index metadatas every 30 seconds.

First, it does not consider that an ingest API source can be disabled.
Secondly, once the Control Plane is working, it will assign indexing tasks to indexers, so we will need to update this garbage collector or even replace it with a garbage collector that will clean ingest queues but also take care of other sources cleanup, for example, we may want to clean the split caching directory as having many split files can slow down the startup time of a node.

fmassot:
Fixes #2605

I refactored the indexing service tests which were a bit too long.

We may want to remove the `ShutdownPipelines` and `ShutdownPipelines` messages as they are not used anymore. I kept them for now as it can help a bit in tests.

Additional notes (2023/02/17):

- To shutdown pipeline actors, we call now `kill`.
- I want to remove `ShutdownPipelines` and `ShutdownPipelines`, @guilload do you feel there is still a need for them?
- Discussed with @guilload, we may want to distinguish a master service (`Quicklet`?) running on top of the indexing service and ingest API service to handle incoming requests from the control plane or admin users. My feeling right now is that it's a bit too early to do that, I'm unsure if it will fit well the next developments as we have not yet defined them.


quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -508,7 +517,8 @@ impl IndexingService {
         for pipeline_id_to_remove in running_pipeline_ids.difference(&updated_pipeline_ids) {
             match self.detach_pipeline(pipeline_id_to_remove).await {
                 Ok(pipeline_handle) => {
-                    pipeline_handle.quit().await;
+                    // Killing the pipeline ensure that all pipeline actors will stop.
+                    pipeline_handle.kill().await;
```
fmassot, line 521: Use `kill`.

fmassot:
PTAL @guilload

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -558,6 +583,52 @@ impl IndexingService {
             );
         }
     }
+
+    /// Garbage collects ingest API queues of deleted indexes.
+    async fn run_ingest_api_queues_gc(&mut self) -> anyhow::Result<()> {
+        let ingest_api_service =
```
guilload, line 589:
```rust
        let Some(ingest_api_service) = &self.ingest_api_service else {
            return Ok(());
        }
```

quickwit/quickwit-indexing/src/actors/indexing_service.rs
guilload: Do we know it's safe to kill the pipeline? I worry about ongoing merges.

guilload:
Let's remove the `ShutdownPipeline` messages if they are no longer used.

guilload (approved):

quickwit/quickwit-indexing/src/actors/indexing_service.rs
fmassot: nice indeed

quickwit/quickwit-indexing/src/actors/indexing_service.rs
fmassot: Actually not sure what to test. The indexing pipeline refers to the merge planner mailbox... so killing the pipeline will not affect the merge planner.

PR #2832, opened 2023-02-17
Optimize `FindTraceIdsCollector`

guilload:
### Description
Per title.

### How was this PR tested?
- [x] add unit tests
- [x] fix broken tests
- [x] test within Jaeger UI


quickwit/quickwit-search/src/find_trace_ids_collector.rs
```diff
@@ -0,0 +1,508 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(into = "TraceIdFruitForSerde")]
+#[serde(try_from = "TraceIdFruitForSerde")]
+pub struct TraceIdFruit {
+    pub trace_id: B64TraceId,
+    pub span_timestamp: DateTime,
+}
+
+impl TraceIdFruit {
+    pub fn new(trace_id_bytes: &[u8], span_timestamp: DateTime) -> Self {
```
fulmicoton, line 47: There is a hidden contract here. Trace_id_bytes needs to have a specific len otherwise it panics, and this struct and method are pub.
Ideally we should get rid of the hidden contract.

fulmicoton: Or at least, we  want to make that private.

fulmicoton: The size of the bytes is the result of the index containing the right information. This is too loose. I think we need to return an error if trace_id_bytes does not have the right size.

quickwit/quickwit-search/src/find_trace_ids_collector.rs
```diff
@@ -0,0 +1,508 @@
...
+        let mut trace_id_bytes = Vec::with_capacity(24);
+
+        self.select_trace_ids
+            .harvest()
+            .into_iter()
+            .map(|trace_id| {
+                let found = self.inverted_index_reader
+                    .terms()
+                    .ord_to_term(trace_id.term_ord, &mut trace_id_bytes)
+                    .expect("The term dict should be available.");
```
fulmicoton, line 292: The expect message is wrong...

ord_to_term returns an error if the term ordinal does not exist.

quickwit/quickwit-search/src/find_trace_ids_collector.rs
```diff
@@ -0,0 +1,508 @@
...
+impl Hash for TraceIdFruit {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.trace_id.hash(state);
+    }
+}
+
+impl Ord for TraceIdFruit {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.span_timestamp
+            .cmp(&other.span_timestamp)
```
fulmicoton, line 103: I'm not super comfortable with having Ord inconsistent with Eq.
Is it necessary?

quickwit/quickwit-search/src/find_trace_ids_collector.rs
```diff
@@ -0,0 +1,508 @@
...
+        &self,
+        segment_fruits: Vec<<Self::Child as SegmentCollector>::Fruit>,
+    ) -> tantivy::Result<Self::Fruit> {
+        let mut trace_ids = Vec::with_capacity(self.num_traces);
+        let mut seen_trace_ids = FnvHashSet::default();
+
+        for trace_id in segment_fruits
+            .into_iter()
+            .map(BinaryHeap::from)
+            .map(binary_heap::IntoIterSorted::new)
```
fulmicoton, line 234: Why do we need these two lines?

It would be less complicated and faster to just sort the segment fruits, and apply the kmerge on that.

quickwit/quickwit-search/src/find_trace_ids_collector.rs
```diff
@@ -0,0 +1,508 @@
...
+            ),
+            select_workbench: Vec::with_capacity(2 * num_traces),
+            running_term_ord: None,
+            running_span_timestamp: DateTime::default(),
+            span_timestamp_sentinel: DateTime::from_timestamp_micros(i64::MIN),
+        }
+    }
+
+    fn collect(&mut self, term_ord: TermOrd, span_timestamp: DateTime) {
+        if self.running_term_ord.is_none() {
```
fulmicoton, line 325: nitpick...

I suspect the code could be midly cleaner with an Option<RunningTerm>

```
struct RunningTerm {
   term_ord,
   max_timestamp,
} 
```

You could then use `running_term_opt.get_or_insert()`

quickwit/quickwit-search/src/find_trace_ids_collector.rs
```diff
@@ -0,0 +1,508 @@
...
+        }
+        self.select_workbench.clear();
+
+        for (term_ord, span_timestamp) in self.dedup_workbench.drain() {
+            let trace_id = TraceIdTermOrd::new(term_ord, span_timestamp);
+            self.select_workbench.push(trace_id);
+        }
+        let select_len = self.num_traces.min(self.select_workbench.len());
+        let select_index = select_len - 1;
+        self.select_workbench.select_nth_unstable(select_index);
```
fulmicoton, line 376: TIL that it uses a version of quickselect/quicksort that does not exhibit the worst case nightmares of quicksort.

quickwit/quickwit-search/src/find_trace_ids_collector.rs
```diff
@@ -0,0 +1,508 @@
...
+    num_traces: usize,
+    dedup_workbench: FnvHashMap<TermOrd, DateTime>,
+    select_workbench: Vec<TraceIdTermOrd>,
+    running_term_ord: Option<TermOrd>,
+    running_span_timestamp: DateTime,
+    span_timestamp_sentinel: DateTime,
+}
+
+impl SelectTraceIds {
+    fn new(num_traces: usize) -> Self {
```
fulmicoton, line 310: I think some of your code breaks if num_traces == 0.
The `select_index` underflows for instance. Do we want to make a NonZeroUsize?

fulmicoton (changes requested):
a couple of comments.

quickwit/quickwit-search/src/find_trace_ids_collector.rs
guilload: Changed to `&[u8; 24]`.

quickwit/quickwit-search/src/find_trace_ids_collector.rs
guilload: What's the difference between the function returning an error and `Ok(false)`? How do we know the error is an actual io error vs. the term ord does not exist?

quickwit/Cargo.lock
```diff
@@ -2253,6 +2226,12 @@ dependencies = [
  "libc",
 ]
 
+[[package]]
+name = "hermit-abi"
```
fulmicoton, line 2230: do you know what ended up adding this dep?
if possible it would be nice to disable it

quickwit/quickwit-indexing/src/test_utils.rs
```diff
@@ -69,6 +69,7 @@ impl TestSandbox {
         indexing_settings_yaml: &str,
         search_fields: &[&str],
     ) -> anyhow::Result<Self> {
+        let universe = Universe::new();
```
fulmicoton, line 72: you add trouble iwth accelerated time too?

quickwit/quickwit-jaeger/src/lib.rs
```diff
@@ -63,8 +64,7 @@ mod metrics;
 // OpenTelemetry to Jaeger Transformation
 // <https://opentelemetry.io/docs/reference/specification/trace/sdk_exporters/jaeger/>
 
-/// A base64-encoded 16-byte array.
-type TraceId = String;
+type B64TraceId = [u8; 24];
```
fulmicoton, line 67: We have two types called B64TraceId.

One is a String the other is `[u8; 24]`;
It is a bit confusing. Can we rename one?

quickwit/quickwit-jaeger/src/lib.rs
```diff
@@ -294,8 +297,10 @@ impl JaegerService {
                 query.push_str(" OR ");
             }
             query.push_str("trace_id:");
-            query.push_str(trace_id);
+            query.push_str(std::str::from_utf8(trace_id).expect("Base64 should be valid UTF-8."));
```
fulmicoton, line 300: I don't see the justification for trace_id to be valid utf-8.
Is the `trace_id` something that is passed in a request?

Maybe we could replace the type def by a `struct Base64TraceId([u8; 24])`;
and check for the str validity  in the constructors.

quickwit/Cargo.lock
fulmicoton: Ok I checked. This is architecture specific, and we don't pull it.

fulmicoton (approved):

PR #2836, opened 2023-02-17
Make test_merge_executor_controlled_directory_kill_switch more deterministic

### LINKED ISSUE #2835: test_merge_executor_controlled_directory_kill_switch fails occasionally
**Describe the bug**
This test fails 4 times out of 5 on my machines. 

1. I can easily reproduce it by running `make test-failpoints` a few times.

**Expected behavior**
Test passes.

**Root cause analysis:**
On my machine sleep in the following line doesn't seem to be long enough for the actor to reach the pause point:
https://github.com/quickwit-oss/quickwit/blob/51440cf0344af4b7267d6632756f583b8699a636/quickwit/quickwit-indexing/failpoints/mod.rs#L321
The universe is killed before actor gets the message and as a result returns status code Killed instead of expected Failure.

imotov:
### Description

Removes sleep from test_merge_executor_controlled_directory_kill_switch and makes it more deterministic. Instead of realying on the sleep to allow the test to reach the "pause" point this change replaces it with two barriers to guarantee it.

### How was this PR tested?

It is fixing the test.

Closes #2835


quickwit/quickwit-indexing/failpoints/mod.rs
```diff
@@ -307,20 +307,27 @@ async fn test_merge_executor_controlled_directory_kill_switch() -> anyhow::Resul
...
+    let before_universe_kill_clone = before_universe_kill.clone();
+    let after_universe_kill_clone = after_universe_kill.clone();
+    fail::cfg_callback("before-merge-split", move || {
+        before_universe_kill_clone.wait();
+        after_universe_kill_clone.wait();
+    })
+    .unwrap();
     merge_executor_mailbox.send_message(merge_scratch).await?;
-
-    tokio::time::sleep(Duration::from_millis(10)).await;
```
fulmicoton, line 321: Alternatively you can put an arbitrary amount of time here. We work with simulated time.

imotov: I couldn't quite figure out how to make this sleep simulated. Even if I replace `tokio::time::sleep()` with `universe.sleep()` this sleep is still real. The envelope of the merge_scratch message created on the line above is not getting dropped until `merge_executor_handle.join`, which happens after sleep and while this envelope is alive `no_advance_time_guard` in it prevents the simulated sleep from happening. So, while putting 500ms here does indeed help these are real 500ms.

fulmicoton: ah right because we pause within a handler. that makes sense

fulmicoton (approved):

PR #2840, opened 2023-02-18
Improve scheduling triggers. Fix #2659.

### LINKED ISSUE #2659: Avoid triggering a scheduling each time a running plan differs from the last applied plan
Copy-pasting context from [here](https://github.com/quickwit-oss/quickwit/pull/2633#discussion_r1071945813):

> Triggering a new schedule plan seems a little bit harsh here.
Can you open a ticket to revisit this logic?
>
> I suspcet it could easily end up in a crazy loop if a single node is faulty, rendering the entire cluster unusablle.
For instance, imagine a node that is missing hard disk.. or a node that cannot connect to kafka. etc.
>
> In fact it is probably worth thinking whetehr an errorr is ndoe speicfic (the node cannot run a task), or source specific (won't work on another node).... Try to infer it... and act differently depending on that.
>
> We could:
> a) Send an RPC to try to enforce the plan (also I agree this is strange considering we pass the plan through gRPC already). Afterall. maybe chitchat is just lagging. The node should then tell us if
it fails to start the pipeline in case we should do b). If we get a success, then we wait.
> b) If the node cannot be accessed, manually marking the faulty node as faulty, putting out of the membership and trigger a replan. This requires refining our membership definition
>
> More generally, my hunch is that we could have decoupled the two following tasks:
>computing a new plan if necessary.
> making sure the plan is applied.

fmassot:
# Description

Several improvements were made:
- [x] Avoid scheduling if the difference between the last applied plan and the running plan is a difference between tasks only.
- [x] Avoid scheduling if the last applied plan is the same as the newly built plan.
- [x] I removed the usage of an empty plan when there is no indexer, the code was confusing.
- [x] The diff between plans is fixed (it was wrong...). The current code is ugly, I still want to improve it...

# How this PR was tested
- [x] New tests have been added, so the coverage is way better.
- [x] Add a test to check no plan is scheduled during the `MIN_DURATION_BETWEEN_SCHEDULING_SECS` period.
- [x] Add unit test on indexing tasks diff.

### DIFF
--- a/quickwit/quickwit-cluster/src/cluster.rs
+++ b/quickwit/quickwit-cluster/src/cluster.rs
@@ -626,7 +626,7 @@ mod tests {
             .unwrap(),
         );
         let mut random_generator = rand::thread_rng();
-        let indexing_tasks = (0..10_000)
+        let indexing_tasks = (0..1_000)
             .map(|_| {
                 let index_id = random_generator.gen_range(0..=10_000);
                 let source_id = random_generator.gen_range(0..=100);
@@ -705,12 +705,14 @@ mod tests {
         cluster: Arc<Cluster>,
         gossip_public_address: SocketAddr,
         indexing_tasks: Vec<IndexingTask>,
-    ) -> anyhow::Result<bool> {
+    ) -> bool {
         let members = cluster.ready_members_from_chitchat_state().await;
-        let node = members
+        let node_opt = members
             .iter()
-            .find(|member| member.gossip_advertise_addr == gossip_public_address)
-            .ok_or_else(|| anyhow!("no cluster member"))?;
+            .find(|member| member.gossip_advertise_addr == gossip_public_address);
+        let Some(node) = node_opt else {
+            return false;
+        };
         let node_grouped_tasks: HashMap<IndexingTask, usize> = node
             .indexing_tasks
             .iter()
@@ -724,7 +726,7 @@ mod tests {
             .into_iter()
             .map(|(key, group)| (key, group.count()))
             .collect();
-        Ok(node_grouped_tasks == grouped_tasks)
+        node_grouped_tasks == grouped_tasks
     }
 
     #[tokio::test]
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -413,7 +413,7 @@ pub async fn create_cluster_for_test_with_id(
 /// Creates a failure detector config for tests.
 fn create_failure_detector_config_for_test() -> FailureDetectorConfig {
     FailureDetectorConfig {
-        phi_threshold: 6.0,
+        phi_threshold: 3.0,
```
fmassot, line 416: Reduce the threshold to detect faster a dead node during tests.

quickwit/quickwit-common/src/test_utils.rs
```diff
@@ -31,11 +31,11 @@ pub async fn wait_until_predicate<Fut>(
     retry_interval: Duration,
 ) -> Result<(), Elapsed>
 where
-    Fut: Future<Output = Result<bool, anyhow::Error>>,
+    Fut: Future<Output = bool>,
 {
     tokio::time::timeout(timeout, async move {
         loop {
-            if predicate().await.is_ok() {
+            if predicate().await {
```
fmassot, line 38: Forgot to fix it in my previous PR, we were not testing the right condition before...

quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -626,7 +626,7 @@ mod tests {
             .unwrap(),
         );
         let mut random_generator = rand::thread_rng();
-        let indexing_tasks = (0..10_000)
+        let indexing_tasks = (0..1_000)
```
fmassot, line 629: Modifications here are only made to make the test pass as we are now testing the right condition in `wait_untIl_predicate`

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -44,10 +46,9 @@ use crate::indexing_plan::{
 const REFRESH_PLAN_LOOP_INTERVAL: Duration = if cfg!(any(test, feature = "testsuite")) {
     Duration::from_secs(3)
 } else {
-    Duration::from_secs(60)
+    Duration::from_secs(60 * 5)
```
fmassot, line 49: Not sure about this one. I found 60 seconds a bit aggressive.

fmassot:
PTAL @guilload

guilload (approved):

PR #2843, opened 2023-02-18
Standardize API error response.

fmassot:
### Description

Currently, the body of API error responses can be very different. 
It can be of the form 
```json
{
  "error": "error message that corresponds to error.to_string()",
}
```

or any errors that are JSON serializable. For example, with a `SearchError::InvalidQuery`, we are getting:

```json
{
  "InvalidQuery": "Invalid query: a {message}"
}
```

The status code can be retrieved from the HTTP response.

I would like to make the following changes: make all errors return a JSON following this format:

```json
{
  "message": "typically error.to_string()",
}
```

I have a preference for the field "message" rather than "error".

### DIFF
--- a/quickwit/quickwit-serve/src/format.rs
+++ b/quickwit/quickwit-serve/src/format.rs
@@ -17,11 +17,14 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+use std::convert::Infallible;
+
 use hyper::header::CONTENT_TYPE;
 use hyper::StatusCode;
 use quickwit_proto::{ServiceError, ServiceErrorCode};
 use serde::{self, Deserialize, Serialize, Serializer};
 use warp::reply::{self, WithHeader, WithStatus};
+use warp::{Filter, Rejection, Reply};
 
 const JSON_SERIALIZATION_ERROR: &str = "JSON serialization failed.";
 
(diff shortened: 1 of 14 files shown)


fmassot:
@guilload I did not find an elegant way to use once and for all the `make_response` or the extraction of the format in the query params.

There are several reasons why I would not push for it now:

- we need to return the proper HTTP code 201 for "create" endpoints. This means we should return the result and a status code. `make_response` is not sufficient
- each handler currently returns `impl Reply` which is conv

quickwit/quickwit-serve/src/format.rs
```diff
@@ -55,49 +58,69 @@ impl Serialize for Format {
...
-#[derive(Serialize)]
-pub(crate) struct FormatError {
-    #[serde(skip_serializing)]
-    pub code: ServiceErrorCode,
-    pub error: String,
+/// This struct represents a QueryString passed to
+/// the REST API.
+#[derive(Deserialize, Debug, Eq, PartialEq, utoipa::IntoParams)]
+#[into_params(parameter_in = Query)]
+pub(crate) struct FormatQueryString {
```
guilload, line 65: `Format` should be `BodyFormat`.

guilload, line 65:
```suggestion
struct FormatQueryString {
```

quickwit/quickwit-serve/src/format.rs
```diff
@@ -55,49 +58,69 @@ impl Serialize for Format {
...
+    pub format: Format,
 }
 
-impl ToString for FormatError {
-    fn to_string(&self) -> String {
-        self.error.clone()
-    }
+pub(crate) fn extract_format_from_qs() -> impl Filter<Extract = (Format,), Error = Rejection> + Clone
+{
+    serde_qs::warp::query::<FormatQueryString>(serde_qs::Config::default()).and_then(
```
guilload, line 73:
```rust
serde_qs::warp::query::<FormatQueryString>(serde_qs::Config::default())
        .map(|format_qs: FormatQueryString| format_qs.format)
```

guilload (approved):

Issue #2790, opened 2023-02-10
Check sort by score is taken into account with the new CLI search command


fmassot:

evanxg852000:
The scoring works but the search API error message is not displayed correctly on the CLI when the index is node configured correctly to support scoring.

PR #2854, opened 2023-02-20
Check sort-by-score CLI

evanxg852000:
Make use of FormatError in make rest reply to display a more meaningful error message in CLI upon SearApiError

Closes https://github.com/quickwit-oss/quickwit/issues/2790


evanxg852000:
A standard API error message is provided in https://github.com/quickwit-oss/quickwit/pull/2843

PR #2855, opened 2023-02-20
Updated scoring example

evanxg852000:
Updated scoring query example with a term that has an observable effect
Closes https://github.com/quickwit-oss/quickwit/issues/2790


fmassot:
@evanxg852000 can you update `cli_doc_ext.toml` too?

evanxg852000:
Oops! That should even be the first file to update.

fmassot (approved):

Issue #2857, opened 2023-02-20
ingesting via file source often fails for no obvious reason to the user


trinity-1686a:
**Describe the bug**
the new file "source" sends files by chunks of 5k lines, which on some dataset can go over the limit of what the ingest api accepts.
On gh-archive, I had to lower that number to 800 to not actually get errors, but on worst dataset (or worst batch of the same dataset), this might not even be enough.

Instead of counting lines, we should count bytes and send just what the api will allow

**Steps to reproduce (if applicable)**

try to ingest a few hours of gh-archive dataset with the file source.

**Expected behavior**
it ingests without throwing an error

fmassot:
I had the same issue on the wikipedia dataset. We should get the size of each json line and ensure that we respect the API limit.

PR #2870, opened 2023-02-22
Limit ingest CLI payload

evanxg852000:
Limit ingest CLI payload chunk to 9MiB since the allows 10MB which is not exactly the same as 10MiB 
Currently using a peelable stream to limit more. A simpler and less accurate solution is to 
```rust
  loop {
      if buf_reader.read_until(b'\n', &mut buffer).await? == 0 {
          break;
      }
      if buffer.len() >= INGEST_CHUNK_SIZE_IN_BYTES {
          break;
      }
  }
```
But this is still error-prone since some dataset payload lines can be lengthier than others thus exceeding the allowed payload. 

Closes https://github.com/quickwit-oss/quickwit/issues/2857

### DIFF
--- a/quickwit/quickwit-rest-client/src/rest_client.rs
+++ b/quickwit/quickwit-rest-client/src/rest_client.rs
@@ -20,6 +20,8 @@
 use std::time::{Duration, Instant};
 
 use bytes::Bytes;
+use futures::pin_mut;
+use futures::stream::StreamExt;
 use quickwit_common::FileEntry;
 use quickwit_config::{ConfigFormat, SourceConfig};
 use quickwit_metastore::{IndexMetadata, Split};
@@ -31,13 +33,14 @@ use serde::Serialize;
 use serde_json::json;
 use tokio::fs::File;
 use tokio::io::{AsyncBufReadExt, AsyncRead, BufReader};
+use tokio_stream::wrappers::LinesStream;
 
 use crate::error::Error;
 use crate::models::{ApiResponse, IngestSource};
 
 pub static DEFAULT_ADDRESS: &str = "http://127.0.0.1:7280";
 pub static DEFAULT_CONTENT_TYPE: &str = "application/json";
-pub static INGEST_CHUNK_SIZE_IN_BYTES: usize = 6 * 1024 * 1024; // 6 MiB
+pub static INGEST_CHUNK_SIZE_IN_BYTES: usize = 9 * 1024 * 1024; // 9MiB
 
 pub struct Transport {
     base_url: Url,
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-rest-client/src/rest_client.rs
```diff
@@ -31,13 +33,14 @@ use serde::Serialize;
...
 use tokio::io::{AsyncBufReadExt, AsyncRead, BufReader};
+use tokio_stream::wrappers::LinesStream;
 
 use crate::error::Error;
 use crate::models::{ApiResponse, IngestSource};
 
 pub static DEFAULT_ADDRESS: &str = "http://127.0.0.1:7280";
 pub static DEFAULT_CONTENT_TYPE: &str = "application/json";
-pub static INGEST_CHUNK_SIZE_IN_BYTES: usize = 4 * 1024 * 1024;
+pub static INGEST_CHUNK_SIZE_IN_BYTES: usize = 9 * 1024 * 1024; // 9MiB
```
guilload, line 43: Why not 10?

quickwit/quickwit-rest-client/src/rest_client.rs
```diff
@@ -149,17 +152,30 @@ impl QuickwitClient {
...
         let start = Instant::now();
         let mut num_bytes_sent = 0;
 
         loop {
             let mut buffer = Vec::with_capacity(INGEST_CHUNK_SIZE_IN_BYTES);
-            for _ in 0..5_000 {
-                if buf_reader.read_until(b'\n', &mut buffer).await? == 0 {
-                    break;
-                }
+            while let Some(read_result) = peekable_reader
```
guilload, line 164: Indeed, this is not very readable. I prefer the "simple" solution, which we can improve to be accurate. We have to pop the last line from the current buffer and push it to a new one.

guilload (changes requested):

quickwit/quickwit-rest-client/src/rest_client.rs
evanxg852000: 10 MiB is a bit above what's allowed which is specified in 10MB (10 * 1000 * 1000) as opposed to (9 * 1024 * 1024)
We can harmonize to MB (10 * 1000 * 1000) because this is more readable.

guilload: Let's harmonize.

quickwit/quickwit-rest-client/src/rest_client.rs
evanxg852000: Yeah, I thought of this but I felt that's even more complicated and less idiomatic than the peelable solution. 
Please see the changes.

guilload:
Superseded by #2871.

PR #2871, opened 2023-02-22
Ingest files in batches of 10MiB or less

guilload:
### Description
REST client ingests files in batches of 10MiB or less.

### How was this PR tested?
- Added unit test for `BatchLineReader`
- Ingest one GH archive file with client

### DIFF
--- a/quickwit/quickwit-rest-client/src/lib.rs
+++ b/quickwit/quickwit-rest-client/src/lib.rs
@@ -17,6 +17,166 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+use std::path::Path;
+use std::{io, mem};
+
+use bytes::Bytes;
+use tokio::fs::File;
+use tokio::io::{AsyncBufReadExt, AsyncRead, BufReader};
+use tracing::warn;
+
 pub mod error;
 pub mod models;
 pub mod rest_client;
+
+pub(crate) struct BatchLineReader {
+    buf_reader: BufReader<Box<dyn AsyncRead + Send + Sync + Unpin>>,
+    buffer: Vec<u8>,
+    alloc_num_bytes: usize,
+    max_batch_num_bytes: usize,
+    num_lines: usize,
+}
+
+impl BatchLineReader {
+    pub async fn from_file(filepath: &Path, max_batch_num_bytes: usize) -> io::Result<Self> {
+        let file = File::open(&filepath).await?;
+        Ok(Self::new(Box::new(file), max_batch_num_bytes))
+    }
+
+    pub fn from_stdin(max_batch_num_bytes: usize) -> Self {
+        Self::new(Box::new(tokio::io::stdin()), max_batch_num_bytes)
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-rest-client/src/lib.rs
```diff
@@ -17,6 +17,124 @@
...
+
 pub mod error;
 pub mod models;
 pub mod rest_client;
+
+pub(crate) struct BatchLineReader {
+    buf_reader: BufReader<Box<dyn AsyncRead + Send + Sync + Unpin>>,
+    buffer: Vec<u8>,
+    alloc_num_bytes: usize,
+    batch_num_bytes: usize,
```
fmassot, line 35: maybe `max_batch_num_bytes`

quickwit/quickwit-rest-client/src/lib.rs
```diff
@@ -17,6 +17,124 @@
...
+            buf_reader: BufReader::new(reader),
+            buffer: Vec::with_capacity(alloc_num_bytes),
+            alloc_num_bytes,
+            batch_num_bytes,
+        }
+    }
+
+    pub async fn next_batch(&mut self) -> io::Result<Option<Bytes>> {
+        loop {
+            let num_bytes = self.buf_reader.read_until(b'\n', &mut self.buffer).await?;
```
fmassot, line 60: could be `line_num_bytes`

quickwit/quickwit-rest-client/src/lib.rs
```diff
@@ -17,6 +17,124 @@
...
+
+            if self.buffer.len() > self.batch_num_bytes {
+                let mut new_buffer = Vec::with_capacity(self.alloc_num_bytes);
```
fmassot, line 63: This little buffer dance is a bit annoying. My initial thought on this was to use `memchr::memchr` directly instead of `read_until`

quickwit/quickwit-rest-client/src/lib.rs
```diff
@@ -17,6 +17,124 @@
...
+            alloc_num_bytes,
+            batch_num_bytes,
+        }
+    }
+
+    pub async fn next_batch(&mut self) -> io::Result<Option<Bytes>> {
+        loop {
+            let num_bytes = self.buf_reader.read_until(b'\n', &mut self.buffer).await?;
+
+            if self.buffer.len() > self.batch_num_bytes {
```
fmassot, line 62: Even if it's not likely to happen, we could protect the code against a huge line that has size > `self.batch_num_bytes`

fmassot (approved):

quickwit/quickwit-rest-client/src/lib.rs
guilload: Yeah, I agree. I was not keen on adding a new dependency, though.

quickwit/quickwit-rest-client/src/lib.rs
guilload: Good point. Will skip the line and log a warning. Will add a unit test too :)

PR #2888, opened 2023-02-24
Fix source toggle bug in REST client.

fmassot:
### Description

Fix source toggle, wrong path, and wrong body. What is annoying is that the API did not return an error.

### How was this PR tested?

- [x] Added a test on the ingest API toggling.
- [x] Check why the API did not return an error

### DIFF
--- a/quickwit/quickwit-cli/src/source.rs
+++ b/quickwit/quickwit-cli/src/source.rs
@@ -415,13 +415,13 @@ async fn toggle_source_cli(args: ToggleSourceArgs) -> anyhow::Result<()> {
 pub async fn toggle_ingest_api_cli(args: ToggleIngestApiArgs) -> anyhow::Result<()> {
     debug!(args=?args, "toggle-ingest-api");
     println!("❯ Toggling ingest API...");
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-cli/src/source.rs
```diff
@@ -415,13 +415,13 @@ async fn toggle_source_cli(args: ToggleSourceArgs) -> anyhow::Result<()> {
 pub async fn toggle_ingest_api_cli(args: ToggleIngestApiArgs) -> anyhow::Result<()> {
     debug!(args=?args, "toggle-ingest-api");
     println!("❯ Toggling ingest API...");
```
guilload, line 417:
```suggestion
    println!("❯ {}abling ingest API...", if args.enabled { "En" } else { "Dis" });
```

guilload (approved):

quickwit/quickwit-cli/src/source.rs
fmassot: This suggestion is gold :)

PR #2890, opened 2023-02-24
Enable multiple pipelines for pulsar source.

### LINKED ISSUE #2889: Pulsar source can only have one running pipeline
Just need to change the code here:

```rust
    /// Returns `desired_num_pipelines` if it's a Kafka source, else 1.
    pub fn desired_num_pipelines(&self) -> usize {
        match &self.source_params {
            SourceParams::Kafka(_) => self.desired_num_pipelines,
            _ => 1,
        }
    }

    /// Returns `max_num_pipelines_per_indexer` if it's a Kafka source, else 1.
    pub fn max_num_pipelines_per_indexer(&self) -> usize {
        match &self.source_params {
            SourceParams::Kafka(_) => self.max_num_pipelines_per_indexer,
            _ => 1,
        }
    }
```

fmassot:
Fix #2889

### DIFF
--- a/quickwit/quickwit-config/src/source_config/mod.rs
+++ b/quickwit/quickwit-config/src/source_config/mod.rs
@@ -48,7 +48,6 @@ pub const INGEST_API_SOURCE_ID: &str = "_ingest-api-source";
 pub struct SourceConfig {
     pub source_id: String,
 
-    #[doc(hidden)]
     /// Maximum number of indexing pipelines spawned for the source on a given indexer.
     /// The maximum is reached only if there is enough `desired_num_pipelines` to run.
     /// The value is only used by sources that Quickwit knows how to distribute accross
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-config/src/source_config/mod.rs
```diff
@@ -135,15 +136,15 @@ impl SourceConfig {
     /// Returns `desired_num_pipelines` if it's a Kafka source, else 1.
```
ChillFish8, line 136: Nitpick, could do with updating this doc

fmassot: nice catch

quickwit/quickwit-config/src/source_config/mod.rs
```diff
@@ -135,15 +136,15 @@ impl SourceConfig {
...
     pub fn desired_num_pipelines(&self) -> usize {
         match &self.source_params {
-            SourceParams::Kafka(_) => self.desired_num_pipelines,
+            SourceParams::Kafka(_) | SourceParams::Pulsar(_) => self.desired_num_pipelines,
```
ChillFish8, line 139: What is the actual implication here with this on a source?

Does this prevent us from having multiple pulsar sources for different indexes per indexer, or is this for the specific source itself?

fmassot: > Does this prevent us from having multiple pulsar sources for different indexes per indexer, or is this for the specific source itself?

It's for the specific source. I'm testing pulsar on EKS and because of this, the control plane believes the pulsar source can only have 1  `desired_num_pipelines` and I can't test the distribution!!!!

fmassot: I hesitate to remove `pub fn desired_num_pipelines` and `pub fn max_num_pipelines_per_indexer`, I will open an issue on this.

ChillFish8 (approved):

PR #2903, opened 2023-02-27
QW-2897: Update pulsar message ID comment

### LINKED ISSUE #2897: Confirm this layout is correct?
https://github.com/quickwit-oss/quickwit/blob/1312b4fd21c31239b5f87f0988ecaf684aa96163/quickwit/quickwit-indexing/src/source/pulsar_source.rs#L379

ChillFish8:
Closes #2897 

Confirms the layout of the position ID for pulsar message IDs and adds an explanation as to why the order is what it is.

### DIFF
--- a/quickwit/quickwit-indexing/src/source/pulsar_source.rs
+++ b/quickwit/quickwit-indexing/src/source/pulsar_source.rs
@@ -376,7 +376,16 @@ async fn create_pulsar_consumer(
 fn msg_id_to_position(msg: &MessageIdData) -> Position {
     // The order of these fields are important as they affect the sorting
     // of the checkpoint positions.
-    // TODO: Confirm this layout is correct?
+    //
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -376,7 +376,16 @@ async fn create_pulsar_consumer(
 fn msg_id_to_position(msg: &MessageIdData) -> Position {
     // The order of these fields are important as they affect the sorting
     // of the checkpoint positions.
-    // TODO: Confirm this layout is correct?
+    //
+    // The key parts of the ID used for ordering are:
+    // - The ledger ID which is a sequentially increasing ID.
```
ChillFish8, line 381: To be confirmed, which unfortunately just involves a lot of source code reading :(

PR #2911, opened 2023-02-28
Issue/2910 dynamic by default

fulmicoton:
A Non-consensual PR.

Makes the dynamic mode the default.


fulmicoton:
Closed as it is in a different PR.

PR #2912, opened 2023-02-28
Redirect only GET request to UI.

### LINKED ISSUE #2830: REST server returns 200 OK for endpoints which should return 404s
While creating an index via the rest API I accidentally used just the base address of `http://127.0.0.1:7280`, unexpectedly, this returned a 200 OK from the POST request with a JSON body. It shouldn't as it's incredibly confusing to then wonder why the index doesn't actually appear.

I suspect this is an issue for any endpoints which don't exist, maybe our 404 handlers isn't working correctly?

fmassot:
### Description

Redirect only GET request to UI. Now POST request on the root url returns a 405. Fix #2830.

### How was this PR tested?

Test added.

### DIFF
--- a/quickwit/quickwit-serve/src/rest.rs
+++ b/quickwit/quickwit-serve/src/rest.rs
@@ -110,8 +110,9 @@ pub(crate) async fn start_rest_server(
         .or(elastic_api_handlers());
 
     let api_v1_root_route = api_v1_root_url.and(api_v1_routes);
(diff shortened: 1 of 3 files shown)


fmassot:
@ChillFish8 can you confirm that it's fix on your machine?

ChillFish8:
LGTM! Fixes the issue for my side, now getting a `405 Method not allowed` which is what we'd expect to see :)

PR #2917, opened 2023-02-28
QW-2905: Add pulsar tests with authorization

ChillFish8:
This adds token authorization to pulsar so that we can test authorization handling while testing pulsar.


guilload:
The test passes, but when I change the token's value, the test does not fail.

Issue #2595, opened 2022-12-19
Add docs to explain what Quickwit is doing in `qw_data` directory


fmassot:
Add also a recommendation on the split cache capacity.

PR #2919, opened 2023-03-01
Doc for qwdata W.I.P

evanxg852000:
Closes https://github.com/quickwit-oss/quickwit/issues/2595


docs/internals/data-directory.md
```diff
@@ -0,0 +1,43 @@
+# Quickwit data directory
+
+Quickwit operates on a local directory to store intermediates files. By default, this working directory is named `qwdata` and placed right next to the Quickwit binary. The following is a tree of how that directory might look like after some time of operating Quickwit.
```
fmassot, line 3:
```suggestion
Quickwit operates on a local directory to store temporary files. By default, this working directory is named `qwdata` and placed right next to the Quickwit binary. The following is a tree of how that directory might look like after some time of operating Quickwit.
```

docs/internals/data-directory.md
```diff
@@ -0,0 +1,43 @@
...
+│   └── wikipedia
+│       └── _ingest-api-source
+│           └── scratch
+└── queues
+    ├── partition_id
+    ├── wal-00000000000000000028
+    └── wal-00000000000000000029
+```
+
+- `/queues`: This directory contains write ahead log files allowing hight performance data ingestion on the ingest API. When data is sent to Quickwit via the ingest API, the data is stored in these queues to guard against data lost. The indexing pipelines then consume those queues at their own pace. Whenever an indexing pipeline commits based on `commit_timeout_secs`, the queue is truncated to free up the unnecessarily occupied storage. This directory gets created only on node running the ingest API service.
```
fmassot, line 24:
```suggestion
- `/queues`: This directory contains write ahead log files of the ingest API. When data is sent to Quickwit via the ingest API, the data is stored in these queues to guard against data lost. The indexing pipelines then consume those queues at their own pace. Whenever an indexing pipeline commits based on `commit_timeout_secs`, the queue is truncated to free up the unnecessarily occupied storage. This directory is created only if the ingest API service is running on your node.
```

docs/internals/data-directory.md
```diff
@@ -0,0 +1,43 @@
...
+
+- `indexing`: This directory hold the local indexing directory of each index managed by Quickwit. The local indexing directories are further separated by indexing source. In the above tree, you can see there is an indexing folder corresponding to `_ingest-api-source` source for `wikipedia` index. This directory gets created only on node running the indexing service.
```
fmassot, line 26:
```suggestion
- `indexing`: This directory hold the local indexing directory of each index managed by Quickwit. The local indexing directories are further separated by indexing source. In the above tree, you can see there is an indexing folder corresponding to `_ingest-api-source` source for `wikipedia` index. This directory gets created only on nodes running the indexing service.
```

docs/internals/data-directory.md
```diff
@@ -0,0 +1,43 @@
...
+
+- `delete_task_service`: This directory is used by the Janitor service to apply deletes on indexes. During this process, splits are downloaded, a new split is created while applying deletes and uploaded to the target storage. This directory gets created only on node running Janitor service.
```
fmassot, line 28:
```suggestion
- `delete_task_service`: This directory is used by the Janitor service to apply deletes on indexes. During this process, splits are downloaded, a new split is created while applying deletes and uploaded to the target storage. This directory gets created only on nodes running Janitor service.
```

docs/internals/data-directory.md
```diff
@@ -0,0 +1,43 @@
...
+
+- `cache`: This directory is used for caching splits created on this node. As we want to reduce time-to-search, we often creates splits that are far from mature in size and number of document. This leads to new splits going through a set of merges before getting mature. In order to avoid downloading a split that already existed on disk, we keep hold of newly created splits in cache as much as possible. This cache limits on number of splits as well as the overall size of splits it holds. Splits are evicted based on age only (oldest gets evicted).
```
fmassot, line 30: I think we should shorten the text that explains why and talk about the cache policy: `split_store_max_num_bytes` and `split_store_max_num_splits` and the fact that we evict splits older than 2 days and maybe add a link to the node config page.

The start could be:
```
This directory is used for caching splits created on this node. This avoids downloading young splits that will go through merges.
```

Ideally we should link to the page/section that explain merges.

docs/internals/data-directory.md
```diff
@@ -0,0 +1,43 @@
...
+
+- `delete_task_service`: This directory is used by the Janitor service to apply deletes on indexes. During this process, splits are downloaded, a new split is created while applying deletes and uploaded to the target storage. This directory gets created only on nodes running Janitor service.
+
+- `cache`: This directory is used for caching splits created on this node. As we want to reduce time-to-search, we often creates splits that are far from mature in size and number of document. This leads to new splits going through a set of merges before getting mature. In order to avoid downloading a split that already existed on disk, we keep hold of newly created splits in cache as much as possible. This cache limits on number of splits as well as the overall size of splits it holds. Splits are evicted based on age only (oldest gets evicted).
+
+## Keeping disk space in check
```
fmassot, line 32: I think this part is too long, it's fine for internal docs but we should make the user life easier on this and we should probably be more concrete and present a good rule of thumb.
Let me find some time this weekend to propose something.

evanxg852000: I will try to shorten it then

fmassot:
Thanks @evanxg852000 for the PR, the first part will definitely help users. I think the last part is a bit more tricky, so let's give us a few days to have something nicer on this.

As for the right section in the docs, I don't think it should be put that `concept`, my initial goal on this is to help users to understand how to operate correctly quickwit.

PR #2926, opened 2023-03-02
Address a bad interaction between rustfmt and codegeneration.

fulmicoton:
With the addition of services codegeneration, we end up creating new files that are not formatted in the sense of rustfmt.

As a result, every new run of `make fmt` results in a diff. Currently, the problem is solved by commit the formatted version of the code.

This is still not great, because now every call to `cargo build` shows a diff until we run `make fmt`.

Finally, the current code does not make it clear enough what is generated and what is not. The last thing we want is for people to start editing/committing modified generated code.

This PR introduces the following solution:

- generated code stays in src/. Putting it in target is a common way to fix these problem but it tends to confuse source code editor GoToDefinition. If generated code is not something we want to write, we often want to read it.

- generated code goes into a codegen subdirectories. That subdirectory is not covered by rustfmt anymore. We commit non formatted version of the file (it does not hurt readability).


quickwit/quickwit-codegen-example/src/hello.rs
```diff
@@ -16,7 +16,10 @@ pub struct HelloResponse {
 #[cfg_attr(any(test, feature = "testsuite"), mockall::automock)]
 #[async_trait::async_trait]
 pub trait Hello: std::fmt::Debug + dyn_clone::DynClone + Send + Sync + 'static {
-    async fn hello(&mut self, request: HelloRequest) -> crate::HelloResult<HelloResponse>;
+    async fn hello(
```
fulmicoton, line 19: This is the version of the file as produced by codegen.

guilload:
Thanks for looking into this. Codegen tries to format the code using `prettyplease`, but the output still differs from what you would obtain with `rustfmt`.

> This is still not great, because now every call to cargo build shows a diff until we run make fmt.

This is not exactly accurate. After checking out a branch, the first call to `cargo build` usually re-runs the codegen build script and emits code formatted differently. Still, the subsequent calls won't unless proto files have been updated, the build script has been modified, or the codegen logic has changed. Reasons for Cargo to re-run the build scripts include upgrading dependencies or updating the `Cargo.lock` file. There are probably others.

Anyway, your fix applies regardless and it's nice to put the generated files in a `codegen` directory.

quickwit/quickwit-codegen/quickwit-codegen-example/src/lib.rs
```diff
@@ -18,6 +18,8 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 mod error;
+
+#[path = "codegen/hello.rs"]
```
guilload, line 22: I didn't know that was a thing.

guilload (approved):

Issue #2929, opened 2023-03-02
azure is pulling openssl


fulmicoton:
I removed the deps from quickwit-elasticapi thing to openssl, but somehow azure is also pulling openssl. 
We need to work on finding the right feature cocktail to remove this deps.

Also I don't understand why we pull azure on a simple `cargo build` without any extra release flag.

PR #2930, opened 2023-03-02
Remove openssl dependency on azure 

evanxg852000:
Currently using `enable_reqwest_rustls` feature on all `azure_*` crates

Closes https://github.com/quickwit-oss/quickwit/issues/2929


fulmicoton (approved):
Thank you that was fast. Did you check that `cargo tree -i openssl` was  returning nothing?

evanxg852000:
Yes, also the simple `cargo build` was not pulling azure.

PR #2934, opened 2023-03-02
Generate service clients that support Tower middlewares

guilload:
### Description
You can now user stack Tower middlewares on top of any service:
```rust
let layer = ServiceBuilder::new()
    .timeout(Duration::from_secs(30))
    .concurrency_limit(1).into())

let mut hello_tower: HelloClient = HelloClient::tower()
    .hello_layer(layer)
    .service(hello_impl);
    
hello_tower.hello(...)
```

### How was this PR tested?
Updated example, which serves as a unit test.


quickwit/quickwit-ingest-api/src/codegen/ingest_service.rs
```diff
@@ -227,6 +230,138 @@ impl tower::Service<TailRequest> for IngestServiceClient {
...
+            Box<dyn IngestService>,
+            TailRequest,
+            FetchResponse,
+            crate::IngestServiceError,
+        >,
+    ) -> Self {
+        self.tail_layer = Some(layer);
+        self
+    }
+    pub fn service<T>(self, instance: T) -> IngestServiceClient
```
fmassot, line 337: `build` instead of `searvice`?

fmassot (approved):

quickwit/quickwit-ingest-api/src/codegen/ingest_service.rs
guilload: I hesitated on that one. I'm mimicking the Tower API, which uses `.layer().layer().service()`.

quickwit/quickwit-codegen/quickwit-codegen-example/src/codegen/hello.rs
```diff
@@ -92,6 +95,69 @@ impl tower::Service<HelloRequest> for HelloClient {
...
+            HelloRequest,
+            HelloResponse,
+            crate::HelloError,
+        >,
+    >,
+}
+impl HelloTowerBlockBuilder {
+    pub fn hello_layer(
+        mut self,
+        layer: quickwit_common::tower::BoxLayer<
```
guilload, line 137: This is where I could be happier about the API. I wanted layer build methods to accept `*_layer<S>(layer: Layer<S>) where ...` and perform the boxing inside the functions. Unfortunately, I could not do so while making the compiler happy, and I gave up, given the pressing need to ship this feature for the next release.

Anybody, feel free to take a stab.

PR #2941, opened 2023-03-04
Fix vector configuration in tutorial

Maniarr:
### Description

In the tutorial for sending logs to the vector. There is an error on the vector configuration that causes an error, by default the sink vector `http` uses the HTTP method `GET` and quickwit responds with an HTTP 405 error when ingesting logs.

We need to explicitly define the use of the `POST` method.

Error in vector logs:

```
vector::internal_events::http_client: HTTP response. status=405 Method Not Allowed version=HTTP/1.1 ...
```

### How was this PR tested?

PR changes only documentation file.

### DIFF
--- a/docs/get-started/tutorials/send-logs-from-vector-to-quickwit.md
+++ b/docs/get-started/tutorials/send-logs-from-vector-to-quickwit.md
@@ -163,6 +163,7 @@ source = '''
 
 [sinks.quickwit_logs]
 type = "http"
+method = "post"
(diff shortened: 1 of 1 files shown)


fmassot:
@Maniarr thanks a lot!

fmassot (approved):

fmassot:
@Maniarr I have just updated the docs on logs for the main branch: https://quickwit.io/docs/main-branch/log-management/overview

PR #2945, opened 2023-03-05
Protect code against empty or negative partition delta.

### LINKED ISSUE #2944: Protect the code against recording negative partition deltas
Currently it's possible to record negative partition deltas:

The following test is passing:

```rust
    fn test_record_negative_partition_delta() {
        let mut delta = SourceCheckpointDelta::from_partition_delta(
            PartitionId::from("a"),
            Position::from("10"),
            Position::from("20"),
        );
        delta.record_partition_delta(
            PartitionId::from("a"),
            Position::from("20"),
            Position::from("10"),
        ).unwrap();
    }
```

Of course, this should never happen...  but with Pulsar failover mechanism, Quickwit can receive messages from a given partition that has already been received, see #2893. 

I'm gonna fix the pulsar source that should skip old messages. But IMHO we should also be defensive when recording partition deltas.

fmassot:
### Description

This PR prevents from instantiating a `SourceCheckpointDelta` or recording a partition delta if the delta is empty or negative.

- Fix #2944 
- Replace `From<Range<u64>>` implementation of `SourceCheckpoint` by `TryFrom`
- Add `SourceCheckpoint::from_range` only for testing
- Remove assert in `record_partition_delta` as we now have a dedicated error.

For example:
- `record_partition_delta(partition_1, 5, 5)` is invalid, the recorded range `(5, 5]` is empty
- `record_partition_delta(partition_1, 5, 4)` is invalid

### How was this PR tested?

Tests added.


quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -318,26 +318,27 @@ mod tests {
...
 
     #[tokio::test]
-    async fn test_file_source_resume_from_checkpoint() -> anyhow::Result<()> {
+    async fn test_file_source_resume_from_checkpoint() {
         quickwit_common::setup_logging_for_tests();
         let universe = Universe::with_accelerated_time();
         let (doc_processor_mailbox, doc_processor_inbox) = universe.create_test_mailbox();
         use tempfile::NamedTempFile;
-        let mut temp_file = NamedTempFile::new()?;
+        let mut temp_file = NamedTempFile::new().unwrap();
```
fmassot, line 326: just me replacing `?` by `unwrap` in tests I udpate.

fulmicoton (approved):

PR #2948, opened 2023-03-05
Move ingest data tutorials in dedicated sections, improve navigation.

fmassot:
### Description

- Move ingest tutorials in dedicated section
- Add an index page to show the different possibilities
- Add a tutorial for ingest API so we cover the main sources.

### How was this PR tested?

Tested website rendering.


docs/ingest-data/index.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Ingest data from multiple sources
+---
+
+import DocCardList from '@theme/DocCardList';
+
+<DocCardList />
```
fmassot, line 7: It's a react component that renders one div block per ingest tutorial (see screenshot).

Issue #2834, opened 2023-02-17
Records in the queues should be more than just a payload


fulmicoton:
The PushAPI relies on `mrecordlog` to store the documents that were ingested as is.

This is fine for the moment, but it will prevent us from further evolution...
In particular #2699 will most likely require us to append some commit Command to the queue.

We need the record in the queue to not be directly documents, but rather something closer to 

```rust
enum Command<'a> {
     Ingest {
         json_payload: &'a [u8]
     },
     Commit,
     // ... more to come?
}
```
with an adhoc simple, versioned and extensible format.

imotov:
I can see two driving forces that can influence how this change should be implemented. First, since the data that we handle here can be huge, it is performance sensitive. That part is pretty clear. The second aspect (that I don't quite understand the motivation behind) is having `mrecordlog` as a separate project. Depending on how generic we want `mrecordlog` to remain I can see 3 ways of implementing this change.

1) Leave `mrecordlog` as is and add some logic to encode `Command` into `&'a [u8]`. This will allow `mrecordlog` to remain unchanged, but that will mean we will have to clone the payload twice, which is not good from the performance perspective. So, I would say this option is no go.
2) Change `mrecordlog`'s `append_records` function to accept an iterator over something like `mrecordlog::record::Serializable` instead of `&'a [u8]`. 
3) Make `mrecordlog` aware of `Commit` and possibly other marker records that we might want to add to the log in the future. 

Assuming that we want to keep `mrecordlog` generic 2) seems to be the best solution. But as any abstraction it comes at the price of `mrecordlog` not knowing what it is storing. So if later we will want to add a viewer to `mrecordlog` or make it more intelligent about that it actually stores the option 3) might be more useful, but in this case it might also make sense to consider merging `mrecordlog` into quickwit.

guilload:
@imotov, we're planning on switching from `&[u8]` to [`Bytes`](https://docs.rs/bytes/latest/bytes/) (reference counted slice of memory) in the ingest pipeline in 0.6.0, so we can still consider 1.

imotov:
@guilload, did you mean using `bytes::Buf` instead of `&'a [u8]` in 2)? In other words, we would change the `mrecordlog` interface to look something like this:

```
    pub async fn append_records<T: Iterator<Item = Buf>>(
        &mut self,
        queue: &str,
        position_opt: Option<u64>,
        payloads: T,
    ) -> Result<Option<u64>, AppendError> {
....
    }

[...]

PR #2950, opened 2023-03-05
Convert records in ingest payload into commands

imotov:
### Description

Makes it possible to inject different commands such as force commit alongside with documents into
the ingest stream.

See #2834

### How was this PR tested?

Added unit tests for the new code. Ran all test for the old code.


fulmicoton:
@trinity-1686a can you have a look?

quickwit/quickwit-opentelemetry/src/otlp/logs.rs
```diff
@@ -343,25 +341,21 @@ impl OtlpGrpcLogsService {
...
         }
         let current_span = RuntimeSpan::current();
         current_span.record("num_log_records", num_log_records);
-        current_span.record("num_bytes", doc_batch.concat_docs.len());
+        current_span.record("num_bytes", doc_batch.len());
         current_span.record("num_parse_errors", num_parse_errors);
 
         let parsed_spans = ParsedLogRecords {
-            doc_batch,
+            doc_batch: doc_batch.build(),
```
trinity-1686a, line 358: It's unclear to me whether this change causes double-counting in metrics, or fix not counting at all

quickwit/quickwit-opentelemetry/src/otlp/trace.rs
```diff
@@ -603,25 +601,21 @@ impl OtlpGrpcTraceService {
...
         }
         let current_span = RuntimeSpan::current();
         current_span.record("num_spans", num_spans);
-        current_span.record("num_bytes", doc_batch.concat_docs.len());
+        current_span.record("num_bytes", doc_batch.len());
         current_span.record("num_parse_errors", num_parse_errors);
 
         let parsed_spans = ParsedSpans {
-            doc_batch,
+            doc_batch: doc_batch.build(),
```
trinity-1686a, line 618: same wrt maybe causing double-counting  or fixing not  counting

quickwit/quickwit-serve/src/ingest_api/rest_handler.rs
```diff
@@ -156,15 +156,12 @@ async fn ingest(
...
-        ..Default::default()
-    };
+    let mut doc_batch = DocBatchBuilder::new(index_id);
     for doc_payload in lines(&payload) {
-        add_doc(doc_payload.as_bytes(), &mut doc_batch);
+        doc_batch.ingest_doc(doc_payload.as_bytes());
     }
     let ingest_req = IngestRequest {
-        doc_batches: vec![doc_batch],
+        doc_batches: vec![doc_batch.build()],
```
trinity-1686a, line 164: here the behavior doesn't change, but I think `INGEST_METRICS.ingested_num_bytes` will be updated twice, at ingest (here), and when fetching from the queue

quickwit/quickwit-serve/src/ingest_api/rest_handler.rs
```diff
@@ -259,16 +256,18 @@ async fn elastic_ingest(
...
 
-        add_doc(source.to_string().as_bytes(), doc_batch);
+        doc_batch.ingest_doc(source.to_string().as_bytes());
     }
 
     let ingest_request = IngestRequest {
-        doc_batches: batches.into_values().collect(),
+        doc_batches: batches
+            .into_values()
+            .map(|builder| builder.build())
```
trinity-1686a, line 269: same wrt no change in behavior but possible double-counting

quickwit/quickwit-ingest-api/src/doc_batch.rs
```diff
@@ -0,0 +1,420 @@
...
+{
+    Ingest { payload: T },
+    Commit,
+    // ... more to come?
+}
+
+/// We can use this byte to track both commands and their version changes
+/// If serialization protocol changes, we can just use the next number
+const INGEST_CODE_V1: u8 = 0;
+const COMMIT_CODE_V1: u8 = 1;
```
trinity-1686a, line 40: this should probably be a `repr(u8)` enum

quickwit/quickwit-ingest-api/src/doc_batch.rs
```diff
@@ -0,0 +1,420 @@
...
+
+impl<T> DocCommand<T>
+where T: Buf
+{
+    /// Returns the
```
trinity-1686a, line 45: the end of the sentence is missing

quickwit/quickwit-ingest-api/src/doc_batch.rs
```diff
@@ -0,0 +1,420 @@
...
+    pub fn code(&self) -> u8 {
+        match self {
+            DocCommand::Ingest { payload: _ } => INGEST_CODE_V1,
+            DocCommand::Commit => COMMIT_CODE_V1,
+        }
+    }
+
+    /// Builds a command for bytes::Buf
+    pub fn read(buf: T) -> Self {
+        let mut buf = buf;
```
trinity-1686a, line 55:
```suggestion
    pub fn read(mut buf: T) -> Self {
```

quickwit/quickwit-ingest-api/src/doc_batch.rs
```diff
@@ -0,0 +1,420 @@
...
+    pub fn new(index_id: String) -> Self {
+        Self {
+            index_id,
+            concat_docs: BytesMut::new(),
+            doc_lens: Vec::new(),
+        }
+    }
+
+    /// Adds an ingest command to the batch
+    pub fn ingest_doc(&mut self, payload: impl Buf + Default) -> usize {
```
trinity-1686a, line 95: why does it requires `Default`?

quickwit/quickwit-ingest-api/src/doc_batch.rs
```diff
@@ -0,0 +1,420 @@
...
+        self.concat_docs.len()
+    }
+
+    /// Returns true if the batch is empty
+    pub fn is_empty(&self) -> bool {
+        self.doc_lens.is_empty()
+    }
+
+    /// Returns the number of document in the batch
+    pub fn count(&self) -> usize {
```
trinity-1686a, line 142: the doc comment should clarify it includes Commit commands

trinity-1686a (dismissed):
some small change here and there, but overall LGTM!

You can disregards comments on possible double-counting. I'll open a ticket on that as I'm unsure of the state it was in before and how it is now, so I (or someone else) will check how it actually is

guilload:
Can we also use `num_docs` and `num_bytes` instead of `count` and `len`, please?

quickwit/quickwit-opentelemetry/src/otlp/logs.rs
imotov: I don't think we are double counting since I removed increment in add_doc and moved it into build() and I cannot find any other places where it would be incremented, but it seems we were not counting at all in the past. @guilload is there a reason not to count these records in logs.rs and trace.rs?

quickwit/quickwit-ingest-api/src/doc_batch.rs
imotov: I might be missing something, but it didn't look like it would be worth it. If DocCommand was a plain enum that we could have easily casted it to u8, but since it is using generics and payload, I couldn't find any way of getting this code except through https://doc.rust-lang.org/std/mem/fn.discriminant.html and I wasn't sure if introducing unsafe code was justified here.

quickwit/quickwit-opentelemetry/src/otlp/logs.rs
guilload: I was not aware that counting was performed when building the doc batch. `IngestApiService::ingest` might be a better place.

imotov: Thanks! I will take a look at pushing it back there then. It is a bit weird since we are counting bytes and records in different places, but I wasn't sure if there was a reason for it.

quickwit/quickwit-ingest-api/src/doc_batch.rs
imotov: Leftover from one of the iterations where it was needed. Don't need it anymore. Great catch!

quickwit/quickwit-ingest-api/src/doc_batch.rs
trinity-1686a: I was thinking of a 2nd enum mirroring `DocCommand`, instead of a bunch of free-floating constants

imotov: Great idea!

trinity-1686a:
@imotov can you change the name of a few functions as guiload asked? After that I think we are good to go

guilload (approved):

PR #2959, opened 2023-03-07
QW-2958:  Compress large responses on the REST server

### LINKED ISSUE #2958: Add the ability to compress response payloads
Currently, the REST server will never compress the response data, which is fine for most cases.

The issue comes if you're performing big aggregations and return 50 - 100MB responses or more.
Compression would be very nice here and should be very easy to add as middleware.

ChillFish8:
Closes #2958 

Adds middleware that compresses large response bodies with gzip.
This technically applies to all routes but will most of the time only be applied in things like serving the UI or large aggregations.


quickwit/quickwit-serve/src/rest.rs
```diff
@@ -125,10 +131,24 @@ pub(crate) async fn start_rest_server(
...
+    let compression_predicate =
+        DefaultPredicate::new().and(SizeAbove::new(MINIMUM_RESPONSE_COMPRESSION_SIZE));
+
+    let service = ServiceBuilder::new()
+        .layer(
+            CompressionLayer::new()
+                .gzip(true)
+                .compress_when(compression_predicate),
+        )
+        .service(warp_service);
```
ChillFish8, line 144: To clarify why we use tower_http here rather than warp's version.

Warp's compression middleware compresses all responses without any adjustment, making it compress at times we don't need or want, the tower middleware is more adjustable for this purpose.

It also lays the ground work in situations if we want to be able to send compressed bodies to Quickwit via the ingest API in future.

quickwit/quickwit-serve/src/rest.rs
```diff
@@ -125,10 +131,24 @@ pub(crate) async fn start_rest_server(
...
+
     info!("Searcher ready to accept requests at http://{rest_listen_addr}/");
-    let rest_server: Pin<Box<dyn Future<Output = ()> + Send>> =
-        Box::pin(warp::serve(rest_routes).run(rest_listen_addr));
-    rest_server.await;
+
+    // let service = Buffer::new(service, 1024);
```
fmassot, line 148: comment to remove

quickwit/quickwit-serve/src/rest.rs
```diff
@@ -44,6 +46,10 @@ use crate::search_api::{search_get_handler, search_post_handler, search_stream_h
 use crate::ui_handler::ui_handler;
 use crate::{with_arg, BodyFormat, QuickwitServices};
 
+/// The minimum size a response body must be in order to
+/// be automatically compressed with gzip.
+const MINIMUM_RESPONSE_COMPRESSION_SIZE: u16 = 10 << 10;
```
fmassot, line 51: Sounds reasonable. Do you have a rule of thumb to estimate the  right `MINIMUM_RESPONSE_COMPRESSION_SIZE`?

ChillFish8: Not really, it's a fairly arbitrary size, but a size derived so that the general REST API requests won't get needlessly compressed, and small searches won't either while catching the bigger things.

fmassot (dismissed):

fmassot:
> I want to check the compression predicate stuff, I think we are compressing image with your change

Ok so as you use the default compression predicate, we don't compression image

fmassot (approved):

PR #2960, opened 2023-03-08
Deactorify ingest service

guilload:
### Description
Deactorify ingest service so we can serve multiple requests concurrently.

### How was this PR tested?
In progress


fulmicoton:
Can you describe in the commit message a little bit more what problem you are trying to solve?

Nicolas-Richard:
> Can you describe in the commit message a little bit more what problem you are trying to solve?

+ 1

quickwit/quickwit-indexing/src/source/ingest_api_source.rs
```diff
@@ -125,12 +127,13 @@ impl Source for IngestApiSource {
             start_after: self.counters.current_offset,
             num_bytes_limit: None,
         };
+        println!("fetch_req: {:?}", fetch_req);
```
fulmicoton, line 130:
```suggestion
```

quickwit/quickwit-ingest-api/src/ingest_api_service.rs
```diff
@@ -93,41 +91,63 @@ impl IngestApiService {
...
+
+    pub async fn drop_queue(&self, request: DropQueueRequest) -> crate::Result<()> {
+        self.queues.drop_queue(&request.queue_id).await?;
+        Ok(())
+    }
+
+    async fn ingest_inner(&self, request: IngestRequest) -> crate::Result<IngestResponse> {
+        // Check all indexes exist assuming existing queues always have a corresponding index.
+        for doc_batch in &request.doc_batches {
+            if !self.queues.queue_exists(&doc_batch.index_id).await {
```
fulmicoton, line 128: We have a race condition now... If someone deletes the queues after the existence check, the ingest request can fail partially.

fulmicoton: Alternatively we can just properly identify & accept the new behavior of this function...

fulmicoton (approved):
I'm approving this PR to keep thing smooth, but it needs more work, or at least more investigation on the race condition I pointed.

Removing the burden of bookeeping the existence of the queues from the recordlog miht be a good idea, but if it is the way to go, it is  a decent amount of work. let's not do this for this release.

I recommend against waiting on the memory being available and would rather just return a 413... And work on tweaking the rate limiting logic (wait 1s for every call if we reach a threshold of the memory) to make sure it never happens if things are working correctly, but see what you see fit!

guilload:
Ok, I'm in for not waiting for memory to become available and returning a 429 immediately instead. Let's handle everything in the rate limiter and modulator.

Then, the ingest service can remain an actor, which simplifies things in terms of handling concurrency. I'm going to revert most of the changes in that PR and just keep the minor code clean-ups that I made.

guilload:
@imotov also touches this file in his next PR, so I'll push the code clean up post-release.

PR #2963, opened 2023-03-08
Track memory usage of ingest service

guilload:


quickwit/quickwit-ingest-api/src/memory_capacity.rs
```diff
@@ -0,0 +1,166 @@
...
+        Self {
+            inner: Arc::new(InnerMemoryCapacity {
+                max_capacity,
+                capacity: AtomicUsize::new(max_capacity),
+                notify: Notify::new(),
+            }),
+        }
+    }
+
+    pub async fn acquire(&self, num_bytes: usize) {
```
fulmicoton, line 51: /// Attempt to reserve a given `capacity`. Waits until the capacity is available.

quickwit/quickwit-ingest-api/src/memory_capacity.rs
```diff
@@ -0,0 +1,166 @@
...
+        f.debug_struct("MemoryCapacity")
+            .field("capacity", &self.capacity())
+            .field("max_capacity", &self.max_capacity())
+            .finish()
+    }
+}
+
+impl MemoryCapacity {
+    pub fn new(max_capacity: usize) -> Self {
+        Self {
```
fulmicoton, line 42:
```suggestion
        assert!(max_capacity > 0, "The memory capacity is required to be > 0.");
        Self {
```

quickwit/quickwit-ingest-api/src/memory_capacity.rs
```diff
@@ -0,0 +1,166 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("MemoryCapacity")
+            .field("capacity", &self.capacity())
+            .field("max_capacity", &self.max_capacity())
+            .finish()
+    }
+}
+
+impl MemoryCapacity {
+    pub fn new(max_capacity: usize) -> Self {
```
fulmicoton, line 41: YOu have a hidden contract here.

Ideally make it 
```suggestion
    pub fn new(max_capacity: NonZeroU64) -> Self {
```
but sometimes NonZeroU64 are annoying. An alternative is to put an assert and document the contract.
or

quickwit/quickwit-ingest-api/src/memory_capacity.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    pub fn capacity(&self) -> usize {
+        self.inner
+            .capacity
+            .load(std::sync::atomic::Ordering::Relaxed)
+    }
+
+    pub fn capacity_ratio(&self) -> f64 {
+        self.capacity() as f64 / self.max_capacity() as f64
```
fulmicoton, line 95:
```suggestion
      if self.max_capacity() == 0 {
          return 0f64;
      }
      self.capacity() as f64 / self.max_capacity() as f64
```

fulmicoton (approved):

guilload:
Since we're no longer waiting to be notified that some memory is available, I'm removing the `Notify` attribute from the `MemoryCapacity` object.

Since the ingest API service remains an actor, the `MemoryCapacity` no longer requires to be thread-safe, but I'm not undoing that part.

PR #2964, opened 2023-03-08
Implement sma rate estimator and plug into rate limiter

guilload:


quickwit/quickwit-common/src/tower/rate_estimator.rs
```diff
@@ -17,10 +17,236 @@
...
+            / initial_rate.period().as_millis() as u64;
+        for i in 0..self.inner.num_buckets {
+            self.inner.buckets[i as usize].update(initial_work);
+        }
+        self
+    }
+}
+
+impl Rate for SmaRateEstimator {
+    fn work(&self) -> u64 {
```
fulmicoton, line 89: The definitin of what work seems non trivial. 
Can you document?

// Returns the estimated amount of work to be processed in a given `period`.

fulmicoton, line 89: // Returns an estimate of the amount of work to be achieved in a given `period`.

quickwit/quickwit-common/src/tower/rate_estimator.rs
```diff
@@ -17,10 +17,236 @@
...
+    bucket_period_secs: u64,
+    bucket_period_millis: u64,
+    num_buckets: u64,
+    period: Duration,
+    period_millis: u64,
+}
+
+impl SmaRateEstimator {
+    /// Creates a new simple moving average rate estimator.
+    ///
```
fulmicoton, line 51:
```suggestion
    ///
    /// This rate estimator is bucket based, and simply outputs that would be output in `period`,
    /// assuming the rate of work is the same as the average over the previous closed` N-1` bucket.
```

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -276,6 +305,52 @@ pub async fn serve_quickwit(config: QuickwitConfig) -> anyhow::Result<()> {
...
+        }
+    }
+}
+
+impl<R> Rate for RateModulator<R>
+where R: Rate
+{
+    fn work(&self) -> u64 {
+        let memory_usage_ratio = self.memory_capacity.usage_ratio();
+        let work = self.rate.work();
```
fulmicoton, line 328: The tricky part is that ideally we would like work to be a "max indexing throughpput estimator", however right now it is a actual_throughput estimator.

If we don't have any traffic, it drops to 0.

The obvious quickfix is to 

```
let work = self.rate.work().max(MIN_WORK);
````

And if you have nicer ways to measure the maximum throughput it would be better.

fulmicoton (approved):

PR #2970, opened 2023-03-09
WIP: QW-2597: Initial `quickwit init` command implementation.

ChillFish8:
### Description
This will add the initial `quickwit init` command, designed to make creating configs and setting up Quickwit as simple and as pain-free as possible.

The system reads a schema defined as a yaml which describes the various properties of the config (validation, nested objects, selects, etc...) and walks the user through populating the config with descriptions to match.

This initial PR focuses on the main configs and currently doesn't intend to implement the templating logic that we want further down the line.

**Command Goals:**
- `quickwit init` - Sets up the base Quickwit config itself for when you're first setting up a node.
- `quickwit init index` - Sets up an index config.
- `quickwit init source` - Sets up a source config.

**Checklist:**
- [x] Implement data source schema.
- [ ] Implement index schema.
- [ ] Implement Quickwit config schema.

Originally this was going to be written as a derived macro but it very quickly became apparent that it would be way too much time to implement and test vs going with the yaml approach, either way, they save writing hundreds of lines of rust code to prompt the user for each config.

**Preview**:
image

### How was this PR tested?

Currently, there is not testing, I'm still trying to work out how best to go about unit testing something which takes input from stdin.

### DIFF
--- a/quickwit/Cargo.lock
+++ b/quickwit/Cargo.lock
@@ -4141,6 +4141,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "promptea"
+version = "0.1.0"
+source = "git+https://github.com/quickwit-oss/promptea.git?rev=29a6310#29a631086e49f196e69dd7220c1cf197b180d9e4"
+dependencies = [
+ "Inflector",
+ "console",
+ "dialoguer",
+ "indexmap",
+ "regex",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
 name = "proptest"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


fmassot:
This is great, I tested it, and I liked the dev XP. And I also found the idea of defining this in a `YAML` nice.

A few remarks:
- we should add the `pulsar://` address validation in quickwit pulsar source params, I thought about it before but forgot to open an issue, can you open it? :)
- I saw some `null` values in the output files
- (minor) The help text of the file source is not up to date

ChillFish8:
> we should add the pulsar:// address validation in quickwit pulsar source params, I thought about it before but forgot to open an issue, can you open it? :)

Have done :)

> I saw some null values in the output files

Yes currently the `skipped` fields are made into `null` which isn't ideal, and currently still something to be ironed out.

> The help text of the file source is not up to date

Ah yes, I think currently it's on the 0.4.0 help-wise, I'll make sure everything is up to date before pbulishing

PR #2979, opened 2023-03-11
Switch `RawDocBatch.docs` from `Vec<String>` to `Vec<Bytes>`

guilload:
### Description
This has two advantages:
1. Sometimes, when the source returns a `Bytes` object (mrecordlog, Kinesis), we can reuse the object as is and save an allocation.

2. We no longer have to check that the doc is valid UTF-8 in each source. The check is performed once in the doc processor when going from `&[u8]` to `JsonValue`

### How was this PR tested?
- Added unit tests

### DIFF
--- a/quickwit/quickwit-indexing/src/models/raw_doc_batch.rs
+++ b/quickwit/quickwit-indexing/src/models/raw_doc_batch.rs
@@ -18,26 +18,55 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::fmt;
+use std::ops::Range;
 
+use bytes::Bytes;
 use quickwit_metastore::checkpoint::SourceCheckpointDelta;
 
-#[derive(Clone, Default)]
+#[derive(Default)]
 pub struct RawDocBatch {
-    pub docs: Vec<String>,
+    pub docs: Vec<Bytes>,
     pub checkpoint_delta: SourceCheckpointDelta,
     pub force_commit: bool,
 }
 
 impl RawDocBatch {
     pub fn new(
-        docs: Vec<String>,
+        docs: Vec<Bytes>,
         checkpoint_delta: SourceCheckpointDelta,
         force_commit: bool,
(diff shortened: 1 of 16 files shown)


quickwit/quickwit-indexing/src/source/ingest_api_source.rs
```diff
@@ -145,18 +145,15 @@ impl Source for IngestApiSource {
...
-        let mut raw_doc_batch = RawDocBatch::default();
-        raw_doc_batch.docs.extend(
-            doc_batch
-                .iter()
-                .map(|command| match command {
-                    DocCommand::Ingest { payload } => payload,
-                    // TODO: We will add a commit logic in a follow-up PR, for now - just panic
-                    _ => panic!("Unknown command {command:?}"),
-                })
-                .map(|buff| String::from_utf8_lossy(buff.as_ref()).to_string()),
```
guilload, line 157: This line illustrates the purpose of this PR.

quickwit/quickwit-indexing/src/models/raw_doc_batch.rs
```diff
@@ -18,29 +18,56 @@
...
+        }
+    }
 }
 
 impl fmt::Debug for RawDocBatch {
     fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
         formatter
             .debug_struct("RawDocBatch")
-            .field("docs_len", &self.docs.len())
+            .field("num_docs", &self.docs.len())
```
imotov, line 70:
```suggestion
            .field("num_docs", &self.num_docs())
```

quickwit/quickwit-indexing/src/source/ingest_api_source.rs
imotov: Makes sense.

imotov (approved):
LGTM. I added a couple of cosmetic suggestions.

Issue #2981, opened 2023-03-12
Control plane subscription on metastore event is not working


fmassot:
The subscription handle is dropped just after being created.

fmassot:
Fixed by #2982

PR #2982, opened 2023-03-12
Fix control plane notifications

fmassot:
### Description

The control plane notifications are not working for 2 reasons:
- one is described in #2981
- in a cluster mode, the control plane is not notified as we don't instantiate the control plane client to send the notifications, so nothing is sent!

### How was this PR tested?

- [x] add a test on control plane notification on a standalone server
- [x] add a test on control plane notification on a multi-node cluster

### DIFF
--- a/quickwit/quickwit-serve/src/lib.rs
+++ b/quickwit/quickwit-serve/src/lib.rs
@@ -55,7 +55,7 @@ use itertools::Itertools;
 use once_cell::sync::OnceCell;
 use quickwit_actors::{Mailbox, Universe};
 use quickwit_cluster::{Cluster, ClusterMember};
-use quickwit_common::pubsub::EventBroker;
+use quickwit_common::pubsub::{EventBroker, EventSubscriptionHandle};
 use quickwit_common::tower::{
     BufferLayer, ConstantRate, EstimateRateLayer, Rate, RateLimitLayer, SmaRateEstimator,
 };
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-serve/src/lib.rs
```diff
@@ -176,20 +180,37 @@ pub async fn serve_quickwit(config: QuickwitConfig) -> anyhow::Result<()> {
...
+    } else if config
+        .enabled_services
+        .contains(&QuickwitService::Metastore)
+    {
+        let (channel, _) = create_balance_channel_from_watched_members(
+            cluster.ready_member_change_watcher(),
+            QuickwitService::ControlPlane,
+        )
+        .await?;
+        Some(ControlPlaneServiceClient::from_channel(channel))
```
fulmicoton, line 204: classy.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -176,20 +180,37 @@ pub async fn serve_quickwit(config: QuickwitConfig) -> anyhow::Result<()> {
...
             start_control_plane_service(&universe, cluster.clone(), metastore.clone()).await?;
-        let control_plane_service = ControlPlaneServiceClient::from_mailbox(control_plane_mailbox);
-        event_broker.subscribe::<MetastoreEvent>(control_plane_service.clone());
-        Some(control_plane_service)
+        Some(ControlPlaneServiceClient::from_mailbox(
+            control_plane_mailbox,
+        ))
+    } else if config
+        .enabled_services
+        .contains(&QuickwitService::Metastore)
```
fulmicoton, line 197: silly question... why do we need the metastore to be running on the node to attempt to connect to the control plane?

fulmicoton (approved):
Approved but can we discuss my question when you get up?

Issue #2699, opened 2023-01-24
Add ability to force an immediate commit after indexing


imotov:
**Is your feature request related to a problem? Please describe.**
In my experience the simplest way to explain a difficult concept to a search engine user is by demonstrating this concept using a shell script that consists of several curl statements. I have been using this method for many years. You can find some examples of such scripts in [imotov/elasticsearch-test-scripts](https://github.com/imotov/elasticsearch-test-scripts) repo.

Trying the same approach with quickwit uncovered 2 issues, one of which is absence of refresh operation that would force all indexed data to be searchable. It is necessary since most of the scripts need to perform the following operations:

- recreate an index with certain mapping
- index some test data
- perform the search operation on such data

**Describe the solution you'd like**
Add a flag to the bulk indexing operation that would trigger the commit operation and block execution until all data indexed so far is available to searchers.

**Describe alternatives you've considered**
An alternative is to reduce `commit_timeout_secs` and add sleep to the script. It is a viable solution if you need to run this script only once, but it is detrimental to quick iterations while debugging the script. 

**Additional context**
I found a relevant discussion in https://github.com/quickwit-oss/quickwit/issues/1674#issuecomment-1168220633 but since that issue is stalled and my motivation for this feature is quite different I would like to open a separate feature request.

fulmicoton:
It should be feasible in the pushapi at least...

imotov:
I will give it a shot.

guilload:
Let me give you some pointers then:
1. request somehow reaches the control plane*
2. control plane locates the relevant indexers and forwards them the request
3. some actor on the target node receives the message and forwards it to the indexer actor, which then finally commits the split (I'm voluntarily imprecise here to let you discover on your own the actors that compose an indexing pipeline)

\* we don't have a proper routing layer at the moment. That's my goal for February. On Kubernetes, our current solution is to route requests using ingress rules:

```
`api/v1/<index ID>/ingest*` → `indexers.<namespace>.svc.cluster.local`
`api/v1/<index ID>/search*` → `searchers.<namespace>.svc.cluster.local`
...
```

imotov:
[...]

Issue #2984, opened 2023-03-13
Add a quickwit-integration-tests crate


fmassot:
We started a bunch of "e2e" tests [there](https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-serve/src/tests.rs)

As we need to do more of them to cover the basic actions a user will do when using quickwit, we should probably isolate them and use the `quickwit-rest-client` for that.

Thus `quickwit-integration-tests`.

PR #2995, opened 2023-03-14
Add an option to force or wait for commit

imotov:
### Description

Adds an option to force a commit or wait for the next commit while issuing an ingest command. This commit only implements it for the ingest endpoint. If all records in the batch fail, the command will hang until at least one good document is indexed until #2826 is merged in or the document parsing is moved to the early stages.

See #2699

### How was this PR tested?

Added tests.


fulmicoton:
@trinity-1686a can you have a look?

trinity-1686a (commented):
It would be nice to have an e2e style integration test that ingest a few docs forcing commit and immediately search for them, but that'll possibly be best done once #2984 has been worked on

trinity-1686a (approved):

fmassot:
Let's wait for 0.5 release before merging.

imotov:
As much as release branches can be a pain, the balance of cost vs benefit of having them will only move towards the benefit as the team and customer base grow. Maybe be something to consider for 0.6?

fmassot:
@imotov absolutely, we are releasing tomorrow so we can merge your PR tomorrow.

PR #3010, opened 2023-03-14
QW-2972: Add validator to pulsar address

### LINKED ISSUE #2972: Ensure pulsar address is validated.
Currently, we require the pulsar address to start with `pulsar://` and conform with a URI, for a better dev exp we should add validation for this when creating the source.

ChillFish8:
Closes #2972 

Adds the missing validator check to ensure the pulsar address starts with `pulsar://`.

This adds about as much validation as is possible with the limited amount of documentation available.

### DIFF
--- a/quickwit/quickwit-config/src/source_config/mod.rs
+++ b/quickwit/quickwit-config/src/source_config/mod.rs
@@ -361,9 +361,25 @@ fn pulsar_uri<'de, D>(deserializer: D) -> Result<String, D::Error>
 where D: Deserializer<'de> {
     let uri: String = Deserialize::deserialize(deserializer)?;
 
-    if !uri.starts_with("pulsar://") {
-        return Err(Error::custom("Pulsar uri must start with `pulsar://`."))
-    }
+    let remaining = uri
+        .strip_prefix("pulsar://")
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-config/src/source_config/mod.rs
```diff
@@ -355,6 +356,23 @@ pub enum PulsarSourceAuth {
...
+fn pulsar_uri<'de, D>(deserializer: D) -> Result<String, D::Error>
+where D: Deserializer<'de> {
+    let uri: String = Deserialize::deserialize(deserializer)?;
+
+    let err_msg = format!(
+        "Invalid Pulsar uri provided, must be in the format of `pulsar://host:port/path`. Got: \
+         `{uri}`"
+    );
+
+    let _remaining = uri
```
fmassot, line 369: nitpick: an `if` condition would have been a bit clearer.

fmassot (approved):

quickwit/quickwit-config/src/source_config/mod.rs
ChillFish8: AH yes, originally this was used but then I remembered IPV6 was allowed which broke everything :(

Shall fix!

PR #3026, opened 2023-03-16
Minor changes meant to simplify the UX of tracing

fulmicoton:
- tracing commits every 5s by default
- removes QW_ENABLE_OTLP_ENDPOINT=true as it is the default
- we only talk about disabling that endpoint in one place and link that information somewhere else.
- adding a link to the tracing tutorial.
- Fixed instruction for linux

I could not test the changes so leaving as draft for the moment.

### DIFF
--- a/docs/log-management/index.md
+++ b/docs/log-management/index.md
@@ -66,6 +66,22 @@ The logs received by this endpoint are indexed on  the `otel-logs-v0` index. Thi
 
 You can also send your logs directly to this index by using the [ingest API](/docs/reference/rest-api.md#ingest-data-into-an-index).
 
+## OpenTelemetry service
+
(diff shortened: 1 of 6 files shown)


docs/log-management/index.md
```diff
@@ -166,7 +170,8 @@ search_settings:
 Currently, Quickwit provides a simplistic UI to get basic information from the cluster, indexes and search documents.
 If a simple UI is not sufficient for you and you need additional features, Grafana and Elasticsearch query API support are planned for Q2 2023, stay tuned!
 
-You can also send traces to Quickwit that you can visualize in Jaeger UI. Tutorial coming soon.
+You can also send traces to Quickwit that you can visualize in Jaeger UI, as explained in the following [tutorial](./distributed-tracing/instrument-python-and-send-traces-to-quickwit).
```
fulmicoton, line 173: We may have a better link after #3021 is merged

fmassot (approved):

PR #3034, opened 2023-03-16
Fix forward of ingest requests.

fmassot:


fulmicoton (approved):

quickwit/quickwit-serve/src/test_utils/rest_client.rs
```diff
@@ -42,6 +42,25 @@ impl QuickwitRestClient {
...
+            .header("content-type", "application/json")
+            .body(Body::from(ndjson_doc.to_string()))
+            .unwrap();
+        let response = self.client.request(request).await.unwrap();
+        if response.status() == StatusCode::OK {
+            return Ok(());
+        }
+        let body_bytes = hyper::body::to_bytes(response.into_body()).await.unwrap();
+        let body_string = String::from_utf8(body_bytes.to_vec()).unwrap();
+        Err(anyhow::anyhow!("error when creating index: {body_string}"))
```
fulmicoton, line 61:
```suggestion
        bail!("error when creating index: {body_string}")
```

fulmicoton (changes requested):
see change requests inline

quickwit/quickwit-serve/src/test_utils/rest_client.rs
fulmicoton, line 61:
```suggestion
        let response = self.client.request(request).await.context("Failed to emit request")?;
        if response.status() != StatusCode::OK {
            let body_bytes = hyper::body::to_bytes(response.into_body()).await.unwrap();
            let body_str = std::str::from_utf8_lossy(&body_bytes);
            bail!("error when creating index: {body_str}"));
        }
        Ok(())
```

super minor espeically considered it is in tests.

early return on success is a bit twisted.

fulmicoton:
Good job.

I tested the PR. It works as intended for forwarding the Rest ingest API, however it does not work for the Otel stuff.
I assume the OTEL gRPC is not running on the searcher or something like that. I'll open a separate issue.

fulmicoton (approved):

PR #3043, opened 2023-03-17
Fix index side bar width and index combobox.

fmassot:


quickwit/quickwit-ui/src/views/SearchView.tsx
```diff
@@ -66,8 +69,8 @@ function SearchView() {
     setSearchRequest(updatedSearchRequest);
     setQueryRunning(true);
     setSearchError(null);
+    navigate('/search?' + toUrlSearchRequestParams(updatedSearchRequest).toString());
```
fmassot, line 72: Update the URL once the "run" action is triggered, was bad to update it only if search request was successful.

quickwit/quickwit-ui/src/views/SearchView.tsx
```diff
@@ -42,6 +42,9 @@ function updateSearchRequestWithIndex(index: Index | null, searchRequest: Search
     searchRequest.startTimestamp = null;
     searchRequest.endTimestamp = null;
   }
+  if (index?.metadata.index_config.index_id) {
```
fmassot, line 45: this is the fix. We update the index ID.

PR #3062, opened 2023-03-21
Retry on flaky Kinesis test failure

guilload:
### Description
Localstack no longer supports `Kinesalite` and uses `kinesis-mock` by default, which is flaky. Since this test is critical, I'm adding retries to the coverage workflow.

### How was this PR tested?
In progress

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -32,9 +32,6 @@ jobs:
           - "4571:4571"
           - "8080:8080"
         env:
-          # `kinesalite` provides a more accurate implementation than
-          # the default Kinesis provider (`kinesis-mock`).
-          KINESIS_PROVIDER: kinesalite
           SERVICES: kinesis,s3
         options: >-
           --health-cmd "curl -k https://localhost:4566"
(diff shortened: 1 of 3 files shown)


.github/workflows/coverage.yml
```diff
@@ -32,9 +32,6 @@ jobs:
           - "4571:4571"
           - "8080:8080"
         env:
-          # `kinesalite` provides a more accurate implementation than
-          # the default Kinesis provider (`kinesis-mock`).
-          KINESIS_PROVIDER: kinesalite
```
guilload, line 37: Deprecated.

ChillFish8 (approved):
LGTM, shame they removed the more accurate implementation but at least this will catch other flaky tests as well in future.

.github/workflows/coverage.yml
```diff
@@ -131,17 +128,17 @@ jobs:
...
 
       - uses: Swatinem/rust-cache@v2
 
       - name: Generate code coverage
         run: |
           cargo llvm-cov clean --workspace
-          cargo llvm-cov --test failpoints --no-report --features fail/failpoints
-          cargo llvm-cov --no-report --all-features
+          cargo llvm-cov nextest failpoints --no-report --features fail/failpoints
+          cargo llvm-cov nextest --no-clean --no-report --all-features --retries 1
```
ChillFish8, line 141: It seems here we need to add `--workspace` here

PR #3063, opened 2023-03-21
QW-3054: Add a way to adjust OpenAPI routes sorting.

ChillFish8:
This adds the ability to adjust the ordering of how routes are displayed in swagger.

Swagger sorts the route in the order the tags are presented, so this ensures a stable order in which tags are defined so we can sort the routes.

Previously swagger got the tags from the routes themselves which were alphabetically sorted by the path.

### DIFF
--- a/docs/distributed-tracing/send-traces/using-otel-sdk-python.md
+++ b/docs/distributed-tracing/send-traces/using-otel-sdk-python.md
@@ -71,13 +71,14 @@ Let's first install the dependencies:
 ```bash
 pip install flask
 pip install opentelemetry-distro
+pip install opentelemetry-exporter-otlp
 ```
 
 The opentelemetry-distro package installs the API, SDK, and the opentelemetry-bootstrap and opentelemetry-instrument tools that you’ll use.
 
 Here is the code of our app:
 
-```python title=app.py
+```python title=my_app.py
 import random
 import time
 import requests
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-serve/src/openapi.rs
```diff
@@ -57,6 +57,21 @@ pub fn build_docs() -> utoipa::openapi::OpenApi {
...
+        Tag::new("Indexes"),
+        Tag::new("Ingest"),
+        Tag::new("Delete Tasks"),
+        Tag::new("Sources"),
+        Tag::new("Get Metrics"),
+        Tag::new("Cluster Info"),
+        Tag::new("Indexing"),
+        Tag::new("Splits"),
+        Tag::new("Node Health"),
+    ];
```
fulmicoton, line 72:
```suggestion
    let tags = vec![
        Tag::new("Search"),
        Tag::new("Indexes"),
        Tag::new("Ingest"),
        Tag::new("Delete Tasks"),
        Tag::new("Node Health"),
        Tag::new("Sources"),
        Tag::new("Get Metrics"),
        Tag::new("Cluster Info"),
        Tag::new("Indexing"),
        Tag::new("Splits"),
    ];
```

fulmicoton (approved):

Issue #3068, opened 2023-03-23
Cors options


joepio:
**Is your feature request related to a problem? Please describe.**
I'm building a front-end for QuickWit, and I'm trying to send a `POST` from a different origin. This triggers a CORS error in the client:

```
Access to fetch at 'http://localhost:7280/api/v1/stackoverflow/search' from origin 'http://localhost:5173' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

**Describe the solution you'd like**
Either default to adding an `Access-Control-Allow-Origin` or give the option to  set this. 

**Describe alternatives you've considered**
I tried adding [`mode: "no-cors"`](https://github.com/quickwit-oss/quickwit/blob/fa4c7e755887a97ea8ce5146f5574b6bee2e086f/quickwit/quickwit-ui/src/services/client.ts#L103) in my `fetch` request, but this prevents my browser to actually read the responses. 

**Additional context**
Running `9b40d05b6ac396ea5ebec698a03f97630237c636c984a4edea7edf97350dce10` docker image

joepio:
This is fixed, you can try in the `qw-cors` tagged docker image.

joepio:
We had a discussion on discord, basically letting it default to allow all is a danger if you're running quickwit on a private network, as it opens up the possibility of a cors attack.

PR #3070, opened 2023-03-23
Add CORS to REST API

ChillFish8:
Adds CORS to the rest api (currently allowing all origins) so users can use a dashboard in a browser from a different port than Quickwit.


joepio:
closes #3068

joepio:
Can confirm this works properly! Thank @ChillFish8  :)

fmassot (approved):
LGTM.

I would prefer to open an issue on this rather than write a comment that is harder to track.

PR #3083, opened 2023-03-24
QW-3077: Add CORS config

ChillFish8:
Adds the ability to configure the CORS change made earlier.

This denies all origins by default and can be optionally enabled with the new config:

```yaml
version: 0.5

rest_api:
  cors_allow_origins: *
  # cors_allow_origins: http://localhost:3000
  # cors_allow_origins:
  #   - http://localhost:3000
  #   - http://my.domain.com:6000
```


quickwit/quickwit-serve/src/args.rs
ChillFish8, line 1: So oddly while finding the right config I discovered this file is seemingly not used anywhere?

docs/configuration/index-config.md
```diff
@@ -5,12 +5,13 @@ sidebar_position: 2
 
 This page describes how to configure an index.
 
-In addition to the `index_id`, the index configuration lets you define five items:
+In addition to the `index_id`, the index configuration lets you define six items:
```
ChillFish8, line 8: Just noticing that this should probably still be 5 items?  Was this an error left over from previous changes?

quickwit/quickwit-serve/src/rest.rs
```diff
@@ -107,14 +109,24 @@ pub(crate) async fn start_rest_server(
...
+            cors = cors.allow_origin(tower_http::cors::Any);
+        } else {
+            info!(origins = ?cors_origins, "CORS is enabled, the following origins will be allowed");
+            let origins = cors_origins
+                .iter()
+                .map(|origin| origin.parse::<HeaderValue>().unwrap())
+                .collect::<Vec<_>>();
+            cors = cors.allow_origin(origins);
+        };
+    }
```
ChillFish8, line 129: Warp's CORS middleware did not seem to allow you to specify origins *and not* reject any request with an invalid origin. This is why we went with Tower so that we do enough to satisfy the browsers without interfering with the CLI or other applications.

docs/configuration/index-config.md
guilload, line 1: This is the index config doc page. You want the node config page.

ChillFish8: Good catch, will change now

quickwit/quickwit-config/src/quickwit_config/mod.rs
```diff
@@ -226,6 +226,18 @@ impl Default for JaegerConfig {
     }
 }
 
+#[serde_with::serde_as]
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, Default)]
+#[serde(deny_unknown_fields)]
+pub struct RestConfig {
+    #[serde(default)]
+    #[serde_as(deserialize_as = "serde_with::OneOrMany<_>")]
```
fulmicoton, line 234: cool! I could have used that recently.

fulmicoton:
Have you tested it?
Also is this something we can unit test?

ChillFish8:
> Have you tested it?

Yes, I think we can probably add this as a unit test with warps test client

ChillFish8:
I've added a unit test checking the behaviour of the cors middleware is as expected.

quickwit/quickwit-serve/src/rest.rs
```diff
@@ -229,3 +223,254 @@ fn get_status_with_error(rejection: Rejection) -> ApiError {
...
+
+    cors
+}
+
+#[cfg(test)]
```
fmassot, line 254: Nice tests.

fmassot (changes requested):
I tested the cors stuff with Quickwit UI, it worked so we are good on this.

I'm a bit concerned by the new configuration section `rest_api` that does not take into account the current parameter `rest_listen_port`.

There are several solutions here, let's start with 3:

1. Having a `rest` section

```yaml
rest:
  listen_port: 7280
  cors_allow_origins: *
```

We could add later `compression` and `max_content_length`  parameters.

2. Having a `cors` section

```yaml
rest_listen_port: 7280
rest_cors:
  allow_origins: *
```

We could add later more accurate parameters on cors stuff.

3.  Add `rest_cors_allow_origins` variable

```yaml
rest_listen_port: 7280
rest_cors_allow_origins: *
```

ChillFish8:
I decided to leave the rest port untouched for the following reasons:

- It's a breaking change if we go ahead and move the parameter
- Personally, I prefer having all of the grpc, gossip and rest ports next to each other in the config where it's easier to read and change, generally I don't particularly appreciate having them under separate sections for such a commonly configured option.
- The `rest_api` config in general gives us future expandability for adding things like authorization, etc... which will want to be configured.

fmassot:
> It's a breaking change if we go ahead and move the parameter
we can break things but not too often :)

> Personally, I prefer having all of the grpc, gossip and rest ports next to each other in the config where it's easier to read and change, generally I don't particularly appreciate having them under separate sections for such a commonly configured option.
I agree.

> The rest_api config in general gives us future expandability for adding things like authorization, etc... which will want to be configured.
I don't know where we will put authorization stuff, we will probably need it for grpc requests too.


I have a preference to name the section`rest` rather than `rest_api`: it's shorter and consistent with `rest_listen_port`.

guilload:
I vote for 1' or 3.
```yaml
rest_server:
  listen_port: 7280
  cors_allow_origins: *
  ...
  
grpc_server:
  listen_port: 7281
  ...
```

fmassot:
Let's do the 3rd then, I like the simplicity and I don't want to anticipate too much on the next parameters that we could add in a rest section (for example, auth stuff does not concern only the rest part).

@ChillFish8 can you update the PR?

fmassot (approved):
LGTM

PR #3091, opened 2023-03-26
File-backed metastore check connectivity with existence of indexes st…

### LINKED ISSUE #3088: File backed metastore check connectivity should not rely on `list_objects_v2` call
Currently, the file-backed metastore `check_connectivity` method calls the underlying storage `check_connectivity`, which calls 
rusoto method `list_objects_v2`. 

One goal of the file-backed metastore was to limit the usage of s3 API to maximize the compatibility with s3 compatible storage.

I suggest replacing the storage `check_connectivity` call with a simple `GET` request on the file `indexes_states.json` that is always there once a file-backed metastore is created. 

Update: actually, if no index has already been created, `indexes_states.json` does not exist.

fmassot:
…ates file.

### Description

Fix #3088 

The check connectivity is now using the existence of indexes states file. To make this work, the file-backed metastore is now creating this file in `try_new` if it does not exist.

### How was this PR tested?

Test added.

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/file_backed_metastore/store_operations.rs
+++ b/quickwit/quickwit-metastore/src/metastore/file_backed_metastore/store_operations.rs
@@ -73,6 +73,13 @@ fn convert_error(index_id: &str, storage_err: StorageError) -> MetastoreError {
     }
 }
 
+/// Check if `INDEXES_STATES_FILENAME` file exists.
+pub(crate) async fn check_indexes_states_exist(storage: Arc<dyn Storage>) -> anyhow::Result<()> {
+    let indexes_list_path = Path::new(INDEXES_STATES_FILENAME);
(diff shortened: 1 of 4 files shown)


fmassot:
@guilload can you take a look?

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/store_operations.rs
```diff
@@ -73,9 +73,16 @@ fn convert_error(index_id: &str, storage_err: StorageError) -> MetastoreError {
     }
 }
 
+/// Check if `INDEXES_STATES_FILENAME` file exists.
```
guilload, line 76:
```suggestion
/// Checks whether `INDEXES_STATES_FILENAME` file exists.
```

guilload (approved):

PR #3113, opened 2023-03-28
Add OpenAPI schema resolver unit test

ChillFish8:
This adds a unit test that attempts to resolve the paths of all schemas defined within the generated OpenAPI doc.

This comes out of how little validation Utoipa does to produce a correct schema, which is often missed during PRs and is very easy to break and very hard to notice unless you open every single request on the docs.

This also adds some other minor changes like making it clearer what the base path is for each request, and ensuring paths start with `/`

### DIFF
--- a/quickwit/quickwit-serve/src/openapi.rs
+++ b/quickwit/quickwit-serve/src/openapi.rs
@@ -130,6 +130,11 @@ impl OpenApiMerger for utoipa::openapi::OpenApi {
     fn with_path_prefix(mut self, prefix: &str) -> Self {
         let paths = mem::take(&mut self.paths.paths);
         for (path, detail) in paths {
+            if !path.starts_with('/') {
+                // We can panic here as I will be raised during unit tests.
+                panic!("Path {path:?} does not start with `/`.");
+            }
+
             let adjusted_path = if path != "/" {
                 format!("{prefix}{path}")
             } else {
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-serve/src/openapi.rs
```diff
@@ -122,11 +127,271 @@ impl OpenApiMerger for utoipa::openapi::OpenApi {
         }
     }
 
-    fn with_path_prefix(mut self, path: &str) -> Self {
-        for details in self.paths.paths.values_mut() {
-            details.servers = Some(vec![Server::new(path)]);
+    fn with_path_prefix(mut self, prefix: &str) -> Self {
+        let paths = mem::take(&mut self.paths.paths);
+        for (path, detail) in paths {
+            if !path.starts_with('/') {
```
fmassot, line 133: Can you use `assert!()` here?

fmassot (approved):
Please use `assert!` as suggested.

One note concerning the tests: while it's really nice to add them I'm a bit concerned about their maintainability. We will see how it goes with the upgrade of utoipa.
I let you judge if it's worth opening an issue to track that.

ChillFish8:
I thought about this too, but this is by far the more maintainable solution only having to ensure one thing is correct (and is backed by compiler errors) rather than relying on individual PRs to ensure that the docs don't break with no compile errors.

For reference, the docs have been broken 3 types due to the schema resolve error, unfortunately it's just very easy to break and very hard to detect during review.

fmassot:
> For reference, the docs have been broken 3 types due to the schema resolve error, unfortunately, it's just very easy to break and very hard to detect during the review.

Yes totally agree with that. We will see how it goes with utoipa upgrades.

PR #3118, opened 2023-03-29
Implement metrics Tower layer

guilload:
### Description
This PR aims to provide a Tower service that we can easily layer on a client or server to track the number of requests, errors, and latencies. The tricky part is labeling the metrics. Requests must implement the `PrometheusLabels` trait to provide labels. I added a macro to derive that trait. Let me know what you think.

### How was this PR tested?
Added unit tests


quickwit/quickwit-common/src/tower/metrics.rs
```diff
@@ -0,0 +1,245 @@
...
+            requests_total: self.requests_total.clone(),
+            request_errors_total: self.request_errors_total.clone(),
+            request_duration_seconds: self.request_duration_seconds.clone(),
+        }
+    }
+}
+
+/// Response future for [`PrometheusMetrics`].
+#[pin_project]
+pub struct ResponseFuture<const N: usize, F> {
```
guilload, line 135: `pin-project-lite` does not support const generics.

fulmicoton (approved):

fulmicoton:
Approved.
Nitpick: I think the macro to generate  `PrometheusLabels` implementation is overkill though.
I suggest we remove it and implement stuff manually if it gets in the way (hit on compilation times, or making things too opaque).

PR #3120, opened 2023-03-30
Add warnings on AWS credentials errors.

fmassot:


quickwit/quickwit-aws/src/lib.rs
```diff
@@ -98,11 +99,15 @@ impl ExtendedChainProvider {
 #[async_trait]
 impl ProvideAwsCredentials for ExtendedChainProvider {
     async fn credentials(&self) -> Result<AwsCredentials, CredentialsError> {
-        if let Ok(credentials) = self.web_identity_provider.credentials().await {
-            return Ok(credentials);
+        match self.web_identity_provider.credentials().await {
+            Ok(credentials) => return Ok(credentials),
+            Err(error) => {
+                warn!(err=?error, "Failed to find AWS credentials from `WebIdentityProvider`.")
```
guilload, line 105: We will always display this warning for users who rely on the "regular" `ChainProvider`. Let's store the error from the first provider and display it only if both authentication methods fail. We can also add debug statements to display which method was elected.

quickwit/quickwit-aws/src/lib.rs
```diff
@@ -98,12 +99,21 @@ impl ExtendedChainProvider {
...
+        let chain_provider = self.chain_provider.credentials().await;
+
+        match (identity_provider, chain_provider) {
+            (Ok(credentials), _) => return Ok(credentials),
+            (_, Ok(credentials)) => return Ok(credentials),
+            (Err(identity_error), Err(chain_error)) => {
+                error!(
+                    web_identity_error=?identity_error,
+                    chain_provider_error=?chain_error,
+                    "Failed to find AWS credentials from `WebIdentityProvider` or `ChainProvider`.",
```
fmassot, line 112:
```suggestion
                    "Failed to find AWS credentials from `WebIdentityProvider` and `ChainProvider`.",
```

quickwit/quickwit-aws/src/lib.rs
```diff
@@ -98,12 +99,21 @@ impl ExtendedChainProvider {
 #[async_trait]
 impl ProvideAwsCredentials for ExtendedChainProvider {
     async fn credentials(&self) -> Result<AwsCredentials, CredentialsError> {
-        if let Ok(credentials) = self.web_identity_provider.credentials().await {
-            return Ok(credentials);
-        }
-        if let Ok(credentials) = self.chain_provider.credentials().await {
-            return Ok(credentials);
+        let identity_provider = self.web_identity_provider.credentials().await;
+        let chain_provider = self.chain_provider.credentials().await;
```
guilload, line 103: Why resolve the second one if the first one succeeded?

```rust
let identity_provider_error = match self.web_identity_provider.credentials().await {
  Ok(credentials) => return Ok(credentials),
  Err(error) => error,
};
let chain_provider_error = match self.chain_provider.credentials().await {
  Ok(credentials) => return Ok(credentials),
  Err(error) => error,
};

error!(...)
Err(...)
```

fmassot: nice, this is better indeed

fmassot:
Closing as this should become obsolete with #3168

PR #3148, opened 2023-04-10
QueryAST and ElasticSearchDSL

fulmicoton:
Redesign the way we handle query.

- Introducing a QueryAST. It is the form used to send the query to the leaves.
- Added an ES compatible search API. It is now "truely" end-to-end compatible for a subset of the ES API. The query string and the body is used to build a quickwit SearchRequest. In particular the elastic query DSL is converted into the QueryAST.
- Query validation is only done against the last version of the doc  mapper. On the leaf side, this is only "best effort". For instance, searching on a field that used ot not be indexed will not return an error any more. The old splits will simply not return any results in that case. The idea is to allow docmapper change in the future.
- The semantics of default_fields and its interaction with JSON fields is changed. A default_fields used to have to be a Field in the sense of tantivy. A JSON field being default used to mean that we would use it as a potential prefix for all `field:term` like query. After this PR, _dynamic field has a special status. If present all `field:term`  are resolved against the dynamic field if field is not part of the schema. On the other hand, it is now possible to have a field within the dynamic field defined as a search default field.

Missing:
- [ ] complete/fix `quickwit-query/README.md`. Make it more like a Life of a Query
- [x] facet search (todo!())
- [x] bytes search  (todo!())
- [ ] range query on json field  (todo!())
- [ ] range query on bytes field   (todo!())
- [x] add unit tests in particular, test default fields logic.
- [x] re-add content limit in search body
- [ ] ditch query building from the docmapper.
- [ ] add the python REST-API level ES compatibility & non-regression test into quickwit's repo


quickwit/quickwit-query/src/query_ast/user_text_query.rs
```diff
@@ -0,0 +1,219 @@
...
+};
+use tantivy::schema::Schema;
+
+use crate::not_nan_f32::NotNaNf32;
+use crate::query_ast::tantivy_query_ast::TantivyQueryAst;
+use crate::query_ast::{self, BuildTantivyAst, QueryAst};
+use crate::{DefaultOperator, InvalidQuery, JsonLiteral};
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct UserTextQuery {
```
guilload, line 36: `UserInputQuery`?

quickwit/quickwit-query/src/query_ast/user_text_query.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// The default_search_fields argument on the other hand, is the default search fields defined
+    /// in the `DocMapper`.
+    pub fn parse_user_query(&self, default_search_fields: &[String]) -> anyhow::Result<QueryAst> {
+        let search_fields = self
+            .default_fields
+            .as_ref()
+            .map(|search_fields| &search_fields[..])
+            .unwrap_or(default_search_fields);
+        let user_input_ast = tantivy::query_grammar::parse_query(&self.user_text)
+            .map_err(|_| anyhow::anyhow!("Failed to parse query: `{}`.", &self.user_text))?;
```
guilload, line 65: .with_context

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -18,57 +18,45 @@
...
 use std::collections::{HashMap, HashSet};
+use std::convert::Infallible;
 
-use anyhow::{bail, Context};
-use quickwit_proto::SearchRequest;
-use tantivy::query::{Query, QueryParser, QueryParserError as TantivyQueryParserError};
-use tantivy::query_grammar::{UserInputAst, UserInputLeaf, UserInputLiteral};
-use tantivy::schema::{Field, FieldEntry, FieldType, Schema};
+use quickwit_query::query_ast::{QueryAst, QueryAstVisitor, RangeQuery};
+use tantivy::query::Query;
```
guilload, line 24: At this point, I wonder if we should systematically rename tantivy imports, at least in this crate and the query crate, with a `Tantivy*` prefix.

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -18,57 +18,45 @@
...
-    }
+/// Build a `Query` with field resolution & forbidding range clauses.
+pub(crate) fn build_query(
+    query_ast: &QueryAst,
+    schema: Schema,
+    search_fields: &[String],
+    with_validation: bool,
+) -> Result<(Box<dyn Query>, WarmupInfo), QueryParserError> {
+    let mut range_query_fields = RangeQueryFields::default();
+    range_query_fields.visit(query_ast).unwrap();
```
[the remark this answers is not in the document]
fulmicoton: TLDR the error is Infallible. unwrap_unfallible is still unstable :( ill adda comment.

quickwit/quickwit-doc-mapper/src/query_builder.rs
fulmicoton: agreed

fmassot:
One remark so that you are aware of that. The [endpoint](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-multi-search.html) `_msearch` is often used for UI (kibana, grafana plugin). I suspect this could be needed for the airmail project.

fmassot:
@fulmicoton one more thing to track. It's about the timestamp field. I don't know if in this PR, when you define a filter on the timestamp field, it will populate the `start_timestamp` end `end_timestamp` query attributes.

quickwit/quickwit-query/src/query_ast/user_text_query.rs
fulmicoton: It is not possible as the error is not StdErr.

quickwit/quickwit-proto/protos/quickwit/metastore_api.proto
```diff
@@ -186,9 +186,11 @@ message DeleteQuery {
   // If set, restrict search to documents with a `timestamp < end_timestamp``.
   optional int64 end_timestamp = 3;
   // Query text. The query language is that of tantivy.
-  string query = 4;
-  // Search fields.
-  repeated string search_fields = 5;
+  // Query AST serialized in JSON
+  string query_ast = 6;
+
+  // Reserved field due to deprecation
```
fmassot, line 192: Nitpick: given that all our breaking changes, I'm wondering if we really need to keep the deprecated reserved field.

quickwit/quickwit-proto/src/lib.rs
```diff
@@ -341,6 +344,49 @@ impl TryFrom<&str> for IndexingTask {
...
+///
+/// If it is not supplied, the docmapper search fields are meant to be used.
+pub fn query_ast_from_user_text(
+    user_text: &str,
+    default_fields: Option<Vec<String>>,
+) -> QueryAst {
+    quickwit_query::query_ast::UserInputQuery {
+        user_text: user_text.to_string(),
+        default_fields,
+        default_operator: quickwit_query::DefaultOperator::Or,
```
fmassot, line 385: I'm surprised by the default "OR", I thought we went for "AND".

fmassot (approved):

quickwit/quickwit-query/src/query_ast/bool_query.rs
```diff
@@ -0,0 +1,87 @@
...
+/// - named queries
+///
+/// Edge cases of BooleanQuery are not obvious,
+/// and different beahvior could be justified.
+///
+/// Here we aligne ourselves with ElasticSearch.
+/// A boolean query is to be interpreted like a filtering predicate
+/// over the set of documents.
+///
+/// If all clauses are empty, then the full set of documents is returned.
```
fmassot, line 38: <3

quickwit/quickwit-proto/protos/quickwit/metastore_api.proto
fulmicoton: it is absolutely useless at the moment but considering how immature we are about backward compatibility, I think a repetition of the good practise is a good idea.

quickwit/quickwit-proto/src/lib.rs
fulmicoton: Ah yes! It was done for elastic, but I can probably have a different behavior for elastic and quickwit.

PR #3153, opened 2023-04-11
update tantivy

PSeitz:
includes:
- percentiles aggregation
- seperate tokenizer for fastfield - This is mainly to allow lower case normalization on fast field data. Unfinished on quickwit side. (The datastructure is compatible with previous de/serialization of `true` and `false`) 
- remove required restriction for fast fields

### Description

### How was this PR tested?


quickwit/quickwit-search/src/collector.rs
```diff
@@ -473,14 +473,16 @@ fn merge_leaf_responses(
...
+            let mut fruit_iter = fruits.into_iter();
+            if let Some(first_fruit) = fruit_iter.next() {
+                let mut merged_fruit = first_fruit;
+                for fruit in fruit_iter {
+                    merged_fruit.merge_fruits(fruit)?;
+                }
+                Some(serde_json::to_string(&merged_fruit)?)
+            } else {
+                None
+            }
```
fulmicoton, line 485:
```suggestion
            let mut fruit_iter = fruits.into_iter();
            let mut merged_fruit = fruits_iter.next()?;
            for fruit in fruit_iter {
                    merged_fruit.merge_fruits(fruit)?;
            }
            Some(serde_json::to_string(&merged_fruit)?)
```

fulmicoton (approved):

quickwit/quickwit-search/src/collector.rs
PSeitz: the early exit is not possible here

Issue #3146, opened 2023-04-07
Is our Tokio runtime too loaded?


fulmicoton:

fulmicoton:
The branch bench-perf-investigation  fixes part of the problem, by moving some of the processing in the spawn_blocking pool. The staircase pattern between the start of the different leaf search task is now of 60microsecs which is probably acceptable.

The elephant in the room now is the the docstore. We decompress blocks in the tokio loop which is a catastrophy.

trinity-1686a:
this is a timeline of what I believe happens on the 1st half (leaf_search):
- `leaf_search_single_split(A)` starts, an IO get scheduled by `get_split_footer_from_cache_or_fetch(A)`
- `leaf_search_single_split(B)` starts, an IO get scheduled by get_split_footer_from_cache_or_fetch(B)`
- ... (repeat for more splits)
- all IO are ready roughly at the same time
- `get_split_footer_from_cache_or_fetch(A)` finishes, index gets opened. `warmup(A)` schedules an IO
- `get_split_footer_from_cache_or_fetch(B)` finishes, index gets opened. `warmup(B)` schedules an IO
- ... (repeat for more splits)
- IO starts getting completed
- `warmup(A)` finishes,, off-pool `tantivy-search(A)` begin
- `warmup(B)` finishes,, off-pool `tantivy-search(B)` begin

`leaf_search_single_split` starting in staircase is to be expected, each job gets started after the previous scheduled its first IO. There isn't really a point yielding before there: it will just delay the time before the 1st IO starts, but not impact the time before the last IO does.
The staircase of `warmup` does look bad: each IO is ready roughly at the same time, but after an IO from `get_split_footer_from_cache_or_fetch` get completed and before `warmup` finish scheduling its IOs, about 300µs are spent doing cpu work. This delays the end of `get_split_footer_from_cache_or_fetch` for the next split, which then itself spend 300µs before yielding to a 3rd split... Again I don't think actually yielding would change much: each IO would finish more quickly, but warmup from the 1st split would then start just after the last split finished opening its index, and last split warmup would start same time as before.

I don't think this should be interpreted as a runtime being too loaded, but rather a single worker being alternatively waiting for many IOs (which is fine), and crowded from many joined futures doing not-that-long things individually, but which all needs cpu at the same time.
[...]

PR #3162, opened 2023-04-11
fix some issues found while investigating search latency

trinity-1686a:
### Description

related to #3146 
spawn a few futures, instrument some functions and use the cpu_intensive pool instead of tokio blocking pool

### How was this PR tested?
doing queries on a local node and seeing `elapsed_time_micros` is much smaller

### DIFF
--- a/quickwit/quickwit-search/src/fetch_docs.rs
+++ b/quickwit/quickwit-search/src/fetch_docs.rs
@@ -18,10 +18,11 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::collections::{BTreeMap, HashMap};
+use std::future::Future;
 use std::sync::Arc;
 
 use anyhow::{Context, Ok};
-use futures::stream::{StreamExt, TryStreamExt};
+use futures::StreamExt;
 use itertools::Itertools;
 use quickwit_doc_mapper::DocMapper;
 use quickwit_proto::{FetchDocsResponse, PartialHit, SearchRequest, SplitIdAndFooterOffsets};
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-search/src/fetch_docs.rs
```diff
@@ -240,8 +242,37 @@ async fn fetch_docs_in_split(
...
+{
+    assert!(max_running > 0);
+
+    let mut results = Vec::new();
+    let mut future_set = futures::stream::FuturesUnordered::new();
+    for future in (&mut futures).take(max_running) {
+        future_set.push(tokio::spawn(future.in_current_span()));
+    }
+
+    while let Some(result) = future_set.next().await {
```
fulmicoton, line 268: You might as well use 
https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffer_unordered

directly I think...

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -402,21 +407,25 @@ pub async fn leaf_search(
         Vec<(String, SearchError)>,
     ) = split_search_results
         .into_iter()
-        .partition_map(|split_search_res| match split_search_res {
-            Ok(split_search_resp) => Either::Left(Ok(split_search_resp)),
-            Err(err) => Either::Right(err),
-        });
+        .partition_map(
+            |split_search_res| match split_search_res.expect("spawned future died") {
```
fulmicoton, line 411: I understand that this is the same behavior as before, but could we make that expect into a SearchError sayin the taks pannicked or was cancelled?

fulmicoton (approved):

quickwit/quickwit-search/src/fetch_docs.rs
trinity-1686a: hum, actually you are right, iter are lazy so `map()` won't call `tokio::spawn` until `buffer_unordered` tries to schedule more tasks

PR #3164, opened 2023-04-12
Improve helm deployment docs

### LINKED ISSUE #3163: Add a few clarifications to the helm deployment instructions
I had some difficulties while trying to follow the [helm deployment instructions](https://quickwit.io/docs/deployment/kubernetes) on my local cluster until I realized that PostgreSQL is not included in the helm chart, and it is not listed in requirements either. The `image.tag` value in the `values.yml` file is also pointing to an old version of quickwit, which is incompatible with the latest version of helm charts.

imotov:
### Description

Adds PostgreSQL as a requirement and fixes obsolete image.tag form value.yaml example.

Fixes #3163

### How was this PR tested?

By following instruction on my local Kubernetes cluster.


docs/deployment/kubernetes.md
```diff
@@ -16,41 +16,49 @@ To deploy Quickwit on Kubernetes, you will need:
 - kubectl, compatible with your cluster (+/- 1 minor release from your cluster) (`kubectl version`)
 - Helm v3 (`helm version`)
 - A Kubernetes cluster
+- An instance of PostgreSQL server
```
fmassot, line 19: You can have a quickwit cluster without postgresql server, a file backed metastore on S3 works. The current helm chart needs to be clearer on that, there is an open PR to ease the setup of the S3 metastore https://github.com/quickwit-oss/helm-charts/pull/43

```suggestion
- An instance of PostgreSQL server or an S3-backed metastore
```

fmassot: For now, you have to set QW_METASTORE_URI env variable to make it work with an S3 metastore.

imotov: Hmm. That will simplify things quite a bit at the expense of performance. Let me play with it. Maybe we can add file store as the main path and move postgresql to the end as a scalable but more complicated alternative.

docs/deployment/kubernetes.md
```diff
@@ -82,10 +66,29 @@ helm install <deployment name> quickwit/quickwit -f values.yaml
...
 
-It might take some time for the cluster to start. During the startup process individual pods might restart themselves. The command that you typed on previous step should print out instructions on how to set port forwarding to connect to the cluster. While this command is running, you can access the cluster UI by navigating to http://127.0.0.1:8080 in your browser. You can also check the version of Quickwit by running:
+It might take some time for the cluster to start. During the startup process individual pods might restart themselves several times. The command on the previous step will print the instructions on how to connect to the cluster. This endpoint can be used to access the quickwit search UI, as well execute standard API commands against.
 
-```bash
-curl http://127.0.0.1:8080/api/v1/version
+## Using PostgreSQL as a metadata store
+
+The file-backed metastore is mainly useful for testing purposes. Though a file-backed metastore might be easier to setup, we strongly encourage you to use a PostgreSQL metastore in production. For the quickwit installation to work with PostgreSQL metadata you need to provide connection PostgreSQL information instead of metastore url:
```
fmassot, line 73:
```suggestion
The file-backed metastore is mainly useful for testing purposes. Though a file-backed metastore might be easier to setup, we strongly encourage you to use a PostgreSQL metastore in production. For the quickwit installation to work with PostgreSQL metadata you need to provide connection PostgreSQL information instead of metastore URI:
```

fmassot (approved):
Nice PR, thanks.

PR #3168, opened 2023-04-12
QW-885: Move to AWS SDK

ChillFish8:
Moves from the now unmaintained Rusto library to the official AWS SDK.

### TODO List

- [x] Migrate `quickwit-aws` crate
- [x] Move across S3 handling
- [x] Fix Kinesis 
- [x] Fix unit tests
- [x] Check semantics with other providers (Google storage, B2, MinIO, is the same as before)
- [x] Replug the different authentication methods
- [x] Update documentation in https://quickwit.io/docs/guides/aws-setup#aws-credentials
- [x] Test virtual path on s3


quickwit/quickwit-aws/src/lib.rs
```diff
@@ -19,94 +19,55 @@
...
-    fn new() -> Self {
-        let mut chain_provider = ChainProvider::new();
-        chain_provider.set_timeout(CREDENTIALS_FETCH_TIMEOUT);
-        let web_identity_provider = WebIdentityProvider::from_k8s_env();
-        Self {
-            chain_provider,
-            web_identity_provider,
-        }
-    }
+            aws_config::from_env()
```
fmassot, line 60: Did you check if it covers the k8s case? (it seems so but just want you to double check). We could add a task such as "test k8s credentials on k8s cluster" in the to-do list.

fmassot:
Sounds like a good start. One thing you can add to the to-do list is to test virtual-hosted–style requests.

quickwit/quickwit-storage/src/error.rs
```diff
@@ -47,6 +49,10 @@ pub enum StorageResolverError {
     /// A protocol is required for the URI.
     #[error("Invalid format for URI: required: `{message}`")]
     InvalidUri { message: String },
+    #[error("The system does not have a initialised AWS config available")]
+    /// This should never happen if the `quickwit_aws::try_init_aws_config` method
+    /// has been called, if this error occurs this is a bug.
+    MissingAWSConfig,
```
fmassot, line 55: Can we rename `MissingAWSConfig`? I think it concerns any s3 compatible storage. 
And I'm a bit confused here, the issue is not really that the config is missing right? I think we should be clearer and say something like "S3 compatible storage configuration has not been initialized. This should never happen".

quickwit/quickwit-aws/src/lib.rs
ChillFish8: I haven't checked, what system does it normally use? (I.e. ENV vars, etc...)

fmassot: I think you work on it in the PR related to credentials errors (web identity provider etc.). But the defaults seem quite nice so hopefully we don't have much to do there.

ChillFish8:
Marking as ready just to try get CI running

fulmicoton:
@fmassot can we land this?

quickwit/quickwit-storage/src/error.rs
guilload: I got rid of it altogether.

fmassot (approved):

Issue #3147, opened 2023-04-07
`quickwit index delete` does not wait for confirmation before deleting an index


trinity-1686a:
**Describe the bug**
When running `quickwit index delete`, a confirmation prompt appear, but the index is deleted before the prompt is answered

**Steps to reproduce (if applicable)**

1. run `quickwit run`. Open a new terminal
1. run `quickwit index create --index-config ./datasets/index_config/gh-archive.yaml`
2. run `quickwit index delete --index gh-archive`, answer no, or don't answer a open a new terminal
3. run `quickwit index describe --index gh-archive` => ``Command failed: Api error: (code=404 Not Found, message=Index `gh-archive` does not exist.)``

**Expected behavior**
Deletion to not happen **before** answering the prompt

**Configuration:**
Please provide:

1. Output of `quickwit --version`: `Quickwit 0.5.0-nightly (a68c012 2023-04-07T14:38:01Z)`
2. The [index_config.yaml](https://raw.githubusercontent.com/quickwit-oss/quickwit/main/config/tutorials/gh-archive/index-config.yaml)

related to #2201

PR #3175, opened 2023-04-13
Fix the delete index user confirmation

ph:
### Description

Ensure that when a user answer `no` the index will not be deleted, this is done by moving the actual delete logic after confirming with the user.

### How was this PR tested?

The test suite was run locally and the following command were run to confirm the changes in behavior.

```sh
cargo r index create --index-config ../config/tutorials/hdfs-logs/index-config.yaml

cargo r index create --index-config ../config/tutorials/hdfs-logs/index-config.yaml # Answered no.

cargo r index describe --index hdfs-logs # return the information about the index.

cargo r index create --index-config ../config/tutorials/hdfs-logs/index-config.yaml # Answered yes.

cargo r index describe --index hdfs-logs # return 404 not found
```

I've looked at the test suite, we seem to run all the CLI test with `assume_yes` and skipping any prompt. 
Maybe we should also validate the confirmation flow, I am not sure  how to add it to the test suite, I am not familiar enough with the code base.

Fixes: #3147

### DIFF
--- a/quickwit/quickwit-cli/src/index.rs
+++ b/quickwit/quickwit-cli/src/index.rs
@@ -866,6 +866,13 @@ pub async fn search_index_cli(args: SearchIndexArgs) -> anyhow::Result<()> {
 pub async fn delete_index_cli(args: DeleteIndexArgs) -> anyhow::Result<()> {
     debug!(args=?args, "delete-index");
     println!("❯ Deleting index...");
+    if !args.dry_run && !args.assume_yes {
+        let prompt = "This operation will delete the index. Do you want to proceed?".to_string();
+        if !prompt_confirmation(&prompt, false) {
+            return Ok(());
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


fmassot:
Thanks @ph for your PR. Seems good, let me test it before merging.
We don't test the prompts unfortunately :/. I will open an issue dedicated to that because we will certainly break this again.

quickwit/quickwit-cli/src/index.rs
```diff
@@ -866,6 +866,13 @@ pub async fn search_index_cli(args: SearchIndexArgs) -> anyhow::Result<()> {
 pub async fn delete_index_cli(args: DeleteIndexArgs) -> anyhow::Result<()> {
     debug!(args=?args, "delete-index");
     println!("❯ Deleting index...");
```
guilload, line 868: Can you move this line after the user has confirmed?

guilload (approved):

fmassot:
All good, it will be merged in a few minutes. Thanks!

Issue #2875, opened 2023-02-23
Using a random node ID on startup leads to a suboptimal experience


guilload:
Merges are fenced by node ID. When stopping and resuming a server, the node ID changes and merges between "indexing sessions" do not happen. I fixed my local setup by messing with the node ID property of splits in the metastore.

We may want to use the hostname as the default node ID.

guilload:
Issues with this change:
- detect collisions at the instance level (multiple nodes running on the same machine in dev mode) 
- detect collisions at the cluster level
- allow characters such as `.` in node ID ([Valid characters of a hostname?](https://stackoverflow.com/questions/3523028/valid-characters-of-a-hostname))

Idea:
Use node ID as prefix in QW dir: `qwdata/<node ID>/...`

PR #3181, opened 2023-04-14
Use short hostname as default node ID

guilload:
### Description
Use hostname as default node ID. The problem I'm trying to solve is described in the following [issue](https://github.com/quickwit-oss/quickwit/issues/2875).

### How was this PR tested?
In progress:
- [x] unit test
- [x] macOS
- [x] linux
- [x] Kube?
- [x] Docker

### DIFF
--- a/quickwit/quickwit-common/src/net.rs
+++ b/quickwit/quickwit-common/src/net.rs
@@ -17,7 +17,9 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+use std::ffi::OsString;
 use std::fmt::Display;
+use std::io;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, TcpListener};
 use std::str::FromStr;
 
@@ -222,6 +224,7 @@ pub fn find_private_ip() -> Option<(String, IpAddr)> {
     _find_private_ip(&datalink::interfaces())
 }
 
+// Inner function for testing purposes.
 fn _find_private_ip(interfaces: &[NetworkInterface]) -> Option<(String, IpAddr)> {
     // The way we do this is the following:
     // 1. List the network interfaces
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-common/src/net.rs
```diff
@@ -330,7 +330,7 @@ fn is_private_ip(ip_addr: &IpAddr) -> bool {
 /// - Its labels (characters separated by `.`) are not empty.
 /// - Its labels are 63 or fewer characters.
 /// - Its labels do not start or end with '-' or '.'.
-fn is_valid_hostname(hostname: &str) -> bool {
+pub fn is_valid_hostname(hostname: &str) -> bool {
```
imotov, line 333: I wonder if it makes more sense to chop off the part after `.` instead of allowing FQDNs here.

imotov (commented):
Looks good to me in general. 

I tested it on macOS, linux and using helm. In my environment macOS gave me FQDN while linux returned just the hostname, which is consistent with return value of `hostname` command, but I wonder if it makes sense to just alway use hostname without the domain part. In my experience `hostname` output can be a bit finicky on macs.

I didn't observe any difference while using helm chart since we are using pod name as the node name, but it works, so you can probably check that off.

I think it would be also great to update [docs](https://github.com/quickwit-oss/quickwit/blob/a68c0122cf3e4189b76ae66f7be27516718bc346/docs/configuration/node-config.md?plain=1#L20) to reflect that it is no longer that random.

fmassot:
I made some tests on EKS. Just to confirm the expected behavior, the node ID was set to the pod name.

On docker, by default the hostname is set to the container ID. If you pass the arg `-h`, you can set the hostname and then Quickwit will have the node ID set to this hostname.

```bash
docker run -d -h test-1 quickwit/quickwit:qw-hostname run
```

leads to

```bash

2023-04-18T20:48:02.850Z  INFO quickwit_cli: Loaded Quickwit config. config_uri=file:///quickwit/config/quickwit.yaml config=QuickwitConfig { cluster_id: "quickwit-default-cluster", node_id: "test-1", enabled_services: {...}...

```

quickwit/quickwit-common/src/net.rs
guilload: In cloud environments, a common practice is to set the hostname to `<instance ID>.<az>.<region>.<azure|aws|...>.<company domain>.`, so it might be nice to actually allow FQDNs.

imotov: That's a good point. I don't have strong feelings about that, but besides consistency with linux standalone nodes that only return hostname, I think long node ids might hunt us later when we will want to show them in different management UIs.

imotov: 👍 Than my only concern is the doc updates here.

imotov (approved):
Re-tested it on Mac, works fine now giving only a short name. Thanks!

PR #3198, opened 2023-04-19
add lowercase tokenizer, document fast field tokenizer

### LINKED ISSUE #3126: add and document tokenizer option on fastfields
After https://github.com/quickwit-oss/tantivy/pull/1945, we can set tokenizers on fastfields. 

project airmail uses this for lowercase transformation. So we also need a lowercase tokenizer

### LINKED ISSUE #3158: Document fast field tokenizer

PSeitz:
### Description

add lowercase tokenizer
document fast field tokenizer

### How was this PR tested?
unit test for serialization
locally tested config with custom tokenizer

closes #3126 closes #3158

### DIFF
--- a/docs/configuration/index-config.md
+++ b/docs/configuration/index-config.md
@@ -136,7 +136,7 @@ fieldnorms: true
 | `default`     | Chops the text on according to whitespace and punctuation, removes tokens that are too long, and converts to lowercase |
 | `en_stem`     |  Like `default`, but also applies stemming on the resulting tokens  |
 | `chinese_compatible` |  Chop between each CJK character in addition to what `default` does. Should be used with `record: position` to be able to properly search |
-| `lowercase` |  Applies a lowercase transformation on the text. |
+| `lowercase` |  Applies a lowercase transformation on the text. It does not tokenize the text. |
 
 **Description of record options**
 
--- a/docs/configuration/index-config.md
+++ b/docs/configuration/index-config.md
@@ -40,7 +40,7 @@ doc_mapping:
       type: text
       tokenizer: raw
       fast: 
-        - with_tokenizer: lowercase
+        - tokenizer: lowercase
     - name: body
       type: text
       tokenizer: default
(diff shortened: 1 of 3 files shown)


fulmicoton:
Should we differentiate tokenizers from normalizing analyzers?
For instance, add a `_normalize` suffix
lower_case -> lowercase_normalize.

docs/configuration/index-config.md
```diff
@@ -134,6 +136,7 @@ fieldnorms: true
 | `default`     | Chops the text on according to whitespace and punctuation, removes tokens that are too long, and converts to lowercase |
 | `en_stem`     |  Like `default`, but also applies stemming on the resulting tokens  |
 | `chinese_compatible` |  Chop between each CJK character in addition to what `default` does. Should be used with `record: position` to be able to properly search |
+| `lowercase` |  Applies a lowercase transformation on the text. |
```
fulmicoton, line 139: We need to mention explicitly this tokenizer does not tokenize.

fulmicoton (approved):

PSeitz:
> Should we differentiate tokenizers from normalizing analyzers? For instance, add a `_normalize` suffix lower_case -> lowercase_normalize.
> 

The way as it is used now does the job, but the API is not great.
Ideally the user could pick normalizer and tokenizer separately. Mixing them is confusing.

I would not add `_normalize` suffix, because we may have later a normalizer category.

> We could also consider having two tokenizer manager. One for indexing (with tokenizing + normalizing tokenizers) and one for fast fields (with only normalizing tokenizers)

We would need to support that in tantivy. But that sounds more like we want to support for tokenizer and normalizer in tantivy.

docs/configuration/index-config.md
PSeitz: done

fmassot:
@PSeitz is it possible to use `tokenizer` instead of `with_tokenizer`?

We could also use (not now) the term `analyzer`, an analyzer being a combination of a tokenizer and filters. I think this would help.

PSeitz:
> @PSeitz is it possible to use `tokenizer` instead of `with_tokenizer`?
> 
> We could also use (not now) the term `analyzer`, an analyzer being a combination of a tokenizer and filters. I think this would help.

yes, I'll change it to `tokenizer`. `analyzer` now would also be fine, with the restriction that for now only tokenizer are available

Issue #3073, opened 2023-03-23
Quickwit fails to handle SIGTERM inputs.


ChillFish8:
Services like Docker and K8s send a `SIGTERM` signal to the containers to gracefully shut down.

When benchmarking we observe that Quickwit does not do this and instead continues until it is forcefully killed.

PR #3200, opened 2023-04-19
Add a quickwit-integration-tests crate

### LINKED ISSUE #2984: Add a quickwit-integration-tests crate

imotov:
### Description

Fixes #2984

### How was this PR tested?

The whole PR is basically a test infrastructure.


quickwit/quickwit-integration-tests/src/test_utils/test_client.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub fn new(addr: SocketAddr) -> Self {
+        let client = hyper::Client::builder()
+            .pool_idle_timeout(Duration::from_secs(30))
+            .http2_only(true)
+            .build_http();
+        let root_url = format!("http://{addr}");
+        Self { root_url, client }
+    }
+
+    pub async fn is_live(&self) -> anyhow::Result<bool> {
```
fmassot, line 42: Is there a particular reason to not put these endpoints in the rest client?

imotov: I had a reason, but now I see it was a wrong reason 😄 . Somehow, I thought these endpoints are enabled only during a test run, but now I see that I misinterpreted this part. I will fix that.

quickwit/quickwit-cli/src/service.rs
```diff
@@ -80,7 +81,12 @@ impl RunCliCommand {
         quickwit_telemetry::send_telemetry_event(telemetry_event).await;
         // TODO move in serve quickwit?
         start_actor_runtimes(&config.enabled_services)?;
-        serve_quickwit(config).await?;
+        let _ = serve_quickwit(config, async move {
```
fmassot, line 84: I think we should take into account other signals (I had this [kind of stuff in mind](https://github.com/vectordotdev/vector/blob/68b54a9bc0ae07d916ec48e997a03f7681e54ccc/src/signal.rs#L44)
I suggest to do that in another PR so that you can advance on your original issue :)

fmassot: There is an issue related to this https://github.com/quickwit-oss/quickwit/issues/3073

fmassot (approved):
We now have this integration tests project, that's really cool, thanks for taking care of this issue. I want to review the shutdown signals but let's do it in another PR.

PR #3204, opened 2023-04-20
check for executor liveliness

trinity-1686a:
### Description

related to #3146 
this adds a "busy detector" which logs a warn when the main runtime doesn't park itself for too long. I tried to instead spawn a future on each threads, sadly it proved impossible to pin futures to each threads. [`LocalSet`](https://docs.rs/tokio/latest/tokio/task/struct.LocalSet.html) isn't as flexible as I'd thought. This implementation can then claim the runtime is overloaded when the task(s) overloading it actually yield often enough, but are collectively always doing something

this also adds the possibility to use a patched version of tracing which records the duration of the longest poll (or rather, entered section) for a span as `max_pool_ns`, and logs when a poll starts and end at DEBUG level, so it's possible to see more precisely when a future is running vs waiting when looking at an otlp trace

### How was this PR tested?

made a search query requesting many document and verified the logs appears as expected. Looked at the traces in jaeger and verified they looks as expected too.
image


quickwit/quickwit-cli/src/main.rs
```diff
@@ -113,50 +113,106 @@ fn setup_logging_and_tracing(
...
-    );
+fn main() -> anyhow::Result<()> {
+    tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .on_thread_unpark(busy_detector::thread_unpark)
+        .on_thread_park(busy_detector::thread_park)
+        .build()
+        .unwrap()
+        .block_on(async {
+            #[cfg(feature = "openssl-support")]
```
fulmicoton, line 124: I think we can move that to a function `main_impl` or something

quickwit/quickwit-cli/src/main.rs
```diff
@@ -113,50 +113,106 @@ fn setup_logging_and_tracing(
...
+                    let id = my_id.load(Ordering::Relaxed);
+                    if id == 0 {
+                        let new_id = NEXT_ID.fetch_add(1, Ordering::Relaxed);
+                        my_id.store(new_id, Ordering::Relaxed);
+                        new_id
+                    } else {
+                        id
+                    }
+                });
+                warn!("Thread{id} wasn't parked for {delta}µs, is the runtime too busy?");
```
fulmicoton, line 212: We might want some mechanism to avoid logging this more than once in 30s or something.
or make it a counter?

quickwit/Cargo.toml
```diff
@@ -238,3 +238,11 @@ encoding_rs = "=0.8.29"
 
 [patch.crates-io]
 sasl2-sys = { git = "https://github.com/quickwit-oss/rust-sasl/", rev = "daca921" }
+
+#tracing = { git = "https://github.com/trinity-1686a/tracing.git", branch = "tracing-subscriber-0.3.16-poll-duration" }
```
fulmicoton, line 242: do we want to keep those?

quickwit/quickwit-cli/src/main.rs
```diff
@@ -159,6 +168,87 @@ async fn main() -> anyhow::Result<()> {
...
+    use std::time::Instant;
+
+    use once_cell::sync::Lazy;
+    use tracing::warn;
+
+    // we need that time reference to use an atomic and not a mutex for LAST_UNPARK
+    static TIME_REF: Lazy<Instant> = Lazy::new(Instant::now);
+
+    const ALLOWED_DELAY_MICROS: u64 = 5000;
+    const WARN_SUPPRESSION_DURATION: u64 = 30_000_000;
```
fulmicoton, line 182:
```suggestion
    const WARN_SUPPRESSION_MICROS: u64 = 30_000_000;
```

quickwit/quickwit-cli/src/main.rs
```diff
@@ -159,6 +168,87 @@ async fn main() -> anyhow::Result<()> {
...
+
+    thread_local!(static MY_ID: AtomicU64 = AtomicU64::new(0));
+    static NEXT_ID: AtomicU64 = AtomicU64::new(1);
+
+    // LAST_UNPARK and NEXT_WARN are semantically timestamps
+    thread_local!(static LAST_UNPARK: AtomicU64 = AtomicU64::new(0));
+    static NEXT_WARN: AtomicU64 = AtomicU64::new(0);
```
fulmicoton, line 189:
```suggestion
    static NEXT_WARN_TIMESTAMP: AtomicU64 = AtomicU64::new(0);
```

quickwit/quickwit-cli/src/main.rs
```diff
@@ -159,6 +168,87 @@ async fn main() -> anyhow::Result<()> {
...
+    static SUPPRESSED_WARN: AtomicU64 = AtomicU64::new(0);
```
fulmicoton, line 190:
```suggestion
    static SUPPRESSED_WARN_COUNT: AtomicU64 = AtomicU64::new(0);
```

quickwit/Cargo.toml
trinity-1686a: yes. This is what provides the changes on traces (visible on Jaeger). I should probably move the fork to `quickwit-oss/` though

quickwit/quickwit-cli/src/main.rs
```diff
@@ -159,6 +168,87 @@ async fn main() -> anyhow::Result<()> {
...
+    use tracing::warn;
+
+    // we need that time reference to use an atomic and not a mutex for LAST_UNPARK
+    static TIME_REF: Lazy<Instant> = Lazy::new(Instant::now);
+
+    const ALLOWED_DELAY_MICROS: u64 = 5000;
+    const WARN_SUPPRESSION_DURATION: u64 = 30_000_000;
+
+    thread_local!(static MY_ID: AtomicU64 = AtomicU64::new(0));
+    static NEXT_ID: AtomicU64 = AtomicU64::new(1);
```
fulmicoton, line 185: NEXT_THREAD_ID

quickwit/quickwit-cli/src/main.rs
```diff
@@ -159,6 +168,87 @@ async fn main() -> anyhow::Result<()> {
...
+    use once_cell::sync::Lazy;
+    use tracing::warn;
+
+    // we need that time reference to use an atomic and not a mutex for LAST_UNPARK
+    static TIME_REF: Lazy<Instant> = Lazy::new(Instant::now);
+
+    const ALLOWED_DELAY_MICROS: u64 = 5000;
+    const WARN_SUPPRESSION_DURATION: u64 = 30_000_000;
+
+    thread_local!(static MY_ID: AtomicU64 = AtomicU64::new(0));
```
fulmicoton, line 184: If I understand correctly, MY_ID/NEXT_ID are here to assign a thread id to our thread. Why don't we just use the thread id?

trinity-1686a: Getting the inner id is unstable, and `ThreadId` Debug representation is `ThreadId(1)` which isn't very pretty

fulmicoton (approved):

quickwit/Cargo.toml
fulmicoton: thanks for adding the comment

fulmicoton (approved):

PR #3210, opened 2023-04-21
Fix rate estimator bucket race condition

guilload:
### Description
Per title.

### How was this PR tested?
- Added some test cases

### DIFF
--- a/quickwit/quickwit-common/src/tower/rate_estimator.rs
+++ b/quickwit/quickwit-common/src/tower/rate_estimator.rs
@@ -60,8 +60,8 @@ impl SmaRateEstimator {
         assert!(period.as_millis() > 0);
 
         let mut buckets = Vec::with_capacity(num_buckets.get());
-        for bucket_ord in 0..num_buckets.get() {
-            buckets.push(Bucket::new(bucket_ord as u64, 0));
+        for _ in 0..num_buckets.get() {
+            buckets.push(Bucket::default());
         }
         let inner = InnerSmaRateEstimator {
             anchor: Instant::now(),
@@ -81,8 +81,8 @@ impl SmaRateEstimator {
     pub fn with_initial_rate(self, initial_rate: ConstantRate) -> Self {
         let initial_work = initial_rate.work() * self.inner.bucket_period_millis
             / initial_rate.period().as_millis() as u64;
-        for i in 0..self.inner.num_buckets {
-            self.inner.buckets[i as usize].update(initial_work);
+        for bucket_ord in 0..self.inner.num_buckets {
+            self.inner.buckets[bucket_ord as usize].increment_work(initial_work, 0);
         }
         self
     }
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-common/src/tower/rate_estimator.rs
```diff
@@ -119,59 +124,57 @@ impl RateEstimator for SmaRateEstimator {
...
-    fn cumulated_work(&self) -> u64 {
-        self.cumulated_work.load(Ordering::Relaxed)
+    fn work(&self) -> u64 {
+        self.bits.load(Ordering::Relaxed) & Self::WORK_MASK
     }
 
-    fn update(&self, work: u64) {
-        self.cumulated_work.fetch_add(work, Ordering::Relaxed);
-    }
+    fn increment_work(&self, work: u64, bucket_ord: u64, num_buckets: u64) {
```
guilload, line 153: This should be `self.bits.fetch_add(work & Self::WORK_MASK, Ordering::Relaxed)` but that seems overkill.

quickwit/quickwit-common/src/tower/rate_estimator.rs
```diff
@@ -119,59 +124,57 @@ impl RateEstimator for SmaRateEstimator {
...
 }
 
 impl Bucket {
-    fn new(bucket_ord: u64, cumulated_work: u64) -> Self {
-        Self {
-            bucket_ord: AtomicU64::new(bucket_ord),
-            cumulated_work: AtomicU64::new(cumulated_work),
-        }
-    }
+    const COLOR_MASK: u64 = 0x8000000000000000;
```
fulmicoton, line 145:
```suggestion
    const COLOR_MASK: u64 = 1u64 << 63;
```

quickwit/quickwit-common/src/tower/rate_estimator.rs
```diff
@@ -119,59 +124,57 @@ impl RateEstimator for SmaRateEstimator {
...
 
-    fn bucket_ord(&self) -> u64 {
-        self.bucket_ord.load(Ordering::Acquire)
-    }
+    const WORK_MASK: u64 = 0x7FFFFFFFFFFFFFFF;
```
fulmicoton, line 147:
```suggestion
    const WORK_MASK: u64 = u64::MAX - COLOR_MASK;
```

quickwit/quickwit-common/src/tower/rate_estimator.rs
fulmicoton, line 153: We could pass color directly in this method maybe?

fulmicoton (approved):

quickwit/quickwit-common/src/tower/rate_estimator.rs
```diff
@@ -119,59 +124,57 @@ impl RateEstimator for SmaRateEstimator {
     fn update(&mut self, _started_at: Instant, ended_at: Instant, work: u64) {
```
fulmicoton, line 124: could we ditch `_start_at`? I know it is used in the dummy impl, but do we use it? Could it be internalized in the Dummy impl?

guilload: I should actually use it... The argument should be used to check whether the duration spans multiple buckets, and if so, we should spread the amount of work evenly between buckets. However, in our case, the buckets are large (10s), and the requests are short, so the approximation seems ok. What do you think?

PR #3212, opened 2023-04-22
Add an internal incarnation id to the index

### LINKED ISSUE #3041: Add an internal incarnation number/id to the index id for the control plane
Currently, the control plane has a dreadful glitch if we delete an index and swiftly recreate an index with the exact same name. (This is a rather common operation.)

The control plane will not have time to stop existing pipelines, and will be perfectly happy to see a pipeline already running with the specified index id.

As a fix for #3036, I suggest we introduce the notion of incarnation number for our index id.

This number is never used/shown to the user. 
The index id is still unique at any moment of time.

The point of this incarnation id is to make it natural for the control plane to  notice that the instance before the delete/create operaiton has happened is obsolete.

In the future, this incarnatin is also an elegant way to allow a change of conifguration.
We can simply bump the incarnation number when a change is operated.

We also need to take change the incarnation of `clear index` for the same reasons.

imotov:
### Description

Adds an internal incarnation id to the index in order to deferentiate a newly created index from a recently deleted index with the same name.

Fixes #3041

### How was this PR tested?

Added unit and integration tests.


quickwit/quickwit-integration-tests/src/test_utils/cluster_sandbox.rs
```diff
@@ -187,6 +188,68 @@ impl ClusterSandbox {
...
+                == required_pipeline_num
+            {
+                return Ok(());
+            }
+            num_attempts += 1;
+        }
+        if num_attempts == max_num_attempts {
+            anyhow::bail!("Too many attempts to get expected number of pipelines.");
+        }
+        Ok(())
```
fulmicoton, line 218: const MAX_NUM_ATTEMPTS: u64 = 3;
        for num_attempts in 0..MAX_NUM_ATTEMPTS {
            if num_attempts > 0 {
                tokio::time::sleep(Duration::from_millis(100 * num_attempts)).await;
            }
            if self
                .indexer_rest_client
                .node_stats()
                .indexing()
                .await
                .unwrap()
                .num_running_pipelines
                == required_pipeline_num
            {
                return Ok(());
            }
        }
        anyhow::bail!("Too many attempts to get expected number of pipelines.")

quickwit/quickwit-integration-tests/src/test_utils/cluster_sandbox.rs
```diff
@@ -187,6 +188,68 @@ impl ClusterSandbox {
...
+                .await
+                .unwrap()
+                .len()
+                == required_splits_num
+            {
+                return Ok(());
+            }
+            num_attempts += 1;
+        }
+        anyhow::bail!("Too many attempts to get expected number of published splits.");
```
fulmicoton, line 250:
```suggestion
                const MAX_NUM_ATTEMPTS: u64 = 3;
        for num_attempts in 0..MAX_NUM_ATTEMPTS {
            if num_attempts > 0 {
                tokio::time::sleep(Duration::from_millis(100 * num_attempts)).await;
            }
            if self
                .indexer_rest_client
                .node_stats()
                .indexing()
                .await
                .unwrap()
                .num_running_pipelines
                == required_pipeline_num
            {
                return Ok(());
            }
        }
        anyhow::bail!("Too many attempts to get expected number of pipelines.")
```

quickwit/quickwit-integration-tests/src/test_utils/cluster_sandbox.rs
fulmicoton, line 218:
```suggestion
        let mut num_attempts = 0;
        let max_num_attempts = 3;
        while num_attempts < max_num_attempts {
            if num_attempts > 0 {
                tokio::time::sleep(Duration::from_millis(100 * (num_attempts))).await;
            }
            if self
                .indexer_rest_client
                .node_stats()
                .indexing()
                .await
                .unwrap()
                .num_running_pipelines
                == required_pipeline_num
            {
                return Ok(());
            }
            num_attempts += 1;
        }
        if num_attempts == max_num_attempts {
            anyhow::bail!("Too many attempts to get expected number of pipelines.");
        }
        Ok(())
```

quickwit/quickwit-integration-tests/src/test_utils/cluster_sandbox.rs
```diff
@@ -187,6 +188,68 @@ impl ClusterSandbox {
         Ok(())
     }
 
+    // Waits for the needed number of indexing pipeline to start.
+    pub async fn wait_for_indexing_pipelines(
+        &self,
```
fulmicoton, line 193: I opened a ticket to deal with those...
https://github.com/quickwit-oss/quickwit/issues/3214

The original is pretty terrible by the way...
Why a while loop ? why the condition at the end?
Why the condition at the beginning?

quickwit/quickwit-metastore/src/metastore/index_metadata/serialize.rs
```diff
@@ -67,6 +69,9 @@ impl From<IndexMetadata> for IndexMetadataV0_5 {
 
 #[derive(Clone, Debug, Serialize, Deserialize, utoipa::ToSchema)]
 pub(crate) struct IndexMetadataV0_5 {
+    #[schema(value_type = String)]
+    #[serde(default)]
```
fulmicoton, line 73:
```suggestion
    // Defaults to nil for backward compatibility.
    #[serde(default)]
```

quickwit/quickwit-metastore/test-data/index-metadata/v0.5-48d02d9017e35d54e3c67cf2e5d3002a.expected.json
```diff
@@ -0,0 +1,109 @@
+{
+  "checkpoint": {
+    "kafka-source": {
+      "00000000000000000000": "00000000000000000042"
+    }
+  },
```
fulmicoton, line 6: we can remove this file

quickwit/quickwit-metastore/test-data/index-metadata/v0.5-48d02d9017e35d54e3c67cf2e5d3002a.json
```diff
@@ -0,0 +1,109 @@
+{
+  "checkpoint": {
+    "kafka-source": {
+      "00000000000000000000": "00000000000000000042"
+    }
+  },
+  "create_timestamp": 1789,
+  "incarnation_id": "11111111111111111111111111",
```
fulmicoton, line 8: we can remove this file..

quickwit/quickwit-proto/src/lib.rs
```diff
@@ -287,10 +289,105 @@ impl<'a> Extractor for MetadataMap<'a> {
...
+            anyhow!(
+                "Invalid index task format, cannot find source_id in `{}`",
+                index_task_str
+            )
+        })?;
+        let incarnation_id = iter.next().map_or_else(
+            || Ok(ulid::Ulid::default()),
+            |encoded| {
+                Ulid::from_string(encoded).map_err(|err| {
+                    anyhow!(
```
fulmicoton, line 329: nitpick:
alternatively you can use 

`.with_context(|| {  ... })`

quickwit/quickwit-proto/src/lib.rs
```diff
@@ -287,10 +289,105 @@ impl<'a> Extractor for MetadataMap<'a> {
...
+                index_task_str
+            )
+        })?;
+        let source_id = iter.next().ok_or_else(|| {
+            anyhow!(
+                "Invalid index task format, cannot find source_id in `{}`",
+                index_task_str
+            )
+        })?;
+        let incarnation_id = iter.next().map_or_else(
```
fulmicoton, line 325: nitpick:
Rust std has a massive number of functional methods. I personally think 
the most obscure ones just make the code less readable. I'd put `.map_or_else(...)` into that category. 


Here `map` and `unwrap_or_else` would work.

My personal favorite, especial when Error is involved like this, is to just use an if statement.

```rust
let incarnation_id = 
     if let Some(incarnation_id_str) = iter.next() {
           Ulid::from_string(encoded)
                 .context("Invalid index task format, cannot parse incarnation_id")?
     } else {
           Ulid::default()
     };
 ```

**Review by fulmicoton — CHANGES_REQUESTED** [review]
Your PR ended up generating too many non-regression test files. (probably because of an intermediary state).

You should have only one new file per category, and some edits in the expected files.
We should also probably bump the index version to avoid duplicating all of these new files, when we bump the version.

File: quickwit/quickwit-metastore/test-data/index-metadata/v0.5-97f3a089a54400241da1cab788695aca.expected.json [hunk 9]
```diff
@@ -0,0 +1,109 @@
+{
+  "checkpoint": {
+    "kafka-source": {
+      "00000000000000000000": "00000000000000000042"
+    }
+  },
+  "create_timestamp": 1789,
+  "incarnation_id": "0001EHZADJ3VAF00JK8RQ0VMSA",
+  "index_config": {
```
fulmicoton, line 9: two "index-metadata" files where created.
One is sufficient (see PR comment). Ideally v0.6.

quickwit/quickwit-metastore/test-data/index-metadata/v0.5-1a77cc5755c53213debe38b455a499a5.json
```diff
@@ -5,6 +5,7 @@
     }
   },
   "create_timestamp": 1789,
+  "incarnation_id": "00000000000000000000000000",
```
fulmicoton, line 8: we are not supposed to edit this file.

quickwit/quickwit-metastore/test-data/index-metadata/v0.4-b5e855552221fc7a3d31f4751a9348e3.json
```diff
@@ -5,6 +5,7 @@
     }
   },
   "create_timestamp": 1789,
+  "incarnation_id": "00000000000000000000000000",
```
fulmicoton, line 8:
```suggestion
```

quickwit/quickwit-metastore/test-data/file-backed-index/v0.5-8ee262f0219f8e823ae6cbaf9681548e.json
```diff
@@ -17,6 +17,7 @@
       }
     },
     "create_timestamp": 1789,
+    "incarnation_id": "00000000000000000000000000",
```
fulmicoton, line 20:
```suggestion
```

quickwit/quickwit-metastore/test-data/file-backed-index/v0.4-333b645a318ab8f1dac4fd5b5162a638.json
```diff
@@ -17,6 +17,7 @@
       }
     },
     "create_timestamp": 1789,
+    "incarnation_id": "00000000000000000000000000",
```
fulmicoton, line 20:
```suggestion
```

quickwit/quickwit-metastore/test-data/file-backed-index/v0.4-1cfee07237d8cbbe9240d5a883bf98c2.json
```diff
@@ -17,6 +17,7 @@
       }
     },
     "create_timestamp": 1789,
+    "incarnation_id": "00000000000000000000000000",
```
fulmicoton, line 20:
```suggestion
```

fulmicoton (changes requested):
We are not supposed to update existing non-regression .json.
Only the `.expected.json` files.

The latter have their changes auto generated by the test. The role of the dev & reviewer is to eyeball the diff.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -478,22 +485,23 @@ impl IndexingService {
...
+        let indexes_metadata_by_index_id: HashMap<IndexConfigId, IndexMetadata> = indexes_metadata
             .into_iter()
-            .map(|index_metadata| (index_metadata.index_config.index_id.clone(), index_metadata))
+            .map(|index_metadata| (index_metadata.index_config_id(), index_metadata))
             .collect();
 
         let mut failed_spawning_pipeline_ids: Vec<IndexingPipelineId> = Vec::new();
 
         // Add new pipelines.
         for new_pipeline_id in updated_pipeline_ids.difference(&running_pipeline_ids) {
```
fulmicoton, line 499: While you are editing this file, can you do the necessary change to reuse `new_pipeline_ids`?
The current code is weird.

```suggestion
        for new_pipeline_id in new_pipeline_ids {
```

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -478,22 +485,23 @@ impl IndexingService {
...
             info!(pipeline_id=?new_pipeline_id, "Spawning indexing pipeline.");
+
             let index_metadata = indexes_metadata_by_index_id
-                .get(&new_pipeline_id.index_id)
-                .expect("`indexes_metadata_by_index_id` must contain the index ID.");
+                .get(&new_pipeline_id.index_config_id)
+                .expect("`indexes_metadata_by_index_id` must contain the index incarnation ID."); // TODO: should we now just warn here?
```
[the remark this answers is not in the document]
imotov: Oops. I was planning to get back to that TODO but completely forgot with integration test yak shaving. 

> This is important, because we have a race condition that did not really exist before here.

That's what I thought when I added this TODO comment, but after thinking about it more today, it doesn't seem to me that it is a completely new race condition here. I think you get the same condition today if you create and immediately delete an index. The only difference in behavior between the current main branch and this PR can be triggered by also creating an index with the same name afterwards. In this case the current code gets confused and code in this PR continues to fail the same way as if the index wasn't re-created at all or the new index had a different name.

Anyway, a more important question here is should we fail the entire plan if one of the indices fail? Panic here seems like a nuclear option for a theoretically very possible scenario. So, I feel like simply ignoring this index might be a better solution. If I am reading the scheduler code correctly we are checking the applied plan against chitchat server, so any discrepancies should be reconcile on the next run.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
imotov: Let's get #3215 in and then I will see if I can make it cleaner overall.

quickwit/quickwit-metastore/test-data/index-metadata/v0.5-48d02d9017e35d54e3c67cf2e5d3002a.expected.json
imotov: Not this one but we another one that was also created. This one keeps getting recreated.

quickwit/quickwit-metastore/test-data/index-metadata/v0.5-97f3a089a54400241da1cab788695aca.expected.json
imotov: I am not sure how to change it to v0.6 without migrating the protocol version, which I thought we decided to postpone until later in the release cycle.

quickwit/quickwit-metastore/test-data/index-metadata/v0.5-1a77cc5755c53213debe38b455a499a5.json
imotov: Yeah. I got really confused with that.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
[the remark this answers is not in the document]
imotov: @fulmicoton I pushed some changes to improve readability and no longer panic here. Could you take a look?

quickwit/quickwit-metastore/test-data/index-metadata/v0.5-97f3a089a54400241da1cab788695aca.expected.json
fulmicoton: ok Iguess we can clean up those files later.

fulmicoton (approved):

PR #3216, opened 2023-04-24
improve dynamic mapping docs

PSeitz:
add missing dynamic_mapping configuration example
document fast parameter
add dynamic_mapping paragraph
rephrase


docs/configuration/index-config.md
```diff
@@ -379,21 +380,29 @@ Quickwit offers you three different modes:
...
 
 ```yaml
-- indexed: true
-- stored: true
-- tokenizer: default
-- record: basic
-- expand_dots: true
+version: 0.5
+index_id: my-dynamic-index
+doc_mapping:
```
fulmicoton, line 392: ah yes that's much better!

fulmicoton (approved):

PR #3227, opened 2023-04-25
Sets fixed version for docker compose images

### LINKED ISSUE #3184: Using :latest tag in docker compose
Currently, we are using `:latest` tag for several images in docker-compose script that starts services for our test environment. Since docker compose doesn't pull new images in our current configuration, it means that we are testing against different versions of several services. Basically everybody are using whatever was the "latest" version when they ran this script the first time. I see a couple of issues with that:

1. We don't detect breaking changes in these dependencies, which can lead to bad newcomer experience. See #3177, for example, I only noticed the issue after 2 weeks because I recently rebuilt my linux machine.
2. We might be running old images that might have security vulnerabilities on our development machines. That can be mitigated to a certain degree by #3183, but at the moment I think it is an issue.

Any thoughts/suggestions?

imotov:
### Description

Sets fixed versions for docker images, moves data volumes from temporary local directories to docker volumes and tweaks the healthcheck timeouts to reduce startup time.

Fixes #3184

### How was this PR tested?

I ran it multiple times on Mac OS and Ubuntu. In my testing the startup time for the docker-compose scripts is reduced from 30+ second (with occasional timeout) to 9-11 seconds.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,5 @@
 version: "3.9"
-name: quickwit
+
 networks:
   default:
     name: quickwit-network
@@ -35,7 +35,8 @@ services:
       retries: 100
 
   postgres:
-    image: postgres:${POSTGRES_VERSION:-9.6.20} # Oldest version we support.
+    # The oldest supported version. EOL November 9, 2023
+    image: postgres:${POSTGRES_VERSION:-11.19-alpine} 
     container_name: postgres
     ports:
       - "5432:5432"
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -20,22 +20,22 @@ services:
...
+      - localstack_data:/var/lib/localstack
     healthcheck:
       test: ["CMD", "curl", "-k", "-f", "https://localhost:4566/quickwit-integration-tests"]
       interval: 1s
       timeout: 5s
       retries: 100
 
   postgres:
-    image: postgres:9.6.20 # Oldest version we support.
+    image: postgres:${POSTGRES_VERSION:-9.6.20} # Oldest version we support.
```
guilload, line 38: 9 and 10 are no longer supported. Shall we set `11.19`?

docker-compose.yml
```diff
@@ -1,5 +1,5 @@
 version: "3.9"
-
+name: quickwit
```
guilload, line 2: maybe `quickwit-compose`?

guilload (approved):

docker-compose.yml
imotov: Sure, I will fix that. I thought it was about the oldest version **we** support, not the version that a vendor supports.

Issue #2921, opened 2023-03-02
pulsar_broker_tests are consistently failing on an arm64 mac


imotov:
**Describe the bug**
Every time I run the tests on my M1-base mac pulsar_broker_tests are failing with the following message:

```
failures:

---- source::pulsar_source::pulsar_broker_tests::test_doc_batching_logic stdout ----
thread 'source::pulsar_source::pulsar_broker_tests::test_doc_batching_logic' panicked at 'Setup pulsar source: Connection error: fatal error when connecting to the Pulsar server

Caused by:
    fatal error when connecting to the Pulsar server', quickwit-indexing/src/source/pulsar_source.rs:834:14

---- source::pulsar_source::pulsar_broker_tests::test_multi_topic_ingestion stdout ----
thread 'source::pulsar_source::pulsar_broker_tests::test_multi_topic_ingestion' panicked at 'Create source: Failed to create source `test-pulsar-source--source-RPFBm` of type `pulsar`. Cause: Connection error: fatal error when connecting to the Pulsar server

Caused by:
    fatal error when connecting to the Pulsar server

Caused by:
    0: Connection error: fatal error when connecting to the Pulsar server
    1: fatal error when connecting to the Pulsar server', quickwit-indexing/src/source/pulsar_source.rs:968:10

---- source::pulsar_source::pulsar_broker_tests::test_partitioned_topic_multi_consumer_ingestion stdout ----
thread 'source::pulsar_source::pulsar_broker_tests::test_partitioned_topic_multi_consumer_ingestion' panicked at 'Send admin request: reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("localhost")), port: Some(8081), path: "/admin/v2/persistent/public/default/test-pulsar-source--partitioned-multi-consumer--topic-gTKk2/partitions", query: None, fragment: None }, source: hyper::Error(IncompleteMessage) }', quickwit-indexing/src/source/pulsar_source.rs:700:14

---- source::pulsar_source::pulsar_broker_tests::test_partitioned_topic_single_consumer_ingestion stdout ----
[...]

PR #3229, opened 2023-04-26
Map docker compose services to localhost

### LINKED ISSUE #3183: Should we switch docker-compose environment to listen on localhost by default?
As a follow up to #3177 I would like to see if anybody sees any issues with switching docker-compose to listen on localhost instead of 0.0.0.0 by default. We can control this with `.env` for example so switching back to listening on external interfaces should be possible without modifying docker-compose.yml file, but by default services will only be available to the host and not to the outside world.

imotov:
### Description

Maps docker compose test services to localhost by default. This can be changed by changing the MAP_HOST_<SERVICE> in .env file to 0.0.0.0. It also removes .env file from the version control to make it available for making these changes.

Fixes #3183

### How was this PR tested?

Tested it on MacOS and Ubuntu.


fulmicoton:
ah sweet. does that mean after this change the docker-compose file works without trouble on linux and Mac? Should we update our tutorials accorindly?

right now on linux we just suggest using the host network mode.

imotov:
On Mac it still has the issue outlined in #2921, so I still have to use the workaround I outlined in my [last comment on that issue](https://github.com/quickwit-oss/quickwit/issues/2921#issuecomment-1509528950). I don't think it will change until we have an official arm64 image in 3.0. With this workaround it works fine though.

If I understand it correctly, the suggestion to use host network mode will need to stay, since it provides a solution for accessing quickwit running directly on the host from a service running in a docker container. This PR doesn't change this part. It only prevents somebody sitting on the same coffeeshop network with you from accessing such service on your machine unless you explicitly open that service to outside world.

.env.example
```diff
@@ -0,0 +1,26 @@
...
+
+# Update all services to the latest versions
+LOCALSTACK_VERSION=latest
+POSTGRES_VERSION=latest
+PULSAR_VERSION=latest
+CP_VERSION=latest
+AZURITE_VERSION=latest
+JAEGER_VERSION=latest
+OTEL_VERSION=latest
+PROMETHEUS_VERSION=
```
guilload, line 14:
```suggestion
PROMETHEUS_VERSION=latest
```

guilload (approved):

PR #3241, opened 2023-04-27
timestamp pruning guide [docs]

evanxg852000:
Added a guide on timestamp pruning guide

Closes https://github.com/quickwit-inc/devrel/issues/3


docs/guides/timestamp-pruning.md
```diff
@@ -0,0 +1,83 @@
...
+http://127.0.0.1:7280/api/v1/hdfs/search
+```
+:::
+
+We understand that it's not convenient:
+
+- To use another set of parameters to take advantage of timestamp pruning. Extracting these from the search query feels more natural.
+- To specify timestamp pruning parameters only in timestamp `seconds`. Supporting other datetime formats would yield a better user experience.
+
+These are known [issues](https://github.com/quickwit-oss/quickwit/issues/3169) we intend to address in an up-coming version of Quickwit.
```
trinity-1686a, line 82: this part may become wrong in a very close future, see #3400 .

docs/guides/timestamp-pruning.md
```diff
@@ -0,0 +1,76 @@
+---
+title: Timestamp pruning
+sidebar_position: 4
+---
+
+Quickwit indexes documents in bacthing fashion by outputing self contained index files called [splits](https://quickwit.io/docs/overview/architecture/#splits). Overtime, the number of splits created in a Quickwit index grows as long as new documents are being indexed. This number of splits linearly hurts query performance because Quickwit needs to examine all the splits in an index during query execution.
```
fmassot, line 6:
```suggestion
Quickwit indexes documents in batching fashion by outputting self-contained index files called [splits](../architecture/#splits). Over time, the number of splits created in a Quickwit index grows as long as new documents are being indexed. This number of splits linearly hurts query performance because Quickwit needs to examine all the splits in an index during query execution.
```

we need to use relative path for the docs

PR #3243, opened 2023-04-28
Refactor quickwit-cluster

guilload:
### Description
I have a few things left to do, but this PR is in good enough shape to receive a review. My goal for this PR was to refactor the cluster logic so we could:
- share HTTP connections between multiple balance channels and client pools
- easily access up-to-date ready nodes
- streamline access to cluster change events
- provide a balance channel with a better balance policy

In addition, a few things have changed:
- Chitchat now always returns the self node in the list of lives nodes
- Readiness is now handled in Quickwit
- We now distinguish the ready nodes from the live nodes in the cluster snapshot

### How was this PR tested?
- Updated unit tests
- Launched multiple nodes locally

### DIFF
--- a/quickwit/Cargo.toml
+++ b/quickwit/Cargo.toml
@@ -43,14 +43,14 @@ assert-json-diff = "2"
 async-speed-limit = "0.4"
 async-trait = "0.1"
 atty = "0.2"
-azure_core = {version = "0.5.0", features = ["enable_reqwest_rustls"] }
-azure_storage = { version = "0.6.0", default-features = false, features = ["enable_reqwest_rustls"]}
-azure_storage_blobs = { version = "0.6.0", default-features = false, features = ["enable_reqwest_rustls"]}
+azure_core = { version = "0.5.0", features = ["enable_reqwest_rustls"] }
+azure_storage = { version = "0.6.0", default-features = false, features = ["enable_reqwest_rustls"] }
+azure_storage_blobs = { version = "0.6.0", default-features = false, features = ["enable_reqwest_rustls"] }
 backoff = { version = "0.4", features = ["tokio"] }
 base64 = "0.21"
 byte-unit = { version = "4", default-features = false, features = ["serde"] }
 bytes = "1"
-chitchat = { git = "https://github.com/quickwit-oss/chitchat", rev = "4973853" }
+chitchat = { git = "https://github.com/quickwit-oss/chitchat", rev = "bc29598" }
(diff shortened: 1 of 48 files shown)


quickwit/Cargo.toml
```diff
@@ -50,7 +50,7 @@ backoff = { version = "0.4", features = ["tokio"] }
 base64 = "0.21"
 byte-unit = { version = "4", default-features = false, features = ["serde"] }
 bytes = "1"
-chitchat = { git = "https://github.com/quickwit-oss/chitchat", rev = "4973853" }
+chitchat = { path = "../../chitchat/chitchat" }
```
fulmicoton, line 53: this needs to point to a rev within main before merging.

guilload: We're almost there on the Chitchat front, there's one more PR to be merged, and then we'll be good to go.

quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -58,81 +62,76 @@ const MARKED_FOR_DELETION_GRACE_PERIOD: usize = if cfg!(any(test, feature = "tes
...
-    members_rx: watch::Receiver<Vec<ClusterMember>>,
-
-    /// A stop flag of cluster monitoring task.
-    /// Once the cluster is created, a task to monitor cluster events will be started.
-    /// Nodes do not need to be monitored for events once they are detached from the cluster.
-    /// You need to update this value to get out of the task loop.
-    stop: Arc<AtomicBool>,
+    cluster_id: String,
+    self_chitchat_id: ChitchatId,
+    gossip_listen_addr: SocketAddr,
```
fulmicoton, line 69:
```suggestion
    /// The current node gossip UDP socket address.
    gossip_listen_addr: SocketAddr,
```

quickwit/quickwit-common/src/sorted_iter.rs
```diff
@@ -0,0 +1,294 @@
...
+                    .next()
+                    .expect("The right iterator should not be empty.");
+                Some(Diff::Added(right))
+            }
+            (None, None) => None,
+        }
+    }
+}
+
+impl<T> SortedIterator for T where T: Iterator + Sorted {}
```
fulmicoton, line 104: probably a silly question, but why not just have a `SortedIterator` trait?
(and not have a Sorted trait).

Would implementating

```rust
impl<K, V> SortedIterator for btree_map::IntoKeys<K, V> {}
...
```
have worked?

Issue #2878, opened 2023-02-23
Clean up and upgrade dependencies


guilload:
- [x] remove time dependency from `chrono` (opt out of `oldtime` feature)
- [x] remove dependency on `datadog-grok` (opt out of `parse_grok(s)` features)
- [x] depend on last release version of `vrl-compiler/stdlib`
- [x] remove unused deps (`cargo +nightly udeps`)
- [x] upgrade deps

guilload:
Fixed via #3245. Thanks @saroh.

PR #3245, opened 2023-04-30
Update deps.

saroh:
### Description

- [x] remove time dependency from `chrono` (opt out of `oldtime` feature)
- [x] remove dependency on `datadog-grok` (opt out of `parse_grok(s)` features)
- [x] depend on last release version of `vrl-compiler/stdlib`
- [x] remove unused deps (`cargo +nightly udeps`)
- [x] upgrade deps (`cargo upgrade`)

Refs #2878 

### How was this PR tested?

UTs.


guilload:
AWESOME!

Two questions:
- What functions depend on the `wasmbind` chrono feature?
- How hard would it be to automate the upgrade and remove unused deps process?

guilload (approved):

quickwit/Cargo.toml
```diff
@@ -51,25 +51,25 @@ base64 = "0.21"
 byte-unit = { version = "4", default-features = false, features = ["serde"] }
 bytes = "1"
 chitchat = { git = "https://github.com/quickwit-oss/chitchat", rev = "4973853" }
-chrono = "0.4.23"
+chrono = { version = "0.4.24", default-features = false, features = ["clock", "std", "wasmbind"] }
 clap = { version = "=3.1", features = ["env"] }
```
fulmicoton, line 55: do we actually use wasmbind?

quickwit/Cargo.toml
```diff
@@ -51,25 +51,25 @@ base64 = "0.21"
...
 colored = "2.0.0"
 console-subscriber = "0.1.8"
 criterion = { version = "0.4", features = ["async_tokio"] }
 cron = "0.11.0"
-dialoguer = "0.10.3"
+dialoguer = "0.10.4"
```
fulmicoton, line 60: that's not useful...

It is a bit counterintuitive, but 
```dialoguer = "0.10.4"```

actually means any version greater than 0.10.4  and compatible with semver.

We can keep 0.10.4 and just delete refresh `Cargo.lock`.

fulmicoton (changes requested):
Most of the version updates in this PR are not very useful.
Can you revert those and keep the good stuff in?
(VRL update, default-features = false)


Copy pasting my inner comment:

It is a bit counterintuitive, but
dialoguer = "0.10.3"

actually means any version greater than 0.10.4 and compatible with semver.

We can keep 0.10.4 and just delete refresh Cargo.lock.

Keeping it low has some virtue.
If one deps depends on "=0.10.3", (contrived example of course), then that Cargo will do the right thing and stick to 0.10.3 to avoid having two versions of the same package.

saroh:
> How hard would it be to automate the upgrade and remove unused deps process?

THe upgrade is easy but seeing the other comments it's not necessary here :) 
The remove is harder: it has false positives and for instance in our case there's one caught by clippy in `quickwit-metadata`

guilload:
At one point, we were only depending on minor versions, i.e. `x.y` instead of `x.y.z`. Should we revert back to that practice?

fulmicoton:
@guilload. That practise was due to me being ignorant.

crate = "x.y.z" actually is interpreted as 
pick anything semver compatible and greater than "x.y.z".
If possible pick the latest one.

This is exactly what we want.

Using a ".z" is a good practise, because let's say you are using a crate today and test it over 4.1.4.
You only know that it works on 4.1.4, and nothing tells you it is compatible with 4.1.3.

For instance, you might be using a function that did not exist in 4.1.3.

fulmicoton (approved):

fulmicoton:
@saroh awesome job!

PR #3252, opened 2023-05-01
Adds a magic number and versioning to hot directory data and bundle

fulmicoton:
storage.

### DIFF
--- a/quickwit/quickwit-storage/src/versioned_component.rs
+++ b/quickwit/quickwit-storage/src/versioned_component.rs
@@ -1,0 +1,211 @@
+// Copyright (C) 2023 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-storage/src/versioned_component.rs
```diff
@@ -0,0 +1,211 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use std::io::Read;
+
+use anyhow::Context;
+use tantivy::directory::OwnedBytes;
+
+/// Trait to help for versioning.
```
fmassot, line 25: Nitpick: we could move the trait to the common crate.

Nitpick 2: one thing that I found unclear (though it's late there): which components are concerned by this trait? Adding some docs about the use case can help in understanding when it's relevant to use it.

fmassot (approved):

quickwit/quickwit-storage/src/versioned_component.rs
fulmicoton: Nitpick 1: It is not possibly as it depends on Ownedbytes, and we get that from tantivy.
Nitpick 2: I am not fond of the idea.

PR #3254, opened 2023-05-02
add cache for leaf search result

### LINKED ISSUE #3219: Add a leaf level response cache
Split immutability makes it a great candidate for caching.

The idea is that we could keep on leaf sides a `leaf request -> compressed leaf response` cache.
This kind of cache would be especially potent for dashboard generation, (dashboard at T+1 strongly overlaps with dashboard at T) and queries that are being reemitted.

No need to implmeent this now, but the cache will need in the future to tell whether an element is in cache, and protect it from eviction for a few seconds.

trinity-1686a:
### Description

fix #3219
add a cache keyed by request+split_id for leaf search result, reducing io and overall duration of a repeated search query, with some leniency wrt the timestamp covered by the request. Reuse `MemorySizedCache`, making it generic over cache value type.

Size of the cache is pretty small compared to other caches, `LeafSearchResponse` isn't very big, and I suspect the `SearchRequest` key might actually use more memory (which isn't accounted for) for queries with low `max_hits`

### How was this PR tested?

manual testing: making a few request and verify request duration is impacted as expected

##### TODO
add some unit tests for `LeafSearchCache`, notably verify it resists to certains changes in timerange as expected


quickwit/quickwit-proto/Cargo.toml
```diff
@@ -19,6 +19,7 @@ tracing = { workspace = true }
 tracing-opentelemetry = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+tantivy = { workspace = true }
```
fulmicoton, line 22: Noooo let's not add random dependencies like this.

quickwit/quickwit-proto/protos/quickwit/search_api.proto
```diff
@@ -166,7 +166,10 @@ message SplitIdAndFooterOffsets {
   uint64 split_footer_start = 2;
   // The offset of the end of the footer in split bundle. The footer contains the file bundle metadata and the hotcache.
   uint64 split_footer_end = 3;
-
+  // The lowest timestamp appearing in the split
+  optional int64 timestamp_start = 4;
```
fulmicoton, line 170: why do  we need this?

quickwit/quickwit-search/src/leaf_cache.rs
```diff
@@ -0,0 +1,288 @@
...
+
+        self.content.put(key, result);
+    }
+}
+
+#[derive(Debug, Hash, PartialEq, Eq)]
+struct CacheKey {
+    split_id: String,
+    request: SearchRequest,
+    request_time_range: Range,
```
fulmicoton, line 61: isn't the request_time_range part of SearchRequest?

quickwit/quickwit-search/src/leaf_cache.rs
```diff
@@ -0,0 +1,288 @@
...
+            request: search_request,
+            request_time_range,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+struct Range {
+    start: Option<i64>,
+    end: Option<i64>,
```
fulmicoton, line 93: It is crucial to comment what kind of interval this is supposed to be.
Is it closed or semi-open?
(The confusion is the source of many bugs in software)

Implementing https://doc.rust-lang.org/std/ops/trait.RangeBounds.html is nice too.

quickwit/quickwit-search/src/leaf_cache.rs
fulmicoton: ok so my understanding is that it has been cropped to match the split, to increase cache hits.

This should be explained in a comment.

quickwit/quickwit-search/src/leaf_cache.rs
```diff
@@ -0,0 +1,288 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use quickwit_proto::{LeafSearchResponse, SearchRequest, SplitIdAndFooterOffsets};
+use quickwit_storage::MemorySizedCache;
+
+pub struct LeafSearchCache {
+    content: MemorySizedCache<CacheKey, LeafSearchResponse>,
```
fulmicoton, line 24: We would be better off storing `LeafSearchResponse` in serialized form.
(serialized_len is a bad approximation of the amount of memory held by
LeafSearchResponse).

If the code forces you to deserialize/serialize instead of `serialize` this is ok.

fulmicoton (changes requested):
See comments.

quickwit/quickwit-search/src/leaf_cache.rs
```diff
@@ -0,0 +1,351 @@
...
+
+        CacheKey {
+            split_id: split_info.split_id,
+            request: search_request,
+            merged_time_range,
+        }
+    }
+}
+
+/// A (half-open) range bounded inclusively below and exclusively above (start..end).
```
fulmicoton, line 96:
```suggestion
/// A (half-open) range bounded inclusively below and exclusively above [start..end).
```

quickwit/quickwit-search/src/leaf_cache.rs
```diff
@@ -0,0 +1,351 @@
...
+            (Some(this), None) => Some(this),
+            (None, other) => other,
+        };
+
+        let end = match (self.end, other.end) {
+            (Some(this), Some(other)) => Some(this.min(other)),
+            (Some(this), None) => Some(this),
+            (None, other) => other,
+        };
+        Range { start, end }
```
fulmicoton, line 160:
```suggestion
        Range { start, end }.normalize()
```

quickwit/quickwit-search/src/leaf_cache.rs
```diff
@@ -0,0 +1,345 @@
...
+            Bound::Unbounded => None,
+        };
+
+        Range { start, end }
+    }
+
+    /// Normalize empty ranges to be 0..0
+    fn normalize(self) -> Range {
+        match self {
+            Range {
```
fulmicoton, line 136: Range {
  start: None
 end: Some(i64::MIN) 
}
is also empty.

fulmicoton (approved):

fulmicoton (approved):

PR #3266, opened 2023-05-04
Replace index_id with index_uid in metastore methods

### LINKED ISSUE #3231: Replace index_id with index_config_id in metastore methods
As a follow up to #3212 we should either remove index_id from metastore's stage_splits and publish_splits methods since it is already present in supplied split metadata or replace it with index_config_id for additional consistency checks.

imotov:
### Description

Renames index_config_id to index_uid and replaces index_id parameter with index_uid parameter on all metastore methods expect index_metadata and index_exists.

Fixes #3231

### How was this PR tested?

Modified existing tests to support and check for index_uid.


quickwit/quickwit-proto/src/lib.rs
```diff
@@ -297,11 +296,60 @@ pub fn set_parent_span_from_request_metadata(request_metadata: &tonic::metadata:
...
+/// Index identifiers that uniquely identify not only the index, but also
+/// its incarnation allowing to distinguish between deleted and recreated indexes.
+/// It is represented as a stiring in index_id:incarnation_id format.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq, Ord, PartialOrd, Hash)]
+pub struct IndexUid(String);
+
+impl IndexUid {
+    /// Creates a new index uid form index_id and incarnation_id
+    pub fn new(index_id: impl ToString) -> Self {
+        Self::from_parts(index_id,  Ulid::new().to_string())
```
guilload, line 308: `Ulid` are 26-char long, but we actually don't need that many random chars because of the index ID prefix. Should we only use a few random characters to increase readiblity?

quickwit/quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -790,18 +799,29 @@ mod tests {
...
             .returning(|list_split_query: ListSplitsQuery| {
-                if list_split_query.index_id == "quickwit-demo-index"
+                if list_split_query.index_uid.index_id() == "quickwit-demo-index"
                     && list_split_query.split_states
                         == vec![SplitState::Published, SplitState::Staged]
                     && list_split_query.time_range.start == Bound::Included(10)
                     && list_split_query.time_range.end == Bound::Excluded(20)
                     && list_split_query.create_timestamp.end == Bound::Excluded(2)
                 {
+                    println!("OK");
```
guilload, line 821: remove

quickwit/quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -790,18 +799,29 @@ mod tests {
...
                     return Ok(vec![mock_split("split_1")]);
                 }
+                println!("Err");
```
guilload, line 824: remove

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -948,47 +1000,50 @@ impl Metastore for PostgresqlMetastore {
...
     /// Lists the delete tasks with opstamp > `opstamp_start`.
-    #[instrument(skip(self), fields(index_id=index_id))]
+    #[instrument(skip(self), fields(index_id=index_uid.index_id()))]
     async fn list_delete_tasks(
         &self,
-        index_id: &str,
+        index_uid: IndexUid,
         opstamp_start: u64,
     ) -> MetastoreResult<Vec<DeleteTask>> {
+        println!("111111");
```
guilload, line 1033: remove

quickwit/quickwit-metastore/src/metastore/postgresql_model.rs
```diff
@@ -95,7 +96,7 @@ impl Split {
     /// Deserializes and returns the split's metadata.
     fn split_metadata(&self) -> MetastoreResult<SplitMetadata> {
         serde_json::from_str::<SplitMetadata>(&self.split_metadata_json).map_err(|error| {
-            error!(index_id=%self.index_id, split_id=%self.split_id, error=?error, "Failed to deserialize split metadata.");
+            error!(index_id=%IndexUid::from(self.index_uid.clone()).index_id(), split_id=%self.split_id, error=?error, "Failed to deserialize split metadata.");
```
guilload, line 99: `index_id=%self.index_uid.index_id()` does not work?

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
```diff
@@ -198,7 +205,20 @@ impl FileBackedMetastore {
...
+        if locked_index.index_uid() == index_uid {
+            view(&locked_index)
+        } else {
+            Err(MetastoreError::IndexDoesNotExist {
+                index_id: index_id.to_string(),
+            })
+        }
+    }
+
+    async fn read_any<T, F>(&self, index_id: &str, view: F) -> MetastoreResult<T>
```
guilload, line 221: `ready_any` is used once, we could inline it and then have `fn get_locked_index(index_uid: IndexUid)` instead.

guilload (approved):
LGTM.

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
guilload: nvm.

quickwit/quickwit-proto/src/lib.rs
```diff
@@ -297,11 +296,60 @@ pub fn set_parent_span_from_request_metadata(request_metadata: &tonic::metadata:
     Span::current().set_parent(parent_cx);
 }
 
+/// Index identifiers that uniquely identify not only the index, but also
+/// its incarnation allowing to distinguish between deleted and recreated indexes.
+/// It is represented as a stiring in index_id:incarnation_id format.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq, Ord, PartialOrd, Hash)]
+pub struct IndexUid(String);
```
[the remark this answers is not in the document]
imotov: I didn't, but I don't see how this will be better than what I had here - https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-metastore/src/metastore/index_metadata/mod.rs#L40

quickwit/quickwit-proto/src/lib.rs
imotov: Yeah, we can probably safely cut it in half considering frequency of index creation.

quickwit/quickwit-proto/src/lib.rs
```diff
@@ -297,11 +296,60 @@ pub fn set_parent_span_from_request_metadata(request_metadata: &tonic::metadata:
...
+    }
+}
+
+impl ToString for IndexUid {
+    fn to_string(&self) -> String {
+        self.0.clone()
+    }
+}
+
+impl From<String> for IndexUid {
```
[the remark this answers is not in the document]
imotov: It is a conversion from `String` to `IndexUid(String)`. I don't think it is more dangerous than passing index name in String, which it essentially replaces. It is also used only for internal deserialization from our messages and not something provided by the end-user. It can be abused but only if somebody is messing with messages on gprs level. If we want to protect that level from fuzzing, it would require a much bigger effort.

Issue #3281, opened 2023-05-06
Improve readability of aggregation error response


fmassot:
Example::

```json
{"query":"*","aggs":{"1":{"avg":{}}}}
```

```json
{"message":"Invalid aggregation request: data did not match any variant of untagged enum QuickwitAggregations"}
```

fulmicoton:
@PSeitz if we already have a ticket for this, can you close one of the two?
It think it will require an adhoc serializer... let me know if you need help.

PSeitz:
The issue is that there's another `#[serde(untagged)]`. It seems unnecessary to merge the two requests. Can we unsplit it?
@fulmicoton  @guilload 

```rust
/// Available aggregation types.
#[derive(Debug, Clone, Deserialize)]
#[serde(untagged)]
pub enum QuickwitAggregations {
    /// Aggregation used by the Jaeger service to find trace IDs that match a
    /// [`quickwit_proto::jaeger::storage::v1::FindTraceIDsRequest`].
    FindTraceIdsAggregation(FindTraceIdsCollector),
    /// Your classic Tantivy aggregation.
    TantivyAggregations(Aggregations),
}
```

fulmicoton:
Yes I think it would be a good idea to remove this untagged.

It might be painful to add a new type of request (as in RPC service), but we can probably simply add a field in LeafRequest for tracing stuff.

PSeitz:
This is fixed by only returning the tantivy error in case aggregation parsing fails.
The correct solution would be to split tracing and tantivy aggregations.

PR #3286, opened 2023-05-08
improve agg error message

PSeitz:
return only the tantivy error in case aggregation parsing fails.
The correct solution would be to split tracing and tantivy aggregations.

### Description

### How was this PR tested?


PSeitz:
https://github.com/quickwit-oss/quickwit/issues/3281

fulmicoton (approved):

PSeitz:
> Approved, but This is not entirely solving #3281

Yes, the list of allowed aggregations is still missing

PR #3305, opened 2023-05-11
Pass aggregation limits when finalizing aggregation on the root node.

fmassot:
Tested on an aggregation that was failing because the default aggregation limits were still used in the finalization step.

### DIFF
--- a/quickwit/quickwit-search/src/root.rs
+++ b/quickwit/quickwit-search/src/root.rs
@@ -379,7 +379,7 @@ pub async fn root_search(
     let aggregation: Option<String> = finalize_aggregation(
         leaf_search_response.intermediate_aggregation_result,
         aggregations,
-        &searcher_context,
+        searcher_context,
     )?;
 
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-search/src/root.rs
```diff
@@ -394,6 +396,7 @@ pub async fn root_search(
 pub fn finalize_aggregation(
     intermediate_aggregation_result: Option<Vec<u8>>,
     aggregations: Option<QuickwitAggregations>,
+    searcher_context: &Arc<SearcherContext>,
```
guilload, line 399: nit: Can you attempt to change the signature of `aggregation_limits_from_searcher_context ` and this function to `searcher_context: &SearcherContext`. If we're not cloning the `Arc` anywhere, it would be cleaner.

`SearcherContext` could also have an `aggregation_limits` accessor.

guilload (approved):

PR #3313, opened 2023-05-12
Query resource attributes when Jaeger request carries tags

guilload:
### Description
Also populate `service.version` resource attribute in OTLP exporter.

### How was this PR tested?
- Updated unit tests
- Launched quickwit locally with OTLP exporter enabled and queried `service.version` resource attribute via Jaeger UI


guilload:
@fulmicoton, PTAL

PR #3317, opened 2023-05-12
Use index UID in Kafka/Pulsar consumer groups

guilload:
### Description
It is safer than using index IDs when indexes are deleted/recreated.

### How was this PR tested?
`make test-all`


quickwit/quickwit-indexing/src/source/mod.rs
```diff
@@ -115,9 +116,11 @@ impl SourceExecutionContext {
         queues_dir_path: PathBuf,
         source_config: SourceConfig,
     ) -> Arc<SourceExecutionContext> {
+        use ulid::Ulid;
+
         Arc::new(Self {
             metastore,
-            index_id: index_id.to_string(),
+            index_uid: IndexUid::from_parts(index_id, Ulid::nil()),
```
imotov, line 123: I think it might be better to pass `index_uid` here instead of generating a fake one. I tried the fake generation approach in #3212 but very quickly regretted it except for very small unit tests.

imotov (commented):
Thanks for picking it up and pushing it forward!

quickwit/quickwit-indexing/src/source/mod.rs
guilload: Ok!

quickwit/quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -345,12 +346,16 @@ impl KafkaSource {
         assignment_tx: oneshot::Sender<Vec<(i32, Offset)>>,
     ) -> anyhow::Result<()> {
         let index_metadata = ctx
-            .protect_future(self.ctx.metastore.index_metadata(&self.ctx.index_id))
+            .protect_future(
+                self.ctx
+                    .metastore
+                    .index_metadata(self.ctx.index_uid.index_id()),
+            )
```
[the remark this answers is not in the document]
guilload: I agree with you. We need a more systemic approach. In the meantime, I'll make the change you suggested in the Kafka and Pulsar sources.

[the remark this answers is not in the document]
guilload: Well... I have introduced a new metastore method named `index_metadata_strict` with a default implementation that does that.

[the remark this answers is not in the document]
imotov: Thanks!

quickwit/quickwit-indexing/src/source/mod.rs
```diff
@@ -117,7 +118,7 @@ impl SourceExecutionContext {
     ) -> Arc<SourceExecutionContext> {
         Arc::new(Self {
             metastore,
-            index_id: index_id.to_string(),
+            index_uid: IndexUid::new(index_id),
```
imotov, line 121: That's not what I meant, but considering the current usage, that will work for now.

quickwit/quickwit-proto/src/lib.rs
```diff
@@ -339,7 +339,7 @@ impl IndexUid {
         } else {
             Self (
                 format!("{}:{}", index_id.into(), incarnation_id)
-            )                
+            )
```
imotov, line 342: Interesting. Did you format that manually? It seem that `make fix` doesn't pick that up.

imotov (approved):
LGTM

quickwit/quickwit-proto/src/lib.rs
guilload: My editor is set up to remove trailing whitespaces.

PR #3315, opened 2023-05-12
Added rest-api-tests.

fulmicoton:
This PR does not add them to CI yet.

See README.md

Issue #3296, opened 2023-05-09
[Discussion] Improve mapping configuration for a better log search experience


fmassot:
While testing our Grafana plugin, I went though a few frustrating cases:

I indexed most of the fields in a schemaless manner. Everything worked fine but:

- I was not able to search on a long UID. This is due to the fact the default tokenizer will omit tokens longer than 40 bytes.
- I was not able to search on UID with "-" separator. This is annoying especially when you are able to make an aggregation on the untokenized version.
- Same story on IP and several fields where you have short name that contains a character that is not a number nor a letter.

It seems the right dynamic setup is to not tokenize text by default, and on very specific field tokenize the text. And on this specific field, you want to set `fast: false`.

I would change the default tokenizer and raise the max token length to `255` bytes as done in elasticsearch.

- [ ] Raise max token limit to 255  (#3372)
- [ ] Make static mapping of object in array possible (#3371)
- [ ] Interpret implicit phrase queries as an intersection when position are not available (#3364)
- [ ] Set the default tokenizer for json & dynamic stuff to raw (#3373)

fulmicoton:
> I was not able to search on UID with "-" separator. This is annoying especially when you are able to make an aggregation on the untokenized version.

For confirmation:
When you run the query using the normal search UI the spec is that it should run as a phrase query...
Did you receive an error message saying positions where not indexed and hence phrase query cannot?
Should we run those as intersection?

fmassot:
> Did you receive an error message saying positions where not indexed and hence phrase query cannot?
Yes I did, and this was frustrating. Running those as intersections can improve the UX.

fulmicoton:
I added a bunch of concrete actions and created separate tickets for each them. I need your input on #3371 @fmassot

guilload:
This issue also demonstrates the need for a "log" tokenizer that would be better at identifying dates, IPs, and UUIDs. I'm adding #1143 to the 0.7 project.

fulmicoton:
Closing as everything is tracked in other issues.

PR #3321, opened 2023-05-13
Added match query support. First stab

### LINKED ISSUE #3285: Add support for ES "match" requests.

fulmicoton:
Closes #3285

This PR also does some refactoring, to
extend the usage of the PhraseQuery (now FullTextQuery)
to cover the MatchQuery as well.

# How is tested?

Unit test.
I'll add integration tests once #3315 is merged.


quickwit/quickwit-query/src/elastic_query_dsl/match_query.rs
```diff
@@ -0,0 +1,197 @@
...
+    #[serde(default)]
+    default_operator: BooleanOperand,
+    #[serde(default)]
+    zero_terms_query: MatchAllOrNone,
+}
+
+impl ConvertableToQueryAst for MatchQuery {
+    fn convert_to_query_ast(self) -> anyhow::Result<QueryAst> {
+        // TODO fix this.
+        // This is not exactly what we want. This will interpret boolean operators for instance.
```
trinity-1686a, line 54: I don't think this is true?

fulmicoton: ah yes this is an old comment.

fulmicoton, line 54:
```suggestion
```

quickwit/quickwit-query/src/query_ast/term_set_query.rs
```diff
@@ -37,7 +41,11 @@ impl TermSetQuery {
         for (full_path, values) in &self.terms_per_field {
             for value in values {
                 // We simplify the logic by calling compute_query, and extract the resulting terms.
```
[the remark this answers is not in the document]
fulmicoton: I fixed the comment.

quickwit/quickwit-query/src/query_ast/tantivy_query_ast.rs
```diff
@@ -144,12 +180,40 @@ impl TantivyBoolQuery {
...
+            MatchAllOrNone::MatchNone,
+            has_no_positive_ast_so_far,
+        );
+        if self.must.is_empty()
+            && self.should.is_empty()
+            && self.must_not.is_empty()
+            && self.filter.is_empty()
+        {
+            return TantivyQueryAst::match_none();
+        }
```
trinity-1686a, line 216: I have a hard time wrapping my head around this bit.
1st I don't think I understand why the same condition above should `match_all`, but this one, after more filtering, should `match_none`.
2nd I'm not convinced the condition can ever evaluate to `true`. If any of the 4 lists is not empty before, due to the cascading `stop_before_empty = has_no_positive_ast_so_far`, then at least one of the list shouldn't be empty afterward?

fulmicoton: you are correct!

fulmicoton:
Merging to allow pursuing work on #3296. Can you have a look at the last few commits on Monday @trinity-1686a ?

trinity-1686a:
looks good

PR #3332, opened 2023-05-15
Implement generic client pool

guilload:
### Description
Implement a generic client pool that listens to a stream of changes.

### How was this PR tested?
- Added unit tests


quickwit/quickwit-common/src/tower/pool.rs
```diff
@@ -0,0 +1,184 @@
...
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+        }
+    }
+}
+
+#[derive(Default)]
+struct InnerPool<K, S> {
+    nodes: HashMap<K, S>,
```
fmassot, line 60: This is a pool of services... so why not replacing `nodes` by `services`? 

Same for `get_node` which is a bit strange when you get a service in return. Why not `get_service` or `get`?

If I understand correctly the usage, we will have a pool of services identified by a `node_id` key or a `grpc_address` key.

guilload: We can stay generic in the naming then, because it can be a pool of anything really:
For the search client, it's probably going to be `Pool<NodeUid, SearchServiceClient>`. For the control plane, it'll be probably a `Pool<NodeUid, (Vec<IndexingTasks>, IndexingServiceClient)>`, for the event broker, it will be `Pool<NodeUid, EventHandlers>`, etc.

fmassot: Yes, I was going to suggest being agnostic about the value types. I like to generic way here.

Issue #3273, opened 2023-05-05
Fetching 10000 docs in parallel is failing with seaweedfs storage


fmassot:
A user reported the following error when executing 5 requests in parallel. Each request fetches 10 000 docs.

```bash

[2m2023-05-05T09:26:20.710Z[0m [31mERROR[0m [1mfetch_docs[0m[2m:[0m [2mquickwit_search::fetch_docs[0m[2m:[0m Error when fetching docs in splits. [3msplit_ids[0m[2m=[0m["01GYF19T9WRH371R8APMAPDA3R", "01GYF4D88AHBYQ9YATRRFS1XNF", "01GYFBSFKEKWYW7J6CE3H28HZT", "01GYF13J00NP6DDARY2PA8XS96", "01GYF2NB3XKQAH1X3M2NBZ4P2J", ....
 
Caused by:
    An IO error occurred: 'Error obtaining chunk: error reading a body from connection: end of file before message length reached Ctx:Failed to fetch slice 914729148..914974812 for object: s3://quickwit/indexes/index/01GYFADW547Q67QA5KPHPXYT5Q.split'

```

When fetching 1000 documents, there are no errors.

This is probably linked to the number of requests we are making on the storage. Currently, we limit the number of requests per split to `NUM_CONCURRENT_REQUESTS`. But in the case where there are many splits and many requests  in parallel, we have no limitation.

@trinity-1686a is already working on optimization fetch docs requests, it's worth waiting for the merge of @trinity-1686a 
 work and then tackling this issue.

fulmicoton:
We can probably add some configurable limit on the number of concurrent fetch request. A global semaphore will do the trick.

trinity-1686a:
fixed by #3356

PR #3356, opened 2023-05-17
add a semaphore to limit s3 concurrent access

trinity-1686a:
### Description

possible fix for #3273 

I'm not very satisfied by it. I made a 1st iteration in [`trinity--storage-rate-limiter`](https://github.com/quickwit-oss/quickwit/tree/trinity--storage-rate-limiter) which tries to make it configurable per index, but while working on it I found we create a new `Storage` for each request, so a local semaphore doesn't do the trick.

This one is only configurable by an env-var `QW_S3_MAX_CONCURRENCY`.


quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -52,6 +53,13 @@ use crate::{
     STORAGE_METRICS,
 };
 
+static REQUEST_SEMAPHORE: Lazy<Semaphore> = Lazy::new(|| {
```
fulmicoton, line 56: Can you comment above the effect of this semaphore, and why it was introduced?

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -500,6 +508,8 @@ async fn download_all(byte_stream: ByteStream, output: &mut Vec<u8>) -> io::Resu
 #[async_trait]
 impl Storage for S3CompatibleObjectStorage {
     async fn check_connectivity(&self) -> anyhow::Result<()> {
+        // we ignore error as we never close the semaphore
+        let _permit = REQUEST_SEMAPHORE.acquire().await;
```
fulmicoton, line 512: Nitpick: Alternatively we could have used a guard pattern.

```
struct SharedResourceWithConcurrencyLimits<T> {
    object: T,
    semaphore,
}

struct SharedResourceWithConcurrencyLimitsGuard<T> {
    object: &T,
    semaphore_guard:
}

impl Deref<Target=T> for SharedResourceWithConcurrencyLimitsGuard<T> {
  //...
}
```

fulmicoton (approved):

trinity-1686a:
> Why are we limiting the rate of requests at the storage level? Wasn't the plan to limit the number of concurrent fetch docs requests?

I've choose that way because this issue could easily show-up with a mix of fetching docs and fetching terms for a concurrent request, where fetching just alone isn't the issue

PR #3362, opened 2023-05-17
Move OTEL logs and traces indexes to datetime nanos

guilload:
### Description
Move to native datetime field.

### How was this PR tested?
- [x] Updated unit tests and added new ones
- [x] Tested manually with local Jaeger UI
- [ ] Wrote new integration test


quickwit/quickwit-opentelemetry/src/otlp/traces.rs
```diff
@@ -91,19 +93,20 @@ doc_mapping:
...
-      type: u64
-      indexed: false
-    - name: span_start_timestamp_secs
+    - name: span_start_timestamp
       type: datetime
       input_formats: [unix_timestamp]
+      output_format: unix_timestamp_nanos
       indexed: false
       fast: true
       precision: seconds
```
fmassot, line 102: We may want to make range queries with milliseconds precisions. Does Jaeger make that possible?

guilload: Yeah, we also get nanos on the Jaeger side, but the min/max timestamps are coming from the lookback selector with values such as `last x minutes`, or `last x hours`, so the end user does not have the opportunity to issue queries that would take advantage of the millisecond precision.

We do have a fast field in milliseconds for the span duration.

fmassot: Ok. I'm anticipating a bit, but I would expect to support milliseconds of precision in the future. Grafana typically sends date range queries at millisecond precision.

fmassot (approved):

quickwit/quickwit-opentelemetry/src/otlp/traces.rs
```diff
@@ -139,7 +141,7 @@ doc_mapping:
       type: array<json>
       tokenizer: raw
 
-  timestamp_field: span_start_timestamp_secs
+  timestamp_field: span_start_timestamp
 
   # partition_key: hash_mod(service_name, 100)
   # tag_fields: [service_name]
```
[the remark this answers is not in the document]
guilload: Does that work with JSON fields?

fmassot:
@guilload as requested by matou on discord, we should set `resource_attributes` as `fast`. (I thought it was by default ://).

I'm wondering if we should put other json fields as fast too.

PR #3369, opened 2023-05-18
Differentiates absence/presence of quotation marks in user query.

### LINKED ISSUE #3364: Implicit vs Explicit PhraseQueries in the query parser
We really have two kinds of phrase queries emitted by the query parser.

Those that were generated as a result of the user using the quotation mark.
In that case, it can make sense to return an error if the position were not indexed.

`year:"2023-01-03"`
vs
`year:2023-01-03`

Phrase queries that are the result of the query parser tokenizer less than the fieldl tokenizer.
For instance.

- `year:2023-01-03`
- `uuid:123e4567-e89b-12d3-a456-426614174000`
- `jobtitle:editor-in-chief`

with the standard tokenizer, this ends up interpreted as the phrase query

`year:[2023, 01, 03]`
`uuid:[123e4567, e89b, 12d3, 456, 426614174000]`
`jobtitle:[editor, in, chief]`

Of course in both case a raw tokenizer would have been better.

The three cases above are in practise very different.
a) year will yield a lot (50%) of false positives silently. The user may have preferred to have an error.
b) UUID is unlikely to yield false positives
c) jobtitle may cause a few false positives, but this falls under the common precision/recall problem and the user is very unlikely to blame the system.

It is impossible to satisfy the best solution for a) b) and c) here.
If positions are not available we should treat such implicit phrase queries as intersections.

fulmicoton:
Before this change, a user text yielding more than one token would be always interpreted as a phrase query.

jobtitle:editor-in-chief

Would therefore fail if the field was indexed tokenized but without positions.

After this change
jobtitle:"editor-in-chief" is interpreted as a phrase query. jobtitle:'editor-in-chief' is tokenized with the field tokenizer, but interpreted as an intersection.
jobtitle:editor-in-chief is tokenized and interpreted as a phrase query if positions are available for the field at the scale of the split, and will return an error if it is not available.

Closes #3364
Related with #3296


fmassot:
One thing I'm unsure about this. The difference between `jobtitle:"editor-in-chief"`  and `jobtitle:'editor-in-chief'` is subtle and I expect users to enter `jobtitle:"editor-in-chief"` and thus get an error. They won't find the solution easily.

Why not have the fallback on the intersection even with the double quote? 

And if you have strong reasons against it, we can make life easier for the user with a good error message encouraging him to use single quotes.

fulmicoton:
The best fix is to type `jobtitle:editor-in-chief`, not `'jobtitle:editor-in-chief'`.
People should almost never use the `'`. I added it in this PR, because I found it useful from time to time, and it has been demanded on tantivy side.


> Why not have the fallback on the intersection even with the double quote?

The discussion is definitely worth having but focusing on `editor-in-chief` made the discussion a little bit too simple.
What we do with quotation marks will also influence the behavior of `body:"the who"`. 
I choose an horrible edge case here, but I think the worst case scenario is when you get a lower number of crap (< 20%), because you wont' notice it easily.

If someone runs this query, and positions are not indexed, do we want to return an error, or do we want to return the results `body:the` AND `body:who`.

fulmicoton:
Merging, and opening a separate issue to host discussions #3386. (I need to update tantivy).

PR #3388, opened 2023-05-22
Support input-format arg for local-ingest command

evanxg852000:
Added `input-format` argument to Quickwit `local-ingest` command.
Closes https://github.com/quickwit-oss/quickwit/issues/3389


fmassot (changes requested):
Can you add a test that parses the arguments?

PR #3396, opened 2023-05-23
avoid downloading fast fields when avoidable

### LINKED ISSUE #1748: Avoid filter when the entire split is within the timestamp bounds
If we search for `[ta, tb)` and the split meta informs us that the split's timestamp are within the  `[tA..tB]` range,
with  `[tA..tB] ⊆[ta, tb)`  there is no need to do any filtering.

This issue is not entirely straightforward, because the split meta are not passed in the leaf requests.

One possibility might be to get the meta from the fast field and make sure those metadata are added into the hotcache if it is not there yet.

The logic must be implemented for the search **and** search stream.

trinity-1686a:
### Description

fix #1748

### How was this PR tested?

querying a dataset and looking for the log line containing `WarmupInfo`, to verify what doesn't need to be loaded is effectively not preloaded.

### DIFF
--- a/quickwit/quickwit-search/src/leaf.rs
+++ b/quickwit/quickwit-search/src/leaf.rs
@@ -304,11 +304,12 @@ async fn warm_up_fieldnorms(searcher: &Searcher, requires_scoring: bool) -> anyh
 #[instrument(skip(searcher_context, search_request, storage, split, doc_mapper,))]
 async fn leaf_search_single_split(
     searcher_context: &SearcherContext,
-    search_request: &SearchRequest,
+    mut search_request: SearchRequest,
     storage: Arc<dyn Storage>,
     split: SplitIdAndFooterOffsets,
     doc_mapper: Arc<dyn DocMapper>,
 ) -> crate::Result<LeafSearchResponse> {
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -348,14 +349,38 @@ async fn leaf_search_single_split(
...
+        .put(split, search_request, leaf_search_response.clone());
     Ok(leaf_search_response)
 }
 
+/// Rewrite a request removing parts which incure additional download or computation with no
+/// effect.
+///
+/// This include things such as sorting result by a field or _score when no document is requested,
+/// or applying date range when the range covers the entire split.
+fn rewrite_request(search_request: &mut SearchRequest, split: &SplitIdAndFooterOffsets) {
```
fulmicoton, line 363: could we do some factorization with a 
```
fn rewrite_start_end_time_bounds(start_timestamp_opt: &mut Option<i64>, end_timestamp_opt: &mut Option<i64>, split: &SplitIdAndFooterOffsets) {
```

fulmicoton (approved):

PR #3400, opened 2023-05-24
extract timestamp range from query ast

### LINKED ISSUE #3256: Extract timestamp range to enable pruning
We can sometime identify that a range query on the timestamp is implied by the AST.
In that case, we should extract the boundaries and enable timestamp pruning.

related with #3194 but higher priority

(@fmassot  I opened an issue since it is actually different from #3194)

trinity-1686a:
### Description

fix #3256 

### How was this PR tested?

manual test + added unit test


quickwit/quickwit-search/src/root.rs
```diff
@@ -389,6 +400,87 @@ pub async fn root_search(
...
+    }
+    // TODO visiting those could be usefull too,
+    // fn visit_term(
+    // &mut self,
+    // _term_query: &'a TermQuery
+    // ) -> Result<(), Self::Err> { ... }
+    // fn visit_term_set(
+    // &mut self,
+    // _term_query: &'a TermSetQuery
+    // ) -> Result<(), Self::Err> { ... }
```
evanxg852000, line 481: Are these comments intended?

quickwit/quickwit-search/src/root.rs
```diff
@@ -389,6 +402,106 @@ pub async fn root_search(
     })
 }
 
+pub(crate) struct ExtractTimestampRange<'a> {
+    pub timestamp_field: &'a str,
+    pub start_timestamp: Option<i64>,
```
fulmicoton, line 407:
```suggestion
    // Boundary identified as being implied by the QueryAst.
    // `start_timestamp` is to be interpreted as Inclusive (or Unbounded)
    // `end_timestamp` is to be interpreted as Exclusive (or Unbounded)
    // In other word, this is a `[start_timestamp..end_timestamp)` interval.
    pub start_timestamp: Option<i64>,
```

quickwit/quickwit-search/src/root.rs
```diff
@@ -389,6 +402,106 @@ pub async fn root_search(
     })
 }
 
+pub(crate) struct ExtractTimestampRange<'a> {
```
fulmicoton, line 405: I think having a function 

```pub(crate) fn refine_start_end_timestamp_from_ast(query_ast: &QueryAst, start_timestamp: &mut Option<i64>, end_timestamp: &mut Option<i64>)```

Would have been beneficial here...
ExtractTimestampRange could then have been private... And possibly in the query ast files.

fulmicoton (approved):

fulmicoton:
Approved, because this is definitely an improvement over the current situation... Can you however open an extra issue?
Right now this PR extract start/end boundaries,which will help with split pruning.

However as a result, we will add timestamp filtering collector while our RangeQuery is still in our QueryAST and will be executed.

In other words, our timestamp filtering is happening twice.

PR #3404, opened 2023-05-25
add fastfield normalizer

PSeitz:
### Description
split tokenizer into tokenizer and normalizer
use normalizer for fastfield.
handle two tokenizer manager from tantivy
rise length limit for tokens to 255 bytes
update docs for normalizer
enable fast field tokenizer on Json

### How was this PR tested?

tested aggreggations on project airmail with lowercase

### DIFF
--- a/docs/configuration/index-config.md
+++ b/docs/configuration/index-config.md
@@ -115,6 +115,8 @@ type: text
 tokenizer: default
 record: position
 fieldnorms: true
+fast:
+  normalizer: lowercase
 ```
 
 **Parameters for text field**
@@ -126,17 +128,24 @@ fieldnorms: true
 | `tokenizer` | Name of the `Tokenizer`. ([See tokenizers](#description-of-available-tokenizers)) for a list of available tokenizers.  | `default` |
 | `record`    | Describes the amount of information indexed, choices between `basic`, `freq` and `position` | `basic` |
 | `fieldnorms` | Whether to store fieldnorms for the field. Fieldnorms are required to calculate the BM25 Score of the document. | `false` |
-| `fast`     | Whether value is stored in a fast field. The fast field will contain the term ids and the dictionary. The effective cardinality depends on the tokenizer. The default behaviour for `true` is to store the original text unchanged. The tokenizer on the fast field is seperately configured. It can be configured via `{"tokenizer": "lowercase"}`. ([See tokenizers](#description-of-available-tokenizers)) for a list of available tokenizers. | `false` |
+| `fast`     | Whether value is stored in a fast field. The fast field will contain the term ids and the dictionary. The default behaviour for `true` is to store the original text unchanged. The normalizers on the fast field is seperately configured. It can be configured via `normalizer: lowercase`. ([See normalizers](#description-of-available-normalizers)) for a list of available normalizers. | `false` |
 
 #### **Description of available tokenizers**
 
 | Tokenizer     | Description   |
 | ------------- | ------------- |
-| `raw`         | Does not process nor tokenize text  |
-| `default`     | Chops the text on according to whitespace and punctuation, removes tokens that are too long, and converts to lowercase |
-| `en_stem`     |  Like `default`, but also applies stemming on the resulting tokens  |
+| `raw`         | Does not process nor tokenize text. Filters out tokens larger than 255 bytes.  |
+| `default`     | Chops the text on according to whitespace and punctuation, removes tokens that are too long, and converts to lowercase. Filters out tokens larger than 40 bytes. |
+| `en_stem`     |  Like `default`, but also applies stemming on the resulting tokens. Filters out tokens larger than 40 bytes.  |
 | `chinese_compatible` |  Chop between each CJK character in addition to what `default` does. Should be used with `record: position` to be able to properly search |
 | `lowercase` |  Applies a lowercase transformation on the text. It does not tokenize the text. |
+
+#### **Description of available normalizers**
+
+| Normalizer     | Description   |
+| ------------- | ------------- |
+| `raw`         | Does not process nor tokenize text. Filters token larger than 255 bytes.  |
+| `lowercase` |  Applies a lowercase transformation on the text. Filters token larger than 255 bytes. |
 
 **Description of record options**
 
@@ -312,6 +321,8 @@ stored: true
 indexed: true
 tokenizer: "default"
 expand_dots: false
+fast: 
+  normalizer: lowercase
 ```
 
 **Parameters for JSON field**
@@ -321,7 +332,7 @@ expand_dots: false
 | `description` | Optional description for the field. | `None` |
 | `stored`    | Whether value is stored in the document store | `true` |
 | `indexed`   | Whether value is indexed | `true` |
-| `fast`   | Whether value is fast | `false` |
+| `fast`     | Whether value is stored in a fast field. The default behaviour for text in the JSON is to store the text unchanged. An normalizer can be configured via `normalizer: lowercase`. ([See normalizers](#description-of-available-normalizers)) for a list of available normalizers. | `true` |
 | `tokenizer` | **Only affects strings in the json object**. Name of the `Tokenizer`, choices between `raw`, `default`, `en_stem` and `chinese_compatible` | `default` |
 | `record`    | **Only affects strings in the json object**. Describes the amount of information indexed, choices between `basic`, `freq` and `position` | `basic` |
 | `expand_dots`    | If true, json keys containing a `.` should be expanded. For instance, if `expand_dots` is set to true, `{"k8s.node.id": "node-2"}` will be indexed as if it was `{"k8s": {"node": {"id": "node2"}}}`. The benefit is that escaping the `.` will not be required at query time. In other words, `k8s.node.id:node2` will match the document. This does not impact the way the document is stored.  | `true` |
--- a/docs/configuration/index-config.md
+++ b/docs/configuration/index-config.md
@@ -135,8 +135,8 @@ fast:
 | Tokenizer     | Description   |
 | ------------- | ------------- |
 | `raw`         | Does not process nor tokenize text. Filters out tokens larger than 255 bytes.  |
-| `default`     | Chops the text on according to whitespace and punctuation, removes tokens that are too long, and converts to lowercase. Filters out tokens larger than 40 bytes. |
-| `en_stem`     |  Like `default`, but also applies stemming on the resulting tokens. Filters out tokens larger than 40 bytes.  |
+| `default`     | Chops the text on according to whitespace and punctuation, removes tokens that are too long, and converts to lowercase. Filters out tokens larger than 255 bytes. |
+| `en_stem`     |  Like `default`, but also applies stemming on the resulting tokens. Filters out tokens larger than 255 bytes.  |
 | `chinese_compatible` |  Chop between each CJK character in addition to what `default` does. Should be used with `record: position` to be able to properly search |
 | `lowercase` |  Applies a lowercase transformation on the text. It does not tokenize the text. |
 
--- a/quickwit/quickwit-doc-mapper/src/default_doc_mapper/date_time_type.rs
+++ b/quickwit/quickwit-doc-mapper/src/default_doc_mapper/date_time_type.rs
@@ -22,7 +22,7 @@ use quickwit_datetime::{DateTimeInputFormat, DateTimeOutputFormat};
 use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::Value as JsonValue;
 use tantivy::schema::Value as TantivyValue;
-use tantivy::DatePrecision as DateTimePrecision;
+use tantivy::DateTimePrecision;
 
 use super::default_as_true;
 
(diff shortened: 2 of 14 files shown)


docs/configuration/index-config.md
```diff
@@ -126,18 +126,25 @@ fieldnorms: true
...
 | `record`    | Describes the amount of information indexed, choices between `basic`, `freq` and `position` | `basic` |
 | `fieldnorms` | Whether to store fieldnorms for the field. Fieldnorms are required to calculate the BM25 Score of the document. | `false` |
-| `fast`     | Whether value is stored in a fast field. The fast field will contain the term ids and the dictionary. The effective cardinality depends on the tokenizer. The default behaviour for `true` is to store the original text unchanged. The tokenizer on the fast field is seperately configured. It can be configured via `{"tokenizer": "lowercase"}`. ([See tokenizers](#description-of-available-tokenizers)) for a list of available tokenizers. | `false` |
+| `fast`     | Whether value is stored in a fast field. The fast field will contain the term ids and the dictionary. The default behaviour for `true` is to store the original text unchanged. The normalizers on the fast field is seperately configured. It can be configured via `{"normalizer": "lowercase"}`. ([See normalizers](#description-of-available-normalizers)) for a list of available normalizers. | `false` |
```
fmassot, line 129: can you complete the yaml example above?

```yaml
name: body
description: Body of the document
type: text
tokenizer: default
record: position
fieldnorms: true
fast:
  normalizer: lowercase
```

quickwit/quickwit-query/src/tokenizers.rs
```diff
@@ -160,6 +171,13 @@ pub fn get_quickwit_tokenizer_manager() -> &'static TokenizerManager {
     QUICKWIT_TOKENIZER_MANAGER.borrow()
 }
 
+pub fn get_quickwit_fastfield_normalizer_manager() -> &'static TokenizerManager {
+    /// Quickwits default tokenizer
+    static QUICKWIT_TOKENIZER_MANAGER: Lazy<TokenizerManager> =
```
fmassot, line 176:
```suggestion
    static QUICKWIT_FAST_FIELD_NORMALIZER_MANAGER: Lazy<TokenizerManager> =
```

quickwit/quickwit-query/src/tokenizers.rs
```diff
@@ -160,6 +171,13 @@ pub fn get_quickwit_tokenizer_manager() -> &'static TokenizerManager {
     QUICKWIT_TOKENIZER_MANAGER.borrow()
 }
 
+pub fn get_quickwit_fastfield_normalizer_manager() -> &'static TokenizerManager {
+    /// Quickwits default tokenizer
```
fmassot, line 175:
```suggestion
```

docs/configuration/index-config.md
```diff
@@ -321,7 +328,7 @@ expand_dots: false
 | `description` | Optional description for the field. | `None` |
 | `stored`    | Whether value is stored in the document store | `true` |
 | `indexed`   | Whether value is indexed | `true` |
-| `fast`   | Whether value is fast | `false` |
+| `fast`     | Whether value is stored in a fast field. The default behaviour for text is to store the original text unchanged. It can be configured via `{"normalizer": "lowercase"}`. ([See normalizers](#description-of-available-normalizers)) for a list of available normalizers. | `true` |
```
fmassot, line 331: can you update the mapping example above too?
I would also replace `{"normalizer": "lowercase"}` by `normalizer: lowercase` as we use YAML in the docs.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/date_time_type.rs
```diff
@@ -301,7 +301,7 @@ mod tests {
                 "description": "When the record was last updated.",
                 "input_formats": ["iso8601"],
                 "output_format": "rfc3339",
-                "precision": "seconds",
+                "precision": "second",
```
fmassot, line 304: do we have backward compatibility?

We also need to update all examples in the code.

quickwit/quickwit-indexing/src/actors/merge_executor.rs
```diff
@@ -367,6 +367,7 @@ impl MergeExecutor {
         let mut merged_index = Index::open(controlled_directory.clone())?;
         ctx.record_progress();
         merged_index.set_tokenizers(get_quickwit_tokenizer_manager().clone());
+        merged_index.set_fast_field_tokenizers(get_quickwit_fastfield_normalizer_manager().clone());
```
fmassot, line 370: The fast field tokenizers are useful only when indexing and merging right?

I saw two places where we don't set the fast field tokenizers:
- https://github.com/quickwit-oss/quickwit/blob/d7d4990133cbd4f0f2953d2b444f587abb8eb489/quickwit/quickwit-indexing/src/actors/merge_executor.rs#L515
- https://github.com/quickwit-oss/quickwit/blob/d7d4990133cbd4f0f2953d2b444f587abb8eb489/quickwit/quickwit-search/src/leaf.rs#L121

docs/configuration/index-config.md
fulmicoton: +1

docs/configuration/index-config.md
```diff
@@ -126,18 +126,25 @@ fieldnorms: true
...
+| `fast`     | Whether value is stored in a fast field. The fast field will contain the term ids and the dictionary. The default behaviour for `true` is to store the original text unchanged. The normalizers on the fast field is seperately configured. It can be configured via `{"normalizer": "lowercase"}`. ([See normalizers](#description-of-available-normalizers)) for a list of available normalizers. | `false` |
 
 #### **Description of available tokenizers**
 
 | Tokenizer     | Description   |
 | ------------- | ------------- |
-| `raw`         | Does not process nor tokenize text  |
-| `default`     | Chops the text on according to whitespace and punctuation, removes tokens that are too long, and converts to lowercase |
-| `en_stem`     |  Like `default`, but also applies stemming on the resulting tokens  |
+| `raw`         | Does not process nor tokenize text. Filters token larger than 255 bytes.  |
```
fulmicoton, line 135:
```suggestion
| `raw`         | Does not process nor tokenize text. Filters out token larger than 255 bytes.  |
```

docs/configuration/index-config.md
```diff
@@ -126,18 +126,25 @@ fieldnorms: true
...
+| `default`     | Chops the text on according to whitespace and punctuation, removes tokens that are too long, and converts to lowercase. Filters token larger than 40 bytes. |
```
fulmicoton, line 136:
```suggestion
| `default`     | Chops the text on according to whitespace and punctuation, removes tokens that are too long, and converts to lowercase. Filters out tokens larger than 40 bytes. |
```

docs/configuration/index-config.md
```diff
@@ -126,18 +126,25 @@ fieldnorms: true
...
+| `en_stem`     |  Like `default`, but also applies stemming on the resulting tokens. Filters token larger than 40 bytes.  |
```
fulmicoton, line 137:
```suggestion
| `en_stem`     |  Like `default`, but also applies stemming on the resulting tokens. Filters out tokens larger than 40 bytes.  |
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/date_time_type.rs
PSeitz: | do we have backward compatibility?

Do you mean in tantivy or in the metastore? tantivy accepts both currently

docs/configuration/index-config.md
fulmicoton, line 331: I think we need an entire example here. 
(That's where we pay for the idea of accepting bool and object in the same place)

```yaml
fast:
    normalize: lowercase
```

quickwit/quickwit-indexing/src/actors/merge_executor.rs
PSeitz: Yes, but also I'm not sure. I'm not a fan of having two tokenizers managers, it makes things complicated and doesn't solve anything currently

docs/configuration/index-config.md
PSeitz: This would be an alternative API.
I think it's nice when having both `enabled` and `normalizer`, but `enabled` on it's own is a little strange.
What do you think?

```yaml
fast:
    enabled: true
    normalizer: lowercase
```
```yaml
fast:
    enabled: true
```
```yaml
fast:
    enabled: false
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/date_time_type.rs
```diff
@@ -206,7 +206,7 @@ mod tests {
                     "rfc3339"
                 ],
                 "output_format": "unix_timestamp_secs",
-                "precision": "milliseconds",
+                "precision": "millisecond",
```
fulmicoton, line 209: I think we should revert this change in tantivy.

PSeitz: That's the PR: https://github.com/quickwit-oss/tantivy/pull/2051

`millisecond` or `milliseconds`, it's the same for me (as conclusion I don't think breaking the API is worth it)

`DatePrecision` to `DateTimePrecision` is the correct rename though

quickwit/quickwit-doc-mapper/src/default_doc_mapper/date_time_type.rs
fulmicoton: We have backward compatibility. I think we should revert that change, but in case Adrien has a good case for it I opened two PRs to fix the doc and simplify the tantivy stuff.

quickwit/quickwit-indexing/src/actors/merge_executor.rs
fulmicoton: Let's set it both tokenizer manager in all sites without thinking about whether it is useful or not.

> I'm not a fan of having two tokenizers managers, it makes things complicated and doesn't solve anything currently

You are being unfair. You can say you would have preferred another solution, but you cannot say it is not solving a problem.

Let me repeat what I explained several times in PRs:
Before this addition we had one tokenizer manager that was targetting both indexing and fast fields. You introduced a tokenizer called `lowercase` for fast fields.

Now here is the problem I explained two or three times...
We had tokenizers called:
- `default`: splitting + lowercasing
- `lowercase`: not splitting + lowercasing
- `en_stem`: splitting + stemming.

This is quite confusing for a user isn't it? Normalizer and tokenizer are living in the same namespace with no hint in their names of who is doing what.

Another solution would have been to give more explicit names to tokenizers.
(note you did not do that)

I prefer to avoid any footgun and have two tokenizer managers.
You may argue that users may still populate the tokenizer manager correctly...
Or that the possibly colliding names of the tokenizer can be confusing...
But with the default settings of tantivy, it does remove a massive footgun.

PSeitz: Sorry, I think I my statement wasn't clear, I was only talking about tantivy's two tokenizer API, not separating tokenizer and normalizers.

I agree having `normalizer` and `tokenizer` separated in quickwit is good and solves the problem of mixing them. The validation of them and mapping to the actual tokenizer name (via `get_name`) is done on quickwit side, so they are separated in the API and can't be mixed. In that context the tantivy's two tokenizer API doesn't solve anything or provide a functionality that is required. I see two tokenizer manager as a bigger footgun, than internally mapping a tokenizer to a normalizer or the other way around (this is not a quickwit user API anymore)

docs/configuration/index-config.md
fmassot: Yes, I'm not fond of having `enabled: false` alone. But we then need two examples.

PSeitz: We also have other field types that can't have a normalizer, which are configured with `fast: true`

quickwit/quickwit-doc-mapper/src/default_doc_mapper/date_time_type.rs
fulmicoton: Let's revert the s part.
Can you review this PR and point to the new main?
https://github.com/quickwit-oss/tantivy/pull/2065

docs/configuration/index-config.md
PSeitz: Update both

docs/configuration/index-config.md
PSeitz: Updated example and replaced JSON with YAML

docs/configuration/index-config.md
```diff
@@ -126,18 +128,25 @@ fieldnorms: true
...
 
 #### **Description of available tokenizers**
 
 | Tokenizer     | Description   |
 | ------------- | ------------- |
-| `raw`         | Does not process nor tokenize text  |
-| `default`     | Chops the text on according to whitespace and punctuation, removes tokens that are too long, and converts to lowercase |
-| `en_stem`     |  Like `default`, but also applies stemming on the resulting tokens  |
+| `raw`         | Does not process nor tokenize text. Filters out tokens larger than 255 bytes.  |
+| `default`     | Chops the text on according to whitespace and punctuation, removes tokens that are too long, and converts to lowercase. Filters out tokens larger than 40 bytes. |
```
fmassot, line 138: the limit is now 255 I believe.

fmassot (approved):

docs/configuration/index-config.md
PSeitz: It wasn't for `default`, I updated them all to have the 255 limit

PR #3410, opened 2023-05-25
use QueryAst in jaeger compat layer

### LINKED ISSUE #3318: Rely on the QueryAST in the tracer.
Right now the tracing endpoint builds a text query.
Now that we have a proper QueryAST it would make more sense to work over the QueryAST

trinity-1686a:
### Description

fix #3318 

### How was this PR tested?

updated unit tests


guilload (approved):
LGTM. Thanks for the cleanup job!

PR #3424, opened 2023-05-28
Introduce REST client timeouts

### LINKED ISSUE #3137: longer timeout for ingestion query with wait/force commit
Currently the timeout is always 10s, which is fine to just push data, but when waiting for a commit, it can easily take a good minute (commit timeout), even more if too much data got buffered. The last ingest query made by the cli should be longer to account for that, and possibly tunable

imotov:
### Description

Adds ability to specify operation timeouts for the rust client as well as different CLI commands. Takes a first stab at more conservative timeout defaults. These timeouts are still pretty much guesstimates at this point, but at least we now have a way to adjust default and override them from CLI.

Fixes #3137

### How was this PR tested?

By running automated tests as well as using our large examples from tutorial to see if they fail with new defaults.

### DIFF
--- a/quickwit/quickwit-cli/src/lib.rs
+++ b/quickwit/quickwit-cli/src/lib.rs
@@ -104,8 +104,6 @@ fn client_args() -> Vec<Arg> {
 #[derive(Debug, Eq, PartialEq)]
 pub struct ClientArgs {
     pub cluster_endpoint: Url,
-    /// The outer option represents the presense of the argument,
-    /// the inner option represents the presence of timeout.
     pub connect_timeout: Option<Timeout>,
     pub timeout: Option<Timeout>,
     pub commit_timeout: Option<Timeout>,
--- a/quickwit/quickwit-cli/src/lib.rs
+++ b/quickwit/quickwit-cli/src/lib.rs
@@ -20,10 +20,11 @@
 #![deny(clippy::disallowed_methods)]
 
 use std::collections::HashSet;
+use std::str::FromStr;
 use std::time::Duration;
 
 use anyhow::{bail, Context};
-use clap::{arg, Arg};
+use clap::{arg, Arg, ArgMatches};
 use dialoguer::theme::ColorfulTheme;
 use dialoguer::Confirm;
 use once_cell::sync::Lazy;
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-cli/src/lib.rs
```diff
@@ -73,13 +76,127 @@ fn config_cli_arg() -> Arg {
...
+            .env("QW_CLUSTER_ENDPOINT")
+            .required(false)
+            .display_order(1)
+            .global(true),
+        Arg::new("timeout")
+            .long("timeout")
+            .help("Duration of the timeout.")
+            .required(false)
+            .global(true)
+            .display_order(1),
```
fmassot, line 92: display order 2?

quickwit/quickwit-cli/src/lib.rs
```diff
@@ -73,13 +76,127 @@ fn config_cli_arg() -> Arg {
...
+        Arg::new("connect-timeout")
+            .long("connect-timeout")
+            .help("Duration of the connect timeout.")
+            .required(false)
+            .global(true)
+            .display_order(1),
```
fmassot, line 98: display order 3?

quickwit/quickwit-cli/src/lib.rs
```diff
@@ -73,13 +76,127 @@ fn config_cli_arg() -> Arg {
...
+    ]
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct ClientArgs {
+    pub cluster_endpoint: Url,
+    /// The outer option represents the presense of the argument,
+    /// the inner option represents the presence of timeout.
+    pub connect_timeout: Option<Option<Duration>>,
```
fmassot, line 107: It's weird to have option of option. Is it possible to simplify that?

imotov: I tried and this seems to be the lesser evil. The issue is that we need to represent 2 situations here. 1) whether or not the user specified the commit parameters and 2) where we have a timeout or we have no timeout. We could have teach CLI the default timeouts but then we wouldn't be able to reuse the same flag `--timeout` for different timeouts such as search timeout in case of search and ingest_timeout in case of ingest, so we would be pushing this complexity to the user.

fmassot: I see. 
But I don't understand why we need an option of option for `connect_timeout`. Why not having `pub const DEFAULT_CLIENT_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);`

The user can of course override this. But I don't see the need of an option. In the client args, we can only have an `Option<Duration>` for this variable. Or am I missing something?

imotov: The current setting for connect_timeout is not set. So, since I am setting it pretty aggressively down in this PR I wanted to have an ability to undo that if it doesn't work for some reason, which would be expressed as Some(None) on the CLI level. The main reason I am setting it is because I also increase commit timeout all the way up. It would be a bad user experience to wait half an hour for ingest if we cannot even connect properly.

fmassot: I think it was bad to have not set the `connect_timeout` before, you are fixing that so I suggest to remove the option. 5 seconds is maybe a bit too agressive, something like 10 - 15 seconds sounds reasonable.

fmassot (approved):

PR #3426, opened 2023-05-29
Makes it possible to search bytes using base16.

### LINKED ISSUE #3385: [Tracing] Allow users to query trace and spans IDs with hex encoded IDs
We should also output them in hex.

fulmicoton:
The logic goes
- base16 first
- base64 then

If someone really wants to search something in base64, that happens to be base16, they can prefix the string with a ! to hint that this is base64.

Closes #3385


quickwit/Cargo.toml
```diff
@@ -46,6 +46,7 @@ async-speed-limit = "0.4"
 async-trait = "0.1"
 atty = "0.2"
 backoff = { version = "0.4", features = ["tokio"] }
+base16 = "0.2.1"
```
guilload, line 49: Seems like the `hex` crate is more commonly used and was updated more recently.

quickwit/quickwit-query/src/json_literal.rs
```diff
@@ -159,6 +160,31 @@ impl<'a> InterpretUserInput<'a> for tantivy::DateTime {
     }
 }
 
+/// Lenient base64 engine that allow user to use padding or not.
```
guilload, line 163:
```suggestion
/// Lenient base64 engine that allows users to use padding or not.
```

guilload (approved):

quickwit/Cargo.toml
fulmicoton: switched to hex

PR #3427, opened 2023-05-29
Default docmapper with dynamic mode.

### LINKED ISSUE #2910: Make dynamic mode the default, and allow for the mapping to be empty
With the addition of fast fields, it is more and more appealing to make the dynamic mode the default, and to allow for 
the mapping to be unspecified.

Users could then supply quickwit a very minimal document mapping:

```yaml
version: 0.4
index_id: my-dynamic-index
```

The bummer of course is the absence of timestamps at this point :-(

fulmicoton:
Makes the dynamic mode the default, and allow unspecified doc mapping

Accepting sort by score even if one of the default field does not have fieldnorm.
Minor code cleanup

Closes #2910

### DIFF
--- a/docs/configuration/index-config.md
+++ b/docs/configuration/index-config.md
@@ -376,9 +376,9 @@ field_mappings:
 The `mode` describes how Quickwit should behave when it receives a field that is not defined in the field mapping.
 
 Quickwit offers you three different modes:
-- `lenient` (default value): unmapped fields are dismissed by Quickwit.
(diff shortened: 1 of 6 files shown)


fmassot (approved):

fmassot:
@fulmicoton note that an index config like this one:

```yaml
version: 0.4
index_id: my-dynamic-index

```

will not be created. The returned error message from the API is 

```
"message": "Invalid config: Failed to read YAML file.: missing field `doc_mapping`."
```

fulmicoton:
@fmassot yes it used to be possible in my previous PR, but I removed that.

Should I readd it?

PR #3428, opened 2023-05-29
Use debug level for busy detector and add metric.

fmassot:
Metric tested with a local prometheus + grafana

### DIFF
--- a/quickwit/quickwit-cli/src/metrics.rs
+++ b/quickwit/quickwit-cli/src/metrics.rs
@@ -1,0 +1,41 @@
+// Copyright (C) 2023 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-cli/src/metrics.rs
```diff
@@ -0,0 +1,41 @@
...
+
+pub struct CliMetrics {
+    pub thread_unpark_duration_microseconds: HistogramVec<0>,
+}
+
+impl Default for CliMetrics {
+    fn default() -> Self {
+        CliMetrics {
+            thread_unpark_duration_microseconds: new_histogram_vec(
+                "thread_unpark_duration_milliseconds",
```
guilload, line 31:
```suggestion
                "thread_unpark_duration_microseconds",
```

quickwit/quickwit-cli/src/metrics.rs
```diff
@@ -0,0 +1,41 @@
...
+                "Duration during tokio main thread was unparked",
```
guilload, line 32:
```suggestion
                "Duration for which the tokio main thread is unparked.",
```

guilload (approved):

quickwit/quickwit-cli/src/metrics.rs
fmassot: thanks!

PR #3431, opened 2023-05-29
Index and search span status

guilload:
### Description
- index span status
- modify Jaeger query builder to allow `error=true/false` queries

### How was this PR tested?
- added unit tests
- updated existing unit tests

### DIFF
--- a/quickwit/quickwit-jaeger/src/integration_tests.rs
+++ b/quickwit/quickwit-jaeger/src/integration_tests.rs
@@ -128,21 +128,25 @@ async fn test_otel_jaeger_integration() {
             .await
             .unwrap()
             .into_inner();
-        assert_eq!(get_operations_response.operations.len(), 3);
+        assert_eq!(get_operations_response.operations.len(), 4);
         assert_eq!(
             get_operations_response.operations,
             vec![
                 Operation {
-                    name: "stage_splits".to_string(),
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-jaeger/src/integration_tests.rs
```diff
@@ -226,7 +230,47 @@ async fn test_otel_jaeger_integration() {
...
+            tags: HashMap::from([("error".to_string(), "true".to_string())]),
+            start_time_min: None,
+            start_time_max: None,
+            duration_min: None,
+            duration_max: None,
+            num_traces: 10,
+        };
+        let find_trace_ids_request = FindTraceIDsRequest { query: Some(query) };
+        let find_trace_ids_response = jaeger_service
+            .find_trace_i_ds(tonic::Request::new(find_trace_ids_request))
```
fmassot, line 248: is `find_trace_i_ds` the right name? it seems like `find_trace_ids` is.

guilload: codegen

fmassot (approved):
Approved. Though I think we need to test that in Jaeger UI, is that right?

guilload:
Seems to be working. I generated some `error!` spans in some of our traces and the queries `error:true/false` are working.

PR #3433, opened 2023-05-29
Store span and trace IDs in byte fields

guilload:
### Description
Store span and trace IDs in byte fields rather than text fields.

### How was this PR tested?
- Updated unit tests
- Tested in Jaeger UI


quickwit/quickwit-opentelemetry/src/otlp/span_id.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
+
+    pub fn to_vec(&self) -> Vec<u8> {
+        self.0.to_vec()
+    }
+
+    pub fn base64_display(&self) -> Base64Display<'_, '_, GeneralPurpose> {
```
fmassot, line 44: why keeping the base64 display?

fmassot (approved):
With #3417 I'm not sure if it's better to query with base64 or base16. 

see https://github.com/quickwit-oss/quickwit/blob/81d2957842be6f4749fa41c2618d535bddadeb11/quickwit/quickwit-jaeger/src/lib.rs#L294

PR #3434, opened 2023-05-29
Add env variable to set actor's heartbeat. Distinguish heartbeat and …

fmassot:
- Add env variable to customize the heartbeat. Useful for internal usage.
- Default heartbeat is set to 30 seconds.
- Separates duration intervals that should not be bound to the actor heartbeat.

Drawback: I used a oncecell to store the heartbeat but it's a bit annoying to use it throughout the code. I don't know if it's possible to use something else.

### DIFF
--- a/quickwit/quickwit-control-plane/src/scheduler.rs
+++ b/quickwit/quickwit-control-plane/src/scheduler.rs
@@ -26,7 +26,7 @@ use std::time::{Duration, Instant};
 use anyhow::Context;
 use async_trait::async_trait;
 use itertools::Itertools;
-use quickwit_actors::{Actor, ActorContext, ActorExitStatus, Handler, HEARTBEAT};
+use quickwit_actors::{Actor, ActorContext, ActorExitStatus, Handler};
 use quickwit_cluster::{Cluster, ClusterMember};
 use quickwit_config::service::QuickwitService;
 use quickwit_config::SourceConfig;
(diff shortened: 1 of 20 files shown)


quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -42,10 +42,21 @@ use crate::indexing_plan::{
 };
 use crate::{NotifyIndexChangeRequest, NotifyIndexChangeResponse};
 
+/// Interval between two controls (or checks) of the desired plan VS running plan.
+const CONTROL_PLAN_LOOP_INTERVAL: Duration = if cfg!(any(test, feature = "testsuite")) {
```
fmassot, line 46: Was based on the heartbeat. I just resuse the heartbeat default value.

quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -42,10 +42,21 @@ use crate::indexing_plan::{
...
+
+/// Interval between two scheduling of indexing plans. No need to be faster than the
+/// control plan loop.
+// Note: it's currently not possible to define a const duration with
+// `CONTROL_PLAN_LOOP_INTERVAL * number`.
 const REFRESH_PLAN_LOOP_INTERVAL: Duration = if cfg!(any(test, feature = "testsuite")) {
     Duration::from_secs(3)
 } else {
-    Duration::from_secs(60 * 5)
+    Duration::from_secs(60)
```
fmassot, line 59: Ok... it's just me here... I prefer to check this every minute... one refresh is making one call. I can revert if needed.

quickwit/quickwit-janitor/src/janitor_service.rs
```diff
@@ -44,27 +44,14 @@ impl JanitorService {
...
-    fn harvest_health(&self) -> Health {
-        let mut janitor_health = Health::Healthy;
-
-        for supervisable in self.supervisables() {
-            let actor_health = supervisable.harvest_health();
-
-            if actor_health != Health::Healthy {
-                janitor_health = actor_health;
-            }
+    fn is_healthy(&self) -> bool {
```
guilload, line 47: nit
```rust
fn is_healthy(&self) -> bool {
        self.delete_task_service_handle.state() != ActorState::Failure
            && self.garbage_collector_handle.state() != ActorState::Failure
            && self.retention_policy_executor_handle.state() != ActorState::Failure
    }
    ```

guilload (approved):

PR #3437, opened 2023-05-30
Ensure that index names with 255 characters work

### LINKED ISSUE #3308: Inconsistent handling of long index names
**Describe the bug**
A file-based metastore allows index names up to 255 characters long, while PostgreSQL metastore only allocates 50 characters for the index name. 

1. Try create an index with more than 50 characters in its name using qw instance with PostgreSQL metastore.

**Expected behavior**
The index name limit should be consistent.

**Configuration:**
All version are affected as far as I can tell.

imotov:
### Description

Expands PostgreSQL tables to accommodate long index names up to 255 characters. Also increases the size of incarnation_id to full ulid size.

Fixes #3308

### How was this PR tested?

Added one unit-test and one integration test.


quickwit/quickwit-integration-tests/src/tests/index_tests.rs
```diff
@@ -352,6 +354,114 @@ async fn test_commit_modes() {
     sandbox.shutdown().await.unwrap();
 }
 
+#[tokio::test]
+async fn test_very_large_index_name() {
```
guilload, line 358: Do we really need an integration test for that? They're slow and hard to maintain, I'd rather have a bunch of some unit tests in `quickwit_config::index_config` and `quickwit_metastore` (and maybe even `quickwit_core::index`)

guilload (approved):

quickwit/quickwit-integration-tests/src/tests/index_tests.rs
imotov: The rationale for adding this test was that I wanted to ensure that the basic functionality works in the future no matter what. A bunch of unit tests could protect us form known unknowns, such as modification of metastore. An integration test here would add some protection against unknown unknowns, such as somebody adds something like SSD cache and forgets to properly handle directories for large index names while working on the directory structure.

It is possible to catch such unknown unknowns on the unit-test level or by slightly modifying an existing integration test. In order to do that we need to add a better test sandbox (and better cluster sandbox), that would create different randomized setups, with randomly created indices having randomly selected names, randomly selected storage types for both data and metadata, and in case of the cluster sandbox - different cluster topology. 

However, adding test randomization without having a robust CI infrastructure will be detrimental and will only cause developer frustration since it will basically add a lot of flaky tests without any means of analyzing, prioritizing and making sure that they are getting fixed. Therefore I thought an integration test would be a good temporary solution for now until we get there. 

I don't think I can add an equivalent unit tests in the current test infrastructure, but I am ok with removing that integration test for now if these arguments are not convincing.

PR #3446, opened 2023-05-30
Add `StorageConfig` and `MetastoreConfig` objects in index config

guilload:
### Description
Introduce `StorageConfig` and `MetastoreConfig` objects. The storage config will allow us to configure virtual vs. path-style URLs and disable the multi-objects delete API for GCS.

The `MetastoreConfig` is only useful for setting the max num of connections for Postgres for now, but it can host more settings in the future.

### How was this PR tested?
In progress

### DIFF
--- a/quickwit/Cargo.lock
+++ b/quickwit/Cargo.lock
@@ -3494,9 +3494,9 @@ dependencies = [
 
 [[package]]
 name = "openssl"
-version = "0.10.53"
+version = "0.10.54"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "12df40a956736488b7b44fe79fe12d4f245bb5b3f5a1f6095e499760015be392"
+checksum = "69b3f656a17a6cbc115b5c7a40c616947d213ba182135b014d6051b73ab6f019"
 dependencies = [
  "bitflags",
  "cfg-if",
@@ -4897,6 +4897,7 @@ dependencies = [
  "quickwit-rest-client",
  "quickwit-search",
  "quickwit-serve",
+ "quickwit-storage",
  "rand 0.8.5",
  "reqwest",
  "serde",
@@ -5269,6 +5270,7 @@ dependencies = [
  "proptest",
  "quickwit-aws",
  "quickwit-common",
+ "quickwit-config",
  "rand 0.8.5",
  "regex",
  "serde",
--- a/quickwit/quickwit-config/src/metastore_config.rs
+++ b/quickwit/quickwit-config/src/metastore_config.rs
@@ -1,0 +1,231 @@
+// Copyright (C) 2023 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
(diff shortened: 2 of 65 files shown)


quickwit/Cargo.lock
```diff
@@ -78,6 +78,12 @@ version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
 
+[[package]]
+name = "android-tzdata"
```
fulmicoton, line 82: if someone is wondering this is not a real dep. It only is enabled when targetting android.

quickwit/quickwit-config/src/metastore_config.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    pub fn as_postgres(&self) -> Option<&PostgresMetastoreConfig> {
+        match self {
+            Self::PostgreSQL(postgres_metastore_config) => Some(postgres_metastore_config),
+            _ => None,
+        }
+    }
+
+    pub fn redact(&mut self) {
+        // TODO: Implement this method when we end up storing secrets in the
```
fmassot, line 63: this is the case for postgresql

quickwit/quickwit-core/src/lib.rs
```diff
@@ -68,7 +68,7 @@ mod tests {
         }
         // Now delete the index.
         let index_service =
-            IndexService::new(test_sandbox.metastore(), StorageUriResolver::for_test());
+            IndexService::new(test_sandbox.metastore(), StorageResolver::unconfigured());
```
fmassot, line 71: is `StorageResolver::unconfigured()` really unconfigured? Sound like `StorageUriResolver::for_test()` is doing more and that's what I would expect for a test config.

quickwit/quickwit-metastore/src/metastore_resolver.rs
```diff
@@ -20,163 +20,169 @@
...
-        let test_database_url = env::var("TEST_DATABASE_URL").unwrap_or_else(|_| {
-            "postgres://quickwit-dev:quickwit-dev@localhost/quickwit-metastore-dev".to_string()
-        });
-        let (_uri_protocol, uri_path) = test_database_url.split_once("://").unwrap();
-        for protocol in &["postgres", "postgresql"] {
-            let postgres_uri = Uri::from_well_formed(format!("{protocol}://{uri_path}"));
-            metastore_resolver.resolve(&postgres_uri).await.unwrap();
-        }
-    }
+    // #[tokio::test]
```
fmassot, line 163: uncomment or rm

fmassot (approved):

quickwit/quickwit-config/src/metastore_config.rs
guilload: I have not added the database password in the config in this PR.

quickwit/quickwit-core/src/lib.rs
guilload: They're equivalent. They both register the same factories with no config for the former and the default config for the latter. No config is also equivalent to the default config after this change.

PR #3218, opened 2023-04-24
Auto tagging of partition key fields

saroh:
Refs #3000

### Description

Partition key fields are added to tag fields.

:warning: Fails retrocompatibility tests where the partition key is not defined in the field mappings, hence cannot be validated as a tag. Cf third commit. Is this ok or not? 

### How was this PR tested?

UT.

PR #3449, opened 2023-05-31
Saroh 3000 auto tag partition key

fulmicoton:
Importing @saroh's branch 
https://github.com/quickwit-oss/quickwit/pull/3218 for history


quickwit/quickwit-config/src/index_config/mod.rs
```diff
@@ -427,7 +427,7 @@ impl TestableForRegression for IndexConfig {
             store_source: true,
             mode: ModeType::Dynamic,
             dynamic_mapping: None,
-            partition_key: Some("tenant".to_string()),
+            partition_key: Some("tenant_id".to_string()),
```
fulmicoton, line 430: O_O nice catch

fulmicoton:
@saroh good to go. I made some minor modifications but they were for the most part additions motivated by another ticket rather than corrections.

PR #3450, opened 2023-05-31
Change datetime precision labels in doc

fulmicoton:


fulmicoton:
closing... removed the added s and added aliases instead.

PR #3454, opened 2023-05-31
Add default search field for otel logs index.

fmassot:
I did not find a good default search field for the otel traces index.

I can change that if @guilload you have a good idea.

Tested locally.


guilload (approved):
That's a good start. How do you send OTLP logs to Quickwit for testing?

fmassot:
Good question! I made this example exactly for that: https://quickwit.io/docs/log-management/send-logs/using-otel-collector

You will note that we need to add into the json  `observed_time_unix_nano`  so that the ingest does not fail.
If not, the OTEL collector will send the 0 value to quickwit... and it will fail which is expected as now `observed_timestamp_nanos` is a datetime.

I checked the OTEL log model and I think it's fine to have this additional constraint.

PR #3457, opened 2023-05-31
Fix error on log on SIGINT signal

guilload:
### Description
The indexer fails to finalize because the serializer has already quit when handling the SIGINT signal. I don't think this change prevents us from discovering bugs in the future because actors already do a good job at surfacing their own errors.

### How was this PR tested?
Before:
Screenshot 2023-05-31 at 9 02 31 AM

After:
Screenshot 2023-05-31 at 9 06 43 AM

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexer.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexer.rs
@@ -388,8 +388,9 @@ impl Actor for Indexer {
             | ActorExitStatus::Failure(_)
             | ActorExitStatus::Panicked => return Ok(()),
             ActorExitStatus::Quit | ActorExitStatus::Success => {
-                self.send_to_serializer(CommitTrigger::NoMoreDocs, ctx)
-                    .await?;
+                let _ = self
+                    .send_to_serializer(CommitTrigger::NoMoreDocs, ctx)
+                    .await;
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-indexing/src/actors/indexer.rs
```diff
@@ -388,8 +388,9 @@ impl Actor for Indexer {
             | ActorExitStatus::Failure(_)
             | ActorExitStatus::Panicked => return Ok(()),
             ActorExitStatus::Quit | ActorExitStatus::Success => {
-                self.send_to_serializer(CommitTrigger::NoMoreDocs, ctx)
-                    .await?;
+                let _ = self
```
fulmicoton, line 391: ( O_O ) I don't understand the difference?
Are there files missing?

fulmicoton (approved):

quickwit/quickwit-indexing/src/actors/indexer.rs
fmassot: In the case where the serializer has already exited, the `send_to_serializer` function will return a `TrySendError::Disconnected` and this will make the `finalizer` fail. The log is quite ugly and I agree with Adrien on this.

PR #3458, opened 2023-05-31
Store span and trace IDs in byte fields (logs)

guilload:
### Description
- [Store span and trace IDs in byte fields (logs)](https://github.com/quickwit-oss/quickwit/commit/25211ea9e0443a1219a81c278fd4cdd309a2279a)
- [Handle null timestamps](https://github.com/quickwit-oss/quickwit/commit/c6232ea5f13f1106b779f8634fa6f43a0a79f70a)
- [Skip serializing empty/null fields](https://github.com/quickwit-oss/quickwit/commit/90ede2043ada3cd3ff6fa967a1d5c8e1c4cd7638)

### How was this PR tested?
I'll add more tests in a follow-up PR. This needs the same refactor as the traces part to implement better tests.


guilload:
I tested the log ingest manually with [this script](https://gist.github.com/guilload/c603197deef1524817f393d1597e3d8d), though.

fmassot (approved):

PR #3460, opened 2023-05-31
Add features to telemetry info.

fmassot:
Notable changes:
- Add features to telemetry info
- Each event send the activated features and the services that are running.

I need to make a small test.


quickwit/quickwit-telemetry/src/lib.rs
```diff
@@ -52,7 +52,7 @@ fn get_telemetry_sender_singleton() -> &'static TelemetrySender {
 /// We voluntarily use an enum here to make it easier for reader
 /// to audit the type of information that is send home.
 pub async fn send_telemetry_event(event: TelemetryEvent) {
-    get_telemetry_sender_singleton().send(event).await
+    TELEMETRY_SENDER.get().map(|sender| sender.send(event));
```
fulmicoton, line 55:
```suggestion
    if let Some(telemetry_sender) = TELEMETRY.get() {
        sender.send(event);
    }
```

It is really a if here. We don't return an expression.

fulmicoton (approved):
approved, but we need to fix the bug.

PR #3467, opened 2023-06-01
Add tests for bulk delete and virtual-hosted-style vs path-style access

guilload:
### Description
Added tests and fixed one bug.

### How was this PR tested?
Tested on Amazon S3, Alibaba, GCS, and R2.

### DIFF
--- a/quickwit/Makefile
+++ b/quickwit/Makefile
@@ -16,10 +16,11 @@ fix:
 # `make test-all` starts the Docker services and runs all the tests.
 # `make -k test-all docker-compose-down`, tears down the Docker services after running all the tests.
 test-all:
-	QW_S3_ENDPOINT=http://localhost:4566 \
-	AWS_REGION=us-east-1 \
 	AWS_ACCESS_KEY_ID=ignored \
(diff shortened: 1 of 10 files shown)


quickwit/Makefile
```diff
@@ -16,10 +16,11 @@ fix:
 # `make test-all` starts the Docker services and runs all the tests.
 # `make -k test-all docker-compose-down`, tears down the Docker services after running all the tests.
 test-all:
-	QW_S3_ENDPOINT=http://localhost:4566 \
-	AWS_REGION=us-east-1 \
 	AWS_ACCESS_KEY_ID=ignored \
 	AWS_SECRET_ACCESS_KEY=ignored \
+	AWS_REGION=us-east-1 \
+	QW_S3_ENDPOINT=http://localhost:4566 \
+	QW_S3_FORCE_PATH_STYLE_ACCESS=1 \
```
fmassot, line 23: so we never use virtual host style in tests?

guilload: Not against LocalStack. It did not seem to work.

PR #3475, opened 2023-06-02
Fix CLI options in the generated documentation

### LINKED ISSUE #3474: Autogenerated CLI options are mangled in a browser
The auto-generated markdown for CLI options doesn't look good in the browser:

image

The markdown looks reasonable, but for some reason it is not handled correctly when the final HTML is generated.

imotov:
### Description

Changes the way the options are generated in markdown to make them look more readable in the browser.

Fixes #3474

### How was this PR tested?

Eyeballing.


guilload (approved):

quickwit/quickwit-cli/src/generate_markdown.rs
```diff
@@ -141,7 +141,7 @@ fn markdown_for_command_helper(
                 "".to_string()
             };
             println!(
-                "`--{}` {}{} \\", // Insert line break after option
+                "`--{}` {}{}  ", // Insert line break after option
```
guilload, line 144: I can't remember the rendering issue we got with using `  ` as a line break, maybe something related to Docusaurus, but we ended up using `\` to fix it.

imotov: Spoke with @guilload about it. That solution works but it involves trailing spaces, which some editors remove, which will most likely cause this issue to come back eventually. I will try to convert options to a table or some other more reliable structure as a follow up.

PR #3479, opened 2023-06-02
Handle correclty std::env::var errors.

fmassot:
Tested locally.

Now the log appears only if `QW_ACTOR_HEARTBEAT_SECS` cannot be parsed.

```console
2023-06-02T07:19:42.341Z  WARN quickwit_actors: Failed to parse `QW_ACTOR_HEARTBEAT_SECS=0A` in seconds > 0, using default heartbeat (30 seconds).
```

### DIFF
--- a/quickwit/quickwit-actors/src/lib.rs
+++ b/quickwit/quickwit-actors/src/lib.rs
@@ -91,21 +91,25 @@ fn heartbeat_from_env_or_default() -> Duration {
         // We use a shorter heartbeat to reduce the time running unit tests.
         return Duration::from_millis(500);
     }
-    if let Ok(actor_hearbeat_secs_str) = std::env::var("QW_ACTOR_HEARTBEAT_SECS") {
-        if let Ok(actor_hearbeat_secs) = actor_hearbeat_secs_str.parse::<NonZeroU64>() {
-            info!("Set the actor heartbeat to {actor_hearbeat_secs} seconds.");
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-actors/src/lib.rs
```diff
@@ -91,21 +91,25 @@ fn heartbeat_from_env_or_default() -> Duration {
         // We use a shorter heartbeat to reduce the time running unit tests.
         return Duration::from_millis(500);
     }
-    if let Ok(actor_hearbeat_secs_str) = std::env::var("QW_ACTOR_HEARTBEAT_SECS") {
-        if let Ok(actor_hearbeat_secs) = actor_hearbeat_secs_str.parse::<NonZeroU64>() {
-            info!("Set the actor heartbeat to {actor_hearbeat_secs} seconds.");
-            return Duration::from_secs(actor_hearbeat_secs.get());
-        } else {
+    match std::env::var("QW_ACTOR_HEARTBEAT_SECS") {
```
guilload, line 94: Well-intentioned, but at the end of the day, it will only be used by advanced users, most likely us exclusively. So I'd panic and call it a day.

guilload (approved):

quickwit/quickwit-actors/src/lib.rs
fmassot: Agreed but... the program will not exit with a panic in the initialization. We could parse it at the beginning as we do for the config.

PR #3502, opened 2023-06-05
Use Azure storage account from storage config

guilload:
### Description
The Azure storage account is the equivalent of an AWS region, and it's better not to hardcode them in URIs stored in the metastore.

This PR adds the ability to configure the storage account via the env var `QW_AZURE_STORAGE_ACCOUNT` or storage config key `storage.azure.account`.

### How was this PR tested?
- `cargo test --manifest-path quickwit/Cargo.toml -p quickwit-storage --all-features -- --nocapture`
- `make test-all`


fulmicoton (approved):

fulmicoton:
I'll taking the liberty to merge that for 0.6.0, the PR looked good. I hope that's ok.

PR #3511, opened 2023-06-06
Add Pipfile and instructions for running REST API tests

guilload:
### Description
Per title.

### How was this PR tested?
Ran commands described in README.

### DIFF
--- a/quickwit/rest-api-tests/Pipfile
+++ b/quickwit/rest-api-tests/Pipfile
@@ -1,0 +1,13 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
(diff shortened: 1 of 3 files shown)


quickwit/rest-api-tests/Pipfile
```diff
@@ -0,0 +1,13 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+requests = "*"
```
fulmicoton, line 7: maaaaaaaaaan it seems everytime I wake up there is a different way to describe dependencies and environemnt in python.

This one is nice. It looks like rust :D

fulmicoton (approved):

Issue #3516, opened 2023-06-07
0.6 documentation still refers to 0.6 in the future tense 


imotov:
**Describe the bug**
The 0.6 documentation still says:
```
There are a few limitations on the log management setup in Quickwit 0.6:
- Aggregations are not available on sparse fields and JSON field, this will be fixed in 0.6. This means that only the timestamp field can support aggregations.
```
It appears on 2 pages:
https://quickwit.io/docs/distributed-tracing/otel-service/#known-limitations
https://quickwit.io/docs/log-management/otel-service/#known-limitations

```
You will soon be able to do aggregations on dynamic fields (planned for 0.6).
```
appears on 
https://quickwit.io/docs/log-management/send-logs/using-fluentbit/#further-improvements

PR #3521, opened 2023-06-07
Update docs on 0.5 limitations

evanxg852000:
Updated docs on limitations not lifted in 0.6

Closes https://github.com/quickwit-oss/quickwit/issues/3516


imotov (approved):
LGTM. Thanks for taking care of it!

fmassot (changes requested):
@evanxg852000 can you update the index config too of the trace and log index?

config/tutorials/otel-trace/index-config.yaml
```diff
@@ -5,7 +5,7 @@
 
 version: 0.6
 
-index_id: otel-trace-v0
+index_id: otel-traces-v0_6
```
fmassot, line 8: can you update the whole index config? I think the mapping is not up to date

docs/log-management/otel-service.md
```diff
@@ -19,14 +19,14 @@ indexer:
...
-Quickwit sends OpenTelemetry logs into the `otel-logs-v0` index which is automatically created if you enable the OpenTelemetry service.
+Quickwit sends OpenTelemetry logs into the `otel-logs-v0_6` index which is automatically created if you enable the OpenTelemetry service.
 The doc mapping of this index described below is derived from the [OpenTelemetry logs data model](https://opentelemetry.io/docs/reference/specification/logs/data-model/).
 
 ```yaml
 
 version: 0.6
 
-index_id: otel-logs-v0
+index_id: otel-logs-v0_6
```
fmassot, line 29: same thing, the mapping is not up to date I believe.

config/tutorials/otel-trace/index-config.yaml
fmassot: This config file seems useless, do you know where it's used?

config/tutorials/otel-logs/index-config.yaml
```diff
@@ -5,30 +5,68 @@
```
fmassot, line 5: actually, this log index config file is used in a tutorial that does not use the otel schema. It's a simpler use case. Maybe we should review the tutorial so that we ingest to the otel index directly?

evanxg852000: I have not been able to spot the tutorial you are referring to.

evanxg852000: I am updating that tutorial here https://github.com/quickwit-oss/quickwit/pull/3534

config/tutorials/otel-trace/index-config.yaml
fmassot: if this index config is not used in a tutorial, can you remove it?

config/tutorials/otel-logs/index-config.yaml
fmassot: Cool! Can you revert the index config update then in this PR and do it in #3534 ?

evanxg852000: Yes that's what i did to avoid conflicts only the tutorials files are now changed

config/tutorials/otel-trace/index-config.yaml
fmassot: But then we need to maintain this file. The trace index config is automatically created. You can use the API to get the index config for example. 
What is shown in the docs should the index config written in the rust code.

Currently the directory is `config/tutorials` and if it's not used for tutorials then why keeping it? It's confusing.

config/tutorials/otel-logs/index-config.yaml
fmassot: But I still see changes here. In the tutorial, I think we should not create the index and use the OTEL one created by default. So we can remove this index config file.

Issue #1585, opened 2022-06-01
Link to the nightly build binaries on the main-branch docs


fmassot:
On quickwit.io, the `main-branch` install docs points to the latest release binaries. It could be nice to point to the nightly binaries.

evanxg852000:
@fmassot RIght now I think the easiest to do is to have a Nighlty build section in the  docs with the links, because if we want to have the current install docs for main and release, we will have to change our install script, and possibly the helm install to support nightly

fmassot:
It's possible with docusaurus to make a section optional and I think we should be able to get the `version` and have an if statement.

PR #3524, opened 2023-06-07
Uploading nightly builds to a release

evanxg852000:
Uploading nightly builds to a specific nightly release so that we can refer to the nightly build binaries using as stable URL.
Related to https://github.com/quickwit-oss/quickwit/issues/1585

### DIFF
--- a/.github/actions/cargo-build-macos-binary/action.yml
+++ b/.github/actions/cargo-build-macos-binary/action.yml
@@ -67,4 +67,5 @@ runs:
       with:
         file: ${{ env.ASSET_FULL_NAME }}.tar.gz
         overwrite: true
+        draft: false
         tag_name: ${{ inputs.version }}
--- a/.github/actions/cargo-build-macos-binary/action.yml
+++ b/.github/actions/cargo-build-macos-binary/action.yml
@@ -61,7 +61,6 @@ runs:
         path: ./${{ env.ASSET_FULL_NAME }}.tar.gz
         retention-days: 3
     - name: Deploy archive to GitHub release
-      if: "${{ inputs.version != 'nightly' }}"
       uses: quickwit-inc/upload-to-github-release@v1
       env:
         GITHUB_TOKEN: ${{ inputs.token }}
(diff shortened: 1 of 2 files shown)


fmassot:
Interesting. I think artifacts of a draft release are not public.

One thing we can do is upload the artefacts to an s3 bucket and choose a domain to serve the files like `download.quickwit.com/nightly/`.

.github/actions/cargo-build-macos-binary/action.yml
```diff
@@ -61,11 +61,11 @@ runs:
...
         retention-days: 3
     - name: Deploy archive to GitHub release
-      if: "${{ inputs.version != 'nightly' }}"
       uses: quickwit-inc/upload-to-github-release@v1
       env:
         GITHUB_TOKEN: ${{ inputs.token }}
       with:
         file: ${{ env.ASSET_FULL_NAME }}.tar.gz
         overwrite: true
+        draft: false
```
fmassot, line 70: Ok so if you remove the draft, then it's public.

This is fine for a nightly build but for a real release, we should keep `draft: true` rigth?

.github/actions/cross-build-binary/action.yml
```diff
@@ -56,12 +56,23 @@ runs:
...
-    - name: Upload archive
+    - name: Deploy archive to GitHub nightly release
+      if: "${{ inputs.version == 'nightly' }}"
+      uses: quickwit-inc/upload-to-github-release@v1
+      env:
+        GITHUB_TOKEN: ${{ inputs.token }}
+      with:
+        file: ${{ env.ASSET_FULL_NAME }}.tar.gz
+        overwrite: true
+        draft: false
```
fmassot, line 67: does `draft: ${{ inputs.version == 'nightly' }}` can work?

.github/actions/cargo-build-macos-binary/action.yml
evanxg852000: I have fixed it. it's rather verbose which I was trying to avoid. Ideally, I need an action that just leaves the state of the release as is. Given the action is our own, can we update it?

.github/actions/cross-build-binary/action.yml
evanxg852000: I don't think so but let me try. I am not a fan of the verbose solution so I am still trying to find a better way.

evanxg852000: Thanks @fmassot 🎖️  the expression `draft: ${{ inputs.version != 'nightly' }}` works so no need for a huge duplicate.

fmassot (approved):

PR #3530, opened 2023-06-08
Enhancement of docker section

evanxg852000:
- Added a list of available environment variables in the docker installation section
- Updated `node-config.md` to add the environment variables to the respective config option

@guilload  The idea is to link this section from Quickwit Dockerhub page 
https://quickwit.io/docs/get-started/installation#use-the-docker-image

### DIFF
--- a/docs/configuration/node-config.md
+++ b/docs/configuration/node-config.md
@@ -39,11 +39,8 @@ There are also other parameters that can be only defined by env variables:
 
 | Env variable | Description |
 | --- | --- |
-| `QW_S3_ENDPOINT` | Custom S3 endpoint. |
-| `QW_S3_MAX_CONCURRENCY` | Limit the number of concurent requests to S3 |
+| `QW_S3_MAX_CONCURRENCY` | Limit the number of concurrent requests to S3 |
 | `QW_ENABLE_JAEGER_EXPORTER` | Enable trace export to Jaeger. |
-| `QW_AZURE_STORAGE_ACCOUNT` | Azure Blob Storage account name. |
-| `QW_AZURE_STORAGE_ACCESS_KEY` | Azure Blob Storage account access key. |
 
 More details about [storage configuration](../reference/storage-uri.md).
 
@@ -57,10 +54,10 @@ If a storage configuration is not explicitly set, Quickwit will rely on the defa
 
 ### Azure storage configuration
 
-| Property | Description | Default value |
-| --- | --- | --- |
-| `account` | The Azure storage account name. | |
-| `access_key` | The Azure storage account access key. | |
+| Property | Description | Env variable | Default value |
+| --- | --- | --- | --- |
+| `account` | The Azure storage account name. | `QW_AZURE_STORAGE_ACCOUNT` | |
+| `access_key` | The Azure storage account access key. | `QW_AZURE_STORAGE_ACCESS_KEY` | |
 
 Example of a storage configuration for Azure in YAML format:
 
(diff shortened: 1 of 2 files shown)


docs/configuration/node-config.md
```diff
@@ -39,11 +39,8 @@ There are also other parameters that can be only defined by env variables:
 
 | Env variable | Description |
 | --- | --- |
-| `QW_S3_ENDPOINT` | Custom S3 endpoint. |
-| `QW_S3_MAX_CONCURRENCY` | Limit the number of concurent requests to S3 |
+| `QW_S3_MAX_CONCURRENCY` | Limit the number of concurrent requests to S3 |
```
fmassot, line 42: why not putting `QW_S3_MAX_CONCURRENCY` in s3 section like others?

docs/get-started/installation.md
```diff
@@ -150,6 +150,46 @@ docker run --rm quickwit/quickwit --version
 # You can also safely ignore jemalloc warnings.
```
fmassot, line 150: Can you remove this line? we now support arm64 arch

```suggestion
```

docs/get-started/installation.md
```diff
@@ -150,6 +150,46 @@ docker run --rm quickwit/quickwit --version
...
 docker run --rm --platform linux/amd64 quickwit/quickwit --version
```
fmassot, line 151:
```suggestion
docker run --rm quickwit/quickwit --version
```

docs/get-started/installation.md
```diff
@@ -150,6 +150,46 @@ docker run --rm quickwit/quickwit --version
...
+    └── qwdata
+```
+
+- `config/quickwit.yaml`: the default configuration file.
+- `quickwit`: the quickwit executable binary.
+- `qwdata/`: the default data directory.
+
+Here is a list of the available environment variables and ports options that can be used to configure Quickwit:
+
+[Node configuration](../configuration/node-config.md):
```
fmassot, line 169: It feels weird to put this list without defining the variables. I would definitely avoid that and put just the link which is a very good idea.

docs/get-started/installation.md
```diff
@@ -150,6 +150,46 @@ docker run --rm quickwit/quickwit --version
...
+- `QW_DEFAULT_INDEX_ROOT_URI`
+- `QW_ENABLE_JAEGER_EXPORTER`
+- `QW_ENABLE_JAEGER_ENDPOINT`
+- `QW_S3_ENDPOINT`
+- `QW_S3_MAX_CONCURRENCY`
+- `QW_AZURE_STORAGE_ACCOUNT`
+- `QW_AZURE_STORAGE_ACCESS_KEY`
+
+[Ports configuration](../configuration/ports-config.md):
+- `QW_REST_LISTEN_PORT`
```
fmassot, line 187: Same for ports configuration, it's nice to have the link but I definitely want to avoid env variables in this page.

docs/get-started/installation.md
```diff
@@ -150,6 +150,46 @@ docker run --rm quickwit/quickwit --version
...
+- `QW_GOSSIP_LISTEN_PORT`
+- `QW_GRPC_LISTEN_PORT`
+
+[Telemetry](../telemetry.md):
```
fmassot, line 191: I think we don't need to add this here.

docs/configuration/node-config.md
evanxg852000: There was no corresponding config option, this is just an environment variable. but I have now moved it into that section as an env variable. I think we can create a config option for it if it makes sense like the others.

Issue #3535, opened 2023-06-09
Update otel index configs with relevant fast fields


fmassot:
traces:
- `resource_attributes` should be fast
- `span_attributes` should be fast

logs:
- `severity_text` should be fast
- `severity_number` should be fast
- `attributes` should be fast
- `resource_attributes` should be fast

PR #3536, opened 2023-06-09
Update otel indexes config

evanxg852000:
Updated otel index configs with relevant fast fields
Note: aggregation also works

Closes https://github.com/quickwit-oss/quickwit/issues/3535

### DIFF
--- a/quickwit/quickwit-opentelemetry/src/otlp/logs.rs
+++ b/quickwit/quickwit-opentelemetry/src/otlp/logs.rs
@@ -67,13 +67,16 @@ doc_mapping:
     - name: severity_text
       type: text
       tokenizer: raw
+      fast: true
     - name: severity_number
       type: u64
+      fast: true
     - name: body
       type: json
     - name: attributes
       type: json
       tokenizer: raw
+      fast: true
     - name: dropped_attributes_count
       type: u64
       indexed: false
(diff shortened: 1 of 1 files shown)


config/tutorials/otel-trace/index-config.yaml
```diff
@@ -5,51 +5,71 @@
 
 version: 0.6
```
fmassot, line 6: Is this index config used somewhere?

config/tutorials/otel-logs/index-config.yaml
```diff
@@ -5,30 +5,72 @@
 
 version: 0.6
 
-index_id: otel-log-v0
+index_id: otel-logs-v0_6
```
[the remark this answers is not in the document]
evanxg852000: I updated them here because these configs seem more like a reference to me. we should keep them up to date whether they are used in a tut or not. 
I can move the change to the other PR if it makes more sense.

config/tutorials/otel-trace/index-config.yaml
evanxg852000: Yes in tut https://github.com/quickwit-oss/quickwit/blob/main/docs/distributed-tracing/otel-service.md?plain=1#L30

fmassot (approved):

PR #3537, opened 2023-06-09
Timestamp field can now be defined at any field mapping level

### LINKED ISSUE #3531: Timestamp field is only valid if defined at the root document level
Regression coming from this commit: https://github.com/quickwit-oss/quickwit/commit/eeee44b3f075d29868932bfcad24b3aa32d8bef2#diff-2b06f568140d8993ed3037c07df5573caa9f697ce13d55222243f9903327ccd8

fmassot:
Fix #3531 

When we define the `timestamp_field` in the index config, we implicitly define it as a field path. Dots '.' is used as a field name separator. If a dot is part of a field name, it must be escaped.
This is the same logic for tag fields.

Example:
- `timestamp_field: object.timestamp` for a document like `{"object": {"timestmap": 10}}`
- `timestamp_field: my\\.timestamp` for a document like `{"my.timestamp": 10}`

Tests added.

### DIFF
--- a/quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
+++ b/quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
@@ -126,17 +126,13 @@ fn validate_timestamp_field(
     };
     if let FieldMappingType::DateTime(date_time_option, cardinality) = &timestamp_field_type {
         if cardinality != &Cardinality::SingleValue {
-            bail!(
-                "Timestamp field `{timestamp_field_path}` should be single-valued."
-            );
+            bail!("Timestamp field `{timestamp_field_path}` should be single-valued.");
         }
         if !date_time_option.fast {
             bail!("Timestamp field `{timestamp_field_path}` should be a fast field.");
         }
     } else {
-        bail!(
-            "Timestamp field `{timestamp_field_path}` should be a datetime field."
-        );
+        bail!("Timestamp field `{timestamp_field_path}` should be a datetime field.");
     }
     Ok(())
 }
@@ -601,6 +597,20 @@ mod tests {
         }"#,
         )
         .unwrap();
+
+        serde_yaml::from_str::<DefaultDocMapper>(
+            r#"
+            field_mappings:
+              - name: some_obj
+                type: object
+                field_mappings:
+                  - name: timestamp
+                    type: datetime
+                    fast: true
+            timestamp_field: some_obj.timestamp
+        "#,
+        )
+        .unwrap();
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -117,30 +117,25 @@ impl DefaultDocMapper {
...
-            };
-    if let FieldMappingType::DateTime(date_time_option, cardinality) =
-        &timestamp_field_entry.mapping_type
-    {
+    if let FieldMappingType::DateTime(date_time_option, cardinality) = &timestamp_field_type {
         if cardinality != &Cardinality::SingleValue {
             bail!(
-                "Multiple values are forbidden for  the timestamp field \
-                 (`{timestamp_field_name}`)."
+                "Multiple values are forbidden for the timestamp field (`{timestamp_field_path}`)."
```
guilload, line 130:
```suggestion
                "Timestamp field `{timestamp_field_path}` should be single-valued."
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -584,15 +581,54 @@ mod tests {
     }
 
     #[test]
-    fn test_fail_to_build_doc_mapper_with_non_datetime_timestamp_field() {
+    fn test_timestamp_field_in_object_is_valid() {
```
guilload, line 584: Let's test field names/paths without quotes. That's also valid, YAML.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -541,6 +566,33 @@ fn field_name_for_field_path(field_path: &[&str]) -> String {
...
+    let mut current_path_fragment = String::new();
+    let mut escaped = false;
+    for char in field_path_as_str.chars() {
+        if escaped {
+            current_path_fragment.push(char);
+            escaped = false;
+        } else if char == '\\' {
+            escaped = true;
+        } else if char == '.' {
+            field_path.push(current_path_fragment.clone());
```
guilload, line 584: `mem::replace`

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -1096,4 +1148,23 @@ mod tests {
             serde_json::json!("010203")
         );
     }
+
+    #[test]
+    fn test_field_path_for_field_name() {
```
guilload, line 1153: What about string starting or ending with a `.` vs `\.`?

guilload (approved):

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
fmassot: Yes... I thought about this. We don't do any validation for tags and timestamp fields, I was going to open an issue on this.

PR #3545, opened 2023-06-12
optimize sort by date and docid

### LINKED ISSUE #3298: Optimize sort by date
This optimization is more important than it might look like. In particular it considerably improves
the AllQuery we run when reaching the UI screen, and the same logic could apply to tracing. (although we will have to replicate this logic)

The optim can be done on the leaf side (although it can be done on the root side in the case of the MatchAllQuery). 

On the leaf side, if a leaf node has a lot of splits, it could for instance process the splits in the most recent order. 
If a split about to be processed has max(timestamp) < current bottom k, we know it won't contribute to the top K.
We can disable the top K computation, hence removing the need to download the timestamp column.

If there are no aggregation, in the future we can also just run a count query. 

- [x] Avoid downlading the sorting field if K's top K = 0 
- [ ] Nullify K depending on the split metadata and previously processed splits.
- [ ] Integrate a similar optim in tracing. 

To complicate stuff a little more... In the future we might want to remove "accurate count", and let quickwit tell us "More than 10000 docs matching" for instance. Entire split could then be skipped.

trinity-1686a:
### Description

fix #3298 
make it so sorting by date or document id takes advantage of metadata about splits to not search as much in splits which would give worst result than what we already have.
I tried benching it (setup below), and it made very little to no difference as is. However it made a huge difference if we allow non exact hit count.

### How was this PR tested?

added a few unit tests.

Created an index with a day worth of GH (2.7m docs), with 510 splits of ~5k docs each, and run a few queries, and with `searcher.max_num_concurrent_split_searches: 5`. This isn't representative of a real workload (it would have larger splits, and more concurrent search). 
It had no significant impact as is, however setting  `let run_all_splits = false;` in leaf.rs (i.e. ignore splits with worst results instead of just counting hits inside them) caused a 10x improvement.


quickwit/quickwit-common/src/binary_heap.rs
```diff
@@ -118,9 +118,70 @@ impl<O: Ord, T> PartialEq for OrderItemPair<O, T> {
...
+            let mut head = self.heap.peek_mut().unwrap();
+            let order = (self.sort_key_fn)(&item);
+            if head.0.order < order {
+                *head = Reverse(OrderItemPair { order, item });
+            }
+        }
+    }
+
+    /// Get a reference to the worst entry.
+    pub fn peek_worst(&self) -> Option<&T> {
```
fulmicoton, line 169: I like the name.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -100,7 +102,7 @@ impl SortingFieldComputer {
                     SortFieldType::U64 => SortValue::U64(sort_value),
                     SortFieldType::I64 => SortValue::I64(i64::from_u64(sort_value)),
                     SortFieldType::F64 => SortValue::F64(f64::from_u64(sort_value)),
-                    SortFieldType::DateTime => SortValue::U64(sort_value),
```
fulmicoton, line 103: O_O

quickwit/quickwit-search/src/collector.rs
```diff
@@ -695,13 +699,149 @@ pub(crate) fn make_merge_collector(
     })
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct PartialHitSortingKey {
+    score: Option<SortValue>,
+    address: GlobalDocAddress,
+    asc: bool,
```
fulmicoton, line 706: I suspect we have a SortOrder { Ascending, Descending } in this crate

quickwit/quickwit-search/src/collector.rs
```diff
@@ -695,13 +699,149 @@ pub(crate) fn make_merge_collector(
...
+    fn cmp(&self, other: &PartialHitSortingKey) -> Ordering {
+        assert_eq!(
+            self.asc, other.asc,
+            "comparing two PartialHitSortingKey of different ordering"
+        );
+        if self.asc {
+            self.score
+                .map(Reverse)
+                .cmp(&other.score.map(Reverse))
+                .then(other.address.cmp(&self.address))
```
fulmicoton, line 719:
```suggestion
            self.score
                .map(Reverse)
                .cmp(&other.score.map(Reverse))
                .then(other.address.cmp(&self.address))
```

alternatively 

`(other.score, other.address).cmp(&(self.score, self.address))`

quickwit/quickwit-search/src/collector.rs
```diff
@@ -695,13 +699,149 @@ pub(crate) fn make_merge_collector(
...
+        } else {
+            self.score
```
fulmicoton, line 721: Alternatively 

`(self.score, self.address).cmp(&(other.score, other.address))`

quickwit/quickwit-search/src/collector.rs
fulmicoton: (nitpick)

quickwit/quickwit-search/src/collector.rs
fulmicoton: (nitpick)

fulmicoton: You can also do something like:

let descending_order =  (self.score, self.address).cmp(&(other.score, other.address));
if self.asc {
   descending_order.reverse()
} else {
   descending_order
}
```

File: quickwit/quickwit-search/src/collector.rs [hunk 6]
```diff
@@ -695,13 +699,149 @@ pub(crate) fn make_merge_collector(
...
+    failed_splits: Vec<SplitSearchError>,
+    num_attempted_splits: u64,
+}
+
+impl IncrementalCollector {
+    /// Create a new incremental collector
+    pub(crate) fn new(inner: QuickwitCollector) -> Self {
+        let sort_key_fn = match inner.sort_by.sort_order() {
+            SortOrder::Asc => sort_asc,
+            SortOrder::Desc => sort_desc,
```
fulmicoton, line 764: Yeah at this point I think you would have been better off using SortOrder in the PartialHitSortingKey :D

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -414,6 +415,111 @@ pub(crate) fn rewrite_start_end_time_bounds(
...
+            MutCanSplitDoBetter::SplitIdHigher(split_id) => *split_id = Some(hit.split_id.clone()),
+            MutCanSplitDoBetter::SplitTimestampHigher(timestamp) => {
+                if let Some(SortValue::I64(timestamp_ns)) = hit.sort_value {
+                    // if we get a timestamp of, says 1.5s, we need to check up to 2s to make
+                    // sure we don't throw away something like 1.2s, so we should round up while
+                    // dividing.
+                    let mut timestamp_s = timestamp_ns / 1_000_000_000;
+                    if timestamp_ns % 1_000_000_000 != 0 {
+                        timestamp_s += 1;
+                    }
```
fulmicoton, line 490: I think the usual way to compute ceil_div is the following
```
#[inline]
fn ceil_div(n: i64, q: i64) {    
    assert!(q != 0i64);
    assert!(q != i64::MAX);
    (n + (q  - 1)) / q
}
```

```suggestion
      ceil_div(timestamp_ns, 1_000_000)
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -424,31 +530,78 @@ pub async fn leaf_search(
...
     request: &SearchRequest,
     index_storage: Arc<dyn Storage>,
-    splits: &[SplitIdAndFooterOffsets],
+    splits: &mut [SplitIdAndFooterOffsets],
     doc_mapper: Arc<dyn DocMapper>,
 ) -> Result<LeafSearchResponse, SearchError> {
     let request = Arc::new(request.clone());
-    let leaf_search_single_split_futures: Vec<_> = splits
+
+    let split_filter = if request.sort_by_field.is_none() {
```
fulmicoton, line 538:
```suggestion
    // The leaf search code contains some logic that makes it possible to skip entire splits
    // when we are confident they won't make it into top K.
    // To make this optimization as potent as possible, we sort the splits so that the first splits
    // are the most likely to fill our Top K.
    let split_filter = if request.sort_by_field.is_none() {
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -424,31 +530,78 @@ pub async fn leaf_search(
...
+        .sort_by_field
+        .as_deref()
+        .zip(doc_mapper.timestamp_field_name())
+    {
+        if sort_by_field == timestamp_field {
+            if request.sort_order() == SortOrder::Desc {
+                splits.sort_unstable_by_key(|b| std::cmp::Reverse(b.timestamp_end()));
+                CanSplitDoBetter::desc_timestamp()
+            } else {
+                splits.sort_unstable_by_key(|a| a.timestamp_start());
```
fulmicoton, line 551:
```suggestion
                splits.sort_unstable_by_key(|split| split.timestamp_start());
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -424,31 +530,78 @@ pub async fn leaf_search(
...
+                CanSplitDoBetter::asc_timestamp()
+            }
+        } else {
+            CanSplitDoBetter::uninformative()
+        }
+    } else {
+        CanSplitDoBetter::uninformative()
+    };
+
+    // In the future this should become `request.aggregation_request.is_some() ||
```
fulmicoton, line 561: great comment.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -424,31 +530,78 @@ pub async fn leaf_search(
...
+                // TODO currently permits are not always given in order when thread pool has more
+                // than one thread. This means sorting splits is only half useful.
                 let _leaf_split_search_permit = searcher_context_clone.leaf_search_split_semaphore
                     .acquire()
                     .await
                     .expect("Failed to acquire permit. This should never happen! Please, report on https://github.com/quickwit-oss/quickwit/issues.");
+
+                if !split_filter.can_be_better(&split) {
+                    if run_all_splits {
+                        request.max_hits = 0;
```
fulmicoton, line 585: ah yes interesting actually.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -414,6 +415,111 @@ pub(crate) fn rewrite_start_end_time_bounds(
...
+
+    /// Returns whether it is useful to iteratively merge instead of merging all at once.
+    pub fn iterative_merge(&self) -> bool {
+        !matches!(
+            *self.inner.read().unwrap(),
+            MutCanSplitDoBetter::Uninformative
+        )
+    }
+
+    /// Whether we should wash score and rely only on document GlobalAddress.
```
fulmicoton, line 514: I don't understand what `washing` means here.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -459,46 +612,86 @@ pub async fn leaf_search(
...
-    // It should be executed by Tokio's blocking threads.
-    let span = info_span!("merge_search_responses");
-    let mut merged_search_response = crate::run_cpu_intensive(move || {
-        let _span_guard = span.enter();
-        merge_collector.merge_fruits(split_search_responses)
-    })
-    .await
-    .context("Failed to merge split search responses.")??;
+    let wash_score = split_filter.wash_score();
+    if split_filter.iterative_merge() {
```
fulmicoton, line 621: I don't understand why we need a different logic for the iterative case and for the non iterative case.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -459,46 +612,86 @@ pub async fn leaf_search(
...
-            error: format!("{err}"),
-            retryable_error: true,
-        }));
-    Ok(merged_search_response)
+            if let Some(last_hit) = incremental_merge_collector.peek_worst_hit() {
+                split_filter.record_new_worst_hit(last_hit);
+            }
+        }
+
+        incremental_merge_collector.finalize().map_err(Into::into)
```
fulmicoton, line 655: can we run that in the executor? Merging aggregation can be quite CPU intensive.

quickwit/quickwit-search/src/leaf.rs
trinity-1686a: sadly, I don't measure any meaningful difference from that. I don't know if we have any optimization going on when `max_hits == 0` (don't download the date fastfield for instance, or somehow use less cpu)

quickwit/quickwit-search/src/leaf.rs
trinity-1686a: I expect iterative to be somewhat slower than non iterative merge, though this is probably a non-problem. Most of the filtering already happened before and this part is less performance critical. I'll bench things to see if having both code paths is really worth the trouble

quickwit/quickwit-search/src/collector.rs
trinity-1686a: part of why this is needed is because I did not want TopK to have a type parameter (which is required when using closures, unless you box it), so I used function pointers directly.
Having a type parameter on TopK means IncrementalCollector would have one too, and I don't think it would be possible to name the type output of `IncrementalCollector::new` due to that anonymous closure.
Alternatively, we could have a simple `SortKeyMapper` trait or something along the line, with a single function doing the same as `sort_key_fn`, and an implementer which can be named

wdyt?

fulmicoton: Yes yes. I understood you wanted to avoid having a generic type. This is looking good.

My comment was more in reference to another place were you used a bool to say Ascending / Descending. Had 2 constructors for each, and an if statement. Only to eventually convert it to the `SortOrder` enum.

fulmicoton: (I think you already modified the code)

quickwit/quickwit-common/src/lib.rs
```diff
@@ -174,6 +174,16 @@ where T: Debug
     }
 }
 
+#[inline]
+pub fn ceil_div(n: i64, q: i64) -> i64 {
+    assert!(q != 0i64);
+    // TODO trinity-1686a: i'm unsure about this assertion, i64::MIN makes more sense to me,
```
fulmicoton, line 180: good point.

the `div_ceil` in nightly goes

``
   pub const fn div_ceil(self, rhs: Self) -> Self {
            let d = self / rhs;
            let r = self % rhs;
            if (r > 0 && rhs > 0) || (r < 0 && rhs < 0) {
                d + 1
            } else {
                d
            }
        }
```

which is closer to you original value.

Comment (fulmicoton) [reply quickwit/quickwit-common/src/lib.rs:180]: So.... my bad. Also I am not sure what I suggested is exactly what you want for signed stuff. So forget my original comment and:
- do the implementation you see fit. performance is not an issue here.
- comment the behavior on negative int.
- maybe rename it ceil_div to align with rust

File: quickwit/quickwit-search/src/leaf.rs [hunk 16]
```diff
@@ -459,46 +618,88 @@ pub async fn leaf_search(
...
-            error: format!("{err}"),
-            retryable_error: true,
-        }));
-    Ok(merged_search_response)
+            if let Some(last_hit) = incremental_merge_collector.peek_worst_hit() {
+                split_filter.record_new_worst_hit(last_hit);
+            }
+        }
+
+        crate::run_cpu_intensive(|| incremental_merge_collector.finalize().map_err(Into::into))
```
fulmicoton, line 661: thx

quickwit/quickwit-search/src/leaf.rs
fulmicoton: yes if you can check it is worth. I'd prefer have a single path here if possible.

fulmicoton (approved):

trinity-1686a:
I believe this conflicts with #3614, I'll have to see how both should go together

quickwit/quickwit-search/src/tests.rs
```diff
@@ -666,7 +666,7 @@ async fn test_sort_bm25() {
         let hits: Vec<(f32, u32)> = search_hits("nofreq:two").await;
         assert_eq!(
             &hits[..],
-            &[(0.15965714, 1), (0.12343242, 0), (0.12343242, 2)]
+            &[(0.15965714, 1), (0.12343242, 2), (0.12343242, 0)]
```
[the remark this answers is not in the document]
fulmicoton: It really does not matter much. I agree your order is nicer though: if we sort by dates in a descending fashion, we want higher doc_id first.

(timestamp could be tie, but documents tend to be order with time so we want to use them that way to break the tie)
It seems to be the way your code works, so that's ok with me.

Is this order respected both in the merge and the leaf code?

trinity-1686a:
While trying to assess if sorting was coherent between leaf and merge, I found it currently isn't due to bugs in leaf sorting in the presence of `None`. 

My test was composed of ingesting the following synthetic dataset, both everything-in-one-split, and a split per document. Then the result between both were compared with the results I expected, as defined in the python scripts that follows 

<details><summary>dataset</summary>

```json
{"sort1": 4, "sort2": 3, "global_id": 0}
{"sort1": 6, "sort2": 7, "global_id": 1}
{"sort1": 9, "sort2": 7, "global_id": 2}
{"sort1": 7, "sort2": 4, "global_id": 3}
{"sort1": 2, "sort2": null, "global_id": 4}
{"sort1": null, "sort2": 1, "global_id": 5}
{"sort1": null, "sort2": 3, "global_id": 6}
{"sort1": 8, "sort2": 10, "global_id": 7}
{"sort1": 4, "sort2": 5, "global_id": 8}
{"sort1": 4, "sort2": null, "global_id": 9}
{"sort1": 1, "sort2": 3, "global_id": 10}
{"sort1": null, "sort2": null, "global_id": 11}
{"sort1": 4, "sort2": null, "global_id": 12}
{"sort1": 7, "sort2": 10, "global_id": 13}
{"sort1": 1, "sort2": 4, "global_id": 14}
{"sort1": 5, "sort2": null, "global_id": 15}
{"sort1": 9, "sort2": 10, "global_id": 16}
{"sort1": 9, "sort2": 5, "global_id": 17}
{"sort1": 10, "sort2": 5, "global_id": 18}
{"sort1": 7, "sort2": 1, "global_id": 19}
{"sort1": 7, "sort2": 7, "global_id": 20}
{"sort1": 1, "sort2": 3, "global_id": 21}
{"sort1": 8, "sort2": 7, "global_id": 22}
{"sort1": null, "sort2": null, "global_id": 23}
{"sort1": null, "sort2": 3, "global_id": 24}
{"sort1": 2, "sort2": 4, "global_id": 25}
{"sort1": 3, "sort2": 6, "global_id": 26}
{"sort1": 3, "sort2": 10, "global_id": 27}
{"sort1": 2, "sort2": 3, "global_id": 28}
{"sort1": 8, "sort2": 2, "global_id": 29}
{"sort1": null, "sort2": 6, "global_id": 30}
{"sort1": 3, "sort2": 8, "global_id": 31}
{"sort1": 1, "sort2": 2, "global_id": 32}
{"sort1": 3, "sort2": 8, "global_id": 33}
{"sort1": 7, "sort2": null, "global_id": 34}
{"sort1": 3, "sort2": 4, "global_id": 35}
{"sort1": 2, "sort2": 7, "global_id": 36}
{"sort1": 2, "sort2": 3, "global_id": 37}
{"sort1": 9, "sort2": 5, "global_id": 38}
{"sort1": null, "sort2": 8, "global_id": 39}
{"sort1": 7, "sort2": 3, "global_id": 40}
{"sort1": 6, "sort2": 6, "global_id": 41}
{"sort1": null, "sort2": 4, "global_id": 42}
{"sort1": 6, "sort2": 9, "global_id": 43}
{"sort1": 7, "sort2": 3, "global_id": 44}
{"sort1": 3, "sort2": 6, "global_id": 45}
{"sort1": null, "sort2": 3, "global_id": 46}
{"sort1": 1, "sort2": 9, "global_id": 47}
{"sort1": null, "sort2": 3, "global_id": 48}
{"sort1": null, "sort2": 2, "global_id": 49}
{"sort1": 4, "sort2": 6, "global_id": 50}
{"sort1": null, "sort2": null, "global_id": 51}
{"sort1": 5, "sort2": 4, "global_id": 52}
{"sort1": 6, "sort2": 8, "global_id": 53}
{"sort1": 9, "sort2": 1, "global_id": 54}
{"sort1": null, "sort2": 9, "global_id": 55}
{"sort1": null, "sort2": 2, "global_id": 56}
{"sort1": null, "sort2": 8, "global_id": 57}
{"sort1": 8, "sort2": null, "global_id": 58}
{"sort1": 8, "sort2": 1, "global_id": 59}
{"sort1": 8, "sort2": 1, "global_id": 60}
{"sort1": 6, "sort2": 4, "global_id": 61}
{"sort1": 6, "sort2": null, "global_id": 62}
{"sort1": 10, "sort2": 7, "global_id": 63}
{"sort1": 2, "sort2": 7, "global_id": 64}
{"sort1": null, "sort2": 3, "global_id": 65}
{"sort1": 7, "sort2": 2, "global_id": 66}
{"sort1": 1, "sort2": null, "global_id": 67}
{"sort1": 8, "sort2": 5, "global_id": 68}
{"sort1": 1, "sort2": 10, "global_id": 69}
{"sort1": 8, "sort2": 2, "global_id": 70}
{"sort1": 2, "sort2": 9, "global_id": 71}
{"sort1": null, "sort2": 2, "global_id": 72}
{"sort1": 7, "sort2": 3, "global_id": 73}
{"sort1": 8, "sort2": 1, "global_id": 74}
{"sort1": 7, "sort2": 2, "global_id": 75}
{"sort1": null, "sort2": 10, "global_id": 76}
{"sort1": 3, "sort2": 10, "global_id": 77}
{"sort1": null, "sort2": 8, "global_id": 78}
{"sort1": 8, "sort2": 6, "global_id": 79}
{"sort1": 2, "sort2": 8, "global_id": 80}
{"sort1": null, "sort2": 10, "global_id": 81}
{"sort1": 8, "sort2": null, "global_id": 82}
{"sort1": 10, "sort2": 7, "global_id": 83}
{"sort1": 7, "sort2": null, "global_id": 84}
{"sort1": null, "sort2": null, "global_id": 85}
{"sort1": 2, "sort2": 4, "global_id": 86}
{"sort1": 7, "sort2": 2, "global_id": 87}
{"sort1": 2, "sort2": 5, "global_id": 88}
{"sort1": null, "sort2": 6, "global_id": 89}
{"sort1": null, "sort2": 2, "global_id": 90}
{"sort1": 9, "sort2": 9, "global_id": 91}
{"sort1": 5, "sort2": 9, "global_id": 92}
{"sort1": 6, "sort2": 3, "global_id": 93}
{"sort1": null, "sort2": 1, "global_id": 94}
{"sort1": 2, "sort2": 10, "global_id": 95}
{"sort1": 4, "sort2": 1, "global_id": 96}
{"sort1": null, "sort2": null, "global_id": 97}
{"sort1": 3, "sort2": 2, "global_id": 98}
{"sort1": 10, "sort2": 10, "global_id": 99}
```

</details>

<details><summary>script</summary>

```python
#!/usr/bin/env python

import requests,json

base_url = "http://127.0.0.1:7280/api/v1/sorted-multi/search?query=*&max_hits={count}&sort_by={order}"

with open('/home/trinity/dev/datasets/dataset/sorted/dataset.json', 'r') as json_file:
    json_list = list(json_file)

raw_results = []
for json_str in json_list:
    result = json.loads(json_str)
    raw_results.append(result)

def cmp(field, x):
    reverse = False
    if field.startswith("-"):
        field = field[1:]
        reverse = True

    x = x[field]
    if x is None:
        return 0
    if reverse:
        return x
    else:
        return 100 - x

sorted_results = {}

def chain(fields):
    def chain_inner(fields, x):
        reverse_first = None
        score = 0
        for field in fields:
            if reverse_first is None:
                reverse_first = field.startswith("-")
            score += cmp(field, x)
            score *= 1000
            
        if reverse_first != False:
            return score + cmp("-global_id", x)
        else:
            return score + cmp("global_id", x)
    return lambda x: chain_inner(fields, x)


for name in ["", "sort1", "-sort1", "sort1,sort2", "-sort1,sort2", "sort1,-sort2", "-sort1,-sort2"]:
    
    args = name.split(",")
    if args[0] == "":
        args = []
    fn = chain(args)

    sorted_results[name] = sorted(raw_results, key=fn, reverse=True)

def run_case(count, order):
    url = base_url.format(count=count, order=order)

    res = requests.get(url)
    res = json.loads(res.content)
    if len(res["hits"]) != count:
        print("count error: ", count, order)
        exit(1)

    expected = sorted_results[order][:count]
    processed_res = [
        {"sort1": hit.get("sort1"),
         "sort2": hit.get("sort2"),
         "global_id": hit["global_id"]}
        for hit in res["hits"]]
    if processed_res != expected:
        print("content error: ", count, order)
        print("got     ", res["hits"])
        print("expected", expected)
        exit(1)

for kind in ["", "sort1", "-sort1", "sort1,sort2", "-sort1,sort2", "sort1,-sort2", "-sort1,-sort2"]:
    print("test case:", kind)
    for i in range(101):
        run_case(i, kind)
```

</details>

This allows testing thoroughly both single-leaf sort and incremental (single node) sort, however "normal" merge which happens when merging splits from different nodes isn't tested just yet. I will try to somehow make this python test into a test we can integrate in CI.

I found current leaf sort code was hard to wrap my head around, so I reused the incremental merge infrastructure as it is, imo, a lot easier to reason about.

trinity-1686a:
this is currently blocked by having an agreed upon spec for sorting (#3726), and fixing some of the bugs currently in main

trinity-1686a:
superseded by #3849

PR #3566, opened 2023-06-20
Add a parameter to limit the size of a query string

### LINKED ISSUE #3546: Add a parameter to limit the size of a query string
We want to avoid a user making a query with a huge text: this can indeed generate a lot of requests to the object storage.

We could set it as a parameter at the index level. When updating the index config, the root searcher will get the info from the config and will be able to validate the query.

shimatar0:
### Description

fix #3546 
Added a parameter to set the maximum number of characters in a query.
Default is set to 512 for the following reasons.
Because the query for this test was 395 bytes, and I thought the default of 256 byte might be too short even for a normal use case. (I'd like to hear from maintainers here)
https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-jaeger/src/integration_tests.rs#L197

### How was this PR tested?

add a few unit tests.
Please let us know if you feel your test case is missing :)

### DIFF
--- a/quickwit/quickwit-search/src/lib.rs
+++ b/quickwit/quickwit-search/src/lib.rs
@@ -199,14 +199,13 @@ pub async fn single_node_search(
     let metas = list_relevant_splits(index_uid, &search_request, metastore).await?;
     let split_metadata: Vec<SplitIdAndFooterOffsets> =
         metas.iter().map(extract_split_and_footer_offsets).collect();
-    validate_request(&*doc_mapper, &search_request)?;
+    let searcher_context = Arc::new(SearcherContext::new(SearcherConfig::default()));
+    validate_request(&*doc_mapper, &search_request, &searcher_context)?;
 
     // Verifying that the query is valid.
     doc_mapper
         .query(doc_mapper.schema(), &query_ast_resolved, true)
         .map_err(|err| SearchError::InvalidQuery(err.to_string()))?;
-
-    let searcher_context = Arc::new(SearcherContext::new(SearcherConfig::default()));
 
     let leaf_search_response = leaf_search(
         searcher_context.clone(),
--- a/quickwit/quickwit-search/src/root.rs
+++ b/quickwit/quickwit-search/src/root.rs
@@ -173,6 +173,7 @@ fn validate_sort_by_field(field_name: &str, schema: &Schema) -> crate::Result<()
 pub(crate) fn validate_request(
     doc_mapper: &dyn DocMapper,
     search_request: &SearchRequest,
+    searcher_context: &SearcherContext,
 ) -> crate::Result<()> {
     let schema = doc_mapper.schema();
 
@@ -204,6 +205,14 @@ pub(crate) fn validate_request(
         )));
     }
 
+    if search_request.query_ast.len() > searcher_context.query_string_max_length {
+        return Err(SearchError::InvalidArgument(format!(
+            "max length for query string is {}, but got {}",
+            searcher_context.query_string_max_length,
+            search_request.query_ast.len()
+        )));
+    }
+
     Ok(())
 }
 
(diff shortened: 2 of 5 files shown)


quickwit/quickwit-search/src/lib.rs
```diff
@@ -199,15 +199,14 @@ pub async fn single_node_search(
     let metas = list_relevant_splits(index_uid, &search_request, metastore).await?;
     let split_metadata: Vec<SplitIdAndFooterOffsets> =
         metas.iter().map(extract_split_and_footer_offsets).collect();
-    validate_request(&*doc_mapper, &search_request)?;
+    let searcher_context = Arc::new(SearcherContext::new(SearcherConfig::default()));
```
fulmicoton, line 202:
```suggestion
    let searcher_context = SearcherContext::new(SearcherConfig::default());
```

quickwit/quickwit-search/src/service.rs
```diff
@@ -303,6 +303,8 @@ pub struct SearcherContext {
     pub split_stream_semaphore: Semaphore,
     /// Recent sub-query cache.
     pub leaf_search_cache: LeafSearchCache,
+    /// Query string max length.
+    pub query_string_max_length: usize,
```
fulmicoton, line 307:
```suggestion
```

fulmicoton: it is in the searcher_config, which is already in the SearcherContext.

quickwit/quickwit-search/src/root.rs
```diff
@@ -173,6 +173,7 @@ fn validate_sort_by_field(field_name: &str, schema: &Schema) -> crate::Result<()
 pub(crate) fn validate_request(
     doc_mapper: &dyn DocMapper,
     search_request: &SearchRequest,
+    searcher_context: &SearcherContext,
```
fulmicoton, line 176: could we pass a `&SearcherConfig` instead?

quickwit/quickwit-search/src/root.rs
```diff
@@ -204,6 +205,14 @@ pub(crate) fn validate_request(
         )));
     }
 
+    if search_request.query_ast.len() > searcher_context.query_string_max_length {
```
fulmicoton, line 208: I'm not fond of enforcing that on the query_ast. On the other hand, it might be difficult to do something at the user query level. Maybe we should have a limit not in number of chars but in number of terms?

shimatar0: Ok.
The query equivalent that was easiest to handle was `search_request.query_ast`

I have a few questions about this response.
- Is it necessary to change `query_ast` from a json format to a structure again for simplicity?
- Which is the recognition that indicates which is the `term` here?
  (Is it in the `QueryAst` field??)

guilload:
> Default is set to 512 for the following reasons.
I'd personally go even higher with 1,000 bytes or more. Queries generated programmatically can be long.

quickwit/quickwit-search/src/service.rs
shimatar0: fix
https://github.com/quickwit-oss/quickwit/pull/3566/commits/9d02896663ee4181c52422baf6c61cf55b4b29f5

quickwit/quickwit-search/src/root.rs
shimatar0: fix
https://github.com/quickwit-oss/quickwit/pull/3566/commits/9d02896663ee4181c52422baf6c61cf55b4b29f5

quickwit/quickwit-search/src/lib.rs
shimatar0: fix(This `searcher_context` here has been reverted to its previous state)
https://github.com/quickwit-oss/quickwit/pull/3566/commits/9d02896663ee4181c52422baf6c61cf55b4b29f5

PR #3580, opened 2023-06-26
config: Added more S3StorageConfig options

jpds:
### Description

Adds the other options of `S3StorageConfig` to the example configuration and sets examples to defaults per https://quickwit.io/docs/configuration/node-config/#s3-storage-configuration

### DIFF
--- a/config/quickwit.yaml
+++ b/config/quickwit.yaml
@@ -73,6 +73,9 @@ version: 0.6
 # default_index_root_uri: s3://your-bucket/indexes
 #
 # -------------------------------- Storage settings --------------------------------
+# Hardcoding credentials into configuration files is not secure and strongly
+# discouraged. Prefer the alternative authentication methods that your storage
+# backend may provide.
 #
 # storage:
 #   azure:
@@ -80,7 +83,10 @@ version: 0.6
 #     access_key: ${QW_AZURE_STORAGE_ACCESS_KEY}
 #
 #   s3:
+#     access_key_id: ${AWS_ACCESS_KEY_ID}
+#     secret_access_key: ${AWS_SECRET_ACCESS_KEY}
 #     endpoint: ${QW_S3_ENDPOINT}
+#     region: ${AWS_REGION}
 #     force_path_style_access: ${QW_S3_FORCE_PATH_STYLE_ACCESS:-true}
 #     disable_multi_object_delete_requests: true
 #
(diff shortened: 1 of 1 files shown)


config/quickwit.yaml
```diff
@@ -80,9 +80,12 @@ version: 0.6
 #     access_key: ${QW_AZURE_STORAGE_ACCESS_KEY}
 #
 #   s3:
+#     access_key_id: ${AWS_ACCESS_KEY_ID}
```
guilload, line 83: We don't want to incentivize users to use the config file to set up their credentials. For instance, this is what you'll find in the documentation of the AWS SDK:

>  It is __NOT__ secure to hardcode credentials into your application, and the credentials
  providers that come with the AWS SDK should be preferred.
  
  I'm saddened we must do so for Azure, but the Azure SDK leaves us no choice. Can you remove `access_key_id` and `secret_access_key`?

config/quickwit.yaml
```diff
@@ -80,9 +80,12 @@ version: 0.6
...
+#     secret_access_key: ${AWS_SECRET_ACCESS_KEY}
 #     endpoint: ${QW_S3_ENDPOINT}
-#     force_path_style_access: ${QW_S3_FORCE_PATH_STYLE_ACCESS:-true}
-#     disable_multi_object_delete_requests: true
+#     region: ${AWS_REGION}
+#     force_path_style_access: ${QW_S3_FORCE_PATH_STYLE_ACCESS:-false}
+#     disable_multi_object_delete_requests: false
```
guilload, line 88: I'd rather leave the value to `true`. If you are uncommenting this line, you most likely want to change the default behavior.

jpds: From a new user's perspective if I see an option in the example - I assume that's the recommended and default value - however in this case I then read the documentation and it turns out that the docs say the default is something different.

config/quickwit.yaml
jpds: > We don't want to incentivize users to use the config file to set up their credentials.

That's interesting - from experience this certainly isn't what other applications choose to be doing:

- https://garagehq.deuxfleurs.fr/documentation/connect/apps/
- https://thanos.io/tip/thanos/storage.md/#s3

Hence why this looked a little broken to me at first (and I'm also trying to avoid the craziness of what Thanos has with its configuration of random percentages of options being split between environment variables and YAML configuraiton (on top of which different percentages apply to different components)).

guilload: I'm ok with leaving those if we add a disclaimer:
```
Hardcoding credentials into configuration files is not secure and strongly discouraged. Prefer the alternative authentication methods that your storage backend may provide.
```

I'll add the same disclaimer to our documentation.

guilload (approved):

guilload:
Thanks, @jpds!

Issue #3578, opened 2023-06-26
DNS failure using Garage endpoint


jpds:
**Describe the bug**

I just tried using Quickwit locally on my laptop pointing at a remote Garage installation, and with the following configuration:

```
metastore_uri: s3://qw-test/indexes
default_index_root_uri: s3://qw-test/indexes
storage:
   s3:
     access_key_id: "..."
     secret_access_key: "..."
     endpoint: https://s3.example.com
     region: "garage"
     force_path_style_access: true
     disable_multi_object_delete_requests: false
```

With this, it would simply return:

```
[...]

PR #3583, opened 2023-06-26
Fix metastore resolver

guilload:
### Description
Ensure the metastore resolver uses the configured storage resolver.

### How was this PR tested?
- [x] add unit test


trinity-1686a (approved):
LGTM

fix #3578

PR #3575, opened 2023-06-25
Add support for custom tokenizers ngram and regex.

fmassot:
Fix #3056 and #3392

This PR adds support for custom analyzers, aka tokenizer + filters.

## Doc mapping config with custom tokenizers

```yaml
doc_mapping:
  tokenizers:
    - name: service_regex
      type: regex
      pattern: "\\w*"
  field_mappings:
    - name: service
      type: text
      tokenizer: service_regex
```

## Endpoint to test a custom tokenizer

`POST /analyze`

with payload

```json
{
  "type": "ngram",
  "min_gram": 3,
  "max_gram": 5,
  "text": "hello"
}
```

```console
curl -XPOST http://localhost:7280/api/v1/analyze -H "content-type: application/json" --data '{"type": "regex", "pattern": "\\w+", "text":"hello world"}'
[
  {
    "offset_from": 0,
    "offset_to": 5,
    "position": 0,
    "position_length": 1,
    "text": "hello"
  },
  {
    "offset_from": 6,
    "offset_to": 11,
    "position": 1,
    "position_length": 1,
    "text": "world"
  }
]
```

## Follow-up in PRs to come

- [ ] Add documentation.
- [ ] Add multilang tokenizer
- [ ] What else?

PR #3584, opened 2023-06-26
Add multilang tokenizer

fmassot:
Wait for #3575

### DIFF
--- a/quickwit/Cargo.toml
+++ b/quickwit/Cargo.toml
@@ -222,7 +222,7 @@ quickwit-serve = { version = "0.6.1", path = "./quickwit-serve" }
 quickwit-storage = { version = "0.6.1", path = "./quickwit-storage" }
 quickwit-telemetry = { version = "0.6.1", path = "./quickwit-telemetry" }
 
-tantivy = { git = "https://github.com/quickwit-oss/tantivy/", rev = "924fc70", default-features = false, features = [
+tantivy = { git = "https://github.com/quickwit-oss/tantivy/", branch = "fmassot/add-box-token-filter-and-refactor", default-features = false, features = [
   "mmap",
   "lz4-compression",
   "zstd-compression",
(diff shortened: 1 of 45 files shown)


quickwit/Cargo.toml
```diff
@@ -88,6 +88,8 @@ indicatif = "0.17.3"
 itertools = "0.10.5"
 json_comments = "0.2"
 libz-sys = "1.1.8"
+lindera-core = "0.25.0"
```
fulmicoton, line 91: can you put all that behind a feature flag?

(this is a very independent feature and I don't want it to impact our day to day dev compilation time.)

quickwit/quickwit-cli/Cargo.toml
```diff
@@ -65,6 +65,7 @@ quickwit-doc-mapper = { workspace = true }
 quickwit-indexing = { workspace = true }
 quickwit-metastore = { workspace = true }
 quickwit-proto = { workspace = true }
+quickwit-query = { workspace = true }
```
fulmicoton, line 68: this is not looking good...

I understand you want to add this feature to forward the feature to quickwit-query

fulmicoton: Investigated a little bit more..

So originally we have 
a) `quickwit-query` <- `quickwit-doc-mapper`
b) `quickwit-query` <- `quickiwtindexing`

That are quite discutable.
b) exists to get the tokenizer managers for indexing, but also unfortunately, due the delete pipeline which gets a query. The latter will make it difficult to inverse.

a) is less justified. I have previously opened a ticket to refactor this dependency away.
I dont think your PR reinforce that link so I guess it should be fine.

fulmicoton: Marking as resolved to avoid confusing.

PR #3585, opened 2023-06-27
Add maturity date in split metadata and use it to filter mature splits for merges

### LINKED ISSUE #3576: Merge pipeline failed when index has too many splits
**Describe the bug**
quickwit-indexer got this error
```
2023-06-26T07:47:49.057Z  WARN spawn_merge_pipeline{index=xxx-index gen=0}: quickwit_metastore::metastore::retrying_metastore::retry: Request failed attempt_count=5
2023-06-26T07:47:49.057Z ERROR quickwit_indexing::actors::merge_pipeline: Error while spawning indexing pipeline, retrying after some time. error=Internal error: `Error, message length too large: found 50114113 bytes, the limit is: 4194304 bytes` Cause: ``. retry_count=0 retry_delay=2s
```

**Expected behavior**
no error

**Configuration:**
quickwit version: v0.6.1
deploy with helm, default settings (with cpu/memory request/limit increased)

----

I found the same error(message length too large) when request `/api/v1/indexes/<index-id>/splits` on searcher, but successed on metastore. I found this error template in tonic's source, so I guess that error occurs on large grpc response of splits.

fmassot:
Fix #3576 

## MergePolicy update

This PR modifies the `MergePolicy` by replacing `is_mature` method  by `split_maturity`

```rust

trait MergePolicy {
    fn split_maturity(
        &self,
        split_num_docs: usize,
        split_num_merge_ops: usize,
    ) -> SplitMaturity;
}

/// `SplitMaturity` defines the maturity of a split, is is either `Mature`
/// or becomes mature after a given period.
#[derive(Clone, Copy, Debug, Default, Eq, Serialize, Deserialize, PartialEq)]
#[serde(tag = "type")]
#[serde(rename_all = "snake_case")]
pub enum SplitMaturity {
    #[default]
    Mature,
    Immature {
        #[serde(with = "serde_duration_millisecs")]
        maturation_period: Duration,
    },
}


```

## Migration

Splits from the old quickwit version will be considered mature.

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexer.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexer.rs
@@ -458,6 +458,7 @@ impl Handler<NewPublishLock> for Indexer {
 }
 
 impl Indexer {
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         pipeline_id: IndexingPipelineId,
         doc_mapper: Arc<dyn DocMapper>,
@@ -559,6 +560,7 @@ impl Indexer {
         // Dropping the indexing permit explicitly here for enhanced readability.
         drop(indexing_permit);
 
+        // Update the time to maturity of the splits as they won't receive documents anymore.
         let mut splits: Vec<IndexedSplitBuilder> = indexed_splits.into_values().collect();
 
         if let Some(other_split) = other_indexed_split_opt {
@@ -767,10 +769,9 @@ mod tests {
         let batch = messages.into_iter().next().unwrap();
         assert_eq!(batch.commit_trigger, CommitTrigger::NumDocsLimit);
         assert_eq!(batch.splits[0].split_attrs.num_docs, 4);
-        assert_eq!(
-            batch.splits[0].split_attrs.delete_opstamp,
-            last_delete_opstamp
-        );
+        for split in batch.splits.iter() {
+            assert_eq!(split.split_attrs.delete_opstamp, last_delete_opstamp);
+        }
         let index_checkpoint = batch.checkpoint_delta.unwrap();
         assert_eq!(index_checkpoint.source_id, "test-source");
         assert_eq!(
--- a/quickwit/quickwit-indexing/src/actors/indexer.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexer.rs
@@ -458,7 +458,6 @@ impl Handler<NewPublishLock> for Indexer {
 }
 
 impl Indexer {
-    #[allow(clippy::too_many_arguments)]
     pub fn new(
         pipeline_id: IndexingPipelineId,
         doc_mapper: Arc<dyn DocMapper>,
@@ -560,7 +559,6 @@ impl Indexer {
         // Dropping the indexing permit explicitly here for enhanced readability.
         drop(indexing_permit);
 
-        // Update the time to maturity of the splits as they won't receive documents anymore.
         let mut splits: Vec<IndexedSplitBuilder> = indexed_splits.into_values().collect();
 
         if let Some(other_split) = other_indexed_split_opt {
--- a/quickwit/quickwit-indexing/src/merge_policy/mod.rs
+++ b/quickwit/quickwit-indexing/src/merge_policy/mod.rs
@@ -23,14 +23,13 @@ mod stable_log_merge_policy;
 
 use std::fmt;
 use std::sync::Arc;
-use std::time::Duration;
 
 pub(crate) use const_write_amplification::ConstWriteAmplificationMergePolicy;
 use itertools::Itertools;
 pub use nop_merge_policy::NopMergePolicy;
 use quickwit_config::merge_policy_config::MergePolicyConfig;
 use quickwit_config::IndexingSettings;
-use quickwit_metastore::SplitMetadata;
+use quickwit_metastore::{SplitMaturity, SplitMetadata};
 use serde::Serialize;
 pub(crate) use stable_log_merge_policy::StableLogMergePolicy;
 use tracing::{info_span, Span};
@@ -111,24 +110,7 @@ pub trait MergePolicy: Send + Sync + fmt::Debug {
     fn operations(&self, splits: &mut Vec<SplitMetadata>) -> Vec<MergeOperation>;
     /// Returns the duration after which a split becomes mature.
     /// If the split is already mature given its num docs and merge ops, returns `None`.
-    fn split_time_to_maturity(
-        &self,
-        split_num_docs: usize,
-        split_num_merge_ops: usize,
-    ) -> Option<Duration>;
-
-    /// Returns the timestamp at which a split with the given create timestamp becomes mature.
-    /// If the split is already mature given its num docs and merge ops, returns `0`.
-    // fn split_maturity_timestamp(
-    //     &self,
-    //     split_create_timestamp: i64,
-    //     split_num_docs: usize,
-    //     split_num_merge_ops: usize,
-    // ) -> i64 {
-    //     self.split_time_to_maturity(split_num_docs, split_num_merge_ops)
-    //         .map(|time_to_maturity| split_create_timestamp + time_to_maturity.as_secs() as i64)
-    //         .unwrap_or(0)
-    // }
+    fn split_maturity(&self, split_num_docs: usize, split_num_merge_ops: usize) -> SplitMaturity;
 
     /// Checks a bunch of properties specific to the given merge policy.
     /// This method is used in proptesting.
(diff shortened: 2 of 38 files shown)


quickwit/quickwit-indexing/src/actors/indexer.rs
```diff
@@ -456,6 +456,7 @@ impl Handler<NewPublishLock> for Indexer {
 }
 
 impl Indexer {
+    #[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
fmassot: yes... I added one and then remove it... will clean that.

quickwit/quickwit-indexing/src/merge_policy/mod.rs
```diff
@@ -108,8 +109,27 @@ impl fmt::Debug for MergeOperation {
 pub trait MergePolicy: Send + Sync + fmt::Debug {
     /// Returns the list of merge operations that should be performed.
     fn operations(&self, splits: &mut Vec<SplitMetadata>) -> Vec<MergeOperation>;
-    /// A mature split is a split that won't undergo a merge operation in the future.
-    fn is_mature(&self, split: &SplitMetadata) -> bool;
+    /// Returns the duration after which a split becomes mature.
```
fulmicoton, line 112:
```suggestion
    /// Returns the duration after which a split becomes mature.
    /// A mature split is a split that does not undergo more merge operations,
```

quickwit/quickwit-metastore/src/split_metadata.rs
```diff
@@ -153,6 +159,19 @@ impl SplitMetadata {
...
     }
 
+    /// Returns true if the split is mature at `utc_now_timetamp()`.
+    pub fn is_mature(&self) -> bool {
+        self.maturity_timestamp() <= utc_now_timestamp()
+    }
+
+    /// Returns the timestamp at which the split becomes mature.
+    /// If the split has no time to maturity defined, it returns 0.
+    pub fn maturity_timestamp(&self) -> i64 {
```
fulmicoton, line 169: Why 0. Why a i64 as opposed to a DateTime?

quickwit/quickwit-metastore/src/split_metadata.rs
```diff
@@ -153,6 +159,19 @@ impl SplitMetadata {
         &self.split_id
     }
 
+    /// Returns true if the split is mature at `utc_now_timetamp()`.
+    pub fn is_mature(&self) -> bool {
+        self.maturity_timestamp() <= utc_now_timestamp()
+    }
+
+    /// Returns the timestamp at which the split becomes mature.
```
fulmicoton, line 167:
```suggestion
    /// Returns the unix timestamp at which the split becomes mature.
```

quickwit/quickwit-metastore/src/split_metadata.rs
```diff
@@ -101,6 +102,11 @@ pub struct SplitMetadata {
     /// Timestamp for tracking when the split was created.
     pub create_timestamp: i64,
 
+    /// Timestamp for tracking when the split becomes mature.
+    /// If a split is already mature, this timestamp is set to 0.
```
fulmicoton, line 106: the comment is wrong

fulmicoton:
> Ideally, I don't want to use Duration but to have a number of seconds directly.

Why? Types are really great for this...

> The split metadata comes with two new methods, one returning the maturity_timestamp. This maturity_timestamp is set to 0 if no time to maturity is defined for this split.
> I could have hidden this hackish maturity_timestamp = 0 in the metastore implementations, but I thought having this rule defined at the SplitMetadata level could be helpful for readers.

Is the idea here to make sure that existing splits end up mature after the upgrade of quickwit?

fmassot:
> > Ideally, I don't want to use Duration but to have a number of seconds directly.
> 
> Why? Types are really great for this...

Yes indeed, what I really would like is to have a struct at the seconds precision only.

> Is the idea here to make sure that existing splits end up mature after the upgrade of quickwit? 

Not really. In the case where you want to filter splits by `maturity_timestamp`, I thought it could be useful to expose publicly the way it is built.

Now the truth is that I'm not satisfied with the solution...

I'm thinking of this other solution:

- `time_to_maturity: Duration` instead of `Option<Duration>`. If the maturity split is already mature, `time_to_maturity` is 0.
- `maturity_timestamp: i64`, equals to `create_timestamp + time_to_maturity`

This solution has one drawback: the time of the system can be different between servers, so one indexer can potentially fetch splits that are already mature;

quickwit/quickwit-indexing/src/split_store/indexing_split_store.rs
```diff
@@ -66,11 +66,6 @@ struct InnerIndexingSplitStore {
     remote_storage: Arc<dyn Storage>,
 
     local_split_store: Arc<LocalSplitStore>,
-
-    /// The merge policy is useful to identify whether a split
-    /// should be stored in the local storage or not.
-    /// (mature splits do not need to be stored).
-    merge_policy: Arc<dyn MergePolicy>,
```
fulmicoton, line 73: cool

quickwit/quickwit-janitor/src/actors/delete_task_service.rs
```diff
@@ -167,7 +167,6 @@ impl DeleteTaskService {
             index_metadata.index_uid.clone(),
             self.metastore.clone(),
             self.search_job_placer.clone(),
-            index_config.indexing_settings,
```
fulmicoton, line 170: cool

quickwit/quickwit-metastore/src/metastore/mod.rs
```diff
@@ -336,6 +338,15 @@ pub struct ListSplitsQuery {
 
     /// The create timestamp range to filter by.
     pub create_timestamp: FilterRange<i64>,
+
+    /// Is the split mature or immature at a given datetime.
+    pub maturity: Option<SplitMaturityFilter>,
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct SplitMaturityFilter {
```
fulmicoton, line 347: nitpick:
`Bound<OffsetDateTime>` instead of `Option<SplitMaturityFilter>` could have also done the job in an unambiguous manner.

quickwit/quickwit-metastore/src/metastore/mod.rs
```diff
@@ -496,6 +508,16 @@ impl ListSplitsQuery {
         self.create_timestamp.start = Bound::Excluded(v);
         self
     }
+
+    /// Set the maturity filter to match splits that are mature or immature
+    /// at a given datetime.
+    pub fn is_mature(mut self, mature: bool, evaluation_datetime: OffsetDateTime) -> Self {
```
fulmicoton, line 514: Second nitpick...

`is_xxxx` are usually function that accessor returning a `bool`. Here it is tricky because you are building a predicate, which makes it sensible to use this vocab.

`filter_xxx` is ironically even worse, because `filter` itself is ambiguous. Some program call filter stuff that `filter out`, some that `retain`.

I would suggest two methods (we don't use it dynamically I think) called 

`fn retain_mature(now: OffsetDateTime)`
`fn retain_immature(now: OffsetDateTime)`

quickwit/quickwit-metastore/src/split_metadata.rs
```diff
@@ -243,6 +263,42 @@ impl FromStr for SplitState {
...
+    Immature {
+        /// Time to maturity.
+        #[serde(with = "serde_duration_millisecs")]
+        #[serde(rename = "maturation_period_millisecs")]
+        maturation_period: Duration,
+    },
+}
+
+// Serializer/Deserializer of `Duration` in milliseconds.
+pub mod serde_duration_millisecs {
```
fulmicoton, line 287: serde_with has an implem too if you want to add the dependency (I agree it is always a tiny bit sad)

I'd prefer secs than millisecs here. I'm not too fond of reading pinball high scores. :)

fulmicoton (approved):
Approved. Good job. 

I had a couple of nitpicks. Can you at least consider putting duration in secs (it will be a pain to change in the future)? Strongest point being: the timestamp we compre it with is in secs.

The other change I would like you to consider is
`Bound<OffsetDatetime>` instead of `Option<SplitMaturityFilter>`.

quickwit/quickwit-metastore/src/split_metadata.rs
fmassot: Actually, there is a plan led by @guilload about putting all split metadata "timestamps"  in milliseconds and adding a small wrapper `Timestamp` to avoid having `i64` everywhere.

So we will eventually move to a world where everything is in milliseconds. Thus the duration in millis.

quickwit/quickwit-metastore/src/metastore/mod.rs
fmassot: good idea, `retain` really improves the API.

quickwit/quickwit-metastore/src/split_metadata.rs
guilload: Milliseconds are much better for unit testing. We have a bunch of time::sleep(1s) in our metastore unit tests, making them slow and flaky, because of the poor resolution of our timestamps.

fmassot: For this particular duration, it does not impact the unit tests.

On the contrary, we don't control `create_timestamp`, `update_timestamp`, and `publish_timestamp` values, thus the need for `sleep`.

PR #3591, opened 2023-06-28
Add storage flavors

guilload:
### Description
Add storage flavors for S3-compatible storage providers that deviate from the reference API.

### How was this PR tested?
- Added unit tests

### DIFF
--- a/quickwit/quickwit-config/src/quickwit_config/serialize.rs
+++ b/quickwit/quickwit-config/src/quickwit_config/serialize.rs
@@ -204,7 +204,7 @@ struct QuickwitConfigBuilder {
 
 impl QuickwitConfigBuilder {
     pub async fn build_and_validate(
-        self,
+        mut self,
         env_vars: &HashMap<String, String>,
     ) -> anyhow::Result<QuickwitConfig> {
         let enabled_services = self
@@ -265,6 +265,7 @@ impl QuickwitConfigBuilder {
             .unwrap_or_else(|| default_index_root_uri(&data_dir_uri));
 
         self.storage_configs.validate()?;
+        self.storage_configs.apply_flavors();
 
         let quickwit_config = QuickwitConfig {
             cluster_id: self.cluster_id.resolve(env_vars)?,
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-config/src/storage_config.rs
```diff
@@ -39,6 +39,16 @@ pub enum StorageBackend {
     S3,
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum StorageBackendFlavor {
```
guilload, line 44: @trinity-1686a, do you think it's worth adding a "Garage" flavor that forces path-style access? I know Garage supports both path and virtual-hosted style but the latter requires some DNS support.

trinity-1686a: it would probably be a good idea. Garage doesn't profit in any way from vhost style, so it's just additional constraints for people deploying it to have a DNS wildcard (and probably the TLS certificate that comes with it)

fmassot: I would stick to what we know well today, that is `Garage`, `MinIO`, `Gcs`, there are many unknown storages and I'm afraid that `StorageBackendFlavor::SelfHosted` will not be appropriate for all self-hosted storage.

quickwit/quickwit-config/src/quickwit_config/serialize.rs
```diff
@@ -465,11 +467,14 @@ mod tests {
         );
 
         let s3_storage_config = config.storage_configs.find_s3().unwrap();
+        assert_eq!(s3_storage_config.flavor.unwrap(), StorageBackendFlavor::Gcs,);
```
fmassot, line 470: nitpick

```suggestion
        assert_eq!(s3_storage_config.flavor.unwrap(), StorageBackendFlavor::Gcs);
```

fmassot (approved):

Issue #3582, opened 2023-06-26
GET on `/api/v1/_elastic` should return basic info of the cluster


fmassot:
Something like this should be sufficient: 

```json
{
  "name" : "node-id",
  "cluster_name" : "cluster-name",
  "version" : {
    "distribution" : "quickwit",
    "number" : "0.6.1",
    "build_hash" : "...",
    "build_date" : "...."
  }
}
```

PR #3594, opened 2023-06-29
Add cluster info endpoint for `/_elastic`

evanxg852000:
- Implements `/_elastic` endpoint for showing basic cluster and build info
- Activated OpenAPI docs for the endpoints: `/version`, `/cluster`, and `/config`
Closes https://github.com/quickwit-oss/quickwit/issues/3582


quickwit/quickwit-serve/src/node_info_handler.rs
```diff
@@ -23,16 +23,24 @@ use quickwit_config::QuickwitConfig;
...
+
 pub fn node_info_handler(
     build_info: &'static BuildInfo,
     runtime_info: &'static RuntimeInfo,
     config: Arc<QuickwitConfig>,
 ) -> impl Filter<Extract = (impl warp::Reply,), Error = Rejection> + Clone {
-    node_version_handler(build_info, runtime_info).or(node_config_handler(config))
+    node_version_handler(build_info, runtime_info)
+        .or(node_config_handler(config.clone()))
+        .or(es_compat_info_handler(build_info, config))
```
fmassot, line 40: This is not the place to put ES handler right?

evanxg852000: I think these are the top-level handlers for cluster info. That's why it is built in ES folder but mounted here. functionally they belong to the same group. But I can just mount it there if you think this is more appropriate to you.

fmassot: But we want to keep **all** elasticsearch handlers mounted at the same place, under the `_elastic` path, that's the logic currently coded in there.

evanxg852000: I have changed to mount all elastic-related endpoints from the same location.

fulmicoton:
Can we add a rest-api test for this? It is (hopefully) relatively easy https://github.com/quickwit-oss/quickwit/blob/main/quickwit/rest-api-tests/README.md

fmassot (approved):
LGTM

PR #3601, opened 2023-07-03
Remove unnecessary `Sync` bounds in actor framework

guilload:
### Description
- Actors don't need to be `Sync` because they're behind channels and process messages sequentially
- Messages sent to actors don't need to be `Sync` either.

I used `SyncWrapper` in this PR to make the actor loop future `Sync`.

### How was this PR tested?
- added unit tests
- `make test-all`

### DIFF
--- a/quickwit/quickwit-actors/src/actor.rs
+++ b/quickwit/quickwit-actors/src/actor.rs
@@ -105,9 +105,9 @@ impl From<SendError> for ActorExitStatus {
 /// - update its state;
 /// - emits one or more messages to other actors.
 #[async_trait]
-pub trait Actor: Send + Sync + Sized + 'static {
+pub trait Actor: Send + Sized + 'static {
     /// Piece of state that can be copied for assert in unit test, admin, etc.
-    type ObservableState: Send + Sync + Clone + serde::Serialize + fmt::Debug;
+    type ObservableState: fmt::Debug + serde::Serialize + Send + Sync + Clone;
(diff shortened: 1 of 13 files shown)


quickwit/quickwit-actors/src/actor.rs
```diff
@@ -105,9 +105,9 @@ impl From<SendError> for ActorExitStatus {
 /// - update its state;
 /// - emits one or more messages to other actors.
 #[async_trait]
-pub trait Actor: Send + Sync + Sized + 'static {
+pub trait Actor: Send + Sized + 'static {
     /// Piece of state that can be copied for assert in unit test, admin, etc.
-    type ObservableState: Send + Sync + Clone + serde::Serialize + fmt::Debug;
+    type ObservableState: fmt::Debug + serde::Serialize + Send + Sync + Clone;
```
trinity-1686a, line 110: does the observable state needs to be Sync?

trinity-1686a (approved):
lgtm

quickwit/quickwit-actors/src/actor.rs
guilload: Yes, it needs to be `Sync` because of the watch channel. Under the hood, it's using a RwLock, I found out the hard way in the parking_lot documentation:

> The type parameter `T` represents the data that this lock protects. It is
/// required that `T` satisfies `Send` to be shared across threads and `Sync` to
/// allow concurrent access through readers.

PR #3605, opened 2023-07-04
GCP PubSub Source

magnalite:
This is an initial implementation of a pubsub source. While it works there are a few concerns and unimplemented areas I'll outline. Most of them have an associated comment in the code. I have tested this on a live pubsub instance (but added no unit/integration tests) and it seems to scale linearly very well. I don't believe there is any practical limit on how far it can scale horizontally, similar to the kafka source.

This source works by generating a random partition id (to support multiple pipelines) and then pulls as many messages as it can. As we have little control over how many messages we can pull in one pubsub request it sends multiple requests concurrently (typically the request will return the max of 2000, but can be as low as 1 even when there are many messages to pull). `pull_parallelism` is the current terminology for how many requests it will send in parallel. It then pushes these messages to a batch and acks the messages on `suggest_truncate`. The usefulness of this pull parallelism depends on the latency to the pubsub api. When operating on GCP it is fairly unnecessary and mostly adds overhead but at higher latency it significantly increases throughput per pipeline.

The biggest concern at the moment is it does not guarantee at-least-once delivery. Currently it will blindly ack all messages being processed including any not included in submitted batches. This was a shortcut just while I was testing and should be easy to fix.

It also relies upon the ack deadline being configured on gcp to be suitably large enough (in my case I went for the max deadline of 600 seconds). While there is a mechanism for us to extend the ack deadline programatically, the package I chose to use does not expose it publicly for some reason. It does however implement deadline extension internally but not in a way that is useful to us.

This also currently has no support for ordered subscriptions or subscriptions configured for exactly once delivery. These are significantly more difficult to support mainly due to the requirement to ack every message in the same order it was given and refusing to send more messages until they have been acked. Achieving high throughput in these circumstances seems prohibitively convoluted and I'm not sure is even possible given the typical ingest delay of publishing to a split. The source would need a fast WAL to be able to ack messages fast enough and even then it seems difficult.

Additionally the volume of data sent just confirming ack ids is significant. Each ack id is a huge 192 bytes. While they are usually similar to each other, often only varying by one character, they change randomly and sometimes change every character. There may be a way around this by using snapshots however they are complicated and I'm not convinced they are worth the effort.


fulmicoton:
the review is for you @guilload !

AyWa:
@magnalite any help we can provide on the PR ?  I am still pretty new to contributing in Quickwit, but I have some experience with google pubsub.

I think we could use [pubsub emulator](https://cloud.google.com/pubsub/docs/emulator) to make some integration test with docker ? 
Any help to achieve the at least once guarantee ? I think this pr is almost there but need to add the logic to ack only after the `suggest_truncate` ?

> While there is a mechanism for us to extend the ack deadline programatically, the package I chose to use does not expose it publicly for some reason. It does however implement deadline extension internally but not in a way that is useful to us.

I don't know the rust package, but other high level clients (go etc) are usually extending automatically the deadline by calling a `modify_ack_deadline` req. In go it is configurable with the `MaxExtension`. 
I will try to have a look into the rust package. If possible, we could just set `MaxExtension` as the max `quickwit 
 index.commit_timeout_secs` ? 
 
 > This also currently has no support for ordered subscriptions or subscriptions configured for exactly once delivery
 
I would suggest to also avoid exactly once delivery until there is strong usecase for it. At least it would be better to have a first version with a at least once ?

guilload:
Hello @AyWa,

@magnalite is exploring alternative solutions for ingesting data into Quickwit and will no longer work on this PR. You can take it over.

Here is what I suggest:
- step 1: fix the compilation/formatting issues in this PR, put the code behind a `gcp` flag, and merge the PR.
- step 2: focus on supportting at-least-once delivery semantics for now and improve the current implementation as you suggested
- step 3: add unit and integration tests for this source
- step 4: test at scale (I can help with that and do so on our GCP account)
- step 5: opportunistically support exactly-once delivery semantics

Otherwise,

> I think we could use pubsub emulator to make some integration test with docker ?
Sounds good.

> we could just set MaxExtension as the max quickwit index.commit_timeout_secs?
The approach sounds good. However, `commit_timeout_secs` does not take into account the time required to commit/serialize/upload/publish the split, so I suggest `commit_timeout_secs` * 2.

Please feel free to open a thread on #quickwit-dev on Discord and ping me if you need anything.

guilload:
I merged this branch into another one living in this repo called [`gcp-pubsub-source`](https://github.com/quickwit-oss/quickwit/tree/gcp-pubsub-source), and I gave write access to both of you.

guilload:
It is acceptable, but since there's no GC for checkpoint partitions at the moment, the checkpoint will grow indefinitely. In addition, except for logging purposes, the checkpoint does not help us with queues such as AWS SQS or GCP PubSub, where concurrent consumption of a topic is performed via visibility timeouts.

For now, I'd go with `<node_id>/<pipeline_ord>` for the partition ID used by each pipeline, I would enumerate the doc batches sent downstream, and store ranges in the position so I could ack them properly on suggest truncate.

PR #3614, opened 2023-07-07
sort by two fast fields

PSeitz:
### Description

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-proto/protos/quickwit/search_api.proto
+++ b/quickwit/quickwit-proto/protos/quickwit/search_api.proto
@@ -95,21 +95,30 @@ message SearchRequest {
   // deprecated tag field
   reserved 8;
 
-  // Sort order
-  optional SortOrder sort_order = 9;
+  // deprecated `sort_order``
+  reserved 9;
 
   // Sort by fast field. If unset sort by docid
   // sort_by_field can be:
   // - a field name
   // - _score
   // - None, in which case the hits will be sorted by (SplitId, doc_id).
-  optional string sort_by_field = 10;
+  // deprecated `sort_by_field``
+  reserved 10;
 
   // json serialized aggregation_request
   optional string aggregation_request = 11;
 
   // Fields to extract snippet on
   repeated string  snippet_fields = 12;
+
+  // Optional sort by one or more fields (limited to 2 at the moment). 
+  repeated SortField sort_fields = 14;
+}
+
+message SortField {
+  string field_name = 1;
+  SortOrder sort_order = 2;
 }
 
 enum SortOrder {
@@ -226,16 +235,13 @@ message PartialHit {
   // - the split_id,
   // - the segment_ord,
   // - the doc id.
-  oneof sort_value {
-    uint64 u64 = 5;
-    int64 i64 = 6;
-    double f64 = 7;
-    bool boolean = 8;
-  }
+
   // Deprecated
   reserved 1;
   // Room for eventual future sorted key types.
-  reserved 9 to 20;
+  reserved 12 to 20;
+  SortBy sort_value = 10;
+  SortBy sort_value2 = 11;
 
   string split_id = 2;
 
@@ -245,6 +251,17 @@ message PartialHit {
 
   // The DocId identifies a unique document at the scale of a tantivy segment.
   uint32 doc_id = 4;
+}
+
+message SortBy {
+  oneof sort_value {
+    uint64 u64 = 1;
+    int64 i64 = 2;
+    double f64 = 3;
+    bool boolean = 4;
+  }
+  // Room for eventual future sorted key types.
+  reserved 5 to 20;
 }
 
 message LeafSearchResponse {
--- a/quickwit/quickwit-proto/protos/quickwit/search_api.proto
+++ b/quickwit/quickwit-proto/protos/quickwit/search_api.proto
@@ -235,8 +235,8 @@ message PartialHit {
   reserved 1;
   // Room for eventual future sorted key types.
   reserved 12 to 20;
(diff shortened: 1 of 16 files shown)


quickwit/quickwit-proto/protos/quickwit/search_api.proto
```diff
@@ -95,21 +95,30 @@ message SearchRequest {
...
-  // Sort order
-  optional SortOrder sort_order = 9;
+  // deprecated `sort_order``
+  reserved 9;
 
   // Sort by fast field. If unset sort by docid
   // sort_by_field can be:
   // - a field name
   // - _score
   // - None, in which case the hits will be sorted by (SplitId, doc_id).
```
fulmicoton, line 105:
```suggestion
```

quickwit/quickwit-search/src/collector.rs
```diff
@@ -44,17 +44,24 @@ pub(crate) enum SortBy {
         field_name: String,
         order: SortOrder,
     },
+    TwoFastField {
```
fulmicoton, line 47:
```suggestion
    TwoFastField {
```

I'm not fond of the approach...
We end up not being able to support `DocId` and score for not much reason at all.

I think we can decouple the concept of 
- "source of single scalar sort value": the SortBy object before this PR
- using a pair of value for sorting.

I suggest we keep the `SortBy` object and rename it as `SortByComponent`. 
It would have the same `sort_order` and `recover_typed_sort_value` as before.

We can then have a `SortBy` object that goes
```
struct SortByPair {
    first: SortByComponent,
    second: Option<SortByComponent>,
}
```

and have it implement the methods with the prototypes you introduced in this PR 
`pub fn sort_orders(&self) -> (SortOrder, SortOrder)`
and
`fn recover_typed_sort_value(
        &self,
        sort_value1: Option<u64>,
        sort_value2: Option<u64>,
    ) -> (Option<SortValue>, Option<SortValue>)`

quickwit/quickwit-proto/src/lib.rs
```diff
@@ -464,7 +463,27 @@ pub fn query_ast_from_user_text(user_text: &str, default_fields: Option<Vec<Stri
 // Prost imposes the PartialEq derived implementation.
 // This is terrible because this means Eq, PartialEq are not really in line with Ord's implementation.
 // if in presence of NaN.
```
fulmicoton, line 465: the comment above was targettting SortValue.

quickwit/quickwit-proto/src/lib.rs
```diff
@@ -464,7 +463,27 @@ pub fn query_ast_from_user_text(user_text: &str, default_fields: Option<Vec<Stri
...
+impl PartialOrd for SortBy {
+    fn partial_cmp(&self, other: &SortBy) -> Option<Ordering> {
+        self.sort_value.partial_cmp(&other.sort_value)
+    }
```
fulmicoton, line 469: I think this is derivable no? You can add derive attributes in the prost config.

quickwit/quickwit-proto/protos/quickwit/search_api.proto
```diff
@@ -226,16 +230,13 @@ message PartialHit {
...
-    double f64 = 7;
-    bool boolean = 8;
-  }
+
   // Deprecated
   reserved 1;
   // Room for eventual future sorted key types.
-  reserved 9 to 20;
+  reserved 12 to 20;
+  SortBy sort_value = 10;
```
fulmicoton, line 238: can we call the property sort_by too?

quickwit/quickwit-proto/protos/quickwit/search_api.proto
```diff
@@ -226,16 +230,13 @@ message PartialHit {
...
+  SortBy sort_value2 = 11;
```
fulmicoton, line 240: same as above.

quickwit/quickwit-proto/protos/quickwit/search_api.proto
fulmicoton: (There is also a SortValue thing that makes everything confusing)

fulmicoton (changes requested):
Good job. Can you:
- have a look at the rename?
- add a rest api test? (probably in es-compat-blablabla)

quickwit/quickwit-proto/protos/quickwit/search_api.proto
PSeitz: I renamed it to `SortByValue` as SortBy is not really accurate. It's unnecessary, but I don't know how to get rid of it with prost

quickwit/quickwit-proto/src/lib.rs
PSeitz: yes

fulmicoton (approved):

quickwit/rest-api-tests/scenarii/sort_orders/0001-sort-elasticapi.yaml
```diff
@@ -10,10 +10,12 @@ json:
 expected:
   hits:
     total:
-      value: 5
+      value: 7
       relation: "eq"
     hits:
       - fields: { "count": 15, "id": 2 }
+      - fields: { "count": 10, "id": 0 }
```
fulmicoton, line 17: the bummer with those test is that they do not test against elasticsearch.

We need to add a set up for elastic like it is done in the es-compat-tests

PSeitz: Yes they are just regular integration tests, but that's also fine I think

PR #3623, opened 2023-07-11
Run coverage workflow on BuildJet

guilload:
### Description
Run coverage workflow on BuildJet. The key was to increase the amount of memory while not increasing the number of jobs as quickly. Thanks for the tip @imotov.

### How was this PR tested?
Ran workflow multiple times:
https://github.com/quickwit-oss/quickwit/actions/runs/5520393533
https://github.com/quickwit-oss/quickwit/actions/runs/5520702941
https://github.com/quickwit-oss/quickwit/actions/runs/5520566576
https://github.com/quickwit-oss/quickwit/actions/runs/5521004488

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -14,7 +14,7 @@ on:
 env:
   AWS_ACCESS_KEY_ID: "placeholder"
   AWS_SECRET_ACCESS_KEY: "placeholder"
-  AWS_REGION: "us-east-1"
+  AWS_REGION: us-east-1
   CARGO_INCREMENTAL: 0
   QW_DISABLE_TELEMETRY: 1
   QW_S3_ENDPOINT: "http://localhost:4566" # Services are exposed as localhost because we are not running coverage in a container.
@@ -24,7 +24,7 @@ env:
 jobs:
   test:
     name: Coverage
-    runs-on: ubuntu-latest
+    runs-on: buildjet-8vcpu-ubuntu-2204 # ubuntu-latest
     # Setting a containing will require to fix the QW_S3_ENDPOINT to http://localstack:4566
     services:
       localstack:
@@ -34,15 +34,15 @@ jobs:
           - "4571:4571"
           - "8080:8080"
         env:
-          # `kinesalite` provides a more accurate implementation than
-          # the default Kinesis provider (`kinesis-mock`).
-          KINESIS_PROVIDER: kinesalite
           SERVICES: kinesis,s3
         options: >-
           --health-cmd "curl -k https://localhost:4566"
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
+        credentials:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_ACCESS_TOKEN }}
 
       postgres:
         image: postgres:latest
@@ -57,6 +57,9 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
+        credentials:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_ACCESS_TOKEN }}
 
       kafka-broker:
         image: confluentinc/cp-kafka:7.2.1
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yml
```diff
@@ -24,7 +24,7 @@ env:
 jobs:
   test:
     name: Coverage
-    runs-on: ubuntu-latest
+    runs-on: buildjet-8vcpu-ubuntu-2204 # ubuntu-latest
```
guilload, line 27: 8 vCPUs, 32 GB memory.

.github/workflows/coverage.yml
```diff
@@ -34,15 +34,15 @@ jobs:
           - "4571:4571"
           - "8080:8080"
         env:
-          # `kinesalite` provides a more accurate implementation than
```
guilload, line 37: `kinesalite` is deprecated.

.github/workflows/coverage.yml
```diff
@@ -34,15 +34,15 @@ jobs:
...
-          # the default Kinesis provider (`kinesis-mock`).
-          KINESIS_PROVIDER: kinesalite
           SERVICES: kinesis,s3
         options: >-
           --health-cmd "curl -k https://localhost:4566"
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
+        credentials:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
```
guilload, line 44: Avoid Docker pull failures. Unlogged pulls are heavily rate limited.

.github/workflows/coverage.yml
```diff
@@ -106,7 +118,6 @@ jobs:
           path: |
             ~/.cargo/git
             ~/.cargo/registry
-            target
```
guilload, line 109: No need to cache the `target` file because we always start with `cargo clean`.

.github/workflows/coverage.yml
```diff
@@ -127,24 +138,20 @@ jobs:
...
+          tool: cargo-llvm-cov,nextest,protoc
 
+      # We limit the number of jobs to 4 to avoid OOM errors when linking the binary.
       - name: Generate code coverage
         run: |
           cargo llvm-cov clean --workspace
-          cargo llvm-cov --test failpoints --no-report --features fail/failpoints
-          cargo llvm-cov --no-report --all-features
-          cargo llvm-cov --no-run --lcov --output-path lcov.info
+          cargo llvm-cov nextest --no-report --test failpoints --features fail/failpoints --retries 2
```
guilload, line 152: Run tests in parallel with `nextest`.

.github/workflows/coverage.yml
```diff
@@ -127,24 +138,20 @@ jobs:
...
+          CARGO_BUILD_JOBS=4 cargo llvm-cov nextest --no-report --all-features --retries 2
```
guilload, line 153: Can we go a bit higher? 5 or 6?

imotov: I think hiding flaky tests here by increasing the number might create some issues for use during normal PR runs.

docker-compose.yml
```diff
@@ -82,14 +79,14 @@ services:
       - "${MAP_HOST_PULSAR:-127.0.0.1}:6650:6650"
       - "${MAP_HOST_PULSAR:-127.0.0.1}:8081:8080"
     environment:
-      PULSAR_MEM: " -Xms512m -Xmx512m -XX:MaxDirectMemorySize=2g"
+      PULSAR_MEM: " -Xmx256M"
```
imotov, line 82: I would keep `-Xms512m` to avoid pauses for memory allocations. It's one of the best practices to specify both -Xms and -Xmx and set them to the same value.

docker-compose.yml
```diff
@@ -110,6 +107,7 @@ services:
       KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
       KAFKA_JMX_PORT: 9101
       KAFKA_JMX_HOSTNAME: localhost
+      KAFKA_HEAP_OPTS: -Xmx256M
```
imotov, line 110: Same here.

docker-compose.yml
```diff
@@ -127,6 +125,7 @@ services:
       - all
       - kafka
     environment:
+      KAFKA_HEAP_OPTS: -Xmx256M
```
imotov, line 128: And here.

.github/workflows/coverage.yml
imotov: It doesn't fail here?

imotov (approved):
LGTM

Added a couple of suggestions for java flags.

.github/workflows/coverage.yml
guilload: Not to my knowledge.

.github/workflows/coverage.yml
guilload: Can we start at 2, then gradually go to 0? We still have a few flaky tests that have not been fixed. Moreover, unless you commit to Azure storage, Kafka, Kinesis, and Pulsar sources, which we seldom do, the "regular" CI workflow should have already been executed on our code.

docker-compose.yml
guilload: So you mean "-Xms256M -Xmx256M" or "-Xms512M -Xmx512M"? I did multiple runs with 256 and that did not cause any issue.

.github/workflows/coverage.yml
imotov: I think 2 is reasonable at the moment, my comment was against going to 5-6. 2 gives test a second chance and works great reducing noise from low frequency failures. But if something requires 3+ attempts it to succeeded it is no longer a low frequency failure and should probably be fixed.

docker-compose.yml
imotov: Yes, setting them to the same value removes the need for heap resizes, which is expensive and tricky operation.

.github/workflows/coverage.yml
guilload: I was talking about the number of compile jobs :D

guilload: Let's stick with 4 as discussed offline.

PR #3636, opened 2023-07-14
Do not ignore storage errors silently when deleting splits

guilload:
### Description
- Do not ignore storage errors silently when deleting splits
- Remove `FileEntry` in favor of `SplitInfo`
- Move `BulkDeleteError` in `storage::error`

### How was this PR tested?
- Added unit tests


quickwit/quickwit-janitor/src/garbage_collection.rs
```diff
@@ -243,100 +235,104 @@ async fn delete_splits_marked_for_deletion(
...
+            "Failed to delete split files {:?} from storage.",
+            PrettySample::new(&storage_failures, 5),
+        );
+        storage_error = Some(bulk_delete_error);
+    }
+    if !split_infos.is_empty() {
+        let split_ids: Vec<&str> = split_infos
+            .values()
+            .map(|split_info| split_info.split_id.as_str())
+            .collect();
```
trinity-1686a, line 282: I'm a bit worried a bug in `bulk_delete` could not add a split in any list (`failures`, `unattempted`, `success`), and we would consider it a success for that split. I think it would be more defensive to use `bulk_delete_error.success` if `bulk_delete` failed, and only consider everything in `split_infos` is okay when `delete_result` is `Ok(_)`

quickwit/quickwit-core/src/lib.rs
```diff
@@ -17,61 +17,6 @@
...
-            let split_num_bytes = test_sandbox
-                .storage()
-                .file_num_bytes(Path::new(file_entry.file_name.as_str()))
-                .await?;
-            assert_eq!(split_num_bytes, file_entry.file_size_in_bytes);
-        }
-        // Now delete the index.
-        let index_service =
-            IndexService::new(test_sandbox.metastore(), StorageResolver::unconfigured());
-        let deleted_file_entries = index_service.delete_index(index_id, false).await?;
```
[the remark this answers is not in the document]
guilload: The CLI tests cover it. Unfortunately, there are no tests in `quickwit_core::index` :/ This test is garbage, though, I really want to remove it. I'll add better ones for the entire file before merging this PR.

quickwit/quickwit-janitor/src/garbage_collection.rs
guilload: I checked the storage code before going with this strategy and we seemed to be doing the right, but I'm ok the more belt and suspenders approach that you describe. The ` .flat_map(|split_path| split_infos.remove(split_path))` was bad anyway. That is what you get for coding on Sunday mornings at 6 am.

PR #3674, opened 2023-07-21
Stop relying on builder image for CI.

fulmicoton:


.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ env:
   RUST_BACKTRACE: 1
   RUSTFLAGS: -Dwarnings
   RUSTDOCFLAGS: -Dwarnings -Arustdoc::private_intra_doc_links
-  TEST_DATABASE_URL: postgres://quickwit-dev:quickwit-dev@postgres:5432/quickwit-metastore-dev
+  TEST_DATABASE_URL: postgres://quickwit-dev:quickwit-dev@localhost:5432/quickwit-metastore-dev
```
fulmicoton, line 19: Without the extra container nesting, postgres is actually accessible as localhost.

PR #3676, opened 2023-07-24
Refactoring of the CI.

fulmicoton:
The previous CI was strangely relying on workflows to split the CI and have it run in parallel.

A lot of the work was made several times:
e.g. installing rust stable + nightly, installing packages, starting our builder image etc.

Overall it was done x6 times. It was strongly overkill considering some of the steps run in (<1s).

After this refactoring we run in two "jobs" with a different set of steps.
- Lint runs clippy + all of the smaller jobs (cargo fmt, cargo deny, checking the license)
- Test runs the unit test.

In addition, we stop relying on the builder image and the associated confusion. For instance, the rust stable installed in the CI workflow was actually not used. The quickwit-builder image one was taken because of the rust-toolchain.toml.

After the PR, we run on stock ubuntu-latest directly. We then rely on rustup to install the right version of cargo/rust as define in the rust-toolchain.
Updating the rust version does not require any extra changes.

cargo deny / cargo nextest need to be build unfortunately. That step is however cached.


fulmicoton:
After the PR is merged, we need to edit the settings of the main branch.

.github/workflows/ci.yml
```diff
@@ -69,35 +45,95 @@ jobs:
...
+        if: steps.modified.outputs.rust_src == 'true'
         uses: Swatinem/rust-cache@v2
         with:
-          key: "v1-${{ matrix.task.name }}"
           workspaces: "./quickwit -> target"
-      - name: Install nextest
-        if: matrix.task.name == 'cargo nextest'
-        uses: taiki-e/install-action@nextest
-      - name: ${{ matrix.task.name }}
-        if: steps.modified-files.outputs.qw-src == 'true' && steps.modified-files.outputs.ui-src == 'false'
```
fulmicoton, line 101: that was broken.

guilload (approved):

PR #3678, opened 2023-07-24
Migrate Control Plane codegen to `quickwit-proto` crate

guilload:
### Description
I'm also trying to align naming and packaging conventions between services.

### How was this PR tested?
`make test-all`


quickwit/quickwit-proto/src/lib.rs
```diff
@@ -21,144 +21,38 @@
...
 use tracing_opentelemetry::OpenTelemetrySpanExt;
+use ulid::Ulid;
 
-pub use tonic;
-
-#[path = "quickwit/quickwit.indexing.rs"]
+pub mod control_plane;
 pub mod indexing;
-#[path = "quickwit/quickwit.metastore.rs"]
+#[path = "codegen/quickwit/quickwit.metastore.rs"]
```
guilload, line 39: At this point, we want some code formatting in `quickwit_proto`, so I moved all the generated files to a `codegen` folder.

imotov (approved):
Really nice clean-up! 

It looks like we could now move the service_client_pool from grpc-clients into search crate and get rid of grpc-clients crate. This crate confused me earlier because I didn't realize it is part of the legacy structure. So, maybe we should also deprecate service_client_pool for good measure.

guilload:
Yes. We might be able to get rid of `quickwit-grpc-clients` altogether because I've already migrated the searcher pool to the "new" pool. Let me see if we use anything from the `ServiceClient` trait.

PR #3690, opened 2023-07-26
Pave way for metastore migration to code-generated gRPC client

guilload:


guilload:
Death by too many conflicts.

PR #3691, opened 2023-07-26
revisit parsing of configuration

### LINKED ISSUE #3472: Revisit the way we handle default / Option::None in configuration
Our configuration has default values. 
In some complex case, there is some value to be able to keep the information
"the value is missing vs it was the default".

Let's pick an example.
To make sure a text is indexed the user just needs to mark the field as indexed. It happens to be the default.
They can also set a tokenizer, or record option.

We currently store those in an `Option`. The `Option` gives us the possibility to do an extra validation phase to let the user know that their configuration does not make sense.

For instance:
```yaml
indexed: false
tokenizer: raw //< the user did set a tokenizer, when in fact the field is not indexed. This is probably an error on their side.
```

When we build the tantivy TextOptions or whatever, we simply use a default value upon access.

This zealous validation leads us to a more serious problem.

We actually never replace this None by the default value. What we do, is that upon access, when we build the tantivy TextOptions, we use the default value in place of None.
 
As a result, when we serialize the configuration to store in the metastore, it will still contain `None`.
This makes backward compatibility a bit of a nightmare. Now getting the behavior correctly will force us to add 
adhoc conversion code when we want to change our default values.

Instead we want to interpret those defaults, and make sure that after reserialization in the Metastore those defaults value are there.

Worst even... This change in semantics cannot be caught by our regression tests as they rely on deserialization / serialization.

There are several ways to get there, involving more or less code.

trinity-1686a:
### Description

fix #3472
Change how configuration is parsed so validation and replacement of default values happens during deserialization.

Sadly Serde doesn't provide anything for manipulating multiple fields (or a whole struct) at once, and I couldn't find a crate to do that either, so I wrote a procmacro to generate a mirror of our strucs for parsing, and a TryFrom/Into impl to use that for serializing/deserializing our config structs, which are now entirely validated.

I've done some special handling of utoipa, but I don't really know how to verify


fulmicoton:
@trinity-1686a The description does not explain what this is improving, whether this is no-op or not etc. Can you explain?

fulmicoton:
@trinity-1686a That's indeed the behavior that we want for most of the values.

fulmicoton:
`#[serde(flatten)]` does what you described. However it does not work in combination with deny_unknown_fields (and probably ruins the error messages, although I am not sure)

quickwit/quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
```diff
@@ -253,24 +253,131 @@ impl QuickwitTextNormalizer {
...
+        }
+    }
+
+    fn to_parts_text(
+        this: Option<Self>,
+    ) -> (
+        bool,
+        Option<QuickwitTextTokenizer>,
+        Option<IndexRecordOption>,
+        bool,
```
fulmicoton, line 312:
```suggestion
        bool,  // indexed
        Option<QuickwitTextTokenizer>,
        Option<IndexRecordOption>,
        bool,  // fieldnorms
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
```diff
@@ -253,24 +253,131 @@ impl QuickwitTextNormalizer {
...
+                this.fieldnorms,
+            ),
+            None => (false, None, None, false),
+        }
+    }
+
+    fn to_parts_json(
+        this: Option<Self>,
+    ) -> (
+        bool,
```
fulmicoton, line 328:
```suggestion
        bool,  // indexed
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
```diff
@@ -253,24 +253,131 @@ impl QuickwitTextNormalizer {
...
+        Option<QuickwitTextTokenizer>,
+        Option<IndexRecordOption>,
+    ) {
+        let (indexed, tokenizer, record, _fieldorm) = TextIndexingOptions::to_parts_text(this);
```
fulmicoton, line 332: thx for having reused `to_parts_text`

fulmicoton (approved):

fulmicoton:
LGTM... Thanks for dealing with the tricky ticket.

trinity-1686a:
> `#[serde(flatten)]` does what you described. However it does not work in combination with deny_unknown_fields

yes, and that's exactly why I decided not to use it. Not rejecting unknown fields in config is imo not an option (it can catch important typos), so `#[serde(flatten)]` wasn't either. While the way it works is different, the result is overall similar

Issue #2991, opened 2023-03-13
Quickwit panics while attempting to parse `index_uri`.


ChillFish8:
Quickwit panics when parsing `http://localhost:9000/quickwit` as the index_uri config value.
The value should be reported as an invalid error since  index_uri expect a [storage_uri](https://quickwit.io/docs/reference/storage-uri)

guilload:
What's the use case for an HTTP index URI?

ChillFish8:
I should clarify, it wasn't meant to be an HTTP uri, but if you give it one it'll panic and say it's a bug rather than say it's invalid.

PR #3692, opened 2023-07-26
[fix] Quickwit panics while attempting to parse index_uri

AyWa:
### Description

Fix https://github.com/quickwit-oss/quickwit/issues/2991
The goal is to ensure that protocol are well validated, instead of having some panic in later stage.

- [x] ensure that protocol are validated during de-serialization
- [x] add test

Before
```
url -XPOST http://localhost:7280/api/v1/indexes -H "content-type: application/yaml" --data-binary @example.yaml
curl: (52) Empty reply from server
```

After
```
curl -XPOST http://localhost:7280/api/v1/indexes -H "content-type: application/yaml" --data-binary @example.yaml
{
  "message": "Invalid config: Failed to read YAML file.: Unknown URI protocol `http`.."
}
```

### How was this PR tested?

* manually by creating an index with an invalid `index_uri`
* added test in Uri

### DIFF
--- a/quickwit/quickwit-common/src/uri.rs
+++ b/quickwit/quickwit-common/src/uri.rs
@@ -486,6 +486,12 @@ mod tests {
             Uri::from_str("azure://account/container/homer/docs/../dognuts").unwrap(),
             "azure://account/container/homer/docs/../dognuts"
         );
+
+        assert_eq!(
+            Uri::from_str("http://localhost:9000/quickwit").unwrap_err().to_string(),
+            "Unknown URI protocol `http`."
+        );
+
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


AyWa:
I understood why the test failed. `pg" | "postgres" | "postgresql` are valid protocol for postgres. And it is correctly parsed.
But for the test assertion, we will transform those into `postgresql` so assertion will failed.

There is few possible way to fix:
* fix the assertion to always expect `postgressql` as they should not impact the connection. And it is also what is happening in the code anyway
* make change so that, the protocol postgres is `PostgreSQL(String)` so that we can keep the "base protocol" inside

fulmicoton (approved):

fmassot:
Ok, with this PR we are now normalizing the PostgreSQL protocol to always have the string representation `postgresql`. I'm good with that as this is the standard, [ref](https://www.postgresql.org/docs/current/libpq-connect.html#id-1.7.3.8.3.6).

fulmicoton (approved):

PR #3704, opened 2023-08-02
Add coerce and output format options for numeric fields

guilload:
### Description
- Add `coerce` option to convert string values to numeric values for numerics fields by default.
- Add `output_format` option to output numeric values as strings

I will update the doc when we agree on the spec.

### How was this PR tested?
Add unit tests


quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -231,41 +238,76 @@ fn insert_json_val(
...
+            JsonValue::String(str_val) if coerce => {
+                str_val.parse::<Self>().map(|num| num.into()).map_err(|_| {
+                    format!(
+                        "Expected {}, got inconvertible JSON string `{}`.",
+                        type_name::<Self>(),
+                        str_val,
+                    )
+                })
+            }
+            _ => Err(format!("Expected JSON number, got `{json_val}`.",)),
```
fulmicoton, line 264: The error is a bit misleading if coerce is set to `true`.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -231,41 +238,76 @@ fn insert_json_val(
...
+                }),
+            JsonValue::String(str_val) if coerce => {
+                str_val.parse::<Self>().map(|num| num.into()).map_err(|_| {
+                    format!(
+                        "Expected {}, got inconvertible JSON string `{}`.",
+                        type_name::<Self>(),
+                        str_val,
+                    )
+                })
+            }
```
fulmicoton, line 263: It would be nice to improve discoverability of the feature as follows...
```
            JsonValue::String(str_val) => {
                let num_val_res: TantivyValue = str_val.parse().map(Self::into);
                if coerce {
                    //...
                } else {
                    if num_val_res.is_ok() {
                         return Err("some message mentioning the existence of the coerce parameter")
                    } else {
                         return Err("message explaining we expected a number")
                    }
                }
            }
```

fulmicoton (approved):
approved but need update on regression tests and doc I assume.

(I'm ok with this spec.)

guilload:
Maybe I should have named the parameter `output_type` instead of `output_format`. Opinions?

Issue #3657, opened 2023-07-17
Improvements for `quickwit_core::index`


guilload:
Todo:
- [ ] test file more thoroughly
- [ ] rename crate to `quickwit-index-management`
- [ ] move some logic (delete index (storage + metastore), garbage collect index) from `quickwit_janitor` to `quickwit-index-management` and inverse dependency. Bonus point for not pulling actors in `quickwit-index-management`.
- [ ] remove needless errors (most of them end up being mapped to `ServiceErrorCode::Internal`

PR #3708, opened 2023-08-03
Index management cleanup

smiklos:
### Description

Follow up on #3657 

```move some logic (delete index (storage + metastore), garbage collect index) from quickwit_janitor to quickwit-index-management and inverse dependency. Bonus point for not pulling actors in quickwit-index-management```

Basically moved the garbage_collection logic to index-management and removed actor dependency from the code. 
janitor now depends on index-management, and not the other way around

Avoided adding new tests to keep the pr small. 

### How was this PR tested?

Using the existing tests (and the compiler 💥)


guilload (approved):
LGTM. Nice clean up. Thanks!

PR #3720, opened 2023-08-07
Gcp pubsub source

AyWa:
### Description

Continuation of the great work of @magnalite and following advice of @guilload 
This PR, fix the compilation/formatting issues of the previous PR and put the code behind a flag `gcp-pubsub`

Also renamed everywhere `pubsub` by `gcp-pubsub` or `GcpPubSub` to reduce confusion with other pubsub system.

### How was this PR tested?

I didnt test it. It will be added in step 3

### Next Item

  step 2: focus on supporting at-least-once delivery semantics for now and improve the current implementation as you suggested
  step 3: add unit and integration tests for this source
  step 4: test at scale (I can help with that and do so on our GCP account)
  step 5: opportunistically support exactly-once delivery semantics

### DIFF
--- a/quickwit/quickwit-config/src/source_config/mod.rs
+++ b/quickwit/quickwit-config/src/source_config/mod.rs
@@ -86,6 +86,7 @@ impl SourceConfig {
         match self.source_params {
             SourceParams::File(_) => "file",
             SourceParams::Kafka(_) => "kafka",
+            SourceParams::PubSub(_) => "pubsub",
             SourceParams::Kinesis(_) => "kinesis",
             SourceParams::Vec(_) => "vec",
             SourceParams::Void(_) => "void",
@@ -100,6 +101,7 @@ impl SourceConfig {
         match &self.source_params {
             SourceParams::File(params) => serde_json::to_value(params),
             SourceParams::Kafka(params) => serde_json::to_value(params),
+            SourceParams::PubSub(params) => serde_json::to_value(params),
             SourceParams::Kinesis(params) => serde_json::to_value(params),
             SourceParams::Vec(params) => serde_json::to_value(params),
             SourceParams::Void(params) => serde_json::to_value(params),
@@ -204,6 +206,8 @@ pub enum SourceParams {
     File(FileSourceParams),
     #[serde(rename = "kafka")]
     Kafka(KafkaSourceParams),
+    #[serde(rename = "pubsub")]
+    PubSub(PubSubSourceParams),
     #[serde(rename = "kinesis")]
     Kinesis(KinesisSourceParams),
     #[serde(rename = "pulsar")]
@@ -245,7 +249,9 @@ pub struct FileSourceParams {
 
 // Deserializing a filepath string into an absolute filepath.
 fn absolute_filepath_from_str<'de, D>(deserializer: D) -> Result<Option<PathBuf>, D::Error>
-where D: Deserializer<'de> {
+where
+    D: Deserializer<'de>,
+{
     let filepath_opt: Option<String> = Deserialize::deserialize(deserializer)?;
     if let Some(filepath) = filepath_opt {
         let uri = Uri::from_str(&filepath).map_err(D::Error::custom)?;
@@ -285,6 +291,24 @@ pub struct KafkaSourceParams {
     #[serde(default)]
     #[serde(skip_serializing_if = "is_false")]
     pub enable_backfill_mode: bool,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, utoipa::ToSchema)]
+#[serde(deny_unknown_fields)]
+pub struct PubSubSourceParams {
+    /// Name of the subscription that the source consumes.
+    pub subscription: String,
+    /// When backfill mode is enabled, the source exits after reaching the end of the topic.
+    #[serde(default)]
+    #[serde(skip_serializing_if = "is_false")]
+    pub enable_backfill_mode: bool,
+    /// GCP service account credentials (None will use default via GOOGLE_APPLICATION_CREDENTIALS)
+    pub credentials: Option<String>,
+    /// How many threads spread pubsub pull requests over (default 10)
+    /// Higher values can mean higher throughput but higher overhead
+    pub pull_parallelism: Option<u64>,
+    /// The max messages to pull per pull request (default 1000)
+    pub max_messages_per_pull: Option<i32>,
 }
 
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, utoipa::ToSchema)]
@@ -385,7 +409,9 @@ pub enum PulsarSourceAuth {
 
 // Deserializing a string into an pulsar uri.
 fn pulsar_uri<'de, D>(deserializer: D) -> Result<String, D::Error>
-where D: Deserializer<'de> {
+where
+    D: Deserializer<'de>,
+{
     let uri: String = Deserialize::deserialize(deserializer)?;
 
     if uri.strip_prefix("pulsar://").is_none() {
--- a/quickwit/quickwit-config/src/source_config/mod.rs
+++ b/quickwit/quickwit-config/src/source_config/mod.rs
@@ -86,7 +86,7 @@ impl SourceConfig {
         match self.source_params {
             SourceParams::File(_) => "file",
             SourceParams::Kafka(_) => "kafka",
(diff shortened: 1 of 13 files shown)


quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -0,0 +1,355 @@
...
+
+        loop {
+            let mut handles = vec![];
+
+            for _ in 1..self.pull_parallelism {
+                handles.push(self.pull_message_batch(Arc::clone(&batch_lock)));
+            }
+
+            tokio::select! {
+                _ = futures::future::join_all(handles) => {
```
guilload, line 213: We should revisit how we handle pull parallelism in this source. Let's spawn one tokio task per `pull_parallelism` and have each task send a pull batch into a shared bounded channel. You can take a look at how it is done in the Kinesis source.

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -0,0 +1,355 @@
...
+        )
+    }
+
+    fn observable_state(&self) -> JsonValue {
+        JsonValue::Object(Default::default())
+    }
+}
+
+impl GcpPubSubSource {
+    async fn ack_ids(&mut self) -> anyhow::Result<()> {
```
guilload, line 282: This also acknowledges batches that are in-flight and have not been published yet. We need to encode our position in the checkpoint that is sent downstream so that when we get a checkpoint back in the suggest truncate, we can ack the published messages only.

I think we can attach a unique ID to each batch sent downstream associated with a set of message IDs to ack in the future:

```rust
type BatchId = Ulid; // Ulids are monotonically increasing
type MessageIDs = Vec<String>

let in_flight_batches: VecDeque<(BatchId, MessageIDs)>

async fn emit_batches(
        &mut self,
        doc_processor_mailbox: &Mailbox<DocProcessor>,
        ctx: &SourceContext,
    ) -> Result<Duration, ActorExitStatus> {
    let now = Instant::now();
    let bach_id = Ulid::new();
    let mut batch_builder = ...;
    let mut message_ids = Vec::new();
    let deadline = time::sleep(*quickwit_actors::HEARTBEAT / 2);
    tokio::pin!(deadline);
    
    loop {
            tokio::select! {
                batch = self.batch_tx.recv() => {
                    for message in batch {
                      batch_builder.add_doc(message.doc);
                      message_ids.push(message.id);
                    }
                    if batch_builder.num_bytes >= BATCH_NUM_BYTES_LIMIT {
                        break;
                    }
                }
                _ = &mut deadline => {
                    break;
                }
            }
            ctx.record_progress();
        }
        let message = batch_builder.build();
        self.in_flight_batches.push((batch_id, message_ids));
         ctx.send_message(doc_processor_mailbox, message).await?;
        }
        Ok(Duration::default())
    }
    
    async fn suggest_truncate(
        &mut self,
        source_checkpoint: SourceCheckpoint
        ctx: &SourceContext,
    ) -> Result<Duration, ActorExitStatus> {
      let max_batch_id = // extract last batch ID published from checkpoint
     // pop batches from `in_flight_batches` from the left as long as batch_id <= max_batch_id and ack them
    }
```

For simplicity, we can do 0 parallelism and no suggest truncate in this PR and implement that in separate PRs.

quickwit/quickwit-config/src/source_config/mod.rs
```diff
@@ -204,6 +206,8 @@ pub enum SourceParams {
     File(FileSourceParams),
     #[serde(rename = "kafka")]
     Kafka(KafkaSourceParams),
+    #[serde(rename = "gcp-pubsub")]
```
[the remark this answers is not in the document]
AyWa: Did it except for `ingest-api`, `ingest-cli` because rename those will be breaking change right ?

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -0,0 +1,241 @@
...
+        );
+
+        Ok(GcpPubSubSource {
+            ctx,
+            subscription,
+            state: GcpPubSubSourceState::default(),
+            subscription_source,
+            backfill_mode_enabled,
+            // TODO: get the real value
+            partition_id: "<node_id>/<pipeline_ord>".to_string(),
```
AyWa, line 120: @guilload can you give some guidance where I can get those information ? like you suggested in https://github.com/quickwit-oss/quickwit/pull/3605#issuecomment-1668563149

AyWa (commented):
I tried to applied all your comment @guilload. Let me know if I misunderstood something.

In next PR, I will try to add the ACK of the message with some test.

Then I will try to add parallelism if needed, and some integration test.

quickwit/quickwit-config/src/source_config/mod.rs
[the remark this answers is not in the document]
guilload: Yes, that's correct.

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
guilload: Yes, a refactor of `SourceExecutionContext` is required to have access to those parameters. Let me push a PR before the end of the week.

quickwit/quickwit-indexing/src/source/pulsar_source.rs
```diff
@@ -33,18 +33,15 @@ use pulsar::{
...
 use quickwit_proto::IndexUid;
 use serde_json::{json, Value as JsonValue};
 use tokio::time;
 use tracing::{debug, info, warn};
 
 use crate::actors::DocProcessor;
-use crate::models::RawDocBatch;
 use crate::source::{
-    Source, SourceActor, SourceContext, SourceExecutionContext, TypedSourceFactory,
+    BatchBuilder, Source, SourceActor, SourceContext, SourceExecutionContext, TypedSourceFactory,
```
guilload, line 44: The Kafka source is also relying on a duplicated `BatchBuilder`.

quickwit/quickwit-indexing/src/source/mod.rs
```diff
@@ -369,6 +377,35 @@ impl Handler<SuggestTruncate> for SourceActor {
...
+
+impl BatchBuilder {
+    pub fn build(self) -> RawDocBatch {
+        RawDocBatch {
+            docs: self.docs,
+            checkpoint_delta: self.checkpoint_delta,
+            force_commit: false,
+        }
+    }
+    pub fn build_force(self) -> RawDocBatch {
```
guilload, line 395: This method is never used, right? Forcing a commit can only be done via the ingest source.

AyWa: current code is doing that:
```
        if self.should_exit() {
            info!(subscription = %self.subscription, "Reached end of subscription.");
            ctx.ask(doc_processor_mailbox, batch.build_force())
                .await
                .context("Failed to force commit last batch!")?;
            ctx.send_exit_with_success(doc_processor_mailbox).await?;
            return Err(ActorExitStatus::Success);
        }
```

I think if it is not working, then, I can remove this code, we could exit after we got call back of commit

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
AyWa: I updated my code, I put `append_random_suffix` to not be block by it. I can add those parameter after the refactor is done

quickwit/quickwit-indexing/src/source/mod.rs
AyWa: I removed the code (build force). In next PR (at_least once impl), I will ensure we exit only when all msg are ack

guilload: Sounds good. Thanks.

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
guilload: Sounds good.

guilload (approved):

guilload:
Merged! Awesome job @magnalite and @AyWa. It's really great to land large features from OSS contributors.

@AyWa, as a next step, I would welcome a PR that adds some tests before adding more new functionalities to the source (parallelism, ack on truncate).

fmassot:
@AyWa @magnalite Ping me on Discord if you want some swag :)

PR #3723, opened 2023-08-08
POC for ssd cache

### LINKED ISSUE #3443: SSD Split "cache"

imotov:
### Description

Proof of concept for adding SSD cache to make sure this architecture is feasible. It is still in the very early stages, not all features are implemented.

To setup add the following node configuration is needed:

```
storage:
  cache:
    storage_uri: file:///tmp/storage
    max_cache_storage_disk_usage: "5G"
```

On the index side the following setting should be added to the config:

```
    "index_uri": "cache://file:///tmp/path_to/index/index_id'",
```

Fixes #3443

### How was this PR tested?

Still need to add tests.

### Outstanding issues

This is still WIP. The following tasks are still need to be finished for MVP:
- [ ] Clean up cache storage configuration (needs some reasonable defaults)
- [ ] Comprehensive unit testing of subcomponents
- [ ] Rehydration of the current cache split state on the node startup
- [ ] Limiting the maximum space used by downloaded and in-flight splits
- [ ] Postponing split rebalancing until the cluster reached the desired number of nodes or a certain time passed after the number of nodes threshold has been breached.
- [ ] Throttling the number of simulations split downloads and network load on search nodes.
- [ ] Make storage copy_to_storage more robust
- [ ] Optimize reading metadata with a large rate of change
- [ ] Modify URI format to `cache:file:///tmp`

Beyond MVP:
- [ ] Refactor Storage Configuration


quickwit/quickwit-config/src/storage_config.rs
```diff
@@ -140,6 +143,15 @@ impl StorageConfigs {
...
             })
     }
+
+    pub fn find_cache(&self) -> Option<&CacheStorageConfig> {
+        self.0
+            .iter()
+            .find_map(|storage_config| match storage_config {
+                StorageConfig::Cache(cache_storage_config) => Some(cache_storage_config),
+                _ => None,
+            })
```
fulmicoton, line 153: personal nitpick. I tend to prefer to avoid rare combinators.

```rust
for storage_config in &self.0 {
    if let StorageConfig::Cache(cache_storage_config) = storage_config {
        return Some(cache_storage_config);
    }
}
None
```

quickwit/quickwit-control-plane/src/cache_storage_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+};
+use serde::Serialize;
+use tokio::sync::RwLock;
+use tower::timeout::Timeout;
+use tracing::{debug, error};
+
+#[derive(Debug, Clone, Default, Serialize)]
+pub struct CacheStorageControllerState {}
+
+/// Resides in the control plane. Responsible for recieving notification about the changes in
```
fulmicoton, line 47:
```suggestion
/// Resides in the control plane. Responsible for receiving notification about the changes in
```

quickwit/quickwit-control-plane/src/cache_storage_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+/// published splits and forwarding these notification to CacheStorageServices that resides on
```
fulmicoton, line 48:
```suggestion
/// published splits and forwarding these notifications to CacheStorageServices that resides on
```

quickwit/quickwit-control-plane/src/cache_storage_controller.rs
```diff
@@ -0,0 +1,299 @@
...
+/// search nodes, that in turn are responsible for maintaining the local cache.
+pub struct CacheStorageController {
+    metastore: Arc<dyn Metastore>,
+    state: CacheStorageControllerState,
+    split_to_node_map: HashMap<String, HashSet<String>>,
+    pool: CacheStorageServicePool,
+}
+
+impl fmt::Debug for CacheStorageController {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
fulmicoton, line 58:
```suggestion
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```

quickwit/quickwit-config/src/storage_config.rs
imotov: This code is repeated 5 times in this file. We should probably replace all occurrences then.

imotov:
Superseded by #3786

PR #3724, opened 2023-08-08
Scroll and exists query

fulmicoton:
[Added docs for scroll and exists query.](https://github.com/quickwit-oss/quickwit/pull/3724/commits/b491ad0d3a45f4d0104dfad7c7474d58d39b417e) 

Closes https://github.com/quickwit-oss/quickwit/issues/3715
Closes https://github.com/quickwit-oss/quickwit/issues/3713


docs/reference/es_compatible_api.md
```diff
@@ -158,6 +196,31 @@ The payload is expected to alternate:
...
+```
+
+#### Supported Request Body parameters
+
+| Variable      | Type       | Description                                                      | Default value |
+|---------------|------------|------------------------------------------------------------------|---------------|
+| `scroll_id`     | Scroll id (obtained from a search response)  | Required
+
+
+The `_scroll`, in combination with the `_search` API makes it possible successive pages of search results.
```
imotov, line 213: I think something is missing between `possible` and `successive`. Perhaps, it should be `possible to retrieve successive pages`? I think it also should be `_search/scroll` instead of just `_scroll`.

docs/reference/es_compatible_api.md
```diff
@@ -158,6 +196,31 @@ The payload is expected to alternate:
 - a `header` json object, containing the targetted index id.
 - a `search request body` as defined in the [`_search` endpoint section].
 
+
+### `_scroll` &nbsp; Scroll API
```
imotov, line 200:
```suggestion
### `_search/scroll` &nbsp; Scroll API
```

docs/reference/es_compatible_api.md
```diff
@@ -158,6 +196,31 @@ The payload is expected to alternate:
...
+
+#### Supported Request Body parameters
+
+| Variable      | Type       | Description                                                      | Default value |
+|---------------|------------|------------------------------------------------------------------|---------------|
+| `scroll_id`     | Scroll id (obtained from a search response)  | Required
+
+
+The `_scroll`, in combination with the `_search` API makes it possible successive pages of search results.
+First, the client needs to call the `search api` with a `scroll` query parameter, and then pass the `scroll_id` returned in the response payload to  `_scroll` endpoint.
```
imotov, line 214: I couldn't find corresponding docs on the search API. Also `_scroll` should probably be replaced with `_search/scroll`.

docs/reference/es_compatible_api.md
```diff
@@ -131,10 +132,47 @@ If a parameter appears both as a query string parameter and in the JSON payload,
...
 | `size` | `Integer` | Number of hits to return. |  10 |
-| `sort` | `JsonObject[]` | Describes how documents should be ranked. | `[]` |
+| `sort` | `JsonObject[]` | Describes how documents should be ranked. See [Sort order](#sort-order) | `[]` |
 | `aggs` | `Json object` | Aggregation definition. See [Aggregations](aggregation.md). | `{}` | `
 
 
+#### Sort order
+
+You can define up to two criteria on which to apply sort.
+If two criteria are supplied, the lexicographical order is applied.
```
imotov, line 142: This phrase is open for interpretation if you have two text fields and can cause some confusion. I would just omit it and go straight to describing how the second field is use as a tie breaker if the first field have the same value in multiple documents.

docs/reference/es_compatible_api.md
```diff
@@ -131,10 +132,47 @@ If a parameter appears both as a query string parameter and in the JSON payload,
...
+
+```json
+{ //...
+  "sort" : ["_score", "timestamp"]
+  // ...
+}
+```
+
+The default order is `asc`, except for `_score` for which it is `desc`.
+When sorting by a fast field, if the doc is associated to several values, only the first one is considered.
```
imotov, line 174:
```suggestion
When sorting by a fast field and this field contains several values in a single document, only the first value is used for sorting.
```

docs/reference/es_compatible_api.md
```diff
@@ -131,10 +132,47 @@ If a parameter appears both as a query string parameter and in the JSON payload,
...
+
+You can define up to two criteria on which to apply sort.
+If two criteria are supplied, the lexicographical order is applied.
+In other words, the second criterion will only be used in presence of a tie for the first criterion.
+
+A given criterion can either be
+- the name of a fast field (explicitly defined in the schema or captured by the dynamic mode)
+- `_score` to sort by BM25.
+
+It is also possible to define whether the different component should be taken as descending/ascending using the
```
imotov, line 149: I think I would flip the order here and put the default syntax first. And then I would say something like "The default sort order can be overwritten using the following syntax" and put this example. That will allow us to avoid saying "It is also possible" multiple times and introducing a new term "component" that could be a bit confusing for readers.

imotov (approved):
LGTM. Added a few suggestion to clarify some some aspects that I found a bit confusing.

PR #3726, opened 2023-08-08
RFC: sorting specification

trinity-1686a:
### Description

This PR tries to formalize the state of sorting/top-k in Quickwit, and propose changes both in behavior and in code organization to enable a more correct, simpler to maintain and simpler to optimize behavior.
The goal for now is to gather opinions on the state of sorting, and how it should move forward. Feel free to comment if you disagree with the proposed new behavior

### How was this PR tested?

For now, this is a text-only PR. It was prompted by tests done [here](https://github.com/quickwit-oss/quickwit/pull/3545#issuecomment-1662380517)


docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Behavior
+
+Sorting is controlled by the `sort_by` query parameter. It accepts a comma separated list of fields
+to use for sorting. Sorting is Ascending by default. The sorting order can be reversed by prefixing
+a field name with a hyphen `-`.
+The special value `_score` means sorting by score, with a sort that is then Descending by default
+(documents matching better goes first).
+
+In case of equality between two documents, the GlobalDocId, composed of (SplitId, SegmentId, DocId)
```
saroh, line 54: Hey :wave:,
This might be based on wrong assumptions, disregard if so :D
With [cursor pagination](https://betterprogramming.pub/understanding-the-offset-and-cursor-pagination-8ddc54d10d98?gi=72175bea8ea5) in mind: shouldn't the `timestamp` be somewhere in the tie breaker, espacially if it's the posting's sort order? It probably is as it seems like an optim for top-k on your prefered "timestamp" sort?

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+
+Both orders allow an optimization when sorting by date (either direction), by leveraging splits
+meta-data to know in advance if a split can, or not, contain better results. Changing the sorting
+order for "not sorted" queries allows to leverage SplitId as a way to know whether a split can
+contain or not better results (if its SplitId is more/less than the current worst best-hit, the
+split does not need to be searched).
+
+If we allow unsorted requests, we can go further and stop searching as soon as we have k hits,
+without even looking at other splits metadata. Argument can be made in favor of this because
+GlobalDocId is not stable, and can change during a merge, so order is not guaranteed anyway,
```
saroh, line 88: > GlobalDocId is not stable
:'( 
Good to know.

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+It also tries to describe optimizations that may be enabled (but are not necessarily implemente)
+by this behavior.
+
+# Current state
+(this section shall be removed once proposed changes are accepted)
+
+## Behavior
+
+Sorting is controlled by the `sort_by` query parameter. It accepts a comma separated list of fields
+to use for sorting. Sorting is Ascending by default. The sorting order can be reversed by prefixing
```
fulmicoton, line 14: Crazy idea:

Should we want descending by default?

In elasticsearch it is descending by default for _score, ascending for everything else.
For timestamp, descending is better too, so maybe descending by default is a good idea.

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+a field name with a hyphen `-`.
+The special value `_score` means sorting by score, with a sort that is then Descending by default
```
fulmicoton, line 16: ah yes you spotted that.

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+(documents matching better goes first).
+
+In case of equality between two documents, the GlobalDocId, composed of (SplitId, SegmentId, DocId)
+is used as a tie breaker. It is always used in Ascending order *inside a split*. It is used in the
```
fulmicoton, line 20: This behavior is not great. You are offering a spec where the repartition of docs in splits as an effect on search results.

Could we do ascending always?

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+same order as the first field *when merging results*
+// TODO is this really the case? That's at least what my own tests seems to show.
+
+If a document doesn't have a value for a sorting field, that document is considered to go after any
+document which has a value, independantly of sort order. That is, when sorting the value 1,2 and
+None, ascending sort would give `[1, 2, None]`, and descending sort would give `[2, 1, None]`.
```
fulmicoton, line 26: yes

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+
+If the client does not request sorting, documents are sorted using (DocId, SplitId, SegmentId), in
```
fulmicoton, line 28: `(SplitId, SegmentId, DocId)` would make way more sense no?

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+Descending order.
```
fulmicoton, line 29: I suspect you made an error.

```suggestion
If the client does not request sorting, documents are sorted using (SplitId, SegmentId, DocId), in
Descending order. In other words, everything happens as if documents were sorted by a constant value.
```

(This suggests that descending order by default is not a bad idea)

docs/internals/sorting.md
fulmicoton: Aaaaaaah I had missed this was a description of the current state!

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+to use for sorting. Sorting is Ascending by default. The sorting order can be reversed by prefixing
+a field name with a hyphen `-`.
+The special value `_score` means sorting by score, with a sort that is then Descending by default
+(documents matching better goes first).
+
+In case of equality between two documents, the GlobalDocId, composed of (SplitId, SegmentId, DocId)
+is used as a tie breaker. It is used to sort in the same order as the first field being sorted by.
+This means it is in Ascending order by default, except when the first field being sorted by is
+`_score` .
+// TODO or should it always be Ascending by default? I don't think it matters much, but would require
```
fulmicoton, line 58: I suspect descending by default would be better.

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+some more special handling of `_score`.
+
+If a document doesn't have a value for a sorting field, that document is considered to go after any
+document which has a value, independantly of sort order. That is, when sorting the value 1,2 and
```
fulmicoton, line 62:
```suggestion
document which has a value, independently of sort order. That is, when sorting the value 1,2 and
```

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+None, ascending sort would give `[1, 2, None]`, and descending sort would give `[2, 1, None]`.
+
+If a client does not request sorting, documents are sorted using (SplitId, SegmentId, DocId), on
```
fulmicoton, line 65: I think it is easier to define that behavior by saying "everything should happen as we sorted by a constant field"
in descending order. (Assuming we go for descending by default.)

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+## Behavior
+
+Sorting is controlled by the `sort_by` query parameter. It accepts a comma separated list of fields
+to use for sorting. Sorting is Ascending by default. The sorting order can be reversed by prefixing
+a field name with a hyphen `-`.
+The special value `_score` means sorting by score, with a sort that is then Descending by default
+(documents matching better goes first).
+
+In case of equality between two documents, the GlobalDocId, composed of (SplitId, SegmentId, DocId)
+is used as a tie breaker. It is used to sort in the same order as the first field being sorted by.
```
fulmicoton, line 55: ok

docs/internals/sorting.md
```diff
@@ -0,0 +1,93 @@
...
+Both orders allow an optimization when sorting by date (either direction), by leveraging splits
+meta-data to know in advance if a split can, or not, contain better results. Changing the sorting
+order for "not sorted" queries allows to leverage SplitId as a way to know whether a split can
+contain or not better results (if its SplitId is more/less than the current worst best-hit, the
+split does not need to be searched).
+
+If we allow unsorted requests, we can go further and stop searching as soon as we have k hits,
+without even looking at other splits metadata. Argument can be made in favor of this because
+GlobalDocId is not stable, and can change during a merge, so order is not guaranteed anyway,
+at least not until Quickwit has support for a Point In Time mechanism.
```
fulmicoton, line 89: Now, when paginating using the scroll API we have such a PIT mechanism.

fulmicoton:
Great job! I suspect we might prefer Descending order for everything by default but otherwise I did not have any comment.

docs/internals/sorting.md
trinity-1686a: as surprising as it seems, this is the actual order used. We use `(SplitId, SegmentId, DocId)` as a tie breaker, but sort primarily based on a [value which is the DocId](https://github.com/quickwit-oss/quickwit/blob/5d2c66e38e17ba5952196f01123d87f5a98b6a3c/quickwit/quickwit-search/src/collector.rs#L203). It might be more explicit to write that we sort by `(DocId, SplitId, SegmentId, DocId)`, which is a bit redundant, but more of what's actually happening

docs/internals/sorting.md
trinity-1686a: I've seen that, but it's not really usable from outside the scroll API, is it? I meant mostly for a "general purpose PIT" like the one ES has, and recommend using with things like search-after

docs/internals/sorting.md
trinity-1686a: I think if you want `timestamp` to be used as a tie breaker of sorts, you should put it last in the `sort_by` query parameter. Right now it's limited to two fields, but that limitation will be lifted off at some point. I don't think it would help in optimization to have `timestamp` as a tie breaker, on the contrary it means you have to always download that fastfield. However a "more stable sort order" is something that would be nice to have

fulmicoton (approved):

trinity-1686a:
I've added some test for the sort logic inside a split, as well as fixed the test for sort merging (the ascending test wasn't performing an asc sort, and the test vector were ordered in a way that made it hard to notice any issue). I also fixed a bug previously mentioned (though I otherwise kept the old sorting behaviour).
If nobody has an objection to the new ~~world~~ sort order, I think this is ready to merge, with a followup implementing the changes described here

docs/internals/sorting.md
```diff
@@ -0,0 +1,84 @@
+# Sorting
+
+Quickwit can sort results based on fastfield values or score. This document discuss where and ho
+ it happens.
+It also tries to describe optimizations that may be enabled (but are not necessarily implemente)
+by this behavior.
+
+# Current state
```
fulmicoton, line 8: I think we can remove the `Current State`

fulmicoton (approved):

PR #3727, opened 2023-08-09
Added PhraseMatchQuery and MultiMatchQuery.

### LINKED ISSUE #3725: Add support for ES multimatch query
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html

fulmicoton:
Closes #3725


trinity-1686a (commented):
I don't see [per-field boosting and wildcard fields](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html#field-boost). I don't know if we want to support them, if we don't, we should probably return an explicit error when getting one of those

fulmicoton:
> per-field boosting and wildcard fields

I added code to return an error in that case.

trinity-1686a (approved):

PR #3734, opened 2023-08-10
Add multi indices search to elasticsearch + quickwit search endpoints.

### LINKED ISSUE #2770: Target multiple indexes during a search
Elastic search allows users to target multiple indexes with the same query in one search request.

fmassot:
Fix #2770.

This is a partial implementation of the multi-index syntax of elasticsearch, which is defined [here](https://www.elastic.co/guide/en/elasticsearch/reference/8.9/api-conventions.html#api-multi-index): only comma-separated and wildcard syntax are supported in this PR.

```
In multi-target syntax, you can use a comma-separated list to run a request on multiple resources, such as data streams, indices, or aliases: test1,test2,test3. You can also use [glob-like](https://en.wikipedia.org/wiki/Glob_(programming)) wildcard (*) expressions to target resources that match a pattern: test* or *test or te*t or *test*.
```

Examples:
- `/index-*/search`
- `/index-1,index-2/search`
- `/_elastic/*/_search`

## Limitations

NB: those limitations can be lifted, but it adds complexity due to managing multiple `QueryAST`, `TagFilterAst`,  and `timestamp_field`.

My opinion on this is that the limitations are reasonable from the user's point of view.

1. Limitation on queries with different resolved query AST

Queries on multiple indexes that lead to different `QueryAst` per index are not supported.

Example:
- index 1 has two fields, `field1` and `field2`, with the default field on `field1`.
- index 2 has two fields, `field1` and `field2`, with the default field on `field2`.

A query like `test` will be resolved as:
- `field1:test` on index 1.
- `field2:test` on index 2.

2. Limitation on queries with different timestamp fields

Queries on multiple indexes with different timestamp fields per index are not supported. Note that if one index has no timestamp field, but the others have, this does not raise any issue.

Supporting different timestamp fields will require to:
- call a modified version of `refine_start_end_timestamp_from_ast` for each timestamp field of each index and extract the start/end timestamp from the query
- merge the start/end timestamp correctly: take the max of start timestamps, min of end timestamps and if None values are present ignore the refinements.

Also, it can lead to side effects in query performance on corner cases (start and end timestamps can be very narrowed on all indexes but one, the last one has no refinements on start/end timestamps, and we end up with no refinement. So adding just one index to the query will lead to a very different performance).

I implemented a version of it, but I decided not to add it to this PR. I would rather add it if users need this.

## Follow-up in next PRs

- [ ] Resolve with the default fields in the delete task planner.
- [ ] Add docs.
- [ ] Listing splits into several indexes introduced a discrepancy between our file-backed metastore and PostgreSQL metastore. In the file-backed metastore, if one index ID among all the requested index IDs does not exist, it will return an `IndexesDoNotExistError`. In the PostgreSQL metastore, we only check the existence of index IDs if no splits are returned. I suggest not checking for index existence in the `list_split` method and fixing that in a follow-up PR.

## TODO

- [x] Remove `list_all_indexes_metadatas` as it is redundant with `list_indexes_metadatas`
- [x] Clean up the code.
- [x] Add more tests.

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
+++ b/quickwit/quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
@@ -33,10 +33,11 @@ use std::time::Duration;
 use async_trait::async_trait;
 use futures::future::try_join_all;
 use quickwit_common::uri::Uri;
-use quickwit_config::{IndexConfig, SourceConfig};
+use quickwit_config::{validate_index_id_pattern, IndexConfig, SourceConfig};
 use quickwit_proto::metastore::{DeleteQuery, DeleteTask};
 use quickwit_proto::IndexUid;
 use quickwit_storage::Storage;
+use regex::Regex;
 use tokio::sync::{Mutex, OwnedMutexGuard, RwLock};
 
 use self::file_backed_index::FileBackedIndex;
@@ -46,6 +47,7 @@ use self::store_operations::{
     check_indexes_states_exist, delete_index, fetch_index, fetch_or_init_indexes_states,
     index_exists, put_index, put_indexes_states,
 };
+use super::ListIndexesQuery;
 use crate::checkpoint::IndexCheckpointDelta;
 use crate::{
     IndexMetadata, ListSplitsQuery, Metastore, MetastoreError, MetastoreResult, Split,
@@ -163,8 +165,8 @@ impl FileBackedMetastore {
         let index_id = index_uid.index_id();
         let mut locked_index = self.get_locked_index(index_id).await?;
         if locked_index.index_uid() != index_uid {
-            return Err(MetastoreError::IndexDoesNotExist {
-                index_id: index_id.to_string(),
+            return Err(MetastoreError::IndexesDoNotExist {
+                index_ids: vec![index_id.to_string()],
             });
         }
         let mut index = locked_index.clone();
@@ -212,8 +214,8 @@ impl FileBackedMetastore {
         if locked_index.index_uid() == index_uid {
             view(&locked_index)
         } else {
-            Err(MetastoreError::IndexDoesNotExist {
-                index_id: index_id.to_string(),
+            Err(MetastoreError::IndexesDoNotExist {
+                index_ids: vec![index_id.to_string()],
             })
         }
     }
@@ -377,8 +379,8 @@ impl Metastore for FileBackedMetastore {
         if !per_index_metastores_wlock.contains_key(index_id)
             && !index_exists(&*self.storage, index_id).await?
         {
-            return Err(MetastoreError::IndexDoesNotExist {
-                index_id: index_id.to_string(),
+            return Err(MetastoreError::IndexesDoNotExist {
+                index_ids: vec![index_id.to_string()],
             });
         }
 
@@ -402,7 +404,7 @@ impl Metastore for FileBackedMetastore {
             Ok(()) |
             // If the index file does not exist, we still need to return an error,
             // but it makes sense to ensure that the index state is removed.
-            Err(MetastoreError::IndexDoesNotExist { .. }) => {
+            Err(MetastoreError::IndexesDoNotExist { .. }) => {
                 per_index_metastores_wlock.remove(index_id);
                 if let Err(error) = put_indexes_states(&*self.storage, &per_index_metastores_wlock).await {
                     per_index_metastores_wlock.insert(index_id.to_string(), IndexState::Deleting);
@@ -546,9 +548,14 @@ impl Metastore for FileBackedMetastore {
     /// Read-only accessors
 
     async fn list_splits(&self, query: ListSplitsQuery) -> MetastoreResult<Vec<Split>> {
-        let query_clone = query.clone();
-        self.read(query.index_uid, |index| index.list_splits(query_clone))
-            .await
+        let mut splits = Vec::new();
+        for index_uid in query.index_uids.iter() {
+            let index_splits = self
+                .read(index_uid.clone(), |index| index.list_splits(query.clone()))
+                .await?;
+            splits.extend(index_splits);
+        }
+        Ok(splits)
     }
 
     async fn index_metadata(&self, index_id: &str) -> MetastoreResult<IndexMetadata> {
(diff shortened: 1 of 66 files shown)


fulmicoton:
@fmassot is it still draft ?

quickwit/quickwit-query/src/query_ast/mod.rs
```diff
@@ -272,15 +272,19 @@ fn parse_user_query_in_asts(
 /// # Panics
 ///
 /// Panics if the user text is invalid.
-pub fn qast_helper(user_text: &str, default_fields: &[&'static str]) -> String {
+pub fn qast_string_helper(user_text: &str, default_fields: &[&'static str]) -> String {
```
guilload, line 275: `qast_json_helper`? `to_qast_json`? `query_ast_json_from_user_text`?

quickwit/quickwit-search/src/lib.rs
```diff
@@ -120,28 +121,24 @@ fn extract_split_and_footer_offsets(split_metadata: &SplitMetadata) -> SplitIdAn
 
 /// Extract the list of relevant splits for a given search request.
 async fn list_relevant_splits(
-    // TODO: switch search request to index_uid and remove this.
-    index_uid: IndexUid,
-    search_request: &SearchRequest,
+    indexes_uids: Vec<IndexUid>,
```
guilload, line 124:
```suggestion
    index_uids: Vec<IndexUid>,
```

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -98,8 +98,8 @@ message GetKVResponse {
 // -- Search -------------------
 
 message SearchRequest {
-  // Index ID
-  string index_id = 1;
+  // Index ID patterns
+  repeated string index_id_patterns = 1;
```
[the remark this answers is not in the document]
guilload: My bad. It's ok to go from optional to repeated, but also bytes/string to repeated bytes/string.

quickwit/quickwit-serve/src/search_api/rest_handler.rs
```diff
@@ -229,31 +230,51 @@ async fn search_endpoint(
...
+        validate_index_id_pattern(index_id_pattern).map_err(|error| {
+            warp::reject::custom(crate::rest::InvalidArgument(error.to_string()))
+        })?;
+        index_ids_patterns.push(index_id_pattern.to_string());
+    }
+    if index_ids_patterns.is_empty() {
+        return Err(warp::reject::custom(crate::rest::InvalidArgument(
+            "Missing index ID.".to_string(),
+        )));
+    }
```
fulmicoton, line 274: your error handling code can never be triggered.
```suggestion
    assert!(index_ids_patterns.is_empty());
```

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -229,7 +230,16 @@ fn write_sql_filter<V: Display>(
...
-    sql.push_str(" WHERE index_uid = $1");
+    // Note: `ListSplitsQuery` builder enforces a non empty `index_uids` list.
+    sql.push_str(" WHERE (");
+    for (idx, index_uid) in query.index_uids.iter().enumerate() {
+        let _ = write!(sql, "index_uid = '{}'", index_uid);
+
+        if idx < query.index_uids.len() - 1 {
+            sql.push_str(" OR ");
+        }
+    }
```
fulmicoton, line 241:
```suggestion
    let where_clause: String =  query.index_uids
        .iter()
        .map(|index_uid| format!("index_uid={index_uid}")
        .join();
```

fulmicoton: (that requires itertools though)

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -229,7 +230,16 @@ fn write_sql_filter<V: Display>(
 }
 
 fn build_query_filter(mut sql: String, query: &ListSplitsQuery) -> String {
```
fulmicoton, line 232: what an ugly function this is...

quickwit/quickwit-metastore/src/metastore/instrumented_metastore.rs
```diff
@@ -156,9 +163,15 @@ impl Metastore for InstrumentedMetastore {
     }
 
     async fn list_splits(&self, query: ListSplitsQuery) -> MetastoreResult<Vec<Split>> {
+        let index_uids = query
+            .index_uids
+            .iter()
+            .map(|index_uid| index_uid.to_string())
+            .collect_vec()
```
fulmicoton, line 170: nitpick

```suggestion
```
Itertools has that for iterator.

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
```diff
@@ -676,6 +698,76 @@ async fn get_index_mutex(
...
 }
 
+/// Index ID patterns matcher which matches one of the given patterns with the following
+/// rules:
+/// - If the given pattern does not contain a `*` char, it matches the exact pattern.
+/// - If the given pattern contains one or more `*`, it matches the regex built from a regex where
+///   `*` is replaced by `.*`. All other regular expression meta characters are escaped.
+struct IndexIdPatternsMatcher {
+    match_all: bool,
+    index_ids: Vec<String>,
```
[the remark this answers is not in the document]
fulmicoton: +1 : this struct is an antipattern.

We can also just simply go with a `RegexSet`.

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
```diff
@@ -676,6 +698,76 @@ async fn get_index_mutex(
...
+    validate_index_id_pattern(index_pattern)?;
+    let mut re: String = String::new();
+    re.push('^');
+    for tok in index_pattern.chars() {
+        if tok == '*' {
+            re.push_str(".*");
+        } else {
+            re.push_str(&regex::escape(&tok.to_string()));
+        }
+    }
```
fulmicoton, line 765:
```suggestion
   index_pattern.split("*")
          .map(|pattern_part| regex::espace(&pattern_part))
          .join(".*");
```

quickwit/quickwit-search/src/root.rs
```diff
@@ -124,6 +130,103 @@ impl From<FetchDocsJob> for SplitIdAndFooterOffsets {
...
+        .map_err(|err| {
+            SearchError::InternalError(format!("Failed to build doc mapper. Cause: {err}"))
+        })?;
+        let query_ast_resolved_for_index = query_ast
+            .clone()
+            .parse_user_query(doc_mapper.default_search_fields())
+            // We convert the error to return a 400 to the user (and not a 500).
+            .map_err(|err| SearchError::InvalidQuery(err.to_string()))?;
+
+        // Validate uniqueness of resolved query AST.
```
fulmicoton, line 176: We had three possible specs here:
a) pick the `query_ast_resolved` of a given index.. Maybe the freshest one, or the first defined in the list.
b) use a different `query_ast_resolved` for each indexed.
c) return an error when facing an inconsistency (that's what you picked)

Obviously this is a trade-off, but I strongly prefer b).
I don't think there is a strong case to be defensive here.

fulmicoton (approved):
Approved, I think we should merge this BUT I think we should have gone with one query ast resolved per index.

This also had some wrangling on the models of the LeafRequest.
A dead simple implementation could have been something: 
phase1: keep the leaf request unchanged (as in like it is in main) and do emit several gRPC
to each node, apply the existing merging logic on the root side.
phase2: add a batch leaf_request GRPC endpoint, that computes the different leaf request and pre-merge their results before sending the result back to the root.

If you agree, can we open a ticket to revisit this?

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
guilload: https://crates.io/crates/sea-query ?

PR #3735, opened 2023-08-10
Fix broken doc link in quickwit-storage

mgattozzi:
### Description

In #3709 there was a refactor that removed `StorageConfig` from the scope of the changed file in this commit. This was fine as the function `register` in StorageResolveBuilder did not need it anymore. However, the docs alluded to registering both a factory and a config. When running `make build-docs` for a different change I ran into this causing the command to fail. This change updates the docs so that they can build properly again.

### How was this PR tested?

I ran `make build-docs` to make sure they built and checked the output.

### DIFF
--- a/quickwit/quickwit-storage/src/storage_resolver.rs
+++ b/quickwit/quickwit-storage/src/storage_resolver.rs
@@ -130,7 +130,7 @@ pub struct StorageResolverBuilder {
 }
 
 impl StorageResolverBuilder {
-    /// Registers a [`StorageFactory`] and a [`StorageConfig`].
+    /// Registers a [`StorageFactory`].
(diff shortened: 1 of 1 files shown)


guilload (approved):

guilload:
CI used to perform that check but no longer does because of recent refactor of the workflow. Can you fix that too?

mgattozzi:
@guilload do you mean this [part right here](https://github.com/quickwit-oss/quickwit/blob/main/.github/workflows/ci.yml#L140-L143)?

Would this not always run now if the source changed?

guilload:
`cargo doc` runs each time the source is changed as expected, but the CI workflow and `make build-docs` commands differ.

fulmicoton:
Merging. I don't think @mgattozzi has the permissions to do that.

PR #3744, opened 2023-08-14
Add initial test for gcp pubsub indexer

AyWa:
### Description

**PR is still in draft, few things I need to do before review**

Before adding new functionality to gcp pubsub indexer (`at_least_once` support), we want to be sure that basics functionality are working.
As suggested in https://github.com/quickwit-oss/quickwit/pull/3720#issuecomment-1675038458 this PR is adding basic integration test. 

- [x] test: client return error if subscription does not exist
- [x] test: ensure we process message and exit when backfill_mode

We are also doing small improvement: upgrade the dependency, better support for GCP auth.

### How was this PR tested?

This pr is adding integration test :)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -209,6 +209,17 @@ services:
     extra_hosts:
       - "host.docker.internal:host-gateway"
 
+  gcp-pubsub-emulator:
+    # It is not an official docker image
+    # if we prefer we can build a docker from the official docker image (gcloud cli)
+    # and install the pubsub emulator https://cloud.google.com/pubsub/docs/emulator
+    image: thekevjames/gcloud-pubsub-emulator:${GCLOUD_EMULATOR:-7555256f2c}
+    ports:
+      - "${MAP_HOST_GCLOUD_EMULATOR:-127.0.0.1}:8681:8681"
+    environment:
+      # create a fake gcp project and a topic / subscription
+      - PUBSUB_PROJECT1=quickwit-emulator,emulator_topic:emulator_subscription
+
 volumes:
   localstack_data:
   postgres_data:
--- a/quickwit/quickwit-config/src/source_config/mod.rs
+++ b/quickwit/quickwit-config/src/source_config/mod.rs
@@ -295,7 +295,7 @@ pub struct GcpPubSubSourceParams {
     pub enable_backfill_mode: bool,
     /// GCP service account credentials (`None` will use default via
     /// GOOGLE_APPLICATION_CREDENTIALS)
-    /// Path to a [`google_cloud_auth::credentials::CredentialsFile`] serialized in JSON. See also
+    /// Path to a google_cloud_auth::credentials::CredentialsFile serialized in JSON. See also
     /// `<https://github.com/yoshidan/google-cloud-rust/tree/main/pubsub#automatically>`.
     pub credentials_file: Option<String>,
     /// GCP project ID (Defaults to credentials file project ID).
--- a/quickwit/quickwit-config/src/source_config/mod.rs
+++ b/quickwit/quickwit-config/src/source_config/mod.rs
@@ -296,7 +296,9 @@ pub struct GcpPubSubSourceParams {
     /// GCP service account credentials (`None` will use default via
     /// GOOGLE_APPLICATION_CREDENTIALS)
     /// Path to a google_cloud_auth::credentials::CredentialsFile serialized in JSON. See also
-    /// `<https://github.com/yoshidan/google-cloud-rust/tree/main/pubsub#automatically>`.
+    /// `<https://cloud.google.com/docs/authentication/application-default-credentials>` and
+    /// `<https://github.com/yoshidan/google-cloud-rust/tree/main/pubsub#automatically>` and
+    /// `<https://docs.rs/google-cloud-auth/0.12.0/google_cloud_auth/credentials/struct.CredentialsFile.html>`.
     pub credentials_file: Option<String>,
     /// GCP project ID (Defaults to credentials file project ID).
     pub project_id: Option<String>,
--- a/quickwit/quickwit-config/src/source_config/mod.rs
+++ b/quickwit/quickwit-config/src/source_config/mod.rs
@@ -296,8 +296,8 @@ pub struct GcpPubSubSourceParams {
     /// GCP service account credentials (`None` will use default via
     /// GOOGLE_APPLICATION_CREDENTIALS)
     /// `<https://github.com/yoshidan/google-cloud-rust/tree/main/pubsub#automatically>`
-    pub credentials_file_path: Option<String>,
-    /// GCP project id. (`None` will use default like credentials_file_path)
+    pub credentials_file: Option<String>,
+    /// GCP project ID (Defaults to credentials file project ID).
     pub project_id: Option<String>,
     /// Maximum number of messages returned by a pull request (default 1,000)
     pub max_messages_per_pull: Option<i32>,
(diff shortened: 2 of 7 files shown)


quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -145,13 +157,15 @@ impl Source for GcpPubSubSource {
...
-
-        // TODO: lets add parallelism support in next PR
         // TODO: ensure we ACK the message after being commit: at least once
         // TODO: ensure we increase_ack_deadline for the items
         loop {
             tokio::select! {
-                _ = self.pull_message_batch(&mut batch) => {
+                resp = self.pull_message_batch(&mut batch) => {
+                    if let Err(err) = resp {
+                        warn!("failed to pull message {:?}", err);
```
guilload, line 166:
```suggestion
                        warn!("Failed to pull messages from subscription `{}`: {:?}", self.state.subscription_name, err);
```

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -241,3 +273,170 @@ impl GcpPubSubSource {
...
+    #[tokio::test]
+    async fn test_gcp_source_invalid_subscription() -> anyhow::Result<()> {
+        quickwit_common::setup_logging_for_tests();
+        let sub = append_random_suffix("test-gcp-pubsub-source--sub");
+        let source_config = get_source_config(&sub);
+
+        let index_id = append_random_suffix("test-gcp-pubsub-source--process-message--index");
+        let index_uid = IndexUid::new(&index_id);
+        let metastore = metastore_for_test();
+        let params = if let SourceParams::GcpPubSub(params) = source_config.clone().source_params {
```
guilload, line 347:
```suggestion
        let SourceParams::GcpPubSub(params) = source_config.clone().source_params else {
            panic!("Expected `SourceParams::GcpPubSub` source params, got {:?}", source_config.clone().source_params);
        }
```

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -241,3 +273,170 @@ impl GcpPubSubSource {
...
+
+        let created_topic = c.create_topic(topic, None, None).await?;
+        c.create_subscription(subscription, topic, subscription_config, None)
+            .await?;
+
+        anyhow::Ok(created_topic.new_publisher(None))
+    }
+
+    #[tokio::test]
+    async fn test_gcp_source_invalid_subscription() -> anyhow::Result<()> {
```
guilload, line 339:
```suggestion
    async fn test_gcp_source_invalid_subscription() {
```

docker-compose.yml
```diff
@@ -209,6 +209,17 @@ services:
     extra_hosts:
       - "host.docker.internal:host-gateway"
 
+  gcp-pubsub-emulator:
+    # It is not an official docker image
+    # if we prefer we can build a docker from the official docker image (gcloud cli)
+    # and install the pubsub emulator https://cloud.google.com/pubsub/docs/emulator
+    image: thekevjames/gcloud-pubsub-emulator:${GCLOUD_EMULATOR:-7555256f2c}
```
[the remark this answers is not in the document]
AyWa: [this image](https://cloud.google.com/pubsub/docs/emulator) google doesn't provide a docker image with gcloud emulator on it.
But they provide a docker image with `gcloud` so in the docker build you can just do:

```
gcloud components install pubsub-emulator
gcloud components update
# set entry point to gcloud beta emulators pubsub start --project
```

this is basically what all the "community" docker image are doing. 
My question is more, if you don't want to have a docker image that is not own by a community or a big project, it is maybe better to setup a dockerfile and push to quickwit dockerhub etc.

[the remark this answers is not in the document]
guilload: Thanks for the explanation. A community-maintained image is fine.

quickwit/quickwit-config/src/source_config/mod.rs
```diff
@@ -295,9 +295,10 @@ pub struct GcpPubSubSourceParams {
     pub enable_backfill_mode: bool,
     /// GCP service account credentials (`None` will use default via
     /// GOOGLE_APPLICATION_CREDENTIALS)
-    pub credentials: Option<String>,
-    /// Number of pull requests issued in parallel by the source (default 1)
-    pub pull_parallelism: Option<u64>,
+    /// `<https://github.com/yoshidan/google-cloud-rust/tree/main/pubsub#automatically>`
```
guilload, line 298:
```suggestion
    /// Path to a [`google_cloud_auth::credentials::CredentialsFile`] serialized in JSON. See also `<https://github.com/yoshidan/google-cloud-rust/tree/main/pubsub#automatically>`.
```

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -92,14 +92,24 @@ impl GcpPubSubSource {
...
-        // let pull_parallelism = params.pull_parallelism.unwrap_or(DEFAULT_PULL_PARALLELISM);
         let max_messages_per_pull = params
             .max_messages_per_pull
             .unwrap_or(DEFAULT_MAX_MESSAGES_PER_PULL);
-        let client_config = ClientConfig::default()
-            .with_auth()
-            .await
-            .context("Failed to authenticate GCP PubSub source.")?;
+
+        let mut client_config: ClientConfig = match params.credentials_file_path {
```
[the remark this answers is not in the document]
guilload: This suggestion remains.

AyWa:
> You're almost there, so I will approve this PR, assuming you fix the little things here and there. Worst case, we'll take care of them in the next PRs. Thanks for your great work!

yeah I am fixing the PR now and I will likely add `an assert for the checkpoint delta.` likely tomorrow. Thank you for your details review !

AyWa:
I wanted to quickly do https://github.com/quickwit-oss/quickwit/pull/3744#discussion_r1294665268 but it is actually a bit more tricky because almost everything is random for now:
* the partition_id
* the message_id
* the timestamp

I will check more tomorrow, but does there is an easy way to assert any string, any timestamp greater than X etc inside a struct or I need to write a custom function that will do that ?

Anyway I put the PR in ready, I also had to revert a suggested change https://github.com/quickwit-oss/quickwit/pull/3744/commits/c524b6d3ce3b81572868117a7a427c565301a457 because we likely don't want to important `google_auth` in this package just for the link ? or is it worth it ?

AyWa:
@guilload Also I wanted to ask, isn't better we don't add parallelism in the gcp indexer, but just adding `max_num_pipelines_per_indexer`  and `desired_num_pipelines` ?

guilload:
It depends on the throughput delivered by the source with no parallelism. We need to achieve a throughput of at least 20-30MB/s to saturate the indexing pipeline, after which we can start spawning more indexing pipelines.

AyWa:
> It depends on the throughput delivered by the source with no parallelism. We need to achieve a throughput of at least 20-30MB/s to saturate the indexing pipeline, after which we can start spawning more indexing pipelines.

Okay sounds good ! Then next PR will focus on `at_least_once`. Then we can test / benchmark. Then optimize / add parallelism if needed ? 

Also in parallel I am working with @sravan-s on a Nats implementation. https://github.com/AyWa/quickwit/pull/2
We are planing to follow similar step as what we do for gcp pubsub :)

Issue #3485, opened 2023-06-02
Observation messages are stacking up


fulmicoton:
Some of our actors can have long running tasks.
For instance, the merge actor can run for 30minutes or so, during large merges.

While it happens, the pipeline supervisor will keep stacking observe message into its high prioirty pipeline, and htey will keep timeouting.
They will then all be processed when the merge finishes.

We are talking about a few thousands messages, and a few thousand observation per index. This is not catastrophic, but we should fix this.

(There are several solutions to this problem)

PR #3765, opened 2023-08-18
Prevent `Observe` messages from stacking up

fulmicoton:
# Prevents Observe message emitted by Supervisors from stacking up

This introduce a refresh observation method, that does not wait
for anything but just triggers an eventual refresh of the last
observation.

This method includes debouncing logic to avoid stacking up messages.

We then use it in supervisors, which the other merit of making the
Supervise handler return much faster.

Before we were waiting for the observation to timeout.

Closes https://github.com/quickwit-oss/quickwit/issues/3485

### DIFF
--- a/quickwit/quickwit-actors/src/actor_context.rs
+++ b/quickwit/quickwit-actors/src/actor_context.rs
@@ -68,7 +68,7 @@ pub struct ActorContextInner<A: Actor> {
     actor_state: AtomicState,
     backpressure_micros_counter_opt: Option<IntCounter>,
     observable_state_tx: watch::Sender<A::ObservableState>,
-    // Counter that gets incremented after each observation.
+    // Boolean marking the presence of an observe message in the actor's high priority queue.
     observe_enqueued: AtomicBool,
 }
 
--- a/quickwit/quickwit-actors/src/actor_context.rs
+++ b/quickwit/quickwit-actors/src/actor_context.rs
@@ -21,6 +21,7 @@ use std::convert::Infallible;
 use std::fmt;
 use std::future::Future;
 use std::ops::Deref;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-actors/src/actor_context.rs
```diff
@@ -67,6 +68,8 @@ pub struct ActorContextInner<A: Actor> {
     actor_state: AtomicState,
     backpressure_micros_counter_opt: Option<IntCounter>,
     observable_state_tx: watch::Sender<A::ObservableState>,
+    // Counter that gets incremented after each observation.
```
fmassot, line 71: Not a counter but a boolean.

quickwit/quickwit-actors/src/actor_handle.rs
```diff
@@ -144,10 +144,35 @@ impl<A: Actor> ActorHandle<A> {
...
     /// The observation will be scheduled as a high priority message, therefore it will be executed
     /// after the current active message and the current command queue have been processed.
+    ///
+    /// This method does not do anything to avoid Observe messages from stacking up.
+    /// In supervisors, prefer using `refresh_observation`.
     pub async fn observe(&self) -> Observation<A::ObservableState> {
         self.observe_with_priority(Priority::High).await
     }
 
+    // Triggers an observation.
```
fmassot, line 154: `///` is better than `//` for docs

quickwit/quickwit-actors/src/actor_handle.rs
```diff
@@ -144,10 +144,35 @@ impl<A: Actor> ActorHandle<A> {
...
+    // It is scheduled as a high priority
+    // message, and will hence be executed as soon as possible.
+    //
+    // This method does not enqueue an Observe requests if one is already enqueue.
```
fmassot, line 158:
```suggestion
    // This method does not enqueue an Observe request if one is already enqueued.
```

fmassot (approved):

Issue #1107, opened 2022-01-25
Supporting Google PubSub as a source


fulmicoton:
Google PubSub is a little bit different from Kinesis, Kafka, Pulsar.

First there are no way to manuallly handle partitions. The only way to control "which consumer gets what" is by having several topics.
Second, there are no real notion of offsets.

Instead, one can create and resume from snaphosts, or by publish timestamps.
Snapshots might be very difficult for us to use, as it would require an interaction between the indexer (downstream) and the source (upstream).
Using the timestamp as a checkpoint should be ok.

TLDR:
We can ack message right away, store timestamps as checkpoints, and seek to timestamp when we restart the source.

AyWa:
@fulmicoton Can I have a look into it with @sravan-s ? 

I used Google pubsub quite a lot before, and also Nats. So we could try to implements pubsub then Nats if you are good with it. 
Maybe I can make a plan and share with you (especially on the way to not have data loss / ack message right away or wait before ack etc)

fmassot:
@magnalite opened the PR #3605 on this. We did not have the time to review it though.
 In this PR, at-least-once delivery is not yet there, but it seems achievable. Exactly-once delivery seems way harder.

AyWa:
@fmassot oh I see, so do you think it is better we have a look at an other provider like NATS?
or you want us to try to help on the Google pubsub effort?

PR #3767, opened 2023-08-20
[DRAFT] gcp pubsub source: at_least_once

AyWa:
### Description

Following on https://github.com/quickwit-oss/quickwit/pull/3720 and https://github.com/quickwit-oss/quickwit/pull/3744 for https://github.com/quickwit-oss/quickwit/issues/1107

Goal of this PR, is to support `at_least_once` delivery, and add test for it. 

### How was this PR tested?

PR is still draft


AyWa:
> Looking good so far. This PR opens the door for proper unit testing of the checkpointing logic.

Actually I am now looking how to test the `suggest_truncate`. Because in the current test, nothing is calling `suggest_truncate`. Because there is no `uploader` right ?  Does there is a test util to trigger `suggest_truncate` or other ?

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -219,10 +228,25 @@ impl Source for GcpPubSubSource {
...
         _ctx: &ActorContext<SourceActor>,
     ) -> anyhow::Result<()> {
-        // TODO: add ack of ids
+        if let Some(Position::Offset(offset_str)) = checkpoint.position_for_partition(&self.partition_id) {
+            let commited_until = offset_str.parse::<Ulid>()?;
+            while let Some((position, messages)) = self.in_flight_batches.front() {
+                if position > &commited_until {
+                    break
+                }
+                // TODO, we could avoid this clone if we pop and push back if fail to ack ?
```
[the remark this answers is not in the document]
guilload: Yes, indeed, the handler of the `SuggestTruncate` message only logs an error:
```rust
#[async_trait]
impl Handler<SuggestTruncate> for SourceActor {
    type Reply = ();

    async fn handle(
        &mut self,
        suggest_truncate: SuggestTruncate,
        ctx: &SourceContext,
        let SuggestTruncate(checkpoint) = suggest_truncate;
        if let Err(err) = self.source.suggest_truncate(checkpoint, ctx).await {
            // Failing to process suggest truncate does not
            // kill the source nor the indexing pipeline, but we log the error.
            error!(err=?err, "suggest-truncate-error");
        }
        Ok(())
    }
}
```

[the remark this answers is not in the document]
guilload: Then on ack failure, should we also extend the deadline? This can also fail, I guess. If we are never able to ack, we have to gc the message IDs at some point. Whatever we do, we just get a better shade of at-least-once but the code complexity increases dramatically. I suggest we ignore ack errors for now.

Does the client handle retries automatically for us?

guilload:
> > Looking good so far. This PR opens the door for proper unit testing of the checkpointing logic.
> 
> Actually I am now looking how to test the `suggest_truncate`. Because in the current test, nothing is calling `suggest_truncate`. Because there is no `uploader` right ? Does there is a test util to trigger `suggest_truncate` or other ?

You can move the ack logic outside the `suggest_truncate` method of the `Source` trait and call the method directly on the `GcpPubSubSource`. Some Kafka tests do that. Or you can send `SuggestTruncate` messages to the actor` and perform some asserts via the observed state.

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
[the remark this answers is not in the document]
AyWa: > Does the client handle retries automatically for us?

it does with the default settings. (in the ack methods I cant set the retry).
```
impl Default for RetrySetting {
    fn default() -> Self {
        Self {
            from_millis: 10,
            max_delay: Some(Duration::from_secs(1)),
            factor: 1u64,
            take: 5,
            codes: vec![Code::Unavailable, Code::Unknown, Code::Aborted],
        }
    }
}

```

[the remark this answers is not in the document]
guilload: Awesome. We're in good shape if we apply this retry policy for all the GCP API calls.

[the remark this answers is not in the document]
AyWa: I think it is by default, so we should be good. Some `fn` take an optional Retry, so we could tune it later (I guess default should be fine for now)

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -216,10 +237,29 @@ impl Source for GcpPubSubSource {
...
+                    break;
+                }
+                // if we failed to ack the message, for now we just return an error that will be log
+                // we might consider to push it back to the queue later, to be a bit more resilient
+                // but we need to implements logics to extends ack first
+                let (_, messages) = self.in_flight_batches.pop_front().unwrap();
+                self.subscription
+                    .ack(messages)
+                    .await
+                    .map_err(anyhow::Error::from)
```
guilload, line 258: I suspect we don't need this line.

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -251,32 +292,28 @@ impl GcpPubSubSource {
...
         }
-        let to_position = Position::from(format!(
-            "{}:{message_id}:{publish_timestamp_millis}",
-            self.state.num_messages_processed
-        ));
+
+        let position_id: Ulid = Ulid::new();
+        self.in_flight_batches.push_back((position_id, message_ids));
+
+        let to_position = Position::from(format!("{position_id}"));
```
guilload, line 316:
```suggestion
        let to_position = Position::from(position_id.to_string());
```

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -351,6 +392,32 @@ mod gcp_pubsub_emulator_tests {
...
+            .as_str()
+            .unwrap()
+            .to_string()
+    }
+
+    fn ack_id_nb(ack_id: &str) -> u64 {
+        ack_id.split(':').last().unwrap().parse::<u64>().unwrap()
+    }
+
+    async fn publish_sample_msg(publisher: &Publisher, start: u64, end: u64) {
```
guilload, line 406:
```suggestion
    async fn publish_sample_messages(publisher: &Publisher, start: u64, end: u64) {
```

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -416,13 +472,32 @@ mod gcp_pubsub_emulator_tests {
             .await
             .unwrap();
 
+        let partition = get_partition_id(&(*source));
```
guilload, line 475:
```suggestion
        let partition = get_partition_id(&*source);
```

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -416,13 +472,32 @@ mod gcp_pubsub_emulator_tests {
...
         let source_actor = SourceActor {
             source,
             doc_processor_mailbox: doc_processor_mailbox.clone(),
         };
-        let (_source_mailbox, source_handle) = universe.spawn_builder().spawn(source_actor);
+        let (source_mailbox, source_handle) = universe.spawn_builder().spawn(source_actor);
+        let suggest_truncate_partition = partition.clone();
+        let trigger_suggest_truncate = tokio::spawn(async move {
+            loop {
+                let to_position = Position::from(format!("{}", Ulid::new()));
```
guilload, line 485:
```suggestion
                let to_position = Position::from(Ulid::new().to_string());
```

quickwit/quickwit-indexing/src/source/mod.rs
```diff
@@ -351,6 +351,17 @@ pub async fn check_source_connectivity(source_config: &SourceConfig) -> anyhow::
                 Ok(())
             }
         }
+        #[allow(unused_variables)]
+        SourceParams::GcpPubSub(params) => {
+            #[cfg(not(feature = "gcp-pubsub"))]
+            bail!("Quickwit binary was not compiled with the `gcp_pubsub` feature.");
```
guilload, line 357:
```suggestion
            bail!("Quickwit binary was not compiled with the `gcp-pubsub` feature.");
```

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -440,11 +515,182 @@ mod gcp_pubsub_emulator_tests {
...
+        gcp_source
+            .emit_batches(&doc_processor_mailbox, &source_ctx)
+            .await
+            .unwrap();
+        assert_eq!(gcp_source.in_flight_batches.len(), 1);
+        assert_eq!(
+            gcp_source
+                .in_flight_batches
+                .iter()
+                .fold(0, |acc, e| acc + e.1.len()),
```
guilload, line 587: We rarely use the `fold` combinator in this codebase. The preferred way here would be `.map(...).sum()`.

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -440,11 +515,182 @@ mod gcp_pubsub_emulator_tests {
...
+            18
+        );
+        let ulid_after_third_batch = Ulid::new();
+
+        // ensure that nothing is ack if ulid is before
+        let to_position = Position::from(format!("{}", ulid_before_batch));
```
guilload, line 632:
```suggestion
        let to_position = Position::from(ulid_before_batch.to_string());
```

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -440,11 +515,182 @@ mod gcp_pubsub_emulator_tests {
...
+
+        // we make a void actor. We will call the real source fn manually
+        let source_actor = SourceActor {
+            source: Box::new(VoidSource {}),
+            doc_processor_mailbox: doc_processor_mailbox.clone(),
+        };
+        let (source_mailbox, _source_handle) = universe.spawn_builder().spawn(source_actor);
+        let (observable_state_tx, _observable_state_rx) = watch::channel(json!({}));
+
+        let source_ctx = SourceContext::for_test(&universe, source_mailbox, observable_state_tx);
```
guilload, line 567: We don't need to spawn a dummy source to get a source context for test. There's already `ActorContext::for_test`.

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -440,11 +515,182 @@ mod gcp_pubsub_emulator_tests {
...
             "num_bytes_processed": 54,
             "num_messages_processed": 6,
             "num_invalid_messages": 0,
             "num_consecutive_empty_batches": 4,
         });
         assert_eq!(exit_state, expected_exit_state);
     }
+
+    #[tokio::test]
+    async fn test_gcp_pubsub_source_ack() {
```
guilload, line 528: We could drop the line `self.subscription.ack(message_ids).await`, and this test would still pass. Does the PubSub API have some way to tell us that we acknowledged the messages correctly?

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
```diff
@@ -216,10 +237,29 @@ impl Source for GcpPubSubSource {
...
+        {
+            let commited_until = offset_str.parse::<Ulid>()?;
+            while let Some((position, _)) = self.in_flight_batches.front() {
+                if position > &commited_until {
+                    break;
+                }
+                // if we failed to ack the message, for now we just return an error that will be log
+                // we might consider to push it back to the queue later, to be a bit more resilient
+                // but we need to implements logics to extends ack first
+                let (_, messages) = self.in_flight_batches.pop_front().unwrap();
```
guilload, line 254:
```suggestion
                let (_, message_ids) = self.in_flight_batches.pop_front().unwrap();
```

quickwit/quickwit-indexing/src/source/gcp_pubsub_source.rs
AyWa: Yeah... sadly I think there is no way to really know if the message are ack. 
My guess is because gcp doesn't garantee you that the message will not be redeliver even when you call ack.

When I was working with go, we usually tested that with an interface mock...

## Options1:
One way, would be to set an ack deadline really small. And at the end of the test just subscribe and wait to see if we receive message. It will not be perfect but it should be "fine". Problem is we might need to wait few seconds... 

## Options2:

having some interface so we can mock / ensure that the function was call.

guilload: I'm leaning towards Option 2. We can only mock the call to `ack` for now. There are examples in the codebase of mocked traits (`Metastore`, code-generated clients) with [`mockall`](https://docs.rs/mockall/latest/mockall/).

PR #3787, opened 2023-09-01
Fix file-backed index backward compat serde test

guilload:
### Description
We forgot to compare delete tasks in `test_equality`. When I added the equality check, some backward compatibility tests failed. I set `query_ast` to `"Harry Potter"` for the tests to pass again, but that's not a valid query AST. The correct fix requires more work. I suggest we drop backward compatibility with 0.4 and 0.5 in the next release instead.


fmassot (approved):
Agreed on the drop of 0.4 and 0.5 backward compatibility. Can you open an issue on this?

PR #3789, opened 2023-09-01
Use tantivy's ExistsQuery for fast fields

### LINKED ISSUE #3721: Fast field column index based field presence.
We recently introduced field presence by indexing a special field presence field.

This however comes at a cost. That field presence field is redundant with the docvalue column index if the field is a docvalue.
In that case, we should not populate the special field presence field and instead return a DocSet built using the `DocSet` leveraging the ColumnIndex.

imotov:
### Description

Updates ExistsQuery in quickwit to use tantivy's ExistsQuery for the fast fields.

Fixes #3721

### How was this PR tested?

Modified existing test to ensure they hit both use cases.


imotov:
I openned it as a draft since it is pointing to an unmerged PR in tantivy at the moment. This should be updated before merging.

imotov:
Ready for review.

quickwit/quickwit-query/src/query_ast/field_presence.rs
```diff
@@ -80,13 +80,24 @@ impl BuildTantivyAst for FieldPresenceQuery {
...
+        if field_entry.is_fast() {
+            let full_path = if path.is_empty() {
+                field_entry.name().to_string()
+            } else {
+                format!("{}.{}", field_entry.name(), path)
+            };
+            let exists_query = tantivy::query::ExistsQuery::new_exists_query(full_path);
+            Ok(TantivyQueryAst::from(exists_query))
+        } else {
+            // fallback to the presense field
```
fulmicoton, line 93:
```suggestion
            // fallback to the presence field
```

quickwit/rest-api-tests/scenarii/es_compatibility/0016-misc-query.yaml
```diff
@@ -87,7 +87,7 @@ json:
 expected:
   hits:
     total:
-      value: 60
+      value: 59  # There are actually 60 documents where this field is not empty, but one of them has a field longer than 255 chars
```
fulmicoton, line 90: Excellent point. We need to keep this side effect in check.

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -47,6 +47,21 @@ impl<'a> QueryAstVisitor<'a> for RangeQueryFields {
...
+#[derive(Default)]
+struct ExistsQueryFields {
+    exists_query_field_names: HashSet<String>,
+}
+
+impl<'a> QueryAstVisitor<'a> for ExistsQueryFields {
+    type Err = Infallible;
+
+    fn visit_exists(&mut self, exists_query: &'a FieldPresenceQuery) -> Result<(), Infallible> {
+        self.exists_query_field_names
```
fulmicoton, line 59:
```suggestion
        /// If the field is a fast field, we will rely on the `ColumnIndex`.
        /// If the field is not a fast, we will rely on the field presence field.
        ///
        /// In both case we add the field for warmup. The warmup will be no op if the field is not 
        /// a fast field.
        self.exists_query_field_names
```

quickwit/rest-api-tests/scenarii/es_compatibility/0016-misc-query.yaml
```diff
@@ -87,7 +87,7 @@ json:
 expected:
   hits:
     total:
```
fulmicoton, line 89: Can we add an integration test where we see what happens if we target a field that is not a fast field?

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -501,7 +501,8 @@ impl DocMapper for DefaultDocMapper {
             );
             for FieldValue { field, value } in document.field_values() {
                 let field_entry = self.schema.get_field_entry(*field);
-                if !field_entry.is_indexed() {
+                if !field_entry.is_indexed() || field_entry.is_fast() {
+                    // We are using an tantivy's ExistsQuery for fast fields.
```
fulmicoton, line 505: Cool you though about this!

fulmicoton (approved):

PR #3792, opened 2023-09-02
Jaeger now accepts OTLP protocol, remove jaeger exporter stuff.

fmassot:
Tested locally.


guilload (approved):
Nice. We could update the CHANGELOG with a deprecation note.

PR #3800, opened 2023-09-04
add support for exist in query language

### LINKED ISSUE #3714: Add support for field:* in the query grammar.
... and plug it to field presence.

trinity-1686a:
### Description

fix #3714 

### How was this PR tested?

tested manually

Could probably use some tests


fulmicoton (approved):

PSeitz:
> it seems recent tantivy wants more memory for some arena. I increased a value in some place and tests now passes, but I'm suspicious of whether this other place needs changes too.
> 
> https://github.com/quickwit-oss/quickwit/blob/3d7f39aaaa56f5ea1fea52e90c75bd8385682dfc/quickwit/quickwit-indexing/src/models/indexed_split.rs#L101
> 
> @PSeitz would you know that by any chance?

Yes, `15_000_000` is the minimum now.

PR #3830, opened 2023-09-13
Create lambda infrastructure

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/3862

### How was this PR tested?

- comparison between custom Docker image and provided Amazon Linux: much lower latency and faster builds with the later, so we'll continue with that one only for now
- Deployed and running for multiple days
- Benchmarks that will serve for an upcoming blogpost


distribution/lambda/README.md
```diff
@@ -0,0 +1,58 @@
+
+# CDK template for running Quickwit on AWS Lambda
+
```
fmassot, line 3: ## Prerequisite

### Install AWS CDK Toolkit (cdk command)

```
npm install -g aws-cdk 
```

### Setup python environment

...

distribution/lambda/README.md
```diff
@@ -0,0 +1,58 @@
...
+pip install -r requirements.txt
+pip install -r requirements-dev.txt
+```
+
+
+## Deploy and run
+
+The Makefile is a usefull entrypoint to show how the Lambda deployement can used.
+```bash
+export CDK_ACCOUNT=123456789
```
fmassot, line 44: I guess it's the AWS account id.

distribution/lambda/README.md
```diff
@@ -0,0 +1,58 @@
...
+export CDK_REGION=us-east-1
+make init
```
fmassot, line 46: Maybe we need to execute `make build` first?

I got this error

```
RuntimeError: Cannot find asset at /Users/fmassot/Documents/quickwit/repos/quickwit/quickwit/target/lambda/indexer
```

rdettai: I don't really understand why but looks like you do!

distribution/lambda/cdk/stacks/services/searcher_service.py
```diff
@@ -0,0 +1,33 @@
...
+            code=aws_lambda.Code.from_asset("../../quickwit/target/lambda/searcher"),
+            runtime=aws_lambda.Runtime.PROVIDED_AL2,
+            handler="N/A",
+            environment={
+                "INDEX_BUCKET": store_bucket.bucket_name,
+                "METASTORE_BUCKET": store_bucket.bucket_name,
+                "INDEX_ID": index_id,
+            },
+            timeout=aws_cdk.Duration.seconds(30),
+            memory_size=1024,
```
fulmicoton, line 29: memory_size = cpu = network.

distribution/lambda/Makefile
```diff
@@ -0,0 +1,68 @@
...
+	
+invoke-hdfs-indexer: check-env
+	python -c 'from cdk import cli; cli.upload_hdfs_src_file()'
+	python -c 'from cdk import cli; cli.invoke_hdfs_indexer()'
+
+invoke-hdfs-searcher-severity: check-env
+	python -c 'from cdk import cli; cli.invoke_hdfs_searcher("""{"query": "severity_text:ERROR"}""")'
+
+invoke-hdfs-searcher-histogram: check-env
+	python -c 'from cdk import cli; cli.invoke_hdfs_searcher("""{ "query": "*", "max_hits": 0, "aggs": { "events": { "date_histogram": { "field": "timestamp", "fixed_interval": "1h" }, "aggs": { "log_level": { "terms": { "size": 10, "field": "severity_text", "order": { "_count": "desc" } } } } } } }""")'
```
fmassot, line 34: A fixed interval of 1h is not very relevant on this dataset. I would set it to 1d.

distribution/lambda/cdk/cli.py
```diff
@@ -0,0 +1,350 @@
...
+    )
+    uri = f"s3://{bucket_name}/{example_file}"
+    try:
+        resp = session.client("s3").head_object(Bucket=bucket_name, Key=example_file)
+        print(f"{uri} already exists ({resp['ContentLength']} bytes), skipping upload")
+        return
+    except botocore.exceptions.ClientError as e:
+        if e.response["Error"]["Code"] != "404":
+            raise e
+    print(f"upload src file to {uri}")
```
fmassot, line 140:
```suggestion
    print(f"download dataset")
```

distribution/lambda/cdk/cli.py
```diff
@@ -0,0 +1,350 @@
...
+        unzipped_resp = gzip.GzipFile(mode="rb", fileobj=response)
+        while True:
+            chunk = unzipped_resp.read(1024 * 1024)
+            if len(chunk) == 0:
+                break
+            tmp.write(chunk)
+        tmp.flush()
+        print(f"downloaded {tmp.tell()} bytes")
+        session.client("s3").upload_file(
+            Bucket=bucket_name, Filename=tmp.name, Key=example_file
```
fmassot, line 157: we upload an unzip file, can you remove the `.gz` from the file name?

distribution/lambda/Makefile
```diff
@@ -0,0 +1,68 @@
...
+
+deploy-hdfs: build check-env
+	cdk deploy -a cdk/app.py HdfsStack
+	
+invoke-hdfs-indexer: check-env
+	python -c 'from cdk import cli; cli.upload_hdfs_src_file()'
+	python -c 'from cdk import cli; cli.invoke_hdfs_indexer()'
+
+invoke-hdfs-searcher-severity: check-env
+	python -c 'from cdk import cli; cli.invoke_hdfs_searcher("""{"query": "severity_text:ERROR"}""")'
```
fmassot, line 31: Limiting max_hits to 10 would be nicer as elasticsearch default is 10

config/tutorials/hdfs-logs/index-config.yaml
```diff
@@ -20,6 +20,7 @@ doc_mapping:
     - name: severity_text
       type: text
       tokenizer: raw
+      fast: true
     - name: body
       type: text
       tokenizer: default
```
fmassot, line 26: For the lambda, we should limit the `split_num_docs_target` to 2 million otherwise, we won't benefit from parallelization.
I indexed the dataset and got a split of 14 million documents. The aggregation takes 1400ms on this split...

distribution/lambda/Makefile
```diff
@@ -0,0 +1,127 @@
...
+	python -c 'from cdk import cli; cli.upload_hdfs_src_file()'
+	python -c 'from cdk import cli; cli.invoke_hdfs_indexer()'
+
+invoke-hdfs-searcher-term: check-env
+	python -c 'from cdk import cli; cli.invoke_hdfs_searcher("""{"query": "severity_text:ERROR", "max_hits": 10}""")'
+
+invoke-hdfs-searcher-histogram: check-env
+	python -c 'from cdk import cli; cli.invoke_hdfs_searcher("""{ "query": "*", "max_hits": 0, "aggs": { "events": { "date_histogram": { "field": "timestamp", "fixed_interval": "1d" }, "aggs": { "log_level": { "terms": { "size": 10, "field": "severity_text", "order": { "_count": "desc" } } } } } } }""")'
+
+bench-index:
```
rdettai, line 85: Leaving the benchmark scripts as part of the code base. We might want to remove them to avoid clutter and improve readability.

distribution/lambda/README.md
rdettai, line 1: This README somewhat overlaps with the tutorial but goes more in depth. I would argue this redundancy is fine despite the maintenance burden.

rdettai (commented):
I left some comments regarding the parts I'm less confident about. I think overall that this is ready to be merged.

distribution/lambda/Makefile
```diff
@@ -3,7 +3,8 @@
 SHELL := bash
 .SHELLFLAGS := -eu -o pipefail -c
 
-QW_LAMBDA_VERSION?=beta
+# Update this when cutting a new release
+QW_LAMBDA_VERSION?=beta-01
```
fmassot, line 7: why not use info from the tag? `lambda-beta-01` for example?

fmassot: or maybe I don't understand how this is used.

rdettai: you mean
```
QW_LAMBDA_VERSION?=$(shell git tag --sort=committerdate | grep -E "lambda-beta-" | tail -1 | cut -b 8-14)
```
?

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -152,18 +153,34 @@ impl TypedSourceFactory for FileSourceFactory {
...
-            storage
-                .get_slice_stream(
-                    file_name,
-                    Range {
-                        start: offset,
-                        end: file_size,
-                    },
-                )
-                .await?
+            // If it's a gzip file, we can't seek to a specific offset, we need to start from the
```
fmassot, line 156: I added the handling of gzip files. I think we can't have a decent tutorial on quickwit on lambda without this feature. Putting uncompressed JSON on S3 is really a bad practice.

distribution/lambda/Makefile
fmassot: Yes, after playing a bit with cdk, I think the best is to keep in the repository only the stuff related to the build of the binaries and move everything else in a dedicated repository.
We can do this cleanup just after the release.

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -152,18 +153,35 @@ impl TypedSourceFactory for FileSourceFactory {
...
+            // beginning of the file, decompress and skip the first `offset` bytes.
+            if filepath.extension().map_or(false, |ext| ext == "gz") {
```
fulmicoton, line 158: nitpick. not even sure if that works, but would
```suggestion
            if filepath.extension() == Some("gz") {
```

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -152,18 +153,35 @@ impl TypedSourceFactory for FileSourceFactory {
...
+                let stream = storage
+                    .get_slice_stream(
+                        file_name,
+                        Range {
+                            start: 0,
+                            end: file_size,
+                        },
```
fulmicoton, line 165:
```suggestion
                        0..file_size
```

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -152,18 +153,35 @@ impl TypedSourceFactory for FileSourceFactory {
...
+                    )
+                    .await?;
+                assert!(file_size >= offset, "file size should be >= offset");
```
fulmicoton, line 168: that is not a relevant reason to panic. we should return an error here.

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -172,13 +190,48 @@ impl TypedSourceFactory for FileSourceFactory {
...
+    fn new(reader: Box<dyn AsyncRead + Send + Unpin>, num_bytes_to_skip: usize) -> Self {
+        Self {
+            reader: BufReader::new(reader),
+            num_bytes_to_skip,
+        }
+    }
+
+    async fn skip(&mut self) -> io::Result<()> {
+        // Allocate once a 64kb buffer.
+        let mut buf = [0u8; 64000];
```
fulmicoton, line 215: but this is a regression for non-gzip files too no?

fmassot: No as we will never call the `skip` method

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -152,18 +153,35 @@ impl TypedSourceFactory for FileSourceFactory {
...
+                assert!(file_size >= offset, "file size should be >= offset");
+                FileSourceReader::new(Box::new(GzipDecoder::new(BufReader::new(stream))), offset)
+            } else {
+                let stream = storage
+                    .get_slice_stream(
+                        file_name,
+                        Range {
+                            start: offset,
+                            end: file_size,
+                        },
```
fulmicoton, line 177: offset..file_size
```suggestion
offset..file_size
```

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -172,13 +190,48 @@ impl TypedSourceFactory for FileSourceFactory {
...
+
+impl FileSourceReader {
+    fn new(reader: Box<dyn AsyncRead + Send + Unpin>, num_bytes_to_skip: usize) -> Self {
+        Self {
+            reader: BufReader::new(reader),
+            num_bytes_to_skip,
+        }
+    }
+
+    async fn skip(&mut self) -> io::Result<()> {
```
fulmicoton, line 213:
```suggestion
    // This function is only called for GZIP file.
    // Because they cannot be seeked into, we have to scan them to the right initial position.
    async fn skip(&mut self) -> io::Result<()> {
```

fulmicoton (approved):
see nitpicks

PR #3832, opened 2023-09-14
Upgrade Node.js

### LINKED ISSUE #3831: Update deprecated node version
**Describe the bug**
Node 16 is deprecated and should not be used anymore. 

I propose to update to node 18 (LTS until april 2025).

N/A

**Expected behavior**
The frontend keeps working as it does currently but is built using node 18

**Configuration:**
N/A

rdettai:
### Description

Closes #3831

### How was this PR tested?

- [x] Docker build
- [x] Local run
- [x] CI run

### DIFF
--- a/.github/workflows/ui-ci.yml
+++ b/.github/workflows/ui-ci.yml
@@ -5,21 +5,22 @@ on:
   pull_request:
     paths:
       - "quickwit/quickwit-ui/**"
+      - ".github/workflows/ui-ci.yml"
   push:
     branches:
       - main
       - trigger-ci-workflow
     paths:
       - "quickwit/quickwit-ui/**"
+      - ".github/workflows/ui-ci.yml"
 
 jobs:
   tests:
-    name: ${{ matrix.task.name }} (${{ matrix.os }})
-    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.task.name }}
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
         task:
           - name: Cypress run
             command: |
@@ -32,7 +33,6 @@ jobs:
             command: yarn --cwd quickwit-ui lint
           - name: Unit Test
             command: yarn --cwd quickwit-ui test
-    container: public.ecr.aws/l6o9a3f9/quickwit-builder:latest
     services:
       # PostgreSQL service container
       postgres:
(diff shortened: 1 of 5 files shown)


rdettai:
The environment for the UI CI seem to come from a [mysterious image](https://github.com/quickwit-oss/quickwit/blob/551b51130668ceb751d6b07131abb8b34f246df1/.github/workflows/ui-ci.yml#L35C40-L35C56). I couldn't find any CI that could have generated it, and it probably uses node16. Any idea how to upgrade it?

rdettai:
when running locally, I get some ugly warning:
```
WARNING in ./node_modules/autolinker/dist/es2015/version.js
Module Warning (from ./node_modules/source-map-loader/dist/cjs.js):
Failed to parse source map from '/home/ubuntu/workspace/quickwit/quickwit/quickwit-ui/node_modules/autolinker/src/version.ts' file: Error: ENOENT: no such file or directory, open '/home/ubuntu/workspace/quickwit/quickwit/quickwit-ui/node_modules/autolinker/src/version.ts'
```

It seems those were already there before as I also get there when I switch my local setup back to node 16.

rdettai:
any good idea on how to test the CI part of the change? The `cross-build-binary` action also publishes the binaries...

fmassot:
> The environment for the UI CI seem to come from a [mysterious image](https://github.com/quickwit-oss/quickwit/blob/551b51130668ceb751d6b07131abb8b34f246df1/.github/workflows/ui-ci.yml#L35C40-L35C56). I couldn't find any CI that could have generated it, and it probably uses node16. Any idea how to upgrade it?

I think we can ditch it and use a docker image with node 18 already in it.

Dockerfile
```diff
@@ -1,4 +1,14 @@
-FROM rust:bullseye AS builder
+FROM node:18 as ui-builder
```
fmassot, line 1: Nice. Can you do the same thing in the docker image?  
see https://github.com/quickwit-oss/quickwit/blob/main/Dockerfile#L25-L33

.github/workflows/ui-ci.yml
```diff
@@ -57,17 +57,24 @@ jobs:
...
-      - uses: c-hive/gha-yarn-cache@v2
+      - uses: actions/setup-node@v3
         with:
-          directory: ./quickwit/quickwit-ui
+          node-version: 18
+          cache: "yarn"
+          cache-dependency-path: quickwit/quickwit-ui/yarn.lock
+      - name: Install rust
+        run: curl https://sh.rustup.rs -sSf | sh -s -- -y
+      - uses: arduino/setup-protoc@v2
```
fmassot, line 67: why do we need that?

.github/workflows/ui-ci.yml
```diff
@@ -57,17 +57,24 @@ jobs:
...
       - name: Install JS dependencies
         run: yarn --cwd quickwit-ui install
         working-directory: ./quickwit
-      - name: Setup cache
+      - name: Setup Rust cache
         if: matrix.task.name == 'Cypress run'
         uses: Swatinem/rust-cache@v2
         with:
-          key: "v1-${{ matrix.task.name }}"
+          key: "default"
```
fmassot, line 77: I guess we could remove the `key`

Dockerfile
fmassot, line 1: nice!

fmassot (commented):
I tested the UI on the docker image, and the aarch64 build and did not find any issue.

Can you just look at the two comments?

.github/workflows/ui-ci.yml
rdettai: I had a failure because protoc was not installed. Let me try again.

fmassot: ah yes... we need that for building quickwit :/
the "cypress run" name made me forget that we are doing a `cargo build` and we need protoc...

rdettai: now I'm wondering how the release build works, I don't see any install of `protoc` anywhere

rdettai: ah through [`cross`](https://github.com/cross-rs/cross)! didn't know that tool!

fmassot (approved):

PR #3834, opened 2023-09-14
Replace action-rs/toolchain with manual config

### LINKED ISSUE #3833: Get rid of actions-rs/toolchain@v1
The github action `actions-rs/toolchain` hasn't been maintained for years now and is using features like `set-ouput` that are deprecated and might stop working anytime.

Replace with something like:
```
steps:
  - name: Install rust
    run: |
      curl https://sh.rustup.rs -sSf | sh -s -- -y
      rustup default stable-${{ inputs.target }}
      rustup component add rustfmt
```

We could also use some alternative github actions like https://github.com/dtolnay/rust-toolchain

**Additional context**
Add any other context or information about the feature request here.

rdettai:
### Description

Closes #3833

### How was this PR tested?

- [x] Run ci.yaml
- [ ] Run publish_release_packages.yml

### DIFF
--- a/.github/actions/cargo-build-macos-binary/action.yml
+++ b/.github/actions/cargo-build-macos-binary/action.yml
@@ -25,11 +25,12 @@ runs:
     - name: Install protoc
       run: brew install protobuf
       shell: bash
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-        targets: ${{ inputs.target }}
-        override: true
+    - name: Install rustup
+      shell: bash
+      run: |
+        curl https://sh.rustup.rs -sSf | sh -s -- -y
+        rustup target add ${{ inputs.target }}
+        rustup override set stable-${{ inputs.target }}
     - name: Add target aarch64-apple-darwin
       if: "${{ inputs.target == 'aarch64-apple-darwin' }}"
       run: rustup target add aarch64-apple-darwin
--- a/.github/actions/cargo-build-macos-binary/action.yml
+++ b/.github/actions/cargo-build-macos-binary/action.yml
@@ -27,10 +27,10 @@ runs:
       shell: bash
     - name: Install rustup
       shell: bash
-      run: |
-        curl https://sh.rustup.rs -sSf | sh -s -- -y
-        rustup target add ${{ inputs.target }}
-        rustup override set stable-${{ inputs.target }}
+      run: curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain none -y
+    - name: Configure rustup
+      shell: bash
+      run: rustup override set stable-${{ inputs.target }}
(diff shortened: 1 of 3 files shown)


fmassot:
@rdettai why not using https://github.com/dtolnay/rust-toolchain ?

fmassot:
I'm okay with both. I don't have strong arguments on this.
@guilload, any preference?

guilload:
`dtolnay/rust-toolchain` for conciseness.

rdettai:
@guilload I took a deeper look into `dtolnay/rust-toolchain`. It's actually pretty confusing in many ways:
- it's quite unclear what the library's behavior is with `rust-toolchain.toml`:
  - Looks from the [issues](https://github.com/dtolnay/rust-toolchain/issues/29) that it is discarded
  - yet from what I see in the sources it's really just setting `rustup default xxx`
- the [way the action versions are mapped to toolchain versions](https://github.com/dtolnay/rust-toolchain#example-workflow) is very strange. They shouldn't be correlated, actions versions should be used to indicate breaking changes in the actions behavior itself.

If my understanding is right and the action doesn't support the override option, it means that:
```
    - name: Install rustup
      run: |
        curl https://sh.rustup.rs -sSf | sh -s -- -y
        rustup target add ${{ inputs.target }}
        rustup override set stable-${{ inputs.target }}
```

is written as: 
```
    - name: Install rustup
      uses: dtolnay/rust-toolchain@v1
      with:
        toolchain: stable
        targets: ${{ inputs.target }}
    - name: Set rustup override
      run: rustup override set ${{steps.toolchain.outputs.name}}
```
That's actually even longer as the explicit definition 😅 

With these observations, do you still believe we should use this action here?

guilload:
rustup is fine too.

.github/actions/cargo-build-macos-binary/action.yml
```diff
@@ -25,11 +25,12 @@ runs:
     - name: Install protoc
       run: brew install protobuf
       shell: bash
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-        targets: ${{ inputs.target }}
-        override: true
```
rdettai, line 32: @guilload now that I think about it, why were we even overriding here in the first place? Doesn't it make more sense to build the releases with the toolchain versions pinned in `rust-toolchain.toml`?

guilload: I don't know :) There was a time when we did not have a `rust-toolchain.toml` file. It should be the source truth indeed.

rdettai:
@fmassot should we do the same tagging trick as for https://github.com/quickwit-oss/quickwit/pull/3832 to test the release workflows?

fmassot (approved):

Issue #1990, opened 2022-09-21
Warn when using filesystem?


PSeitz:
We could print a warning with a link to the docs outlining the limitations when using the filesystem as datadir or meta-store. It may ease onboarding new users.

guilload:
datadir on local fs is fine and is actually the recommended way to deploy Quickwit (SSDs).

What is not fine (in cluster mode), or at least risky if don't know what you're doing:
- file-backed metastore (S3 is ok in read-only mode)
- file-backed indexes
- default index root URI on local file system

I've recently added some warning but adding a warning for the default index root URI on local file system and linking to the docs would be nice.

JerryKwan:
@guilload 
Maybe we can add the check and warning logic after the call of load_node_config_with_env()
https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-config/src/node_config/mod.rs#L281

JerryKwan:
@guilload 
Or add the check and warning logic around https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-config/src/node_config/serialize.rs#L257

JerryKwan:
Or add the logic in the validate function https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-config/src/node_config/serialize.rs#L300

JerryKwan:
And should add warning when validating the index config at https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-config/src/index_config/serialize.rs#L88

JerryKwan:
@guilload @PSeitz 
Please help to review the pull request https://github.com/quickwit-oss/quickwit/pull/3839 which added some warning about file-backed meta-store and index

PR #3839, opened 2023-09-15
add warning about file-backed storage

JerryKwan:
add waning when using file-backend metastore
add waning when using file-backend index

### Description

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-config/src/index_config/serialize.rs
+++ b/quickwit/quickwit-config/src/index_config/serialize.rs
@@ -20,7 +20,7 @@
 use anyhow::Context;
 use quickwit_common::uri::Uri;
 use serde::{Deserialize, Serialize};
-use tracing::info;
+use tracing::{info, warn};
 
 use crate::{
     build_doc_mapper, validate_identifier, ConfigFormat, DocMapping, IndexConfig, IndexingSettings,
@@ -87,6 +87,9 @@ impl IndexConfigForSerialization {
 
         let index_uri = self.index_uri_or_fallback_to_default(default_index_root_uri)?;
 
+        if index_uri.protocol().is_file() {
+            warn!("index_uri = {}. Using file-backed index comes with certain limitations, please refer to the specific documentation for details.", index_uri);
+        }
         if let Some(retention_policy) = &self.retention_policy {
             retention_policy.validate()?;
 
--- a/quickwit/quickwit-config/src/node_config/serialize.rs
+++ b/quickwit/quickwit-config/src/node_config/serialize.rs
@@ -308,6 +308,11 @@ fn validate(node_config: &NodeConfig) -> anyhow::Result<()> {
     if node_config.peer_seeds.is_empty() {
         warn!("Peer seed list is empty.");
     }
+
+    if node_config.metastore_uri.protocol().is_file() {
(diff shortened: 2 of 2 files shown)


quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -308,6 +308,15 @@ fn validate(node_config: &NodeConfig) -> anyhow::Result<()> {
     if node_config.peer_seeds.is_empty() {
         warn!("Peer seed list is empty.");
     }
+
+    if node_config.metastore_uri.protocol().is_file() {
+        warn!(
+            "metastore_uri = {}. Using file-backed metastore comes with certain limitations, \
+             please refer to the specific documentation for details.",
+            node_config.metastore_uri
+        );
```
fulmicoton, line 317:
```suggestion
        warn!(
            metastore_uri = %node_config.metastore_uri, "Using file-backed metastore comes with certain limitations, please refer to the specific documentation for details."
        );
```

fulmicoton: (also I am not sure uri implements display `Display` but you get the gist.)

quickwit/quickwit-config/src/index_config/serialize.rs
```diff
@@ -87,6 +87,13 @@ impl IndexConfigForSerialization {
 
         let index_uri = self.index_uri_or_fallback_to_default(default_index_root_uri)?;
 
+        if index_uri.protocol().is_file() {
+            warn!(
+                "index_uri = {}. Using file-backed index comes with certain limitations, please \
```
fulmicoton, line 92: Same as below

quickwit/quickwit-config/src/node_config/serialize.rs
fulmicoton: The comment is not super helpful. We should have a link to the doc or something for people to actually know what the limitations are.

JerryKwan: @fulmicoton 
Thanks for reviewing
I mean to give the link as https://quickwit.io/docs/configuration/metastore-config
but I am not sure if the link will be persistent

quickwit/quickwit-config/src/index_config/serialize.rs
JerryKwan: How about we give the link as https://quickwit.io/docs/configuration/index-config ?

JerryKwan:
@fulmicoton 
I have updated the warning content, would you please help to review again? thank you

fmassot:
I think this does not fix the initial issue #1990 for 2 reasons:
- the pointed docs are not clear on the limitation
- the current warning is shown only if the metastore is file backed. But we say nothing about the the index storage (index_uri).

This issue is harder than it appears to do it right. We would need to first describe correctly the limitations in the docs and then see if it's worth to add warnings.

JerryKwan:
@fmassot Thanks for reviewing
Currently it outputs warnings when using filebacked index
(https://github.com/quickwit-oss/quickwit/pull/3839/commits/cf377dbe1bf7562de049797c5e0389847f372dde#diff-248b478cda2addc18cdbe2c474a75a2cba1ea230c7679bd771778e4b940cca45R92)
Regarding to the docs pointed, did you mean we should use something like this https://quickwit.io/docs/configuration/index-config#index-uri?

PR #3849, opened 2023-09-19
optimize sort by date and docid

trinity-1686a:
### Description

fix https://github.com/quickwit-oss/quickwit/issues/3298

supersedes #3545. Recent changes in sorting and how the search semaphore is acquired made that PR very hard to rebase. Porting the changes manually to a new branch seemed easier

### DIFF
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -784,6 +784,7 @@ impl PartialOrd for PartialHitSortingKey {
     }
 }
 
+#[derive(Clone)]
 struct HitSortingMapper {
     order1: SortOrder,
     order2: SortOrder,
@@ -803,6 +804,7 @@ impl SortKeyMapper<PartialHit> for HitSortingMapper {
 }
 
 /// Incrementally merge segment results.
+#[derive(Clone)]
 pub(crate) struct IncrementalCollector {
     inner: QuickwitCollector,
     top_k_hits: TopK<PartialHit, PartialHitSortingKey, HitSortingMapper>,
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -23,7 +23,9 @@ use std::collections::HashSet;
 use itertools::Itertools;
 use quickwit_common::binary_heap::{SortKeyMapper, TopK};
 use quickwit_doc_mapper::{DocMapper, WarmupInfo};
-use quickwit_proto::search::{LeafSearchResponse, PartialHit, SearchRequest, SortOrder, SortValue};
+use quickwit_proto::search::{
+    LeafSearchResponse, PartialHit, SearchRequest, SortOrder, SortValue, SplitSearchError,
+};
 use serde::Deserialize;
 use tantivy::aggregation::agg_req::{get_fast_field_names, Aggregations};
 use tantivy::aggregation::intermediate_agg_result::IntermediateAggregationResults;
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -446,88 +539,151 @@ pub async fn leaf_search(
     searcher_context: Arc<SearcherContext>,
     request: Arc<SearchRequest>,
     index_storage: Arc<dyn Storage>,
-    splits: &[SplitIdAndFooterOffsets],
+    splits: &mut [SplitIdAndFooterOffsets],
```
fulmicoton, line 542: I understand that you need it to be mut to sort it.
Let's pass ownership. The `&mut []` makes the API confusing.


```suggestion
    mut splits: Vec<SplitIdAndFooterOffsets>,
```

quickwit/quickwit-search/src/collector.rs
```diff
@@ -548,46 +550,27 @@ fn map_error(err: postcard::Error) -> TantivyError {
...
-    mut leaf_responses: Vec<LeafSearchResponse>,
-    sort_order1: SortOrder,
-    sort_order2: SortOrder,
-    max_hits: usize,
-) -> tantivy::Result<LeafSearchResponse> {
-    // Optimization: No merging needed if there is only one result.
-    if leaf_responses.len() == 1 {
-        return Ok(leaf_responses.pop().unwrap());
-    }
+    intermediate_aggregation_results: impl Iterator<Item = &'a [u8]>,
```
fulmicoton, line 555:
```suggestion
    intermediate_aggregation_results: &dyn Iterator<Item = &'a [u8]>,
```

I think we don't need the generics here, do we?

quickwit/quickwit-search/src/collector.rs
```diff
@@ -548,46 +550,27 @@ fn map_error(err: postcard::Error) -> TantivyError {
...
-fn merge_leaf_responses(
+fn merge_intermediate_aggregation_result<'a>(
     aggregations_opt: &Option<QuickwitAggregations>,
-    mut leaf_responses: Vec<LeafSearchResponse>,
-    sort_order1: SortOrder,
-    sort_order2: SortOrder,
-    max_hits: usize,
-) -> tantivy::Result<LeafSearchResponse> {
-    // Optimization: No merging needed if there is only one result.
-    if leaf_responses.len() == 1 {
```
fulmicoton, line 559: It's a bummer we lose this optimization.

trinity-1686a: We don't lose it, it's at the start of `fn merge_leaf_responses()`, which call this function. In a previous iteration it was called from 2 places, but right now it's only called from one, so it could re-inlined there actually

quickwit/quickwit-search/src/collector.rs
trinity-1686a: it's only monomorphized once for `FilterMap<std::slice::Iter, closure@>`, we can use dynamic dispatch, but I don't think there is an advantage to it in this case

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -435,6 +435,99 @@ pub(crate) fn rewrite_start_end_time_bounds(
     }
 }
 
+#[derive(Debug, Clone)]
+pub enum CanSplitDoBetter {
```
fulmicoton, line 439:
```suggestion
enum CanSplitDoBetter {
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -446,88 +539,151 @@ pub async fn leaf_search(
...
+    while let Some(sub_task) = futures.next().await {
+        match sub_task {
+            Either::Left(None) => {
+                // no more splits to startup
+            }
+            Either::Left(Some((split, permit, splits_iter))) => {
+                // we got a split to run, but first setup so the split iterator is pulled for more
+                // splits to run after.
+                futures.push(Either::Left(
+                    split_iter_get_one(
```
fulmicoton, line 607: I don't understand why this is not a deadlock.

fulmicoton (approved):

quickwit/quickwit-search/src/collector.rs
fulmicoton: It gets two implementations. It is also called in `finalize`

quickwit/quickwit-search/src/collector.rs
fulmicoton: Ah awesome!

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -446,88 +542,134 @@ pub async fn leaf_search(
...
-    merged_search_response
-        .failed_splits
-        .extend(errors.into_iter().map(|(split_id, err)| SplitSearchError {
-            split_id,
-            error: format!("{err}"),
-            retryable_error: true,
-        }));
-    Ok(merged_search_response)
+#[allow(clippy::too_many_arguments)]
+fn launch_leaf_single_split(
```
fulmicoton, line 626: nitpick:

`launch_leaf_single_split` -> `spawn_leaf_single_split`

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -446,88 +542,134 @@ pub async fn leaf_search(
...
+    request: SearchRequest,
+    searcher_context: Arc<SearcherContext>,
+    index_storage: Arc<dyn Storage>,
+    doc_mapper: Arc<dyn DocMapper>,
+    split: SplitIdAndFooterOffsets,
+    split_filter: Arc<Mutex<CanSplitDoBetter>>,
+    incremental_merge_collector: Arc<Mutex<IncrementalCollector>>,
+    leaf_split_search_permit: tokio::sync::OwnedSemaphorePermit,
+) -> tokio::task::JoinHandle<()> {
+    tokio::spawn(
```
fulmicoton, line 636: could we have moved the `tokio::spawn` to the caller side?
(and made this function `async fn` with the body in the async move block here.

fulmicoton: (or were you fighting with something tricky?

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -446,88 +542,134 @@ pub async fn leaf_search(
...
+                    split_id: split.split_id.clone(),
+                    error: format!("{err}"),
+                    retryable_error: true,
+                }),
+            }
+            if let Some(last_hit) = locked_incremental_merge_collector.peek_worst_hit() {
+                split_filter.lock().unwrap().record_new_worst_hit(last_hit);
+            }
+            // We explicitly drop it, to highlight it to the reader and to force the move.
+            std::mem::drop(leaf_split_search_permit);
```
fulmicoton, line 669: Maybe this is playing with fire, but couldn't we drop that lock before appending to the merge collector?

fulmicoton (approved):

quickwit/quickwit-search/src/collector.rs
trinity-1686a: ho yes, mb

Issue #3847, opened 2023-09-19
splits are not in a deletable state


dojiong:
I found some errors in quickwit-indexer recently:
```
message: quickwit_actors::spawn_builder: actor-failure cause=Failed to publish splits.

Caused by:
Splits `["01HAN8GBZ3F0NQFE7GBKWAC4R6"]` are not in a deletable state. exit_status=Failure(Failed to publish splits.

Caused by:
Splits `["01HAN8GBZ3F0NQFE7GBKWAC4R6"]` are not in a deletable state.)
```

After some days of logging (more logs on merge pipeline's splits tracing), I'm almost certain it's caused by:
1. merge-pipeline:publisher finished its works, passed some splits to planner's mailbox
2. merge-pipeline crashed with some reason (upload-to-object-store-fail in my scene)
3. merge-pipeline respawn the whole pipeline after a period
4. merge-pipeline:planner created, pull all the splits from metastore (*includes the splits from step-1*)
5. merge-pipeline:planner planned to merge splits from step-4, passed it to downstream
6. merge-pipeline:planner received the NewSplits message from step-1, then planned to merge

step-5/step-6 started two task contains the same splits(not all the same)

------
Temporary fix: clear the planner's inbox before respawn merge-pipeline

Futher question: how to avoid similar message problems in actor model

quickwit version: v0.6.2

dojiong:
@fulmicoton Is there a better plan to fix these problems (dead actor's messages)?

fulmicoton:
@dojiong First congratulations on understanding our indexing/merge pipeline in depth.

This bug is most likely a duplicate of #3627. Are you working with @geek-frio by any chance?

While investigating with the logs I had access to, I came down to the same conclusion as you, and then ruled out that possible cause due to some guard in the code, and due to the absence of merge pipeline error in the log... but maybe I overlook something.

The temporary fix you suggest is definitely not a good idea.
Let me reprioritize this. This bug is important.

dojiong:
I read that issue, it was also caused by duplicate plans.

According to my dig on the merge-planner's code, the most possible reason is the NewSplits message and merge-planner-init contains the same split (I also checked related actor's lifetime, it's not the reason in my scene).

This is my temprary fix: @geek-frio
```diff
diff --git a/quickwit/quickwit-actors/src/mailbox.rs b/quickwit/quickwit-actors/src/mailbox.rs
index e1571682..e23fd94d 100644
[...]

PR #3854, opened 2023-09-20
fix: duplicate splits planned on pipeline crash-respawn

dojiong:
### Description

MergePlanner will plan the same splits to downstream when MergePipeline crash-then-respawn.

The reason is that MergePlanner fetched all the splits from metastore, these splits maybe duplicated with the NewSplits message in mailbox.

*This fix is unfinished*


fulmicoton:
Attempts to fix #3847

fulmicoton:
Awesome. If I understand correctly this fix should address most of our problem... Except there is still the (rare) possibility that we had a PlanMerge message in our mailbox.

dojiong:
> Except there is still the (rare) possibility that we had a PlanMerge message in our mailbox.

Right! I dont have a better solution, so I changed it to draft.

one not good but works: add a field to PlanMerge to indicate a bootstrap PlanMerge, merge-planner reject all the PlanMerge before a bootstrap PlanMerge received

fmassot:
Indeed, it should help.

In the current implementation, we cannot have duplicated splits in the merge planner thanks to [this condition](https://github.com/quickwit-oss/quickwit/blob/8baf1ec702bd60a2dd7edad4925d173691c03419/quickwit/quickwit-indexing/src/actors/merge_planner.rs#L160-L167).

But... if we plan merges, we send splits in the pipeline, and they are removed from the merge planner.... so the above condition is insufficient to prevent duplicated splits in the whole merge pipeline.

We could enhance the condition by checking that the new split is not the ongoing merge operations tracked in `ongoing_merge_operations_inventory`.

fulmicoton (approved):

dojiong:
I think this should cover all the scenarios, I'll try on my production cluster.

fulmicoton:
I think the scenario cannot really happen today due to a technicality: the message queue of the merge planner have a capacity of 1 :).

fulmicoton:
If there are no objections I'll merge this!

fulmicoton (approved):

dojiong:
> the message queue of the merge planner have a capacity of 1 :).
The merge-publisher may await on sending NewSplits to merge-planner, so the capacity of 1 or 100 is not matter.

Is this possible?
1. merge-publisher send NewSplits to merge-planner, used the capacity of merge-planner's mailbox
2. merge-publisher send NewSplits to merge-planner, no capacity available, current future wait mailbox available
3. merge-pipeline crashed
4. merge-pipeline respawn, fetch all the splits from metastore
5. new merge-planner send PlanMerge to self, will fail (mailbox is full)
6. merge-planner received NewSplits from step-1, make a new plan includes splits from step-4
7. merge-planner received NewSplits from step-2, make a new plan includes splits from step-1/2, **splits duplicated**

I got a new `deletable error` after this patch @fulmicoton

PR #3855, opened 2023-09-20
Refactoring of the quickwit services initialization.

fulmicoton:
Refactoring of the quickwit services initialization.

The big NON trivial change, is that this removes the forward of events to the control plane through the EventBroker as it seemed redundant with the control plane proxy.

The second non trivial change, is the uncoupling of the ingest service and the indexing service. This is made possible using the universe actor registry.


fulmicoton:
Reverting to draft as it seems like this is actually not working (most likely the event broker stuff).

guilload (approved):

fulmicoton:
moved to #3874

Issue #3868, opened 2023-09-22
Support Object Storage in the file source


rdettai:
**Is your feature request related to a problem? Please describe.**
Currently, the file source only supports the local filesystem. I would be very convenient if we could reference files in any blob storage (e.g ` s3://mybucket/myfile`).
- enable the Lambda deployments to interact with data without having to first download it
- make experimentation with Quickwit easier, instead of having to have the data on the node running Quickwit users could load data they have on their buckets
- simplify the [S3 tutorial](https://quickwit.io/docs/get-started/tutorials/tutorial-hdfs-logs-distributed-search-aws-s3#index-logs), instead of having to download the file we could directly have `./quickwit index ingest --index hdfs-logs --input-path s3://quickwit-datasets-public/hdfs-logs-multitenants.json.gz` 

**Describe the solution you'd like**
The file source would automatically handle any paths like ` s3://`  and `gs://` thanks to the `Storage` abstraction.

**Describe alternatives you've considered**
Currently the workaround is to first download the file to the local storage.

**Additional context**
This is related yet mostly orthogonal to the discussions regarding making the file source more capable in terms of checkpointing (https://github.com/quickwit-oss/quickwit/issues/1159)

PR #3869, opened 2023-09-22
Read from Storage in file source

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/3868

### How was this PR tested?

Running full test suite: 

[Code coverage](https://github.com/quickwit-oss/quickwit/actions/workflows/coverage.yml?query=branch%3Aobject-source)


guilload:
I read the comment for the `resolve_streamable` method but don't understand it. Why can't the `Storage` trait have a `get_stream` method directly?

rdettai:
> I read the comment for the resolve_streamable method but don't understand it.

Noted, I'll rephrase it if we move forward in this direction

> Why can't the Storage trait have a get_stream method directly?

That's one of the questions (I actually made a typo in my previous comment, sorry!). 
**Pros:** The idea behind composing traits instead of having one big `Storage` trait that does everything is to avoid leaving methods unimplemented. I believe the `Storage` trait should already be split into a `ReadableStorage` and a `WriteableStorage`, thus avoiding unimplemented methods in `BundledStorage` and `StorageWIthCache`. I didn't want to add yet another method that doesn't make sense in all cases. You will usually want to avoid `get_stream` when the actual data is already backed by a byte array (like the cached storage) as it implies an extra unnecessary copy (I know that's what the `RamStorage` does, but that's mostly for tests). It's implementation might also be a bit tricky on the `BundledStorage`, which is a shame because it will actually never be used 😅 .
**Cons:** the machinery of the supertrait is bit obscure (especially the upcasting but also some generics introduced by it), and most `Storage` implementations are streamable anyway.

fulmicoton:
We can add a `get_stream` method implementation that has a default impl that calls `get_all` and avoid any `unimplemented!()`.

rdettai:
Ok, thanks for your insight! I'll go with a `get_stream` method directly on the `Storage` abstraction then.

quickwit/quickwit-storage/src/local_file_storage.rs
```diff
@@ -98,28 +98,28 @@ impl LocalFileStorage {
...
-            }
-        }
-    }
-    Ok(())
-}
+// /// Ensure that the path given does not include any ".." for security reasons.
+// ///
+// /// In order to reduce the attack surface, we want to make sure the `FileStorage`
+// /// only access/delete files that are children of its root_directory.
+// fn ensure_valid_relative_path(path: &Path) -> StorageResult<()> {
```
fulmicoton, line 105: same question as above?

quickwit/quickwit-storage/src/storage.rs
```diff
@@ -103,6 +103,13 @@ pub trait Storage: fmt::Debug + Send + Sync + 'static {
     /// Downloads a slice of a file from the storage, and returns an in memory buffer
     async fn get_slice(&self, path: &Path, range: Range<usize>) -> StorageResult<OwnedBytes>;
 
+    /// Open a stream handle on the file from the storage
+    async fn get_slice_stream(
```
fulmicoton, line 107: I'm surprised you wanted to stream slices and not the entire object. Is this correct?

quickwit/quickwit-storage/src/local_file_storage.rs
```diff
@@ -58,7 +58,7 @@ impl fmt::Debug for LocalFileStorage {
 
 impl LocalFileStorage {
     fn full_path(&self, relative_path: &Path) -> crate::StorageResult<PathBuf> {
-        ensure_valid_relative_path(relative_path)?;
+        // ensure_valid_relative_path(relative_path)?;
```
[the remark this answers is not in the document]
rdettai: I commented it out temporarifly, I still need to find a way around it: for the file source, we don't want this restriction the file must be a subdirectory of the storage's root.

quickwit/quickwit-storage/src/local_file_storage.rs
rdettai: let's have only one discussion thread for this: https://github.com/quickwit-oss/quickwit/pull/3869#discussion_r1338768504

quickwit/quickwit-storage/src/local_file_storage.rs
```diff
@@ -231,6 +231,18 @@ impl Storage for LocalFileStorage {
...
     }
 
+    #[tracing::instrument(skip(self), level = "debug")]
+    async fn get_slice_stream(
+        &self,
+        path: &Path,
+        range: Range<usize>,
+    ) -> StorageResult<Box<dyn AsyncRead + Send + Unpin>> {
+        let full_path = self.full_path(path)?;
+        let mut file = fs::File::open(full_path).await?;
```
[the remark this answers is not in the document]
rdettai: actually, it's a very valid nitpick! helps readability a lot because using `tokio::fs::File` has quite some implications

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -747,6 +767,23 @@ impl Storage for S3CompatibleObjectStorage {
...
+        path: &Path,
+        range: Range<usize>,
+    ) -> crate::StorageResult<Box<dyn AsyncRead + Send + Unpin>> {
+        let permit = REQUEST_SEMAPHORE.acquire().await;
+        let get_object_output = retry(&self.retry_params, || {
+            self.create_get_object_request(path, Some(range.clone()))
+        })
+        .await?;
+        Ok(Box::new(S3AsyncRead {
+            read: get_object_output.body.into_async_read(),
```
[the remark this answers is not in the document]
rdettai: yes! but it implies more changes in the FileSource which does expect an `AsyncRead` because it can also use stdin when no path is specified: https://github.com/quickwit-oss/quickwit/blob/46e65f8a6cc5cb3ad5bfd6238b27d54fcbaf7dd5/quickwit/quickwit-indexing/src/source/file_source.rs#L141-L156

quickwit/quickwit-storage/src/prefix_storage.rs
```diff
@@ -79,6 +80,14 @@ impl Storage for PrefixStorage {
         self.storage.get_all(&self.prefix.join(path)).await
     }
 
+    async fn get_slice_stream(
+        &self,
+        path: &Path,
+        range: Range<usize>,
+    ) -> crate::StorageResult<Box<dyn AsyncRead + Send + Unpin>> {
+        self.storage.get_slice_stream(path, range).await
```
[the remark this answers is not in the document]
rdettai: thanks!

quickwit/quickwit-storage/src/storage.rs
rdettai: The file source seeks to the location of the last checkpoint, that's why I needed a `get_stream` with an arbitrary starting point. I actually don't need the end, but it makes things simpler to have it similar to the `get_slice`. I'm open to removing the end!

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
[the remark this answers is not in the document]
fulmicoton: Understood!

quickwit/quickwit-storage/src/local_file_storage.rs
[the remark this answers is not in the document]
fulmicoton: I think we want this restriction though. Let's be careful with PR commenting out that kind of stuff. It could easily slip through the CR process.

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -62,6 +64,24 @@ static REQUEST_SEMAPHORE: Lazy<Semaphore> = Lazy::new(|| {
     Semaphore::new(num_permits)
 });
 
+/// Wrap the async read handle together with a permit to keep the permit alive
+/// until the handle is dropped
```
fulmicoton, line 68: great comment.

quickwit/quickwit-storage/src/storage_resolver.rs
```diff
@@ -51,8 +51,7 @@ impl StorageResolver {
         StorageResolverBuilder::default()
     }
 
-    /// Resolves the given URI.
-    pub async fn resolve(&self, uri: &Uri) -> Result<Arc<dyn Storage>, StorageResolverError> {
+    fn get_storage_factory(&self, uri: &Uri) -> Result<&dyn StorageFactory, StorageResolverError> {
```
fulmicoton, line 54: What is the benefit?

rdettai: this is work in progress, I'm looking for ways to properly disable the security in the LocalFileStorage only for the FileSource. One way will be to have a specific resolve function that creates a LocalFileStorage with a flag that disables the check.

quickwit/quickwit-storage/src/local_file_storage.rs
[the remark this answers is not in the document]
rdettai: right! I am still in the process of building up my list of questions and this PR is still a draft! I hope I won't leave this kind of things in when moving it out of draft 😄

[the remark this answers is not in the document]
rdettai: I actually asked this question [on the discord](https://discordapp.com/channels/908281611840282624/908281611840282627/1156577562391236648) because I find that interactive discussions quickly become confusing on Github. But if you prefer, I can centralize all comments here, I don't mind.

quickwit/quickwit-storage/src/storage_resolver.rs
rdettai: I'll actually revert this, I found a simpler way around the check by splitting the URI into the directory + the filename in the source. The tests now go through without disabling the security check!

rdettai:
This is taking shape. Before moving this PR out of draft, I have a few questions:
1. **Cache implementation**: I added a naive implementation that uses `get_slice` and wraps it into a cursor. This implementation is actually never used. Should we rather leave it un-implemented and let it to the person who will first use it to come up with the implementation they want? (if we leave this dummy implementation, I'll obviously add a unit test for it)
2. **Retry on Azure**: the Azure SDK opens a fully lazy stream, so nothing happens until the stream is first polled. This means that we cannot use the usual mechanism here. One workaround to enable the retry mechanism would be to peak into the stream and wrap that into a retry. Do you think it is worth it? 
3. **Buffered reader**: The S3 SDK documentation seems to suggest that the `AsyncRead` should be wrapped into a `BufReader`. The existing file source implementation was also performing this wrapping around the `tokio::fs::File`. Currently, it is still the file source that performs the wrapping, but my feeling is that it should be the `Storage` implementation that takes care of this when it is relevant. Any opinion on this?
4. **Debouncer**: I'm not sure I fully understand the role of the ins an outs of the debouncer. I added a dummy implementation of `get_slice_stream` that bypasses it. Does that seem sound?

rdettai:
Started the full test suite: https://github.com/quickwit-oss/quickwit/actions/runs/6341303249

fulmicoton:
**Cache implementation**: I added a naive implementation that uses get_slice and wraps it into a cursor. This implementation is actually never used. Should we rather leave it un-implemented and let it to the person who will first use it to come up with the implementation they want? (if we leave this dummy implementation, I'll obviously add a unit test for it)

Yes `unimplemented!()` is ok here.

**Retry on Azure**: the Azure SDK opens a fully lazy stream, so nothing happens until the stream is first polled. This means that we cannot use the usual mechanism here. One workaround to enable the retry mechanism would be to peak into the stream and wrap that into a retry. Do you think it is worth it?

> This means that we cannot use the usual mechanism here.
I'm not entirely sure what is not working in this case. If you want to make sure we poll the stream within the body of the `get_stream` method,  a simple workaround could be to wrap the async reader in a tokio's async BufReader. It has a `fill_buf` method. Make sure you comment why you do that for future generations though.

**Buffered reader**: The S3 SDK documentation seems to suggest that the AsyncRead should be wrapped into a BufReader. The existing file source implementation was also performing this wrapping around the tokio::fs::File. Currently, it is still the file source that performs the wrapping, but my feeling is that it should be the Storage implementation that takes care of this 
when it is relevant. Any opinion on this?

Which documentation?
I wonder if this is relevant here. Under the hood the s3 client wraps a stream of large Bytes of object, which size are independent from the way we use the AsyncRead. Does it prevent polling of the next block or something?

**Debouncer**: I'm not sure I fully understand the role of the ins an outs of the debouncer. I added a dummy implementation of get_slice_stream that bypasses it. Does that seem sound?

Yes that's ok. The debouncer detects that several queries currently in flight are actually fetching the same piece of data, and makes sure we only emit one GET request.

rdettai:
### Cache implementation 
ok, switching to unimplemented

### Buffered reader
> Which documentation?

Very sorry for this, I had the link ready but forgot to paste it in: 
- https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/primitives/struct.ByteStream.html#method.into_async_read

But now that I think about it, it's probably just for the convenience of having `.lines()` which comes from [`AsyncBufReadExt`].(https://docs.rs/tokio/latest/tokio/io/trait.AsyncBufReadExt.html).

I took a deeper look at the code, and the `BufReader` is actually required in the `FileSource` for the same reason as in the docs mentioned above: to go through the file line by line ([here](https://github.com/quickwit-oss/quickwit/blob/ae417ffabf1d3547a3efb036d6c8c634c6244e58/quickwit/quickwit-indexing/src/source/file_source.rs#L80)). So my proposed approach for now is to leave things as they are, i.e return the finer grained `AsyncRead` from the `Storage` (exactly as it is returned by the SDKs/`tokio::fs::File`) and wrap it with `BufReader` in the file source, if not for performance reasons at least to get the utility of the line by line iterator.

### Retry on Azure
> If you want to make sure we poll the stream within the body of the get_stream method, a simple workaround could be to wrap the async reader in a tokio's async BufReader. It has a fill_buf method

I feel it would be relevant to ensure the stream is polled at least once on the body, kind of as an equivalent to "opening the file". This should rarely be useful as connectivity checks seem to be usually performed beforehand. To be consitstend with the approach I proposed above for handling `BufReader`, I'd rather not solve this with a `BufReader` here. If that's fine, I'll stick to my peek idea as it's pretty simple an zero cost.

### Debouncer
:+1:

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -168,6 +175,19 @@ impl TypedSourceFactory for FileSourceFactory {
     }
 }
 
+pub fn dir_and_filename(filepath: &Path) -> anyhow::Result<(Uri, &Path)> {
```
fulmicoton, line 178:
```suggestion
fn dir_and_filename(filepath: &Path) -> anyhow::Result<(Uri, &Path)> {
```

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -168,6 +175,19 @@ impl TypedSourceFactory for FileSourceFactory {
...
+    let dir_uri: Uri = filepath
+        .parent()
+        .ok_or_else(|| anyhow::anyhow!("Parent directory could not be resolved"))?
```
fulmicoton, line 181: you need an import from anyhow... I forgot the name of the trait.


```suggestion
        .context("Parent directory could not be resolved")?
```

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -168,6 +175,19 @@ impl TypedSourceFactory for FileSourceFactory {
...
+        .to_str()
+        .ok_or_else(|| anyhow::anyhow!("Path cannot be turned to string"))?
```
fulmicoton, line 183:
```suggestion
        .context("Path cannot be turned to string"))?
```

quickwit/quickwit-storage/src/cache/storage_with_cache.rs
```diff
@@ -71,6 +72,17 @@ impl Storage for StorageWithCache {
         }
     }
 
+    async fn get_slice_stream(
+        &self,
+        path: &Path,
+        _range: Range<usize>,
+    ) -> StorageResult<Box<dyn AsyncRead + Send + Unpin>> {
+        unimplemented!(
```
fulmicoton, line 80: i still think forwarding to the underlying storage was better O_o

quickwit/quickwit-storage/src/local_file_storage.rs
```diff
@@ -231,6 +230,18 @@ impl Storage for LocalFileStorage {
...
 
+    #[tracing::instrument(skip(self), level = "debug")]
+    async fn get_slice_stream(
+        &self,
+        path: &Path,
+        range: Range<usize>,
+    ) -> StorageResult<Box<dyn AsyncRead + Send + Unpin>> {
+        let full_path = self.full_path(path)?;
+        let mut file = tokio::fs::File::open(&full_path).await?;
+        file.seek(SeekFrom::Start(range.start as u64)).await?;
```
fulmicoton, line 241: You ignored range.end. this is a bug.

`.take`

quickwit/quickwit-indexing/src/source/file_source.rs
rdettai: I actually found quite a few other of these in the codebase:
image

But I would prefer not to touch unrelated pieces of code in the context of this PR. Fine by you?

quickwit/quickwit-storage/src/local_file_storage.rs
rdettai: Good catch, sorry for letting this one sleep through! When wanting to add a test for this I discovered the `storage_test_suite` that I missed until now!

quickwit/quickwit-storage/src/cache/storage_with_cache.rs
rdettai: did I misunderstand this https://github.com/quickwit-oss/quickwit/pull/3869#issuecomment-1740164229 ?

> Yes unimplemented!() is ok here.

quickwit/quickwit-indexing/src/source/file_source.rs
fulmicoton: Yes you can ignore them

quickwit/quickwit-storage/src/cache/storage_with_cache.rs
fulmicoton: I am not sure what I was thinking. Anyway let's stick to `unimplemented!()`

fulmicoton (approved):

PR #3871, opened 2023-09-23
update quickstart document

### LINKED ISSUE #3541: Improve stackoverflow quickwit tutorial
Now that we have dynamic aggregations, the quickstart tutorial can be even simpler.

I suggest removing all fields in the mapping except the timestamp, title, and body.

Quickstart doc page: https://github.com/quickwit-oss/quickwit/blob/main/docs/get-started/quickstart.md

JerryKwan:
update quickstart to show the usage of document dynamic mapping

### Description

Update the quickstart document. and show the usage of dynamic mapping to simplify  index configuration

### How was this PR tested?

Closes #3541

### DIFF
--- a/docs/get-started/quickstart.md
+++ b/docs/get-started/quickstart.md
@@ -79,7 +79,7 @@ Let's create an index configured to receive Stackoverflow posts (questions and a
 curl -o stackoverflow-index-config.yaml https://raw.githubusercontent.com/quickwit-oss/quickwit/main/config/tutorials/stackoverflow/index-config.yaml
 ```
 
-The index config defines nine text fields. Among them there are five text fields: `user`, `tags`, `title`, `type` and `body`. Two of these fields, `body` and `title` are [indexed and tokenized](../configuration/index-config.md#text-type) and they are also used as default search fields, which means they will be used for search if you do not target a specific field in your query. The `tags` field is configured to accept multiple text values. The rest of the text fields are not tokenized and configured as [fast](/docs/configuration/index-config.md#text-type). There are three numeric fields `questionId`, `answerId` and `acceptedAnswerId`. And there is the `creationDate` field that serves as the timestamp for each record.
+The index config defines three fields: `title`, `body` and `createDate`. `title` and `body` are [indexed and tokenized](../configuration/index-config.md#text-type) and they are also used as default search fields, which means they will be used for search if you do not target a specific field in your query. And `creationDate` serves as the timestamp for each record. And there is no more explicit field definitions. We can use the default dynamic [mode](/docs/configuration/index-config.md#mode) to simplify the index configuration.
 
 And here is the complete config:
 
@@ -93,18 +93,6 @@ index_id: stackoverflow
 
 doc_mapping:
   field_mappings:
-    - name: user
-      type: text
-      fast: true
-      tokenizer: raw
-    - name: tags
-      type: array<text>
-      fast: true
-      tokenizer: raw
-    - name: type
-      type: text
-      fast: true
-      tokenizer: raw
     - name: title
       type: text
       tokenizer: default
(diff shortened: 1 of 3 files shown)


docs/get-started/quickstart.md
```diff
@@ -79,7 +79,7 @@ Let's create an index configured to receive Stackoverflow posts (questions and a
...
+The index config defines three fields: `title`, `body` and `createDate`. `title` and `body` are [indexed and tokenized](../configuration/index-config.md#text-type) and they are also used as default search fields, which means they will be used for search if you do not target a specific field in your query. And `creationDate` serves as the timestamp for each record. And there is no more explicit field definitions. We can use the default dynamic [mode](/docs/configuration/index-config.md#mode) to simplify the index configuration.
```
fmassot, line 82:
```suggestion
The index config defines three fields: `title`, `body` and `createDate`. `title` and `body` are [indexed and tokenized](../configuration/index-config.md#text-type), and they are also used as default search fields, which means they will be used for search if you do not target a specific field in your query. `creationDate` serves as the timestamp for each record. There are no more explicit field definitions as we can use the default dynamic [mode](/docs/configuration/index-config.md#mode): the undeclared fields will still be indexed, by default the dynamic mode uses the `raw` tokenizer and set `fast: true` to enable aggregation queries.
```

docs/get-started/quickstart.md
```diff
@@ -133,7 +115,7 @@ search_settings:
   default_search_fields: [title, body]
 
 indexing_settings:
-  commit_timeout_secs: 5
```
fmassot, line 136: Do you have a specific reason to modify this value?

fmassot (changes requested):
Can you look at the two comments? 
Also, can you fix the error you pointed out on discord? (in this index configuration file)

docs/get-started/quickstart.md
JerryKwan: To be honest, I copied the config from https://github.com/quickwit-oss/quickwit/blob/main/config/tutorials/stackoverflow/index-config.yaml#L50
Because the document says stackoverflow-index-config.yaml is downloaded using the following command
curl -o stackoverflow-index-config.yaml https://raw.githubusercontent.com/quickwit-oss/quickwit/main/config/tutorials/stackoverflow/index-config.yaml
What value should we use? 5 or 10?

fmassot: 10 is fine!

docs/get-started/quickstart.md
```diff
@@ -79,7 +79,7 @@ Let's create an index configured to receive Stackoverflow posts (questions and a
...
+The index config defines three fields: `title`, `body` and `createDate`. `title` and `body` are [indexed and tokenized](../configuration/index-config.md#text-type), and they are also used as default search fields, which means they will be used for search if you do not target a specific field in your query. `creationDate` serves as the timestamp for each record. There are no more explicit field definitions as we can use the default dynamic [mode](/docs/configuration/index-config.md#mode): the undeclared fields will still be indexed, by default the dynamic mode uses the `raw` tokenizer and set `fast: true` to enable aggregation queries.
```
fmassot, line 82:
```suggestion
The index config defines three fields: `title`, `body` and `creationDate`. `title` and `body` are [indexed and tokenized](../configuration/index-config.md#text-type), and they are also used as default search fields, which means they will be used for search if you do not target a specific field in your query. `creationDate` serves as the timestamp for each record. There are no more explicit field definitions as we can use the default dynamic [mode](/docs/configuration/index-config.md#mode): the undeclared fields will still be indexed, by default fast fields are enabled to enable aggregation queries. and the `raw` tokenizer is used for text. 
```

fmassot:
Almost good to go, I've just fixed a typo.

docs/get-started/quickstart.md
JerryKwan: It's my fault, thanks for the correction

docs/get-started/quickstart.md
JerryKwan: Got it, thank you

JerryKwan:
@fmassot 
Thanks for the correction. I have updated, would you please help to review again? thank you

fmassot (approved):

PR #3880, opened 2023-09-27
Removed metastore -> controlplane event flow via the EventBroker.

fulmicoton:
Refactoring of the control plane.

The two internal actors are replaced by two internal states.

Their execution is not concurrent anymore,
but we the code and the flow is simplified and we avoid some race conditions.

The control plane is not getting the update notify signal through the
broker anymore.

The Notify RPC is entirely removed.

Most of the test in the indexer scheduler become Control plane tests.


quickwit/quickwit-control-plane/src/scheduler.rs
```diff
@@ -497,373 +498,7 @@ fn get_indexing_tasks_diff<'a>(
 
 #[cfg(test)]
 mod tests {
-    use std::collections::{HashMap, HashSet};
```
fulmicoton, line 586: these test moved to `crate::tests` and are applied to the control plane actor

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -273,42 +334,27 @@ impl Handler<DeleteSourceRequest> for ControlPlane {
...
         &mut self,
         request: GetOpenShardsRequest,
-        _: &ActorContext<Self>,
+        ctx: &ActorContext<Self>,
     ) -> Result<Self::Reply, ActorExitStatus> {
-        let response = handle_ask_res!(self.ingest_controller_mailbox.ask_for_res(request).await);
-        Ok(Ok(response))
+        Ok(self
+            .ingest_controller
+            .get_open_shards(request, ctx.progress())
```
guilload, line 356: I've renamed this event `GetOrCreateOpenShardsRequest` to make it clear that it may create new shards. When it does, the indexing scheduler should be notified.

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -273,42 +334,27 @@ impl Handler<DeleteSourceRequest> for ControlPlane {
             index_uid,
             source_id: request.source_id,
         };
-        // TODO: Notify indexing controller.
+
```
guilload, line 337: I think we need to call ` self.ingest_controller.delete_source(&index_uid, &source_id).await;` here.

quickwit/quickwit-control-plane/src/lib.rs
```diff
@@ -44,113 +34,5 @@ pub struct IndexerNodeInfo {
 
 pub type IndexerPool = Pool<String, IndexerNodeInfo>;
 
-/// Subscribes to various metastore events and forwards them to the control plane using the inner
-/// client. The actual subscriptions are set up in `quickwit-serve`.
-#[derive(Debug, Clone)]
-pub struct ControlPlaneEventSubscriber(ControlPlaneServiceClient);
```
guilload, line 50: Arg. I'm adding two callbacks in my PR :D

quickwit/quickwit-control-plane/src/control_plane.rs
fulmicoton: Yes added.

fulmicoton:
> Approved, but let's replug the delete source thingy.

I think it was never plugged.

quickwit/quickwit-control-plane/src/lib.rs
fulmicoton: We can readd whatever is needed.

PR #3895, opened 2023-09-28
Close local shards and delete mrecordlog queues on ingester startup

guilload:
### Description
On startup, ingesters list their local mrecordlog queues and call the metastore to close the corresponding shards. If a shard no longer exists, its queue is deleted. Closed shards whose records have been indexed are eventually deleted by a GC task after a grace period has elapsed. The metastore notifies the control plane of the close and delete shards event.

### How was this PR tested?
- Added unit tests

### DIFF
--- a/quickwit/quickwit-common/src/tower/pool.rs
+++ b/quickwit/quickwit-common/src/tower/pool.rs
@@ -23,17 +23,16 @@ use std::cmp::{Eq, PartialEq};
 use std::collections::HashMap;
 use std::fmt;
 use std::hash::Hash;
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 
 use futures::{Stream, StreamExt};
-use tokio::sync::RwLock;
 
 use super::Change;
 
 /// A pool of `V` values identified by `K` keys. The pool can be updated manually by calling the
 /// `add/remove` methods or by listening to a stream of changes.
 pub struct Pool<K, V> {
-    inner: Arc<RwLock<InnerPool<K, V>>>,
+    pool: Arc<RwLock<HashMap<K, V>>>,
 }
 
 impl<K, V> fmt::Debug for Pool<K, V>
@@ -49,24 +48,19 @@ where
 impl<K, V> Clone for Pool<K, V> {
     fn clone(&self) -> Self {
         Self {
-            inner: self.inner.clone(),
+            pool: self.pool.clone(),
         }
     }
 }
 
-impl<K, V> Default for Pool<K, V> {
+impl<K, V> Default for Pool<K, V>
+where K: Eq + PartialEq + Hash
+{
     fn default() -> Self {
-        let inner = InnerPool {
-            map: HashMap::new(),
-        };
         Self {
-            inner: Arc::new(RwLock::new(inner)),
+            pool: Arc::new(RwLock::new(HashMap::default())),
         }
     }
-}
-
-struct InnerPool<K, V> {
-    map: HashMap<K, V>,
 }
 
 impl<K, V> Pool<K, V>
@@ -85,10 +79,10 @@ where
                 .for_each(|change| async {
                     match change {
                         Change::Insert(key, service) => {
-                            pool.insert(key, service).await;
+                            pool.insert(key, service);
                         }
                         Change::Remove(key) => {
-                            pool.remove(&key).await;
+                            pool.remove(&key);
                         }
                     }
                 })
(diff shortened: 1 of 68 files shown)


quickwit/quickwit-proto/src/control_plane/mod.rs
```diff
@@ -52,22 +52,22 @@ impl From<ControlPlaneError> for MetastoreError {
...
 
 impl From<tonic::Status> for ControlPlaneError {
     fn from(status: tonic::Status) -> Self {
-        match status.code() {
-            tonic::Code::Unavailable => {
-                ControlPlaneError::Unavailable(status.message().to_string())
-            }
-            _ => ControlPlaneError::Internal(status.message().to_string()),
-        }
+        serde_json::from_str(status.message()).unwrap_or_else(|_| {
```
guilload, line 65: JSON serialization of errors is a bit brittle, but it'll do for now. We need a way to go through the gRPC layer without losing the error variants.

guilload (commented):
This is ready for review!

quickwit/quickwit-common/src/tower/pool.rs
```diff
@@ -20,20 +20,19 @@
...
 use futures::{Stream, StreamExt};
-use tokio::sync::RwLock;
 
 use super::Change;
 
 /// A pool of `V` values identified by `K` keys. The pool can be updated manually by calling the
 /// `add/remove` methods or by listening to a stream of changes.
 pub struct Pool<K, V> {
-    inner: Arc<RwLock<InnerPool<K, V>>>,
+    pool: Arc<DashMap<K, V>>,
```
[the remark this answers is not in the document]
guilload: I'm having second thoughts: the lock is never held across an .await point, so using a sync lock makes sense, but `DashMap` seems overkill t_t

Thoughts?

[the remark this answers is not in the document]
fulmicoton: It is a sharded `RwLock<HashMap>` under the hood.
I'd rather use a vanilla `RwLock<HashMap>` and returns Vec like before.

Dashmap's iterators are hiding the fact that as long as the iterator lives, an Iterator is held.

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -135,13 +135,36 @@ impl Actor for ControlPlane {
...
+
         ctx.schedule_self_msg(CONTROL_PLAN_LOOP_INTERVAL, ControlPlanLoop)
             .await;
 
         Ok(())
     }
 }
 
+#[async_trait]
+impl Handler<RefreshPlanLoop> for ControlPlane {
```
fulmicoton, line 147: I think we can remove it. (I assume it has resurfaced during the rebase?)

quickwit/quickwit-common/src/tower/pool.rs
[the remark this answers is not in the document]
guilload: Sounds good.

quickwit/quickwit-control-plane/src/control_plane.rs
guilload: I'm one rebase behind.

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -213,6 +218,8 @@ impl Handler<DeleteIndexRequest> for ControlPlane {
     }
 }
 
+// This handler is a metastore call proxied through the control plane: we must first forward the
```
fulmicoton, line 221: i love this comment.

quickwit/quickwit-ingest/src/ingest_v2/gc.rs
```diff
@@ -0,0 +1,225 @@
...
+use quickwit_proto::ingest::IngestV2Result;
+use quickwit_proto::metastore::{DeleteShardsRequest, DeleteShardsSubrequest};
+use quickwit_proto::{split_queue_id, IndexUid, QueueId, ShardId, SourceId};
+use tokio::sync::{RwLock, Semaphore};
+use tracing::{error, info};
+
+use super::ingest_metastore::IngestMetastore;
+use super::ingester::IngesterState;
+
+/// Period of time after which shards candidate for deletion are actually deleted.
```
fulmicoton, line 36: It would be nicer to explicit what we mean by deleting shards here:
Of course this is not about truncating data. At this point hopefully the data has been deleted, but this is about deleting from the metastore.

fulmicoton: Also I would but the last two lines of description on the GcTask doc instead of. the GRACE_PERIOD doc

quickwit/quickwit-ingest/src/ingest_v2/gc.rs
```diff
@@ -0,0 +1,225 @@
...
+/// Deletes shards asynchronously after the grace period has elapsed.
+///
+/// 1. deletes the shards from the metastore
+/// 2. removes the shards from in-memory data structures
+/// 3. deletes the associated mrecordlog queues
+pub(super) struct GcTask {
+    metastore: Arc<dyn IngestMetastore>,
+    mrecordlog: Arc<RwLock<MultiRecordLog>>,
+    state: Arc<RwLock<IngesterState>>,
+    gc_candidates: Vec<QueueId>,
```
fulmicoton, line 62: I'm not sure about the `candidates` naming. We want to remove all of those right?
`queues_to_remove`?

quickwit/quickwit-ingest/src/ingest_v2/gc.rs
```diff
@@ -0,0 +1,225 @@
...
+        let gc_task = Self {
+            metastore,
+            mrecordlog,
+            state,
+            gc_candidates,
+        };
+        tokio::spawn(gc_task.run());
+    }
+
+    async fn run(self) -> IngestV2Result<()> {
```
fulmicoton, line 84:
```suggestion
    async fn run_after_grace_period(self) -> IngestV2Result<()> {
```

quickwit/quickwit-ingest/src/ingest_v2/gc.rs
```diff
@@ -0,0 +1,225 @@
...
+
+/// Limits the number of concurrent GC tasks.
+static MAX_CONCURRENCY_SEMAPHORE: Lazy<Arc<Semaphore>> =
+    Lazy::new(|| Arc::new(Semaphore::new(MAX_CONCURRENCY)));
+
+/// Deletes shards asynchronously after the grace period has elapsed.
+///
+/// 1. deletes the shards from the metastore
+/// 2. removes the shards from in-memory data structures
+/// 3. deletes the associated mrecordlog queues
```
fulmicoton, line 57: I am not sure exactly how to fix this, but the structure of the code and the naming made me thing that this task was a loop.

I think I would have been less confused with a simple function that does not include the sleep in this file, and 
a call that includes the grace period sleep on the call site.

quickwit/quickwit-serve/src/ingest_metastore.rs
```diff
@@ -0,0 +1,65 @@
...
+use quickwit_proto::metastore::{
+    CloseShardsRequest, CloseShardsResponse, DeleteShardsRequest, DeleteShardsResponse,
+};
+
+// TODO: Remove when the metastore is code generated in `quickwit-proto`.
+
+/// Implementation of the [`quickwit_ingest::IngestMetastore`] trait. See comment in the module
+/// where it is defined for more details about why this is required.
+#[derive(Clone)]
+pub(crate) struct IngestMetastore {
```
fulmicoton, line 34: :-/ so IngestMetastore is both a trait and struct.

Can we have different names?

fulmicoton (approved):

quickwit/quickwit-ingest/src/ingest_v2/gc.rs
guilload: It's not just the queues, but yes, I can find a better name.

quickwit/quickwit-ingest/src/ingest_v2/gc.rs
guilload: ...because it was a loop at some point :D I will package everything in a function.

quickwit/quickwit-serve/src/ingest_metastore.rs
guilload: Yes, I will rename. Hopefully, it'll go away soon :)

Yes, it does because `IngestMetastore` is defined in `quickwit-ingest`.

PR #3898, opened 2023-09-30
Migrate metastore to code generated metastore service

fmassot:
TODO
- [x] Readd ability to get index metadata by `IndexUid`
- [x] Instrument metastore with a tower layer
- [x] Readd `check_connectivity` and `uris` methods (add a new trait `MetastoreServiceExt`?)
- [x] Readd `index_exists` method
- [x] When testing, it's easy to `MetastoreServiceClient::new(mock_metastore)`, but we need to wrap the mock. Also it's possible to do `MockMetastoreService::new()` and I think we don't want a developer to do that => I added an assert if a dev instantiate with `MetastoreServiceClient::new` instead of `MetastoreServiceClient::from`.
- [ ] ~~Some proto requests have a `default` implementation that will generate an empty json payload, this is bad~~ The solution would be to have only optional fields.... 
- [x] Simplify the building of some gRPC requests.
- [x] Simplify how metrics are added to gRPC requests
- [x] Clean code
- [x] Check how proto files are generated and see how to avoid repetition of ".type_attribute("Shard", "#[derive(Eq)]")"
- [x] make test-all
- [x] test local cluster with 3 nodes and check create index - ingest - delete index works.

Next steps:
- [ ] Put the metastore documentation in the internal docs
- [ ] Replace the control plane metastore implementation by a proxy layer. The annoying part is that we would need to implement each `Service<R>` for each metastore request on the control plane client. I don't have a concrete plan on this and prefer to open an issue to fix it later. I may miss an obvious solution, let's discuss that tomorrow.
- [ ] Open an issue to refactor the metastore factory cache.
- [ ] Replace `ListSplitsRequest::try_from_index_uid` by `ListSplitsRequest::from_index_uid` as we are sure this builder won't fail.


quickwit/quickwit-index-management/src/garbage_collection.rs
```diff
@@ -94,23 +100,31 @@ pub async fn run_garbage_collect(
...
-            .into_iter()
-            .map(|meta| meta.split_metadata)
-            .collect();
+    let list_deletable_staged_request = ListSplitsRequest::try_from_list_splits_query(query)?;
+    let deletable_staged_splits: Vec<SplitMetadata> = protect_future(
+        progress_opt,
+        metastore.list_splits(list_deletable_staged_request),
+    )
+    .await?
+    .deserialize_splits()?
```
guilload, line 109: `.deserialize_split_metadatas()?` ext?

quickwit/quickwit-index-management/src/garbage_collection.rs
```diff
@@ -124,14 +138,18 @@ pub async fn run_garbage_collect(
     }
 
     // Schedule all eligible staged splits for delete
-    let split_ids: Vec<&str> = deletable_staged_splits
+    let split_ids: Vec<String> = deletable_staged_splits
```
guilload, line 141:
```suggestion
    let split_ids: Vec<SplitId> = deletable_staged_splits
```

quickwit/quickwit-index-management/src/garbage_collection.rs
```diff
@@ -174,19 +192,28 @@ async fn delete_splits_marked_for_deletion(
             .with_update_timestamp_lte(updated_before_timestamp)
             .with_limit(DELETE_SPLITS_BATCH_SIZE);
 
-        let list_splits_result = protect_future(progress_opt, metastore.list_splits(query)).await;
+        let list_splits_request = match ListSplitsRequest::try_from_list_splits_query(query) {
+            Ok(request) => request,
+            Err(error) => {
+                error!(error = ?error, "Failed to build list splits request.");
```
guilload, line 198:
```suggestion
                error!(error = ?error, "failed to build list splits request");
```

quickwit/quickwit-index-management/src/garbage_collection.rs
```diff
@@ -174,19 +192,28 @@ async fn delete_splits_marked_for_deletion(
             .with_update_timestamp_lte(updated_before_timestamp)
             .with_limit(DELETE_SPLITS_BATCH_SIZE);
 
-        let list_splits_result = protect_future(progress_opt, metastore.list_splits(query)).await;
+        let list_splits_request = match ListSplitsRequest::try_from_list_splits_query(query) {
```
guilload, line 195: It's tempting to remove the `try_from`. If this fails, we have a bug somewhere.

quickwit/quickwit-index-management/src/garbage_collection.rs
```diff
@@ -280,15 +307,16 @@ pub async fn delete_splits_from_storage_and_metastore(
         }
     };
     if !successes.is_empty() {
-        let split_ids: Vec<&str> = successes
+        let split_ids: Vec<String> = successes
```
guilload, line 310:
```suggestion
        let split_ids: Vec<SplitId> = successes
```

quickwit/quickwit-index-management/src/garbage_collection.rs
```diff
@@ -280,15 +307,16 @@ pub async fn delete_splits_from_storage_and_metastore(
...
+            .map(|split_info| split_info.split_id.to_string())
             .collect();
-        let metastore_result = protect_future(
-            progress_opt,
-            metastore.delete_splits(index_uid.clone(), &split_ids),
-        )
-        .await;
+        let delete_splits_request = DeleteSplitsRequest {
+            index_uid: index_uid.to_string(),
+            split_ids: split_ids.clone(),
```
guilload, line 316: Are you sure we need to clone?

quickwit/quickwit-index-management/src/index.rs
```diff
@@ -237,45 +285,66 @@ impl IndexService {
...
-        let split_ids: Vec<&str> = splits.iter().map(|split| split.split_id()).collect();
-        self.metastore
-            .mark_splits_for_deletion(index_uid.clone(), &split_ids)
-            .await?;
-        let split_metas: Vec<SplitMetadata> = splits
+        let splits_metadata: Vec<SplitMetadata> = self
+            .metastore
+            .list_all_splits(ListAllSplitsRequest::from(&index_uid))
+            .await?
+            .deserialize_splits()?
```
guilload, line 306: Looks like we could also use `deserialize_split_ids`.

quickwit/quickwit-index-management/src/garbage_collection.rs
fmassot: We can remove the error message, is that what you mean? 
I'm fine with removing it, we return the appropriate metastore error that we could log instead.

quickwit/quickwit-index-management/src/garbage_collection.rs
fmassot: yeah... I would rather keep this for now.

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/file_backed_metastore_factory.rs
```diff
@@ -44,11 +44,9 @@ use crate::{FileBackedMetastore, Metastore, MetastoreFactory, MetastoreResolverE
...
 pub struct FileBackedMetastoreFactory {
     storage_resolver: StorageResolver,
-    // We almost never garbage collect the dangling Weak pointers
-    // here. This is judged to not be much of a problem however.
-    //
-    // In a normal run, this cache will contain a single Metastore.
-    cache: Arc<Mutex<HashMap<Uri, Weak<dyn Metastore>>>>,
+    // We never garbage collect unused metastore client instances. This should not be a problem
+    // because during normal use this cache will hold at most a single instance.
+    cache: Arc<Mutex<HashMap<Uri, MetastoreServiceClient>>>,
```
guilload, line 49: If this is true, maybe we should use an `Arc<Mutex<Option<(Uri, MetastoreServiceClient)>>>` and always evict the old client if URIs don't match (and log a warning?)

quickwit/quickwit-metastore/src/metastore/mod.rs
```diff
@@ -357,8 +106,361 @@ pub enum ListIndexesQuery {
...
+    /// Deserializes the `query_json` field of a [`ListIndexesMetadataRequest`] into a
+    /// [`ListIndexesQuery`].
+    fn deserialize_list_indexes_query(&self) -> MetastoreResult<ListIndexesQuery>;
+}
+
+impl ListIndexesMetadataRequestExt for ListIndexesMetadataRequest {
+    fn try_from_list_indexes_query(
+        list_indexes_query: ListIndexesQuery,
+    ) -> MetastoreResult<ListIndexesMetadataRequest> {
+        let query_json = serde_json::to_string(&list_indexes_query).map_err(|error| {
```
guilload, line 129: there's a util to do that in `proto::metastore`

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -466,17 +474,21 @@ where
 }
 
 #[async_trait]
-impl Metastore for PostgresqlMetastore {
-    async fn check_connectivity(&self) -> anyhow::Result<()> {
-        self.connection_pool.acquire().await?;
-        Ok(())
+impl MetastoreService for PostgresqlMetastore {
+    async fn check_connectivity(&mut self) -> anyhow::Result<()> {
+        todo!()
```
guilload, line 479: todo!

guilload (approved):

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/file_backed_metastore_factory.rs
fmassot: I like the idea and I tested it. Only a few tests are failing if we changed this. We need to also to refactor the factory as now the code between the file backed factory cache and the postgres factory cache is exactly the same.

quickwit/quickwit-proto/protos/quickwit/metastore.proto
```diff
@@ -129,12 +126,12 @@ message CreateIndexResponse {
   string index_uid = 1;
 }
```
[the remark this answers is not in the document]
fmassot: yes I just changed the `IndexMetadataRequest` that you can build from an index ID or index UID.

quickwit/quickwit-cli/src/lib.rs
```diff
@@ -256,23 +257,18 @@ fn get_resolvers(
...
 pub async fn run_index_checklist(
-    metastore: &dyn Metastore,
+    metastore: &mut MetastoreServiceClient,
     storage_resolver: &StorageResolver,
     index_id: &str,
     source_config_opt: Option<&SourceConfig>,
 ) -> anyhow::Result<()> {
     let mut checks: Vec<(&str, anyhow::Result<()>)> = Vec::new();
 
-    // The metastore is file-backed, so we must check the storage first.
```
fmassot, line 266: This check is useless. The check connectivity of the metastore will do the right check.

guilload: Leave it. The checklist will tell you it's a metastore issue, whereas it stems from the storage.

fmassot: Ok, will change the comment then.

fmassot: I would prefer to do the right check in the metastore method and make sure the error displayed is clean.

Issue #3797, opened 2023-09-04
Allow larger payload in the Rest API.


fulmicoton:
Currently we limit payload to 10MB by default.

Some clients needs the limit to be increased. Ideally we would prefer to have this configurable.

kamalesh0406:
@fulmicoton are you working on this or can I take up the task?

fulmicoton:
You can take the task! I just assigned it to you. Thanks

guilload:
Closed via #3905.

PR #3905, opened 2023-10-03
Add parameter to configure content limit for ingest api

kamalesh0406:
### Description

This change adds the ability for the user to configure the content limit for the ingest api. It is related to the feature request made in #3797.

### How was this PR tested?

I have tried adding unit tests but they were not working as expected. I wanted the PR out to get the initial comments, I will add the tests in a follow-up commit to this PR.

### DIFF
--- a/quickwit/quickwit-config/src/node_config/mod.rs
+++ b/quickwit/quickwit-config/src/node_config/mod.rs
@@ -178,6 +178,7 @@ pub struct IngestApiConfig {
     pub max_queue_memory_usage: Byte,
     pub max_queue_disk_usage: Byte,
     pub replication_factor: usize,
+    pub content_length_limit: u64
 }
 
 impl Default for IngestApiConfig {
@@ -186,6 +187,7 @@ impl Default for IngestApiConfig {
             max_queue_memory_usage: Byte::from_bytes(2 * 1024 * 1024 * 1024), /* 2 GiB // TODO maybe we want more? */
             max_queue_disk_usage: Byte::from_bytes(4 * 1024 * 1024 * 1024), /* 4 GiB // TODO maybe we want more? */
             replication_factor: 1,
+            content_length_limit: 10 * 1024 * 1024 /* 10 MB */
         }
     }
 }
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -186,6 +187,7 @@ impl Default for IngestApiConfig {
             max_queue_memory_usage: Byte::from_bytes(2 * 1024 * 1024 * 1024), /* 2 GiB // TODO maybe we want more? */
             max_queue_disk_usage: Byte::from_bytes(4 * 1024 * 1024 * 1024), /* 4 GiB // TODO maybe we want more? */
             replication_factor: 1,
+            content_length_limit: 10 * 1024 * 1024 /* 10 MB */
```
ChillFish8, line 190: `10 << 20` would probably be better, but I think 10MB is too low that would be about 250ms worth of data 1/4 of the data a single pipeline could ingest. 100MB might be more suitable.

fulmicoton (changes requested):
Can we add a unit test checking we reject payload above the limit configured?
A test on the filter level is ok.

quickwit/quickwit-config/src/node_config/mod.rs
kamalesh0406: Okay I will modify the default value to 100MB.

kamalesh0406:
> Can we add a unit test checking we reject payload above the limit configured? A test on the filter level is ok.

Sure I was working on the unit test I will push it when its done.

quickwit/quickwit-config/src/node_config/mod.rs
guilload: 10 was perfectly fine. Our gRPC stack does not allow messages bigger than 4 MB at the moment. Not to mention rate limiting only applies after ingest payloads are already in memory.

kamalesh0406: I agree with @guilload as well, I have reverted it back to 10MB.

fulmicoton (approved):

fulmicoton:
Thank you @kamalesh0406 !

PR #3908, opened 2023-10-03
Call `truncate` on leaders AND followers

guilload:
### Description
Before this change, indexers called truncate on leaders, and leaders replicated the truncation request on followers. After this change, indexers are responsible for calling `truncate` on both leaders and followers. Leaders are no longer in charge of replicating the request. 

That way, followers see their queues truncated even if leaders are unavailable. The code is also simplified.

### How was this PR tested?
Updated tests

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/gc.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/gc.rs
@@ -202,7 +202,8 @@ mod tests {
             mrecordlog.clone(),
             state.clone(),
         );
-        tokio::time::sleep(REMOVAL_GRACE_PERIOD * 2).await;
+        // Wait for the removal task to run.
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-ingest/src/ingest_v2/gc.rs
```diff
@@ -202,7 +202,8 @@ mod tests {
             mrecordlog.clone(),
             state.clone(),
         );
-        tokio::time::sleep(REMOVAL_GRACE_PERIOD * 2).await;
+        // Wait for the removal task to run.
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
fulmicoton, line 206: I suspect this test will be flaky.

fulmicoton (approved):

PR #3911, opened 2023-10-04
Use same lock for mrecordlog and ingester state

guilload:
### Description
Use a single lock for mrecordlog and ingester state to avoid deadlocks.

### How was this PR tested?
Updated unit tests but no functional changes.

### DIFF
--- a/quickwit/quickwit-doc-mapper/benches/doc_to_json_bench.rs
+++ b/quickwit/quickwit-doc-mapper/benches/doc_to_json_bench.rs
@@ -19,6 +19,7 @@
 
 use criterion::{criterion_group, criterion_main, Criterion, Throughput};
 use quickwit_doc_mapper::DocMapper;
+use tantivy::TantivyDocument;
 
 const JSON_TEST_DATA: &str = include_str!("data/simple-parse-bench.json");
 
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-common/src/tower/buffer.rs
```diff
@@ -166,10 +166,7 @@ impl<R> fmt::Debug for BufferLayer<R> {
 
 impl<R> Clone for BufferLayer<R> {
```
fulmicoton, line 167: `#[derive(Copy, Clone)]` did not work here?

guilload: No, the derived `Clone` implementation unnecessarily adds `Clone` bounds on `S` and `R`:
```rust
#[automatically_derived]
        impl<S: ::core::clone::Clone, R: ::core::clone::Clone> ::core::clone::Clone
        for Buffer<S, R>
        where
            S: Service<R>,
        {
            #[inline]
            fn clone(&self) -> Buffer<S, R> {
                Buffer {
                    bound: ::core::clone::Clone::clone(&self.bound),
                    inner: ::core::clone::Clone::clone(&self.inner),
                }
            }
        }
```

fulmicoton (approved):

PR #3912, opened 2023-10-04
Optimize replication task process loop

guilload:
### Description
Optimize replication task process loop.

### Todo
- [x] ~~use bounded channels~~
- [ ] ~~handle replication errors~~

### How was this PR tested?
- Updated unit tests


guilload:
Also ready for review. Using unbounded channels is fine as long as the upstream SYN replication channel is bounded, which it is. I added comments where those unbounded channels are used. I'll handle replication errors in a dedicated PR.

fulmicoton (changes requested):
I don't understand this code. I think it needs a markdown doc somewhere to explain the big picture.

I suspect this big picture cannot be inline in the code as it is to understand a lot of the code we have here.

fulmicoton (approved):

PR #3919, opened 2023-10-06
Store versioned records in MRecordLog

guilload:
Attempt to store versioned records in MRecordLog. I'm opening this PR as a draft as a basis for discussion. Records are still formed of a header and a payload, but this time, the header is a ProtocolBuf message, which provides more flexibility and handles backward and forward compatibility more gracefully.

However, in the current state, this change allocates a new vector for each record ingested. The previous scheme knew the size and value of each header, so it was easy to reuse statically allocated headers. We can batch docs in a single record to amortize the creation of a single header, which we discussed before, or revert to the previous way.


quickwit/quickwit-ingest/src/ingest_v2/mrecord.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub fn new_commit() -> Self {
+        Self {
+            header: MRecordHeader::new_commit(),
+            payload: MRecordPayload::Commit,
+        }
+    }
+
+    pub fn encode(&self) -> impl Buf {
+        // TODO: this is going to allocate for each record. No bueno.
+        let header_buf = Bytes::from(self.header.encode_length_delimited_to_vec());
```
guilload, line 51: Here is the header allocation. Suggestions welcome :)

quickwit/quickwit-ingest/src/ingest_v2/mrecord.rs
```diff
@@ -0,0 +1,102 @@
...
+use bytes::{Buf, Bytes};
+use prost::Message;
+use quickwit_proto::ingest::{m_record_header, MRecordHeader};
+
+pub(super) enum MRecordPayload {
+    Doc(Bytes),
+    Commit,
+}
+
+pub(super) struct MRecord {
```
fulmicoton, line 29: I am not fond of the implicitly entangle double `enums` (MRecordPayload and MRecordHeader)

I would suggest
```
enum MRecord {
      Doc(DocRecord),
      Commit(CommitHeader),
}

struct DocRecord {
    header: DocHeader,
    payload: Bytes, 
}
```

or if the intermediary struct does not help (as in, is not useful in other methods prototypes)

```
enum MRecord {
      Doc {
           header: DocHeader,
           payload: Bytes,
      },
      Commit(CommitHeader),
}
```

quickwit/quickwit-ingest/src/ingest_v2/mrecord.rs
fulmicoton: The usual solution is to have a MRecordDocEncoder object that you add as a member of the `Ingester`. It holds an internal buffer and has a 
`fn encode(&mut self, mrecord: &MRecord) -> Buf`

Unfortunately in your case, lifetime will get in the way, because you want to use it in an iterator.

It is a none problem with the Iterator trait. It is also the raison d'etre of the Streamer trait for instance. https://docs.rs/fst/latest/fst/trait.Streamer.html
I don't see any simple solution unfortunately.

fulmicoton (approved):

fulmicoton:
Can you just add a ticket so we remember to remove the allocation?

PR #3920, opened 2023-10-06
Refactoring of the control plane

fulmicoton:
This paves the way for the construction of a partial view of the
metastore, by isolating the model nested in the ingest controller as
a control plane model.

It also change the error handling. Now upon a metastore error
for which we do not know whether the control plane has lost consistency
or not, we kill the control plane.

Finally the control plane is now supervized.


quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -155,6 +162,36 @@ impl Handler<ControlPlanLoop> for ControlPlane {
...
+        | MetastoreError::Db { .. } => false,
+    };
+    if metastore_failure_is_certain {
+        // If the metastore failure is certain, this is actually a good thing.
+        // We do not need to restart the control plane.
+        Ok(Err(ControlPlaneError::Metastore(metastore_error)))
+    } else {
+        // If the metastore failure is uncertain, we need to restart the control plane
+        // so that it gets resynced with the metastore state.
+        Err(ActorExitStatus::from(anyhow::anyhow!(metastore_error)))
```
guilload, line 191: Do we make sure we're correctly logging an error when that happens? Additionally, we should emit a metric when a control plane crash/restart occurs. As an operator, I'd certainly want to be able to set an alert on that event.

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -699,4 +742,126 @@ mod tests {
     async fn test_control_plane_close_shards() {
         // TODO: Write test when the RPC is actually called by ingesters.
     }
+
+    #[tokio::test]
+    async fn test_control_plane_supervision_reload_from_metastore() {
```
guilload, line 747: Cool test.

guilload (approved):

quickwit/quickwit-control-plane/src/control_plane.rs
fulmicoton: nice catch! This is a very good point.

fulmicoton: ill add the log in this PR and add the counter in a different PR. I created an issue to avoid forgetting

PR #3938, opened 2023-10-11
Extending the control plane model

fulmicoton:
The control plane model now offers a concise view of the Metastore.
The indexing scheduler query the metastore at all.
This also cleans up the hack the populates the list of indexing task with ingest v2 tasks by mutating the list of tasks

Refactoring:

We rely on SourceUid more broadly in the code.
The unit test were using IndexUid in a broken way in places.
This is now fixed.

This PR introduces a LogicalIndexTask. The idea is to have it hold all
information necessary for us to do the placement.
Right now this means only the max pipeline per index parameter (allowing
us to remove the ugly lookup in the source config map).
We will need to also add information about the desired number of pipeline, possibly the leader/follower for locality, etc.

In the next PR we will most likely also break the 1:1 relationship
between LogicalIndexTask and IndexTask.

The idea would be to have a single logical index task, and possibly break it into several IndexTask.

### DIFF
--- a/quickwit/quickwit-config/src/source_config/mod.rs
+++ b/quickwit/quickwit-config/src/source_config/mod.rs
@@ -43,6 +43,7 @@ pub const CLI_INGEST_SOURCE_ID: &str = "_ingest-cli-source";
 pub const INGEST_API_SOURCE_ID: &str = "_ingest-api-source";
 
 /// Reserved source ID used for native Quickwit ingest.
+/// (this is for ingest v2)
 pub const INGEST_SOURCE_ID: &str = "_ingest-source";
 
 pub const RESERVED_SOURCE_IDS: &[&str] =
--- a/quickwit/quickwit-control-plane/src/control_plane.rs
+++ b/quickwit/quickwit-control-plane/src/control_plane.rs
@@ -126,14 +126,14 @@ impl Actor for ControlPlane {
         self.model
             .load_from_metastore(&*self.metastore, ctx.progress())
             .await
-            .context("Failed to intiialize the model")?;
+            .context("failed to intialize the model")?;
 
         if let Err(error) = self
             .indexing_scheduler
             .schedule_indexing_plan_if_needed(&self.model)
         {
             // TODO inspect error.
-            error!("Error when scheduling indexing plan: `{}`.", error);
+            error!("error when scheduling indexing plan: `{}`.", error);
         }
 
         ctx.schedule_self_msg(CONTROL_PLAN_LOOP_INTERVAL, ControlPlanLoop)
@@ -175,7 +175,7 @@ fn convert_metastore_error<T>(
 ) -> Result<ControlPlaneResult<T>, ActorExitStatus> {
     // If true, we know that the transactions has not been recorded in the Metastore.
     // If false, we simply are not sure whether the transaction has been recorded or not.
-    let metastore_failure_is_certain = match &metastore_error {
+    let is_transaction_certainly_aborted = match &metastore_error {
         MetastoreError::AlreadyExists(_)
         | MetastoreError::FailedPrecondition { .. }
         | MetastoreError::Forbidden { .. }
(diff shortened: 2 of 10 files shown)


quickwit/quickwit-config/src/source_config/mod.rs
```diff
@@ -43,6 +43,7 @@ pub const CLI_INGEST_SOURCE_ID: &str = "_ingest-cli-source";
 pub const INGEST_API_SOURCE_ID: &str = "_ingest-api-source";
 
 /// Reserved source ID used for native Quickwit ingest.
+/// (this is for ingest v2)
```
fulmicoton, line 46: can I get confirmation that this is for ingest v2?

fulmicoton:
Simple questions for @fmassot and @guillad:

- The logic of the creation of the indexing task was relying heavily on our special source ids.
I change it for a check on the source type. Was there a reason for using source ids?
- Can I get confirmation that  INGEST_API_SOURCE_ID=> ingest v1 while INGEST_SOURCE_ID => ingest v2?

quickwit/quickwit-control-plane/src/lib.rs
```diff
@@ -20,17 +20,26 @@
...
 use quickwit_common::tower::Pool;
 use quickwit_proto::control_plane::{ControlPlaneService, ControlPlaneServiceClient};
 use quickwit_proto::indexing::{IndexingServiceClient, IndexingTask};
 use quickwit_proto::metastore::{CloseShardsRequest, DeleteShardsRequest};
+use quickwit_proto::{IndexUid, SourceId};
 use tracing::error;
 
+/// It can however appear only once in a given index.
+/// In itself, `SourceId` is not unique, but the pair `(IndexUid, SourceId)` is.
+#[derive(PartialEq, Eq, Debug, PartialOrd, Ord, Hash, Clone)]
```
fulmicoton, line 37: This is not a new struct. (It has been moved from indexing_plan I think)

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -272,67 +274,106 @@ impl From<IndexingTask> for SourceUid {
...
+) -> Vec<LogicalIndexingTask> {
+    let mut indexing_tasks: Vec<LogicalIndexingTask> = Vec::new();
+    for (source_uid, source_config) in model
+        .get_source_configs()
         .sorted_by(|(left, _), (right, _)| left.cmp(right))
     {
         if !source_config.enabled {
             continue;
         }
-        if [CLI_INGEST_SOURCE_ID, INGEST_SOURCE_ID].contains(&source_config.source_id.as_str()) {
```
fulmicoton, line 287: I did not understand why the logic here was working off the source_id and not the source_type.
I moved that to a source type based logic. Is this risky?

guilload: I don't know. It does not seem risky.

guilload (approved):

quickwit/quickwit-control-plane/src/control_plane_model.rs
```diff
@@ -178,19 +185,58 @@ impl ControlPlaneModel {
...
-    pub(crate) fn add_source(&mut self, index_uid: &IndexUid, source_id: &SourceId) {
-        self.shard_table.add_source(index_uid, source_id);
+    /// Adds a source to a given index. Returns an error if a source with the same source_id already
+    /// exists.
+    pub(crate) fn add_source(
+        &mut self,
+        index_uid: &IndexUid,
+        source_config: SourceConfig,
+    ) -> ControlPlaneResult<()> {
+        self.shard_table
```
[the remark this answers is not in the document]
fulmicoton: great point!

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -287,21 +289,22 @@ impl Handler<AddSourceRequest> for ControlPlane {
...
+            .add_source(index_uid.clone(), source_config.clone())
             .await
         {
             return convert_metastore_error(metastore_error);
         };
 
-        self.model.add_source(&index_uid, &source_id);
+        self.model
+            .add_source(&index_uid, source_config)
+            .context("Failed to add source.")?;
```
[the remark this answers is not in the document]
fulmicoton: acknowledged!

Issue #3748, opened 2023-08-15
Improve scroll api to rely on search_after parameter


fulmicoton:
Related to #3523

PR #3941, opened 2023-10-11
implement  search after

### LINKED ISSUE #3523: Support elasticsearch `search_after` parameter in a search query
[Elasticseardh doc ref](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html)

trinity-1686a:
### Description

fix #3523
implement ES search_after

### How was this PR tested?

added unit test and rest-api test

fix a test on scroll which was made unrepresentative of what's really happening due to desync between mocking and real code behaviour.
start supporting more than 2 sort values in collector (but not yet outside of it), as it makes easier to reason about the code than having two Option<>

### DIFF
--- a/quickwit/quickwit-proto/protos/quickwit/search.proto
+++ b/quickwit/quickwit-proto/protos/quickwit/search.proto
@@ -167,8 +167,9 @@ message SearchRequest {
   optional uint32 scroll_ttl_secs = 15;
 
   // Document with sort tuple smaller or equal to this are discarded to
-  // enable pagination
-  repeated SortByValue search_after = 16;
+  // enable pagination.
+  // If split_id is empty, no comparhison with _shard_doc should be done
+  optional PartialHit search_after = 16;
 }
 
 message SortField {
--- a/quickwit/quickwit-proto/protos/quickwit/search.proto
+++ b/quickwit/quickwit-proto/protos/quickwit/search.proto
@@ -165,6 +165,10 @@ message SearchRequest {
   // that will make it possible to paginate through the results
   // in a consistent manner.
   optional uint32 scroll_ttl_secs = 15;
+
+  // Document with sort tuple smaller or equal to this are discarded to
+  // enable pagination
+  repeated SortByValue search_after = 16;
 }
 
 message SortField {
--- a/quickwit/quickwit-proto/src/search/mod.rs
+++ b/quickwit/quickwit-proto/src/search/mod.rs
@@ -67,6 +67,26 @@ impl From<SortValue> for SortByValue {
     }
 }
 
+impl SortByValue {
+    pub fn into_json(self) -> serde_json::Value {
+        use serde_json::Value::*;
+        match self.sort_value {
+            Some(SortValue::U64(num)) => Number(num.into()),
+            Some(SortValue::I64(num)) => Number(num.into()),
+            Some(SortValue::F64(num)) => {
+                if let Some(num) = serde_json::Number::from_f64(num) {
+                    Number(num)
+                } else {
+                    // TODO is there a better way to handle infinite/nan?
+                    Null
+                }
+            }
+            Some(SortValue::Boolean(b)) => Bool(b),
+            None => Null,
+        }
+    }
+}
+
 impl Copy for SortValue {}
 impl Eq for SortValue {}
 
--- a/quickwit/quickwit-proto/src/search/mod.rs
+++ b/quickwit/quickwit-proto/src/search/mod.rs
@@ -52,11 +52,6 @@ impl fmt::Display for SplitSearchError {
     }
 }
 
-// !!! Disclaimer !!!
-//
-// Prost imposes the PartialEq derived implementation.
-// This is terrible because this means Eq, PartialEq are not really in line with Ord's
-// implementation. if in presence of NaN.
 impl Eq for SortByValue {}
 impl Copy for SortByValue {}
 impl From<SortValue> for SortByValue {
@@ -64,6 +59,12 @@ impl From<SortValue> for SortByValue {
         SortByValue {
             sort_value: Some(sort_value),
         }
+    }
+}
+
+impl std::hash::Hash for SortByValue {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.sort_value.hash(state);
     }
 }
 
@@ -85,10 +86,38 @@ impl SortByValue {
             None => Null,
         }
     }
+
+    pub fn try_from_json(value: serde_json::Value) -> Option<Self> {
+        use serde_json::Value::*;
+        let sort_value = match value {
+            Null => None,
+            Bool(b) => Some(SortValue::Boolean(b)),
+            Number(number) => {
+                if let Some(number) = number.as_u64() {
+                    Some(SortValue::U64(number))
+                } else if let Some(number) = number.as_i64() {
+                    Some(SortValue::I64(number))
+                } else if let Some(number) = number.as_f64() {
+                    Some(SortValue::F64(number))
+                } else {
+                    // this should never happen as we don't emit such number ourselves
+                    return None;
+                }
+            }
+            // TODO this means we actually ignore _shard_doc
+            String(_) | Array(_) | Object(_) => return None,
+        };
+        Some(SortByValue { sort_value })
+    }
 }
 
-impl Copy for SortValue {}
+// !!! Disclaimer !!!
+//
+// Prost imposes the PartialEq derived implementation.
+// This is terrible because this means Eq, PartialEq are not really in line with Ord's
+// implementation. if in presence of NaN.
 impl Eq for SortValue {}
+impl Copy for SortValue {}
 
 impl Ord for SortValue {
     fn cmp(&self, other: &Self) -> Ordering {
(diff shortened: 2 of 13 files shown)


quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -165,6 +165,11 @@ message SearchRequest {
   // that will make it possible to paginate through the results
   // in a consistent manner.
   optional uint32 scroll_ttl_secs = 15;
+
+  // Document with sort tuple smaller or equal to this are discarded to
+  // enable pagination.
+  // If split_id is empty, no comparhison with _shard_doc should be done
```
fulmicoton, line 171:
```suggestion
  // If split_id is empty, no comparison with _shard_doc should be done
```

quickwit/quickwit-proto/src/search/mod.rs
```diff
@@ -118,6 +166,67 @@ impl PartialOrd for SortValue {
...
+            SortValue::Boolean(b) => {
+                3u8.hash(state);
+                b.hash(state);
+            }
+        }
+    }
+}
+
+impl SortValue {
+    pub fn normalize(&self) -> Self {
```
fulmicoton, line 194:
```suggestion
    /// When possible, casts round floats to an integer.
    pub fn normalize(&self) -> Self {
```

quickwit/quickwit-proto/src/search/mod.rs
```diff
@@ -118,6 +166,67 @@ impl PartialOrd for SortValue {
...
+            }
+            SortValue::F64(number) => {
+                let number = *number;
+                if number.ceil() == number {
+                    // number is not NaN, and is a natural number
+                    if number.is_sign_positive() && number <= u64::MAX as f64 {
+                        // number is positive (including positive zero) and less than u64::MAX, we
+                        // can convert it safely
+                        return SortValue::U64(number as u64);
+                    } else if number.is_sign_negative() && number >= i64::MIN as f64 {
```
fulmicoton, line 213: this goes against what we do outside in tantivy.

We tend to prefer i64 whenever possible.
so if ceil == number {

   if number within i64 range -> cast to i64 
   if number wthin. u64 range -> cast to u64
   else { do nothing }

quickwit/quickwit-search/src/collector.rs
```diff
@@ -293,25 +233,64 @@ enum AggregationSegmentCollectors {
...
+        let mut sort_values = sort_values.into_iter();
+        let sort_value = sort_values.next().flatten();
+        let sort_value2 = sort_values.next().flatten();
 
         let hit = PartialHit {
             sort_value: sort_value.map(Into::into),
             sort_value2: sort_value2.map(Into::into),
             // we actually know the split_id, but the hit is likely to be discarded, so clonning it
-            // would cause a probably useless allocation. TODO use an Arc<str> instead?
+            // would cause a probably useless allocation. TODO use an Arc<str>/Rc<str> instead?
```
fulmicoton, line 293: In a different PR (it won't be merged before yours, no worry), I removed those by introducing a SegmentPartialHit https://github.com/quickwit-oss/quickwit/pull/3915/files#diff-e46a7a27f5f472b1b9897b219c56db65c7af89a6ee04d783a402d4688edd4d13

I think it is  not a bad idea. If you agree could we do this refactoring?

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -165,6 +165,11 @@ message SearchRequest {
   // that will make it possible to paginate through the results
   // in a consistent manner.
   optional uint32 scroll_ttl_secs = 15;
+
+  // Document with sort tuple smaller or equal to this are discarded to
+  // enable pagination.
+  // If split_id is empty, no comparison with _shard_doc should be done
```
fulmicoton, line 171: What is the purpose of making `split_id` empty?

quickwit/quickwit-search/src/lib.rs
```diff
@@ -105,6 +112,40 @@ impl GlobalDocAddress {
...
+        f.write_str(&self.split)?;
+        write!(
+            f,
+            ":{:08x}:{:08x}",
+            self.doc_addr.segment_ord, self.doc_addr.doc_id
+        )
+    }
+}
+
+impl std::str::FromStr for GlobalDocAddress {
```
fulmicoton, line 126: do we use this?

quickwit/quickwit-serve/src/search_api/rest_handler.rs
```diff
@@ -241,6 +241,9 @@ pub fn search_request_from_api_request(
             .map(|agg| serde_json::to_string(&agg).expect("could not serialize JsonValue")),
         sort_fields: search_request.sort_by.sort_fields,
         scroll_ttl_secs: None,
+        // TODO do we want to support search_after on QW api? If so, what should the request look
```
fulmicoton, line 244: i think you can open a ticket and remove the TODO

quickwit/quickwit-search/src/root.rs
```diff
@@ -291,6 +291,9 @@ fn simplify_search_request_for_scroll_api(req: &SearchRequest) -> SearchRequest
         snippet_fields: Vec::new(),
         // We remove the scroll ttl parameter. It is irrelevant to process later request
         scroll_ttl_secs: None,
+        // TODO unclear as to what we should do when scroll and search after are mixed. I guess
+        // honor search_after?
+        search_after: req.search_after.clone(),
```
fulmicoton, line 296: Different subject than your TODO, but can you open a ticket to have scroll use the superior search_after mechanics?

For the TODO: I think it is best to either return an error OR do whatever elasticsearch does.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -293,25 +233,64 @@ enum AggregationSegmentCollectors {
...
+                // TODO actually it's not first, it should be what's in _shard_doc then first then
+                // default
+                let order = self
+                    .top_k_hits
+                    .sort_key_mapper
+                    .orders
+                    .first()
+                    .unwrap_or(&SortOrder::Desc);
+                cmp_result = cmp_result
+                    .then_with(|| order.compare(&self.split_id, &search_after.split_id))
```
fulmicoton, line 274: Can we compute that when we build the `SegmentCollector`? Comparing strings in collect is extremely expensive.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -293,25 +233,64 @@ enum AggregationSegmentCollectors {
...
 
 impl QuickwitSegmentCollector {
     #[inline]
     fn collect_top_k(&mut self, doc_id: DocId, score: Score) {
-        let (sort_value, sort_value2) =
-            self.score_extractor.extract_typed_sort_value(doc_id, score);
+        let sort_values: Vec<_> = self
+            .score_extractors
+            .iter()
+            .map(|extractor| extractor.extract_typed_sort_value_opt(doc_id, score))
```
fulmicoton, line 251: No allocation in this function please. this is called once per document match by the query.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -293,25 +233,64 @@ enum AggregationSegmentCollectors {
 pub struct QuickwitSegmentCollector {
     num_hits: u64,
     split_id: String,
-    score_extractor: SortingFieldExtractorPair,
+    score_extractors: Vec<SortingFieldExtractorComponent>,
```
fulmicoton, line 236: See comment below this is not a Vec today for good reasons.
Did you need it to be a Vec?

quickwit/quickwit-search/src/collector.rs
```diff
@@ -485,31 +465,32 @@ impl Collector for QuickwitCollector {
...
+            .sort_by
+            .iter()
+            .map(|sort_by| sort_by.sort_order())
+            .collect();
+        let sort_key_mapper = HitSortingMapper { orders };
         Ok(QuickwitSegmentCollector {
             num_hits: 0u64,
             split_id: self.split_id.clone(),
-            score_extractor,
+            score_extractors,
```
fulmicoton, line 478: See above, I prefer the old code.

quickwit/quickwit-proto/protos/quickwit/search.proto
trinity-1686a: if you sort by `["my_field1", "my_field2"]`, without a final `_shard_doc`, ES doesn't return a unique doc identifier, so when the client comes back for more, it can't provide such an identifier in `search_after`. That's when `split_id` is empty

quickwit/quickwit-search/src/lib.rs
trinity-1686a: yes, when we parse `search_after` for a doc id

quickwit/quickwit-search/src/root.rs
trinity-1686a: > Different subject than your TODO, but can you open a ticket to have scroll use the superior search_after mechanics?

there is already #3748 for just that. I actually took #3523 to unblock that ticket

quickwit/quickwit-search/src/collector.rs
trinity-1686a: having a pair of options everywhere makes most code weird, plus limit us in term of feature. I'd prefer to have something that can be iterated over, and ideally which can grow beside two elements. Would using a smallvec, to inline up to N elements (with N somewhere somewhere in 3..=5) be fine? It wouldn't allocate for most queries, and would also check my boxes

fulmicoton: See comment below. We cannot do this with without benchmarking it, and let's not do this in this PR.

trinity-1686a: ftr, when collecting half a million docs, here is how long it took (not great for vec and smallvec)
```
double option:
test collector::tests::bench_collection ... bench:  95,939,300 ns/iter (+/- 2,402,897)
vec:
test collector::tests::bench_collection ... bench: 131,689,753 ns/iter (+/- 2,208,591)
smallvec:
test collector::tests::bench_collection ... bench: 137,068,856 ns/iter (+/- 1,300,872)
```

quickwit/quickwit-search/src/root.rs
trinity-1686a: > `search_after` cannot be used in a scroll context.

this is ES behaviour.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -684,6 +730,8 @@ pub(crate) fn sort_by_from_request(search_request: &SearchRequest) -> SortByPair
     let to_sort_by_component = |field_name: &str, order| {
         if field_name == "_score" {
             SortByComponent::Score { order }
+        } else if field_name == "_shard_doc" || field_name == "_doc" {
```
fulmicoton, line 733: I have no clue what `_shard_doc` or `_doc`.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -308,6 +312,33 @@ impl QuickwitSegmentCollector {
...
+            let mut cmp_result = orders
+                .order1
+                .compare_opt(&sort_value, &search_after_value1)
+                .then_with(|| {
+                    orders
+                        .order2
+                        .compare_opt(&sort_value2, &search_after_value2)
+                });
+            if !search_after.split_id.is_empty() {
+                // TODO actually it's not first, it should be what's in _shard_doc then first then
```
fulmicoton, line 328: I don't understand this comment.
Also "_shard_doc" is an elasticsearch concept.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -684,6 +730,8 @@ pub(crate) fn sort_by_from_request(search_request: &SearchRequest) -> SortByPair
     let to_sort_by_component = |field_name: &str, order| {
         if field_name == "_score" {
             SortByComponent::Score { order }
+        } else if field_name == "_shard_doc" || field_name == "_doc" {
+            SortByComponent::DocId { order }
```
fulmicoton, line 734: I don't think this is the place to have `_shard_doc`
If we want to support it, that should be in the Elasticsearch REST API handler.

fulmicoton (approved):

quickwit/quickwit-search/src/collector.rs
trinity-1686a: yes, but I expect people using search_after on a PIT (which we don't really have yet) to want it for tie-breaking. Note that in ES, you can't use `_shard_doc` outside of a PIT. I take it it's really something related to \*that document\* at \*that time\*, so it's something we can reasonably

Issue #3907, opened 2023-10-03
Expose REST API endpoint for listing (and filtering) splits


guilload:
- Add REST endpoint in the Open API spec and update the docs.
- Add cursor pagination based on split ID.
- Update the response format to have `{"size": 1000, "results": [...]}`.

JerryKwan:
@guilload @fmassot 
Would you please help to briefly describe the expected outcome of the endpoint? thank you

JerryKwan:
I think we should expose the functions provided by ListSplitsQuery ?
https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-metastore/src/metastore/mod.rs#L362

fmassot:
So in fact we already have some filters in the code @guilload. I totally forgot that.

The annoying part right now is that we cannot do pagination and we should return an object and not a list as reported in #2491.

Some things we have to keep in mind for this issue:
- we could return tens of thousands of splits, even more... Doing pagination with offset can be an issue. I would prefer to use sorting on a unique key and paginate with a filter.
- we want a response format consistent for the different resources: indexes, splits, and checkpoints (maybe others?).
- we can use the main filters available in the `ListSplitsQuery` struct.


Endpoint REST: 
`GET /api/v1/indexes/{my-index}/splits?split_states=Published&start_timestamp=&end_timestamp=`

Response format:

```json
{
   "size": 1000,
   "results": [...]
}
```

JerryKwan:
@fmassot 
Thanks for the information
I have checked the codes in index rest handler, and it seems like there has already some codes about the endpoint
https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-serve/src/index_api/rest_handler.rs#L280-L313

fmassot:
> Do we need to refactor the output data format and add paginate support?
Yes.

guilload:
I missed that endpoint because it's not documented in our Open API spec. Let's add it.

Also, we can easily implement cursor pagination instead of offset pagination with split IDs.

fmassot:
> Also, we can easily implement cursor pagination instead of offset pagination with split IDs.

Oh indeed, I just checked the properties of ulid and this is perfect for pagination.

fmassot:
@JerryKwan I updated the issue description.

JerryKwan:
@fmassot @guilload 
Thanks for helping, I will work on it asap, thank you

JerryKwan:
[...]

PR #3942, opened 2023-10-12
expose rest api for splits

JerryKwan:
expose rest api for split list
change rest api response data structure
add offset/limit pagniation
change SplitClient list() accordingly

### Description
Solve the issues described in https://github.com/quickwit-oss/quickwit/issues/3907

### How was this PR tested?

Currently it supports the operations like this
curl  "http://127.0.0.1:7280/api/v1/indexes/stackoverflow/splits?offset=10&limit=10"
and this
./quickwit split list --index stackoverflow --offset 10

And the response of the rest api looks like the following
```
{
  "offset": 0,
  "size": 1,
  "splits": [
    {
      "split_state": "Published",
      "update_timestamp": 1695642901,
      "publish_timestamp": 1695642901,
      "version": "0.6",
      "split_id": "01HB632HD8W6WHNM7CZFH3KG1X",
      "index_uid": "stackoverflow:01HB6321TDT3SP58D4EZP14KSX",
      "partition_id": 0,
      "source_id": "_ingest-api-source",
      "node_id": "jerry",
      "num_docs": 10000,
      "uncompressed_docs_size_in_bytes": 6674940,
      "time_range": {
        "start": 1217540572,
        "end": 1219335682
      },
      "create_timestamp": 1695642900,
      "maturity": {
        "type": "immature",
        "maturation_period_millis": 172800000
      },
      "tags": [],
      "footer_offsets": {
        "start": 4714989,
        "end": 4719999
      },
      "delete_opstamp": 0,
      "num_merge_ops": 0
    }
  ]
}
```

Closes: #3907


quickwit/quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -241,6 +241,14 @@ fn describe_index_handler(
 #[derive(Debug, Clone, Deserialize, Serialize, utoipa::IntoParams, utoipa::ToSchema, Default)]
 #[into_params(parameter_in = Query)]
 pub struct ListSplitsQueryParams {
+    /// If set, restrict the number of splits to skip
```
fmassot, line 244:
```suggestion
    /// If set, define the number of splits to skip
```

quickwit/quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -263,12 +271,22 @@ pub struct ListSplitsQueryParams {
...
 }
 
+#[derive(Serialize, Deserialize, Debug, utoipa::ToSchema)]
+pub struct ListSplitsResponse {
+    #[serde(default)]
+    pub offset: usize,
+    #[serde(default)]
+    pub size: usize,
+    #[serde(default)]
+    pub splits: Vec<Split>,
```
fmassot, line 281: We will probably use pagination for other resources: indexes for example. Replacing the field `splits` would be better.

fmassot (changes requested):
I have only one minor comment, can you have a look at it?

quickwit/quickwit-serve/src/index_api/rest_handler.rs
JerryKwan: Currently it's only used as the response of list split operation, that's why I use splits as the name. Maybe we can define different structures to to be used in different responses

fmassot (approved):

PR #3947, opened 2023-10-12
Remove dead --commit-timeout CLI argument

rdettai:
### Description

The --commit-timeout is declared in the docs and the CLI but the provided value is not used. 

This change might break scripts that call the ingest CLI with this argument, but that's probably better than ignoring it.

### How was this PR tested?

Tried to hunt down every usage of this arg.

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -324,7 +324,6 @@ quickwit index ingest
     [--batch-size-limit <batch-size-limit>]
     [--wait]
     [--force]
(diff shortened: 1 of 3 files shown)


guilload (approved):

fmassot:
Look at this place

```
/// Timeout for the ingest operations that require waiting for commit.
```

rdettai:
Looking into this 😃

rdettai:
Indeed the argument is forwarded as part of `client_args`, I missed that, sorry!

PR #3948, opened 2023-10-12
Implement `IndexUId` as `(String, Ulid)`

guilload:
Draft proposal.

### DIFF
--- a/quickwit/quickwit-proto/protos/quickwit/metastore.proto
+++ b/quickwit/quickwit-proto/protos/quickwit/metastore.proto
@@ -23,6 +23,11 @@ package quickwit.metastore;
 
 import "quickwit/ingest.proto";
 
+message IndexUid {
+  string index_id = 1;
+  uint64 incarnation_id_high = 2;
+  uint64 incarnation_id_low = 3;
+}
 
 enum SourceType {
   CLI = 0;
@@ -216,8 +221,8 @@ message DeleteTask {
 }
 
 message DeleteQuery {
-  // Index ID.
-  string index_uid = 1;
+  // Index UID.
+  IndexUid index_uid = 1;
   // If set, restrict search to documents with a `timestamp >= start_timestamp`.
   optional int64 start_timestamp = 2;
   // If set, restrict search to documents with a `timestamp < end_timestamp``.
--- a/quickwit/quickwit-proto/src/metastore/mod.rs
+++ b/quickwit/quickwit-proto/src/metastore/mod.rs
@@ -22,7 +22,9 @@ use std::fmt;
 use quickwit_common::retry::Retryable;
 use serde::{Deserialize, Serialize};
 
-use crate::{queue_id, IndexId, QueueId, ServiceError, ServiceErrorCode, SourceId, SplitId};
+use crate::{
+    queue_id, IndexId, IndexUid, QueueId, ServiceError, ServiceErrorCode, SourceId, SplitId,
+};
 
 pub mod events;
 
@@ -218,6 +220,14 @@ impl CloseShardsSuccess {
 impl CloseShardsFailure {
     pub fn queue_id(&self) -> QueueId {
         queue_id(&self.index_uid, &self.source_id, self.shard_id)
+    }
+}
+
+impl DeleteQuery {
+    pub fn index_uid(&self) -> &IndexUid {
+        self.index_uid
+            .as_ref()
+            .expect("`index_uid` should be a required field")
     }
 }
 
--- a/quickwit/quickwit-proto/src/types/index_uid.rs
+++ b/quickwit/quickwit-proto/src/types/index_uid.rs
@@ -1,0 +1,208 @@
+// Copyright (C) 2023 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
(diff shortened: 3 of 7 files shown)


quickwit/quickwit-proto/protos/quickwit/metastore.proto
```diff
@@ -23,6 +23,11 @@ package quickwit.metastore;
 
 import "quickwit/ingest.proto";
 
+message IndexUid {
```
guilload, line 26: Mapping this message to `quickwit_proto::types::IndexUid` using the extern path feature, we are never exposed to that message.

quickwit/quickwit-proto/protos/quickwit/metastore.proto
```diff
@@ -216,8 +221,8 @@ message DeleteTask {
 }
 
 message DeleteQuery {
-  // Index ID.
-  string index_uid = 1;
+  // Index UID.
+  IndexUid index_uid = 1;
```
guilload, line 225: Now, we have a typed `IndexUid` in protobuf messages. Yay!
Now, we have to deal with an option. Nay!

quickwit/quickwit-proto/src/metastore/mod.rs
```diff
@@ -221,6 +223,14 @@ impl CloseShardsFailure {
     }
 }
 
+impl DeleteQuery {
+    pub fn index_uid(&self) -> &IndexUid {
```
guilload, line 227: One solution to get a `&IndexUid`.

quickwit/quickwit-proto/src/types/index_uid.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl<'de> Deserialize<'de> for IndexUid {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        unimplemented!()
+    }
+}
+
+/// Encodes an index UID into a protobuf message composed of one string and two u64 integers.
+impl ::prost::Message for IndexUid {
```
guilload, line 91: The trick. If this is too complex for your taste, we can always go with `(String, String)`.

quickwit/quickwit-proto/src/types/index_uid.rs
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    fn clear(&mut self) {
+        self.index_id.clear();
+        self.incarnation_id = Ulid::nil();
+    }
+}
+
+#[macro_export]
+macro_rules! index_uid {
```
guilload, line 178: One way to move the `index_uid` field from an option without moving the other fields of the protobuf message.

quickwit/quickwit-proto/src/types/mod.rs
```diff
@@ -47,83 +49,10 @@ pub fn queue_id(index_uid: &str, source_id: &str, shard_id: u64) -> QueueId {
...
-    }
-
-    pub fn as_str(&self) -> &str {
-        &self.0
-    }
-
-    pub fn from_parts(index_id: impl Into<String>, incarnation_id: impl Into<String>) -> Self {
-        let incarnation_id = incarnation_id.into();
-        let index_id = index_id.into();
-        if incarnation_id.is_empty() {
```
guilload, line 81: This is needed for backward compatibility. We should no longer need this as we're dropping backward compatibility with 0.4 and 0.5 in the next release.

quickwit/quickwit-proto/protos/quickwit/metastore.proto
trinity-1686a: I think that's a breaking change and would technically require to change the field number. I don't know how much we care about that though.

trinity-1686a (commented):
I really like the idea of having proper typing as soon as deserialization end. On the other hand, I don't like the manual impl of `prost::Message`, but I don't think that's avoidable.

Do you want me to take it from there?

quickwit/quickwit-proto/src/types/index_uid.rs
```diff
@@ -0,0 +1,208 @@
...
+        match tag {
+            1u32 => {
+                let value = &mut self.index_id;
+                ::prost::encoding::string::merge(wire_type, value, buf, ctx).map_err(|mut error| {
+                    error.push(STRUCT_NAME, "index_id");
+                    error
+                })
+            }
+            2u32 => {
+                let (mut ulid_high, ulid_low) = self.incarnation_id.into();
```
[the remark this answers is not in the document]
trinity-1686a: I doubt we would be able to measure the difference. Our serialization does a fair bit of allocation, for instance for the 1st field of that same struct. The variance from just that would probably make measurements too noisy to figure out if anything changed

quickwit/quickwit-proto/src/types/index_uid.rs
guilload: Trinity suggests using the getter and cloning if an owned object is required. The call to `clone()` will hopefully be optimized away. I agree :)

quickwit/quickwit-proto/src/types/index_uid.rs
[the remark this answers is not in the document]
guilload: The `From` conversions in the `ulid` crate do exactly that.

PR #3877, opened 2023-09-26
Make ui production build work.

fmassot:

PR #3950, opened 2023-10-12
Fix nightly builds

rdettai:
### Description

We are still having errors in the frontend build in the nightly CI after https://github.com/quickwit-oss/quickwit/pull/3877.

From the logs, it seems `node-gyp` is not installed. Even though yarn seems to install it automatically, when we compare the logs of various runs, it seems that it does so in parallel, and there might be some race condition there that would explain the transient failures. Eagerly installing `node-gyp` might help.

### How was this PR tested?

### DIFF
--- a/.github/actions/cargo-build-macos-binary/action.yml
+++ b/.github/actions/cargo-build-macos-binary/action.yml
@@ -20,6 +20,8 @@ runs:
         node-version: 18
         cache: "yarn"
         cache-dependency-path: quickwit/quickwit-ui/yarn.lock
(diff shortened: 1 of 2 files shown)


rdettai:
@guilload I am not 100% sure about this fix, but tried to run it and it worked: https://github.com/quickwit-oss/quickwit/actions/runs/6498384922/job/17649604209. I propose we merge this and see if the nightly passes tonight (and the nights after 😄).

guilload (approved):

PR #3951, opened 2023-10-12
Fix the cluster change stream when nodes rejoin the cluster quickly

guilload:
### Description
The issue is that the cluster logic keeps track of nodes joining and leaving using (`node_id`, `incarnation_id`), while client pools keep track of nodes using only `node_id`.

When a node restarts and becomes ready quickly, the new node (`node_id` + `incarnation_id + 1`) joins the cluster before the previous node leaves and the following cluster change events are triggered:

1. Add (`node_id` + `incarnation_id + 1`)
2. Remove (`node_id` + `incarnation_id`)

It translates at the pool level to:
1. Add `node_id`
2. Remove `node_id`

The bug used to be less visible because we used to diff the whole pool for each Chitchat event, but now the diff is at the node level (incremental).

To fix the issue, I only emit a  `Remove` event when incarnation IDs match. Additionally, when a node rejoins the cluster, instead of doing nothing, I virtually emit two consecutive  `Remove` and then `Add` events to give the pools a chance to react to configuration changes.

### How was this PR tested?
- Updated unit tests
- Verified locally that the issue no longer occurs

### DIFF
--- a/quickwit/quickwit-cli/src/tool.rs
+++ b/quickwit/quickwit-cli/src/tool.rs
@@ -51,6 +51,7 @@ use quickwit_indexing::IndexingPipeline;
 use quickwit_ingest::IngesterPool;
 use quickwit_metastore::Metastore;
 use quickwit_proto::search::SearchResponse;
+use quickwit_proto::NodeId;
 use quickwit_search::{single_node_search, SearchResponseRest};
 use quickwit_serve::{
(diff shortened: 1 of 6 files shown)


fulmicoton (approved):

fulmicoton (approved):
I approved the PR, but before merging can you make sure to copy paster your PR description in your commit description?

It was extremely important to understand what this is about.

Issue #3758, opened 2023-08-17
Craft SQL queries with query builder instead of manual string wrangling


guilload:
[SeaQuery](https://docs.rs/sea-query/latest/sea_query/)?

divinenaman:
@guilload can I work on this ?

kamalesh0406:
Can you assign this to me, I will take a look!

guilload:
Close via #3959. Thanks again @kamalesh0406!

PR #3959, opened 2023-10-16
Update sql strings with sea_query sql statements

kamalesh0406:
### Description

This PR is a draft version to kickstart #3758. I used `sea_query` to update one sql query and I want to get the initial comments from @guilload on whether this is fine. If the `sea_query` implementation seems good, then I will continue with converting the remaining sql strings into the sea query orm syntax. Otherwise, I will try to reach other options.

### How was this PR tested?

This PR was tested using `cargo test --features=postgres`.

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
+++ b/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
@@ -25,7 +25,6 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use async_trait::async_trait;
-use itertools::Itertools;
 use quickwit_common::uri::Uri;
 use quickwit_common::PrettySample;
 use quickwit_config::{
@@ -45,6 +44,10 @@ use quickwit_proto::metastore::{
     ToggleSourceRequest, UpdateSplitsDeleteOpstampRequest, UpdateSplitsDeleteOpstampResponse,
 };
 use quickwit_proto::types::IndexUid;
+use sea_query::{
+    all, any, Asterisk, Cond, Expr, Func, PostgresQueryBuilder, Query, SelectStatement,
+};
+use sea_query_binder::SqlxBinder;
 use sqlx::migrate::Migrator;
 use sqlx::postgres::{PgConnectOptions, PgDatabaseError, PgPoolOptions};
 use sqlx::{ConnectOptions, Pool, Postgres, Transaction};
@@ -53,7 +56,7 @@ use tracing::log::LevelFilter;
 use tracing::{debug, error, info, instrument, warn};
 
 use crate::checkpoint::IndexCheckpointDelta;
-use crate::metastore::postgresql_model::{PgDeleteTask, PgIndex, PgSplit};
+use crate::metastore::postgresql_model::{PgDeleteTask, PgIndex, PgSplit, Splits, ToTimestampFunc};
 use crate::metastore::{instrument_metastore, FilterRange, PublishSplitsRequestExt};
 use crate::{
     AddSourceRequestExt, CreateIndexRequestExt, IndexMetadata, IndexMetadataResponseExt,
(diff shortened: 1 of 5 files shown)


guilload:
Thanks for working on this. Starting small with an example was the way to go. Good call!

It looks good, but I wonder if it's more readable at the end of the day. Can you try to use `sea_query` in the `write_sql_filter` and `build_query_filter` functions where the ugliness lives and SQL injections can happen? We could migrate those exclusively, WDYT?

kamalesh0406:
Yup that sounds good to me. I will start working on updating them.

kamalesh0406:
@guilload I have completed about 95% of the updates for the `write_sql_filter` and `build_query_filter`. I need to add one more expression and fix the code in the unit tests. Can you give me a review of the current state and let me know if you are fine with proceeding further?

guilload:
Looking very good so far. You can keep going. Thank you!

kamalesh0406:
@guilload I have completed the migration for the two functions. Please do check the unit tests because I had to modify the strings due to some quirks in seaquery.

quickwit/quickwit-metastore/src/metastore/postgresql_model.rs
```diff
@@ -22,10 +22,21 @@ use std::str::FromStr;
 
 use quickwit_proto::metastore::{DeleteQuery, DeleteTask, MetastoreError, MetastoreResult};
 use quickwit_proto::IndexUid;
+use sea_query::{Iden, Write};
 use tracing::error;
 
 use crate::{IndexMetadata, Split, SplitMetadata, SplitState};
 
+#[derive(Iden, Clone)]
```
guilload, line 30:
```suggestion
#[derive(Iden, Clone, Copy)]
```

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -223,131 +227,135 @@ async fn index_metadata(
...
-) {
-    match &filter_range.start {
-        Bound::Included(value) => {
-            let _ = write!(sql, " AND {} >= {}", field_name, (value_formatter)(value));
-        }
-        Bound::Excluded(value) => {
-            let _ = write!(sql, " AND {} > {}", field_name, (value_formatter)(value));
-        }
-        Bound::Unbounded => {}
+fn write_sql_filter<'a, V: Display>(
```
guilload, line 230: The combination of passing the base statement as a mutable reference AND having to return the reference is weird. We should either pass the base statement as a mutable reference and return nothing OR pass the base statement by value and return it. I have a preference for the former because it also allows removing the lifetime.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -223,131 +227,135 @@ async fn index_metadata(
...
-    match &filter_range.end {
-        Bound::Included(value) => {
-            let _ = write!(sql, " AND {} <= {}", field_name, (value_formatter)(value));
-        }
-        Bound::Excluded(value) => {
-            let _ = write!(sql, " AND {} < {}", field_name, (value_formatter)(value));
-        }
-        Bound::Unbounded => {}
+    if let Bound::Excluded(value) = &filter_range.start {
+        sql = sql.cond_where(Expr::col(field_name.clone()).gt((value_formatter)(value)))
```
guilload, line 241: With `Copy` the calls to `.clone()` are no longer needed.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -223,131 +227,135 @@ async fn index_metadata(
...
-        }
-        Bound::Unbounded => {}
+fn write_sql_filter<'a, V: Display>(
+    mut sql: &'a mut SelectStatement,
+    field_name: Splits,
+    filter_range: &'a FilterRange<V>,
+    value_formatter: impl Fn(&V) -> Expr,
+) -> &'a mut SelectStatement {
+    if let Bound::Included(value) = &filter_range.start {
+        sql = sql.cond_where(Expr::col(field_name.clone()).gte((value_formatter)(value)))
```
guilload, line 237: No need to reassign since the function is `&mut self`

```suggestion
        sql.cond_where(Expr::col(field_name.clone()).gte((value_formatter)(value)));
```

guilload (commented):
Looking good. The `'tag' = ANY(tags)` might be a deal breaker :/ Let's see what they say.

For testing:
```
make docker-compose-up DOCKER_SERVICES='postgres'
QW_TEST_DATABASE_URL=postgres://quickwit-dev:quickwit-dev@localhost:5432/quickwit-metastore-dev cargo test --manifest-path quickwit/Cargo.toml -p quickwit-metastore --lib --all-features -- --nocapture postgres
```

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1581,192 +1667,313 @@ mod tests {
...
+        let sql = select_statement.column(Asterisk).from(Splits::Table);
+
         let query =
             ListSplitsQuery::for_index(index_uid.clone()).with_tags_filter(TagFilterAst::Tag {
                 is_present: false,
                 tag: "tag-2".to_string(),
             });
-        let sql = build_query_filter(String::new(), &query);
+        append_query_filters(sql, &query);
+        //// PLEASE CONFIRM IF SWITCHING FROM ANY(TAGS) TO ANY(SELECT TAGS FROM SPLITS) IS FINE.
```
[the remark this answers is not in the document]
kamalesh0406: I just realized a way to fix it. We can use a custom expression(https://docs.rs/sea-query/latest/sea_query/expr/struct.Expr.html#method.cust) that returns `ANY(tags)`, I need to test if that will work though.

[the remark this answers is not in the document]
guilload: Awesome!

guilload (approved):

guilload:
We can merge this once the conflict is solved.

PR #3960, opened 2023-10-16
Adding tracing to fetch docs operations and changing the conccurrency

fulmicoton:
logic.

Previously the code was strangely combining spawning each individual tasks in to a different tokio task, and using a futures_unordered to wait for their results.


quickwit/quickwit-search/src/fetch_docs.rs
```diff
@@ -245,12 +245,11 @@ async fn fetch_docs_in_split(
                     snippet_json: Some(snippet_json),
                 },
             ))
-        })
+        }.in_current_span()
```
trinity-1686a, line 248: rustfmt is unhappy here

trinity-1686a (approved):

Issue #3923, opened 2023-10-06
Add kafka source example with several pipelines in the source configuration page


fmassot:
Just something like this will do the job

```yaml
version: 0.6
source_id: kafka-source
source_type: kafka
max_num_pipelines_per_indexer: 2
desired_num_pipelines: 3
params:
  topic: vpc-logs
  client_params:
    bootstrap.servers: broker-1:9092,broker-2:9092
```

PR #3961, opened 2023-10-16
change kafka source examples

JerryKwan:
add documents about max_num_pipelines_per_indexer and desired_num_pipelines 
change kafka source examples

### Description

update kafka source examples

### How was this PR tested?

Closes:#3923


config/tutorials/gh-archive/kafka-source.yaml
```diff
@@ -1,6 +1,8 @@
 version: 0.6
 source_id: kafka-source
 source_type: kafka
+max_num_pipelines_per_indexer: 2
```
fmassot, line 4: Can you set it to 1 as generally it's better to start with this default value?

JerryKwan: Updated

fmassot (approved):

PR #3962, opened 2023-10-16
update tantivy

PSeitz:
### Description

### How was this PR tested?

### DIFF
--- a/quickwit/Cargo.lock
+++ b/quickwit/Cargo.lock
@@ -3517,6 +3517,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "lru"
+version = "0.12.0"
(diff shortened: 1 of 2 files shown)


quickwit/Cargo.lock
```diff
@@ -8954,6 +8964,15 @@ dependencies = [
  "zstd-safe 6.0.6",
 ]
 
+[[package]]
+name = "zstd"
```
fulmicoton, line 8968: can you update zstd and lru in quickwit? This PR doubles the dep.

fulmicoton (approved):

quickwit/Cargo.lock
PSeitz: updated lru, but zstd is a nested dependency

PR #3973, opened 2023-10-17
segment partial hits

fulmicoton:
This introduces a SegmentPartialHit to make the work of the collector
lighter.

I measured a 25% performance improvement on real life data.

There is still a lot of headroom left over.


quickwit/quickwit-search/src/collector.rs
```diff
@@ -328,6 +325,29 @@ impl QuickwitSegmentCollector {
...
+    fn into_partial_hit(self, split_id: String, segment_ord: SegmentOrdinal) -> PartialHit {
+        PartialHit {
+            sort_value: self.sort_value.map(|sort_value| SortByValue {
+                sort_value: Some(sort_value),
+            }),
+            sort_value2: self.sort_value2.map(|sort_value| SortByValue {
+                sort_value: Some(sort_value),
+            }),
+            doc_id: self.doc_id,
+            split_id: split_id.clone(),
```
trinity-1686a, line 345: it's taken by value, no need to clone

trinity-1686a (approved):

quickwit/quickwit-search/src/collector.rs
fulmicoton: ah yes!

Issue #3990, opened 2023-10-19
Support compressed HTTP request bodies in REST API


guilload:
Implementation pointers:
- the HTTP server for the REST API is set up [here](https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-serve/src/rest.rs).
- ~~add [this Tower layer](https://docs.rs/tower-http/latest/tower_http/decompression/struct.RequestDecompressionLayer.html) should do the trick~~

unpervertedkid:
@guilload can i take this up?

etolbakov:
@unpervertedkid I assume you had a look at the linked https://github.com/quickwit-oss/quickwit/pull/4003 comment section about warp limitation and will tackle that problem differently.

guilload:
Hello @unpervertedkid,

I saw you opened a PR to add support for decompression directly in `warp`. Super cool and thank you! I don't know how long it'll take for the review/merge/release process to complete. In the meantime, we could use your `warp::Filter` in Quickwit directly. The only missing piece is to dynamically decompress the payload depending on the value of the `Content-Encoding` header (or its absence).

unpervertedkid:
Sure @guilload . Although I might take a week or two pick it up so in the mean time if anyone is willing to pick it up feel free to go ahead.
Refer to [this pr](https://github.com/seanmonstar/warp/pull/1080).

DCjanus:
> A malicious user can compose a request using gzip that has a compression ratio > 1000x, meaning a body of 1MB could expand >1GB. When the server processes these requests, it's very likely to saturate the memory capacity and cause an outage.
-- https://medium.com/@abhinav.ittekot/why-http-request-compression-is-almost-never-supported-5de68067b245

Maybe we should be carefully handle this situation.

fmassot:
Friendly ping @unpervertedkid

unpervertedkid:
@fmassot I think you should assign this to someone else.
My bandwidth is currently pretty limited😥.

PR #4003, opened 2023-10-20
feat: enable request bodies decompression

etolbakov:
### Description

enable request bodies decompression

### How was this PR tested?

this setting hasn't been tested

### Related issue
https://github.com/quickwit-oss/quickwit/issues/3990

### DIFF
--- a/quickwit/quickwit-serve/src/rest.rs
+++ b/quickwit/quickwit-serve/src/rest.rs
@@ -153,7 +153,7 @@ pub(crate) async fn start_rest_server(
                 .compress_when(compression_predicate),
         )
         .layer(cors)
-        .layer(DecompressionLayer::new().gzip(true))
+        .layer(DecompressionLayer::new().gzip(true).zstd(true))
         .service(warp_service);
 
     info!(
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-serve/src/rest.rs
```diff
@@ -152,6 +153,7 @@ pub(crate) async fn start_rest_server(
                 .compress_when(compression_predicate),
         )
         .layer(cors)
+        .layer(DecompressionLayer::new().gzip(true))
```
guilload, line 156: Can we support `zstd` as well?

guilload:
Would you be able to test this? You could follow our getting started tuto and tweak the cURL command:
https://quickwit.io/docs/main-branch/get-started/quickstart.

etolbakov:
Currently  `warp` (web framework) does not handle compressed bodies out of the box.
https://github.com/seanmonstar/warp/issues/1035

guilload:
Yes, my bad. I assumed that `warp` is compatible with most `tower-http` layers, but that's no the case. Thanks for doing the discovery work.

guilload:
@unpervertedkid, the best way to go about this is adding support in `warp`. I've looked into it, it's not too hard.

PR #4011, opened 2023-10-23
Fix the NO_COLOR environment variable interpretation.

fulmicoton:
Before this PR, NO_COLOR had to be true or false.
This PR interprets the environment variable as true for any value that is non empty as described in https://no-color.org/

This fixes the config in our own configuration:
https://github.com/quickwit-oss/quickwit/blob/kzf2f6aacca1cd645a473b7a9ec0707a6efc6e323/docs/log-management/send-logs/using-otel-collector-with-helm.md?plain=1#L75

### DIFF
--- a/quickwit/quickwit-cli/src/main.rs
+++ b/quickwit/quickwit-cli/src/main.rs
@@ -50,7 +50,7 @@ async fn main_impl() -> anyhow::Result<()> {
 
     let app = build_cli().about(about_text).version(version_text);
     let matches = app.get_matches();
-    let ansi = !matches.get_flag("no-color");
+    let ansi_colors = !matches.get_flag("no-color");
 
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-cli/src/main.rs
```diff
@@ -704,4 +704,34 @@ mod tests {
...
+        {
+            // empty string is false.
+            std::env::set_var("NO_COLOR", "");
+            let app = build_cli().no_binary_name(true);
+            let matches = app.try_get_matches_from(["run"]).unwrap();
+            let no_color = matches.get_flag("no-color");
+            assert!(!no_color);
+        }
+        {
+            // empty string is false.
```
fmassot, line 727:
```suggestion
```

fmassot (approved):

PR #4018, opened 2023-10-24
Added regression test for wildcard in query string queries

fulmicoton:


fmassot:
Closing as a test has already been added in #4059

Issue #3998, opened 2023-10-20
Apply query simplification optimization to find_trace


fulmicoton:
Finding traces could really beneift from the optimization we introduced in https://github.com/quickwit-oss/quickwit/pull/3849

Namingly, if we find out that the split min/max timestamp will never fit the bill, there is no need to run the find trace aggregator.
We probably want to do that in combination with
https://github.com/quickwit-oss/quickwit/issues/3997

PR #4019, opened 2023-10-24
plug exact count optimization

### LINKED ISSUE #3997: Plug the exact count optimization
We recently added an optimization for top K sorted by time.
We process splits in the most promising order in order.

Once we have saturated the order, if a the split metadata informs us that we cannot hope for better results, we currently set max_hits to 0 which has the great side-effect of removing the need to warmup the timestamp fields.

We however still perform search, because to count the documents.

It would be interesting to add an enum property to the search proto to make it possible to remove the need to process the split entirely, (in the absenve of any aggregations of course)

# Disclaimer

Let's not break backward compatibility. Hence the idea of using an enum, which would give us more control on what the default value should be in the absence of a value.

trinity-1686a:
### Description

fix #3997
does not impact #3998 as it would need more work. Tracing uses aggregations, which don't profit from that optimization yet

### How was this PR tested?

created an index with single split documents, made a few query and verified `num_hits` was below what's expected when `count_all=false`, and normal when it's true or absent

### DIFF
--- a/quickwit/quickwit-proto/src/codegen/quickwit/quickwit.search.rs
+++ b/quickwit/quickwit-proto/src/codegen/quickwit/quickwit.search.rs
@@ -107,6 +107,8 @@ pub struct SearchRequest {
     /// If split_id is empty, no comparison with _shard_doc should be done
     #[prost(message, optional, tag = "16")]
     pub search_after: ::core::option::Option<PartialHit>,
+    #[prost(enumeration = "CountHits", tag = "17")]
+    pub count_hits: i32,
 }
 #[derive(Serialize, Deserialize, utoipa::ToSchema)]
 #[derive(Eq, Hash)]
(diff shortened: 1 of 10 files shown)


quickwit/quickwit-serve/src/search_api/rest_handler.rs
```diff
@@ -217,6 +217,12 @@ pub struct SearchRequestQueryString {
     #[serde(skip_serializing_if = "SortBy::is_empty")]
     #[param(value_type = String)]
     pub sort_by: SortBy,
+    #[serde(default = "default_as_true")]
+    pub count_all: bool,
```
fulmicoton, line 221: can we use the CountHits enum here?

quickwit/quickwit-serve/src/search_api/rest_handler.rs
```diff
@@ -565,6 +578,7 @@ mod tests {
                 format: BodyFormat::default(),
                 sort_by: SortBy::default(),
                 aggs: Some(json!({"range":[]})),
+                count_all: true,
```
fulmicoton, line 581: can we add a mini unit tests for the count_all query string parameter?

quickwit/quickwit-serve/src/elastic_search_api/rest_handler.rs
```diff
@@ -156,6 +157,13 @@ fn build_request_for_es_api(
 
     let max_hits = search_params.size.or(search_body.size).unwrap_or(10);
     let start_offset = search_params.from.or(search_body.from).unwrap_or(0);
+    let count_hits = match search_params.track_total_hits {
+        None => CountHits::Underestimate,
+        Some(TrackTotalHits::Track(false)) => CountHits::Underestimate,
+        Some(TrackTotalHits::Count(count)) if count <= max_hits as i64 => CountHits::Underestimate,
+        _ => CountHits::CountAll,
```
fulmicoton, line 164:
```suggestion
        Some(TrackTotalHits::Track(true)) => CountHits::CountAll,
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -549,7 +549,8 @@ pub async fn leaf_search(
 
     // In the future this should become `request.aggregation_request.is_some() ||
     // request.exact_count == true`
-    let run_all_splits = true;
+    let run_all_splits = request.aggregation_request.is_some()
+        || request.count_hits == i32::from(CountHits::CountAll);
```
fulmicoton, line 553: I think prost creates an accessor that gives you the enum. Something like `get_count_hits()` or something

quickwit/quickwit-proto/src/codegen/quickwit/quickwit.search.rs
```diff
@@ -498,6 +500,37 @@ pub struct LeafSearchStreamResponse {
...
+    pub fn as_str_name(&self) -> &'static str {
+        match self {
+            CountHits::CountAll => "COUNT_ALL",
+            CountHits::Underestimate => "UNDERESTIMATE",
+        }
+    }
+    /// Creates an enum from field names used in the ProtoBuf definition.
+    pub fn from_str_name(value: &str) -> ::core::option::Option<Self> {
+        match value {
+            "COUNT_ALL" => Some(Self::CountAll),
```
fulmicoton, line 524: Why is this useful?

quickwit/quickwit-proto/src/codegen/quickwit/quickwit.search.rs
```diff
@@ -498,6 +500,37 @@ pub struct LeafSearchStreamResponse {
...
+    Underestimate = 1,
+}
+impl CountHits {
+    /// String value of the enum field names used in the ProtoBuf definition.
+    ///
+    /// The values are not transformed in any way and thus are considered stable
+    /// (if the ProtoBuf definition does not change) and safe for programmatic use.
+    pub fn as_str_name(&self) -> &'static str {
+        match self {
+            CountHits::CountAll => "COUNT_ALL",
```
fulmicoton, line 517: Why is this useful?

fulmicoton (approved):

quickwit/quickwit-search/src/leaf.rs
guilload: `count_hits()`

Issue #3995, opened 2023-10-19
Introduce a builder for the `Codegen` object


guilload:
In `quickwit-codegen`:

Before:
```rust
Codegen::run_with_config(
        &[
            "protos/quickwit/ingester.proto",
            "protos/quickwit/router.proto",
        ],
        "src/codegen/quickwit",
        "crate::ingest::IngestV2Result",
        "crate::ingest::IngestV2Error",
        false,
        &["protos"],
        prost_config,
    )
    .unwrap();
```

After:
```rust
Codegen::builder()
  .with_protos()
  .with_includes()
  .with_output_dir()
  .with_prost_config()
  .run()
  .unwrap();
```

jmintb:
Can I work on this? :)

guilload:
Yup :)

guilload:
Closed via $4028.

PR #4028, opened 2023-10-25
Introduce a builder for Codegen

jmintb:
Issue: #3995

### Description

I have implemented a builder for the `Codegen` following the example described in #3995. I also replaced existing calls to `Codegen::run` and `Codegen::run_with_config` with the builder.

I removed `Codegen::run` as it is never called and I am unsure if it even makes sense to keep `run_with_config` separate from the builder. What do you think?

There are conversions from &[&str] to Vec<String> and back again in the builder which I don't like. However, I think overall it is the simplest implementation in this case as we avoid lifetimes and trait bounds. Apart from just merging `run_with_config` with the builder.

Upon writing this I realized that I forgot to update the codegen readme. I will update the readme once the implementation is approved (to avoid having to keep updating it).

Let me  know what you think :)

### How was this PR tested?

Tested that everything still builds and existing tests pass. I have a few failing tests on my machine even without these changes so those will have to be tested during CI.

### DIFF
--- a/quickwit/quickwit-codegen/src/codegen.rs
+++ b/quickwit/quickwit-codegen/src/codegen.rs
@@ -30,27 +30,6 @@ use crate::ProstConfig;
 pub struct Codegen;
 
 impl Codegen {
-    pub fn run(
-        protos: &[&str],
-        out_dir: &str,
-        result_type_path: &str,
-        error_type_path: &str,
-        generate_extra_service_methods: bool,
-        generate_prom_labels_for_requests: bool,
-        includes: &[&str],
-    ) -> anyhow::Result<()> {
-        Self::run_with_config(
-            protos,
-            out_dir,
-            result_type_path,
-            error_type_path,
-            generate_extra_service_methods,
-            generate_prom_labels_for_requests,
-            includes,
-            ProstConfig::default(),
-        )
-    }
-
     #[allow(clippy::too_many_arguments)]
     pub fn run_with_config(
         protos: &[&str],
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-codegen/src/codegen.rs
```diff
@@ -89,6 +89,85 @@ impl Codegen {
...
+    prost_config: ProstConfig,
+    result_type_path: String,
+    error_type_path: String,
+    generate_extra_service_methods: bool,
+    generate_prom_label_for_requests: bool,
+}
+
+impl CodegenBuilder {
+    pub fn with_protos(mut self, protos: &[&str]) -> Self {
+        self.protos = protos.iter().map(|p| p.to_string()).collect();
```
guilload, line 112: nit: we like long variable names :)

```suggestion
        self.protos = protos.iter().map(|proto| proto.to_string()).collect();
```

quickwit/quickwit-codegen/src/codegen.rs
```diff
@@ -89,6 +89,85 @@ impl Codegen {
...
+        self
+    }
+
+    pub fn with_includes(mut self, includes: &[&str]) -> Self {
+        self.includes = includes.iter().map(|i| i.to_string()).collect();
```
guilload, line 117:
```suggestion
        self.includes = includes.iter().map(|include| include.to_string()).collect();
```

guilload (approved):
Looking very good!

guilload:
I agree with you; let's not worry too much about the conversions back and forth between `&[&str]` and `Vec`. You can modify the signature of `run_with_config` if that simplifies a few things, but don't sweat it.

Thanks for the contribution, @jmintb!

quickwit/quickwit-codegen/src/codegen.rs
jmintb: Then long variable names you shall have :)

PR #4029, opened 2023-10-26
Avoiding base64 translation when serializing Traces with postcard

fulmicoton:
Avoiding base64 translation when serializing Traces with postcard

This PR also adds unit tests/proptest checking the serializing of
Vec<Span> with postcard.

### DIFF
--- a/quickwit/quickwit-search/src/find_trace_ids_collector.rs
+++ b/quickwit/quickwit-search/src/find_trace_ids_collector.rs
@@ -19,7 +19,6 @@
 
 use std::cmp::{Ord, Ordering};
 use std::collections::HashSet;
-use std::hash::{Hash, Hasher};
 
 use fnv::{FnvHashMap, FnvHashSet};
 use itertools::Itertools;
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-search/src/find_trace_ids_collector.rs
```diff
@@ -185,23 +183,24 @@ impl Collector for FindTraceIdsCollector {
 }
 
 fn merge_segment_fruits(mut segment_fruits: Vec<Vec<Span>>, num_traces: usize) -> Vec<Span> {
+    // Spans are ordered in reverse order of their timestamp.
     for segment_fruit in &mut segment_fruits {
         segment_fruit.sort_unstable()
     }
-    let mut trace_ids = Vec::with_capacity(num_traces);
+    let mut spans: Vec<Span> = Vec::with_capacity(num_traces);
```
fulmicoton, line 190: this is a just a rename.

quickwit/quickwit-search/src/find_trace_ids_collector.rs
```diff
@@ -48,12 +47,6 @@ impl Span {
     }
 }
 
-impl Hash for Span {
-    fn hash<H: Hasher>(&self, state: &mut H) {
```
fulmicoton, line 52: It is not consistent with ord/eq.
it was not used so removed it.

PSeitz (approved):

Issue #4025, opened 2023-10-25
Improve behavior on corrupted checkpoint.


fulmicoton:
If a checkpoint contains an invalid position (for instance not u64) for ingest,
we currently panic.

Ideally we should:
- log an error
- repair the checkpoint by removing the corrupted partition
- start indexing the faulty partition from the beginning.

guilload:
Yes, you can.

Logging an error and skipping the partition is enough. I don't think self-clean-up is helpful because it's likely that the issue comes from either a bug in the source or a user manually editing a checkpoint. In the first case, the bug will reoccur, and self-cleanup will be less helpful. In the second, I'd rather have users clean up their mess themselves and decide from which point they want to resume indexing the partition. Always restarting from the beginning will yield duplicates.

jmintb:
Perfect, do you know whereabouts in the codebase the panic(s) occurs? Sounds like it is in `quickwit-ingest`.

guilload:
if you `rg -e 'expect\(.*offset.*'` in `quickwit/quickwit-indexing/src/source` you should find them all.

jmintb:
Are there any existing tests or test data that would simulate this scenario?

PR #4039, opened 2023-10-27
Improve error handling for corrupt checkpoints

jmintb:
Issue: #4025

### Description

This PR improves handling of corrupt checkpoints by logging the error and skipping the corrupt partition instead of panicking.

I am not sure if this is what you had in mind so this is a draft PR. 

Questions: 

1. Is there any bookkeeping I am missing when skipping a partition?
2. What kind of information would you like to see in the error message? maybe the partition ID?
3. Do you have any thoughts on testing this?

Let me know what you think and I will apply the final version to the remaining panic cases :)

### How was this PR tested?

Not tested yet :(

### DIFF
--- a/quickwit/quickwit-indexing/src/source/kafka_source.rs
+++ b/quickwit/quickwit-indexing/src/source/kafka_source.rs
@@ -45,7 +45,7 @@ use serde_json::{json, Value as JsonValue};
 use tokio::sync::{mpsc, watch};
 use tokio::task::{spawn_blocking, JoinHandle};
 use tokio::time;
-use tracing::{debug, info, warn};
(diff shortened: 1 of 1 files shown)


guilload:
> 1. Is there any bookkeeping I am missing when skipping a partition?
> 2. What kind of information would you like to see in the error message? maybe the partition ID?
> 3. Do you have any thoughts on testing this?

1. No, I don't think so.
2. index ID, source ID, partition ID, value of invalid offset
3. This is unit-testable; you can update `test_kafka_source_process_assign_partitions` and `test_kafka_source_suggest_truncate`.

There are also other `expect`s to remove on other sources.

PR #4043, opened 2023-10-28
Bugfix

fulmicoton:
- Update the control plane model on toggle source.
- Delete source from index config in control plane model on delete source.

### DIFF
--- a/quickwit/quickwit-control-plane/src/control_plane.rs
+++ b/quickwit/quickwit-control-plane/src/control_plane.rs
@@ -327,15 +327,11 @@ impl Handler<ToggleSourceRequest> for ControlPlane {
 
         let has_changed = self.model.toggle_source(&index_uid, &source_id, enable)?;
 
-        // TODO update the internal view.
-        // TODO: Refine the event. Notify index will have the effect to reload the entire state from
-        // the metastore. We should update the state of the control plane.
         if has_changed {
             self.indexing_scheduler.on_index_change(&self.model).await?;
         }
 
-        let response = EmptyResponse {};
-        Ok(Ok(response))
+        Ok(Ok(EmptyResponse {}))
     }
 }
 
@@ -565,6 +561,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_control_plane_toggle_source() {
+        quickwit_common::setup_logging_for_tests();
         let universe = Universe::with_accelerated_time();
 
         let cluster_id = "test-cluster".to_string();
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -317,14 +317,22 @@ impl Handler<ToggleSourceRequest> for ControlPlane {
...
         };
 
+        let has_changed = self.model.toggle_source(&index_uid, &source_id, enable)?;
+
         // TODO update the internal view.
         // TODO: Refine the event. Notify index will have the effect to reload the entire state from
         // the metastore. We should update the state of the control plane.
-        self.indexing_scheduler.on_index_change(&self.model).await?;
+        if has_changed {
+            self.indexing_scheduler.on_index_change(&self.model).await?;
```
guilload, line 334: Do we want to return an error if the `toggle_source` call has succeeded but the scheduling failed? I have left a todo for the same thing below.

quickwit/quickwit-control-plane/src/control_plane_model.rs
```diff
@@ -245,7 +246,34 @@ impl ControlPlaneModel {
...
+    /// Returns `true` if the source status has changed, `false` otherwise.
+    /// Returns an error if the source could not be found.
+    pub(crate) fn toggle_source(
+        &mut self,
+        index_uid: &IndexUid,
+        source_id: &SourceId,
+        enable: bool,
+    ) -> anyhow::Result<bool> {
+        let Some(index_model) = self.index_table.get_mut(index_uid) else {
+            bail!("Index `{}` not found.", index_uid);
```
guilload, line 270:
```suggestion
            bail!("index `{index_uid}` not found");
```

quickwit/quickwit-control-plane/src/control_plane_model.rs
```diff
@@ -245,7 +246,34 @@ impl ControlPlaneModel {
...
+        };
+        let Some(source_config) = index_model.sources.get_mut(source_id) else {
+            bail!("Source `{}` not found.", source_id);
```
guilload, line 273:
```suggestion
            bail!("source `{source_id}` not found.");
```

quickwit/quickwit-control-plane/src/control_plane_model.rs
```diff
@@ -245,7 +246,34 @@ impl ControlPlaneModel {
...
+        };
+        source_config.enabled = enable;
+        Ok(enable)
```
guilload, line 276: This does not return whether the source has changed but rather whether it was enabled or not.

guilload (approved):

quickwit/quickwit-control-plane/src/control_plane_model.rs
fulmicoton: that's a bug indeed.

quickwit/quickwit-control-plane/src/control_plane.rs
fulmicoton: Good point. 
Actually this does not need to be async and does not return an error.
Addressing htat in a different PR.

PR #4047, opened 2023-10-28
Update query to order by split id when offset is provided

kamalesh0406:
### Description

The change adds an order by clause to the sql query when the split id is provided. This fixes the bug reported by #4046 .

### How was this PR tested?

An unit test was to added to check if the correct query was being returned.


guilload (approved):

guilload:
Thanks @kamalesh0406 !

PR #4048, opened 2023-10-29
Bugfix #4042 - An index ptn not matching any index is ok.

### LINKED ISSUE #4042: Elasticsearch compatible search endpoint should return an empty set when no matching index is found
Return empty hits arrays if index search patterns do not match any indexes

* If none of the index patterns match an index. return an empty hits array.
* If some of the patterns match an index, return results for the indexes that are found.

**Additional context**
Currently if an index pattern doesn't match any index in quickwit it will fail the entire request. This is difficult to handle when using a CSV of globbed index patterns.  In most cases, we use globs because we are unsure if certain indexes exists or not. In the current state, it will return a 404 where as elasticsearch will return a 200 and an empty hits array

fulmicoton:
Fixes #4042

Tested with rest integration tests (checked against elasticsearch)

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
+++ b/quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
@@ -60,10 +60,9 @@ use crate::metastore::postgresql_model::{PgDeleteTask, PgIndex, PgSplit, Splits,
 use crate::metastore::{instrument_metastore, FilterRange, PublishSplitsRequestExt};
 use crate::{
     AddSourceRequestExt, CreateIndexRequestExt, IndexMetadata, IndexMetadataResponseExt,
-    ListIndexesMetadataRequestExt, ListIndexesMetadataResponseExt, ListIndexesQuery,
-    ListSplitsQuery, ListSplitsRequestExt, ListSplitsResponseExt, MetastoreFactory,
-    MetastoreResolverError, MetastoreServiceExt, Split, SplitMaturity, SplitMetadata, SplitState,
-    StageSplitsRequestExt,
+    ListIndexesMetadataResponseExt, ListSplitsQuery, ListSplitsRequestExt, ListSplitsResponseExt,
+    MetastoreFactory, MetastoreResolverError, MetastoreServiceExt, Split, SplitMaturity,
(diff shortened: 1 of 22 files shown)


quickwit/quickwit-proto/src/codegen/quickwit/quickwit.metastore.rs
```diff
@@ -20,8 +20,8 @@ pub struct CreateIndexResponse {
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct ListIndexesMetadataRequest {
-    #[prost(string, tag = "1")]
-    pub query_json: ::prost::alloc::string::String,
+    #[prost(string, repeated, tag = "2")]
+    pub index_ptns: ::prost::alloc::vec::Vec<::prost::alloc::string::String>,
```
fmassot, line 24: I would replace `index_ptns` by `index_id_patterns`

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1388,8 +1380,15 @@ fn tags_filter_expression_helper(tags: &TagFilterAst) -> Cond {
...
-fn build_index_id_patterns_sql_query(index_id_patterns: Vec<String>) -> anyhow::Result<String> {
-    assert!(!index_id_patterns.is_empty());
+fn build_index_id_patterns_sql_query(index_id_patterns: &[String]) -> anyhow::Result<String> {
+    if index_id_patterns.is_empty() {
+        anyhow::bail!("The list of index id patterns may not be empty.");
+    }
+    if index_id_patterns == ["*"] {
+        return Ok("SELECT * FROM indexes".to_string());
+    }
+    // TODO Fix me.
```
fmassot, line 1390: Is the `TODO Fix me.` and the following comment are outdated? 
The check is handled in the `root_search` function and not at the metastore level.

fmassot (approved):
Can you have a look at the 2 minor comments?

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
fulmicoton: yes thx

PR #4049, opened 2023-10-29
check yield_after_each_message

PSeitz:
### Description

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-actors/src/spawn_builder.rs
+++ b/quickwit/quickwit-actors/src/spawn_builder.rs
@@ -284,8 +284,9 @@ impl<A: Actor> ActorExecutionEnv<A> {
             while let Some(envelope) = try_recv_envelope(&mut self.inbox, &self.ctx) {
                 self.process_one_message(envelope).await?;
             }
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-actors/src/spawn_builder.rs
```diff
@@ -284,8 +284,9 @@ impl<A: Actor> ActorExecutionEnv<A> {
             while let Some(envelope) = try_recv_envelope(&mut self.inbox, &self.ctx) {
                 self.process_one_message(envelope).await?;
             }
-            self.ctx.yield_now().await;
-
+            if self.actor.get_mut().yield_after_each_message() {
```
fulmicoton, line 287: What is the motivation here? It seems like a good idea to yield here.

PSeitz: This fixes the issue that Indexer and DocProcessor run on one thread

fulmicoton: That's not a good sign :-/

PR #4050, opened 2023-10-30
update tantivy

PSeitz:
### Description

### How was this PR tested?

### DIFF
--- a/quickwit/Cargo.lock
+++ b/quickwit/Cargo.lock
@@ -943,6 +943,14 @@ dependencies = [
 
 [[package]]
 name = "bitpacking"
+version = "0.8.3"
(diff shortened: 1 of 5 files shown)


quickwit/Cargo.lock
```diff
@@ -941,6 +941,14 @@ dependencies = [
  "serde",
 ]
 
+[[package]]
+name = "bitpacking"
+version = "0.8.3"
```
fulmicoton, line 946: :-/ how did we end up with two deps on bitpacking?

PSeitz: Coming from this change https://github.com/quickwit-oss/tantivy/commit/0d4589219bcf8b4f10326e599c23ac102d28e42b

I think only temporary, until a new bitpacking release is done (I don't have access)

PR #4054, opened 2023-10-31
Ensures the load of pipelines does not exceed 80%.

### LINKED ISSUE #4010: Break indexing of ingest v2 into several peipelines whenever necessary.
Follow up of https://github.com/quickwit-oss/quickwit/pull/3991

Right now all shards of a given source end up in the same pipeline.
This could exceed the pipeline capacity.

We want to break the shard into several sources, and ideally, respect the repartiton of any previously existing physical plan.

fulmicoton:
The allocation attempts to match the previous
plan as much as possible.

It also includes some hysteresis effect to avoid
flip flapping between two N and N+1 pipelines.

Closes #4010

### DIFF
--- a/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
+++ b/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
@@ -29,6 +29,7 @@ use quickwit_proto::types::{IndexUid, ShardId};
 pub use scheduling_logic_model::Load;
 use scheduling_logic_model::{IndexerOrd, SourceOrd};
 use tracing::error;
+use tracing::log::warn;
 
 use crate::indexing_plan::PhysicalIndexingPlan;
 use crate::indexing_scheduler::scheduling::scheduling_logic_model::{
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
```diff
@@ -238,8 +255,99 @@ pub enum SourceToScheduleType {
...
+) -> Vec<IndexingTask> {
+    let num_shards = shard_ids.len() as u32;
+    if num_shards == 0 {
+        return Vec::new();
+    }
+    let max_num_shards_per_pipeline = MAX_LOAD_PER_PIPELINE / load_per_shard;
+
+    // We compute the number of pipelines we will create, cooking in some hysteresis effect here.
+    // We have two different threshold to increase and to decrease the number of pipelines.
+    let min_num_pipelines: u32 =
```
fmassot, line 272: Can we protect the code from a division by 0? It's hard to check in the code if `load_per_shard` will always stay under `MAX_LOAD_PER_PIPELINE` (for a Kafka source, it seems possible to have a load higher than `MAX_LOAD_PER_PIPELINE`).

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
```diff
@@ -238,8 +255,99 @@ pub enum SourceToScheduleType {
...
+            }
+        } else {
+            unassigned_shard_ids.push(shard);
+        }
+    }
+
+    // If needed, let's remove some pipelines. We just remove the pipelines that have
+    // the least number of shards.
+    pipelines.sort_by_key(|shards| std::cmp::Reverse(shards.len()));
+    for removed_pipelines in pipelines.drain(num_pipelines as usize..) {
```
fmassot, line 315: nitpick: replace `removed_pipelines` by `removed_shards` or `removed_pipeline_shards`

fmassot (approved):
Approved. One comment is important to take into account to avoid a division by 0.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
fulmicoton: very good point. FYI in the design this is loosely enforced by forcing a shard to deliver at most `5MB/s`...

But even if the assumption that we can always index  faster than `5MB/s`, we will have to catch with some shards...

PR #4059, opened 2023-10-31
implement  trailing wildcard support

### LINKED ISSUE #3982: Add support for trailing wildcard in ES query_string queries
Elasticsearch supports wildcards in query_string queries.

For the moment, we want to only support wildcards at the end of a term, and return an error for internal wildcards.

`hello mr bon*` should match `Hello Mr Bond`.

The term containing the wildcard should go through a reduced analyzer:
an analyzer from whcih we have only kept the tokenfilters affecting casing.

- `hello mr bon*` should match `Hello mr bond` provided the field analyzer contains a lowercaser
-`hello mr Bon*` should match `hello mr bond` if the field analyzer contains a lowercaser
-`hello mr Bon*` should match `hello mr Bond` if the field analyzer does not contains lowercaser
-`hello mr bon*` should not match `hello mr Bond` if the field analyzer does not contains lowercaser

trinity-1686a:
### Description

fix #3982

### How was this PR tested?

integration test from #4018 (slightly updated as it was already passing without the new feature)


quickwit/quickwit-query/src/query_ast/wildcard_query.rs
```diff
@@ -0,0 +1,186 @@
...
+                let tokenizer_name = text_field_indexing.tokenizer();
+                let analyzer_name = if tokenizer_manager
+                    .get_does_lowercasing(tokenizer_name)
+                    .with_context(|| {
+                        format!("no tokenizer named `{}` is registered", tokenizer_name)
+                    })? {
+                    "lowercase"
+                } else {
+                    "raw"
+                };
```
fmassot, line 86: the API of the tokenizer manager is not very nice

quickwit/quickwit-query/src/tokenizers/tokenizer_manager.rs
```diff
@@ -0,0 +1,74 @@
...
+use std::collections::HashMap;
+use std::sync::{Arc, RwLock};
+
+use tantivy::tokenizer::{TextAnalyzer, TokenizerManager as TantivyTokenizerManager};
+
+#[derive(Clone)]
+pub struct TokenizerManager {
+    inner: TantivyTokenizerManager,
+    is_lowercaser: Arc<RwLock<HashMap<String, bool>>>,
+}
```
fmassot, line 29: I'm not very fond of the `TokenizerManager`. It would be nice to enforce by construction that we have the right `normalizers` already in the hashmap and also provide nicer accessors to normalizers for a given tokenizer.

I would try something like this. 

```rust
pub struct AnalyzerManager {
    tokenizers: Arc<RwLock<HashMap<String, TextAnalyzer>>>,
    tokenizer_to_lower_caser_map: Arc<RwLock<HashMap<String, bool>>>,
}


impl QuickwitTokenizerManager {
    /// Creates an empty tokenizer manager.
    pub fn new() -> Self {
        let tokenizers = HashMap::new();
        let raw_tokenizer = TextAnalyzer::builder(RawTokenizer::default())
            .filter(RemoveLongFilter::limit(DEFAULT_REMOVE_TOKEN_LENGTH))
            .build();
        tokenizers.insert(RAW_TOKENIZER_NAME, raw_tokenizer);
        let lower_case_tokenizer = TextAnalyzer::builder(RawTokenizer::default())
            .filter(LowerCaser)
            .filter(RemoveLongFilter::limit(DEFAULT_REMOVE_TOKEN_LENGTH))
            .build();
        tokenizers.insert(LOWERCASE_TOKENIZER_NAME, lower_case_tokenizer);
        let normalizer_to_tokenizer_name = HashMap::new();
        Self {
            tokenizers: Arc::new(RwLock::new(tokenizers)),
            normalizer_to_tokenizer_map: Arc::new(RwLock::new(normalizer_to_tokenizer_name)),
        }
    }

    /// Registers a new tokenizer associated with a given name.
    pub fn register<T, U>(&self, tokenizer_name: &str, tokenizer: T, does_lowercasing: bool)
    where TextAnalyzer: From<T> {
       todo!()
    }
 
    pub fn get_tokenizer(&self, tokenizer_name: &str) -> Option<TextAnalyzer> {
        todo!()
    }

    pub fn get_normalizer(&self, tokenizer_name: &str) -> Option<TextAnalyzer>{
        todo!()
    }
}


```

quickwit/quickwit-query/src/query_ast/user_input_query.rs
```diff
@@ -216,24 +216,33 @@ fn convert_user_input_literal(
         mode,
         zero_terms_query: crate::MatchAllOrNone::MatchNone,
     };
+    // TODO should probably check for escaped * and ?
```
fmassot, line 219: Yes we need to check if * or ? are escaped.

quickwit/quickwit-query/src/query_ast/wildcard_query.rs
```diff
@@ -0,0 +1,186 @@
...
+                    tokens.push(json_term_writer.term().clone());
+                });
+                let term = tokens
+                    .pop()
+                    .with_context(|| "wildcard query generated no term")?;
+                if !tokens.is_empty() {
+                    return Err(InvalidQuery::Other(anyhow::Error::msg(
+                        "wildcard query generated more than one term",
+                    )));
+                }
```
fmassot, line 148: could be nice to isolate a check function. 

Also the errors are a bit misleading, `tokenzing wildcard query generated no term`? same for the second error message

quickwit/quickwit-query/src/query_ast/wildcard_query.rs
```diff
@@ -0,0 +1,186 @@
...
+
+impl BuildTantivyAst for WildcardQuery {
+    fn build_tantivy_ast_impl(
+        &self,
+        schema: &TantivySchema,
+        tokenizer_manager: &TokenizerManager,
+        _search_fields: &[String],
+        _with_validation: bool,
+    ) -> Result<TantivyQueryAst, InvalidQuery> {
+        // TODO handle escaped wildcard and questionmarks?
```
fmassot, line 166: does elasticsearch handle escaped wildcard and `?` ?

fmassot (commented):
Can you have a look at my comments?

quickwit/quickwit-query/src/query_ast/wildcard_query.rs
trinity-1686a: I checked, and it handle them as being literals, not wildcards

quickwit/quickwit-query/src/query_ast/full_text_query.rs
```diff
@@ -130,10 +131,7 @@ impl FullTextParams {
                     .collect();
                 Ok(TantivyBoolQuery::build_clause(operator, leaf_queries).into())
             }
-            FullTextMode::BoolPrefix {
-                operator,
-                max_expansions,
```
fmassot, line 135: I'm a bit worried by the removal of `max_expansions`. It seems that the parameter exists in the [ES API](https://github.com/elastic/elasticsearch/blob/8.11/server/src/main/java/org/elasticsearch/index/query/MatchBoolPrefixQueryBuilder.java)

trinity-1686a: that's interesting, it's not documented in https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-bool-prefix-query.html :confused:

fmassot: I think we should keep the `max_expansions` then.

quickwit/quickwit-query/src/query_ast/wildcard_query.rs
fmassot: If it's not too much work, I would handle that in this PR.

quickwit/quickwit-query/src/elastic_query_dsl/match_bool_prefix.rs
```diff
@@ -40,7 +40,7 @@ impl ConvertableToQueryAst for MatchBoolPrefixQuery {
             tokenizer: None,
             mode: crate::query_ast::FullTextMode::BoolPrefix {
                 operator: self.params.operator,
-                max_expansions: default_max_expansions(),
+                max_expansions: u32::MAX,
```
fmassot, line 43: Is it in line with the default behavior in ES?

trinity-1686a: as far as I can tell, `match_bool_prefix` has a default max_expansions of 50, but does not honor it. For instance if I add 52 documents with a term from `xaa` to `xbz`, I would expect only 50 documents to be returned in a `match_bool_prefix` for `x`, but they all are (in fact, they all are even if I add `"max_expansions": 1` to the query

fmassot: can you add a comment about that so that the next reader won't be lost?

quickwit/rest-api-tests/scenarii/es_compatibility/0005-query_string_query.yaml
```diff
@@ -151,3 +151,28 @@ json:
...
+      default_field: payload.description
+      lenient: true
+      query: "Jour* AND unix"
+expected:
+  hits:
+    total:
+      value: 2
+---
+# trailing wildcard
+json:
```
fmassot, line 168: Is it possible to add a test that returns more than 50 hits so that we can see the max_expansions limit?

fmassot (approved):
Can you have a look at my comments and add a test on the escaping?

PR #4061, opened 2023-10-31
Handle index/source not found and timeout/transport errors

guilload:
### Description
- Split `GetOrCreateOpenShardsReponse` into `successes` and `failures` to keep track of indexes/sources not found
- Remove leaders from ingester pool on transport error

### How was this PR tested?
- Added unit tests
- Tested locally

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
@@ -19,7 +19,10 @@
 
 use std::collections::HashMap;
 
-use quickwit_proto::ingest::ingester::{PersistFailure, PersistSuccess};
+use quickwit_proto::control_plane::{
+    GetOrCreateOpenShardsFailure, GetOrCreateOpenShardsFailureReason,
+};
+use quickwit_proto::ingest::ingester::{PersistFailure, PersistFailureReason, PersistSuccess};
 use quickwit_proto::ingest::router::{
     IngestFailure, IngestFailureReason, IngestResponseV2, IngestSubrequest, IngestSuccess,
 };
@@ -29,14 +32,19 @@ use tracing::warn;
 
 /// A helper struct for managing the state of the subrequests of an ingest request during multiple
 /// persist attempts.
+#[derive(Default)]
 pub(super) struct IngestWorkbench {
-    subworkbenches: HashMap<SubrequestId, IngestSubworkbench>,
-    num_successes: usize,
+    pub subworkbenches: HashMap<SubrequestId, IngestSubworkbench>,
+    pub num_successes: usize,
+    /// The number of batch persist attempts. This is not sum of the number of attempts for each
+    /// subrequest.
+    pub num_attempts: usize,
+    pub max_num_attempts: usize,
 }
 
 impl IngestWorkbench {
-    pub fn new(subrequests: Vec<IngestSubrequest>) -> Self {
-        let subworkbenches = subrequests
+    pub fn new(ingest_subrequests: Vec<IngestSubrequest>, max_num_attempts: usize) -> Self {
+        let subworkbenches: HashMap<SubrequestId, IngestSubworkbench> = ingest_subrequests
             .into_iter()
             .map(|subrequest| {
                 (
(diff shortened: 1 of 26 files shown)


quickwit/quickwit-proto/protos/quickwit/control_plane.proto
```diff
@@ -80,13 +81,27 @@ message ClosedShards {
...
-    repeated quickwit.ingest.Shard open_shards = 3;
+message GetOrCreateOpenShardsSuccess {
+  uint32 subrequest_id = 1;
+  string index_uid = 2;
+  string source_id = 3;
+  repeated quickwit.ingest.Shard open_shards = 4;
+}
+
+enum GetOrCreateOpenShardsFailureReason {
+  GET_OR_CREATE_OPEN_SHARDS_FAILURE_REASON_UNSPECIFIED = 0;
```
fulmicoton, line 97: Hmm. Repeating the prefix might not be useful here? 
When dealing with enum in rust, we rarely `use GetOrCreateOpneShardsFailureReason::*;`
What do you think?

quickwit/quickwit-proto/protos/quickwit/metastore.proto
```diff
@@ -263,22 +263,24 @@ message OpenShardsRequest {
 }
 
 message OpenShardsSubrequest {
-    string index_uid = 1;
-    string source_id = 2;
-    string leader_id = 3;
-    optional string follower_id = 4;
-    uint64 next_shard_id = 5;
+  uint32 subrequest_id = 1;
```
fulmicoton, line 266: weird indentation?

quickwit/quickwit-proto/protos/quickwit/control_plane.proto
fulmicoton: on further thoughts that's specific to rust and your naming is the right one.

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -29,14 +32,17 @@ use tracing::warn;
...
 /// A helper struct for managing the state of the subrequests of an ingest request during multiple
 /// persist attempts.
+#[derive(Default)]
 pub(super) struct IngestWorkbench {
-    subworkbenches: HashMap<SubrequestId, IngestSubworkbench>,
-    num_successes: usize,
+    pub subworkbenches: HashMap<SubrequestId, IngestSubworkbench>,
+    pub num_successes: usize,
+    pub num_attempts: usize,
+    pub max_num_attempts: usize,
```
fulmicoton, line 40: Can we add a comment explaining this si an overall number of attempts. Not a per sub request number.

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -48,22 +54,77 @@ impl IngestWorkbench {
...
+        else {
+            warn!(
+                "could not find subrequest `{}` in workbench",
+                open_shards_failure.subrequest_id
+            );
+            return;
+        };
+        subworkbench.num_attempts += 1;
+
+        let last_failure_opt = match open_shards_failure.reason() {
```
fulmicoton, line 109:
```suggestion
        // Some failure are not retryable. In that case we record them
        // as the last failure reason and the associated subrequest will be ignore by the retry process.
        let last_failure_opt = match open_shards_failure.reason() {
```

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -48,22 +54,77 @@ impl IngestWorkbench {
...
+            }
+            GetOrCreateOpenShardsFailureReason::SourceNotFound => {
+                Some(SubworkbenchFailure::SourceNotFound)
+            }
+            GetOrCreateOpenShardsFailureReason::Unspecified => {
+                warn!(
+                    "failure reason for subrequest `{}` is unspecified",
+                    open_shards_failure.subrequest_id
+                );
+                None
```
fulmicoton, line 121: I don't understand why we record `None` here. 
We seem to have logic to different retryable (you called that transient) from non retryable errors.

Could we reserve `None` to we have never received a failure yet?

fulmicoton (approved):

quickwit/quickwit-proto/protos/quickwit/control_plane.proto
guilload: Discussed offline.

The [best practice for Protobuf](https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum) enums is to reuse the name of the enum as a prefix for the variant and always have `FOO_UNSPECIFIED` for the `0` tag.

Prost does the right when code-generating the corresponding Rust enums and strips the common prefix.

PR #4083, opened 2023-11-06
Exposing the node capacity via chitchat

fulmicoton:


quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -55,6 +55,8 @@ pub struct IndexerConfig {
     pub enable_otlp_endpoint: bool,
     #[serde(default = "IndexerConfig::default_enable_cooperative_indexing")]
     pub enable_cooperative_indexing: bool,
+    #[serde(default = "IndexerConfig::default_capacity")]
+    pub capacity: u32,
```
guilload, line 59: I think `capacity` is too ambiguous. Also, I'd rather expose our users to the pipeline concept as little as possible. I think we should ask our users how many CPUs they want to allocate to their indexers, and we infer how many pipelines we can pack from there. With that in mind, `cpu_capacity` is probably what we want.

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes

fulmicoton: You are right. I introduced a CpuCapacity. We can now configure it either as a string "1000m" or as a float. (1.0)

quickwit/quickwit-proto/src/indexing/mod.rs
```diff
@@ -209,13 +210,146 @@ impl TryFrom<&str> for IndexingTask {
...
+        let milli_cpus: u32 = milli_cpus_without_unit_str
+            .parse::<u32>()
+            .map_err(|_err| format!("invalid cpu capacity: `{cpu_capacity_str}`."))?;
+        Ok(CpuCapacity(milli_cpus))
+    }
+}
+
+impl From<CpuCapacity> for CpuCapacityForSerialization {
+    fn from(cpu_capacity: CpuCapacity) -> CpuCapacityForSerialization {
+        CpuCapacityForSerialization::MilliCpuWithUnit(format!("{}m", cpu_capacity.0))
```
guilload, line 352: I know it's tested in `quickwit-config`, but this could be tested in this file directly.

guilload (approved):
nit: `CpuCapacity` is also sometimes used for measuring a load or a threshold. It might be better to use `CpuUnits` for the type, and let the variable names tell us whether it's a capacity, load, etc.

quickwit/quickwit-proto/src/indexing/mod.rs
fulmicoton: adding tests

PR #4095, opened 2023-11-06
Honor `auto.offset.reset` parameter in Kafka source

guilload:
### Description
Previously, the source ignored the `auto.offset.reset` configuration parameter and always began consumption of a partition from the beginning. With this change, it now honors the client configuration. 

### How was this PR tested?
Added test

### DIFF
--- a/docs/configuration/source-config.md
+++ b/docs/configuration/source-config.md
@@ -64,11 +64,14 @@ The Kafka source consumes a `topic` using the client library [librdkafka](https:
 - `bootstrap.servers`
 Comma-separated list of host and port pairs that are the addresses of a subset of the Kafka brokers in the Kafka cluster.
 
+- `auto.offset.reset`
+Defines the behavior of the source when consuming a partition for which there is no initial offset saved in the checkpoint. `earliest` consumes from the beginning of the partition, whereas `latest` (default) consumes from the end.
+
 - `enable.auto.commit`
 The Kafka source manages commit offsets manually using the [checkpoint API](../overview/concepts/indexing.md#checkpoint) and disables auto-commit.
 
 - `group.id`
-Kafka-based distributed indexing relies on consumer groups. Unless overridden in the client parameters, the default group ID assigned to each consumer managed by the source is `quickwit-{index_uid}-{source_id}`
+Kafka-based distributed indexing relies on consumer groups. Unless overridden in the client parameters, the default group ID assigned to each consumer managed by the source is `quickwit-{index_uid}-{source_id}`.
 
 - `max.poll.interval.ms`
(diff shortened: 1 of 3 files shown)


docs/configuration/source-config.md
```diff
@@ -64,11 +64,14 @@ The Kafka source consumes a `topic` using the client library [librdkafka](https:
 - `bootstrap.servers`
 Comma-separated list of host and port pairs that are the addresses of a subset of the Kafka brokers in the Kafka cluster.
 
+- `auto.offset.reset`
```
fmassot, line 67: @tuziben new parameter for the Kafka source.

guilload: This is not a Quickwit-specific parameter but a Kafka one that you can pass to the client, via the `client_params` property. For instance:

```yaml
version: 0.6
source_id: my-kafka-source
source_type: kafka
params:
  topic: my-topic
  client_params:
    auto.offset.reset: earliest
    bootstrap.servers: localhost:9092
    security.protocol: SSL
```

docs/configuration/source-config.md
```diff
@@ -64,11 +64,14 @@ The Kafka source consumes a `topic` using the client library [librdkafka](https:
...
+Defines the behavior of the source when consuming a partition for which there is no initial offset saved in the checkpoint. `earliest` consumes from the beginning of the partition, whereas `latest` (default) consumes from the end.
```
fmassot, line 68: using the `latest` by default is probably the safest value, but it might surprise some users.

quickwit/quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -148,39 +148,43 @@ impl ConsumerContext for RdKafkaContext {
...
-                    .expect("Failed to find partition in assignment. This should never happen! Please, report on https://github.com/quickwit-oss/quickwit/issues.");
-                partition
-                    .set_offset(offset)
-                    .expect("Failed to convert `Offset` to `i64`. This should never happen! Please, report on https://github.com/quickwit-oss/quickwit/issues.");
+            for (partition_id, offset) in assignment {
+                let Some(mut partition) = tpl.find_partition(&self.topic, partition_id) else {
+                    warn!("partition `{partition_id}` not found in assignment");
+                    continue;
+                };
+                if let Err(error) = partition.set_offset(offset) {
```
fmassot, line 182: shouldn't we abort if there is an error on set_offset? (just a question, I don't know which kind of error we get from this call)

docs/configuration/source-config.md
fmassot: pinging @jeromegn so he won't be surprised of this change.

fmassot (approved):

PR #4104, opened 2023-11-09
Shorten overly verbose debug log

rdettai:
### Description

The "Merged leaf search response." log can get out of hand as it prints the entire response, including some binary vectors. This change keeps only some high level info about the respone.

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-search/src/root.rs
+++ b/quickwit/quickwit-search/src/root.rs
@@ -498,8 +498,13 @@ pub(crate) async fn search_partial_hits_phase(
     .await
     .context("failed to merge leaf search responses")?
     .map_err(|error: TantivyError| crate::SearchError::Internal(error.to_string()))?;
-    debug!(leaf_search_response = ?leaf_search_response, "Merged leaf search response.");
-
+    debug!(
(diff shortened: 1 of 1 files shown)


guilload (approved):

quickwit/quickwit-search/src/root.rs
```diff
@@ -1061,7 +1066,10 @@ pub async fn root_list_terms(
         merged_iter.collect()
     };
 
-    debug!(leaf_list_terms_response = ?leaf_list_terms_response, "Merged leaf search response.");
+    debug!(
+        leaf_list_terms_response_count = leaf_list_terms_response.len(),
+        "Merged leaf search response."
+    );
```
rdettai, line 1072: @trinity-1686a, as you initially wrote this, are you fine with this lightening of the debug logging ? (`leaf_list_terms_response` seems to be binary, so it should be pretty hard to use anyway 😄)

trinity-1686a: I believe I copied that line from higher up in the file (the other part of that file getting modified). I'm not sure I even used it once. Feel free to do what you wish

PR #4110, opened 2023-11-10
Exposing published position thru chitchat

fulmicoton:
Exposing the published position through chitchat.

The info is transmitted via an EventSubscriber at the same time as regular truncate:
- upon fetching the starting offset when assigning a shard to an IngestSource
- upon suggest truncate.

This PR does not include any code to remove sources yet. 
I created #4143 accordingly.

### DIFF
--- a/quickwit/quickwit-cluster/src/cluster.rs
+++ b/quickwit/quickwit-cluster/src/cluster.rs
@@ -238,6 +238,16 @@ impl Cluster {
             .set(key, value);
     }
 
+    pub async fn get_self_key_value(&self, key: &str) -> Option<String> {
+        self.chitchat()
+            .await
+            .lock()
+            .await
+            .self_node_state()
+            .get_versioned(key)
+            .map(|versioned_value| versioned_value.value.clone())
+    }
+
     /// Waits until the predicate holds true for the set of ready members.
     pub async fn wait_for_ready_members<F>(
         &self,
--- a/quickwit/quickwit-common/src/pubsub.rs
+++ b/quickwit/quickwit-common/src/pubsub.rs
@@ -131,7 +131,7 @@ impl EventBroker {
         }
     }
 
-    /// Subscribe to an event with a function callback.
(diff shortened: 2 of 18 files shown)


quickwit/quickwit-proto/src/indexing/mod.rs
```diff
@@ -207,6 +207,19 @@ impl TryFrom<&str> for IndexingTask {
     }
 }
 
+#[derive(Debug, Clone, Copy, Default, Eq, PartialEq, Serialize, Deserialize, utoipa::ToSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum IndexingStatus {
+    #[default]
+    Active,
+    // We have emitted all documents of the pipeline until EOF.
```
guilload, line 215: I think that's dangerous. Why don't we set that status on `suggest_truncate` rather than on receiving an EOF record?

fulmicoton: This was the previous semantic for `Complete`. We can change it if you think it is safe.

quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -238,6 +239,16 @@ impl Cluster {
             .set(key, value);
     }
 
+    pub async fn get_self_key_value(&self, key: &str) -> Option<String> {
```
fulmicoton, line 242: i added it for unit tests.

quickwit/quickwit-common/src/pubsub.rs
```diff
@@ -111,20 +158,18 @@ impl EventBroker {
...
                 }));
             }
         }
     }
 }
 
-#[derive(Debug)]
 struct EventSubscription<E> {
-    #[allow(dead_code)]
-    subscription_id: usize, // Used for the `Debug` implementation.
```
fulmicoton, line 126: I removed the Debug implementation (we never use it.)

guilload (approved):

Issue #3829, opened 2023-09-13
Stop using actor heartbeat duration as the loop interval for sources


guilload:
The actor heartbeat duration was increased, and now sources are slow to respond to other messages (`AssignShards`,  `Observe`).

Let's go for 1s.

guilload:
Closed via #4112.

PR #4112, opened 2023-11-10
fix: replace actor heartbeat duration with emit timeout as the loop i…

etolbakov:
### Description

replace actor heartbeat duration with emit timeout as the loop interval for sources.

### How was this PR tested?
This is a configuration change, no special tests have been made

### Related issues
https://github.com/quickwit-oss/quickwit/issues/3829

### DIFF
--- a/quickwit/quickwit-indexing/src/source/kafka_source.rs
+++ b/quickwit/quickwit-indexing/src/source/kafka_source.rs
@@ -489,7 +489,6 @@ impl Source for KafkaSource {
     ) -> Result<Duration, ActorExitStatus> {
         let now = Instant::now();
         let mut batch = BatchBuilder::default();
-        // let deadline = time::sleep(*quickwit_actors::HEAR_TBEAT / 2);
         let deadline = time::sleep(EMIT_BATCHES_TIMEOUT);
         tokio::pin!(deadline);
 
--- a/quickwit/quickwit-indexing/src/source/kafka_source.rs
+++ b/quickwit/quickwit-indexing/src/source/kafka_source.rs
@@ -51,21 +51,8 @@ use crate::actors::DocProcessor;
 use crate::models::{NewPublishLock, PublishLock};
 use crate::source::{
     BatchBuilder, Source, SourceContext, SourceRuntimeArgs, TypedSourceFactory,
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-indexing/src/source/kafka_source.rs
```diff
@@ -49,7 +49,10 @@ use tracing::{debug, info, warn};
...
 use crate::actors::DocProcessor;
 use crate::models::{NewPublishLock, PublishLock};
-use crate::source::{BatchBuilder, Source, SourceContext, SourceRuntimeArgs, TypedSourceFactory};
+use crate::source::{
+    BatchBuilder, Source, SourceContext, SourceRuntimeArgs, TypedSourceFactory,
+    EMIT_BATCHES_TIMEOUT,
+};
 
 /// Number of bytes after which we cut a new batch.
 ///
```
etolbakov, line 58: I've noticed that here and in other sources we are using something like `BATCH_NUM_BYTES_LIMIT: u64 = 5_000_000`.
1. I'm happy to replace it with Bytsize
2. should it be `5Mb` (as it here) or `5Mib`(like in [mod.rs](https://github.com/quickwit-oss/quickwit/pull/4112/files#diff-a2df65a154da2a6ab79da9866d7a5dfca567d9951ee3bd3eb2687fe3d4681801))?

etolbakov:
Adrien could you please take a look when you have time?

quickwit/quickwit-indexing/src/source/kafka_source.rs
guilload: 1. Yes, good idea. Let's use a shared `Bytsize ` constant defined in the source module for all sources, as you suggested.

2. Up to you :)

guilload (approved):

etolbakov:
Adrien I've made some changes (hopefully improvements 😅) could you please take a look when you have a spare minute?

guilload:
Once the PR is approved, feel free to get the branch in a mergeable state and merge it yourself. You should have access. Let me know if it's not the case.

guilload:
Merged. Thanks a lot @etolbakov !

PR #4117, opened 2023-11-10
Implement ingester graceful shutdown

guilload:
### Description
Add a `Decommission` RPC to ingesters. By default, ingesters are in `Ready` status. Upon receiving a decommissioning request, ingesters change their status to `Decommissioning`, close all their shards, become read-only and wait for them to be indexed by monitoring calls to `truncate`. When the indexing of all the shards is completed, ingesters transition to the `Decommissioned` status at which point they become safe to shut down.

I also added a `OpenObservationStream` RPC to ingesters, so we can track the status of the decommissioning process. In the future, we can enrich the `ObservationMessage` object to track more things.

### How was this PR tested?
- Updated unit tests
- Add unit tests
- Tested manually

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
@@ -34,20 +34,22 @@ use mrecordlog::MultiRecordLog;
 use quickwit_common::tower::Pool;
 use quickwit_common::ServiceStream;
 use quickwit_proto::ingest::ingester::{
-    AckReplicationMessage, CloseShardsRequest, CloseShardsResponse, FetchResponseV2,
-    IngesterService, IngesterServiceClient, IngesterServiceStream, OpenFetchStreamRequest,
-    OpenReplicationStreamRequest, OpenReplicationStreamResponse, PersistFailure,
-    PersistFailureReason, PersistRequest, PersistResponse, PersistSuccess, PingRequest,
-    PingResponse, ReplicateFailureReason, ReplicateRequest, ReplicateSubrequest,
-    SynReplicationMessage, TruncateRequest, TruncateResponse,
+    AckReplicationMessage, CloseShardsRequest, CloseShardsResponse, DecommissionRequest,
+    DecommissionResponse, FetchResponseV2, IngesterService, IngesterServiceClient,
+    IngesterServiceStream, IngesterStatus, ObservationMessage, OpenFetchStreamRequest,
+    OpenObservationStreamRequest, OpenReplicationStreamRequest, OpenReplicationStreamResponse,
+    PersistFailure, PersistFailureReason, PersistRequest, PersistResponse, PersistSuccess,
+    PingRequest, PingResponse, ReplicateFailureReason, ReplicateRequest, ReplicateSubrequest,
+    SynReplicationMessage, TruncateShardsRequest, TruncateShardsResponse,
 };
 use quickwit_proto::ingest::{CommitTypeV2, IngestV2Error, IngestV2Result, ShardState};
 use quickwit_proto::types::{NodeId, Position, QueueId};
-use tokio::sync::RwLock;
+use tokio::sync::{watch, RwLock};
 use tracing::{error, info, warn};
 
 use super::fetch::FetchStreamTask;
-use super::models::{IngesterShard, PrimaryShard};
+use super::models::IngesterShard;
+use super::mrecord::MRecord;
 use super::mrecordlog_utils::{append_eof_record_if_necessary, check_enough_capacity};
 use super::rate_limiter::{RateLimiter, RateLimiterSettings};
 use super::replication::{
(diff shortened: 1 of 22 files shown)


quickwit/quickwit-ingest/src/ingest_v2/mrecordlog_utils.rs
```diff
@@ -104,8 +104,21 @@ pub(super) fn check_enough_capacity(
     Ok(())
 }
 
+/// Determines whether the queue is drained. A queue is considered drained if it is empty or has
+/// one EOF record.
+pub(super) fn is_queue_drained(mrecordlog: &MultiRecordLog, queue_id: &QueueId) -> bool {
+    mrecordlog
```
fulmicoton, line 110: This reminded me to open
https://github.com/quickwit-oss/quickwit/issues/4135

(We should discuss that some day.)

quickwit/quickwit-proto/protos/quickwit/ingester.proto
```diff
@@ -24,24 +24,31 @@ package quickwit.ingest.ingester;
...
+  rpc OpenObservationStream(OpenObservationStreamRequest) returns (stream ObservationMessage);
 
+  // Closes a set of shards.
   rpc CloseShards(CloseShardsRequest) returns (CloseShardsResponse);
 
   // Pings an ingester to check if it is ready to host shards and serve requests.
   rpc Ping(PingRequest) returns (PingResponse);
 
-  // Truncates the shards at the given positions. Indexers should call this RPC on leaders, which will replicate the request to followers.
+  // Truncates the shards at the given positions.
```
fulmicoton, line 46: With replication is the information conveyed somehow to the replication peer?
(either way it is worth commenting here - or somewhere else- I think)

quickwit/quickwit-ingest/src/ingest_v2/models.rs
```diff
@@ -17,190 +17,169 @@
...
-    Primary(PrimaryShard),
-    /// A replica shard hosted on a follower.
-    Replica(ReplicaShard),
-    /// A shard hosted on a single node when the replication factor is set to 1.
-    Solo(SoloShard),
-}
+    pub fn notify_new_records(&mut self) {
+        self.new_records_tx
+            .send(())
+            .expect("channel should be open");
```
fulmicoton, line 106:
```suggestion
            // `new_records_tx` is guaranteed to have a received since one of them is held by `self`.
            .expect("channel should be open");
```

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -139,7 +158,28 @@ impl Ingester {
...
+        if state.status != IngesterStatus::Decommissioning {
+            return;
+        }
+        if state.shards.values().all(|shard| {
+            shard.shard_state.is_closed() && shard.indexing_status == IndexingStatus::Complete
+        }) {
+            info!("ingester fully decommissioned");
+            state.status = IngesterStatus::Decommissioned;
+
+            state.observation_tx.send_if_modified(|observation_result| {
```
fulmicoton, line 172: TIL about send_if_modified.

fulmicoton (approved):

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -156,9 +196,14 @@ impl Ingester {
         for queue_id in queue_ids {
             append_eof_record_if_necessary(&mut state_guard.mrecordlog, &queue_id).await;
 
-            let solo_shard = SoloShard::new(ShardState::Closed, Position::Eof);
-            let shard = IngesterShard::Solo(solo_shard);
-            state_guard.shards.insert(queue_id.clone(), shard);
+            let indexing_status = if is_queue_drained(&state_guard.mrecordlog, &queue_id) {
+                IndexingStatus::Complete
```
fulmicoton, line 200: Isn't this a shift in the semantics of indexing status?
Since it is ambiguous, can we have its semantics as a comment on the enum

fulmicoton: Ok investigating it is a bit difficult.

So in the ingester:
`IndexingStatus::Complete` means the queue has entirely been consumed. The signal is coming from the truncation.

In the ingester source:
`IndexingStatus::Complete` means the EOF record has been met. This is a reversible state. For instance if the pipeline fails.

guilload: It does not have to be called `IndexingStatus` here. I was hoping to reuse the enum that you're about to move in `quickwit-proto` but I agree that it's not the same thing from a semantics perspective.

quickwit/quickwit-proto/protos/quickwit/ingester.proto
guilload: No, indexers must call truncate directly on leaders and followers. I'll update the doc.

Issue #3810, opened 2023-09-07
Implement shard API for PostgreSQL metastore


guilload:
- [x] `open_shards`
- [x] `acquire_shards`
- [x]  `list_shards` 
- [x] `delete_shards`

guilload:
@fmassot, @kamalesh0406 is going to give you a hand for this one. There are four calls to implement: `open_shards`, `acquire_shards`, `list_shards`,  and `delete_shards`. Let's open one PR for each so it's easy to split the work between you. Please leave a comment here and update the to-do when one of you starts working on a new one.

Implementation pointers:
- there already exists a [reference implementation](https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index/shards.rs) for the file-backed metastore
- the test suite for the Shard API lives in [this file](https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-metastore/src/tests/shard.rs): please fill in the blanks and add more tests as you see fit.

kamalesh0406:
I am starting the work for the `open_shards` @fmassot

kamalesh0406:
I added the implementation for list_shards(unit test pending :) ) to my existing PR as well. I will see if I can get delete/acquire_shards done over the next couple of days.

guilload:
Closed via #4403.

PR #4119, opened 2023-11-11
Add migrations for shards table in postgres metastore

kamalesh0406:
### Description

This is the first set of changes to support the shards API in the postgres metastore(#3810). There are couple of items that still need to be addressed:
1. Using `index_uid` vs `index_id` for the table, the ingestv2_api seems to use (index_id, shard_id) whereas the filebacked index uses index_uid. What is everyone's consensus on the right column for this?
2. Do we need to store the `SourceCheckpoint` similar to this https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index/shards.rs#L42 ? If so, I think the right approach might be to create another table for it.
3. We are using an auto increement column type for `next_shard_id` in this PR. But here https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-metastore/src/metastore/file_backed_metastore/file_backed_index/shards.rs#L144 we use the next_shard_id from the subrequest. Can the subrequest have a `next_shard_id` that can deviate from the indexes `next_shard_id`? If yes, then I would need to remove the auto-increement.

### How was this PR tested?

sqlx migrate run  --database-url postgres://quickwit-dev:quickwit-dev@localhost:5432/quickwit-metastore-dev --source migrations/postgresql


guilload:
1. Let's use `index_uid` for this table. The `index_id` -> `index_uid` resolution happens in the control-plane.

2. No, we don't need to store the `SourceCheckpoint` in a similar way. It's just a shortcut that I took to reuse the checkpointing logic that we had, but in reality all we need is a `publish_position` per shard.

3. They should not deviate, so auto-increment is the way to go.

kamalesh0406:
Okay that sounds good @guilload . I think my current changes are fine then!

quickwit/quickwit-metastore/migrations/postgresql/12_create-shards.up.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TYPE SHARDSTATE AS ENUM ('SHARD_STATE_UNSPECIFIED', 'SHARD_STATE_OPEN', 'SHARD_STATE_UNAVAILABLE', 'SHARD_STATE_CLOSED');
+
+CREATE TABLE IF NOT EXISTS shards (
+    index_uid VARCHAR(288) NOT NULL,
```
guilload, line 4: Why `288` and not `282`? (255 + 1 for `:` separator + 26 for ULID)?

quickwit/quickwit-metastore/migrations/postgresql/12_create-shards.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    source_id VARCHAR(50) NOT NULL,
```
guilload, line 5:
```suggestion
    source_id VARCHAR(255) NOT NULL,
```

quickwit/quickwit-metastore/migrations/postgresql/12_create-shards.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    shard_id BIGSERIAL,
+    leader_id VARCHAR(50) NOT NULL,
```
guilload, line 7:
```suggestion
    leader_id VARCHAR(255) NOT NULL,
```

quickwit/quickwit-metastore/migrations/postgresql/12_create-shards.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    follower_id VARCHAR(50),
```
guilload, line 8:
```suggestion
    follower_id VARCHAR(255),
```

quickwit/quickwit-metastore/migrations/postgresql/12_create-shards.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    shard_state SHARDSTATE NOT NULL,
+    publish_position_inclusive bytea NOT NULL,
+    publish_token VARCHAR(50),
```
guilload, line 11:
```suggestion
    publish_token VARCHAR(255),
```

guilload (approved):

quickwit/quickwit-metastore/migrations/postgresql/12_create-shards.up.sql
kamalesh0406: Yup it should be 282, I mistaked it for 288.

PR #4127, opened 2023-11-13
Bugfix: truncating the ingest v1 source on initialization.

fulmicoton:
Tested with a unit test.

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_service.proto
+++ b/quickwit/quickwit-ingest/src/ingest_service.proto
@@ -88,7 +88,7 @@ message IngestResponse {
     uint64 num_docs_for_processing = 1;
 }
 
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-ingest/src/ingest_service.proto
```diff
@@ -88,6 +88,7 @@ message IngestResponse {
     uint64 num_docs_for_processing = 1;
 }
 
+// Fetch messages that have position strictly after `start_after`.
```
fmassot, line 91:
```suggestion
// Fetch messages with position strictly after `start_after`.
```

fmassot (approved):

fulmicoton:
Tested in situ... Both on resume on panic and a crash + restart.

PR #4136, opened 2023-11-14
fix: Correct the percentile values passed to the percentile function in DescriptiveStats for the index describe command

unpervertedkid:
### Description
This PR corrects the percentile values passed to the percentile function in the DescriptiveStats struct. Previously, the percentiles were incorrectly set as 1, 50, 50, 75, and 75. This PR changes these values to correctly represent the 1st, 25th, 50th, 75th, and 99th percentiles.
### How was this PR tested?
This PR was tested by running the existing unit tests for the DescriptiveStats struct. Additionally, manual testing was performed by generating statistics for a sample dataset and verifying that the outputted percentiles were correct.

More assertions were added to prevent the same issue from recurring in the future.


unpervertedkid:
@guilload am still figuring my way around the codebase, could you help point me in the right direction, I am assuming you mean add assertions that should catch similar errors in the future?

quickwit/quickwit-cli/src/index.rs
```diff
@@ -1161,6 +1161,7 @@ mod test {
         Ok(())
     }
 
+    #[test]
```
guilload, line 1164: Remove this line.

guilload (approved):
You got it right! Remove the unnecessary line and we should be good.

quickwit/quickwit-cli/src/index.rs
unpervertedkid: Ooh yeah, I missed this... should be good now

guilload:
You can run make fmt to fix the formatting issues.
You can run make fix to fix some of the linting issues.
You can run make fmt to test the entire project.

You can run cargo test --manifest-path quickwit/Cargo.toml -p quickwit-common to test a single crate.

-> [Contributing guidelines](https://github.com/quickwit-oss/quickwit/blob/main/CONTRIBUTING.md)

unpervertedkid:
@guilload Should be good now.

guilload:
Thanks @unpervertedkid !

PR #4149, opened 2023-11-16
Close shards on IO error

guilload:
### Description
Close shards on IO error. Since now we call `InitShards` on ingesters, I abandoned the idea of decommissioning ingesters on IO error.

### How was this PR tested?
Nope...

I tried:
1. Open the log
2. Change permissions of log files to read-only
3. Append records

It does not fail on my MacBook. I suspect the file descriptor is still "writeable".

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
@@ -20,7 +20,6 @@
 use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 use std::fmt;
-use std::iter::once;
 use std::path::Path;
 use std::sync::{Arc, Weak};
 use std::time::Duration;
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -583,6 +613,28 @@ impl Ingester {
...
             }
         }
+        if !shards_to_close.is_empty() {
+            for queue_id in &shards_to_close {
+                let shard = state_guard
+                    .shards
+                    .get_mut(queue_id)
+                    .expect("shard should exist");
+
+                shard.shard_state = ShardState::Closed;
```
fulmicoton, line 623: when/through which path change of state is this reported to the control plane -> routers?

fulmicoton (approved):
failpoint can be used to return io Error in unit tests.

guilload:
Right, good idea! Let me update this PR with some failpoint-based unit tests, then.

PR #4153, opened 2023-11-16
Use millis for search after values and returned sort values.

fmassot:
Not ready for review.


quickwit/quickwit-proto/src/search/mod.rs
```diff
@@ -104,7 +104,19 @@ impl SortByValue {
                     return None;
                 }
             }
-            String(_) | Array(_) | Object(_) => return None,
+            // Strings that can be converted to a number are accepted.
+            // Some clients (like JS clients) can't easily handle large integers
+            // without losing precision, so we accept them as strings.
+            String(value) => {
+                if let Ok(number) = value.parse::<u64>() {
```
fulmicoton, line 111:
```suggestion
                if let Ok(number) = value.parse::<u64>() {
```
i64 first please.

quickwit/rest-api-tests/scenarii/es_compatibility/0018-search_after.yaml
```diff
@@ -44,3 +44,102 @@ expected:
...
+# Accept timestamp in epoch_millis
+json:
+  size: 100
+  track_total_hits: true
+  query:
+      match_all: {}
+  sort:
+    - created_at:
+        order: asc
+        format: epoch_millis
```
fulmicoton, line 120: did you run this tests against elastic too? If they don't pass, please comment and add 

`engine: quickwit`

fulmicoton (approved):

quickwit/rest-api-tests/scenarii/es_compatibility/0018-search_after.yaml
fmassot: good catch

fmassot:
I finally changed the name of the date time format for the ES API. 
Reason: `epoch_millis` format return milliseconds as strings.... I want to return integers to be able to replicate the default behavior of elasticsearch when no format is provided.

PR #4156, opened 2023-11-17
Make the ShardPositions model into a ShardPositionsService actor

fulmicoton:
Its purpose is to encapsulate all of the logic used to maintained a
distributed eventually consistent view of the published shard positions
over the cluster.

From the user point of view, after instantiation
- indexing pipelines need to feed it with updates. (this happens on
  suggest_truncate). This is done by publishing
  `LocalShardPositionsUpdates` to the event broker.
- clients interested in updates can just subscript the
  `ShardPositionsUpdate` object in the event broker. The event
  received can come from a local indexing pipeline or anywhere in the
  cluster.

The service takes care of deduping/ignoring updates when necessary.

The two object (Local and not) are very similar, but different in
semantics.

This PR includes the connection with the control plane deleting the shards from the metastore and updating its internal model too.

It however does not contains any path for the deletion in the ingester.

Tested manually for the moment, but will add integration test later.

### DIFF
--- a/quickwit/quickwit-control-plane/src/control_plane.rs
+++ b/quickwit/quickwit-control-plane/src/control_plane.rs
@@ -21,6 +21,7 @@ use std::time::Duration;
 
 use anyhow::Context;
 use async_trait::async_trait;
+use fnv::FnvHashSet;
 use quickwit_actors::{
     Actor, ActorContext, ActorExitStatus, ActorHandle, Handler, Mailbox, Supervisor, Universe,
 };
@@ -31,12 +32,14 @@ use quickwit_proto::control_plane::{
     ControlPlaneError, ControlPlaneResult, GetOrCreateOpenShardsRequest,
     GetOrCreateOpenShardsResponse,
 };
+use quickwit_proto::indexing::ShardPositionsUpdate;
 use quickwit_proto::metastore::{
     serde_utils as metastore_serde_utils, AddSourceRequest, CreateIndexRequest,
-    CreateIndexResponse, DeleteIndexRequest, DeleteSourceRequest, EmptyResponse, MetastoreError,
-    MetastoreService, MetastoreServiceClient, ToggleSourceRequest,
+    CreateIndexResponse, DeleteIndexRequest, DeleteShardsRequest, DeleteShardsSubrequest,
+    DeleteSourceRequest, EmptyResponse, MetastoreError, MetastoreService, MetastoreServiceClient,
+    ToggleSourceRequest,
 };
-use quickwit_proto::types::{IndexUid, NodeId};
+use quickwit_proto::types::{IndexUid, NodeId, Position, ShardId, SourceUid};
 use serde::Serialize;
 use tracing::error;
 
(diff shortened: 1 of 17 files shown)


quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -135,6 +138,71 @@ impl Actor for ControlPlane {
     }
 }
 
+impl ControlPlane {
+    async fn delete_shards(
```
guilload, line 142: Should we move this method to the ingest controller? So far, this is where we handle all shard mutations.

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -889,4 +965,142 @@ mod tests {
...
+                assert_eq!(delete_shards_request.subrequests.len(), 1);
+                let subrequest = &delete_shards_request.subrequests[0];
+                assert_eq!(subrequest.index_uid, index_uid_clone);
+                assert_eq!(subrequest.source_id, INGEST_SOURCE_ID);
+                assert_eq!(&subrequest.shard_ids[..], &[17]);
+                Ok(DeleteShardsResponse {})
+            },
+        );
+
+        let mut shard = Shard {
```
guilload, line 1014: `Shard` implements `Default`.

quickwit/quickwit-indexing/src/models/shard_positions.rs
```diff
@@ -18,57 +18,198 @@
...
-use tracing::warn;
+use quickwit_common::pubsub::{Event, EventBroker};
+use quickwit_proto::indexing::ShardPositionsUpdate;
+use quickwit_proto::types::{IndexUid, Position, ShardId, SourceUid};
+use tracing::{error, warn};
 
-#[derive(Debug, Clone)]
-pub struct PublishedShardPositionsUpdate {
+/// Prefix used in chitchat to publish the shard positions.
+const SHARD_POSITIONS_PREFIX: &str = "shard_positions:";
```
guilload, line 36:
```suggestion
const SHARD_POSITIONS_PREFIX: &str = "indexer.shard_positions:";
```

quickwit/quickwit-indexing/src/models/shard_positions.rs
```diff
@@ -78,29 +219,40 @@ impl PublishedShardPositions {
...
 
+    fn publish_shard_updates_to_event_broker(&self, source_uid: SourceUid) {
+        let Some(shard_positions_map) = self.shard_positions_per_source.get(&source_uid) else {
+            return;
+        };
+        let shard_positions: Vec<(ShardId, Position)> = shard_positions_map
+            .iter()
+            .map(|(&shard_id, position)| (shard_id, position.clone()))
+            .collect();
+        self.event_broker.publish(ShardPositionsUpdate {
```
guilload, line 235: At the moment, I don't see the value of sending complete updates. It's just more work for the ingester on truncate or the control plane on shard deletion to detect positions that have already been processed.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -220,7 +222,7 @@ async fn start_control_plane_if_needed(
         .await?;
 
         let cluster_id = cluster.cluster_id().to_string();
-        let self_node_id = cluster.self_node_id().to_string();
+        let self_node_id: NodeId = cluster.self_node_id().to_string().into();
```
guilload, line 225:
```suggestion
        let self_node_id: NodeId = cluster.self_node_id().into();
```

guilload (approved):

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -135,6 +138,71 @@ impl Actor for ControlPlane {
...
+            .into_iter()
+            .filter(|(shard_id, position)| {
+                (position == &Position::Eof) && known_shard_ids.contains(shard_id)
+            })
+            .map(|(shard_id, _position)| shard_id)
+            .collect();
+        if shard_ids_to_close.is_empty() {
+            return Ok(());
+        }
+        self.delete_shards(&shard_positions_update.source_uid, &shard_ids_to_close[..])
```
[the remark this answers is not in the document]
fulmicoton: no. I am adding documentation in the control plane/metastore and adding a warning log if that happens.

PR #4159, opened 2023-11-18
Minor fixes & doc

fulmicoton:
- Bugfix in the metastore: an index deletion should result in a contorl plane restarting depending on the nature of the error.
- Added mssing doc in the metastore.

### DIFF
--- a/quickwit/quickwit-control-plane/src/control_plane.rs
+++ b/quickwit/quickwit-control-plane/src/control_plane.rs
@@ -253,8 +253,8 @@ impl Handler<DeleteIndexRequest> for ControlPlane {
     ) -> Result<Self::Reply, ActorExitStatus> {
         let index_uid: IndexUid = request.index_uid.clone().into();
 
-        if let Err(error) = self.metastore.delete_index(request).await {
-            return Ok(Err(ControlPlaneError::from(error)));
+        if let Err(metastore_error) = self.metastore.delete_index(request).await {
+            return convert_metastore_error(metastore_error);
         };
 
         self.model.delete_index(&index_uid);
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -253,8 +253,8 @@ impl Handler<DeleteIndexRequest> for ControlPlane {
     ) -> Result<Self::Reply, ActorExitStatus> {
         let index_uid: IndexUid = request.index_uid.clone().into();
 
-        if let Err(error) = self.metastore.delete_index(request).await {
-            return Ok(Err(ControlPlaneError::from(error)));
+        if let Err(metastore_error) = self.metastore.delete_index(request).await {
+            return convert_metastore_error(metastore_error);
```
fulmicoton, line 257: that's a bugfix I believe. Without this, if we encounterred an error from the metastore, depending on the error, we could observe a divergence between the control plane model  and the metastore.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1022,11 +1022,11 @@ impl MetastoreService for PostgresqlMetastore {
         &mut self,
         request: IndexMetadataRequest,
     ) -> MetastoreResult<IndexMetadataResponse> {
-        let response = if let Some(index_id) = &request.index_id {
-            index_opt(&self.connection_pool, index_id).await?
-        } else if let Some(index_uid) = &request.index_uid {
+        let response = if let Some(index_uid) = &request.index_uid {
```
fulmicoton, line 1025: I just inverted uid and id.

quickwit/quickwit-control-plane/src/control_plane.rs
fmassot: We should add a test on this I think. Can you open an issue and assigned it to me?

fmassot (approved):

quickwit/quickwit-control-plane/src/control_plane.rs
fulmicoton: created #4185

PR #4164, opened 2023-11-19
Add implementation for open_shards in the metastore

kamalesh0406:
### Description

This PR adds the implementation of open_shards in the metastore for #3810 . The unit tests are missing in this PR and they will be added once I get the initial review on the code.

### How was this PR tested?

Not tested yet, will be done after discussing with @guilload .


quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1291,9 +1296,23 @@ impl MetastoreService for PostgresqlMetastore {
 
     async fn open_shards(
```
guilload, line 1297: There might be a way to perform only one pass per index:
```SQL
INSERT INTO shards (index_uid, source_id, leader_id, follower_id, shard_state, publish_position_inclusive, publish_token)
VALUES ($1, $2, $3, $4, $5, $6, $7)
ON CONFLICT (index_uid, source_id, shard_id)
DO NOTHING
RETURNING *;
```

kamalesh0406: Oh okay I have never tried this out before, I will update my code then.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1379,6 +1398,81 @@ fn tags_filter_expression_helper(tags: &TagFilterAst) -> Cond {
...
+    subrequest: OpenShardsSubrequest,
+) -> MetastoreResult<OpenShardsSubresponse> {
+    let index_uid = subrequest.index_uid;
+
+    let pg_existing_shards: Option<PgShards> = sqlx::query_as::<_, PgShards>(
+        r#"
+            SELECT index_uid, source_id, shard_id FROM shards WHERE index_uid = $1 AND source_id = $2 AND shard_id = $3
+        "#
+    )
+    .bind(index_uid.to_string())
```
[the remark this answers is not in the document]
kamalesh0406: Sure will fix it.

quickwit/quickwit-metastore/src/metastore/postgresql_model.rs
```diff
@@ -210,3 +211,65 @@ impl TryInto<DeleteTask> for PgDeleteTask {
...
+            PgShardState::Unspecified => ShardState::Unspecified,
+            PgShardState::Open => ShardState::Open,
+            PgShardState::Unavailable => ShardState::Unavailable,
+            PgShardState::Closed => ShardState::Closed,
+        }
+    }
+}
+
+#[derive(sqlx::FromRow)]
+pub struct PgShards {
```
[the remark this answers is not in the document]
kamalesh0406: The name of the table in psql is shards so I am a bit skeptical if it is right to keep it as `PgShard`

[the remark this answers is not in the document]
guilload: But this object represents one shard, i.e., one row, right? So it makes sense to call the table shards since it has 0* shards/rows.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1895,4 +1969,51 @@ mod tests {
...
+        let pg_shard: Option<PgShard> = sqlx::query_as::<_, PgShard>("SELECT * FROM shards")
+            .fetch_optional(&pool)
+            .await?;
+
+        assert!(pg_shard.is_none());
+
+        Ok(())
+    }
+
+    #[sqlx::test(migrator = "MIGRATOR")]
```
kamalesh0406, line 1984: @guilload This is not the complete series of test but this a good first test to check if `sqlx::test` works and I feel it does :) . If you happy with this I can add these to the main tests in the other directory.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1291,30 +1296,172 @@ impl MetastoreService for PostgresqlMetastore {
...
+            let next_shard_id = shards
+                .iter()
+                .max_by_key(|shard| shard.shard_id)
+                .unwrap()
+                .shard_id
+                + 1;
+
+            let subresponse = ListShardsSubresponse {
+                index_uid: subrequest.index_uid.to_string(),
+                source_id: subrequest.source_id.to_string(),
```
guilload, line 1415:
```suggestion
                index_uid: subrequest.index_uid,
                source_id: subrequest.source_id,
```

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1291,30 +1296,172 @@ impl MetastoreService for PostgresqlMetastore {
...
+                let pg_shard_state: String =
+                    (PgShardState::try_from(subrequest.shard_state.unwrap()).unwrap()).into();
+                query = query.bind(pg_shard_state);
+            }
+
+            let pg_shards = query
+                .fetch_all(&self.connection_pool)
+                .await
+                .map_err(|err| {
+                    convert_sqlx_err(
```
guilload, line 1398: Same as above.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1379,6 +1526,62 @@ fn tags_filter_expression_helper(tags: &TagFilterAst) -> Cond {
     }
 }
 
+async fn open_pgshards_by_index(
```
guilload, line 1529: I think we can inline this function in `open_shards`.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
guilload: Let's focus on adding tests to the metastore shard testsuite in `src/tests/shard.rs` and only add tests specific to the Pg implementation here.

quickwit/quickwit-metastore/src/metastore/postgresql_model.rs
```diff
@@ -210,3 +211,84 @@ impl TryInto<DeleteTask> for PgDeleteTask {
...
+    type Error = MetastoreError;
+
+    fn try_from(val: i32) -> Result<PgShardState, Self::Error> {
+        match val {
+            0 => Ok(PgShardState::Unspecified),
+            1 => Ok(PgShardState::Open),
+            2 => Ok(PgShardState::Unavailable),
+            3 => Ok(PgShardState::Closed),
+            _ => Err(MetastoreError::InvalidArgument {
+                message: "The shard state is incorrect".to_string(),
```
guilload, line 259:
```suggestion
            other => Err(MetastoreError::InvalidArgument {
                message: format!("unknown shard state `{other}`")
```

quickwit/quickwit-metastore/src/metastore/postgresql_model.rs
```diff
@@ -210,3 +211,84 @@ impl TryInto<DeleteTask> for PgDeleteTask {
...
+#[sqlx(type_name = "SHARD_STATE")]
+#[sqlx(rename_all = "lowercase")]
+pub enum PgShardState {
+    Unspecified,
+    Open,
+    Unavailable,
+    Closed,
+}
+
+impl From<PgShardState> for String {
```
guilload, line 225: I think we can get rid of this and use [`as_json_str_name`](https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-proto/src/ingest/mod.rs#L195) directly when binding values to queries.

kamalesh0406:
@guilload I have not added the unit tests yet, I think it is better if you review this after I add them.

quickwit/quickwit-metastore/src/tests/mod.rs
```diff
@@ -45,6 +48,18 @@ pub trait DefaultForTest {
...
 
+#[non_exhaustive]
+pub enum MetastoreShardsTestInput {
+    FileBackedIndex,
+    #[cfg(feature = "postgres")]
+    PostgresIndex(Pool<Postgres>),
+}
+
+#[async_trait]
+pub trait DefaultForShard {
```
kamalesh0406, line 59: This is my solution to solve the problem of having a single test function for both the postgres and the filebacked metastore. Ideally, we should modify the function `default_for_test` to accept an additional input paramater. But it is being used in a lot of test functions so I did not want to proceed with that change without your approval @guilload .

.github/workflows/ci.yml
```diff
@@ -13,6 +13,7 @@ on:
 env:
   CARGO_INCREMENTAL: 0
   QW_DISABLE_TELEMETRY: 1
+  DATABASE_URL: postgres://quickwit-dev:quickwit-dev@localhost:5432/quickwit-metastore-dev
```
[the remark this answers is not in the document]
kamalesh0406: No it seems like sqlx::test looks for `DATABASE_URL`. I searched if we can modify the env variable but I am not able to find a good solution for it yet.

quickwit/quickwit-control-plane/src/model/mod.rs
```diff
@@ -196,7 +196,7 @@ impl ControlPlaneModel {
     ) -> ControlPlaneResult<()> {
         let index_metadata = self.index_table.get_mut(index_uid).ok_or_else(|| {
             MetastoreError::NotFound(EntityKind::Index {
-                index_id: index_uid.to_string(),
+                index_id: index_uid.index_id().to_string(),
```
kamalesh0406, line 199: I didn't want to waste time creating an issue for this bug I found so I fixed it here myself. Please lemme know if your intentions where to keep `index_uid` then I will revert it.

quickwit/quickwit-metastore/src/tests/mod.rs
```diff
@@ -363,11 +377,22 @@ macro_rules! metastore_test_suite {
                 $crate::tests::split::test_metastore_stage_splits::<$metastore_type>().await;
             }
 
-            /// Shard API tests
+        }
+    };
+}
 
+#[macro_export]
```
kamalesh0406, line 384: I had to remove the shard_tests from the common tests because of import issues caused by using sqlx. I tried adding an if condition based on `#[cfg(feature="postgres")]` but I quickly realized that the tests would be run for the `FileBackedIndex` even in that case. I am open to other ideas to make this better.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1284,9 +1294,110 @@ impl MetastoreService for PostgresqlMetastore {
...
+                             expected next shard ID `{}`, got `{}`",
+                            max_shard_id, subrequest.next_shard_id
+                        );
+                        return Err(MetastoreError::InconsistentControlPlaneState);
+                    }
+                }
+
+                let pg_shards: Vec<PgShard> = sqlx::query_as::<_, PgShard>(
+                        r#"
+                        INSERT INTO shards (index_uid, source_id, shard_id, leader_id, follower_id, shard_state, publish_position_inclusive, publish_token)
```
kamalesh0406, line 1359: 1. `RETURNING *` works only if there was modification to columns, on conflict it doesn't work and we end up with no output. It seems the only way to ensure that it always returns a value is to add a dummy update. This is a dumb solution and I am open to other queries to make this better.
2. I researched using transactions for this query so that we just get a single shard instead of a vector of all shards. The issue is the error `cannot insert multiple commands into a prepared statement` because a psql engine can only handle a single atomic command in a transaction. This is the simplest solution I could come up with.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1284,9 +1294,110 @@ impl MetastoreService for PostgresqlMetastore {
...
+
+            for subrequest in grouped_subrequests {
+                if !index_metadata.sources.contains_key(&subrequest.source_id) {
+                    return Err(MetastoreError::NotFound(EntityKind::Source {
+                        index_id: index_uid.index_id().to_string(),
+                        source_id: subrequest.source_id.to_string(),
+                    }));
+                }
+
+                let max_shard_opt: Option<PgShard> = sqlx::query_as::<_,PgShard>(
```
kamalesh0406, line 1326: This is done similar to the `FileBackedIndex` to ensure that we have a consistent implementation for all types.

quickwit/quickwit-control-plane/src/model/mod.rs
guilload: That's fine.

quickwit/quickwit-metastore/src/metastore/file_backed_metastore/mod.rs
```diff
@@ -994,9 +1005,13 @@ mod tests {
...
     use crate::metastore::MetastoreServiceStreamSplitsExt;
     use crate::tests::DefaultForTest;
-    use crate::{metastore_test_suite, IndexMetadata, ListSplitsQuery, SplitMetadata, SplitState};
+    use crate::{
+        file_backed_index_shards_test_suite, metastore_test_suite, IndexMetadata, ListSplitsQuery,
+        SplitMetadata, SplitState,
+    };
 
     metastore_test_suite!(crate::FileBackedMetastore);
+    file_backed_index_shards_test_suite!(crate::FileBackedMetastore);
```
guilload, line 1014: What are you trying to achieve here? There's already a `metastore_test_suite!` that runs for both implementations and covers the shard API. It is also possible to turn on/off shard tests independantly (see `quickwit-metastore/src/tests/shard.rs`)

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -541,10 +545,16 @@ impl MetastoreService for PostgresqlMetastore {
         request: DeleteIndexRequest,
     ) -> MetastoreResult<EmptyResponse> {
         let index_uid: IndexUid = request.index_uid.into();
+        sqlx::query("DELETE FROM shards WHERE index_uid = $1")
```
guilload, line 548: We want a transactional behavior.

guilload:
```suggestion
        sqlx::query("DELETE FROM shards WHERE index_uid = $1 AND source_id = $2"")
```

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1284,9 +1294,110 @@ impl MetastoreService for PostgresqlMetastore {
 
     async fn open_shards(
         &mut self,
-        _request: OpenShardsRequest,
+        request: OpenShardsRequest,
     ) -> MetastoreResult<OpenShardsResponse> {
-        unimplemented!("`open_shards` is not implemented for PostgreSQL metastore")
+        let mut subresponses = Vec::with_capacity(request.subrequests.len());
+
+        let grouped_requests: HashMap<IndexUid, Vec<OpenShardsSubrequest>> = request
```
guilload, line 1301: The group by thing does not make sense for the Pg implementation. See this comment in the file-backed metastore implementation:
> We must group the subrequests by `index_uid` to mutate each index only once since each mutation triggers an IO.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
```diff
@@ -1284,9 +1294,110 @@ impl MetastoreService for PostgresqlMetastore {
...
+            let index_metadata_request = IndexMetadataRequest {
+                index_id: None,
+                index_uid: Some(index_uid.to_string()),
+            };
+
+            let index_metadata: IndexMetadata = self
+                .index_metadata(index_metadata_request)
+                .await?
+                .deserialize_index_metadata()
+                .unwrap();
```
guilload, line 1316: No `unwrap` please.

quickwit/quickwit-metastore/src/metastore/postgresql_metastore.rs
guilload, line 1326: This will break in the presence of concurrent requests.

PR #4173, opened 2023-11-21
Changing the shutdown logic for the rest server.

### LINKED ISSUE #4169: HTTP Ingest connection can prevent shutdown
See https://github.com/quickwit-oss/quickwit/pull/4168

fulmicoton:
Instead of relying on the hyper graceful shutdown, which does not shutdown in the presence of existing http connections.

Closes #4169

### DIFF
--- a/quickwit/quickwit-serve/src/rest.rs
+++ b/quickwit/quickwit-serve/src/rest.rs
@@ -158,9 +158,19 @@ pub(crate) async fn start_rest_server(
         rest_listen_addr=?rest_listen_addr,
         "Starting REST server listening on {rest_listen_addr}."
     );
-    let serve_fut = hyper::Server::bind(&rest_listen_addr)
-        .serve(Shared::new(service))
-        .with_graceful_shutdown(shutdown_signal);
+
+    // `graceful_shutdown()` seems to be blocking in presence of existing connections.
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-serve/src/rest.rs
```diff
@@ -158,9 +158,13 @@ pub(crate) async fn start_rest_server(
         rest_listen_addr=?rest_listen_addr,
         "Starting REST server listening on {rest_listen_addr}."
     );
-    let serve_fut = hyper::Server::bind(&rest_listen_addr)
-        .serve(Shared::new(service))
-        .with_graceful_shutdown(shutdown_signal);
+
+    let serve_fut = async move {
```
fmassot, line 162: I'm okay with this. Note that dropping the server will not prevent current connections from working ([ref](https://github.com/hyperium/hyper/issues/2386#issuecomment-754912276))
I would either add a comment there or open a low-priority issue on the REST server graceful shutdown.

fmassot (approved):

quickwit/quickwit-serve/src/rest.rs
fulmicoton: ah good point. I am not sure why but it did fix the test terminating after the rest clients issue.

Issue #4132, opened 2023-11-14
The ES compatible API accepts malformed requests


fmassot:
Example of a request that should return a 400 error:

```
{"term": {"actor.id": {"value": "95077794"}}}
```

This is converted into a match all query.

The correct query is:

```
{"query": {"term": {"actor.id": "95077794"}}}
```

tuziben:
```
{
    "query": {
        "term": {
            "actor.id": {
                "value": "95077794"
            }
        }
    }
}
```
BTW, this is also a valid DSL format

fulmicoton:
Yes ! and we accept both.
https://github.com/quickwit-oss/quickwit/blob/main/quickwit/rest-api-tests/scenarii/es_compatibility/0006-term_query.yaml

I think @fmassot's point is that the doc should probably give the full payload (starting from "query")

fmassot:
We could add `deny_unknown_fields` to the [search body struct](https://github.com/quickwit-oss/quickwit/blob/131f64be0d72f353fc1da6582748978f79ebb4b7/quickwit/quickwit-serve/src/elastic_search_api/model/search_body.rs#L56).

@fulmicoton I'm wondering if you did not use it deliberately, can you confirm?

fulmicoton:
It was not deliberate.

fulmicoton:
Thank you @etolbakov !

PR #4175, opened 2023-11-21
fix: avoid parsing search malformed requests

etolbakov:
### Description

ES compatible API does not accept malformed requests and returns `400` and the following body instead
```
{
    "message": "unknown field `term`, expected one of `from`, `size`, `query`, `sort`, `aggs`, `track_total_hits`, `stored_fields`, `search_after` at line 1 column 7"
}
```

### How was this PR tested?

a unit test has been added, local testing with the change has been done as well

### Related issue
https://github.com/quickwit-oss/quickwit/issues/4132

### DIFF
--- a/quickwit/quickwit-serve/src/elastic_search_api/model/search_body.rs
+++ b/quickwit/quickwit-serve/src/elastic_search_api/model/search_body.rs
@@ -237,11 +237,11 @@ mod tests {
         "#;
 
         let search_body = serde_json::from_str::<SearchBody>(json);
-        assert!(search_body.is_err());
-        assert_eq!(
-            search_body.err().unwrap().to_string(),
-            "unknown field `term`, expected one of `from`, `size`, `query`, `sort`, `aggs`, \
-             `track_total_hits`, `stored_fields`, `search_after` at line 3 column 22"
-        );
+        let error_msg = search_body.unwrap_err().to_string();
+        assert!(error_msg.contains("unknown field `term`"));
+        assert!(error_msg.contains(
+            "expected one of `from`, `size`, `query`, `sort`, `aggs`, `track_total_hits`, \
+             `stored_fields`, `search_after`"
+        ));
     }
 }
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-serve/src/elastic_search_api/model/search_body.rs
```diff
@@ -222,4 +223,25 @@ mod tests {
...
+                     }
+                }
+            }
+        "#;
+
+        let search_body = serde_json::from_str::<SearchBody>(json);
+        assert!(search_body.is_err());
+        assert_eq!(
+            search_body.err().unwrap().to_string(),
+            "unknown field `term`, expected one of `from`, `size`, `query`, `sort`, `aggs`, \
```
fulmicoton, line 243: can we make the assert less brittle?

Check that the error message contains a few keyword "unknown field" for instance

etolbakov: sure, thanks!

fulmicoton:
Can we have a look at other nested structure?

I'm sure some of the query alternative in the query dsl are missing the `deny_unknown_fields` annotation

```
    QueryString(QueryStringQuery),
    Bool(BoolQuery),
    Term(TermQuery),
    Terms(TermsQuery),
    MatchAll(MatchAllQuery),
    MatchNone(MatchNoneQuery),
    Match(MatchQuery),
    MatchBoolPrefix(MatchBoolPrefixQuery),
    MatchPhrase(MatchPhraseQuery),
    MatchPhrasePrefix(MatchPhrasePrefixQuery),
    MultiMatch(MultiMatchQuery),
    Range(RangeQuery),
    Exists(ExistsQuery),
```

etolbakov:
@fulmicoton 
I had a look at it and the good news is that almost all of the `ElasticQueryDslInner` members have `deny_unknown_fields`.
I've added it where I find reasonable. Could you please take a look?
As for `TermQuery`, `MatchNoneQuery`, `MultiMatchQuery` not sure if it's needed.

fulmicoton:
Thank you @etolbakov

fulmicoton (approved):

PR #4179, opened 2023-11-21
Extend Shard API test suite

guilload:
### Description
Added some tests for the `open_shards` API.

### DIFF
--- a/quickwit/quickwit-metastore/src/tests/shard.rs
+++ b/quickwit/quickwit-metastore/src/tests/shard.rs
@@ -20,7 +20,8 @@
 use quickwit_common::rand::append_random_suffix;
 use quickwit_config::{IndexConfig, SourceConfig};
 use quickwit_proto::metastore::{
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-metastore/src/tests/shard.rs
```diff
@@ -95,12 +104,50 @@ pub async fn test_metastore_open_shards<
...
+    };
+    let error = metastore
+        .open_shards(open_shards_request)
+        .await
+        .unwrap_err();
+    assert!(
+        matches!(error, MetastoreError::NotFound(EntityKind::Index { index_id }) if index_id == "index-does-not-exist")
+    );
+
+    // Test source not found.
```
guilload, line 132: @kamalesh0406, I don't think this will be easy to implement for Postgres. Feel free to comment.

kamalesh0406: I feels this looks promising https://docs.rs/sqlx/latest/sqlx/attr.test.html . What are your thoughts?

Issue #4174, opened 2023-11-21
IngestV2 plan apply loop is broken.


fulmicoton:
IngestV2 plan apply loop is broken.

```
IndexingPlanDiff(missing_tasks_by_node_id={"node-1": [IndexingTask { index_uid: "gharchive4:01HFRA32Z532P0P7D6SNG5WPP4", source_id: "_ingest-source", shard_ids: [7] }]}, , unplanned_tasks_by_node_id={"node-1": [IndexingTask { index_uid: "gharchive4:01HFRA32Z532P0P7D6SNG5WPP4", source_id: "_ingest-source", shard_ids: [] }]})
```

my understanding is that we don't expose the list of shards in chitchat yet, and that's where the diff is.

- [ ] fix the serialization format of indexing task in chitchat to allow shards
- [ ] fix the exposition of the indexing task in chitchat to have shard (right now we serialize an empty vec all of the time)

PR #4180, opened 2023-11-22
Fixed the serialize deserialization of the indexing state to include 

fulmicoton:
Fixed the serialize deserialization of the indexing state to include
    shard lists.
    
    Before this, there was no notion of shards and indexers were just
    exposing the number of pipelines per source_uid.
    
    With ingest v2, the list of shard is part of indexing tasks, and the
    control plane wants to diff against it.
    
    The serialization format goes
    `[1,2][3]` to express two pipelines with respectively shard [1,2] and
    `[3]`.
    
    For a source that does not have notion of shards, like kafka, two
    pipelines simply look as follows: `[][]`
    
    Also took it as an opportunity to clean up the code
    and increase the test coverage.
     
    I think we also used to not skip over tombstoned entries

Related with  #4174

### DIFF
--- a/quickwit/quickwit-cluster/src/cluster.rs
+++ b/quickwit/quickwit-cluster/src/cluster.rs
@@ -43,8 +43,8 @@ use tracing::{info, warn};
 use crate::change::{compute_cluster_change_events, ClusterChange};
 use crate::member::{
     build_cluster_member, ClusterMember, NodeStateExt, ENABLED_SERVICES_KEY,
-    GRPC_ADVERTISE_ADDR_KEY, INDEXING_TASK_PREFIX, PIPELINE_METRICS_PREFIX, READINESS_KEY,
-    READINESS_VALUE_NOT_READY, READINESS_VALUE_READY,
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -407,6 +398,137 @@ fn spawn_ready_members_task(
     tokio::spawn(fut);
 }
 
+/// Parsed indexing tasks from the chitchat node state.
```
fmassot, line 401:
```suggestion
/// Parses indexing tasks from the chitchat node state.
```

quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -407,6 +398,137 @@ fn spawn_ready_members_task(
...
+        Err(error) => {
+            warn!(error=%error, "failed to parse pipeline shard list");
+            Vec::new()
+        }
+    }
+}
+
+/// Writes the given indexing tasks in the given node state.
+///
+/// If previous indexing tasks were present in the node state, they are marked for deletion.
```
fmassot, line 454:
```suggestion
/// If previous indexing tasks were present in the node state but were not in the given tasks, they are marked for deletion.
```

fmassot (approved):

PR #4189, opened 2023-11-22
Broadcast shard ingestion rate and state via Chitchat

guilload:
### Description
- Broadcast shard ingestion throughput and state via Chitchat
- Update router shard table (closed shards only)

### How was this PR tested?
- Added unit tests

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/rate_limiter.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/rate_limiter.rs
@@ -66,24 +66,25 @@ pub(super) struct RateLimiter {
 }
 
 impl RateLimiter {
+    /// Creates a new rate limiter from the given settings.
     pub fn from_settings(settings: RateLimiterSettings) -> Self {
         let capacity = settings.burst_limit.as_u64();
 
-        let work = settings.rate_limit.work() as u128;
         let refill_period = settings.refill_period;
-        let rate_limit_period = settings.rate_limit.period();
-        let refill_amount = work * refill_period.as_nanos() / rate_limit_period.as_nanos();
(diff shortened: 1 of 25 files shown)


quickwit/quickwit-ingest/src/ingest_v2/rate_limiter.rs
```diff
@@ -63,27 +63,32 @@ pub(super) struct RateLimiter {
     refill_period: Duration,
     refill_period_micros: u64,
     refill_at: Instant,
+    throughput_num_bytes: ByteSize,
```
fulmicoton, line 66: it used to be more generic (not specific to bytes)

It might be good to rename it.

quickwit/quickwit-proto/src/ingest/mod.rs
```diff
@@ -191,6 +191,25 @@ impl ShardState {
...
+            ShardState::Unspecified => "unspecified",
+            ShardState::Open => "open",
+            ShardState::Unavailable => "unavailable",
+            ShardState::Closed => "closed",
+        }
+    }
+
+    pub fn from_json_str_name(shard_state_str: &str) -> Option<Self> {
+        match shard_state_str {
+            "unspecified" => Some(Self::Unspecified),
```
fulmicoton, line 206: a unit test would be nice here

fulmicoton (approved):

quickwit/quickwit-ingest/src/ingest_v2/rate_limiter.rs
guilload: Yes, good point. I'll take it out into a dedicated `RaterMeter` object.

PR #4198, opened 2023-11-27
append compatible elastic header

tuziben:
### Description

We will get `the client noticed that the server is not Elasticsearch and we do not support this unknown product` error message when querying the Qucikwit with the Golang client.

That's because some Elasticsearch clients will check the response header whether the heads contain `X-Elastic-Product` and the value is `Elasticsearch`

Source code in Elasticsearch clients
- Java https://github.com/elastic/elasticsearch-java/blob/2d0e3a5fe76285f91b28784ba19adf124327d533/java-client/src/main/java/co/elastic/clients/transport/ElasticsearchTransportBase.java#L437
- Python https://github.com/elastic/elasticsearch-py/blob/a6bb5f3808ef8616afde81ebc1dc53f98011a2b8/elasticsearch/_sync/client/_base.py#L325
- Golang https://github.com/elastic/go-elasticsearch/blob/537d09014a37fd1288919bd43eb7f1ab1bd374ae/elasticsearch.go#L364

### How was this PR tested?

test code https://github.com/quickwit-oss/quickwit/compare/es_header?expand=1#diff-ff71f681357dc97850f21fb99bd77c44304e9c20b043c1de8182e48ef206208fR415

### DIFF
--- a/quickwit/quickwit-config/src/node_config/mod.rs
+++ b/quickwit/quickwit-config/src/node_config/mod.rs
@@ -303,6 +303,7 @@ impl Default for JaegerConfig {
 pub struct NodeConfig {
     pub cluster_id: String,
     pub node_id: String,
+    pub enable_elastic_header: bool,
     pub enabled_services: HashSet<QuickwitService>,
     pub rest_listen_addr: SocketAddr,
     pub gossip_listen_addr: SocketAddr,
@@ -375,7 +376,12 @@ impl NodeConfig {
 
     #[cfg(any(test, feature = "testsuite"))]
     pub fn for_test() -> Self {
-        serialize::node_config_for_test()
+        serialize::node_config_for_test(false)
+    }
+
+    #[cfg(any(test, feature = "testsuite"))]
+    pub fn for_test_elastic_header() -> Self {
+        serialize::node_config_for_test(true)
     }
 }
 
--- a/quickwit/quickwit-config/src/node_config/serialize.rs
+++ b/quickwit/quickwit-config/src/node_config/serialize.rs
@@ -59,6 +59,10 @@ fn default_node_id() -> ConfigValue<String, QW_NODE_ID> {
     ConfigValue::with_default(node_id)
 }
 
+fn default_enable_elastic_header() -> ConfigValue<bool, QW_ENABLE_ELSTIC_HEADER> {
+    ConfigValue::with_default(false)
+}
+
 #[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq)]
 struct List(Vec<String>);
 
@@ -163,6 +167,8 @@ struct NodeConfigBuilder {
     cluster_id: ConfigValue<String, QW_CLUSTER_ID>,
     #[serde(default = "default_node_id")]
     node_id: ConfigValue<String, QW_NODE_ID>,
+    #[serde(default = "default_enable_elastic_header")]
+    enable_elastic_header: ConfigValue<bool, QW_ENABLE_ELSTIC_HEADER>,
     #[serde(default = "default_enabled_services")]
     enabled_services: ConfigValue<List, QW_ENABLED_SERVICES>,
     #[serde(default = "default_listen_address")]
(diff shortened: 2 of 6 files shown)


quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -375,7 +376,12 @@ impl NodeConfig {
 
     #[cfg(any(test, feature = "testsuite"))]
     pub fn for_test() -> Self {
-        serialize::node_config_for_test()
+        serialize::node_config_for_test(false)
+    }
+
+    #[cfg(any(test, feature = "testsuite"))]
```
fmassot, line 382: We won't do a lot of test for the header, it's a bit overkill to add this method (see commit on my branch fmassot/es_header)

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -59,6 +59,10 @@ fn default_node_id() -> ConfigValue<String, QW_NODE_ID> {
     ConfigValue::with_default(node_id)
 }
 
+fn default_enable_elastic_header() -> ConfigValue<bool, QW_ENABLE_ELSTIC_HEADER> {
```
fmassot, line 62: Given the usage of the elastic header, I would prefer to not add a new env to manage.

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -163,6 +167,8 @@ struct NodeConfigBuilder {
     cluster_id: ConfigValue<String, QW_CLUSTER_ID>,
     #[serde(default = "default_node_id")]
     node_id: ConfigValue<String, QW_NODE_ID>,
+    #[serde(default = "default_enable_elastic_header")]
+    enable_elastic_header: ConfigValue<bool, QW_ENABLE_ELSTIC_HEADER>,
```
fmassot, line 171: see commit on fmassot/es_header

fmassot (changes requested):
See my comments and https://github.com/quickwit-oss/quickwit/compare/es_header...fmassot/es_header?expand=1

quickwit/quickwit-config/src/node_config/mod.rs
tuziben: I merged your PR

config/quickwit.yaml
```diff
@@ -32,7 +32,9 @@ version: 0.6
 #   2. pass `0.0.0.0` and let Quickwit do its best to discover the node's IP (see `advertise_address`)
 #
 # listen_address: 127.0.0.1
-# rest_listen_port: 7280
+#
+# rest:
+#   listen_port: 7280
```
guilload, line 37: We can add examples of extra headers and CORS here.

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -810,38 +870,74 @@ mod tests {
...
+            assert_eq!(
+                node_config.rest_config.listen_addr.to_string(),
+                "127.0.0.1:1789"
+            );
             assert_eq!(node_config.gossip_listen_addr.to_string(), "127.0.0.1:1889");
             assert_eq!(node_config.grpc_listen_addr.to_string(), "127.0.0.1:1989");
         }
+        {
+            // Check that `rest_listen_port` can still be used for backward compatibility.
+            let node_config = NodeConfigBuilder {
```
guilload, line 922: Nit: this probably should go in a dedicated test. This one is already quite lengthy.

quickwit/quickwit-serve/src/rest.rs
```diff
@@ -525,4 +553,59 @@ mod tests {
             Box::pin(fut)
         }
     }
+
+    #[tokio::test]
+    async fn test_additional_headers() {
```
guilload, line 558:
```suggestion
    async fn test_extra_headers() {
```

quickwit/quickwit-serve/src/elastic_search_api/mod.rs
```diff
@@ -363,43 +359,4 @@ mod tests {
             .await;
         assert_eq!(resp.status(), 200);
     }
-
-    #[tokio::test]
-    async fn test_es_compat_cluster_info_handler() {
```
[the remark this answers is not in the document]
fmassot: indeed :/
will revert that

PR #4199, opened 2023-11-27
Issue/4191 added pipeline uid

fulmicoton:
Introducing pipeline uid in indexing tasks/ indexing pipelines.
    
The scheduling algorithm is for the most part unchanged.
Only the allocation of shard to pipelines is modified.

Currently the logic yields similar results as before. In particular, the scheduling algorithm
may in some case, move shard from one pipeline to another.
In the future, we will prevent this behavior and instead close/open new shards.

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexing_service.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexing_service.rs
@@ -49,7 +49,7 @@ use quickwit_proto::indexing::{
 use quickwit_proto::metastore::{
     IndexMetadataRequest, ListIndexesMetadataRequest, MetastoreService, MetastoreServiceClient,
 };
-use quickwit_proto::types::{IndexId, IndexUid};
+use quickwit_proto::types::{IndexId, IndexUid, PipelineUid};
 use quickwit_storage::StorageResolver;
 use serde::{Deserialize, Serialize};
 use tokio::sync::Semaphore;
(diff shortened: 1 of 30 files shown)


quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -112,8 +125,7 @@ pub struct IndexingService {
     ingest_api_service_opt: Option<Mailbox<IngestApiService>>,
     ingester_pool: IngesterPool,
     storage_resolver: StorageResolver,
-    indexing_pipelines:
-        HashMap<IndexingPipelineId, (Mailbox<IndexingPipeline>, ActorHandle<IndexingPipeline>)>,
+    indexing_pipelines: HashMap<PipelineUid, PipelineHandle>,
```
guilload, line 128: Nice.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -472,65 +483,69 @@ impl IndexingService {
...
-            *pipeline_ord += 1;
-            updated_pipeline_ids.insert(pipeline_id);
-        }
-
-        let running_pipeline_ids: HashSet<IndexingPipelineId> =
-            self.indexing_pipelines.keys().cloned().collect();
+        let pipelines_uid_in_plan: FnvHashSet<PipelineUid> = physical_indexing_plan_request
+            .indexing_tasks
+            .iter()
+            .flat_map(|indexing_task| indexing_task.pipeline_uid.into_iter())
```
guilload, line 489: Wherever we flatmap on the `pipeline_uid` option, we silently skip the empty, which in practice should not happen.

I wonder if instead we want to add a `pipeline_uid()` accessor that `unwrap`s or `unwrap_or_default`. The former will let us know right away if something goes wrong at the risk of panicking. The latter will be also silent but will at least appear in the logs. WDYT?

guilload (approved):

PR #4205, opened 2023-11-28
Create and init shards on ingesters via control plane

guilload:
### Description
Per title.

### How was this PR tested?
- Added unit tests
- Updated unit tests

### DIFF
--- a/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
+++ b/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
@@ -20,7 +20,7 @@
 use std::fmt;
 use std::time::Duration;
 
-use fnv::FnvHashSet;
+use fnv::{FnvHashMap, FnvHashSet};
 use itertools::Itertools;
 use quickwit_common::{PrettySample, Progress};
 use quickwit_ingest::IngesterPool;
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -330,6 +333,38 @@ impl IngestController {
             failures: get_or_create_open_shards_failures,
         })
     }
+
+    /// Calls init shards on the leaders hosting newly opened shards.
+    async fn init_shards(
```
guilload, line 338: I will improve error handling and testing in a follow-up PR.

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -218,14 +219,17 @@ impl Ingester {
         Ok(())
     }
 
-    async fn create_shard<'a>(
+    async fn init_primary_shard(
```
fulmicoton, line 222: Can you add a comment explain what `init_primary_shard` does? (create the queue on the leader and possibly the followed and creates the replication stream)

fulmicoton (approved):

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
guilload: Will do.

PR #4208, opened 2023-11-28
Implement shard auto-scaling

guilload:
### Description
Opening as a draft because I have yet to fix some minor issues and write some more tests.

### How was this PR tested?
In progress.


quickwit/quickwit-control-plane/src/control_plane_model.rs
```diff
@@ -18,37 +18,74 @@
...
 use quickwit_proto::metastore::{
     self, EntityKind, ListIndexesMetadataRequest, ListShardsSubrequest, MetastoreError,
     MetastoreService, MetastoreServiceClient, SourceType,
 };
-use quickwit_proto::types::{IndexId, IndexUid, NodeId, NodeIdRef, ShardId, SourceId, SourceUid};
+use quickwit_proto::types::{IndexId, IndexUid, NodeId, ShardId, SourceId, SourceUid};
 use serde::Serialize;
 use tracing::{error, info, warn};
 
+const AUTOSCALING_RATE_LIMITER_SETTINGS: RateLimiterSettings = RateLimiterSettings {
```
guilload, line 44: These values are somewhat arbitrary right now and will require some tuning. It also downscales the number of shards very aggressively; maybe we want one rate limiter for scaling out and another for scaling in?

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -17,28 +17,41 @@
...
 use tokio::time::timeout;
 use tracing::{info, warn};
 
-use crate::control_plane_model::ControlPlaneModel;
+use crate::control_plane_model::{ControlPlaneModel, ShardEntry, ShardStats};
+
+const MAX_SHARD_INGESTION_THROUGHPUT_MIB_PER_SEC: f32 = 5.;
+
+/// Threshold in MiB/s above which we increase the number of shards.
+const SCALE_OUT_SHARDS_THRESHOLD_MIB_PER_SEC: f32 =
```
fulmicoton, line 49: isn't the terminology scale down/ scale up?

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -191,6 +208,25 @@ impl IngestController {
...
+        local_shards_update: LocalShardsUpdate,
+        model: &mut ControlPlaneModel,
+        progress: &Progress,
+    ) {
+        let shard_stats = model.update_shards(
+            &local_shards_update.source_uid,
+            &local_shards_update.shard_infos,
+        );
+        if shard_stats.avg_ingestion_rate >= SCALE_OUT_SHARDS_THRESHOLD_MIB_PER_SEC {
+            self.try_scale_out_shards(local_shards_update.source_uid, shard_stats, model, progress)
```
fulmicoton, line 222: i think there is an important corner case if you have only one shard.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -365,6 +417,147 @@ impl IngestController {
             }
         }
     }
+
+    /// Attempts to increase the number of shards.
```
fulmicoton, line 421: It is worth explaining the throttling happening here.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -365,6 +417,147 @@ impl IngestController {
...
+    async fn try_scale_out_shards(
+        &mut self,
```
fulmicoton, line 423: can we have counters about the number of shards per source or do you think this will be too much?

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
guilload: It's handled in `try_scale_out_shards`, but I'll move the check here.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
guilload: A gauge per source seems like a good idea.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -303,7 +351,8 @@ impl IngestController {
                 .protect_future(self.metastore.open_shards(open_shards_request))
                 .await?;
 
-            self.init_shards(&open_shards_response, progress).await;
+            // TODO: Handle failures.
```
guilload, line 354: See #4217.

PR #4219, opened 2023-11-30
avoid unneeded copy on touching but non-overlaping byte ranges

trinity-1686a:
### Description

when a `ByteRangeCache` has to store bytes for the range 0..100_000 and the range 100_000..100_100, it currently merged both range, issuing a memcpy which takes up to a few milliseconds. In practice this is almost never useful, as future get are unlikely to make a request that cross boundaries. This PR removes that merging on range that touch but do not overlap. When fetching from cache, if the cache realize it can only server a query by merging multiple ranges. This impact mostly the warmup of fastfields. On a sample `find_trace_ids` request, done on a single searcher over ~1b spans, it lowered query time from ~5.8s to ~4.1s

### How was this PR tested?

added unit test, manually checked there is a performance improvement

### DIFF
--- a/quickwit/quickwit-storage/src/cache/byte_range_cache.rs
+++ b/quickwit/quickwit-storage/src/cache/byte_range_cache.rs
@@ -83,6 +83,8 @@ impl<T: 'static + ToOwned + ?Sized + Ord> NeedMutByteRangeCache<T> {
         let key = CacheKey::from_borrowed(tag, byte_range.start);
         let (k, v) = if let Some((k, v)) = self.get_block(&key, byte_range.end) {
             (k, v)
+        } else if let Some((k, v)) = self.merge_ranges(&key, byte_range.end) {
+            (k, v)
         } else {
             self.cache_counters.misses_num_items.inc();
             return None;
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -106,13 +109,13 @@ impl<T: 'static + ToOwned + ?Sized + Ord> NeedMutByteRangeCache<T> {
...
+        // try to find a block with which we overlap (and not just touch)
         let start_key = CacheKey::from_borrowed(tag.borrow(), byte_range.start);
-        let end_key = CacheKey::from_borrowed(tag.borrow(), byte_range.end);
-
         let first_matching_block = self
-            .get_block(&start_key, byte_range.start)
+            .get_block(&start_key, byte_range.start + 1)
             .map(|(k, _v)| k);
 
+        let end_key = CacheKey::from_borrowed(tag.borrow(), byte_range.end - 1);
```
fulmicoton, line 118: my understanding is that with this, we remove (as opposed to merge) any data that is overlapping (not just is included in) with the slice we are trying to put. 

Is my understanding correct?
If so can we comment this more clearly?

possibly isolate in a different method:
`fn remove_overlapping_slices(&mut self, ....)`

fulmicoton: yes: correction.
we merge any data that is overlapping (not just touches) with the slice we are trying to put.

could we comment what we are doing, or even better isolate it into a method?

fulmicoton (approved):

PR #4221, opened 2023-11-30
Update config with additional headers.

fmassot:


quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -41,6 +42,14 @@ use crate::{ConfigFormat, MetastoreConfigs};
 
 pub const DEFAULT_QW_CONFIG_PATH: &str = "config/quickwit.yaml";
 
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, Default)]
+#[serde(deny_unknown_fields)]
+pub struct RestConfig {
+    pub cors_allow_origins: Vec<String>,
+    #[serde(with = "http_serde::header_map")]
+    pub additional_headers: HeaderMap,
```
guilload, line 50: GitHub code search fight votes for `extra`:

```suggestion
    pub extra_headers: HeaderMap,
```

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -370,11 +368,13 @@ pub fn node_config_for_test() -> NodeConfig {
         .to_path_buf();
     let metastore_uri = default_metastore_uri(&data_dir_uri);
     let default_index_root_uri = default_index_root_uri(&data_dir_uri);
-
+    let rest_config = RestConfig {
+        cors_allow_origins: Vec::new(),
+        additional_headers: HeaderMap::new(),
+    };
```
guilload, line 374:
```suggestion
    let rest_config = RestConfig::default();
```

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -370,11 +368,13 @@ pub fn node_config_for_test() -> NodeConfig {
         .to_path_buf();
     let metastore_uri = default_metastore_uri(&data_dir_uri);
     let default_index_root_uri = default_index_root_uri(&data_dir_uri);
-
+    let rest_config = RestConfig {
```
guilload, line 371: We can move `rest_listen_addr` in there and stay backward compatible by duplicating the field in `RestConfig`, and copying from the `NodeConfig` if it's empty while displaying a deprecation warning for `1.0`. We should return an error if both are set.

fmassot: excellent idea! I was not sure of how to do it

quickwit/quickwit-config/src/node_config/mod.rs
tuziben: `extra`  +1

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -340,6 +352,41 @@ impl Default for NodeConfigBuilder {
     }
 }
 
+#[serde_with::serde_as]
+#[derive(Debug, Deserialize, PartialEq, Default)]
+#[serde(deny_unknown_fields)]
+struct RestConfigBuilder {
+    #[serde(default)]
+    listen_port: Option<u16>,
```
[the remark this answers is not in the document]
fmassot: Good question.

This was motivated to handle the case where the config still defines the old `rest_listen_port`.

To check if there is a conflict, I need to check if a value is set in `rest_listen_port` and `rest.listen_port` without knowing if the env variable was set or not or without any default value. Thus the option and the custom resolving. 

My plan was using directly the `ConfigValue` once we remove the `rest_listen_port`.

PR #4223, opened 2023-12-01
Issue/4184 assign shard logic

fulmicoton:


quickwit/quickwit-indexing/src/source/ingest/mod.rs
```diff
@@ -116,10 +116,10 @@ impl ClientId {
...
     Active,
-    // We have emitted all documents of the pipeline until EOF.
-    // Disclaimer: a complete status does not mean that all documents have been indexed.
-    // Some document might be still travelling in the pipeline, and may not have been published
-    // yet.
+    // We have received all documents from the stream. Note they
+    // are not necessarily published yet.
+    EofReached,
+    // All documents have been indexed AND published.
     Complete,
```
fulmicoton, line 123: @guilload... `Complete` was actually used to dismiss fetch error. It does not seem too silly actually.

guilload (approved):

quickwit/quickwit-indexing/src/source/ingest/mod.rs
fulmicoton: Yes I think we can keep it.

PR #4235, opened 2023-12-04
Populates the control plane model will closed and open shards

### LINKED ISSUE #4176: Control plane should delete closed shard on startup.
When loading shards from the metastore on startup, the controlplane currently only get open shards from the metastore.

Closed shards are therefore "leaking".

fulmicoton:
as opposed to only open ones.

Closes shards are not necessarily entirely indexed yet. With this change, after a reload, the closed shards will be scheduling for indexing as usual, and will get cleaned up once they reach eof.

Closes #4176


guilload (approved):
Can you update the `load_from_metastore` unit test?

PR #4238, opened 2023-12-04
Delete shards on truncate EOF

guilload:
### Description
1. Change the signature of fetch stream, which is now a `oneof` of `{FetchPayload, FetchEof}`
1. Change `Position::Eof` to `Position::Eof(Option<Offset>)`
1. Delete shards on truncate `EOF` and stop appending `EOF` records

- `1.` Generates more boilerplate but is safer because the compiler forces us to handle `EOF` messages explicitly. I caught two bugs.
- `2.` Made my life easier when debugging some tests because we no longer lose the offset when a position is at `EOF`.

### How was this PR tested?
Updated unit tests

### DIFF
--- a/quickwit/quickwit-indexing/src/source/ingest/mod.rs
+++ b/quickwit/quickwit-indexing/src/source/ingest/mod.rs
@@ -22,9 +22,10 @@ use std::fmt;
 use std::sync::Arc;
 use std::time::Duration;
 
-use anyhow::{bail, Context};
+use anyhow::Context;
 use async_trait::async_trait;
 use fnv::FnvHashMap;
+use itertools::Itertools;
 use quickwit_actors::{ActorExitStatus, Mailbox};
 use quickwit_common::pubsub::EventBroker;
 use quickwit_common::retry::RetryParams;
@@ -33,7 +34,8 @@ use quickwit_ingest::{
 };
 use quickwit_metastore::checkpoint::{PartitionId, SourceCheckpoint};
 use quickwit_proto::ingest::ingester::{
-    FetchResponseV2, IngesterService, TruncateShardsRequest, TruncateShardsSubrequest,
+    fetch_message, FetchEof, FetchPayload, IngesterService, TruncateShardsRequest,
+    TruncateShardsSubrequest,
 };
 use quickwit_proto::metastore::{
     AcquireShardsRequest, AcquireShardsSubrequest, AcquireShardsSubresponse, MetastoreService,
@@ -42,6 +44,7 @@ use quickwit_proto::metastore::{
 use quickwit_proto::types::{
     IndexUid, NodeId, Position, PublishToken, ShardId, SourceId, SourceUid,
 };
+use serde::Serialize;
 use serde_json::json;
 use tokio::time;
 use tracing::{debug, error, info, warn};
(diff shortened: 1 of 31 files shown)


quickwit/quickwit-indexing/src/source/ingest/mod.rs
```diff
@@ -116,10 +117,7 @@ impl ClientId {
 enum IndexingStatus {
     #[default]
     Active,
-    // We have emitted all documents of the pipeline until EOF.
-    // Disclaimer: a complete status does not mean that all documents have been indexed.
-    // Some document might be still travelling in the pipeline, and may not have been published
-    // yet.
+    ReachedEof,
```
fulmicoton, line 120: It will show up after you rebase, but please add the comment in there.
We don't want any "floating semantics".

quickwit/quickwit-indexing/src/source/ingest/mod.rs
```diff
@@ -432,13 +462,22 @@ impl Source for IngestSource {
...
-                        break;
+                        if batch_builder.num_bytes >= BATCH_NUM_BYTES_LIMIT {
+                            break;
+                        }
                     }
-                }
+                    Some(fetch_message::Message::Eof(fetch_eof)) => {
+                        self.process_fetch_eof(&mut batch_builder, fetch_eof)?;
+                    }
+                    None => {
```
fulmicoton, line 476: When does this happen? Is it normal behavior or not?

quickwit/quickwit-ingest/src/ingest_v2/mrecord.rs
```diff
@@ -37,14 +38,10 @@ const DOC_HEADER_V0: &[u8; MRECORD_HEADER_LEN] = &[HeaderVersion::V0 as u8, 0];
...
 
-/// `Eof` header v0 composed of the header version and the `Eof = 2` record type.
-const EOF_HEADER_V0: &[u8; MRECORD_HEADER_LEN] = &[HeaderVersion::V0 as u8, 2];
-
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum MRecord {
     Doc(Bytes),
     Commit,
-    Eof,
     Unknown,
```
fulmicoton, line 45:
```suggestion
    Unknown, 
```

Unknown is, I believe, an error here.
We can remove it and return an Option<Self> to decode.

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -199,22 +197,47 @@ impl Ingester {
         if queue_ids.is_empty() {
             return Ok(());
         }
-        info!("closing {} shard(s)", queue_ids.len());
+        info!("recovering {} shard(s)", queue_ids.len());
```
fulmicoton, line 200: I think the function itself could use a comment. I cannot add a CR comment on that but something like 
```
/// During the initialization of the ingester, we list all of the queues stored in
/// mrecordlog. Empty queues are simply deleted, while queues that are not empty,
/// are recovered, but closed. 
///
/// Closed means they are in a read-only state: available for their
/// indexing to be resumed, but not available to receive more records.
```

fulmicoton (approved):

quickwit/quickwit-ingest/src/ingest_v2/mrecord.rs
guilload: Yes, we can remove this variant.

quickwit/quickwit-proto/src/ingest/ingester.rs
```diff
@@ -35,6 +59,20 @@ impl FetchResponseV2 {
             0
         }
     }
+
+    pub fn from_position_exclusive(&self) -> Position {
+        self.from_position_exclusive.clone().unwrap_or_default()
```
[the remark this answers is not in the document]
guilload: Yes, these fields are required and should never be empty.

quickwit/quickwit-indexing/src/source/ingest/mod.rs
guilload: It should never happen. It's a by-product of the `oneof`.

PR #4239, opened 2023-12-05
Added json/ip address to the doc.

fulmicoton:
Added more depth to the TOC so that people can jump into the field type they are interested in.


fmassot:
I was not aware of `toc_max_heading_level`, nice!

Ideally I think we want to move the doc mapping in a dedicated page.

fmassot (approved):

PR #4240, opened 2023-12-05
Truncate shards via Chitchat

guilload:
### Description
Ingestes subscribe to `ShardPositionsUpdate` received via Chitchat and truncate or delete shards if necessary.

### How was this PR tested?
Added unit test


quickwit/quickwit-indexing/src/source/ingest/mod.rs
```diff
@@ -305,8 +305,8 @@ impl IngestSource {
             Vec<TruncateShardsSubrequest>,
         > = FnvHashMap::default();
 
-        for (shard_id, to_position_exclusive) in truncate_positions {
-            if matches!(to_position_exclusive, Position::Beginning) {
+        for (shard_id, truncate_position) in truncate_positions {
```
fulmicoton, line 308: the to_*_exclusive part was actually helpful.

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -835,6 +829,78 @@ impl IngesterService for Ingester {
     }
 }
 
+impl IngesterState {
+    /// Truncates the shard identified by `queue_id` up to `truncate_position_inclusive` only if the
+    /// current truncation position of the shard is smaller.
+    async fn truncate_shard(&mut self, queue_id: &QueueId, truncate_position_inclusive: Position) {
+        if let Some(truncate_offset_inclusive) = truncate_position_inclusive.as_u64() {
```
fulmicoton, line 836:
```suggestion
        let Some(truncate_offset_inclusive) = truncate_position_inclusive.as_u64() {
             return;
        }; 
```

early-return style is nice to avoid that kind of nesting.

fulmicoton (approved):

quickwit/quickwit-indexing/src/source/ingest/mod.rs
guilload: Added doc to `TruncateShardsSubrequest` and renamed to `truncate_up_to_position_inclusive`.

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
guilload: Replaced with a bunch of early returns. This snippet is a good contender for if let chains when stabilized.

Issue #3527, opened 2023-06-07
Support elasticsearch field capabilities endpoint


fmassot:
[Elasticsearch doc ref](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-field-caps.html)

Like we do for search, we want to have a gRPC endpoint that provides JSON information that makes sense in Quickwit (using quickwit lingua). The gRPC endpoints does NOT offer a field filter. It returns the data for everything.
The elastic search endpoint will then be a transcription of the field capabilities protobuf into the Elasticsearch response.

Implementation wise, we want all of the necessary information, in an extra file of the bundle, (probably in protobuf format + compressed). If it is judged not (efficient + well compressed) enough. `SSTable -> Protobuf` is also a possibility.
The packager should be in charge of populating that data.

The field capabilities endpoint would then fetch the data for all of the splits and merge it.
Finally we want to cache locally this data, a bit like we cache the hotcache today.

We don't need to bother reducing the cost of merge for the moment.

- [x] spec the data that should be added to the bundle
- [x] implement the logic to build it in the packager
- [x] implement the logic to merge the field capability data
- [x] implement the gRPC endpoint
- [x] implement the elasticsearch REST endpoint facade
- [x] add storage caching at the storage level for the field capability data

PSeitz:
The field capabilities API returns information for all indices.
What information does the UI require?

In a scenario with a lot of splits but few fields, the merging could become relative expensive. (Need some benchmarks/or more info about the merge logic)
In that case, the caching could store some pre-aggregate, that would be invalidated if a split in an aggregation would get removed.

fulmicoton:
> The field capabilities API returns information for all indices.

Not the `GET /<target>/_field_caps?fields=<fields>` endpoint.

Let's skip the preaggregate for the moment maybe, and do something simple.
If you use sstable merging should be fast enough.


E.g. 
- we can either precompute some datastruct in packaging phase OR if the payload is too large, just rely on the leaf cache (currently called partial result cache).
- Only consider the last N splits (N=100) and abort merging and return results right away if it takes too long.

To be consistent with the rest of quickwit, we should use the capability as defined in the current docmapper (not the one of the split).
So if we have a fast field in the dynamic field named "age" in split 0, that split was not a fast field, but is a fastfield in the current doc mapper, it should be considered a fast field.

PSeitz:
## ES Requirements

[...]

PR #4242, opened 2023-12-06
add list_fields api

PSeitz:
- [ ] Retry
- [x] List fields from tantivy
- [x] Fallback to schema if field list does not exist
- [x] `{index_id}/_field_caps` URL
- [ ] `/_field_caps` URL
- [x] `fields` filter (`my-index-000001/_field_caps?fields=rating`)
- [x] API tests
- [x] Caching (very simple list field cache)
- [x] Parallel fetch of field list file

```
➜  quickwit-indices curl -XGET http://localhost:7280/api/v1/_elastic/hdfs/_field_caps -H 'Content-Type: application/json' -d ''
{
  "indices": [
    "hdfs"
  ],
  "fields": {
    "_dynamic.attributes.class": {
      "keyword": {
        "metadata_field": false,
        "searchable": true,
        "aggregatable": true,
        "indices": [
          "hdfs"
        ]
      },
      "text": {
        "metadata_field": false,
        "searchable": true,
        "aggregatable": true,
        "indices": [
          "hdfs"
        ]
      }
    },
    "_field_presence": {
      "long": {
        "metadata_field": false,
        "searchable": true,
        "aggregatable": false,
        "indices": [
          "hdfs"
        ]
      }
    },
    "body": {
      "text": {
        "metadata_field": false,
        "searchable": true,
        "aggregatable": false,
        "indices": [
          "hdfs"
        ]
      }
    },
    "timestamp": {
      "date_nanos": {
        "metadata_field": false,
        "searchable": true,
        "aggregatable": true,
        "indices": [
          "hdfs"
        ]
      }
    },
    "tenant_id": {
      "long": {
        "metadata_field": false,
        "searchable": true,
        "aggregatable": false,
        "indices": [
          "hdfs"
        ]
      }
    },
    "resource.$facet:service": {
      "text": {
        "metadata_field": false,
        "searchable": true,
        "aggregatable": false,
        "indices": [
          "hdfs"
        ]
      }
    },
    "severity_text": {
      "keyword": {
        "metadata_field": false,
        "searchable": true,
        "aggregatable": true,
        "indices": [
          "hdfs"
        ]
      },
      "text": {
        "metadata_field": false,
        "searchable": true,
        "aggregatable": true,
        "indices": [
          "hdfs"
        ]
      }
    }
  }
}⏎                                                                                                                                                                                                                               
```


quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -111,6 +115,57 @@ message ReportSplitsRequest {
...
+  // Optional limit query to a list of fields
+  // Wildcard expressions are supported.
+  repeated string fields = 4;
+}
+
+message ListFieldsResponse{
+  repeated ListFieldsEntryResponse fields = 1;
+}
+
+message ListFieldsEntryResponse{
```
fulmicoton, line 151:
```suggestion
message ListFieldsEntryResponse {
```

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -111,6 +115,57 @@ message ReportSplitsRequest {
...
+  // Index split ids to apply the query on.
+  // This ids are resolved from the index_uri defined in the search_request.
+  repeated SplitIdAndFooterOffsets split_offsets = 3;
+
+  // Optional limit query to a list of fields
+  // Wildcard expressions are supported.
+  repeated string fields = 4;
+}
+
+message ListFieldsResponse{
```
fulmicoton, line 147:
```suggestion
message ListFieldsResponse {
```

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -111,6 +115,57 @@ message ReportSplitsRequest {
...
+  repeated ListFieldsEntryResponse fields = 1;
+}
+
+message ListFieldsEntryResponse{
+  // The field name
+  string field_name = 1;
+  // The tantivy field type
+  int32 field_type = 2;
+  // The index ids the field exists
```
fulmicoton, line 156: I don't understand the phrase.

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -111,6 +115,57 @@ message ReportSplitsRequest {
...
+
+message ListFieldsResponse{
+  repeated ListFieldsEntryResponse fields = 1;
+}
+
+message ListFieldsEntryResponse{
+  // The field name
+  string field_name = 1;
+  // The tantivy field type
+  int32 field_type = 2;
```
fulmicoton, line 155: Can you comment on how it is mapped? Why signed int?
 
```suggestion
  int32 field_type = 2;
```

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -111,6 +115,57 @@ message ReportSplitsRequest {
...
+  // The index ids the field exists
+  repeated string index_ids = 3;
+  // True means the field is searchable (indexed) in at least some indices.
+  // False means the field is not searchable in any indices.
+  bool searchable = 4;
+  // True means the field is aggregatable (fast) in at least some indices.
+  // False means the field is not aggregatable in any indices.
```
fulmicoton, line 162: do you really mean indices in all of this proto, or did you mean split ids?

quickwit/quickwit-search/src/lib.rs
```diff
@@ -165,7 +166,7 @@ fn extract_split_and_footer_offsets(split_metadata: &SplitMetadata) -> SplitIdAn
     }
 }
 
-/// Extract the list of relevant splits for a given search request.
+/// Extract the list of relevant splits for a given query.
```
fulmicoton, line 169: I prefer request here.
We tend to talk about `query` to describe the predicate that selects documents within an index.

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -0,0 +1,623 @@
...
+) -> anyhow::Result<impl Iterator<Item = io::Result<ListFieldsEntryResponse>>> {
+    // TODO: Add fancy caching
+    let (_, split_bundle) =
+        open_split_bundle(searcher_context, index_storage, split_and_footer_offsets).await?;
+
+    let serialized_split_fields = split_bundle
+        .get_all(Path::new(SPLIT_FIELDS_FILE_NAME))
+        .await?;
+    let serialized_split_fields_len = serialized_split_fields.len();
+    let iter = read_split_fields(serialized_split_fields).map_err(|err| {
```
fulmicoton, line 61:
```suggestion
    let iter = read_split_fields(serialized_split_fields).map_err(|err| {
```
proper variable names please.

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -0,0 +1,623 @@
...
+        anyhow::anyhow!(
```
fulmicoton, line 62: nitpick: `.with_context(|| {})` might make more sense here.

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -0,0 +1,623 @@
...
+    })?;
+
+    Ok(iter.map(move |metadata| {
+        metadata.map(|metadata| ListFieldsEntryResponse {
+            field_name: metadata.field_name,
+            field_type: metadata.typ.to_code() as i32,
+            index_ids: vec![index_id.to_string()],
+            searchable: metadata.indexed,
+            aggregatable: metadata.fast,
+            non_searchable_index_ids: Vec::new(),
```
fulmicoton, line 76: I don't understand this contract. 
So you return io::Result and lose the info of the missing split ids? 
What are non_searchable_index_ids ?

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -0,0 +1,623 @@
...
+        stored: true,
+    }
+}
+
+/// `current_group` needs to contain at least one element.
+/// The group needs to be of the same field name and type.
+fn merge_same_field_group(
+    current_group: &mut Vec<ListFieldsEntryResponse>,
+) -> ListFieldsEntryResponse {
+    // Make sure all fields have the same name and type in current_group
```
fulmicoton, line 99: thanks

quickwit/quickwit-search/src/list_fields.rs
fulmicoton: When do we return an ioError as an item.

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -0,0 +1,623 @@
...
+    let mut non_searchable_index_ids = if searchable {
+        // We need to combine the non_searchable_index_ids + index_ids where searchable is set to
+        // false (as they are all non_searchable)
+        current_group
+            .iter()
+            .flat_map(|entry| {
+                if !entry.searchable {
+                    entry.index_ids.iter().map(Clone::clone)
+                } else {
+                    entry.non_searchable_index_ids.iter().map(Clone::clone)
```
fulmicoton, line 123:
```suggestion
                    entry.non_searchable_index_ids.iter().cloned()
```

quickwit/quickwit-serve/src/elastic_search_api/model/field_capability.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+struct FieldCapabilityResponse {
+    indices: Vec<String>,
+    fields: HashMap<String, HashMap<String, FieldCapabilityEntryResponse>>,
+}
+#[derive(Serialize, Deserialize, Debug)]
+struct FieldCapabilityFieldTypesResponse {
```
fulmicoton, line 11: this is not used yet is it?

PSeitz: yes, not yet used

quickwit/quickwit-proto/protos/quickwit/search.proto
PSeitz: I added a comment (`Type::to_code()`) and switched to u32

quickwit/quickwit-proto/protos/quickwit/search.proto
PSeitz: yes indices. There is no use for split ids in the API currently

quickwit/quickwit-search/src/list_fields.rs
PSeitz: Actually I'm not sure what to do with the error, currently we abort and return. This is happening in the merge function. I guess we want to return the failed split_ids for a retry.
```rust
let entry = entry.map_err(|err| crate::error::SearchError::Internal(format!("{:?}", err)))?; // TODO: No early return on error
```

`non_searchable_index_ids` are exemption list of indexes if `searchable: true`. I took this from the elastic search API.

quickwit/quickwit-serve/src/elastic_search_api/model/field_capability.rs
PSeitz: It's used now

quickwit/quickwit-search/Cargo.toml
```diff
@@ -50,6 +50,7 @@ quickwit-opentelemetry = { workspace = true }
 quickwit-proto = { workspace = true }
 quickwit-query = { workspace = true }
 quickwit-storage = { workspace = true }
+quickwit-indexing = { workspace = true }
```
fulmicoton, line 53: do we really need this?

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -0,0 +1,696 @@
...
+use tantivy::FieldMetadata;
+
+use crate::leaf::open_split_bundle;
+use crate::service::SearcherContext;
+use crate::{list_relevant_splits, ClusterClient, SearchError, SearchJob};
+
+/// Get the list of splits for the request which we need to scan.
+pub async fn get_fields_from_split<'a>(
+    searcher_context: &SearcherContext,
+    index_id: String,
```
fulmicoton, line 52:
```suggestion
    index_id: &str,
```

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -0,0 +1,696 @@
...
+
+    if current_group.len() == 1 {
+        return current_group.pop().unwrap();
+    }
+    let metadata = &current_group.last().unwrap();
+    let searchable = current_group.iter().any(|entry| entry.searchable);
+    let aggregatable = current_group.iter().any(|entry| entry.aggregatable);
+    let field_name = metadata.field_name.to_string();
+    let field_type = metadata.field_type;
+    let mut non_searchable_index_ids = if searchable {
```
fulmicoton, line 146: Can you add the type here?
```suggestion
    let mut non_searchable_index_ids = if searchable {
```

Also if perf do not matter, `BTreeSet<&str>` is a simple way to inform users that this is sorted and unique.

PSeitz: This is building the protobuf API response, which requires `Vec`

quickwit/rest-api-tests/scenarii/field_capabilities/0001-field-capabilities.yaml
```diff
@@ -0,0 +1,224 @@
+# Test date _field_caps API
+method: [GET]
+engines:
+  - quickwit
```
fulmicoton, line 4: how far are we from elastic's api?

quickwit/quickwit-serve/Cargo.toml
```diff
@@ -32,6 +32,7 @@ serde_qs = { workspace = true }
 serde_with = { workspace =  true }
 termcolor = { workspace = true }
 thiserror = { workspace = true }
+tantivy = { workspace = true }
```
fulmicoton, line 35: same question, can we avoid this?

quickwit/rest-api-tests/scenarii/field_capabilities/0001-field-capabilities.yaml
PSeitz: The most important features are supported, some minor parameters are not implemented yet

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -111,6 +115,58 @@ message ReportSplitsRequest {
...
+
+message ListFieldsResponse {
+  repeated ListFieldsEntryResponse fields = 1;
+}
+
+message ListFieldsEntryResponse {
+  // The field name
+  string field_name = 1;
+  // The tantivy field type mapped via `to_code`.
+  uint32 field_type = 2;
```
fulmicoton, line 156: make it a proto enum to remove the dependency in quickwit-serve.

quickwit/quickwit-serve/Cargo.toml
PSeitz: we need to convert the value to tantivy::Type

quickwit/quickwit-proto/protos/quickwit/search.proto
PSeitz: I would rather reuse the existing code

fulmicoton (changes requested):
The code goes

- serialize list of fields as a file artifact with code living in quickwit-indexing, and using tantivy's types.
- convert artifact to protobuf

Would using protobuf to begin lead to better code?
We could remove code, and drop the quickwit-search -> quickwit-indexing dependency

PSeitz:
> The code goes
> 
>     * serialize list of fields as a file artifact with code living in quickwit-indexing, and using tantivy's types.
> 
>     * convert artifact to protobuf
> 
> 
> Would using protobuf to begin lead to better code? We could remove code, and drop the quickwit-search -> quickwit-indexing dependency

I don't think this would be better code, since this would be creating duplicated code. The existing type does exactly what we want and tantivy is already in the dependency tree.
The quickwit-search -> quickwit-indexing dependency has nothing to do with the tantivy types, this is because field de/serialziation is in `models/list_fields.rs`

quickwit/quickwit-proto/Cargo.toml
```diff
@@ -25,6 +25,7 @@ prost-types = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 sqlx = { workspace = true, optional = true }
+tantivy = { workspace = true }
```
fulmicoton, line 28: no.

quickwit/quickwit-proto/src/search/mod.rs
```diff
@@ -243,3 +246,65 @@ impl PartialHit {
...
+            format!("Unsupported split field format version: {}", format_version),
+        ));
+    }
+    let reader = zstd::Decoder::new(reader)?;
+    read_split_fields_from_zstd(reader)
+}
+
+/// Reads the Split fields from a stream of bytes
+fn read_split_fields_from_zstd<R: Read>(reader: R) -> io::Result<ListFields> {
+    let all_bytes: Vec<_> = reader.bytes().collect::<Result<_, _>>()?;
```
fulmicoton, line 306:
```suggestion
    let all_bytes: Vec<_> = reader.bytes().collect::<io::Result<_>>()?;
```

quickwit/rest-api-tests/scenarii/field_capabilities/0001-field-capabilities.yaml
fulmicoton: if we drop

```
engines:
- quickwit
```

which part breaks?

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -111,7 +115,82 @@ message ReportSplitsRequest {
...
+    I64 = 2;
+    F64 = 3;
+    BOOL = 4;
+    DATE = 5;
+    FACET = 6;
+    BYTES = 7;
+    IP_ADDR = 8;
+    JSON = 9;
+}
+message ListFieldSerialized {
```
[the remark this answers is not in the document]
PSeitz: I think I can switch to `ListFieldsEntryResponse`

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -0,0 +1,796 @@
...
+        let list_fields_iter = match fields {
+            Ok(fields) => fields,
+            Err(_err) => {
+                // Schema fallback
+                get_fields_from_schema(index_id.to_string(), doc_mapper.clone())
+            }
+        };
+        let list_fields_iter = list_fields_iter.filter(|field| {
+            if let Ok(field) = field {
+                // We don't want to leak the _dynamic hack to the user API.
```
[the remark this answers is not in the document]
PSeitz: Yes, I think so too. I wasn't completely sure if there's a case where we need it.

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -0,0 +1,796 @@
...
+/// Merge iterators of sorted (FieldMetadata, index_id) into a Vec<ListFieldsEntryResponse>.
+///
+/// The iterators need to be sorted by (field_name, fieldtype)
+fn merge_leaf_list_fields(
+    iterators: Vec<impl Iterator<Item = io::Result<ListFieldsEntryResponse>>>,
+) -> crate::Result<Vec<ListFieldsEntryResponse>> {
+    let merged = iterators.into_iter().kmerge_by(|a, b| {
+        match (a, b) {
+            (Ok(ref a_field), Ok(ref b_field)) => {
+                field_metadata_from_list_field_response(a_field)
```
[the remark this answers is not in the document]
PSeitz: Yes, can be avoided

quickwit/quickwit-proto/Cargo.toml
PSeitz: It's coming from the type conversion, which is used in indexing and search (schema fallback)
This is another variant of the quickwit-indexing->search dependency problem.

Since there is no common layer, I think there are 3 options, duplicate the conversion code, use tantivy to store list fields, add a common layer crate.

```rust
impl From<Type> for ListFieldType {
    fn from(typ: Type) -> Self {
        match typ {
            Type::Str => ListFieldType::Str,
            Type::U64 => ListFieldType::U64,
            Type::I64 => ListFieldType::I64,
            Type::F64 => ListFieldType::F64,
            Type::Bool => ListFieldType::Bool,
            Type::Date => ListFieldType::Date,
            Type::Facet => ListFieldType::Facet,
            Type::Bytes => ListFieldType::Bytes,
            Type::Json => ListFieldType::Json,
            Type::IpAddr => ListFieldType::IpAddr,
        }
    }
}
```

quickwit/rest-api-tests/scenarii/field_capabilities/0001-field-capabilities.yaml
PSeitz: it works without, seems to pick quickwit as default

fulmicoton: I mean running it against elasticsearch without it.

quickwit/quickwit-proto/src/search/mod.rs
PSeitz: switched to the bulk API

quickwit/quickwit-search/src/list_fields.rs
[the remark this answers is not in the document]
PSeitz: done

quickwit/quickwit-proto/protos/quickwit/search.proto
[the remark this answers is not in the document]
PSeitz: done

quickwit/quickwit-search/src/list_fields.rs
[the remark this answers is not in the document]
PSeitz: done

quickwit/rest-api-tests/scenarii/field_capabilities/0001-field-capabilities.yaml
PSeitz: The main differences are:
- Mixed types are not supported in elastic, so this test doesn't work in elastic. We get multiple entries for mixed types. https://github.com/quickwit-oss/quickwit/issues/3527#issuecomment-1831166121
- We have different coercion rules (e.g. i64, u64, f64 preference order)
- Mapping of text types. Currently we map text fields with fast to `keyword` and indexed to `text`. `keyword` is not on the same fieldname in elastic but a new one.
- We map our floats to `double`, but elastic seems to use `float` when the type is detected

quickwit/quickwit-proto/Cargo.toml
fulmicoton: on the consumption (search) side, do we even need to ever do conversion?
on the indexing side, we can have the conversion code.

PSeitz: Yes, there's a fallback case in search to build a field list from `Schema`, if the list fields file does not exist

fmassot: I think we should remove the fallback, It's ok to say to the user that we returned nothing with splits built before adding this feature.

PSeitz: I removed the fallback, it's a not great though to have a worse API to circumvent the dependency issue

quickwit/quickwit-indexing/src/actors/packager.rs
```diff
@@ -328,6 +331,47 @@ fn create_packaged_split(
     Ok(packaged_split)
 }
 
+/// Serializes the Split fields.
+///
+/// `fields_metadata` has to be sorted.
+pub fn serialize_field_metadata(fields_metadata: &[FieldMetadata]) -> Vec<u8> {
```
fulmicoton, line 337: does it need to be pub?

quickwit/quickwit-proto/src/search/mod.rs
```diff
@@ -243,3 +245,48 @@ impl PartialHit {
...
+    // Write Header -- Format Version 2
+    let format_version = 2u8;
+    out.push(format_version);
+    // Write Payload
+    out.extend_from_slice(&payload_compressed);
+    out
+}
+
+/// Reads a fixed number of bytes into an array and returns the array.
+fn read_exact_array<R: Read, const N: usize>(reader: &mut R) -> io::Result<[u8; N]> {
```
fulmicoton, line 267: nitpick...
`impl Trait` is nice in that kidn of funciton as it removes the need to `::<_, 4>` on call sites.

```suggestion
fn read_exact_array<const N: usize>(reader: &mut impl std::io::Read) -> io::Result<[u8; N]> {
```

quickwit/quickwit-indexing/src/actors/packager.rs
PSeitz: no, that's an artifact from moving the code

fulmicoton (approved):

Issue #4253, opened 2023-12-11
ShardPositionsUpdate` event handler timed out`


fulmicoton:
`2023-12-11T02:05:21.913Z  WARN quickwit_common::pubsub: `quickwit_proto::indexing::ShardPositionsUpdate` event handler timed out`

Happens regularly when running the 
`./target/release/quickwit-simian-army "chico(rate=1000000)*20" "chico(rate=100000)*20" "chico(rate=12000000)" "chico(rate=20000000)" --endpoint localhost:7280`

fulmicoton:
Here is what we get when it happens.
```
ub: `quickwit_proto::indexing::ShardPositionsUpdate` event handler "quickwit_ingest::ingest_v2::ingester::WeakIngesterState" timed out
2023-12-11T03:27:12.520Z  WARN quickwit_common::pubsub: `quickwit_proto::indexing::ShardPositionsUpdate` event handler "quickwit_ingest::ingest_v2::ingester::WeakIngesterState" timed out
2023-12-11T03:27:12.629Z  WARN quickwit_common::pubsub: `quickwit_proto::indexing::ShardPositionsUpdate` event handler "quickwit_ingest::ingest_v2::ingester::WeakIngesterState" timed out
2023-12-11T03:27:12.892Z  WARN quickwit_common::pubsub: `quickwit_proto::indexing::ShardPositionsUpdate` event handler "quickwit_ingest::ingest_v2::ingester::WeakIngesterState" timed out
2023-12-11T03:27:13.040Z  WARN quickwit_common::pubsub: `quickwit_proto::indexing::ShardPositionsUpdate` event handler "quickwit_ingest::ingest_v2::ingester::WeakIngesterState" timed out
2023-12-11T03:27:13.225Z  WARN quickwit_common::pubsub: `quickwit_proto::indexing::ShardPositionsUpdate` event handler "quickwit_ingest::ingest_v2::ingester::WeakIngesterState" timed out
2023-12-11T03:27:13.322Z  WARN quickwit_common::pubsub: `quickwit_proto::indexing::ShardPositionsUpdate` event handler "quickwit_ingest::ingest_v2::ingester::WeakIngesterState" timed out
2023-12-11T03:27:13.646Z  WARN quickwit_common::pubsub: `quickwit_proto::indexing::ShardPositionsUpdate` event handler "quickwit_ingest::ingest_v2::ingester::WeakIngesterState" timed out
[...]

PR #4254, opened 2023-12-11
Increases the pubsub callback timeout.

fulmicoton:
From 1s to 20s.
We log a warning whenever the callback takes more than 2s.

Related to #4253


guilload (approved):

fulmicoton:
Judged unnecessary after #4623

PR #4263, opened 2023-12-12
Guarding the router writer path with a semaphore.

fulmicoton:
The idea is to ensure we have at most one write request waiting in the implicit mrecordlog mutex queue.

In addition, the semaphore limits the number of waiters, so that we can push the burden of retry to the client.

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/router.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/router.rs
@@ -47,6 +47,7 @@ use super::ingester::PERSIST_REQUEST_TIMEOUT;
 use super::routing_table::RoutingTable;
 use super::workbench::IngestWorkbench;
 use super::IngesterPool;
+use crate::semaphore_with_waiter::SemaphoreWithMaxWaiters;
 use crate::{with_request_metrics, LeaderId};
 
 /// Duration after which ingest requests time out with [`IngestV2Error::Timeout`].
@@ -67,6 +68,7 @@ pub struct IngestRouter {
     ingester_pool: IngesterPool,
     state: Arc<RwLock<RouterState>>,
     replication_factor: usize,
+    write_semaphore: SemaphoreWithMaxWaiters,
 }
 
 struct RouterState {
@@ -101,6 +103,7 @@ impl IngestRouter {
             ingester_pool,
             state,
             replication_factor,
+            write_semaphore: SemaphoreWithMaxWaiters::new(1, 10),
         }
     }
 
@@ -419,6 +422,11 @@ impl IngestRouterService for IngestRouter {
         &mut self,
         ingest_request: IngestRequestV2,
     ) -> IngestV2Result<IngestResponseV2> {
+        let _permit = self
+            .write_semaphore
+            .acquire()
+            .await
+            .map_err(|()| IngestV2Error::TooManyRequests)?;
         with_request_metrics!(
             self.ingest_timeout(ingest_request, INGEST_REQUEST_TIMEOUT)
                 .await,
--- a/quickwit/quickwit-ingest/src/lib.rs
+++ b/quickwit/quickwit-ingest/src/lib.rs
@@ -30,6 +30,7 @@ mod metrics;
 mod notifications;
 mod position;
 mod queue;
+mod semaphore_with_waiter;
 
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
--- a/quickwit/quickwit-ingest/src/semaphore_with_waiter.rs
+++ b/quickwit/quickwit-ingest/src/semaphore_with_waiter.rs
@@ -1,0 +1,104 @@
+// Copyright (C) 2023 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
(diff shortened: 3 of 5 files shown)


fulmicoton:
I did not manage to make that result in a 429 error status code at the rest api level.

quickwit/quickwit-common/src/lib.rs
```diff
@@ -45,6 +45,8 @@ pub mod tower;
 pub mod type_map;
 pub mod uri;
 
+mod semaphore_with_waiter;
```
guilload, line 48: Personally, I'd leave this custom semaphore in `quickwit-ingest` until we need it somewhere else.

quickwit/quickwit-common/src/semaphore_with_waiter.rs
```diff
@@ -0,0 +1,102 @@
...
+use std::sync::Arc;
+
+use tokio::sync::{OwnedSemaphorePermit, Semaphore, TryAcquireError};
+
+/// `SemaphoreWithLimitedWaiters` is an extension of semaphore
+/// that limits the number of waiters.
+///
+/// If more than n-waiters then acquire returns an error.
+#[derive(Clone)]
+pub struct SemaphoreWithLimitedWaiters {
```
guilload, line 29:
```suggestion
pub struct SemaphoreWithMaxWaiters {
```

quickwit/quickwit-common/src/semaphore_with_waiter.rs
```diff
@@ -0,0 +1,102 @@
...
+    permits: Arc<Semaphore>,
```
guilload, line 30: I would get rid of the two `Arc` and let callers wrap the final semaphore with an `Arc` just like for a regular `Semaphore`.

quickwit/quickwit-common/src/semaphore_with_waiter.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Creates a new `SemaphoreWithLimitedWaiters`.
+    pub fn new(num_permits: usize, max_num_waiters: usize) -> Self {
+        Self {
+            permits: Arc::new(Semaphore::new(num_permits)),
+            waiter_count: Arc::new(Semaphore::new(max_num_waiters)),
+        }
+    }
+
+    /// Acquires a permit.
+    pub async fn acquire(&self) -> Result<OwnedSemaphorePermit, ()> {
```
guilload, line 48: Arf, now I see why you wrapped the semaphores in `Arc`. We can leave it as is or do:
```rust
pub async fn acquire_owned(self: Arc<Self>) -> Result<OwnedSemaphorePermit, ()> {
```

quickwit/quickwit-common/src/semaphore_with_waiter.rs
```diff
@@ -0,0 +1,102 @@
...
+        match Semaphore::try_acquire_owned(self.permits.clone()) {
+            Ok(permit) => {
+                return Ok(permit);
+            }
+            Err(TryAcquireError::NoPermits) => {}
+            Err(TryAcquireError::Closed) => {
+                panic!("semaphore closed (this should never happen)");
```
guilload, line 55:
```suggestion
                // The `permits` semaphore should never be closed because we don't expose the `close` API and never call `close` internally.
                panic!("semaphore should not be closed");
```

quickwit/quickwit-common/src/semaphore_with_waiter.rs
```diff
@@ -0,0 +1,102 @@
...
+        };
+        // We bind wait_permit to a variable to extend its lifetime,
+        // (so we keep holding the associated permit).
+        let _wait_permit = match Semaphore::try_acquire_owned(self.waiter_count.clone()) {
+            Ok(wait_permit) => wait_permit,
+            Err(TryAcquireError::NoPermits) => {
+                return Err(());
+            }
+            Err(TryAcquireError::Closed) => {
+                panic!("semaphore closed (this should never happen)");
```
guilload, line 66: ditto

quickwit/quickwit-ingest/src/ingest_v2/router.rs
```diff
@@ -53,7 +54,7 @@ use crate::LeaderId;
 pub(super) const INGEST_REQUEST_TIMEOUT: Duration = if cfg!(any(test, feature = "testsuite")) {
     Duration::from_millis(10)
 } else {
-    Duration::from_secs(35)
+    Duration::from_secs(8)
```
guilload, line 57: Please update `ingest.md` accordingly. This is also screwing up the timeout values for persist and replicate and the number of retries.

guilload (approved):

quickwit/quickwit-common/src/lib.rs
fulmicoton: yes i agree. moving it.

quickwit/quickwit-common/src/semaphore_with_waiter.rs
fulmicoton: yes. so we cannot use the simple acquire and keep the matrioshka of methods because they borrow &mut self.

fulmicoton: the method you suggest does not work

quickwit/quickwit-ingest/src/ingest_v2/router.rs
fulmicoton: ill stick to the previous timeout for the moment then

PR #4273, opened 2023-12-13
Stack layers instead of overriding them

### LINKED ISSUE #3808: Individual and shared Tower layers can override each other
See `HelloTowerBlockBuilder::hello_layer` and `HelloTowerBlockBuilder::shared_layer` in the codegen tutorial for an example of how this can happen.

guilload:
### Description
I attempted to stack two layers (metrics and trace) on a client with `shared_layer` and it did not end well, so I decided to fix this issue. Fixes #3808.

### How was this PR tested?
Updated unit tests.


fmassot (approved):
Nice addition.

Issue #4291, opened 2023-12-18
should url decode the search index pattern


tuziben:
**Describe the bug**
When we use the Elastic or OpenSearch Golang clients to search some indexes in this pattern: `gh-archive*`, will get a error message: `"message":"index ID pattern `gh-archive%2A` is invalid. patterns must match the following regular expression: `^[a-zA-Z\\*][a-zA-Z0-9-_\\.\\*]{0,254}$`"`

**Steps to reproduce (if applicable)**

1. quickwit built with this hash [8a40a4d143937aacdbc5206d57f3ef7ea26ed9bb](https://github.com/quickwit-oss/quickwit/commit/8a40a4d143937aacdbc5206d57f3ef7ea26ed9bb)
2. config in quickwit

```
rest:
  listen_port: 7280
  cors_allow_origins:
    - "*"
  extra_headers:
    x-elastic-product: Elasticsearch
```

use the Elastic and Opensearch client to query Quickwit
Golang code
```go
       // elastic client 7.x, 8.x
       // opensearch 2.*

 	client, err := elasticsearch.NewClient(elasticsearch.Config{
		Addresses: []string{"http://127.0.0.1:7280/api/v1/_elastic"},
	})

	if err != nil {
		fmt.Println(err)
	}

	index := "gh-archive*"
	queryBody := "{\n    \"query\": {\n        \"bool\": {\n            \"filter\": [\n                {\n                    \"match_phrase\": {\n                        \"id\": {\n                            \"query\": \"32230578641\"\n                        }\n                    }\n                }\n            ]\n        }\n    }\n}"
	res, err := client.Search(
		client.Search.WithIndex(index),
		client.Search.WithBody(strings.NewReader(queryBody)),
	)

	if err != nil {
		fmt.Println(err.Error())
		return
	}

	fmt.Println(res)

        {
          "message": "index ID pattern `gh-archive%2A` is invalid. patterns must match the following regular expression: `^[a-zA-Z\\*][a-zA-Z0-9-_\\.\\*]{0,254}$`"
        }
```

This issue seems unique to the Golang client; clients in other programming languages like Python and Java don't have this problem. I've looked through the Golang client's code and couldn't find any part where it's using URL encoding. But, based on the output we're seeing, it looks like the Golang client is URL encoding the asterisk (*).

So, to fix this, I think Quickwit should decode the URLs in all its search indices to get them back to their original format.
[...]

PR #4292, opened 2023-12-18
url decode the search index pattern

tuziben:
### Description

Trying to fix this issue https://github.com/quickwit-oss/quickwit/issues/4291

### How was this PR tested?


fmassot:
@tuziben indeed warp does not decode the path string, we have to do it ourselves. 
Can we use `percent_decode` from this [crate](https://docs.rs/percent-encoding/latest/percent_encoding/fn.percent_decode.html) instead of the urlencoding crate?

tuziben:
> @tuziben indeed warp does not decode the path string, we have to do it ourselves. Can we use `percent_decode` from this [crate](https://docs.rs/percent-encoding/latest/percent_encoding/fn.percent_decode.html) instead of the urlencoding crate?

OK, I will refine this part.

quickwit/quickwit-serve/src/search_api/rest_handler.rs
```diff
@@ -58,9 +58,10 @@ pub struct SearchApi;
 pub(crate) async fn extract_index_id_patterns(
     comma_separated_index_patterns: String,
 ) -> Result<Vec<String>, Rejection> {
-    let index_pattern = percent_decode_str(&*comma_separated_index_patterns)
+    let index_pattern = percent_decode_str(&comma_separated_index_patterns)
         .decode_utf8()
-        .unwrap();
+        .expect("the index_pattern should be url decoded")
```
fmassot, line 63: can you use `.context("index pattern does not contain valid utf8 characters")` instead of `expect`?

quickwit/quickwit-serve/src/elastic_search_api/rest_handler.rs
```diff
@@ -345,15 +346,29 @@ async fn es_compat_index_multi_search(
...
         for index in &request_header.index {
-            validate_index_id_pattern(index).map_err(|err| {
-                SearchError::InvalidArgument(format!(
-                    "request header contains an invalid index: {}",
-                    err
-                ))
-            })?;
+            match percent_decode_str(index).decode_utf8() {
+                Ok(decode_index) => {
+                    validate_index_id_pattern(decode_index.to_string().as_str()).map_err(
```
fmassot, line 354: I don't think we need to decode the index here. We are deserializing the body from the request, so everything should be fine here.
Did you see some requests failing on the multi search endpoint?

tuziben: Yes, thank you for pointing out this. I fixed it.

fmassot (approved):
Thanks @tuziben !

PR #4295, opened 2023-12-18
Add integration tutorial for Grafana

ddelemeny:
Add Grafana tutorial.


fmassot (changes requested):
This is a good start.

I think we should go a step further. We should give more hints to the user so that he can build meaningful charts.

We did it in the [blog post](https://quickwit.io/blog/observing-rust-app-with-quickwit-jaeger-grafana). A nice screenshot will be nicer too.

Example of a query that can be used `service_name:quickwit AND events.event_attributes.level:WARN`

I can push a commit on top of your branch if you think you need more guidance.

fmassot (approved):

PR #4299, opened 2023-12-18
Set search after values in millis by default.

fmassot:
The initial goal was to keep search after values in nanoseconds and use ES sort formatting.

But... the sort format is not supported in ES 7. To ease migration from ES/OS users, I set search after values in millis by default and added a sort format specific to quickwit to allow users to use search after values in nanoseconds.

### DIFF
--- a/quickwit/quickwit-search/src/root.rs
+++ b/quickwit/quickwit-search/src/root.rs
@@ -152,7 +152,7 @@ struct RequestMetadata {
     timestamp_field_opt: Option<String>,
     query_ast_resolved: QueryAst,
     indexes_meta_for_leaf_search: IndexesMetasForLeafSearch,
-    sort_fields_are_datetime: Vec<bool>,
+    sort_fields_is_datetime: HashMap<String, bool>,
 }
 
 /// Validates request against each index's doc mapper and ensures that:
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-search/src/root.rs
```diff
@@ -151,25 +151,40 @@ pub struct IndexMetasForLeafSearch {
...
 
 pub(crate) type IndexesMetasForLeafSearch = HashMap<IndexUid, IndexMetasForLeafSearch>;
-type TimestampFieldOpt = Option<String>;
+
+#[derive(Debug)]
+struct RequestMetadata {
+    timestamp_field_opt: Option<String>,
+    query_ast_resolved: QueryAst,
+    indexes_meta_for_leaf_search: IndexesMetasForLeafSearch,
+    sort_fields_are_datetime_opt: Vec<Option<bool>>,
```
PSeitz, line 160: Do we need Option here or can we just use bool?

fmassot: We don't need the option here indeed, it's only useful in the validate function.

fmassot: I just removed it @PSeitz

quickwit/quickwit-search/src/root.rs
```diff
@@ -151,25 +151,40 @@ pub struct IndexMetasForLeafSearch {
...
 
 pub(crate) type IndexesMetasForLeafSearch = HashMap<IndexUid, IndexMetasForLeafSearch>;
-type TimestampFieldOpt = Option<String>;
+
+#[derive(Debug)]
+struct RequestMetadata {
+    timestamp_field_opt: Option<String>,
+    query_ast_resolved: QueryAst,
+    indexes_meta_for_leaf_search: IndexesMetasForLeafSearch,
+    sort_fields_are_datetime: Vec<bool>,
```
PSeitz, line 160: `sort_field_ordinal_is_datetime`? Or better would be a Hashmap as long it's not performance critical

PR #4320, opened 2023-12-23
Add parameter to limit the size of a query string

kamalesh0406:
### Description

I added the parameter to limit the size of the query string in a user input. Further, I moved `validate_request` out of the for loop since many of the checks inside it did not require to be run for each index_metadata.

An alternate approach for checking the query size could be to match the enums in the serialized query_ast value and check the size of the actual user input.

### How was this PR tested?

New unit tests to check if the error is thrown.

### DIFF
--- a/quickwit/quickwit-config/src/node_config/mod.rs
+++ b/quickwit/quickwit-config/src/node_config/mod.rs
@@ -170,6 +170,7 @@ pub struct SearcherConfig {
     pub partial_request_cache_capacity: ByteSize,
     pub max_num_concurrent_split_searches: usize,
     pub max_num_concurrent_split_streams: usize,
+    pub query_string_size_limit: ByteSize,
     // Strangely, if None, this will also have the effect of not forwarding
     // to searcher.
     // TODO document and fix if necessary.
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -187,6 +188,7 @@ impl Default for SearcherConfig {
             max_num_concurrent_split_searches: 100,
             aggregation_memory_limit: ByteSize::mb(500),
             aggregation_bucket_limit: 65000,
+            query_string_size_limit: ByteSize::b(1024),
```
fmassot, line 191: This limit is too low. For Jaeger, we need to make queries with at least 100 trace ID terms or even bigger ones. A trace ID is 128 bits, we represent it in hex string, so 2 characters for each byte, so we need 32 chars, and I'm not counting the field name size, which has 8 characters.

I feel like we should either put a limit on the number of terms, like 1K terms, or set a `100kb` in terms of size.

@fulmicoton @guilload any thoughts?

kamalesh0406: Can I go ahead with increasing the default size of the terms to 100kb @fulmicoton  @guilload ?

PR #4311, opened 2023-12-20
Improve otel traces index config, fixes #4218

ddelemeny:
- [x] Added fast fields.
- [x] removed `stored:false` for `span_duration_millis`.
- [x] renamed index.
- [x] input and output format of the bytes field as hex
- [x] jaeger gRPC API should hit both indexes otel-traces-v0_6 and otel-traces-v0_7 by using the pattern otel-traces-v0_*
- [x] migrate 0.6 input/output format to hex for bytes fields to be able to query both 0.6 and 0.7 indexes. 
- [x] allow to ingest using the OTLP gRPC API on a custom index by specifying the header `otel-traces-index` or `otel-logs-index.`
- [x] allow the use of Jaeger gRCP API on a custom index by adding the header `otel-traces-index` when starting jaeger-query
- [x] allow the use of Jaeger REST API on a custom index 
- [ ] ~~ensure that we don't validate the index schema if not splits are hit~~
- [x] update the endpoint of the REST Jaeger API so that we can hit `otel-traces-v0_7` index.

Tests:
- [x] integration test added to the otlp gRCP API.
- [x] started jaeger and query both `otel-traces-v0_6` and `otel-traces-v0_7`

PR #4322, opened 2023-12-25
Allow index patterns in list terms.

fmassot:
Prepare for multi otel traces indexes, coming in #4311


PSeitz (approved):
A test in rest-api-tests would be nice

Issue #4298, opened 2023-12-18
Small improvements on the ES field caps endpoint


fmassot:
- [ ] add a test for the URL `_field_caps?=*` to catch regression on this one
- [ ] if one of the index patterns doesn’t match anything, we should return a 200: we should call `check_all_index_metadata_found` method in the root function (as it's done for a classic search endpoint
- [ ] accept `ignore_unavailable` query parameter and ignore it silently

PSeitz:
> add a test for the URL _field_caps?=* to catch regression on this one

What regression? `_field_caps?=*` is an invalid url and returns an error

fmassot:
@PSeitz yes indeed, we need to fix this.

PSeitz:
> @PSeitz yes indeed, we need to fix this.

What needs to be fixed?

fmassot:
It would be nice to be able to limit the number of splits we scan.

I see 2 options: limit the number of splits to 1000 by default or to a time window (last 30 days for example).

PSeitz:
Related Issues
https://github.com/quickwit-oss/quickwit/issues/3527
https://github.com/quickwit-oss/quickwit/pull/4242

PR #4350, opened 2024-01-05
add _elastic/_field_caps API

PSeitz:
add `_elastic/_field_caps` API (`_elastic/{index}/_field_caps` is already supported)
Clear field.indices if it matches global indices

#4298
#3527

### DIFF
--- a/quickwit/quickwit-serve/src/elastic_search_api/filter.rs
+++ b/quickwit/quickwit-serve/src/elastic_search_api/filter.rs
@@ -29,7 +29,7 @@ use super::model::{
 use crate::elastic_search_api::model::{
     ElasticBulkOptions, ScrollQueryParams, SearchBody, SearchQueryParams,
 };
-use crate::search_api::extract_index_id_patterns;
+use crate::search_api::{extract_index_id_patterns, extract_index_id_patterns_default};
 
 const BODY_LENGTH_LIMIT: ByteSize = ByteSize::mib(1);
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-serve/src/elastic_search_api/filter.rs
```diff
@@ -79,6 +79,31 @@ pub(crate) fn elastic_bulk_filter(
         .and(serde_qs::warp::query(serde_qs::Config::default()))
 }
 
+#[utoipa::path(
```
PSeitz, line 82: I just moved this from below, to group the requests

quickwit/quickwit-serve/src/elastic_search_api/filter.rs
```diff
@@ -118,6 +143,22 @@ pub(crate) fn elastic_index_field_capabilities_filter() -> impl Filter<
         .and(json_or_empty())
 }
 
+#[utoipa::path(get, tag = "Metadata", path = "/_field_caps")]
+pub(crate) fn elastic_field_capabilities_filter() -> impl Filter<
```
fulmicoton, line 147: What is this about?

fulmicoton (approved):

quickwit/quickwit-serve/src/elastic_search_api/filter.rs
PSeitz: it handles path "/_field_caps" (same as the other methods in this file)

Issue #4367, opened 2024-01-11
Remove async_trait


fulmicoton:

rdettai:
Note that native async traits don't support dynamic dispatch because they are not object safe (https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#dynamic-dispatch), so most of our traits can probably not be converted yet.

fulmicoton:
I think we can close this issue.

PR #4372, opened 2024-01-11
Update rust version to 1.75.0

melbourne2991:
### Description

Updates Rust to 1.75 (#4367).

**Note:** _missing from this PR is the removal of async_trait as it is not a straightforward removal and the appropriate change for each use (253 instances in the code base) needs to be evaluated on a case by case basis:_

https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html

Dynamic dispatch still requires `async_trait` and public traits require a `trait_variant` macro as outlined in the blog post.

### How was this PR tested?

Running `make test-all`

### DIFF
--- a/quickwit/quickwit-common/src/rendezvous_hasher.rs
+++ b/quickwit/quickwit-common/src/rendezvous_hasher.rs
@@ -63,8 +63,8 @@ mod tests {
         let mut socket_set3 = vec![socket1, socket4];
         sort_by_rendez_vous_hash(&mut socket_set3, "key");
 
-        assert_eq!(socket_set1, &[socket1, socket3, socket2, socket4]);
-        assert_eq!(socket_set2, &[socket1, socket2, socket4]);
-        assert_eq!(socket_set3, &[socket1, socket4]);
+        assert_eq!(socket_set1, &[socket4, socket3, socket1, socket2]);
+        assert_eq!(socket_set2, &[socket4, socket1, socket2]);
+        assert_eq!(socket_set3, &[socket4, socket1]);
     }
 }
--- a/quickwit/quickwit-common/src/rendezvous_hasher.rs
+++ b/quickwit/quickwit-common/src/rendezvous_hasher.rs
@@ -18,14 +18,14 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::cmp::Reverse;
-use std::collections::hash_map::DefaultHasher;
 use std::hash::{Hash, Hasher};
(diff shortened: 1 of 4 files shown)


melbourne2991:
Still to do: Update async_trait instances

quickwit/quickwit-common/src/rendezvous_hasher.rs
```diff
@@ -63,8 +63,8 @@ mod tests {
         let mut socket_set3 = vec![socket1, socket4];
         sort_by_rendez_vous_hash(&mut socket_set3, "key");
 
-        assert_eq!(socket_set1, &[socket1, socket3, socket2, socket4]);
```
fulmicoton, line 66: interesting. is there a change in some hash function?
It might be a good idea to make sure we fix the hash function here.

melbourne2991: Yeah, I noticed in the commit history this has happened a couple of times before during rust version updates.

It looks like we're using the default hasher from std

```
use std::collections::hash_map::DefaultHasher;
```

I did have a quick scan through the Rust changelog and didn't see any changes specific to hashing but I could have missed something. 

The docs do note though:

> The internal algorithm is not specified, and so it and its hashes should not be relied upon over releases.

guilload: Yes. Let's stop using the default hasher and "pin" the hash function instead.

guilload:
As far as drop `async_trait` and using `RPITIT` instead, we rely on trait objects a lot for the `Storage` trait and the gRPC service client traits, which is probably the majority of where we use `async_trait`. The actor framework and its `Handler` trait might be a better candidate: as far as I remember, we don't box anything there.

quickwit/quickwit-common/src/rendezvous_hasher.rs
guilload: Let's use a third-party crate, same as Rust: https://crates.io/crates/siphasher.

guilload:
We can update Rust in this PR and see if we can get rid of some `async_trait` in another one.

quickwit/quickwit-common/src/rendezvous_hasher.rs
melbourne2991: I've updated the PR

guilload:
Hello @melbourne2991,

You can rebase your PR and fix the formatting with `make fmt`, and your PR should ready.

PR #4379, opened 2024-01-14
Replace `shard_id: u64` with generic `shard_id: ShardId(ByteString)`

guilload:
The idea is to use an `Ulid` instead of a sequential `u64` to implement the Shard API for the PostgreSQL metastore (see design doc)

~~It also replaces `PartitionId` with `ShardId`, which paves the way for unifying the Checkpoint v1 and v2 APIs in the future.~~

Todo:
- [ ] ~~remove `PartitionId`~~

### DIFF
--- a/quickwit/quickwit-cluster/src/cluster.rs
+++ b/quickwit/quickwit-cluster/src/cluster.rs
@@ -467,27 +467,23 @@ pub(crate) fn set_indexing_tasks_in_node_state(
 
 fn indexing_task_to_chitchat_kv(indexing_task: &IndexingTask) -> (String, String) {
     let IndexingTask {
-        pipeline_uid: _,
         index_uid,
         source_id,
         shard_ids,
+        pipeline_uid: _,
     } = indexing_task;
     let key = format!("{INDEXING_TASK_PREFIX}{}", indexing_task.pipeline_uid());
-    let shards_str = shard_ids.iter().sorted().join(",");
-    let value = format!("{index_uid}:{source_id}:{shards_str}");
+    let shard_ids_str = shard_ids.iter().sorted().join(",");
+    let value = format!("{index_uid}:{source_id}:{shard_ids_str}");
     (key, value)
 }
 
-fn parse_shards_str(shards_str: &str) -> Option<Vec<ShardId>> {
-    if shards_str.is_empty() {
-        return Some(Vec::new());
-    }
-    let mut shard_ids = Vec::new();
-    for shard_str in shards_str.split(',') {
-        let shard_id = shard_str.parse::<u64>().ok()?;
-        shard_ids.push(shard_id);
-    }
-    Some(shard_ids)
+fn parse_shard_ids_str(shard_ids_str: &str) -> Vec<ShardId> {
+    shard_ids_str
+        .split(',')
+        .filter(|shard_id_str| !shard_id_str.is_empty())
+        .map(ShardId::from)
+        .collect()
 }
 
 fn chitchat_kv_to_indexing_task(key: &str, value: &str) -> Option<IndexingTask> {
@@ -495,7 +491,7 @@ fn chitchat_kv_to_indexing_task(key: &str, value: &str) -> Option<IndexingTask> 
     let pipeline_uid = PipelineUid::from_str(pipeline_uid_str).ok()?;
     let (source_uid, shards_str) = value.rsplit_once(':')?;
     let (index_uid, source_id) = source_uid.rsplit_once(':')?;
-    let shard_ids = parse_shards_str(shards_str)?;
+    let shard_ids = parse_shard_ids_str(shards_str);
     Some(IndexingTask {
         index_uid: index_uid.to_string(),
         source_id: source_id.to_string(),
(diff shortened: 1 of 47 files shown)


quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -1095,7 +1091,7 @@ mod tests {
             );
             chitchat_guard.self_node_state().set(
                 format!("{INDEXING_TASK_PREFIX}01BX5ZZKBKACTAV9WEVGEMMVS1"),
-                "my_index:uid:my_source:3a,5".to_string(),
+                "my_index-uid-my_source:3,5".to_string(),
```
fulmicoton, line 1094: I think the point was to test what happened in presence of a malformed entry

quickwit/quickwit-control-plane/src/indexing_plan.rs
```diff
@@ -69,21 +69,13 @@ impl PhysicalIndexingPlan {
...
-                            .first()
-                            .copied()
-                            .cmp(&right.shard_ids.first().copied())
-                    })
-                    .then_with(|| left.pipeline_uid().cmp(&right.pipeline_uid()))
+                    .then_with(|| left.shard_ids.first().cmp(&right.shard_ids.first()))
+                    .then_with(|| left.pipeline_uid.cmp(&right.pipeline_uid))
             });
-            for task in tasks {
-                task.shard_ids.sort();
```
fulmicoton, line 85: arf. sorry.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
```diff
@@ -97,7 +97,7 @@ impl<'a> IdToOrdMap<'a> {
     }
 
     fn source_ord(&self, source_uid: &SourceUid) -> Option<SourceOrd> {
-        self.source_uid_to_source_ord.get(source_uid).copied()
+        self.source_uid_to_source_ord.get(source_uid).cloned()
```
fulmicoton, line 100: why?

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
```diff
@@ -106,7 +106,7 @@ impl<'a> IdToOrdMap<'a> {
     }
 
     fn indexer_ord(&self, indexer_id: &str) -> Option<IndexerOrd> {
-        self.indexer_id_to_indexer_ord.get(indexer_id).copied()
+        self.indexer_id_to_indexer_ord.get(indexer_id).cloned()
```
fulmicoton, line 109: same as above

quickwit/quickwit-proto/src/types/shard_id.rs
```diff
@@ -0,0 +1,161 @@
...
+use std::fmt::Debug;
+
+use bytestring::ByteString;
+use prost::DecodeError;
+use serde::{Deserialize, Serialize};
+use ulid::Ulid;
+
+/// Shard ID.
+#[derive(Clone, Debug, Default, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub struct ShardId(ByteString);
```
fulmicoton, line 30: Would it be possible to keep it `Copy` with `ShardId(Ulid)` or `Shard([u8; 16])`?

quickwit/quickwit-proto/src/ingest/ingester.rs
```diff
@@ -17,12 +17,20 @@
...
 
-use crate::types::{queue_id, Position, QueueId};
+use crate::types::{queue_id, Position, QueueId, ShardId};
 
 include!("../codegen/quickwit/quickwit.ingest.ingester.rs");
 
 pub use ingester_service_grpc_server::IngesterServiceGrpcServer;
 
+impl FetchEof {
+    pub fn shard_id(&self) -> ShardId {
```
fulmicoton, line 27: if not copy we should probably change it to 
```suggestion
    pub fn shard_id(&self) -> &ShardId {
```

fulmicoton (approved):

quickwit/quickwit-cluster/src/cluster.rs
guilload: It is still malformed. Unfortunately, any string is a valid shard ID now, so `3a` no longer breaks anything.

quickwit/quickwit-control-plane/src/indexing_plan.rs
guilload: All good.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
guilload: Copy is better. I just performed an aggressive search and replace in the file, and this is collateral damage. I'll revert.

quickwit/quickwit-proto/src/types/shard_id.rs
guilload: I did, but I want to eventually replace `PartitionId` with `ShardId` since they're essentially the same thing. The alternative is to use an enum:
```rust
enum ShardId {
  I64(), // Kafka
  Ulid(), // Ingest API
  String(ByteString), // Kafka, Kinesis
}
```

WDYT?

guilload: The Protobuf serialization won't be too bad using a `oneof`.

quickwit/quickwit-proto/src/ingest/ingester.rs
guilload: Good point.

PR #4395, opened 2024-01-16
Fixing unit test

### LINKED ISSUE #4391: `test_sync_with_ingesters` fails silently
```bash
guilload@mandrill ~/W/r/quickwit (main) [101]> cargo test --manifest-path quickwit/Cargo.toml -p quickwit-control-plane --lib -- --nocapture test_sync_with_ingesters

    Finished test [unoptimized] target(s) in 0.19s
     Running unittests src/lib.rs (quickwit/target/debug/deps/quickwit_control_plane-9f99cf38234e3e09)

running 1 test
thread 'ingest::ingest_controller::tests::test_sync_with_ingesters' panicked at quickwit-proto/src/ingest/../codegen/quickwit/quickwit.ingest.ingester.rs:530:46:
MockIngesterService::retain_shards: Expectation(<anything>) called 0 time(s) which is fewer than expected 1
test ingest::ingest_controller::tests::test_sync_with_ingesters ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 61 filtered out; finished in 0.00s
```

fulmicoton:
The test was terminating without the retain shards call to actually happen.

The mockall code was panicking without failing the test due to some tokio::test quirk.

Closes #4391.

### DIFF
--- a/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
+++ b/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
@@ -43,6 +43,7 @@ use rand::seq::SliceRandom;
 use tokio::time::timeout;
 use tracing::{error, info, warn};
 
+use crate::ingest::wait_handle::WaitHandle;
 use crate::metrics::CONTROL_PLANE_METRICS;
 use crate::model::{ControlPlaneModel, ScalingMode, ShardEntry, ShardStats};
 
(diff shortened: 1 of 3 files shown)


trinity-1686a (approved):
should `sync_with_ingesters` and `sync_with_all_ingesters` stay async? They don't appear to need to be anymore

Issue #4285, opened 2023-12-15
Complete the minio config storage with the URI of an index


fmassot:
The storage config page gives the following storage config for MinIO: 

```yaml
storage:
  s3:
    flavor: minio
    endpoint: http://127.0.0.1:9000
```

Good. Then, a user will have to define the `default_index_root_uri`. 

I saw at least two times users set `default_index_root_uri: s3://127.0.0.1:9000/indexes`, which is wrong.  The right value is `default_index_root_uri: s3://indexes`

We could at least give an example in the docs + link to the [minio tutorial](https://blog.min.io/minio-quickwit/). We could also check the config when starting quickwit and raise an error.

PR #4402, opened 2024-01-16
Update docs for 0.7

fmassot:
Fixes:
- #4285
- #4334


trinity-1686a (approved):
we add a bunch of fastfields that *we* don't need, but can be convenient for our users if they query these index by themselves. We should document the minimum index config that works with Quickwit, so that one can replace the default config with that minimum config, and save on fastfield (and possibly indexing and doc_store) size, if they don't plan on doing anything fancy.

(this is a suggestion for future doc, not a change request for this PR)

fmassot:
Yes, you are referring to the Jaeger usage. We could add an "optimization" section to put this type of information.

Issue #4407, opened 2024-01-17
When creating 1000 indexes in a row we get timeouts


fulmicoton:
```
Response { url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("simian-indexer")), port: Some(7280), path: "/api/v1/indexes", query: None, fragment: None }, status: 500, headers: {"content-type": "application/json", "vary": "origin", "vary": "access-control-request-method", "vary": "access-control-request-headers", "transfer-encoding": "chunked", "date": "Wed, 17 Jan 2024 10:13:15 GMT"} }
````

Ingester errors.
```
 failed to get open shards from control plane: an internal error occurred: failed to deserialize control plane error: `Timeout expired`
```

guilload:
File-backed metastore?

fulmicoton:
yeah. probably not too terrible.

I also need to add a short cooldown on reschedules.

PR #4416, opened 2024-01-18
Debouncing reschedule.

fulmicoton:
After this change, the creation of a new plan will happen at most once by window of 5s. Outside of cooldown period, a (relevant) event triggers a RebuildPlan instant.

Related to #4407

### DIFF
--- a/quickwit/quickwit-control-plane/src/control_plane.rs
+++ b/quickwit/quickwit-control-plane/src/control_plane.rs
@@ -63,7 +63,7 @@ pub(crate) const CONTROL_PLAN_LOOP_INTERVAL: Duration = if cfg!(any(test, featur
 };
 
 /// Minimum period between two rebuild plan operations.
-const REBUILD_PLAN_COOLDOWN_PERIOD: Duration = Duration::from_secs(5);
+const REBUILD_PLAN_COOLDOWN_PERIOD: Duration = Duration::from_secs(2);
 
 #[derive(Debug)]
 struct ControlPlanLoop;
--- a/quickwit/quickwit-control-plane/src/control_plane.rs
+++ b/quickwit/quickwit-control-plane/src/control_plane.rs
@@ -18,6 +18,8 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::collections::BTreeSet;
+use std::fmt;
+use std::fmt::Formatter;
 use std::time::Duration;
 
 use anyhow::Context;
@@ -47,6 +49,7 @@ use quickwit_proto::types::{IndexUid, NodeId, ShardId, SourceUid};
 use serde::Serialize;
 use tracing::error;
 
+use crate::debouncer::Debouncer;
 use crate::indexing_scheduler::{IndexingScheduler, IndexingSchedulerState};
 use crate::ingest::IngestController;
 use crate::model::{ControlPlaneModel, ControlPlaneModelMetrics};
@@ -59,10 +62,15 @@ pub(crate) const CONTROL_PLAN_LOOP_INTERVAL: Duration = if cfg!(any(test, featur
     Duration::from_secs(3)
 };
 
+/// Minimum period between two rebuild plan operations.
+const REBUILD_PLAN_COOLDOWN_PERIOD: Duration = Duration::from_secs(5);
+
 #[derive(Debug)]
 struct ControlPlanLoop;
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
+struct RebuildPlan;
+
 pub struct ControlPlane {
     metastore: MetastoreServiceClient,
     model: ControlPlaneModel,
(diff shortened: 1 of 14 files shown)


quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -59,10 +62,15 @@ pub(crate) const CONTROL_PLAN_LOOP_INTERVAL: Duration = if cfg!(any(test, featur
     Duration::from_secs(3)
 };
 
+/// Minimum period between two rebuild plan operation.
```
guilload, line 65:
```suggestion
/// Minimum period between two rebuild plan operations.
```

quickwit/quickwit-control-plane/src/debouncer.rs
```diff
@@ -0,0 +1,199 @@
...
+
+use std::sync::atomic::{AtomicU8, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+
+use quickwit_actors::{Actor, ActorContext, DeferableReplyHandler, Handler};
+
+/// A debouncer is a helper to debounce events.
+///
+/// The debouncing takes a `cool_down_period` parameters and works like you may expect:
```
guilload, line 28:
```suggestion
/// The debouncing takes a `cool_down_period` parameter and works as you may expect:
```

guilload: nit: I'd go with `cooldown` (no dash `-` or `_`) everywhere.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -296,6 +298,9 @@ pub async fn serve_quickwit(
                 .stack_add_source_layer(broker_layer.clone())
                 .stack_delete_source_layer(broker_layer.clone())
                 .stack_toggle_source_layer(broker_layer)
+                .stack_layer(tower::limit::GlobalConcurrencyLimitLayer::new(
+                    MAX_CONCURRENT_METASTORE_REQUESTS,
```
guilload, line 302: This should be configurable, at the minimum, with an env var. I also don't think that's an appropriate value for a PostgreSQL metastore. Let's discuss this.

quickwit/quickwit-control-plane/src/debouncer.rs
```diff
@@ -0,0 +1,198 @@
...
+pub struct Debouncer {
+    cool_down_period: Duration,
+    cooldown_state: Arc<AtomicU8>,
+}
+
+impl Debouncer {
+    pub fn new(cool_down_period: Duration) -> Debouncer {
+        Debouncer {
+            cool_down_period,
+            cooldown_state: Arc::new(AtomicU8::new(NO_COOLDOWN)),
```
guilload, line 48: Why not use an enum and a sync mutex?

quickwit/quickwit-control-plane/src/debouncer.rs
fulmicoton: ah agreed!

quickwit/quickwit-serve/src/lib.rs
fulmicoton: added an env variable and move it to the client side.

quickwit/quickwit-control-plane/src/debouncer.rs
fulmicoton: changed the implem to enum and sync mutex.

I also took it as an opportunity to change it into a proper state machine, with atomic transitions.

quickwit/quickwit-control-plane/src/debouncer.rs
```diff
@@ -0,0 +1,239 @@
...
+        Debouncer {
+            cooldown_period,
+            cooldown_state: Arc::new(Mutex::new(DebouncerState::NoCooldown)),
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+enum DebouncerState {
+    /// More than cooldown_PERIOD has elapsed since the last time we emitted an event.
```
guilload, line 54:
```suggestion
    /// More than `cooldown_period` has elapsed since we last emitted an event.
```

quickwit/quickwit-control-plane/src/debouncer.rs
```diff
@@ -0,0 +1,239 @@
...
+    NoCooldown,
+    // Less than cooldown_PERIOD has elapsed since the last time we emitted an event,
```
guilload, line 56:
```suggestion
    // Less than `cooldown_period` has elapsed since we last emitted an event,
```

quickwit/quickwit-control-plane/src/debouncer.rs
```diff
@@ -0,0 +1,239 @@
...
+    // and no event has been received since then.
+    CooldownNotScheduled,
+    // Less than cooldown_PERIOD has elapsed since the last time we emitted an event,
```
guilload, line 59:
```suggestion
    // Less than `cooldown_period` has elapsed since we last emitted an event,
```

guilload (approved):

PR #4440, opened 2024-01-22
Make io limits for merges a node-level parameter.

### LINKED ISSUE #4439: Make Merge IO Limits per-cluster
In order to avoid disrupting indexer IO, we added limitations to prevent writing on disk faster than a given speed.
That limit is decided by index config undocumented feature called `max_merge_write_throughput`.

It is however very difficult to size it properly in the presence of many indexes.

We should merge the limiter used for all indexes.

fulmicoton:
This PR also removes the index label from the IO metrics. It also removes the limits for the delete pipeline.

Closes #4439

### DIFF
--- a/quickwit/quickwit-config/src/node_config/mod.rs
+++ b/quickwit/quickwit-config/src/node_config/mod.rs
@@ -92,7 +92,7 @@ pub struct IndexerConfig {
     #[serde(default = "IndexerConfig::default_max_concurrent_split_uploads")]
     pub max_concurrent_split_uploads: usize,
     /// Limits the IO throughput of the `SplitDownloader` and the `MergeExecutor`.
-    /// On hardware where IO is constraints, it makes sure that Merges (a batch operation)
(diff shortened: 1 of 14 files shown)


quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -91,6 +91,11 @@ pub struct IndexerConfig {
     pub split_store_max_num_splits: usize,
     #[serde(default = "IndexerConfig::default_max_concurrent_split_uploads")]
     pub max_concurrent_split_uploads: usize,
+    /// Limits the IO throughput of the `SplitDownloader` and the `MergeExecutor`.
+    /// On hardware where IO is constraints, it makes sure that Merges (a batch operation)
```
guilload, line 95:
```suggestion
    /// On hardware where IO is constrained, it makes sure that Merges (a batch operation)
```

guilload (approved):

PR #4446, opened 2024-01-22
Put mrecordlog behind its own lock

guilload:
### Description
Put mrecordlog behind its own lock.

### How was this PR tested?
Ran unit tests

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/state.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/state.rs
@@ -1,0 +1,213 @@
+// Copyright (C) 2024 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -0,0 +1,212 @@
...
+        PartiallyLockedIngesterState {
+            inner: self.inner.lock().await,
+        }
+    }
+
+    pub async fn lock_fully(&self) -> FullyLockedIngesterState<'_> {
+        FullyLockedIngesterState {
+            inner: self.inner.lock().await,
+            mrecordlog: self.mrecordlog.write().await,
+        }
```
fulmicoton, line 91: I suspect the order in which we acquire those locks matters a lot here, and I have actually no clue in which order they are executed in the example below.
(For instance, in C++ in initialization via constructor, it is not done in the same order as the what the code reads, but in the order of the struct members. For rust, I don't know)

I suggest the following order.
(The idea is that I suspect that the mrecordlog lock takes more time to acquire. Let me know if I am wrong)
```suggestion
        let mrecordlog_lock = self.mrecordlog.write().await;
        let inner_lock = self.inner.lock().await;
        FullyLockedIngesterState {
            inner: inner_lock,
            mrecordlog: mrecordlog_lock,
        }
```

PR #4451, opened 2024-01-24
Allow naming tasks to differentiate actors in tokio-console.

fulmicoton:
By default, tokio-console labels its task using its spawning point. Since all actors are spawned from the same place they end up with the same label.

This introduce an API to optionally name tasks.
For actors, the actor's struct Name is used automatically.

If tokio-console is not enabled, this is no cost.


trinity-1686a (approved):
task names appear to be PascalCase, snake_case and kebab-case. Can you make them all the same case family please?

PR #4452, opened 2024-01-24
Fixing bug that was causing the Indexer to stay busy in cooperative m…

### LINKED ISSUE #4448: High cpu usage
As observed on k8s on 10nodes and push api 2
- when indexing at 200MB/s over 1000 index we use 3.6 vCPU.
- when indexing at 150MB/s over 1000 index we use 3.6 CPU.
- when indexing 150MB/s over 100 index we use 3vCPU 

Something seems to loosely take whatever vCPU is available. A looping tokio task that yield_now or at least .await maybe?
tokio-console should be able to tell us.

fulmicoton:
…ode.

Removing the difference between Idle and Processing.
It was originally introduced with the idea of measure how long actors
where busy or idle, but tokio-console does that job very well now.

The code was relying on the enforcement of the non trivial state machine:
calling .idle() in the pause state had no effect for instance.

The real bug however showed up with the following sequence of event:
- indexer drains its queue
- on_drain is called
- indexer "goes to sleep" by putting itself in the Pause state.
- a low priority message is queued into the low priority queue.
- indexer receives a high priority message (Observe is sent every second).
- indexer would go through the loop logic consuming the entire pipeline,
  but only consuming high priority messages. The `inbox.is_empty()` exit
  condiition is never satisfied, so that this loop would keep going,
  yielding at each iteration, until the Resume Command is received.

This is done on the indexer runtime.
The task would always be busy, but yield all of the time, consuming the
whatever room is available on the thread it is consuming.

Closes https://github.com/quickwit-oss/quickwit/issues/4448

# Test

The bug was doing a using yielding loop, so unit testing is a difficult.
I tested locally and on k8s.

A small pipeline of 100KB/s now takes <1% cpu vs 1 full core before.
Outside of merges, indexing 100*200kb=20MB on a single node takes 1 cpu instead of 4.
Outside of merges, indexing 1000*200kb=20MB on a 10 nodes takes 1 cpu instead of 4.

### DIFF
--- a/quickwit/quickwit-actors/src/spawn_builder.rs
+++ b/quickwit/quickwit-actors/src/spawn_builder.rs
@@ -236,14 +236,8 @@ async fn recv_envelope<A: Actor>(inbox: &mut Inbox<A>, ctx: &ActorContext<A>) ->
     }
 }
 
-fn try_recv_envelope<A: Actor>(inbox: &mut Inbox<A>, ctx: &ActorContext<A>) -> Option<Envelope<A>> {
-    if ctx.state().is_running() {
-        inbox.try_recv()
-    } else {
-        // The actor is paused. We only process command and scheduled message.
-        inbox.try_recv_cmd_and_scheduled_msg_only()
-    }
-    .ok()
+fn try_recv_envelope<A: Actor>(inbox: &mut Inbox<A>) -> Option<Envelope<A>> {
+    inbox.try_recv().ok()
 }
 
 struct ActorExecutionEnv<A: Actor> {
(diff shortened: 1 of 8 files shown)


guilload (approved):

quickwit/quickwit-actors/src/spawn_builder.rs
```diff
@@ -294,19 +288,25 @@ impl<A: Actor> ActorExecutionEnv<A> {
...
-        self.ctx.process();
         self.process_one_message(envelope).await?;
-        loop {
-            while let Some(envelope) = try_recv_envelope(&mut self.inbox, &self.ctx) {
-                self.process_one_message(envelope).await?;
-            }
-            self.ctx.yield_now().await;
-            if self.inbox.is_empty() {
-                break;
+        // If the actor is Running (not Paused), we consume all of the message in the mailbox
```
guilload, line 292:
```suggestion
        // If the actor is Running (not Paused), we consume all the messages in the mailbox
```

`All of` is

quickwit/quickwit-actors/src/spawn_builder.rs
```diff
@@ -294,19 +288,25 @@ impl<A: Actor> ActorExecutionEnv<A> {
...
+        // and call `on_drained_message`.
+        if self.ctx.state().is_running() {
+            loop {
+                while let Some(envelope) = try_recv_envelope(&mut self.inbox) {
+                    self.process_one_message(envelope).await?;
+                }
+                // We have reached the last message.
+                // Let's still yield and see if we have more messages:
+                // an upstream actor might have been experience backpressure, and is now waiting for our mailbox
```
guilload, line 301:
```suggestion
                // an upstream actor might have experienced backpressure, and is now waiting for our mailbox
```

PR #4456, opened 2024-01-24
Index template API

guilload:
In progress.

Todo
- [x] REST API
- [x] Postgres impl
- [x] Write more unit tests
- [x] Index template serialization format regression tests
- [x] REST API tests
- [x] Test rollback on error for file-backed impl
- [x] Rename table to `index_templates`
- [x] Misc. clean ups
- [x] Add test for create index
- [ ] ~~Write e2e integration test~~


quickwit/quickwit-config/src/index_config/serialize.rs
```diff
@@ -58,63 +59,51 @@ pub fn load_index_config_from_user_config(
...
         Ok(index_uri)
     }
 
-    pub fn validate_and_build(
+    pub fn build_and_validate(
         self,
         default_index_root_uri: Option<&Uri>,
     ) -> anyhow::Result<IndexConfig> {
-        validate_identifier("Index ID", &self.index_id)?;
+        validate_identifier("index", &self.index_id)?;
```
fmassot, line 88: question: do we want to validate the identifier in `validate_index_config` (I don't have a strong opinion on that)

fmassot: (I just saw the usage in the template validation function, so yes, we want to keep the identifier validation outside this function)

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -168,6 +196,72 @@ impl ControlPlane {
...
+            )
+            .await?;
+
+        let mut create_index_futures = FuturesUnordered::new();
+
+        for index_template_match in find_index_template_matches_response.matches {
+            // TODO: It's a bit brutal to fail the entire operation if applying a single index
+            // template fails. We should return a partial failure instead for the subrequest. I
+            // want to do so in an upcoming refactor where the `GetOrCreateOpenShardsRequest` will
+            // be procssed in multiple steps in a dedicated workbench.
```
fmassot, line 231:
```suggestion
            // be processed in multiple steps in a dedicated workbench.
```

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -360,30 +454,21 @@ impl Handler<CreateIndexRequest> for ControlPlane {
...
+            Ok(response) => response,
             Err(metastore_error) => return convert_metastore_error(metastore_error),
         };
-
-        let index_metadata: IndexMetadata =
-            IndexMetadata::new_with_index_uid(index_uid.clone(), index_config);
-
-        self.model.add_index(index_metadata);
-
-        self.rebuild_plan_debounced(ctx);
```
fmassot, line 381: ok so as there are no shards, we don't need that right? (I guess it's true only for ingest V2)

guilload: Yeah, we shouldn't. The comment is confusing.

fmassot (approved):
I'm just a bit worried with migrations stuff.

I will test the templates stuff tomorrow.

PR #4472, opened 2024-01-29
limit number of warn logs

### LINKED ISSUE #4461: malformed documents log too much
Every single malformed doc logs a warning in the doc processor. We need a way to make sure we log at most N docs per minutes for this.

This ended up causing #4460.

This can be tested easily with `./simian "abu(rate=20mb)" --endpoint localhost:7280`

The solution needs to be "frugal".

PSeitz:
- limit number of warn logs from doc processor to 20 per minute
- limit same error log to 3 per minute

The limit is on the doc processor level and not shared currently

closes #4461

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/doc_processor.rs
+++ b/quickwit/quickwit-indexing/src/actors/doc_processor.rs
@@ -39,6 +39,7 @@ use thiserror::Error;
 use tokio::runtime::Handle;
 use tracing::warn;
 
+use super::log_rate_limiter::LogRateLimiter;
 #[cfg(feature = "vrl")]
 use super::vrl_processing::*;
 use crate::actors::Indexer;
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-indexing/src/actors/doc_processor.rs
```diff
@@ -413,12 +416,15 @@ impl DocProcessor {
                     processed_docs.push(processed_doc);
                 }
                 Err(error) => {
-                    warn!(
-                        index_id = self.counters.index_id,
-                        source_id = self.counters.source_id,
-                        "{}",
-                        error
-                    );
+                    let error_str = error.to_string();
```
fulmicoton, line 419: keying the hashmap on the string representation of the error is not a great idea. What if the error display name contains something dynamic? (like a docid or something)

PSeitz: Then we would still have the 20 logs limit per minute

fulmicoton:
closing for #4483

PR #4474, opened 2024-01-29
Ignore index not found error in file-backed index polling task

guilload:
### Description
While running the REST API test suite, I observed a bunch of errors emitted by the file-backed indexes polling tasks. 

### How was this PR tested?
Ran the REST API test suite, and the error logs no longer appear.

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/file_backed/lazy_file_backed_index.rs
+++ b/quickwit/quickwit-metastore/src/metastore/file_backed/lazy_file_backed_index.rs
@@ -20,14 +20,14 @@
 use std::sync::{Arc, Weak};
 use std::time::Duration;
 
-use quickwit_proto::metastore::MetastoreResult;
+use quickwit_proto::metastore::{EntityKind, MetastoreError, MetastoreResult};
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-metastore/src/metastore/file_backed/lazy_file_backed_index.rs
```diff
@@ -86,19 +86,30 @@ impl LazyFileBackedIndex {
 async fn poll_index_metadata_once(
     storage: &dyn Storage,
     index_id: &str,
-    metadata_mutex: &Mutex<FileBackedIndex>,
+    unlocked_index: &Mutex<FileBackedIndex>,
```
fulmicoton, line 89: it is not unlocked? I don't understand the name of that variable.

fulmicoton (approved):

quickwit/quickwit-metastore/src/metastore/file_backed/lazy_file_backed_index.rs
guilload: Will change to something I'm not the only one to understand :)

PR #4479, opened 2024-01-30
Removing rebuild plan and moving some other around

fulmicoton:


quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -378,8 +378,6 @@ impl Handler<CreateIndexRequest> for ControlPlane {
 
         self.model.add_index(index_metadata);
 
-        self.rebuild_plan_debounced(ctx);
-
```
guilload, line 382: Let's remove the comment below:
`// We do not need to inform the indexing scheduler as there are no shards at this point.`

guilload (approved):

guilload:
I've picked up those changes in my PR to avoid conflicts.

PR #4504, opened 2024-02-03
Only emitting an event for shards that were actually updated.

fulmicoton:


quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -185,6 +185,11 @@ impl FullyLockedIngesterState<'_> {
     /// Deletes the shard identified by `queue_id` from the ingester state. It removes the
     /// mrecordlog queue first and then removes the associated in-memory shard and rate trackers.
     pub async fn delete_shard(&mut self, queue_id: &QueueId) {
+        // This if-statement is here to avoid needless log.
+        if self.inner.shards.contains_key(queue_id) {
```
guilload, line 189: We could also do below:
```rust
let removed_shard = self.shards.remove(queue_id);
self.rate_trackers.remove(queue_id);

if removed_shard.is_some() {
  log
}
```

guilload (approved):

PR #4506, opened 2024-02-05
support compressed data

### LINKED ISSUE #3990: Support compressed HTTP request bodies in REST API

PSeitz:
closes #3990

### Description

### How was this PR tested?


quickwit/rest-api-tests/run_tests.py
```diff
@@ -100,7 +100,10 @@ def run_request_step(method, step, previous_result):
     body_from_file = step.get("body_from_file", None)
     if body_from_file is not None:
         body_from_file = osp.join(step["cwd"], body_from_file)
-        kvargs["data"] = load_data(body_from_file)
+        kvargs["data"] = open(body_from_file, 'rb').read()
+        if body_from_file.endswith("gz"):
+            kvargs.setdefault("headers")["content-encoding"] = "gzip"
```
fulmicoton, line 105: can we add a test with a gzip body?

quickwit/quickwit-serve/src/decompression.rs
```diff
@@ -0,0 +1,90 @@
...
+/// There are two ways to decompress the body:
+/// - Stream the body through an async decompressor
+/// - Fetch the body and then decompress the bytes
+///
+/// The first approach lowers the latency, while the second approach is more CPU efficient.
+/// Ingesting data is usually CPU bound and there is considerable latency until the data is
+/// searchable, so the second approach is more suitable for this use case.
+async fn decompress_body(encoding: Option<String>, body: Bytes) -> Result<Bytes, warp::Rejection> {
+    match encoding.as_deref() {
+        Some("gzip" | "x-gzip") => {
```
fulmicoton, line 38: right now all errors (invalid gzip, or unknown encoding) will be 500, sometime without an error message.

fulmicoton: can we improve this and make sure we return a 400 or a 406 with an error message for these errors. Right now all of them end up as a 500.

fulmicoton (changes requested):
http error handling

quickwit/quickwit-serve/src/decompression.rs
PSeitz: I update the handling to return 415

quickwit/rest-api-tests/run_tests.py
PSeitz: The test already has gzip data (`gharchive-bulk.json.gz`), I just deactivated the client side decompression

quickwit/quickwit-serve/src/decompression.rs
fulmicoton: not the part where the header is there but the file is invalid ?

PSeitz: updated to also cover that case

fulmicoton: thank you!

fulmicoton (approved):

PR #4508, opened 2024-02-05
Solving the shard not found "error"

fulmicoton:
Indexer have some mechanic to treat shard not found errors as eof. The ShardNotFound error however was not properly transmitted through gRPC.

### DIFF
--- a/quickwit/quickwit-proto/src/types/mod.rs
+++ b/quickwit/quickwit-proto/src/types/mod.rs
@@ -230,11 +230,6 @@ impl NodeId {
     pub fn take(self) -> String {
         self.0
     }
-
-    /// Returns a reference to the underlying [`str`].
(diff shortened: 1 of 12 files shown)


quickwit/quickwit-proto/src/types/mod.rs
```diff
@@ -230,11 +230,10 @@ impl NodeId {
     pub fn take(self) -> String {
         self.0
     }
-}
 
-impl AsRef<str> for NodeId {
-    fn as_ref(&self) -> &str {
-        self.as_str()
+    /// Returns a reference to the underlying [`str`].
```
guilload, line 234: There is a `impl PartialEq<&str> for NodeId`.

guilload (approved):

PR #4512, opened 2024-02-05
Open WAL asynchronously

guilload:
Todo:
- [ ] Fix ingester init test


fulmicoton (approved):
I wonder if we could have made it without the Arc<RwLock<Option<>>>, and put the work on the grpc / rest handler to deal with the absence of an ingester.

PR #4514, opened 2024-02-05
Readded quickwit projects in CI

fulmicoton:


quickwit/Cargo.toml
```diff
@@ -36,9 +36,38 @@ members = [
...
   "quickwit-indexing",
-  "quickwit-metastore"
+  "quickwit-ingest",
+  "quickwit-integration-tests",
+  "quickwit-jaeger",
+  "quickwit-janitor",
+  "quickwit-macros",
+  "quickwit-macros/impl",
+  "quickwit-metastore",
+  # Disabling metastore-utils from the quicwit projects to ease build/deps.
```
trinity-1686a, line 60: typo
```suggestion
  # Disabling metastore-utils from the quickwit projects to ease build/deps.
```

quickwit/Cargo.toml
```diff
@@ -36,9 +36,38 @@ members = [
...
+  "quickwit-datetime",
+  "quickwit-directories",
+  "quickwit-doc-mapper",
   "quickwit-indexing",
-  "quickwit-metastore"
+  "quickwit-ingest",
+  "quickwit-integration-tests",
+  "quickwit-jaeger",
+  "quickwit-janitor",
+  "quickwit-macros",
```
trinity-1686a, line 57: I guess we don't want `quickwit-lambda` here, but maybe put it commented to be explicit?

trinity-1686a (approved):

quickwit/Cargo.toml
rdettai: Yes, Lambda tests are not run in CI. @fmassot can we enable them now?

fmassot: I would avoid that for now.

PR #4524, opened 2024-02-06
remove mutex from log rate limit

### LINKED ISSUE #4494: Remove Mutex from log rate limiter
We could have a small platform specific solution for unix, mem::transmute Instant to a u128 on unix,
and store it in two `AtomicU64`.

trinity-1686a:
### Description

fix #4494 

### How was this PR tested?

tested manually by ingesting faulty docs


quickwit/quickwit-common/Cargo.toml
```diff
@@ -14,6 +14,7 @@ anyhow = { workspace = true }
 async-speed-limit = { workspace = true }
 async-trait = { workspace = true }
 bytesize = { workspace = true }
+coarsetime = { workspace = true }
```
fulmicoton, line 17: just learned about coarsetime, neat.

quickwit/quickwit-common/src/rate_limited_tracing.rs
```diff
@@ -17,35 +17,75 @@
...
+
+        use $crate::rate_limited_tracing::coarsetime::{Instant, Duration};
 
-        static COUNT: AtomicU32 = AtomicU32::new(0);
-        // we can't build an Instant from const context, so we pinitialize with a None
-        static LAST_RESET: Mutex<Option<Instant>> = Mutex::new(None);
+        //  This is treated as 2 u32: upper bits count "generation", lower bits count number of
+        //  calls since LAST_RESET. We assume there won't be 2**32 calls to this log in ~60s.
+        //  Generation is free to wrap arround.
+        static COUNT: AtomicU64 = AtomicU64::new(0);
```
fulmicoton, line 33: That is starting to be a lot of code...

Can we move it into a function taking the `should_log(count :&AtomicU64, last_reset: &AtomicU64) -> u64` as an argument and put that outside of the macro?

After that the editor, clippy, the compiler will do a better job...

In addition it might make unit testing the function itself a little bit easier.

quickwit/quickwit-common/src/rate_limited_tracing.rs
```diff
@@ -17,35 +17,75 @@
...
+        const MASK: u64 = 0xffffffff;
```
fulmicoton, line 34: One way to make that code more readable and testable:

```rust
#[derive(Clone, Copy)]
struct Count {
    generation: u32,
    call_counts: u32, 
}

impl From<u64> for Count {
}


impl Into<u64> for Count {
}
```

if you inline the conversion method, it will be 0 cost to use only one of the two values.

quickwit/quickwit-common/src/rate_limited_tracing.rs
```diff
@@ -17,35 +17,75 @@
...
+            LAST_RESET.store(Instant::now().as_ticks(), Ordering::Release);
         }
 
-        let do_log = if count >= $limit {
-            let mut last_reset = LAST_RESET.lock().unwrap();
-            let current_time = Instant::now();
-            let should_reset = last_reset
-                .map(|last_reset| current_time.duration_since(last_reset) >= Duration::from_secs(60))
-                .unwrap_or(true);
+        let do_log = if count & MASK >= $limit {
```
fulmicoton, line 44:
```suggestion
        let do_log = if count & MASK >= $limit {
```

count != 0 please

quickwit/quickwit-common/src/rate_limited_tracing.rs
```diff
@@ -17,44 +17,123 @@
...
+
+    let count: Count = count.into();
+
+    if count.call_count >= limit {
+        let current_time = Duration::from_ticks(now().as_ticks());
+        let last_reset = Duration::from_ticks(last_reset_atomic.load(Ordering::Acquire));
+
+        let should_reset = current_time.abs_diff(last_reset) >= Duration::from_secs(60);
+
+        if should_reset {
```
fulmicoton, line 74: early return. (see our code style)

quickwit/quickwit-common/src/rate_limited_tracing.rs
```diff
@@ -17,44 +17,121 @@
...
-            *LAST_RESET.lock().unwrap() = Some(Instant::now());
+// TODO coasetime has a recent() instead of now() which is essentially free (atomic read instead of
+// vdso call), but needs us to spawn a future/thread updating that value regularly
+
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use coarsetime::{Duration, Instant};
+
+#[derive(Clone, Copy)]
+struct Count {
```
fulmicoton, line 28:
```suggestion
struct LogSiteMetadata {
```

fulmicoton: adding comments explaining how it works would help readers too!

quickwit/quickwit-common/src/rate_limited_tracing.rs
```diff
@@ -17,44 +17,121 @@
...
+    count_atomic: &AtomicU64,
+    last_reset_atomic: &AtomicU64,
+    limit: u32,
+    now: F,
+) -> bool {
+    //  count_atomic is treated as 2 u32: upper bits count "generation", lower bits count number of
+    //  calls since LAST_RESET. We assume there won't be 2**32 calls to this log in ~60s.
+    //  Generation is free to wrap arround.
+
+    let count = count_atomic.fetch_add(1, Ordering::Acquire);
```
fulmicoton, line 60: no shadowing please.
```suggestion
    // Because the lower 32 bits are storing the log count, we can
    // increment the entire u64 to record this log call.
    let logsite_data_as_u64: u64 = count_atomic.fetch_add(1, Ordering::Acquire);
```

quickwit/quickwit-common/src/rate_limited_tracing.rs
```diff
@@ -17,44 +17,121 @@
...
-            let should_reset = last_reset
-                .map(|last_reset| current_time.duration_since(last_reset) >= Duration::from_secs(60))
-                .unwrap_or(true);
-
-            if should_reset {
-                *last_reset = Some(current_time);
-                // we store 1 because we are already about to log something
-                COUNT.store(1, Ordering::Relaxed);
-                true
+    let count: Count = count.into();
```
fulmicoton, line 66:
```suggestion
    let LogSiteMetadata { generation, call_count } = previous_logsite_metadata.into();
```

quickwit/quickwit-common/src/rate_limited_tracing.rs
```diff
@@ -17,44 +17,121 @@
...
+                // we can log, update the count
+                can_log = true;
+                current_count.call_count += 1;
+                Some(current_count.into())
             } else {
-                // we are over-limit and not far enough in time to reset: don't log
-                false
+                // we can't log, save some contention by not recording that we tried to
+                // log
+                can_log = false;
```
fulmicoton, line 107:
```suggestion
```

quickwit/quickwit-common/src/rate_limited_tracing.rs
```diff
@@ -17,44 +17,121 @@
...
+                generation: count.generation.wrapping_add(1),
+                call_count: 1,
+            };
+            Some(new_count.into())
+        } else {
+            // we can't update generation&time, but maybe we can still log?
+            update_time = false;
+            if current_count.call_count < limit {
+                // we can log, update the count
+                can_log = true;
```
fulmicoton, line 101:
```suggestion
```

quickwit/quickwit-common/src/rate_limited_tracing.rs
```diff
@@ -17,44 +17,121 @@
...
+                current_count.call_count += 1;
+                Some(current_count.into())
```
fulmicoton, line 103: i can't create a proper suggestion due to a UI glitch


nitpick. I think you derive can_log from the result.
```
    let update_res = count_atomic.fetch_update(Ordering::Release, Ordering::Acquire, |current_count| {
        let mut current_count: Count = current_count.into();
        if count.generation == current_count.generation {
            // we can update generation&time, so we can definitely log
            let new_count = Count {
                generation: count.generation.wrapping_add(1),
                call_count: 1,
            };
            Some(new_count.into())
        } else {
            // Generation seems to have already been updated. Maybe we can still log?
            update_time = false;
            if current_count.call_count < limit {
                // we can log, update the count
                current_count.call_count += 1;
                Some(current_count.into())
}
..
      let can_log: bool = update_res.is_ok();
```

trinity-1686a: hum, that looks right. In a previous revision I didn't do the "save some contention by not recording that we tried to log" so we would always get Ok(_), but now that it's done, an Err happens iff we can't log

quickwit/quickwit-common/src/rate_limited_tracing.rs
```diff
@@ -17,44 +17,121 @@
...
-                // we store 1 because we are already about to log something
-                COUNT.store(1, Ordering::Relaxed);
-                true
+    let count: Count = count.into();
+
+    if count.call_count < limit {
+        return true;
+    }
+
+    let current_time = Duration::from_ticks(now().as_ticks());
```
fulmicoton, line 72: nitpick.. did we need two atomics?
Could the generation u32 just been the last_reset timing directly?

I assume the trouble is that you do not want to call time within the fetch_update closure to minimize the change of a failing compare and swap.
We have already computed one Instant though.

I think a single compare and swap is ok. We do not really care whether it fails or not. At this level of contention, we just care about limiting logs.

fulmicoton (approved):
See comments.

fulmicoton:
@trinity-1686a can we merge this?

quickwit/quickwit-common/src/rate_limited_tracing.rs
fulmicoton: `let count: Count = count.into();`

this is shadowing.

quickwit/quickwit-common/src/rate_limited_tracing.rs
trinity-1686a: there would be some complexity because the timestamp we have are 64 bits, and the format isn't defined, we could manage things by dividing by a well-known duration (1s)

Issue #4526, opened 2024-02-06
Log quickwit version on startup


fulmicoton:
the thing on display on --version will do the job

etolbakov:
Hell @fulmicoton @PSeitz 
I was just curious if this is something you were expecting:
<img width="1263" alt="Screenshot 2024-02-06 at 20 40 39" src="https://github.com/quickwit-oss/quickwit/assets/3762201/e986cba9-152b-4f97-925c-b70d75210e67">
Provided that this ticket has been already assigned please let me know if you'd like me to push the change.

Regards, Eugene

etolbakov:
I've managed to place it at the very beginning: 
<img width="1249" alt="Screenshot 2024-02-06 at 21 17 59" src="https://github.com/quickwit-oss/quickwit/assets/3762201/1bc94007-7adb-4e11-8f49-19b67f7d87aa">

PR #4528, opened 2024-02-06
chore: log current Quickwit version

etolbakov:
### Description

Log current Quickwit version on startup

### How was this PR tested?

Launched Quickwit with the change
Screenshot 2024-02-06 at 21 51 45
 
### Related issue
https://github.com/quickwit-oss/quickwit/issues/4526

### DIFF
--- a/quickwit/quickwit-cli/src/generate_markdown.rs
+++ b/quickwit/quickwit-cli/src/generate_markdown.rs
@@ -24,11 +24,7 @@ use toml::Value;
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
-    let build_info = BuildInfo::get();
-    let version_text = format!(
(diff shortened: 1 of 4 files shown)


etolbakov:
I did the change in a most straightforward way. Just would like to know if it would make sense to create a function in `BuildInfo`, something like 
```rust
    pub fn get_version_text() -> String {
        let build_info = Self::get();
        format!(
            "{} ({} {})",
            build_info.cargo_pkg_version, build_info.cargo_pkg_version, build_info.commit_date,
        )
    }
```
so it can be used in a couple of other places

fulmicoton:
@etolbakov yes can you add a method on BuildInfo?

fulmicoton:
Thank you @etolbakov

fulmicoton (approved):

PR #4546, opened 2024-02-08
Use workspace version for members

guilload:
### How was this PR tested?
`make test-all`

### DIFF
--- a/quickwit/quickwit-cli/Cargo.toml
+++ b/quickwit/quickwit-cli/Cargo.toml
@@ -1,13 +1,15 @@
 [package]
 name = "quickwit-cli"
-version = "0.7.1"
-authors = ["Quickwit, Inc. <hello@quickwit.io>"]
-edition = "2021"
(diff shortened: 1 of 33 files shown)


quickwit/quickwit-cli/Cargo.toml
```diff
@@ -1,14 +1,13 @@
...
-documentation = "https://quickwit.io/docs/"
-default-run = "quickwit"
+
+version.workspace = true
+edition.workspace = true
+homepage.workspace = true
+documentation.workspace = true
+repository.workspace = true
+authors.workspace = true
+license.workspace = true
```
trinity-1686a, line 10: maybe we want to keep `default-run = "quickwit"`?

trinity-1686a (approved):
I'm unsure if we want to remove `descriptions =` from all crates

quickwit/quickwit-cli/Cargo.toml
guilload: Good catch. I didn't notice that one.

guilload:
I will put the descriptions back and rewrite the useless ones.

PR #4547, opened 2024-02-08
add start_timestamp, end_timestamp filter to _field_caps

### LINKED ISSUE #4430: Add ability to filter field capabilities on a timestamp window
In the Grafana datasource plugin, we use the field capabilities to list the available fields.

Each panel/view in Grafana is defined on a time window. We don't want to retrieve the field caps on all the splits. Instead, we should get the field caps on a given time window.

PSeitz:
closes #4430


quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -123,9 +123,15 @@ message ListFieldsRequest {
...
   repeated string fields = 2;
+  
+  // Time filter, expressed in seconds since epoch.
+  // That filter is to be interpreted as the semi-open interval:
+  // [start_timestamp, end_timestamp).
+  optional int64 start_timestamp = 3;
+  optional int64 end_timestamp = 4;
 
   // Control if the the request will fail if split_ids contains a split that does not exist. 
-  // optional bool fail_on_missing_index = 3;
```
fmassot, line 128: I would not change this, we should increment the number for new fields.

fmassot (changes requested):
can you look at the comment?

quickwit/quickwit-proto/protos/quickwit/search.proto
PSeitz: It's commented, since it is not yet implemented

fmassot (approved):

PR #4553, opened 2024-02-08
add query visitor with mut capabilities and use it to implement redundant range query removal

### LINKED ISSUE #3416: remove range over timestamp_field from ast when extracted to start/end_timestamp

https://github.com/quickwit-oss/quickwit/pull/3400 allow extracting timestamp ranges from the QueryAst, which yields a nice optimization when it allows more pruning. However the range is also kept in the QueryAst, which means matching documents will be filtered twice, which is a bit wasteful.

In addition to the current const Ast visitor, we should probably implement some form of a mut Ast visitor to allow rewriting easily the query when there are obvious optimizations, such as removing said RangeQuery when it has been copied to the query metadata. 

~~**Describe alternatives you've considered**~~

trinity-1686a:
### Description

fix #3416
this seemed a minor optimization, but it should also greatly improve partial request cache hit rate when using ES api (or quickwit api with bounds inside the query body)

### How was this PR tested?

tested by querying quickwit api (bounds inside body) and making sure the cache now works. Also added unit tests


quickwit/quickwit-query/src/query_ast/visitor.rs
```diff
@@ -112,3 +112,123 @@ pub trait QueryAstVisitor<'a> {
         Ok(())
     }
 }
+
+/// Simple trait to implement a Visitor over the QueryAst.
+pub trait QueryAstMutVisitor {
```
fulmicoton, line 117: The name, in combination to the other QueryAstVisitor does not make sense.
`QueryAstVisitor` is also `&mut self`.

Maybe we call that a `QueryAstTransformer`

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -394,15 +398,195 @@ async fn leaf_search_single_split(
...
+        let end_timestamp = tantivy::DateTime::from_timestamp_secs(end_timestamp_s);
+        upper_bound < end_timestamp
+    }
+}
+
+impl<'a> QueryAstMutVisitor for RemoveRedundantTimestampRange<'a> {
+    type Err = std::convert::Infallible;
+
+    fn visit_bool_mut(&mut self, mut bool_query: BoolQuery) -> Result<Option<QueryAst>, Self::Err> {
+        // we only want to visit sub-queries which are strict (positive) requirements
```
fulmicoton, line 511: yes...

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -394,15 +398,195 @@ async fn leaf_search_single_split(
...
+    search_request: &mut SearchRequest,
+    split: &SplitIdAndFooterOffsets,
+    timestamp_field: &str,
+) {
+    let Ok(query_ast) = serde_json::from_str(search_request.query_ast.as_str()) else {
+        // an error will get raised a bit after anyway
+        return;
+    };
+
+    let start_timestamp = match (search_request.start_timestamp, split.timestamp_start) {
```
fulmicoton, line 433: so one thing that is sad with this approach, is that we end up with different execution logic if 
the timestamp is passed in the query , or in the search query parameter.

Could we remove all of the "must" (what you did on the bool query is good) timestamp range queries, and identify the start end timestamp range it suggests, intersect the start timestamp given by the request parameters, and then  build the fastfield field off of this values?

quickwit/quickwit-search/src/leaf.rs
trinity-1686a: ?

quickwit/quickwit-search/src/leaf.rs
trinity-1686a: I'm not totally sure I understand what you mean. In-body ranges have a higher precision than query parameters so we can't just remove them.
Or do you mean removing all the must `RangeQuery`, stop using `{start,end}_timestamp` in the collector, and instead add a single `RangeQuery` at the root of the ast?

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -394,15 +399,283 @@ async fn leaf_search_single_split(
 ///
 /// This include things such as sorting result by a field or _score when no document is requested,
 /// or applying date range when the range covers the entire split.
```
fulmicoton, line 401: can you update the comment with what has changed?

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -394,15 +399,283 @@ async fn leaf_search_single_split(
...
+            if included < excluded {
+                included_total
+            } else {
+                excluded_total
+            }
+        }
+    }
+}
+
+/// remove timestamp range that would be present both in QueryAst and SearchRequest
```
fulmicoton, line 477: you don't tell if ends up expressed in the AST or in the search request in that case

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -394,15 +399,283 @@ async fn leaf_search_single_split(
...
+                must: vec![new_ast],
+                filter: vec![range.into()],
+                ..Default::default()
+            }
+            .into();
+        }
+    }
+
+    search_request.query_ast = serde_json::to_string(&new_ast).unwrap();
+    search_request.start_timestamp = None;
```
fulmicoton, line 577: can you check whether this is a regression or not?

If we go that root we should modify the collector too.

trinity-1686a: i've tested 4 scenarii: query with many hits, few hits, with time bounds, and without. Most seems unchanged, time bounds with many hits gets way worse, 11.2ms=>18.7ms. Maybe there is something to reclaim by removing time bounds handling from the collector, but probably not \*that much\*. It looks like it might be a better idea to support higher precision timestamps inside the collector and handle the time bounds there :-/

trinity-1686a:
While performance testing this PR and some uncommited changes related to it, i found (as already stated) that in its current form it has a substantial (bad) performance impact in some common cases (queries with many hits, filtering on time).  I also found that rewriting the queries in a different way (trying to put things in `{start,end}_timestamp` instead of in ast) had a similar performance *gain* for queries made on the ES api (or passing time ranges through the query ast in quickwit api).
Doing such a change means changing `{start,end}_timestamp` to be either ns precision or auto-detected unit, and also means the rewrite could happen at the root_search level instead of the current happening once for every leaf. (we already extract time bound at the root level, but don't do any modification, so there is some level of code duplication between what is being added here at the leaf level, and what's already there at the root level)

Half this PR should be changed substantially to account for these changes. I'm going to re-open it as two PR: one limited in scope to the `QueryAstTransformer`, and one modifying the root_search to rewrite the query as done here, instead of merely extracting time bounds

quickwit/quickwit-query/src/query_ast/tantivy_query_ast.rs
```diff
@@ -202,6 +202,14 @@ impl TantivyBoolQuery {
             has_no_positive_ast_so_far,
         );
         has_no_positive_ast_so_far &= self.should.is_empty();
+        // we do that a second time in case must happens to have a MatchAll and nothing else,
+        // but filter and/or should had something
+        remove_with_guard(
```
fulmicoton, line 207: can you had a unit test that match this corner case?

quickwit/quickwit-search/src/filters.rs
```diff
@@ -33,11 +33,13 @@ pub struct TimestampFilter {
 }
 
 impl TimestampFilter {
+    /*
```
fulmicoton, line 36: i think we can remove this block

fulmicoton (approved):

PR #4557, opened 2024-02-09
Fix race condition during ingester state initialization

guilload:
### Description
- Fixed race condition: the status was set to ready before setting the log to `Some`...
- Added some checks to return proper errors instead of panicking if some traffic hits the ingester while being initialized or or after failing to initialize.

### How was this PR tested?
- Added unit tests

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/state.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/state.rs
@@ -17,6 +17,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+use std::fmt;
 use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -186,35 +205,53 @@ impl IngesterState {
         if num_deleted_shards > 0 {
             info!("deleted {num_deleted_shards} empty shard(s)");
         }
-        inner_guard.status = IngesterStatus::Ready;
-        let _ = inner_guard.status_tx.send(IngesterStatus::Ready);
-
-        self.mrecordlog.write().await.replace(mrecordlog);
```
guilload, line 192: The race condition, which is a fancy for "being stupid" t_t

fulmicoton (approved):

PR #4559, opened 2024-02-09
add env var to disable delete tasks service

trinity-1686a:
### Description

the delete task service makes numerous request to the metastore on startup (proportional to the number of indexes), but in practice it is often not used at all. This PR adds an environment variable `QW_DISABLE_DELETE_TASK_SERVICE` to disable that service entirely

### How was this PR tested?

verified the delete task service isn't started when the env var is present

### DIFF
--- a/quickwit/quickwit-janitor/src/lib.rs
+++ b/quickwit/quickwit-janitor/src/lib.rs
@@ -51,6 +51,7 @@ pub async fn start_janitor_service(
     search_job_placer: SearchJobPlacer,
     storage_resolver: StorageResolver,
     event_broker: EventBroker,
+    run_delete_task_service: bool,
 ) -> anyhow::Result<Mailbox<JanitorService>> {
     info!("starting janitor service");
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-janitor/src/lib.rs
```diff
@@ -59,17 +60,23 @@ pub async fn start_janitor_service(
...
+            config.data_dir_path.clone(),
+            config.indexer_config.max_concurrent_split_uploads,
+            universe.get_or_spawn_one::<MergeSchedulerService>(),
+            event_broker,
+        )
+        .await?;
+        let (_, delete_task_service_handle) = universe.spawn_builder().spawn(delete_task_service);
+        Some(delete_task_service_handle)
+    } else {
+        tracing::warn!("Delete task service is disabled. Delete queries will not be processed.");
```
guilload, line 77:
```suggestion
        tracing::warn!("delete task service is disabled: delete queries will not be processed");
```

guilload (approved):

PR #4565, opened 2024-02-12
improve error on missing config

PSeitz:
### Description

### How was this PR tested?


quickwit/quickwit-cli/src/lib.rs
```diff
@@ -219,7 +219,9 @@ pub fn start_actor_runtimes(
 async fn load_node_config(config_uri: &Uri) -> anyhow::Result<NodeConfig> {
     let config_content = load_file(&StorageResolver::unconfigured(), config_uri)
         .await
-        .context("failed to load node config")?;
+        .context(r#"
```
guilload, line 222: - `load_file` can fail for other reasons, we should display this error only on not found error
- `config_uri` is not necessarily `quickwit/quickwit.yaml`, depending on what was the argument of `--config` or value of `QW_CONFIG`.

PR #4569, opened 2024-02-12
[do-not-merge] report capacity in metrics

trinity-1686a:


quickwit/quickwit-ingest/src/queue.rs
```diff
@@ -212,7 +212,7 @@ impl Queues {
         self.record_log.disk_usage()
     }
 
-    pub(crate) fn memory_usage(&self) -> usize {
```
fulmicoton, line 215: We need to fix the mrecordlog API. There is no way anyone can understand what this couple is about.

trinity-1686a: as is, it is meant for debugging. having 3 function doing similar things seems to much. If this gets merged, I plan on having a single one returning a struct

PR #4572, opened 2024-02-13
support _source_excludes,_source_includes in ES API

### LINKED ISSUE #4564: Support `_source_excludes` query params in elastisearch API
A user needs to expose only a given list of fields to its users. 

Elasticsearch has a `_source_excludes` query param to exclude some fields returned in the search response. 

[Doc ref](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-get.html#docs-get-api-query-params)

PSeitz:
closes #4564

### DIFF
--- a/quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
+++ b/quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
@@ -487,22 +487,113 @@ async fn es_compat_index_field_capabilities(
     Ok(search_response_rest)
 }
 
+fn filter_source(
+    value: &mut serde_json::Value,
+    _source_excludes: &Option<Vec<String>>,
+    _source_includes: &Option<Vec<String>>,
+) {
+    fn remove_path(value: &mut serde_json::Value, path: &str) {
+        for (prefix, suffix) in generate_path_variants_with_suffix(path) {
+            match value {
+                serde_json::Value::Object(ref mut map) => {
+                    if let Some(suffix) = suffix {
+                        if let Some(sub_value) = map.get_mut(prefix) {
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
```diff
@@ -481,9 +487,22 @@ async fn es_compat_index_field_capabilities(
...
 
-fn convert_hit(hit: quickwit_proto::search::Hit, append_shard_doc: bool) -> ElasticHit {
-    let fields: BTreeMap<String, serde_json::Value> =
+fn convert_hit(
+    hit: quickwit_proto::search::Hit,
+    append_shard_doc: bool,
+    _source_excludes: &Option<Vec<String>>,
+    _source_includes: &Option<Vec<String>>,
+) -> ElasticHit {
+    let mut fields: BTreeMap<String, serde_json::Value> =
```
fmassot, line 496: I don't think this will work with a field like `app.id` where the JSON document is `{"app": {"id": 123}}`

fmassot: We could do it at the `fetch_docs_in_split` level, we won't have to traverse all the json doc by doing it on a `NamedFieldDocument`.

PSeitz: NamedFieldDocument is not flattened, we would still need to traverse the doc

PSeitz: updated to handle paths

fmassot (approved):

PR #4573, opened 2024-02-13
Make acquire connection timeout configurable and increase default to 10s

guilload:
### Description
We've recently observed that some queries take more than 2 seconds to execute: the current default is too low.

### How was this PR tested?
`make test-all`


docs/configuration/node-config.md
```diff
@@ -139,14 +139,22 @@ metastore:
 
 | Property | Description | Default value |
 | --- | --- | --- |
-| `max_num_connections` | Determines the maximum number of concurrent connections to the database server. | `10` |
+| `min_connections` | Minimum number of connections to maintain in the pool at all times. | `0` |
+| `max_connections` | Maximum number of connections to maintain in the pool. | `10` |
+| `acquire_connection_timeout` | Maximum amount of time to spend waiting for an available connection before aborting a query. | `10s` |
```
guilload, line 144: Default values from `sqlx`. They seem to make more sense than our previous defaults.

fmassot (approved):

PR #4591, opened 2024-02-15
Add gRPC metrics layer to clients and servers

guilload:


quickwit/quickwit-serve/src/lib.rs
```diff
@@ -709,34 +735,43 @@ async fn setup_ingest_v2(
...
 
+                    // Remember that movie called "Inception"?
                     if node.is_self_node() {
-                        let ingester_service =
-                            ingester_service_opt.expect("ingester service should be initialized");
+                        let ingester = ingester_opt_clone_clone
+                            .expect("ingester service should be initialized");
+                        let layers = ServiceBuilder::new()
+                            .layer(INGEST_GRPC_CLIENT_METRICS_LAYER.clone())
+                            .layer(INGEST_GRPC_SERVER_METRICS_LAYER.clone())
```
fulmicoton, line 761: do we want both here?

guilload: I updated the PR with the following comment:
> // Here, since the service is available locally, we bypass the gRPC layer
                        // and use the mailbox directly. However, we still want client-side metrics,
                        // so we use both metrics layers.

PR #4595, opened 2024-02-16
Debounce open shards request to control plane

guilload:
### Todo
- [x] Confirm there's a need for debouncing those requests
- [x] Add more unit tests
- [x] Test in the wild

### How was this PR tested?
- Add unit tests

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/debouncing.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/debouncing.rs
@@ -1,0 +1,228 @@
+// Copyright (C) 2024 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use quickwit_proto::control_plane::{
+    GetOrCreateOpenShardsRequest, GetOrCreateOpenShardsSubrequest,
+};
+use quickwit_proto::ingest::ShardIds;
+use quickwit_proto::types::{IndexId, SourceId};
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-ingest/src/ingest_v2/debouncer.rs
```diff
@@ -0,0 +1,137 @@
...
+#[derive(Default)]
+pub(super) struct Rendezvous {
+    permits: Vec<PermitGuard>,
+    barriers: Vec<BarrierGuard>,
+}
+
+impl Rendezvous {
+    /// Releases the permits and waits for the barriers to be released.
+    pub async fn wait(mut self) {
+        self.permits.clear();
```
fulmicoton, line 73:
```suggestion
        // Releasing the permits before waiting for barrier is necessary to avoid
        // dead locks.
        self.permits.clear();
```

quickwit/quickwit-ingest/src/ingest_v2/debouncer.rs
```diff
@@ -0,0 +1,137 @@
...
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use quickwit_proto::control_plane::{
+    GetOrCreateOpenShardsRequest, GetOrCreateOpenShardsSubrequest,
+};
+use quickwit_proto::ingest::ShardIds;
+use quickwit_proto::types::{IndexId, SourceId};
+use tokio::sync::{OwnedRwLockWriteGuard, RwLock};
```
fulmicoton, line 28: I didn't know that existed.

quickwit/quickwit-ingest/src/ingest_v2/debouncer.rs
```diff
@@ -0,0 +1,137 @@
...
+
+/// Debounces [`GetOrCreateOpenShardsRequest`](quickwit_proto::control_plane::GetOrCreateOpenShardsRequest) requests by index and source IDs. It gives away a permit to the first request and a barrier to subsequent requests.
+#[derive(Default)]
+pub(super) struct Debouncer {
+    locks: HashMap<(IndexId, SourceId), Arc<RwLock<()>>>,
```
fulmicoton, line 33: nitpick:

I think Debouncer should have been an object just wrapping `Arc<RwLock<()>>`
and the map part should have been elsewhere.

nitpick2:
An alternative implementation could have been:
```
struct Debouncer {
   in_flight_released: Arc<Notify>,
   in_flight_lock: Mutex<()>,
}
```

```rust
struct Debouncer {
   in_flight_released: watch::Receiver<()>,
   lock: Mutex<watch::Sender>,
}
```

In both case, the idea would then be to clone `in_flight_released` first.
Then try to acquire the Mutex. 

If the mutex is acquire return a permit. On drop of the permit notify others.

The difference with the RwLock is that the mechanism is not a chain of `Notify::notify_one()` + instantenous drop of LockGuard. It would rely on `Notify::notify_waiters()`

Your solution is easier to readproof in my opinion.

guilload: I will address nitpick 1.

Regarding 2., my first approach was using a `Mutex` and a `Notify`, but I found the `RwLock` simpler.

PR #4601, opened 2024-02-18
Limit the number of in-flight ingest requests by capacity

guilload:
### Description
Limit the number of in-flight ingest requests by capacity, which is less unfair to small requests. I'm also removing the max writers semaphore because writers already have some rate-limiting logic (see `persist`).

### How was this PR tested?
`make test-all`


guilload:
Tested on Cicada. Here's the before/after:

Screenshot 2024-02-29 at 4 54 32 PM

In addition, we are now indexing at speeds unseen before on Cicada: ~200 MB/s.

I don't even want to bother testing #4598, which will always penalize small requests and require some tuning to determine how many writers and waiters we want. This is much simpler: you allocate a buffer of in-flight requests to the router, and that's it.

quickwit/quickwit-ingest/src/ingest_v2/mod.rs
```diff
@@ -87,6 +87,29 @@ pub fn get_idle_shard_timeout() -> Duration {
         .unwrap_or(DEFAULT_IDLE_SHARD_TIMEOUT)
 }
 
+const INGEST_ROUTER_BUFFER_SIZE_ENV_KEY: &str = "QW_INGEST_ROUTER_BUFFER_SIZE_BYTES";
+
+const DEFAULT_INGEST_ROUTER_BUFFER_SIZE: u64 = if cfg!(test) { 8 } else { 256 }; // 256 MiB
```
fulmicoton, line 92:
```suggestion
const DEFAULT_INGEST_ROUTER_BUFFER_SIZE_MB: u64 = if cfg!(test) { 8 } else { 256 }; // 256 MiB
```

Or use one of the byte unit crate thing.

quickwit/quickwit-ingest/src/ingest_v2/mod.rs
```diff
@@ -87,6 +87,29 @@ pub fn get_idle_shard_timeout() -> Duration {
...
+                Some(buffer_size)
+            } else {
+                error!(
+                    "failed to parse environment variable \
+                     `{INGEST_ROUTER_BUFFER_SIZE_ENV_KEY}={buffer_size_bytes_str}`"
+                );
+                None
+            }
+        })
+        .unwrap_or(ByteSize::mib(DEFAULT_INGEST_ROUTER_BUFFER_SIZE))
```
fulmicoton, line 110:
```suggestion
        .unwrap_or(DEFAULT_INGEST_ROUTER_BUFFER_SIZE)
```

quickwit/quickwit-ingest/src/ingest_v2/mod.rs
```diff
@@ -87,6 +87,29 @@ pub fn get_idle_shard_timeout() -> Duration {
...
+const DEFAULT_INGEST_ROUTER_BUFFER_SIZE: u64 = if cfg!(test) { 8 } else { 256 }; // 256 MiB
+
+pub(crate) fn get_ingest_router_buffer_size() -> ByteSize {
+    env::var(INGEST_ROUTER_BUFFER_SIZE_ENV_KEY)
+        .ok()
+        .and_then(|buffer_size_bytes_str| {
+            if let Ok(buffer_size_bytes) = buffer_size_bytes_str.parse::<u64>() {
+                let buffer_size = ByteSize::b(buffer_size_bytes);
+                info!("overriding ingest router buffer size to {buffer_size}");
+                Some(buffer_size)
```
fulmicoton, line 101: (if there are no unit, ByteSize conveniently assumes bytes)

```suggestion
            if let Ok(buffer_size) = buffer_size_bytes_str.parse::<ByteSize>() {
                info!("overriding ingest router buffer size to {buffer_size}");
                Some(buffer_size)
```

fulmicoton (approved):

PR #4606, opened 2024-02-19
add support delete index in es API

### LINKED ISSUE #3841: Support index deletes with an index pattern
[elasticsearch muti target doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#api-multi-index)

[elastic index delete API](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-delete-index.html#delete-index-api-path-params)

This feature should be first implemented in the elasticsearch compatible API (under `_elastic`).

Additional notes:
- The delete with index patterns containing wildcards should not be enabled by default. It is ok to give a list of index IDs.
- We should return an elasticsearch-like response.

PSeitz:
closes #3841

### How was this PR tested?
rest-api-test

### DIFF
--- a/quickwit/quickwit-index-management/src/index.rs
+++ b/quickwit/quickwit-index-management/src/index.rs
@@ -229,6 +229,7 @@ impl IndexService {
     pub async fn delete_indexes(
         &self,
         index_id_patterns: Vec<String>,
+        ignore_missing: bool,
         dry_run: bool,
     ) -> Result<Vec<SplitInfo>, IndexServiceError> {
         let list_indexes_metadatas_request = ListIndexesMetadataRequest {
@@ -251,7 +252,7 @@ impl IndexService {
             .await?
             .deserialize_indexes_metadata()?;
 
-        if indexes_metadata.len() != index_id_patterns.len() {
+        if !ignore_missing && indexes_metadata.len() != index_id_patterns.len() {
             let found_index_ids: HashSet<&str> = indexes_metadata
                 .iter()
                 .map(|index_metadata| index_metadata.index_id())
--- a/quickwit/quickwit-index-management/src/index.rs
+++ b/quickwit/quickwit-index-management/src/index.rs
@@ -24,7 +24,7 @@ use std::time::Duration;
 use futures_util::StreamExt;
 use itertools::Itertools;
 use quickwit_common::fs::{empty_dir, get_cache_directory_path};
-use quickwit_common::PrettySample;
+use quickwit_common::pretty::PrettySample;
 use quickwit_config::{validate_identifier, IndexConfig, SourceConfig};
 use quickwit_indexing::check_source_connectivity;
 use quickwit_metastore::{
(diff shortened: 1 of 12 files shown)


quickwit/rest-api-tests/scenarii/es_compatibility/0024-delete_indices.yaml
```diff
@@ -0,0 +1,30 @@
...
+  version: "0.7"
+  index_id: test_index2
+  doc_mapping:
+    mode: dynamic
+sleep_after: 3
+---
+method: DELETE
+api_root: http://localhost:7280/api/v1/
+endpoint: _elastic/test_index1,does_not_exist
+status_code: 404
```
esatterwhite, line 25: This should additionally support the query param `ignore_unavailable=true|false`  where specifying true ignore indexes that do not exist, and return a 200 if the delete operations can otherwise be carried out.

if for example, I'm trying to delete 100 indexes, and one of the doesn't exists, by default, this would fail the entire request. In most practical situations, it shouldn't matter if the index doesn't exist, We're trying to delete it anyway.

PSeitz: I updated to handle `ignore_unavailable` and added a test for it

quickwit/rest-api-tests/scenarii/es_compatibility/0024-delete_indices.yaml
```diff
@@ -0,0 +1,43 @@
+---
```
trinity-1686a, line 1: it's a bummer these tests don't pass with `--engine elasticsearch`. I think we lack index creation to make them pass seamlessly, but it should be possible to create in the index conditionally with
```yaml
engines:
  - quickwit
# quickwit api here
---
engines:
  - elasticsearch
# es api here
  ```
and removing the api_root and _elastic prefix on other requests.

File: quickwit/quickwit-index-management/src/index.rs [hunk 3]
```diff
@@ -216,6 +221,101 @@ impl IndexService {
...
+        index_id_patterns: Vec<String>,
+        ignore_missing: bool,
+        dry_run: bool,
+    ) -> Result<Vec<SplitInfo>, IndexServiceError> {
+        let list_indexes_metadatas_request = ListIndexesMetadataRequest {
+            index_id_patterns: index_id_patterns.to_owned(),
+        };
+        // disallow index_id patterns containing *
+        for index_id_pattern in &index_id_patterns {
+            if index_id_pattern.contains('*') {
```
trinity-1686a, line 240: we may also want to reject `_all` if we don't support it

PSeitz: updated to reject `_all`

quickwit/rest-api-tests/scenarii/es_compatibility/0024-delete_indices.yaml
PSeitz: I updated the tests to cover elastic search compatibility

trinity-1686a (approved):

PR #4618, opened 2024-02-21
Updated chitchat to include a liveness predicate.

fulmicoton:
This guards us from attempting to build ClusterMember / Node when the critical Key Values have not been transmitted yet.

### DIFF
--- a/quickwit/quickwit-cluster/src/cluster.rs
+++ b/quickwit/quickwit-cluster/src/cluster.rs
@@ -134,9 +134,9 @@ impl Cluster {
             gossip_interval,
             marked_for_deletion_grace_period: MARKED_FOR_DELETION_GRACE_PERIOD,
             extra_liveness_predicate: Some(Box::new(|node_state: &NodeState| {
-                node_state.contains_key(GRPC_ADVERTISE_ADDR_KEY) &&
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -134,6 +134,11 @@ impl Cluster {
             failure_detector_config,
             gossip_interval,
             marked_for_deletion_grace_period: MARKED_FOR_DELETION_GRACE_PERIOD,
+            liveness_predicate: Some(Box::new(|node_state: &NodeState| {
+                node_state.contains_key(GRPC_ADVERTISE_ADDR_KEY) &&
+                node_state.contains_key(ENABLED_SERVICES_KEY) &&
+                node_state.get(READINESS_KEY) == Some(READINESS_VALUE_READY)
```
guilload, line 140:
```suggestion
                node_state.contains_key(ENABLED_SERVICES_KEY)
```
There's already some logic in place to distinguish ready/not ready live nodes. When a node is flapping between the two states, we output some logs that are useful for debugging. We just need the first two for correctness.

guilload (approved):

PR #4627, opened 2024-02-25
fix agg split_size parameter, add docs and test

PSeitz:
fix aggregation parameter `split_size` propagation to `segment_size`
improve split_size docs
add split_size parameter test


quickwit/quickwit-search/src/collector.rs
```diff
@@ -611,16 +630,26 @@ impl Collector for QuickwitCollector {
...
-                    )?,
-                ),
-            ),
+            Some(QuickwitAggregations::TantivyAggregations(aggs)) => {
+                let mut aggs = aggs.clone();
+                visit_term_aggregations(&mut aggs, &mut |terms_agg| {
+                    // Forward split_size (also shard_size) parameter to segment_size, as
+                    // this is the same in context of Quickwit
+                    if let Some(split_size) = &terms_agg.split_size {
+                        terms_agg.segment_size = Some(*split_size);
```
fulmicoton, line 639: I don't understand. So we have an argument called `split_size` introduced in tantivy even though the parameter makes no sense there.

And it is unused... So you are adding glue code to actually have us write `split_size` into `segment_size`? 

Can't we just remove the `split_size` property in tantivy and populate the `segment_size` thing directly?

Is this a serialization issue?

PSeitz: `split_size` which aliases to `shard_size` is supposed to set the number of terms that get send from a node to the root node. That part is not handled in tantivy and therefore unused, but would still make sense in the request structure.
Since we have just one segment in Quickwit currently, we can set it to `segment_size`.

We could alias `shard_size` directly to `segment_size`, but that would kind of incorrect.

fulmicoton: Can you think of a better solution?

PSeitz: We could truncate the results to `split_size` before sending the results between nodes, independently of `segment_size`. But I think we can just forward the parameter for now, until we have more insight how `split_size` and `segment_size` should differ.

docs/reference/aggregation.md
```diff
@@ -532,15 +532,22 @@ By default, the top 10 terms with the most documents are returned. Larger values
 
 ###### **split_size**
 
-The get more accurate results, we fetch more than size from each segment/split.
+The get more accurate results, we fetch more than size from each segment/split. Aliases to `shard_size`.
```
fulmicoton, line 535:
```suggestion
The get more accurate results, we fetch more than size from each segment/split.
```

docs/reference/aggregation.md
```diff
@@ -532,15 +532,22 @@ By default, the top 10 terms with the most documents are returned. Larger values
 
 ###### **split_size**
```
fulmicoton, line 533:
```suggestion
###### **shard_size**
```

fulmicoton, line 533:
```suggestion
###### **shard_size**
```

fulmicoton: Maybe we should only call it `shard_size`?

docs/reference/aggregation.md
fulmicoton, line 535:
```suggestion
The get more accurate results, we fetch more than size from each segment/split. 
The parameter is called `shard_size` for opensearch/elasticsearch compatibility reasons, and admits `split_size` as an alias.
```

quickwit/quickwit-search/src/collector.rs
fulmicoton:
```
split_size which aliases to shard_size is supposed to set the number of terms that get send from a node to the root node. That part is not handled in tantivy and therefore unused, but would still make sense in the request structure.
Since we have just one segment in Quickwit currently, we can set it to segment_size.
```

That makes no sense at all.

PSeitz: Sending data between nodes is not handled in tantivy, it's still obviously useful to control that with a parameter.

No idea why would think that is nonsense.

docs/reference/aggregation.md
PSeitz: Yes `split_size` is not a good name. It's supposed to control the amount of data that's sent between nodes, so `node_size` probably? Plus an alias to `shard_size` since it's close enough.

quickwit/rest-api-tests/scenarii/aggregations/0001-aggregations.yaml
```diff
@@ -123,13 +123,115 @@ expected:
...
       buckets:
-      - doc_count: 4
+      - doc_count: 5
         key: nice
       - doc_count: 2
         key: cool
       doc_count_error_upper_bound: 0
       sum_other_doc_count: 0
 --- 
+# Test term aggs with split_size
```
fulmicoton, line 133: shard
```suggestion
# Test term aggs with shard_size
```

quickwit/rest-api-tests/scenarii/aggregations/0001-aggregations.yaml
```diff
@@ -123,13 +123,115 @@ expected:
...
+  - quickwit
+endpoint: _elastic/aggregations/_search
+json:
+  query: { match_all: {} }
+  aggs:
+    names: 
+      terms: 
+        field: "name"
+        size: 1
+        split_size: 1
```
fulmicoton, line 146:
```suggestion
        shard_size: 1
```

fulmicoton:
@PSeitz Don't we need to update the tantivy version for the test to pass?

fulmicoton (changes requested):
I think we need to update the tantivy version, or was this done in a different PR?

fulmicoton (approved):

PSeitz:
> I think we need to update the tantivy version, or was this done in a different PR?

Yes, that was done in a different PR

Issue #4615, opened 2024-02-20
Quickwit consumes a lot of RAM when indexing on 1000 indexes and more


fmassot:
An indexer can take more than 15GB; it was observed on a k8s cluster, which has 1000 indexes.

Possible explanation: we have several queues in a given indexing pipeline, and each queue can stack up to X messages. For example, the queue size of the doc processor is 10, and given that the source is generating a message of size 1MB, we can deduce it will take 10GB if doc processor queues are full for all indexes... 

We should first confirm this is the correct explanation.

Possible solution: define a RAM budget at the server level and use it to limit the number of messages waiting in the doc processor queue (or other actors?). The budget could be equivalent to the ingest API memory budget.

trinity-1686a:
I've instrumented quickwit to know where allocation happened. On a quickwit running just the indexer service, with 1.5k indexes all getting fed lightly, i get 8.9GiB allocated memory with following usages:

- mrecordlog: 750MiB 
- quickwit_ingest: 1.0GiB
        inside a ServiceStream containing a tonic Streaming decoder 
- hyper::proto: 780MiB
        protobuf records for MRecordBatch inside a FetchMessage
        420MiB of serializing things in StoreWriter, some SegmentWriter and FastFieldsWriter...
- IndexerState::index_batch: 5.4GiB
        stacker's MemoryArena: 2.9 GiB (inside Page) 
        stacker'd ArenaHashMap: 1.3 GiB
        SingleSegmentIndexWriter::add_document: 640MiB (about half StoreWriter, then some Segment and FastField writers)


we could maybe reduce the size of an arena Page (currently 1MiB), as this seems to be mostly page allocated while creating the structure, and rarely any new allocation after a page gets filled, but this could cost performance in a workload indexing more on fewer indexes.
Some goes for ArenaHashMap's size

or we could start less indexing pipelines at once, going through all docs in one queue, committing as soon as the queue is empty, and leaving that index alone for some time while working on others, and getting back after a short while

fulmicoton:
are you using cooperative indexing?

fulmicoton:
`enable_cooperative_indexing: true` in the indexer config.

trinity-1686a:
with cooperative indexing, things are much more tame.
If I increase the number of concurrent request (i'm currently at 800), memory is dominated by buffers allocated in warp/hyper, to handle the request's body. I assume they are quickly given to us, and we keep them for some time, passing them around

[...]

PR #4629, opened 2024-02-26
add semaphore on reading http body

trinity-1686a:
### Description

while investigating #4615, i found when one of the ingest apis gets too many requests per second, it can end up consuming a lot of memory, due to many transient buffer used to store the request body being allocated at the same time (on per ongoing request).
this pr adds a semaphore to prevent having too many such bodies going arround

### How was this PR tested?

before, hammering the api with ~1k concurrent requests (most of which receiving 419) got quickwit to allocate ~8-9G of memory. Now there is no noticeable memory usage change between 100 concurrent requests and 1k

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_api_service.rs
+++ b/quickwit/quickwit-ingest/src/ingest_api_service.rs
@@ -241,6 +241,8 @@ impl IngestApiService {
             .await?;
 
         let memory_usage = self.queues.memory_usage();
+        // TODO trinity: i got `an attempt to subtract with overflow` once here. should investigate
+        // why (i was ingesting on 2k indexes at once)
         let new_capacity = self.memory_limit - memory_usage;
         self.memory_capacity.reset_capacity(new_capacity);
 
--- a/quickwit/quickwit-ingest/src/lib.rs
+++ b/quickwit/quickwit-ingest/src/lib.rs
@@ -47,6 +47,7 @@ pub use position::Position;
 pub use queue::Queues;
 use quickwit_actors::{Mailbox, Universe};
 use quickwit_config::IngestApiConfig;
+pub use semaphore_with_waiter::SemaphoreWithMaxWaiters;
 use tokio::sync::Mutex;
 
 pub const QUEUES_DIR_NAME: &str = "queues";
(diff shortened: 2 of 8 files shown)


quickwit/quickwit-ingest/src/ingest_api_service.rs
```diff
@@ -241,6 +241,8 @@ impl IngestApiService {
             .await?;
 
         let memory_usage = self.queues.memory_usage();
+        // TODO trinity: i got `an attempt to subtract with overflow` once here. should investigate
```
guilload, line 244: This is ingest v1. Did you run your tests against ingest v1 or v2? It would make sense that you observed a memory usage spike on ingest v1 when many requests are in flight because there is no concurrency limit on ingest v1.

On the other hand, in ingest v2, the router does limit the number of requests in flight.

trinity-1686a: while testing, i did a good mix of v1 and v2, iirc this error was with v1

guilload: We don't care about ingest v1 at all. If this fix is still working on v2, do we have an understanding as to why the backpressure from the router is not applied to the REST API?

trinity-1686a: that comment is unrelated to the rest of the pr.
my understanding of what happens with the backpressure not working is the following:
- many requests arrives concurrently (let's say 1k). We allocate a buffer for each body, and start receiving the requests.
- once the body of a request is fully read, the route gets called
- the route calls the inner ingest api code, which for the vast majority of requests, will reply rate_limited
at this point, we have already allocated (and at least partially filled) all those buffers. This PR adds a semaphore sooner, just before allocating the buffer for a request body, so that we don't have too much such body around at any given instant. The requests it rejects would be rejected later, when entering ingest api code, but that way we don't pay for the memory usage of these requests

guilload: Ok, in other words, we would allocate HTTP bodies faster than we reject them in the router.

guilload:
We've reproduced the high memory consumption issue on Simian, and the in-flight metrics I introduced show that the problem is not accumulating in-flight HTTP bodies.

PR #4652, opened 2024-02-29
Add REST endpoint to parse a query into a query AST.

fmassot:
Useful for debugging and for a UI that would want to use the AST.


fulmicoton (approved):

fulmicoton:
@fmassot great idea! Note that this AST is transformed depending on the schema.

PR #4662, opened 2024-03-01
Added a handler to dynamically change the log level.

fulmicoton:
Usage example: just visit.

http://localhost:7280/api/v1/log_level?filter=debug,hyper=debug,tantivy=info,quickwit_serve=debug

### DIFF
--- a/quickwit/quickwit-cli/src/logger.rs
+++ b/quickwit/quickwit-cli/src/logger.rs
@@ -18,6 +18,7 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::env;
+use std::sync::Arc;
 
 use anyhow::Context;
 use opentelemetry::sdk::propagation::TraceContextPropagator;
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-cli/src/logger.rs
```diff
@@ -39,7 +40,7 @@ pub fn setup_logging_and_tracing(
     level: Level,
     ansi_colors: bool,
     build_info: &BuildInfo,
-) -> anyhow::Result<()> {
+) -> anyhow::Result<EnvFilterReloadFn> {
     #[cfg(feature = "tokio-console")]
     {
         if std::env::var_os(QW_ENABLE_TOKIO_CONSOLE_ENV_KEY).is_some() {
```
trinity-1686a, line 46: there is a type error when that feature is enabled `Result<()>` vs `Result<EnvFilterReloadFn>`

quickwit/quickwit-serve/src/log_level_handler.rs
```diff
@@ -0,0 +1,60 @@
...
+    filter: String,
+}
+
+#[utoipa::path(get, tag = "Log level", path = "/log_level")]
+pub fn log_level_handler(
+    env_filter_reload_fn: EnvFilterReloadFn,
+) -> impl warp::Filter<Extract = (impl warp::Reply,), Error = Rejection> + Clone {
+    warp::path("log_level")
+        .and(warp::path::end())
+        .and(with_arg(env_filter_reload_fn))
```
trinity-1686a, line 38: there is no http method selector. this also works with `POST`, `PUT`, `DELETE`, `INVALID`...
(nonblocking)

trinity-1686a (approved):
doesn't compile with `--features tokio-console` currently, otherwise LGTM

PR #4664, opened 2024-03-01
Re-enable rebuild indexing plan debouncer

guilload:
Re-enabled debouncer and fixed ingest v2 happy path test. `test_single_node_cluster` needs more work, I'll open a separate PR.

### DIFF
--- a/quickwit/quickwit-control-plane/src/control_plane.rs
+++ b/quickwit/quickwit-control-plane/src/control_plane.rs
@@ -66,9 +66,9 @@ use crate::IndexerPool;
 
 /// Interval between two controls (or checks) of the desired plan VS running plan.
 pub(crate) const CONTROL_PLAN_LOOP_INTERVAL: Duration = if cfg!(any(test, feature = "testsuite")) {
-    Duration::from_millis(500)
+    Duration::from_millis(100)
 } else {
-    Duration::from_secs(3)
+    Duration::from_secs(5)
 };
 
 /// Minimum period between two rebuild plan operations.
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -66,9 +66,9 @@ use crate::IndexerPool;
 
 /// Interval between two controls (or checks) of the desired plan VS running plan.
 pub(crate) const CONTROL_PLAN_LOOP_INTERVAL: Duration = if cfg!(any(test, feature = "testsuite")) {
-    Duration::from_millis(500)
+    Duration::from_millis(100)
```
guilload, line 69: Tests are slow to run.

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -66,9 +66,9 @@ use crate::IndexerPool;
...
 } else {
-    Duration::from_secs(3)
+    Duration::from_secs(5)
```
guilload, line 71: We now react to nodes joining and leaving the cluster, so I think it makes sense to improve that value now.

quickwit/rustfmt.toml
```diff
@@ -6,6 +6,6 @@ comment_width = 120
 format_strings = true
 group_imports = "StdExternalCrate"
 imports_granularity = "Module"
-normalize_comments = false
+normalize_comments = true
```
fulmicoton, line 9: I think this is a bad idea. `/* */` are very useful when commenting out code blocks.

fulmicoton (approved):
I think we should cook some solution to return a one shot receiver, to allow clients to wait for the next build plan to be executed.

quickwit/rustfmt.toml
guilload, line 9:
```suggestion
normalize_comments = false
```

guilload: I just reverted your previous commit, but I can put it back to `false`.

PR #4668, opened 2024-03-03
delay fast field type conversion to harvest phase

PSeitz:
`extract_typed_sort_value_opt` is in the collect hot loop and very performance sensitive. Delay conversion from u64 to actual type to harvest phase.

The impact on the histogram query time seems a little odd, probably related to some inlining. The impact on other sort query times roughly what I'd expect.

comparison

### DIFF
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -170,7 +170,7 @@ impl SortingFieldExtractorComponent {
     #[inline]
     /// Converts u64 fast field values to its correct type.
     /// The conversion is delayed for performance reasons.
-    fn convert_to_sort_value(&self, sort_value: SortValue) -> SortValue {
+    fn convert_u64_ff_val_to_sort_value(&self, sort_value: SortValue) -> SortValue {
         let map_fast_field_to_value = |fast_field_value, field_type| match field_type {
             SortFieldType::U64 => SortValue::U64(fast_field_value),
             SortFieldType::I64 => SortValue::I64(i64::from_u64(fast_field_value)),
@@ -181,12 +181,24 @@ impl SortingFieldExtractorComponent {
         match self {
             SortingFieldExtractorComponent::DocId => sort_value,
             SortingFieldExtractorComponent::FastField {
-                sort_column: _,
-                sort_field_type,
-                ..
+                sort_field_type, ..
             } => match sort_value {
                 SortValue::U64(val) => map_fast_field_to_value(val, *sort_field_type),
                 _ => panic!("Internal error: Got non-U64 sort value for fast field."),
+            },
+            SortingFieldExtractorComponent::Score => sort_value,
+        }
+    }
+    #[inline]
+    /// Converts fast field values into their u64 fast field representation.
+    fn convert_to_u64_ff_val(&self, sort_value: SortValue) -> SortValue {
+        match self {
+            SortingFieldExtractorComponent::DocId => sort_value,
+            SortingFieldExtractorComponent::FastField { .. } => match sort_value {
+                SortValue::U64(_val) => sort_value,
+                SortValue::I64(val) => SortValue::U64(val.to_u64()),
+                SortValue::F64(val) => SortValue::U64(val.to_u64()),
+                SortValue::Boolean(val) => SortValue::U64(val as u64),
             },
             SortingFieldExtractorComponent::Score => sort_value,
         }
@@ -394,7 +406,7 @@ impl SegmentPartialHit {
         PartialHit {
             sort_value: self
                 .sort_value
-                .map(|sort_value| first.convert_to_sort_value(sort_value))
+                .map(|sort_value| first.convert_u64_ff_val_to_sort_value(sort_value))
                 .map(|sort_value| SortByValue {
                     sort_value: Some(sort_value),
                 }),
@@ -404,7 +416,7 @@ impl SegmentPartialHit {
                     second
                         .as_ref()
                         .expect("Internal error: Got sort_value2, but no sort extractor")
-                        .convert_to_sort_value(sort_value)
+                        .convert_u64_ff_val_to_sort_value(sort_value)
                 })
                 .map(|sort_value| SortByValue {
                     sort_value: Some(sort_value),
@@ -675,6 +687,33 @@ impl Collector for QuickwitCollector {
             // This value isn't actually used.
             _ => Ordering::Equal,
         };
+        // Convert search_after into fast field u64
+        let mut search_after = self.search_after.clone();
+        if let Some(search_after) = &mut search_after {
+            search_after
+                .sort_value
+                .as_mut()
+                .and_then(|sort_value| sort_value.sort_value.as_mut())
+                .map(|sort_value| {
+                    *sort_value = score_extractor
+                        .first
+                        .convert_to_u64_ff_val(*sort_value)
+                        .into();
+                });
+
+            search_after
+                .sort_value2
+                .as_mut()
+                .and_then(|sort_value2| sort_value2.sort_value.as_mut())
+                .map(|sort_value| {
+                    *sort_value = score_extractor
+                        .second
+                        .as_ref()
+                        .expect("Internal error: Got sort_value2, but no sort extractor")
+                        .convert_to_u64_ff_val(*sort_value)
+                        .into();
+                });
+        }
         Ok(QuickwitSegmentCollector {
             num_hits: 0u64,
             split_id: self.split_id.clone(),
@@ -683,7 +722,7 @@ impl Collector for QuickwitCollector {
             segment_ord,
             timestamp_filter_opt,
             aggregation,
-            search_after: self.search_after.clone(),
+            search_after,
             precomp_search_after_order,
         })
     }
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -154,7 +154,23 @@ impl SortingFieldExtractorComponent {
     ///
     /// The function returns None if the sort key is a fast field, for which we have no value
     /// for the given doc_id, or we sort by DocId.
+    #[inline]
     fn extract_typed_sort_value_opt(&self, doc_id: DocId, score: Score) -> Option<SortValue> {
+        match self {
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-search/src/collector.rs
```diff
@@ -154,25 +154,53 @@ impl SortingFieldExtractorComponent {
...
+                sort_field_type, ..
+            } => match sort_value {
+                SortValue::U64(val) => map_fast_field_to_value(val, *sort_field_type),
+                _ => panic!("Internal error: Got non-U64 sort value for fast field."),
+            },
+            SortingFieldExtractorComponent::Score => sort_value,
+        }
+    }
+    #[inline]
+    /// Converts fast field values into their u64 fast field representation.
```
fulmicoton, line 193: nitpick: I think we tend to put inline after the comment.
```suggestion
    /// Converts fast field values into their u64 fast field representation.
    #[inline]
```

quickwit/quickwit-search/src/collector.rs
```diff
@@ -638,6 +689,33 @@ impl Collector for QuickwitCollector {
             // this value isn't actually used.
             Ordering::Equal
         };
+        // Convert search_after into fast field u64
+        let mut search_after = self.search_after.clone();
```
fulmicoton, line 693: This piece of code has a few problems:

It use a functional style to mutate something.
It changes the semantics of something without adding a comment, or using a different type.

After this patch, search_after in Collector is a partial hit that contains sort values expressed in their original type.
and in the SegmentCollector, it uses their u64 representation.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -154,25 +154,53 @@ impl SortingFieldExtractorComponent {
     ///
     /// The function returns None if the sort key is a fast field, for which we have no value
     /// for the given doc_id, or we sort by DocId.
+    #[inline]
```
fulmicoton, line 157: The comment of the function and the name of the function need to be changed.
This function is now only return the u64 value.

Can we introduce another type for this?

quickwit/quickwit-search/src/collector.rs
```diff
@@ -154,25 +154,53 @@ impl SortingFieldExtractorComponent {
...
+            SortingFieldExtractorComponent::Score => sort_value,
+        }
+    }
+    #[inline]
+    /// Converts fast field values into their u64 fast field representation.
+    fn convert_to_u64_ff_val(&self, sort_value: SortValue) -> SortValue {
+        match self {
+            SortingFieldExtractorComponent::DocId => sort_value,
+            SortingFieldExtractorComponent::FastField { .. } => match sort_value {
+                SortValue::U64(_val) => sort_value,
```
fulmicoton, line 198: I think there is a bug here, aren't there?

The partial hit could be coming from a different split, where the value was i64 for instance.
The current split on the other hand could be `u64`.

In that case I believe the ordering is broken.

quickwit/quickwit-search/src/collector.rs
fulmicoton: `SortingFieldExtractorComponent::Score` is quite a pain of course. I'm ok to treat it as u64 using tantivy's `u64_to_f64` if that helps making the other code path faster / simpler.

fulmicoton (commented):
A sad bit is that we compare things twice:
- the search_after part is acting as a first fence.
- the head of the top k collector is also acting as a fence.

In reality... there is only one fence of course.
We could have had single set of sort values on the segment collector
(`current_score_fence`) or something like that, initialized it on the search_after parameter, and then just used the regular collect_top_k logic.

What makes everything so difficult is the different treatment of ties ( the `split_search_after_order` thing).

quickwit/quickwit-search/src/collector.rs
PSeitz: Indeed, this is broken currently. 
This needs to reflect `impl Ord for SortValue`.


This comparison on `SortValue` is also something that I couldn't attribute in a profile:
6% time spent in `quickwit_proto::search::<...>::cmp`  in the
quickwit generated log benchmark on query `sort_numeric_desc.json`:
```json
{
  "track_total_hits": true,
  "query": {
    "match_all": {}
  },
  "sort": [
    {
      "metrics.size": {"order": "desc"}
    }
  ]
}
```

quickwit/quickwit-search/src/collector.rs
PSeitz: I switched to u64, the main advantage is having a simpler compare function. It's additionally ~10% faster on average

quickwit/quickwit-search/src/collector.rs
PSeitz: I replaced the functional style, and convert into a new `SearchAfterSegment` struct instead to make the conversion clear.

PSeitz:
> A sad bit is that we compare things twice:
> 
>     * the search_after part is acting as a first fence.
> 
>     * the head of the top k collector is also acting as a fence.
> 
> 
> In reality... there is only one fence of course. We could have had single set of sort values on the segment collector (`current_score_fence`) or something like that, initialized it on the search_after parameter, and then just used the regular collect_top_k logic.
> 
> What makes everything so difficult is the different treatment of ties ( the `split_search_after_order` thing).

True, I think we could maybe have specialized SegmentCollectors with `Box<SegmentCollector>`, to provide specialized implementations via generics for the different use cases. The calling performance overhead in combination with `collect_block` should be fine.

fmassot:
should we merge this PR and open a follow-up issue then?

quickwit/quickwit-search/src/collector.rs
fulmicoton: awesome.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -303,20 +390,76 @@ pub struct QuickwitSegmentCollector {
...
+impl SearchAfterSegment {
+    fn new(
+        search_after: Option<PartialHit>,
+        score_extractor: &SortingFieldExtractorPair,
+    ) -> Option<Self> {
+        let Some(search_after) = search_after else {
+            return None;
+        };
+
+        let mut sort_value = None;
```
fulmicoton, line 414: Can `sort_value` ever be `None`?

quickwit/quickwit-search/src/collector.rs
```diff
@@ -303,20 +390,76 @@ pub struct QuickwitSegmentCollector {
...
     aggregation: Option<AggregationSegmentCollectors>,
-    search_after: Option<PartialHit>,
+    search_after: Option<SearchAfterSegment>,
     // Precomputed order for search_after for split_id and segment_ord
     precomp_search_after_order: Ordering,
 }
 
+/// Search After, but the sort values are converted to the u64 fast field representation.
+struct SearchAfterSegment {
+    sort_value: Option<u64>,
```
fulmicoton, line 400:
```suggestion
    sort_value: u64,
```

quickwit/quickwit-search/src/collector.rs
```diff
@@ -303,20 +390,76 @@ pub struct QuickwitSegmentCollector {
...
+            compare_on_equal: !search_after.split_id.is_empty(),
+            doc_id: search_after.doc_id,
+        })
+    }
+}
+
 impl QuickwitSegmentCollector {
     #[inline]
     fn collect_top_k(&mut self, doc_id: DocId, score: Score) {
         let (sort_value, sort_value2) =
```
fulmicoton, line 457: This really needs a type

quickwit/quickwit-search/src/collector.rs
PSeitz: It will be `None` if the value on the passed value on the protobuf struct is `None`. 
Should we ignore that case? The comparison is on `Option<u64>`  anyway, since the fastfield may not have a value (except we add a specialized version for full columns)

fulmicoton: Ah yes if there are no search_after_sort_value.

fulmicoton, line 414:
```suggestion
        let mut sort_value: Option<u64> = None;
```

fulmicoton:
> should we merge this PR and open a follow-up issue then?

 Let's fix the bug and then merge. People do rely on edge.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -150,29 +150,119 @@ enum SortingFieldExtractorComponent {
...
-            } => sort_column
-                .first(doc_id)
-                .map(|field_val| map_fast_field_to_value(field_val, *sort_field_type)),
-            SortingFieldExtractorComponent::Score { .. } => Some(SortValue::F64(score as f64)),
+                sort_field_type, ..
+            } => map_fast_field_to_value(sort_value, *sort_field_type),
+            SortingFieldExtractorComponent::Score => SortValue::F64(f64::from_u64(sort_value)),
+        }
+    }
+    /// Converts fast field values into their u64 fast field representation.
```
fulmicoton, line 188:
```suggestion
    /// Converts fast field values into their u64 fast field representation.
    /// This is used to convert `search_after` sort value to a u64 representation that will respect the 
    /// same order as the `SortValue` representation.
```

quickwit/quickwit-search/src/collector.rs
```diff
@@ -150,29 +150,119 @@ enum SortingFieldExtractorComponent {
...
+    /// Returns the sort value for the given element in its u64 representation. The returned u64
+    /// representation maintains the ordering of the original value.
     ///
     /// The function returns None if the sort key is a fast field, for which we have no value
     /// for the given doc_id, or we sort by DocId.
-    fn extract_typed_sort_value_opt(&self, doc_id: DocId, score: Score) -> Option<SortValue> {
+    #[inline]
+    fn extract_typed_sort_value_opt(&self, doc_id: DocId, score: Score) -> Option<u64> {
+        match self {
+            SortingFieldExtractorComponent::DocId => None,
```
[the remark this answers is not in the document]
PSeitz: I added a comment, but I don't really understand why we have `SortingFieldExtractorComponent::DocId` at all. (Need to investigate further, but this part is basically unchanged)

fulmicoton (approved):

Issue #4658, opened 2024-02-29
add some indexes for postgresql metastore


trinity-1686a:
currently we create no index on any postgres columns, but some indexes could be useful.
I think `split` is the table with the most potential for indexes: it's often the biggest. 

it's `index_uid` column is used in every query, there is probably some substantial gain from having a simple hash index on that:
```sql
CREATE INDEX index_splits_index_uid ON splits USING HASH (index_uid);
```

other column should probably be looked at too, but would likely have a less sever impact

trinity-1686a:
we addad an index on splits(index_uid), but should still investigate other possible indexes

trinity-1686a:
I've experimented with creating different indexes, with different level of usefulness:
- an index on `split_state` doesn't appear to get used (probably too low selectivity)
- BRIN on `time_range_start` and `time_range_end` wasn't as successful as I expected. I assume this is because in the young days of a split, it can get deleted for a merge, leaving holes behind, which get filed by some new split with different timestamps... however a BTree seems to work well
- on tags, I've experimented with multiple kind of indexes. Most aren't useful as we want to match one of the values, not the whole set of tags. A GIN can do that, with caveats below
- columns in `*_timestamp` could use a BTree to make some bookkeeping tasks faster, but that doesn't seem to important, gaining a few 100ms there is not noticeable in practice as these are background tasks

GIN on tags: we currently match tags with statements looking like `NOT ('field-name!' = ANY(tags)) OR 'field-name:the-value-i-want' = ANY(tags)`. For some reason, Postgres won't use the index for a `= ANY(_)` comparison. It does when doing `'{field-name:the-value-i-want}' && tags` (&& being the 'overlap' operator, returning true if there is anything in common between the list of 1 element we provide, and the list of tags in a split).
[...]

PR #4670, opened 2024-03-04
add postgresql index and use IN instead of many OR

### LINKED ISSUE #4666: when listing splits from multiple index in postgres metastore, use IN instead of many OR
when querying many indexes at once, the list split query on metastore create an sql request that looks like the following:
```sql
SELECT * FROM "splits" WHERE 
("index_uid" = $1 OR "index_uid" = $2 OR "index_uid" = $3 OR "index_uid" = $4 OR "index_uid" = $5 OR ...) 
AND ...
```
a request based on `IN` is [reportedly faster](https://ottertune.com/blog/query-best-practices-when-should-you-use-the-in-instead-of-the-or-operator)
```sql
SELECT * FROM "splits" WHERE 
"index_uid" IN ($1, $2, $3, $4, $5,...)
AND ...
```

trinity-1686a:
### Description

fix part of #4658 (but we'll keep that open to explore other indexes we may want to add)
add an index on `splits(index_uid)`

fix #4666
use `index_uid IN (<list of index>)` instead of `index_uid = "index1" OR index_uid = "index2" OR ...` when querying multiple index.

### How was this PR tested?

tested by measuring time and checking query plan on an index with a few hundred thousand splits.
`SELECT * FROM "splits" WHERE "index_uid" = $1` is made ~1900x faster on that particular dataset using the index (from 5.8s to 3.1ms)
selecting with a list of 1k indexes, covering almost every split, is made ~3.5x faster (from 3s to 850ms) by using `IN` over many `OR`. Execution time and query plan are not affected by `IN` over a single `=` when searching with a single index

### DIFF
--- a/quickwit/quickwit-metastore/migrations/postgresql/16_create-index-split-uid.down.sql
+++ b/quickwit/quickwit-metastore/migrations/postgresql/16_create-index-split-uid.down.sql
@@ -1,0 +1,1 @@
+DROP INDEX IF EXISTS index_splits_index_uid;
--- a/quickwit/quickwit-metastore/migrations/postgresql/16_create-index-split-uid.down.sql
+++ b/quickwit/quickwit-metastore/migrations/postgresql/16_create-index-split-uid.down.sql
@@ -1,1 +1,1 @@
-DROP INDEX IF EXISTS index_splits_index_uid;
+DROP INDEX IF EXISTS splits_index_uid_idx;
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-metastore/migrations/postgresql/16_create-index-split-uid.down.sql
```diff
@@ -0,0 +1 @@
+DROP INDEX IF EXISTS index_splits_index_uid;
```
guilload, line 1:
```suggestion
DROP INDEX IF EXISTS splits_index_uid_idx;
```

guilload: The naming convention for pg constraints/indexes is`{tablename}_{columnname(s)}_{suffix}` where suffix is `pkey`, `fkey`. `idx`, etc.

guilload (approved):

PR #4682, opened 2024-03-06
Fix the ingest rate displayed in the CLI.

fulmicoton:
The CLI was showing the average rate since the CLI was started.

This PR changes it to an estimation over the last 2 seconds.

### DIFF
--- a/quickwit/quickwit-common/src/tower/rate_estimator.rs
+++ b/quickwit/quickwit-common/src/tower/rate_estimator.rs
@@ -354,7 +354,7 @@ mod tests {
         // |0, 0, 0, 0, 0, 100*, 100, 100, 100, 100|
         //
         // Since the current bucket (idx = 5) is not taken into account, this leads
-        // to an average of 500 / 9 = 55 work units.
+        // to an average of 400 / 9 = 44 work units.
         assert_eq!(estimator.work_at(anchor + Duration::from_secs(15)), 44);
     }
 }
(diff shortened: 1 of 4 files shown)


trinity-1686a:
I tried it and it seems somehow broken. After a while, I get this:
```
⠋ [00:20:00] 20.82 GiB/20.82 GiB (91.4 MiB/s)
```
Ingestion rate should be in the order of 17MiB/s, but the number I got never seemed to go below 40MiB/s

fulmicoton:
oh really? let me check!  Thanks for testing it!

fulmicoton:
@trinity-1686a There was a bug in the rate estimator indeed. I added @guilload to the review.

trinity-1686a:
the estimator always seems to give a multiple of 5.71 for this workload. This is because it essentially count batches of (quasi-)fixed size, and there aren't that many batches processed over a 2s window. In my case, there are almost always 3 batches processed, so I get 17.1MiB most of the time, and 11.4 or 22.8 when 2 or 4 batches got processed.
Maybe for that estimator, it would be better to count the time it took to send the last X batches, rather than to count how many batches were sent in X seconds.

quickwit/quickwit-common/src/tower/rate_estimator.rs
```diff
@@ -252,36 +289,72 @@ mod tests {
...
+
+        // Now let's assume there isn't any work ongoing for 4s.
+        // Over the last 9 seconds, we have received 500 works
+        //
+        // After the reset, we should have the following buckets:
+        // We expect a mean of 44 work/s.
+        // |0, 0, 0, 0, 0, 100*, 100, 100, 100, 100|
+        //
+        // Since the current bucket (idx = 5) is not taken into account, this leads
+        // to an average of 500 / 9 = 55 work units.
```
trinity-1686a, line 357:
```suggestion
        // to an average of 400 / 9 = 44 work units.
```

fulmicoton:
I think there are different ways to improve the estimation, but I'll merge this for the bugfix. @trinity-1686a feel free to revisit it.

PR #4686, opened 2024-03-06
use total_cmp for f64

PSeitz:
use total_cmp for f64, which handles nan etc.
add inline (the reverse check halving may still cause double checks)


fulmicoton:
Sweet, I did not know about total_cmp!

PR #4688, opened 2024-03-06
Remove seed resolution validation

rdettai:
### Description

On ECS when using CloudMap the tasks register to the service DNS entries as they are started. This check might fail because the other peers haven't registered yet. But the the DNS-based discovery mechanism implemented in Chitchat would have built the cluster successfully.

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-config/src/node_config/mod.rs
+++ b/quickwit/quickwit-config/src/node_config/mod.rs
@@ -458,6 +458,9 @@ impl NodeConfig {
             }
             peer_seed_addrs.push(peer_seed_addr.to_string())
         }
+        if !self.peer_seeds.is_empty() && peer_seed_addrs.is_empty() {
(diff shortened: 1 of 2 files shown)


rdettai (commented):
What was the original reason for this check?

guilload:
The check is here because downstream errors resulting from the inability to resolve peer seeds will not be as clear and easily understood. You will get `metastore is unavailable` or something like that, which can happen for a zillion reasons.

Please log a warning instead.

rdettai:
We already have a warning for each unresolved host a few lines before:
https://github.com/quickwit-oss/quickwit/blob/d54479ac9d810c8598548e687316e7652dab2077/quickwit/quickwit-config/src/node_config/mod.rs#L455-L458

Do you want an extra one when none of the seed hostnames where resolved?

guilload:
Yes.

fulmicoton (approved):

PR #4689, opened 2024-03-06
Fix race condition related to debouncing.

fulmicoton:
In Ingest v1, queues are created on apply plans.
This means that right after creating an index, documents cannot be inserted.

This was true before, but was dependent on a race condition. It became more apparent after we added debouncing of control plane apply plans.

This PR adds code to make it possible for caller of debounced rebuild plan code, to get notified and wait for a rebuild to be done and applied.


quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -313,9 +313,17 @@ impl ControlPlane {
     ///
     /// This method includes some debouncing logic. Every call will be followed by a cooldown
     /// period.
-    fn rebuild_plan_debounced(&mut self, ctx: &ActorContext<Self>) {
+    ///
+    /// This method returns a future that can awaited to ensure that the relevant rebuild plan
```
guilload, line 317:
```suggestion
    /// This method returns a future that can be awaited to ensure that the relevant rebuild plan
```

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -25,10 +25,10 @@ use std::time::Duration;
 use anyhow::Context;
 use async_trait::async_trait;
 use futures::stream::FuturesUnordered;
-use futures::StreamExt;
+use futures::{Future, StreamExt};
 use quickwit_actors::{
-    Actor, ActorContext, ActorExitStatus, ActorHandle, Handler, Mailbox, Supervisor, Universe,
-    WeakMailbox,
+    Actor, ActorContext, ActorExitStatus, ActorHandle, DeferableReplyHandler, Handler, Mailbox,
```
guilload, line 30: `DeferableReplyHandler` is also something introduced to support ingest v1 that we should not forget to get rid of eventually.

quickwit/quickwit-control-plane/src/indexing_scheduler/change_tracker.rs
```diff
@@ -0,0 +1,125 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use std::sync::{Arc, Mutex};
+
+/// This object makes it possible to track for the completion of the next rebuild.
+pub struct RebuildNotifier {
+    generation_processed_tx: Arc<Mutex<tokio::sync::watch::Sender<usize>>>,
```
guilload, line 24: This seems like a [broadcast channel](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html).

guilload (approved):

quickwit/quickwit-control-plane/src/control_plane.rs
fulmicoton: I think it is actually quite nice.

quickwit/quickwit-control-plane/src/indexing_scheduler/change_tracker.rs
fulmicoton: Not exactly.
- broadcast don't let me look at the head from the sender
- broadcast channel usually have a capacity (we can set it to one, and older values are then dropped), but then we need to handle the associated error.

PR #4692, opened 2024-03-07
Ignore updates from node from previous generations

guilload:


quickwit/quickwit-cluster/src/grpc_gossip.rs
```diff
@@ -159,9 +159,9 @@ async fn wait_for_gossip_candidates(
             .skip_while(|node_states| {
                 node_states.len() < MAX_GOSSIP_ROUNDS
```
fulmicoton, line 160: MAX_GOSSIP_ROUNDS 's name is not exactly aligned with what it does.

fulmicoton (approved):

fulmicoton (approved):

quickwit/quickwit-cluster/src/grpc_gossip.rs
guilload: Yeah, it no longer makes sense. Will fix.

PR #4693, opened 2024-03-07
Add a TTL to positions.

### LINKED ISSUE #4143: Eventually Remove SourceUid/shards eventually from the published shard positions in chitchat
Right now the `PublishedShardPositions` model is ever growing.

We need to add some logic to eventually remove it.

fulmicoton:
This PR flattens the shard position.
Before we were publishing positions as
`index_uid:sourceuid: {shardpositions map}`

This PR breaks it into one entry per shard.

Key: `index_uid:source_id:shard`
Value: `position`

It then solves the deletion problem by simply adding a TTL to all of these key values.

Closes #4143

### DIFF
--- a/quickwit/quickwit-cluster/src/cluster.rs
+++ b/quickwit/quickwit-cluster/src/cluster.rs
@@ -269,15 +269,28 @@ impl Cluster {
             .set(key, value);
     }
 
+    /// Sets a key-value pair on the cluster node's state.
+    pub async fn set_self_key_value_delete_after_ttl(
+        &self,
+        key: impl Display,
+        value: impl Display,
+    ) {
+        let chitchat = self.chitchat().await;
+        let mut chitchat_lock = chitchat.lock().await;
+        let chitchat_self_node = chitchat_lock.self_node_state();
+        let key = key.to_string();
+        chitchat_self_node.set(key.clone(), value);
+        chitchat_self_node.delete_after_ttl(&key);
+    }
+
     pub async fn get_self_key_value(&self, key: &str) -> Option<String> {
         self.chitchat()
             .await
             .lock()
             .await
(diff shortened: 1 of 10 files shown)


quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -269,15 +269,28 @@ impl Cluster {
             .set(key, value);
     }
 
+    /// Sets a key-value pair on the cluster node's state.
+    pub async fn set_self_key_value_delete_after_ttl(
+        &self,
+        key: impl Display,
+        value: impl Display,
```
guilload, line 276:
```suggestion
        key: impl ToString,
        value: impl ToString,
```
nit: `Display` implements `ToString`, but `ToString` is more generic.

quickwit/quickwit-cluster/src/cluster.rs
```diff
@@ -269,15 +269,28 @@ impl Cluster {
...
+    ) {
+        let chitchat = self.chitchat().await;
+        let mut chitchat_lock = chitchat.lock().await;
+        let chitchat_self_node = chitchat_lock.self_node_state();
+        let key = key.to_string();
+        chitchat_self_node.set(key.clone(), value);
```
guilload, line 282: nit: Add `set_with_ttl` in Chitchat?

guilload (approved):

quickwit/quickwit-cluster/src/cluster.rs
fulmicoton: done

quickwit/quickwit-cluster/src/cluster.rs
fulmicoton: done

PR #4701, opened 2024-03-07
Measure amount of data in-flight in various buffers

guilload:
### Description
Per title.

### How was this PR tested?
Tested on Cicada.


fulmicoton (approved):
The call to drop on the GaugeVec guard is not entirely trivial.
It basically does a hash map lookup. We could embed the resolved object in the metrics directly. It would have the minor benefit of making the guard simpler and smaller (no more labels) and applicable to regular Gauge.

We would have the explicit list of labels in the same place, and the Drop would be as cheap as updating an Atomic.

quickwit/quickwit-serve/src/decompression.rs
```diff
@@ -78,10 +79,29 @@ pub(crate) struct UnsupportedEncoding(String);
...
+pub(crate) fn get_body_bytes() -> impl Filter<Extract = (Body,), Error = warp::Rejection> + Clone {
     warp::header::optional("content-encoding")
         .and(warp::body::bytes())
         .and_then(|encoding: Option<String>, body: Bytes| async move {
-            decompress_body(encoding, body).await
+            decompress_body(encoding, body).await.map(Body::from)
         })
 }
+
+pub(crate) struct Body {
```
fulmicoton, line 90: `Deref` (in moderation) can be useful for that kind of job.
By implementing `Deref<Target=Bytes>` on a `struct TrackedBytes { bytes: Bytes, guard }`  you can both get the tracking and all of the method of bytes.

I only use it on small object like this because the downside is that it is a bit magic.

fulmicoton (approved):

PR #4711, opened 2024-03-11
Added a metric for number of indexes

fulmicoton:


PSeitz (approved):
We have `ControlPlaneModelMetrics` and `ControlPlaneMetrics`. They look very similar, do we need both?

PR #4712, opened 2024-03-11
Stop zeroing buffers in the local file storage.

fulmicoton:


rdettai (approved):
If you want to avoid using `unsafe`, I think you can use `.take(num_bytes).read_to_end()`

fulmicoton:
Oh? @rdettai does that do only one alloc?

fulmicoton:
@rdettai this is a neat trick. Digging a little deeper. 
Take does not pass a size hint to `read_to_end(..)`.

This is not a problem per se. That means we would have to preallocate the vector, with `with_capacity(..)`.
We would just have to be careful to allocate with a capacity of max(len, 8000), because read_to_end allocate such a buffer.

I however did not dig too deep into the implementation of `default_read_to_end`. I think we can stick with the unsafe code considering how contains it is.

PSeitz:
> @rdettai this is a neat trick. Digging a little deeper. Take does not pass a size hint to `read_to_end(..)`.
> 
> This is not a problem per se. That means we would have to preallocate the vector, with `with_capacity(..)`. We would just have to be careful to allocate with a capacity of max(len, 8000), because read_to_end allocate such a buffer.
> 
> I however did not dig too deep into the implementation of `default_read_to_end`. I think we can stick with the unsafe code considering how contains it is.

I think with `take` we end up in the non specialized read version for the `Vec`, which is much slower usually.

PR #4720, opened 2024-03-12
Allocate new shards uniformly

guilload:
### Description
- Allocate new shards uniformly
- Remove ping RPC since now there is also an `InitShards` RPC.

### How was this PR tested?
- Added unit test
- Will test on Cicada


quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -416,66 +311,168 @@ impl IngestController {
...
-                let index_uid: IndexUid = open_shards_subresponse.index_uid().clone();
-                let source_id = open_shards_subresponse.source_id.clone();
-                model.insert_shards(
-                    &index_uid,
-                    &source_id,
-                    open_shards_subresponse.opened_shards,
-                );
-                if let Some(open_shard_entries) =
-                    model.find_open_shards(&index_uid, &source_id, &unavailable_leaders)
+            if let Some(leader_follower_pairs) =
```
guilload, line 328: This function has become quite lengthy and messy. I will refactor it in a follow-up PR, most likely with a workbench, since many steps are involved in opening a new shard.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -407,66 +302,171 @@ impl IngestController {
...
+                .saturating_sub(num_open_shards_inner)
+                .min(num_remaining_shards_to_allocate);
+
+            for _ in 0..num_shards_to_allocate_inner {
+                num_remaining_shards_to_allocate -= 1;
+
+                let leader = leader_id.clone();
+                let mut follower_opt = None;
+
+                if self.replication_factor > 1 {
```
guilload, line 448: I think the mirroring of the primary shards balanced allocation "shifted by one ingester" for replicas is broken because of the `unavailable_leaders` thing. I'll rework this part.

fulmicoton (approved):

PR #4721, opened 2024-03-13
Open shards metrics was only updated on scale up/down.

fulmicoton:
This PR also removes the source id label (as it will always be ingest
v2). This is not much of a win since the overall metric cardinality does
not change.

It also removes `.all_shards_mut()` which was error-prone
as a client could break all kinds of invariants.

### DIFF
--- a/quickwit/quickwit-control-plane/src/control_plane.rs
+++ b/quickwit/quickwit-control-plane/src/control_plane.rs
@@ -513,6 +513,7 @@ impl Handler<DeleteIndexRequest> for ControlPlane {
         ctx: &ActorContext<Self>,
     ) -> Result<Self::Reply, ActorExitStatus> {
         let index_uid: IndexUid = request.index_uid().clone();
+        info!(index=%index_uid, "delete index");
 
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -513,6 +513,7 @@ impl Handler<DeleteIndexRequest> for ControlPlane {
         ctx: &ActorContext<Self>,
     ) -> Result<Self::Reply, ActorExitStatus> {
         let index_uid: IndexUid = request.index_uid().clone();
+        info!(index=%index_uid, "delete index");
```
guilload, line 516: nit: add rather put those statements after the successful op.

quickwit/quickwit-control-plane/src/metrics.rs
```diff
@@ -62,7 +62,7 @@ impl Default for ControlPlaneMetrics {
                 "Number of open shards per source.",
```
guilload, line 62:
```suggestion
                "Number of open shards per indexer.",
```

guilload (approved):

PR #4728, opened 2024-03-14
Wal allocated into inflight data

fulmicoton:
This moves the WAL allocated memory into the inflight data metrics.

### DIFF
--- a/quickwit/quickwit-common/src/metrics.rs
+++ b/quickwit/quickwit-common/src/metrics.rs
@@ -252,12 +252,13 @@ impl Default for MemoryMetrics {
 
 #[derive(Clone)]
 pub struct InFlightDataGauges {
-    pub doc_processor_mailbox: IntGauge,
-    pub index_writer: IntGauge,
-    pub indexer_mailbox: IntGauge,
-    pub ingest_router: IntGauge,
     pub rest_server: IntGauge,
+    pub ingest_router: IntGauge,
+    pub wal: IntGauge,
     pub sources: InFlightDataSourceGauges,
+    pub doc_processor_mailbox: IntGauge,
+    pub indexer_mailbox: IntGauge,
+    pub index_writer: IntGauge,
 }
 
 impl Default for InFlightDataGauges {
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-common/src/metrics.rs
```diff
@@ -252,12 +252,13 @@ impl Default for MemoryMetrics {
 
 #[derive(Clone)]
 pub struct InFlightDataGauges {
-    pub doc_processor_mailbox: IntGauge,
-    pub index_writer: IntGauge,
-    pub indexer_mailbox: IntGauge,
-    pub ingest_router: IntGauge,
     pub rest_server: IntGauge,
+    pub ingest_router: IntGauge,
+    pub wal: IntGauge,
```
fulmicoton, line 257: I reorder stuff in the order of the stream.

guilload: 👍

quickwit/quickwit-indexing/src/metrics.rs
```diff
@@ -66,6 +67,12 @@ impl Default for IndexerMetrics {
                 &[],
                 ["component"],
             ),
+            split_builders_total: new_gauge(
+                "split_builders_total",
+                "Number of existing index writer instance",
```
guilload, line 72:
```suggestion
                "Number of existing index writer instances.",
```

quickwit/quickwit-indexing/src/metrics.rs
```diff
@@ -66,6 +67,12 @@ impl Default for IndexerMetrics {
                 &[],
                 ["component"],
             ),
+            split_builders_total: new_gauge(
+                "split_builders_total",
```
guilload, line 71:
```suggestion
                "split_builders",
```
According to the Prom naming conventions, the `_total` suffix is reserved for accumulating count, i.e., counters. Because of that,  I have a hard time naming gauges, too...

guilload (approved):

quickwit/quickwit-indexing/src/metrics.rs
fulmicoton: oh right!

PR #4733, opened 2024-03-14
remove unused DocId enum variant

PSeitz:
The DocId variant in the score extractor is unused since there is nothing to extract.

### Description

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -82,9 +82,10 @@ impl SortByComponent {
     fn to_sorting_field_extractor_component(
         &self,
         segment_reader: &SegmentReader,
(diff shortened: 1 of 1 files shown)


fulmicoton (commented):
Why is this an improvement?

It is `Option<EnumWithoutDocIdVariant>` vs `EnumWithDocIdVariant` isn't it?

PSeitz:
> Why is this an improvement?
> 
> It is `Option<EnumWithoutDocIdVariant>` vs `EnumWithDocIdVariant` isn't it?

The `Option<EnumWithoutDocIdVariant>` is just the consequence, when we remove the unused `SortingFieldExtractorComponent::DocId`. 

But it seems to be still used outside of the segment collector. When reading the segment collector part, it's a little bit misleading.

PR #4747, opened 2024-03-15
Improve logging on `run` command error

guilload:


quickwit/quickwit-serve/src/lib.rs
```diff
@@ -511,18 +524,23 @@ pub async fn serve_quickwit(
...
+            ] {
                 match index_manager.create_index(index_config, false).await {
                     Ok(_)
                     | Err(IndexServiceError::Metastore(MetastoreError::AlreadyExists(
                         EntityKind::Index { .. },
-                    ))) => Ok(()),
-                    Err(error) => Err(error),
-                }?;
+                    ))) => {}
+                    Err(error) => bail!("failed to create {index_name} index: {error}",),
```
guilload, line 542: We could also just log the error here. WDYT?

guilload:
@fulmicoton, friendly nudge.

quickwit/quickwit-serve/src/lib.rs
fulmicoton: Yes... A error log might be useful here too.

fulmicoton (approved):

quickwit/quickwit-serve/src/lib.rs
fulmicoton: ah... I don't have a clear opinion right now.

guilload: Ok. I'll leave it as is for now. Also, all errors in `quickwit_serve` are logged in `main`.

PR #4752, opened 2024-03-16
Add storage metrics for opendal storage.

fmassot:


quickwit/quickwit-storage/src/opendal_storage/base.rs
```diff
@@ -88,20 +88,35 @@ impl Storage for OpendalStorage {
         tokio::io::copy(&mut payload_reader, &mut storage_writer).await?;
         storage_writer.close().await?;
 
+        // FIXME: should be payload.len() / ByteSize::mb(8)?
```
Xuanwo, line 91: Yes.

PR #4753, opened 2024-03-17
collect_block in QuickwitCollector

PSeitz:
collect_block + using `first_vals` to batch fetch sort values

qw_collect_block_bench

### DIFF
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -785,7 +785,7 @@ pub use tantivy::COLLECT_BLOCK_BUFFER_LEN;
 fn compute_filtered_block<'a>(
     timestamp_filter_opt: &Option<TimestampFilter>,
     docs: &'a [DocId],
-    filtered_docs_buffer: &'a mut [DocId; BUFFER_LEN],
+    filtered_docs_buffer: &'a mut [DocId; COLLECT_BLOCK_BUFFER_LEN],
 ) -> &'a [DocId] {
     let Some(timestamp_filter) = &timestamp_filter_opt else {
         return docs;
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -599,7 +599,7 @@ impl QuickwitSegmentTopKCollector {
             &mut self.sort_values1[..],
             &mut self.sort_values2[..],
         );
-        if let Some(_search_after) = &self.search_after {
+        if self.search_after.is_some() {
             // Search after not optimized for block collection yet
             for ((doc_id, sort_value), sort_value2) in docs
                 .iter()
@@ -775,7 +775,7 @@ impl SegmentPartialHit {
     }
 }
 
-const BUFFER_LEN: usize = tantivy::COLLECT_BLOCK_BUFFER_LEN;
+pub use tantivy::COLLECT_BLOCK_BUFFER_LEN;
 /// Store the filtered docs in `filtered_docs_buffer` if `timestamp_filter_opt` is present.
 ///
 /// Returns the number of docs.
@@ -800,7 +800,6 @@ fn compute_filtered_block<'a>(
         };
     }
     &filtered_docs_buffer[..len]
-    // len
 }
 
 impl SegmentCollector for QuickwitSegmentCollector {
@@ -818,7 +817,6 @@ impl SegmentCollector for QuickwitSegmentCollector {
         self.num_hits += filtered_docs.len() as u64;
 
         if let Some(segment_top_k_collector) = self.segment_top_k_collector.as_mut() {
-            //.top_k_hits.max_len() != 0 {
             segment_top_k_collector.collect_top_k_block(filtered_docs);
         }
 
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -150,6 +150,20 @@ enum SortingFieldExtractorComponent {
 }
 
 impl SortingFieldExtractorComponent {
+    fn is_fast_field(&self) -> bool {
+        matches!(self, SortingFieldExtractorComponent::FastField { .. })
+    }
+    /// Loads the fast field values for the given doc_ids in its u64 representation. The returned
+    /// u64 representation maintains the ordering of the original value.
+    #[inline]
+    fn extract_typed_sort_values_block(&self, doc_ids: &[DocId], values: &mut [Option<u64>]) {
+        // In the collect block case we don't have scores to extract
+        if let SortingFieldExtractorComponent::FastField { sort_column, .. } = self {
+            let values = &mut values[..doc_ids.len()];
+            sort_column.first_vals(doc_ids, values);
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-search/src/collector.rs
```diff
@@ -543,15 +578,107 @@ impl SearchAfterSegment {
...
+                    doc_id,
+                    sort_value,
+                    sort_value2,
+                    &self.search_after,
+                    self.precomp_search_after_order,
+                    &mut self.top_k_hits,
+                );
+            }
+        } else {
+            // Probaly would make sense to check the fence against e.g. sort_values1 earlier,
```
fulmicoton, line 611:
```suggestion
            // Probably would make sense to check the fence against e.g. sort_values1 earlier,
```

quickwit/quickwit-search/src/collector.rs
```diff
@@ -635,9 +794,65 @@ impl SegmentPartialHit {
...
+            compute_filtered_block(&self.timestamp_filter_opt, docs, &mut self.filtered_docs);
+
+        // Update results
+        self.num_hits += num_docs as u64;
+
+        if self.top_k_hits.max_len() != 0 {
+            self.collect_top_k_block(num_docs, docs);
+        }
+
+        let docs = get_final_docs(
```
fulmicoton, line 839:
```suggestion
        let docs: &[DocId] = get_final_docs(
```

quickwit/quickwit-search/src/collector.rs
```diff
@@ -635,9 +794,65 @@ impl SegmentPartialHit {
...
+    #[inline]
+    fn collect_block(&mut self, docs: &[DocId]) {
+        let num_docs =
+            compute_filtered_block(&self.timestamp_filter_opt, docs, &mut self.filtered_docs);
+
+        // Update results
+        self.num_hits += num_docs as u64;
+
+        if self.top_k_hits.max_len() != 0 {
+            self.collect_top_k_block(num_docs, docs);
```
fulmicoton, line 836: At this point, docs has not gone through the timestamp filter. Did you mean to put it one line below?

quickwit/quickwit-search/src/collector.rs
```diff
@@ -593,6 +734,24 @@ impl QuickwitSegmentCollector {
     }
 }
 
+/// This is to circumvent lifetime issues and is pretty terrible.
+///
+/// We either use `filtered_docs` buffer (computed in `compute_filtered_block`) or return the passed
+/// docs.
+#[inline]
+fn get_final_docs<'a>(
```
fulmicoton, line 742: Any better solution for this, it is really quite terrible.

fulmicoton: I think it can work for instance if we split the SegmentCollector in two...

The filtered_docs and the rest of the member.

PSeitz: I replaced it by removing `self` from the top_k collection method and passing the parameters instead

quickwit/quickwit-search/src/collector.rs
```diff
@@ -635,9 +788,70 @@ impl SegmentPartialHit {
...
 impl SegmentCollector for QuickwitSegmentCollector {
     type Fruit = tantivy::Result<LeafSearchResponse>;
 
+    #[inline]
+    fn collect_block(&mut self, docs: &[DocId]) {
+        let block_buffer = &mut self.block_buffer;
+        let filtered_docs = &mut block_buffer.filtered_docs;
+        let sort_values1 = &mut block_buffer.sort_values1;
+        let sort_values2 = &mut block_buffer.sort_values2;
+        let docs = compute_filtered_block(&self.timestamp_filter_opt, docs, filtered_docs);
```
fulmicoton, line 827: alternative solution.
https://github.com/quickwit-oss/quickwit/compare/faster_collect2?expand=1

quickwit/quickwit-search/src/collector.rs
```diff
@@ -542,16 +592,100 @@ impl SearchAfterSegment {
...
-        let (sort_value, sort_value2): (Option<u64>, Option<u64>) =
-            self.score_extractor.extract_typed_sort_value(doc_id, score);
+impl QuickwitSegmentTopKCollector {
+    fn collect_top_k_block(&mut self, docs: &[DocId]) {
+        self.score_extractor.extract_typed_sort_values(
+            docs,
+            &mut self.sort_values1[..],
+            &mut self.sort_values2[..],
+        );
+        if let Some(_search_after) = &self.search_after {
```
fulmicoton, line 602:
```suggestion
        if self.search_after.is_some() {
```

quickwit/quickwit-search/src/collector.rs
```diff
@@ -635,17 +775,74 @@ impl SegmentPartialHit {
     }
 }
 
+const BUFFER_LEN: usize = tantivy::COLLECT_BLOCK_BUFFER_LEN;
```
fulmicoton, line 778:
```suggestion
pub use tantivy::COLLECT_BLOCK_BUFFER_LEN;
```

quickwit/quickwit-search/src/collector.rs
```diff
@@ -635,17 +775,74 @@ impl SegmentPartialHit {
...
+    docs: &'a [DocId],
+    filtered_docs_buffer: &'a mut [DocId; BUFFER_LEN],
+) -> &'a [DocId] {
+    let Some(timestamp_filter) = &timestamp_filter_opt else {
+        return docs;
+    };
+    let mut len = 0;
+    for &doc in docs {
+        filtered_docs_buffer[len] = doc;
+        len += if timestamp_filter.is_within_range(doc) {
```
fulmicoton, line 796: Should we try to vectorize this too?

quickwit/quickwit-search/src/collector.rs
```diff
@@ -635,17 +775,74 @@ impl SegmentPartialHit {
...
+            &self.timestamp_filter_opt,
+            unfiltered_docs,
+            &mut self.filtered_docs,
+        );
+
+        // Update results
+        self.num_hits += filtered_docs.len() as u64;
+
+        if let Some(segment_top_k_collector) = self.segment_top_k_collector.as_mut() {
+            //.top_k_hits.max_len() != 0 {
```
fulmicoton, line 821:
```suggestion
```

fulmicoton (changes requested):
a few minor comments.

quickwit/quickwit-search/src/collector.rs
PSeitz: Fetching the values as block, yes definitely. 

Filtering via SIMD maybe, the values are `u64`, so that would be just 4 integers per block on AVX2.
The fastest variant would probably be a pushdown to the codec where we could filter on the compressed data.
Although I'm not sure if the overhead is not too high with a block size of just 64 elements.

I'm also thinking that maybe we should use a branch here. Since the data we have is mostly ordered by timestamp the branch predictor should have a pretty easy job.

fulmicoton: Agreed on all points and doubts.

fulmicoton (approved):

PR #4761, opened 2024-03-19
document scroll based on search after and docstore config

### LINKED ISSUE #4717: document docstore_blocksize and docstore_compression_level
afaict, these are not documented, but tuning them can be reasonable in some cases

### LINKED ISSUE #4718: scroll documentation says we souldn't go further than 10k, that isn't true anymore
since scroll is based on search after, you can now go arbitrarily far (but have to do it before your scroll gets too old and is no longer valid)

trinity-1686a:
### Description

fix #4717
fix #4718
fix out of memory on crafted scroll_id causing to allocate up to 2**32-1 partial hits (roughly 200GB)

### DIFF
--- a/quickwit/quickwit-search/src/scroll_context.rs
+++ b/quickwit/quickwit-search/src/scroll_context.rs
@@ -222,6 +222,9 @@ impl FromStr for ScrollKeyAndStartOffset {
         let scroll_ulid = u128::from_le_bytes(scroll_ulid_bytes.try_into().unwrap()).into();
         let from = u64::from_le_bytes(from_bytes.try_into().unwrap());
         let max_hits = u32::from_le_bytes(max_hits_bytes.try_into().unwrap());
+        if max_hits > 10_000 {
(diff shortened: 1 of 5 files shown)


fmassot (approved):

quickwit/quickwit-search/src/scroll_context.rs
```diff
@@ -222,6 +222,9 @@ impl FromStr for ScrollKeyAndStartOffset {
         let scroll_ulid = u128::from_le_bytes(scroll_ulid_bytes.try_into().unwrap()).into();
         let from = u64::from_le_bytes(from_bytes.try_into().unwrap());
         let max_hits = u32::from_le_bytes(max_hits_bytes.try_into().unwrap());
+        if max_hits > 10_000 {
```
fmassot, line 225: we can get there only if a user tried to forge a bad scroll ID right?

trinity-1686a: yes

PR #4765, opened 2024-03-20
Add terraform module for ECS

rdettai:
### Description

Add Terraform scripts with:
- a module that can be used to deploy a general purpose Quickwit setup using ECS and RDS
- an self contained example stack that uses that module and includes supporting network configurations (VPC and bastion)

### How was this PR tested?

Executed on AWS with various configurations.


distribution/ecs/example/bastion.tf
```diff
@@ -0,0 +1,65 @@
...
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  egress {
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
```
trinity-1686a, line 44: reading [this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group#removing-all-ingress-and-egress-rules), it looks like `egress  = []` is the more natural way to write that

distribution/ecs/example/kafka.tf
```diff
@@ -0,0 +1,58 @@
...
+      mount_points = [
+        {
+          sourceVolume  = "quickwit-keys"
+          containerPath = "/quickwit/keys"
+        }
+      ]
+      secrets = [
+        {
+          name      = "CA_PEM"
+          valueFrom = "arn:aws:secretsmanager:eu-west-1:542709600413:secret:your_kafka_ca_pem"
```
trinity-1686a, line 38: (is that a well known arn or something?)

distribution/ecs/example/logging.tf
```diff
@@ -0,0 +1,94 @@
+# Example configuration for pushing ECS logs to Datadog
```
trinity-1686a, line 1: sending logs to dd as qw strikes me as odd

trinity-1686a (commented):
looks sensible, though I lack knowledge in terraform to confidently review that kind of PR

distribution/ecs/example/kafka.tf
rdettai: no, these are example values. Not sure how to make that more obvious

rdettai: using 123456789 to make it more obvious that this is an example

distribution/ecs/example/bastion.tf
rdettai: my understanding of the docs is that `egress = []` means "no egress rule", i.e "no egress allowed". The rule configured here is the exact opposite, it means "all egress allowed".

rdettai: image

trinity-1686a (approved):

PR #4766, opened 2024-03-20
allow any valid field in routing expression

### LINKED ISSUE #4730: Allow special characters on partition_key
**Describe the bug**

There are some special characters such as @, $, /, which are allowed for field name of a index, but not allowed in partition_key validation.
https://quickwit.io/docs/configuration/index-config#field-name-validation-rules

Quickwit should add support these characters on **partition_key**, for example Kubernetes logs use cases.

trinity-1686a:
### Description

fix #4730
allow any valid field in routing expression
also fix round-trip of `RoutingExpr` <=> `String`

### How was this PR tested?

added unit test

### DIFF
--- a/quickwit/quickwit-doc-mapper/src/routing_expression/mod.rs
+++ b/quickwit/quickwit-doc-mapper/src/routing_expression/mod.rs
@@ -323,8 +323,14 @@ impl Display for InnerRoutingExpr {
                     if index != 0 {
                         f.write_str(".")?;
                     }
-                    // TODO actually we should escape .
-                    f.write_str(part)?;
+                    if !part.contains('.') {
+                        f.write_str(part)?;
+                    } else {
+                        // TODO use std version once stabilized
+                        for subpart in itertools::Itertools::intersperse(part.split('.'), r"\.") {
+                            f.write_str(subpart)?;
+                        }
+                    }
                 }
             }
             InnerRoutingExpr::Composite(children) => {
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-doc-mapper/src/routing_expression/mod.rs
```diff
@@ -323,8 +323,14 @@ impl Display for InnerRoutingExpr {
...
-                    // TODO actually we should escape .
-                    f.write_str(part)?;
+                    if !part.contains('.') {
+                        f.write_str(part)?;
+                    } else {
+                        // TODO use std version once stabilized
+                        for subpart in itertools::Itertools::intersperse(part.split('.'), r"\.") {
+                            f.write_str(subpart)?;
+                        }
+                    }
```
fulmicoton, line 333: I'm  a bit lost. When is this block different from `part.replace(".", r"\.")`

trinity-1686a: hum, never actually. It saves an allocation, but here we don't care

quickwit/quickwit-doc-mapper/src/routing_expression/mod.rs
```diff
@@ -417,7 +416,7 @@ mod expression_dsl {
 
     fn identifier(input: &str) -> IResult<&str, &str> {
```
fulmicoton, line 417: can you add a comment with an example of what an identifier is...
same with key_identifier.

guilload:
@fulmicoton @trinity-1686a, it seems like this PR is close to being mergeable. I invite you not to let it rot for too long.

fulmicoton (approved):

PR #4770, opened 2024-03-21
fetch timestamp in blocks

PSeitz:
### Description

### How was this PR tested?


fulmicoton:
Does it help?

PSeitz:
> Does it help?

Updated with benchmark numbers. On the `range` query yes, which is just a range filter on the timestamp. The rest is not really discernible from noise.

Overall it's a relative small gain, on larger queries spanning multiple splits, the filter can be removed since the time frame spans most of the splits completely and can be removed.

fulmicoton (approved):

PR #4773, opened 2024-03-21
add support for concatenated field type

### LINKED ISSUE #4548: Field concatenation
It would be nice to have a "concatenate" field type.

The idea is to have it as a default field on search, to enable the use case of users searching for a given id.
If different tokenizer are required depending on the field, the user will have to create on tokenizer per concatenate field.

That field will be non-fast, non-stored but indexed.
For numeric types or dates, we should coerce them to string. Maybe we can skip float (!?) as they will likely not be retrievable anyway.

```
type: concatenate
name: "my_default_field"
concatenated_fields:
      - text
      - body
      - params_* # we allow params
      - _dynamic # json field will be browsed
tokenizer: default # tokenizing by default should make sense too
record: basic # by default basic should make sense
```

Open questions:
Should we do RFC-3339 with a space instead of the T between date an hour.
SHould we index positions for RFC-3339?
Any creative ideas there?

trinity-1686a:
### Description

fix #4548
implement concatenated field type: a field which search inside other fields, in a more efficient way than querying each sub-field independently. Currently doesn't support having date and ip sub-fields. These can be added later, but may require some trickery on the search path to get working properly.

### How was this PR tested?
added integration test testing we use the right tokenizer, and are able to search for supported types (int, bool, text, things inside json field, and in dynamic field)


docs/configuration/index-config.md
```diff
@@ -410,6 +412,53 @@ field_mappings:
...
+Quickwit supports mapping the content of multiple fields to a single one. This can be more efficient at query time than
+searching through dozens of `default_search_fields`. It also allow querying inside a json field without knowing the path
+to the field being searched.
+
+```yaml
+name: my_default_field
+type: concatenate
+concatenated_fields:
+  - text # things inside text, tokenized with the `default` tokenizer
+  - resource.author # all fields in resource.author, assuming resource is a `json` field
```
fulmicoton, line 426:
```suggestion
  - resource.author # all fields in resource.author, assuming resource is a `json` field
```
Does it work on an object type too?

docs/configuration/index-config.md
```diff
@@ -410,6 +412,53 @@ field_mappings:
...
+  - _dynamic # content of the dynamic field, assuming indexing mode is `dynamic` (see below)
+tokenizer: default
+record: basic
+```
+
+Concatenate fields don't support fast fields, and are never stored. They uses their own tokenizer, independantly of the
+tokenizer configured on the individual fields.
+At query time, concatenate fields don't support range queries.
+Only the following types are supported inside a concatenate field: text, bool, i64, u64, json. Other types are reject
```
fulmicoton, line 435:
```suggestion
Only the following types are supported inside a concatenate field: text, bool, i64, u64, json. Other types are rejected
```

docs/configuration/index-config.md
```diff
@@ -410,6 +412,53 @@ field_mappings:
...
+at index creation, or discarded during indexation if they are found inside a json field.
```
fulmicoton, line 436:
```suggestion
at index creation, or silently discarded during indexation if they are found inside a json field.
```

docs/configuration/index-config.md
```diff
@@ -410,6 +412,53 @@ field_mappings:
...
+For json fields, they path is not indexed, only values are.
```
fulmicoton, line 437: An example would be useful.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -535,6 +544,18 @@ impl DocMapper for DefaultDocMapper {
 
         if let Some(dynamic_field) = self.dynamic_field {
             if !dynamic_json_obj.is_empty() {
+                if !self.concate_dynamic_fields.is_empty() {
```
fulmicoton, line 547: concate -> concatenate

docs/configuration/index-config.md
```diff
@@ -410,6 +412,53 @@ field_mappings:
...
+searching through dozens of `default_search_fields`. It also allow querying inside a json field without knowing the path
+to the field being searched.
+
+```yaml
+name: my_default_field
+type: concatenate
+concatenated_fields:
+  - text # things inside text, tokenized with the `default` tokenizer
+  - resource.author # all fields in resource.author, assuming resource is a `json` field
+  - _dynamic # content of the dynamic field, assuming indexing mode is `dynamic` (see below)
```
fulmicoton, line 427: I am not fond of this.  "_dynamic" is an internal implementation detail. I don't think this is "public API" at this point. I wonder if this should be a bool in the config.

`include_dynamic_fields: true`

or something like that.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -535,6 +544,18 @@ impl DocMapper for DefaultDocMapper {
...
+                    let json_obj_values = dynamic_json_obj
+                        .iter()
+                        .flat_map(|(_key, val)| JsonValueIterator::new(val.clone()))
```
fulmicoton, line 550:
```suggestion
                        .values()
                        .flat_map(|child_json_value| JsonValueIterator::new(child_json_value.clone()))
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -535,6 +544,18 @@ impl DocMapper for DefaultDocMapper {
...
+                        .flat_map(map_primitive_json_to_tantivy);
+
+                    for value in json_obj_values {
+                        for concate_dynamic_field in &self.concate_dynamic_fields {
+                            document.add_field_value(*concate_dynamic_field, value.clone());
```
fulmicoton, line 555: the clone is a bit brutal

fulmicoton, line 555: Do you know a simple trick to remove the last clone? 

I haven't found anything simple in the library.
It feels sad if we have a single concate_dynamic_fields (the most common case)

quickwit/quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
```diff
@@ -649,6 +723,13 @@ fn deserialize_mapping_type(
             }
             return Ok(FieldMappingType::Object(object_options));
         }
+        QuickwitFieldType::Concatenate => {
+            let concatenate_options: QuickwitConcatenateOptions = serde_json::from_value(json)?;
+            if concatenate_options.concatenate_fields.is_empty() {
+                anyhow::bail!("concatenate type must have at least one sub-field");
```
fulmicoton, line 729: if we do the dynamic property thingy, we need to remove this check.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -54,6 +55,124 @@ pub enum LeafType {
     Text(QuickwitTextOptions),
 }
 
+fn value_to_pretokenized<T: ToString>(val: T) -> PreTokenizedString {
+    let val = val.to_string();
```
fulmicoton, line 59: no shadowing pls

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -54,6 +55,124 @@ pub enum LeafType {
...
+    PreTokenizedString {
+        text: val.clone(),
+        tokens: vec![Token {
+            offset_from: 0,
+            offset_to: 0,
```
fulmicoton, line 64: this is wrong.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -54,6 +55,124 @@ pub enum LeafType {
     Text(QuickwitTextOptions),
 }
 
+fn value_to_pretokenized<T: ToString>(val: T) -> PreTokenizedString {
```
fulmicoton, line 58: is there a discussion somewhere about why we are going for the string route? I think we discussed it but forgot the rationale.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
fulmicoton: Would it have been easier to build a `serde_json::Value` and pass it to `JsonValueIterator::new`?

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -100,13 +219,76 @@ impl LeafType {
             }
         }
     }
+
+    fn string_value_from_json(
```
fulmicoton, line 223: the name says something different from the prototype.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
trinity-1686a: I don't think we had that discussion. I based my work on
> For numeric types or dates, we should coerce them to string.

(from the ticket).
The rational I see would be that if we want to do point search (not range &co), casting everything to string is the simplest way, and work rather well. Maybe we could leverange json field instead, but I don't think there is much benefit, and you end up doing twice as many requests to storage when you don't know if a term is an integer or a string.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -100,13 +219,76 @@ impl LeafType {
...
+            LeafType::IpAddr(_) => Err("unsuported concat type: IpAddr".to_string()),
+            LeafType::DateTime(_date_time_options) => {
+                Err("unsuported concat type: DateTime".to_string())
+            }
+            LeafType::Bytes(_binary_options) => Err("unsuported concat type: DateTime".to_string()),
+            LeafType::Json(_) => {
+                if let JsonValue::Object(json_obj) = json_val {
+                    Ok(OneOrIter::Iter(
+                        json_obj
+                            .into_iter()
```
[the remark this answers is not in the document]
trinity-1686a: `serde_json::Map` doesn't have an `into_values`, only a `values()` and a `values_mut()` which would force a useless clone

docs/configuration/index-config.md
trinity-1686a: i got a bit confused when redoing the documentation. This is for object field, not json fields.  Currently we wouldn't visit inside author if it was an object. The code is architectured so it wouldn't be hard to add, and would likely be done at the same time as adding support for wildcard names (like `param_*`)

quickwit/quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
trinity-1686a: I've made the check take into account `include_dynamic_fields`

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
trinity-1686a: i've set the offset_to to 1, is that alright, or should that be `text.len()`?

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
trinity-1686a: I couldn't find a good way to do that, so I created a custom iterator which does the trick (`ZipCloneable`). Let me know if you think that's better or not

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -501,6 +510,49 @@ fn populate_field_presence_for_json_obj(
...
+fn zip_cloneable<T, I: Iterator<Item = T>, U: Clone>(iter: I, item: U) -> ZipCloneable<T, I, U> {
+    let mut inner = iter.peekable();
+    if inner.peek().is_some() {
+        ZipCloneable::Running { inner, item }
+    } else {
+        ZipCloneable::Ended
+    }
+}
+
+/// An iterator which zip a value alongside another iterator, clonning it each time it yields,
```
fulmicoton, line 522:
```suggestion
/// An iterator which zip a value alongside another iterator, cloning it each time it yields,
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -501,6 +510,49 @@ fn populate_field_presence_for_json_obj(
     }
 }
 
+fn zip_cloneable<T, I: Iterator<Item = T>, U: Clone>(iter: I, item: U) -> ZipCloneable<T, I, U> {
```
fulmicoton, line 513: can we have a unit test for this?

quickwit/quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
```diff
@@ -353,6 +353,13 @@ impl TextIndexingOptions {
         }
     }
 
+    fn from_parts_concatenate(
+        tokenizer: Option<QuickwitTextTokenizer>,
+        record: Option<IndexRecordOption>,
+    ) -> anyhow::Result<Self> {
+        Self::from_parts_text(true, tokenizer, record, false).map(Option::unwrap)
```
fulmicoton, line 360: Nitpick:
`.map(Option::unwrap)` is difficult to read.

```suggestion
        let text_index_options_opt: Option<TextIndexingOptions> =  TextIndexingOptions::from_parts_text(true, tokenizer, record, false)?;
        let text_index_options = text_index_options_opt.expect("an indexed field should never result in a `None` `TextIndexingOptions`.");
        Ok(text_index_options)
```

fulmicoton (changes requested):
Please add unit tests.

fulmicoton:
@trinity-1686a I don't have more comments apart from the need for unit test. integration tests are not a replacement for UT.
Cool! This is quite an important feature for 0.9! I look forward to it being shipped

@PSeitz can you have a look at the CR? If you see optimization hanging fruit, and they can be shipped separately, I think we can revisit that later.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
```diff
@@ -353,6 +353,15 @@ impl TextIndexingOptions {
         }
     }
 
+    fn from_parts_concatenate(
+        tokenizer: Option<QuickwitTextTokenizer>,
+        record: Option<IndexRecordOption>,
+    ) -> anyhow::Result<Self> {
+        let text_index_options_opt = Self::from_parts_text(true, tokenizer, record, false)?;
+        let text_index_options = text_index_options_opt.expect("concatenate field must be indexed");
```
PSeitz, line 361: Can this be triggered by user input?

trinity-1686a: `Self::from_parts_text` the line above will always return Ok(Some(_)) if it's first parameter is `true`

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -246,20 +451,18 @@ fn insert_json_val(
 trait NumVal: Sized + FromStr + ToString + Into<TantivyValue> {
     fn from_json_number(num: &serde_json::Number) -> Option<Self>;
 
-    fn from_json(json_val: JsonValue, coerce: bool) -> Result<TantivyValue, String> {
+    fn from_json_to_self(json_val: JsonValue, coerce: bool) -> Result<Self, String> {
```
fulmicoton, line 454:
```suggestion
    fn from_json(json_val: JsonValue, coerce: bool) -> Result<Self, String> {
```
I think to_self is implied?

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -284,6 +487,10 @@ trait NumVal: Sized + FromStr + ToString + Into<TantivyValue> {
         }
     }
 
+    fn from_json(json_val: JsonValue, coerce: bool) -> Result<TantivyValue, String> {
+        Self::from_json_to_self(json_val, coerce).map(Self::into)
```
fulmicoton, line 491:
```suggestion
        Self::from_json_to_self(json_val, coerce).map(TantivyValue::from)
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -377,6 +584,37 @@ impl MappingNode {
         }
     }
 
+    /// Finds the field mapping type for a given field path in the mapping tree.
```
fulmicoton, line 587: I know this is not part of this PR, but can we simplify `573-577` ? It seems like the look up in branches order makes no sense.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -377,6 +584,37 @@ impl MappingNode {
...
+    fn internal_find_field_mapping_leaf(
+        &mut self,
+        field_path: &[String],
+    ) -> Option<impl Iterator<Item = &mut MappingLeaf>> {
+        let (first_path_fragment, sub_field_path) = field_path.split_first()?;
+        let field_name = self
+            .branches_order
+            .iter()
+            .find(|name| name == &first_path_fragment)?;
+        let child_tree = self.branches.get_mut(field_name).expect("Missing field");
```
fulmicoton, line 607: is it the same as
```suggestion
        let child_tree = self.branches.get_mut(first_path_fragment)?;
```

or can `branches_order` contain a subset of the child nodes?

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -543,18 +781,57 @@ fn build_mapping_tree_from_entries<'a>(
...
+        let name = &concatenate_field_entry.name;
+        if mapping_node.branches.contains_key(name) {
+            bail!("duplicated field definition `{}`", name);
+        }
+        let text_options: TextOptions = options.clone().into();
+        let field = schema.add_text_field(name, text_options);
+        for sub_field in &options.concatenate_fields {
+            for matched_field in
+                mapping_node
+                    .find_field_mapping_leaf(sub_field)
```
fulmicoton, line 816: What happens for json field?

If I have a json field called attributes and I concat
`attributes.color`, is this an error, or is not an error but working in strange ways?

fulmicoton: If the spec is acceptable, can we have a unit test for this?

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -1514,119 +1582,406 @@ mod tests {
         assert_eq!(doc.len(), 0);
     }
 
-    #[test]
-    fn test_dymamic_mode_simple() {
-        let default_doc_mapper: DefaultDocMapper =
-            serde_json::from_str(r#"{ "mode": "dynamic" }"#).unwrap();
+    fn test_doc_mapper_get_all_aux(
```
fulmicoton, line 1585:
```suggestion
    fn test_doc_from_json_test_aux(
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -1514,119 +1582,406 @@ mod tests {
...
+        doc_mapper: &str,
```
fulmicoton, line 1586:
```suggestion
        doc_mapper_json: &str,
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -1514,119 +1582,406 @@ mod tests {
...
+        field: &str,
+        document: &str,
```
fulmicoton, line 1588:
```suggestion
        document_json: &str,
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -543,18 +781,57 @@ fn build_mapping_tree_from_entries<'a>(
     entries: &'a [FieldMappingEntry],
     field_path: &mut Vec<&'a str>,
     schema: &mut SchemaBuilder,
-) -> anyhow::Result<MappingNode> {
+) -> anyhow::Result<(MappingNode, Vec<Field>)> {
```
fulmicoton, line 784: the function name nor the type of the result suggests what this `Vec<Field>` is about.
You need to refactor, rename,  or at least comment.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -543,18 +781,57 @@ fn build_mapping_tree_from_entries<'a>(
...
+            concatenate_dynamic_fields.append(&mut dynamic_fields);
         }
-        let child_tree = build_mapping_from_field_type(&entry.mapping_type, field_path, schema)?;
-        field_path.pop();
-        mapping_node.insert(&entry.name, child_tree);
     }
-    Ok(mapping_node)
+    for concatenate_field_entry in concatenate_fields {
+        let FieldMappingType::Concatenate(options) = &concatenate_field_entry.mapping_type else {
+            continue; // unreachable
```
fulmicoton, line 805: I'm ok with panicking if it is truely unreachable.

fulmicoton (dismissed):
see minor comments

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
trinity-1686a: I'm not sure there is a clean way to do that. The trait is implemented for `Into<TantivyValue>`, so I can't call `from` (from implies into, but not the other way around). fwiw it was already a Self::into before, it just moved a bit

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
trinity-1686a: right now, it's an error at index creation time. In the future, I hope it may become possible to implement. I've added a commented-out test

fmassot (approved):

Issue #4613, opened 2024-02-20
Garbage collect ingested files in the metastore in Lambda indexer


fmassot:
Each time we invoke the lambda indexer to ingest a file, it will index it and store a new checkpoint with the file's name.

The metastore file will become bigger and bigger. We can remove the checkpoints after indexing and leave the responsibility to the caller to not index the same file twice. It could also be a parameter that we can pass to the indexer.

rdettai:
a few points:
- I think deleting the checkpoint right after the indexing is a bit dangerous, as it is not uncommon that triggering events get duplicated and without the checkpoint safe guard, the file would be ingested twice. I propose using a background cleanup process that removes the checkpoint after X hours.
- we identified this issue when doing file based ingestion in Lambda, but in the end it might happen with file based ingestion scripts executed on regular indexers as well. I wonder whether this cleanup code should be part of the Lambda package or rather some general purpose package like the janitor.

rdettai:
> I propose using a background cleanup process that removes the partition checkpoints after X hours

After taking a deeper look, the timestamp info is not available for checkpoints, so doing a time based pruning is not really practical. Unfortunately, pruning by keeping only the last N partitions is not much better, because the partition entries are sorted in lexicographical order and not by insertion. I suggest changing the partition checkpoint order to insertion order (backing in memory data structure to be decided).

rdettai:
We agreed with @fmassot to use the source id to store the timestamp when indexing a file in Lambda and perform time based pruning at the source level instead of the partition level. This has the benefit of containing the workaround to the Lambda package for now.

rdettai:
After running the solution proposed above, we realized the merges were not performed anymore. Indeed, merges are only performed within a source: https://github.com/quickwit-oss/quickwit/blob/f6065a4f4c96d17189c9fac7935a04da8076a309/quickwit/quickwit-indexing/src/actors/merge_planner.rs#L299-L304.

We agreed with @fmassot to go for a more drastic solution where the entire checkpoint list is purged after a limit is reached (default QW_LAMBDA_MAX_CHECKPOINTS=100). In very rare cases, if a duplicate file notification arrives right after the purge, the associated data might be ingested twice, but this should be extremely rare.

PR #4777, opened 2024-03-21
Lambda-prune-checkpoints

rdettai:
### Description

Avoid accumulating file sources when running Lambda indexer

### How was this PR tested?


rdettai (commented):
A bug remains, only one source is created:
```json
    "checkpoint": {
      "ingest-lambda-source-1711043830": {
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711043826.gz": "00000000000016044634",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711044126.gz": "00000000000016043012",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711044426.gz": "00000000000016038941",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711044726.gz": "00000000000016041053",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711045026.gz": "00000000000016041903",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711045326.gz": "00000000000016044080",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711045626.gz": "00000000000016041526",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711045925.gz": "00000000000016042481",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711046225.gz": "00000000000016044227",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711046526.gz": "00000000000016042532",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711046825.gz": "00000000000016043210",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711047126.gz": "00000000000016042689",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711047426.gz": "00000000000016042582",
        "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711047726.gz": "00000000000016042450"
      }
    },
    "create_timestamp": 1711039331,
    "sources": [
      {
        "version": "0.8",
        "source_id": "ingest-lambda-source-1711043830",
        "num_pipelines": 1,
        "enabled": true,
        "source_type": "file",
        "params": {
          "filepath": "s3://mockdatastack-sourcemockdata26422bfc-mpua0jb4rrh1/mock-sales/1711043826.gz"
        },
        "input_format": "json"
      }
    ]
  },
```

quickwit/quickwit-lambda/src/indexer/ingest/helpers.rs
```diff
@@ -131,66 +137,116 @@ pub(super) async fn send_telemetry() {
...
+                let filepath = file_params
+                    .filepath
+                    .as_ref()
+                    .context("Found Lambda file source without filepath")?;
+                Ok((src_id, filepath))
+            } else {
+                bail!("Found Lambda source with non-file source type");
+            }
+        })
+        .collect::<anyhow::Result<Vec<_>>>()?;
```
rdettai, line 170: The filter that matches the source file with the input path went missing here somehow!

quickwit/quickwit-lambda/src/indexer/ingest/helpers.rs
```diff
@@ -131,66 +137,71 @@ pub(super) async fn send_telemetry() {
...
-        let is_not_found = e
-            .downcast_ref()
-            .is_some_and(|meta_error| matches!(meta_error, MetastoreError::NotFound(_)));
-        if !is_not_found {
-            bail!(e);
+) -> anyhow::Result<IndexMetadata> {
+    let metadata_result = metastore
+        .index_metadata(IndexMetadataRequest::for_index_id(INDEX_ID.clone()))
+        .await;
+    let metadata = match metadata_result {
```
trinity-1686a, line 168: this appears to check less than `run_index_checklist` (no check on source and storage). I think skipping storage is fine because it's likely (forced?) to be the same storage as the metastore. Is the source bucket necessarily the same as the store/metastore bucket?

rdettai: You are right, we are softening the checks. `run_index_checklist` was meant to be used when indexing adhoc from the CLI. In that usecase, it's valuable to have a clear error message on the client side at the cost of an extra roundtrip to the metastore backend. When running on Lambda, this roundtrip uses up a expensive milliseconds and the check result appear in the logs of the indexing Lambda anyway, so they are not that helpful anymore.

trinity-1686a (approved):
i'd rather we always keep the last X checkpoints so as to make sure we are resilient to redundant notification, but i'm not sure how that can be done (embed a timestamp in the partition id, after a # maybe?). Anyway that's definitely an improvement

rdettai:
> i'd rather we always keep the last X checkpoints

Yes, couldn't agree more. I logged in https://github.com/quickwit-oss/quickwit/issues/4613 a few things I tried to achieve that and why they failed. We could definitively come up with a solution but it would require a more massive rewrite on the metastore or the source.

PR #4798, opened 2024-03-26
top n collector specialization

PSeitz:
- **move segment top k collector**
- **specialized top n collection**

top_n_specialization

### DIFF
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -151,6 +151,9 @@ pub(crate) enum SortingFieldExtractorComponent {
 }
 
 impl SortingFieldExtractorComponent {
+    pub fn is_score(&self) -> bool {
+        matches!(self, SortingFieldExtractorComponent::Score)
+    }
     pub fn is_fast_field(&self) -> bool {
         matches!(self, SortingFieldExtractorComponent::FastField { .. })
     }
@@ -388,6 +391,14 @@ pub(crate) struct SortingFieldExtractorPair {
 }
 
 impl SortingFieldExtractorPair {
+    pub fn is_score(&self) -> bool {
+        self.first.is_score()
+            || self
+                .second
+                .as_ref()
+                .map(|second| second.is_score())
+                .unwrap_or(false)
+    }
     /// Returns the list of sort values for the given element
     ///
     /// See also [`SortingFieldExtractorComponent::extract_typed_sort_values_block`] for more
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -39,6 +39,7 @@ use tantivy::{DateTime, DocId, Score, SegmentOrdinal, SegmentReader, TantivyErro
 
 use crate::filters::{create_timestamp_filter_builder, TimestampFilter, TimestampFilterBuilder};
 use crate::find_trace_ids_collector::{FindTraceIdsCollector, FindTraceIdsSegmentCollector, Span};
+use crate::top_k_collector::{QuickwitSegmentTopKCollector, SearchAfterSegment};
 use crate::GlobalDocAddress;
 
 #[derive(Clone, Debug)]
@@ -129,7 +130,7 @@ impl SortByComponent {
 }
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
-enum SortFieldType {
+pub(crate) enum SortFieldType {
     U64,
     I64,
     F64,
@@ -139,7 +140,7 @@ enum SortFieldType {
 
 /// The `SortingFieldExtractor` is used to extract a score, which can either be a true score,
 /// a value from a fast field, or nothing (sort by DocId).
-enum SortingFieldExtractorComponent {
+pub(crate) enum SortingFieldExtractorComponent {
     /// If undefined, we simply sort by DocIds.
     DocId,
     FastField {
@@ -150,7 +151,7 @@ enum SortingFieldExtractorComponent {
 }
 
 impl SortingFieldExtractorComponent {
-    fn is_fast_field(&self) -> bool {
+    pub fn is_fast_field(&self) -> bool {
         matches!(self, SortingFieldExtractorComponent::FastField { .. })
     }
     /// Loads the fast field values for the given doc_ids in its u64 representation. The returned
@@ -187,7 +188,7 @@ impl SortingFieldExtractorComponent {
     ///
     /// This is used to convert `search_after` sort value to a u64 representation that will respect
     /// the same order as the `SortValue` representation.
-    fn convert_u64_ff_val_to_sort_value(&self, sort_value: u64) -> SortValue {
+    pub fn convert_u64_ff_val_to_sort_value(&self, sort_value: u64) -> SortValue {
         let map_fast_field_to_value = |fast_field_value, field_type| match field_type {
             SortFieldType::U64 => SortValue::U64(fast_field_value),
             SortFieldType::I64 => SortValue::I64(i64::from_u64(fast_field_value)),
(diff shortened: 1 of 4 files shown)


fulmicoton:
What does the bench looks like?

PSeitz:
> What does the bench looks like?

So far really good, 2x on the last 6 hours query and up to 5x. I added the results above. The impact on the date histogram doesn't make sense though.

fulmicoton:
I expected a big improvment but that's amazing :) Good job!

quickwit/quickwit-search/src/collector.rs
```diff
@@ -1088,32 +880,20 @@ impl Collector for QuickwitCollector {
...
-                .compare(&self.split_id, &search_after.split_id)
-                .then_with(|| order1.compare(&segment_ord, &search_after.segment_ord)),
-            // This value isn't actually used.
-            _ => Ordering::Equal,
-        };
-        // Convert search_after into fast field u64
-        let search_after =
-            SearchAfterSegment::new(self.search_after.clone(), order1, order2, &score_extractor);
 
         let segment_top_k_collector = if leaf_max_hits == 0 {
```
fulmicoton, line 884: Arguable: now that you have dynamic dispatch, the code might be a tiny bit cleaner if we removed the `Option<Box<dyn QuickwitSegmentTopKCollector>>` and just made it a `Box<dyn QuickwitSegmentTopKCollector>`.

The logic `if leaf_max_hits == 0` would then be moved in `specialized_top_k_segment_collector`
and we would return a `DoNothingSegmentTopKCollector`.

(I wouldn't expect any performance difference, one way or another. It is just to remove the Option)

quickwit/quickwit-search/src/top_k_collector.rs
```diff
@@ -0,0 +1,860 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use std::cmp::{Ordering, Reverse};
```
fulmicoton, line 20: I like that you isolated the complexity in a different file.

quickwit/quickwit-search/src/top_k_collector.rs
```diff
@@ -0,0 +1,860 @@
...
+    SortingFieldExtractorComponent, SortingFieldExtractorPair,
+};
+
+pub trait QuickwitSegmentTopKCollector {
+    fn collect_top_k_block(&mut self, docs: &[DocId]);
+    fn collect_top_k(&mut self, doc_id: DocId, score: Score);
+    fn get_top_k(&self) -> Vec<PartialHit>;
+}
+trait IntoOptionU64 {
+    fn is_unit_type() -> bool;
```
fulmicoton, line 39: nitpick. A default implementation might make sense here.
```suggestion
    fn is_unit_type() -> bool {
        false
    }
```

quickwit/quickwit-search/src/top_k_collector.rs
```diff
@@ -0,0 +1,860 @@
...
+    fn into_option_u64(self) -> Option<u64> {
+        None
+    }
+    #[inline]
+    fn from_option_u64(_: Option<u64>) -> Self {}
+}
+
+/// Like Reverse<Option<u64>>, but sorts None as the smallest value.
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+struct ReverseOptionU64(Option<u64>);
```
fulmicoton, line 89: Can we move that above?

fulmicoton, line 89: Could it have been `Option<Reverse<u64>>` ?

quickwit/quickwit-search/src/top_k_collector.rs
```diff
@@ -0,0 +1,860 @@
...
+///
+/// The buffer is truncated to the top_n elements when it reaches the capacity of the Vec.
+/// That means capacity has special meaning and should be carried over when cloning or serializing.
+///
+/// For TopK == 0, it will be relative expensive.
+pub struct TopKComputer<D> {
+    /// Reverses sort order to get top-semantics instead of bottom-semantics
+    buffer: Vec<Reverse<D>>,
+    top_n: usize,
+    pub(crate) threshold: Option<D>,
```
fulmicoton, line 325: Could we remove the Option in threshold?
(and initialize it to Max or Min depending on what we do)

```suggestion
    pub(crate) threshold: Option<D>,
```

quickwit/quickwit-search/src/top_k_collector.rs
```diff
@@ -0,0 +1,860 @@
...
+        TopKComputer {
+            buffer: buffer_clone,
+            top_n: self.top_n,
+            threshold: self.threshold.clone(),
+        }
+    }
+}
+
+impl<D> TopKComputer<D>
+where D: Ord + Clone + Debug
```
fulmicoton, line 343:
```suggestion
where D: Ord + Clone + Copy + Debug
```

quickwit/quickwit-search/src/top_k_collector.rs
```diff
@@ -0,0 +1,860 @@
...
+            top_n,
+            threshold: None,
+        }
+    }
+
+    /// Push a new document to the top n.
+    /// If the document is below the current threshold, it will be ignored.
+    #[inline]
+    pub fn push(&mut self, doc: D) {
+        if let Some(last_median) = self.threshold.clone() {
```
fulmicoton, line 361:
```suggestion
        if let Some(last_median) = self.threshold {
```

quickwit/quickwit-search/src/top_k_collector.rs
fulmicoton: also do we need the Debug boundary?

quickwit/quickwit-search/src/top_k_collector.rs
```diff
@@ -0,0 +1,860 @@
...
+        V2: Copy + PartialEq + PartialOrd + Ord + IntoOptionU64 + Debug,
+    > SpecSortingFieldExtractor<V1, V2>
+{
+    fn new(
+        first: SortingFieldExtractorComponent,
+        second: Option<SortingFieldExtractorComponent>,
+    ) -> Self {
+        Self {
+            _phantom: PhantomData,
+            sort_values1: Box::new([None; COLLECT_BLOCK_BUFFER_LEN]),
```
fulmicoton, line 438: Nitpick.

Unfortunately this still allocates on the stack first. 
I tend to prefer `vec![None; COLLECT_BLOCK_BUFFER_LEN].into_boxed_slice()`?


In this very case this is only `9*64` bytes so it is probably ok.

quickwit/quickwit-search/src/top_k_collector.rs
```diff
@@ -0,0 +1,860 @@
...
+    }
+
+    #[inline(never)]
+    fn truncate_top_n(&mut self) -> D {
+        // Use select_nth_unstable to find the top nth score
+        let (_, median_el, _) = self.buffer.select_nth_unstable(self.top_n);
+
+        let median_score = median_el.clone();
+        // Remove all elements below the top_n
+        self.buffer.truncate(self.top_n);
```
fulmicoton, line 391: do you know if truncate_top_n or `self.buffer.truncate` takes a lot of perf?

fulmicoton (approved):

quickwit/quickwit-search/src/top_k_collector.rs
PSeitz: yes that's better

quickwit/quickwit-search/src/top_k_collector.rs
PSeitz: `self.buffer.truncate` is just setting the len and therefore very cheap, but `self.buffer.select_nth_unstable(self.top_n)` takes the majority of time on some queries, especially if the threshold is not effective

quickwit/quickwit-search/src/top_k_collector.rs
PSeitz: Yes, seems to be slightly faster

quickwit/quickwit-search/src/top_k_collector.rs
PSeitz: I thought this has been resolved, but it hasn't

quickwit/quickwit-search/src/collector.rs
PSeitz: I think it's a little bit more readable to have an option instead of a `DoNothingCollector` behind the trait

PR #4801, opened 2024-03-27
run migrations in a transaction

### LINKED ISSUE #4746: Postgres migrations are not run in a transaction
Notice how we begin a transaction but how the migrations are run with the pool. The pool will run migrations with a separate connection and run each DDL statement as a separate transaction.

```rust
let tx = pool.begin().await?;
let migrate_result = MIGRATOR.run(pool).await;
```

The obvious fix does not compile:
```rust
let mut tx = pool.begin().await?;
let migrate_result = MIGRATOR.run(&mut tx).await;
```

trinity-1686a:
### Description

fix #4746 
we get an `implementation of sqlx::Acquire is not general enough` error. To get around that, we use the hidden `run_direct`, which is made specifically to go arround that error. See also https://github.com/launchbadge/sqlx/issues/2016

### How was this PR tested?

created 2 dummy migrations, one valid and one invalid. Verify the 1st one doesn't get committed because the 2nd failed.


guilload (approved):
Nice!

PR #4804, opened 2024-03-28
add debug endpoint to get mrecordlog summary

### LINKED ISSUE #4800: Add a debugging endpoint with the record log Summary
In https://github.com/quickwit-oss/mrecordlog/commit/a133272cadc5aa45e55dc42e7d09c7960eaddb93
a summary method was added to mrecordlog.

We would like a debugging rest endpoint to expose it. This is for debugging and likely to be removed in the future, no need to sweat it too much.

trinity-1686a:
### Description

update mrecordlog
move debugging endpoints to their own prefix, and add debug endpoint to get mrecordlog state summary
fix #4800 

### How was this PR tested?

tested manually on an indexer and a non-indexer node

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
@@ -45,6 +45,7 @@ use quickwit_proto::ingest::ingester::{
     AckReplicationMessage, CloseShardsRequest, CloseShardsResponse, DecommissionRequest,
     DecommissionResponse, FetchMessage, IngesterService, IngesterServiceClient,
     IngesterServiceStream, IngesterStatus, InitShardsRequest, InitShardsResponse,
+    MrecordlogQueueSummary, MrecordlogSummaryRequest, MrecordlogSummaryResponse,
     ObservationMessage, OpenFetchStreamRequest, OpenObservationStreamRequest,
     OpenReplicationStreamRequest, OpenReplicationStreamResponse, PersistFailure,
     PersistFailureReason, PersistRequest, PersistResponse, PersistSuccess, ReplicateFailureReason,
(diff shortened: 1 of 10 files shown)


trinity-1686a:
blocked on https://github.com/quickwit-oss/mrecordlog/pull/62

quickwit/quickwit-proto/protos/quickwit/ingester.proto
```diff
@@ -268,6 +271,20 @@ message DecommissionRequest {
 message DecommissionResponse {
 }
 
+message MrecordlogSummaryRequest {
```
fulmicoton, line 274:
```suggestion
message MRecordLogSummaryRequest {
```

if clippy does not whine too much...

trinity-1686a: the messages can be renamed (2 upper cases is fine iirc). The rpc could be, but this leads to (imo ugly) `async fn m_recordlog_summary(..)`, which we could do, but could also not.

quickwit/quickwit-proto/protos/quickwit/ingester.proto
```diff
@@ -268,6 +271,20 @@ message DecommissionRequest {
...
+}
+
+message MrecordlogSummaryResponse {
+  repeated MrecordlogQueueSummary queues = 1;
+}
+
+message MrecordlogQueueSummary {
+  string id = 1;
```
fulmicoton, line 282: I don't think we need grpc for this. the data is on the same node.

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -961,6 +961,48 @@ impl Ingester {
 
         Ok(DecommissionResponse {})
     }
+
+    pub async fn mrecordlog_summary(&mut self) -> IngestV2Result<MRecordlogSummary> {
+        let rw_mrecordlog = self.state.mrecordlog();
+        // this is a debug api endpoint, with_lock_metrics! doesn't seem necessary
+        let maybe_mrecordlog = rw_mrecordlog.read().await;
+
+        let summary = match &*maybe_mrecordlog {
```
fulmicoton, line 970: Nitpick: let... else... are  nice for those.


```suggestion
        let Some(summary) = maybe_mrecordlog else {
               return Err(IngestV2Error::Internal(
                    "mrecordlog isn't initialized".to_string(),
                ))
        };
```

You could also use `.ok_or_else(|| {})?`

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -961,6 +961,48 @@ impl Ingester {
...
+            })
+            .collect();
+
+        Ok(MRecordlogSummary { queues })
+    }
+}
+
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct MRecordlogSummary {
+    pub queues: Vec<MRecordlogQueueSummary>,
```
fulmicoton, line 997: Nitpick. You did not like the format of the QuerySummary in record log? I think we could have modified instead of having 2.

quickwit/quickwit-serve/src/debugging_api.rs
```diff
@@ -17,41 +17,102 @@
...
+        ),
+    )
+}
+
 #[utoipa::path(
     get,
-    tag = "Get debug information for node",
-    path = "/",
+    tag = "Get debug informations about the control plane from the node viewpoint",
+    path = "/shard",
```
fulmicoton, line 54: I'm not sure about the path `/shard`.

Note that your description one line above suggests that the name is not great.
Could we call it `indexing_plan`  or `control_plane` or something like that?

fulmicoton (changes requested):
See  comments. (names mostly.)

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
trinity-1686a: for objects inside the language hashmap are better, but for APIs, i tend to prefer lists of key-values. I don't think it really matter, i'm going to use what we have in mrecordlog directly

fulmicoton (approved):

Issue #4734, opened 2024-03-14
More compact logging on Lambda


rdettai:
**Describe the bug**
When running the Lambda indexer, the cloudwatch logging is actually more expensive than the execution time. 

**Steps to reproduce (if applicable)**
Run the mock data stack for a few days. Inspect the AWS cost explorer. You should get something like this:

![image](https://github.com/quickwit-oss/quickwit/assets/7913347/7f0f02e8-cbd6-4a96-9399-09d81ca36a45)

**Expected behavior**
Cloudwatch logs should represent at most 20% of the Lambda compute cost.

PR #4805, opened 2024-03-28
Lambda search endpoints using a warp adapter

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/4614

### How was this PR tested?

Endpoint integration tests can be performed with:
```
# deploy the stack
export CDK_ACCOUNT=123456789
export CDK_REGION=us-east-1
export SEARCHER_API_KEY=my-at-least-20-char-long-key
export QW_LAMBDA_BUILD=1
make bootstrap
make deploy-mock-data
# ...wait a few minutes...
make test-mock-data-endpoints
```


rdettai:
started addressing https://github.com/quickwit-oss/quickwit/issues/4734 by making default fmt logging more compact

quickwit/quickwit-lambda/src/searcher/warp_lambda.rs
```diff
@@ -0,0 +1,173 @@
...
+where
+    S: Service<WarpRequest, Response = WarpResponse, Error = Infallible> + 'a,
+    S::Future: Send + 'a,
+{
+    type Response = Response<LambdaBody>;
+    type Error = LambdaError;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send + 'a>>;
+
+    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        core::task::Poll::Ready(Ok(()))
```
trinity-1686a, line 127: why not call warp_service's own poll_ready?
```suggestion
    fn poll_ready(&mut self, ctx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
        self.warp_service.poll_ready(ctx)
            .map_err(|err| match err {})
```

trinity-1686a (commented):
why is that PR configured to merge into ` lambda-prune-checkpoints`?

So you ended up always keeping the static context instead of supporting generating a new `SearchService` for each request?

quickwit/quickwit-lambda/src/searcher/warp_lambda.rs
rdettai: This code comes from the forked https://github.com/aslamplr/warp_lambda. I think the rational is that when running on Lambda, we don't want the backpressure mechanism offered by `poll_ready` on the service trait. We have only one caller at a time. But I agree that we should make sure that the warp_service is ready, otherwise as per the `call` [method's spec](https://docs.rs/tower/latest/tower/trait.Service.html#tymethod.call) it's allowed to panic

trinity-1686a (approved):

PR #4812, opened 2024-03-29
Fix control plane readiness

guilload:
### Description
Per title.

### How was this PR tested?
- Added unit test
- Verified logs output `2024-03-29T00:38:43.814Z  INFO quickwit_serve: control plane is ready`

### DIFF
--- a/quickwit/quickwit-serve/src/lib.rs
+++ b/quickwit/quickwit-serve/src/lib.rs
@@ -64,8 +64,7 @@ use itertools::Itertools;
 use once_cell::sync::Lazy;
 use quickwit_actors::{ActorExitStatus, Mailbox, SpawnContext, Universe};
 use quickwit_cluster::{
-    start_cluster_service, Cluster, ClusterChange, ClusterChangeStream, ClusterMember,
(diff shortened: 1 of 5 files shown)


guilload:
PTAL @fulmicoton

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -967,7 +970,7 @@ async fn setup_control_plane(
         default_index_root_uri,
         replication_factor,
     };
-    let (control_plane_mailbox, _control_plane_handle) = ControlPlane::spawn(
+    let (cp_mailbox, _cp_handle, mut readiness_rx) = ControlPlane::spawn(
```
fulmicoton, line 973: I personally preferred the longer name...

Now we have `cp` or `control_plane_` depending on the file in we are in in our code base.

fulmicoton (approved):

quickwit/quickwit-serve/src/lib.rs
guilload: fixed

PR #4819, opened 2024-03-30
remove unused deps

trinity-1686a:
### Description

remove as much unused dependencies as possible.
i hoped it would have some impact on compilation time. If it has, i can't say i'm able to  measure it, but the work is done, so i might as well push it

### How was this PR tested?

cargo check individually each modified crate, with `--tests` and `--all-features` where that felled relevant


guilload (approved):
Awesome! Should we set up cargo machete in our CI job?

trinity-1686a:
given the limited of impact from these changes, i don't think we *need* it, but it wouldn't hurt either.

PR #4831, opened 2024-04-03
support negative match with multiple clauses

trinity-1686a:
### Description

change so that `-field:val AND -field2:val2` returns the same as `NOT field:val AND NOT field2:val2`.
Previously it would be different because the former is parsed as `-field:val -field2:val2` while latter is parsed as `+(-field:val) +(-field2:val2)`, and we added a `+*` when there is a ***single*** must_not in a bool query, but wouldn't if there was 2 of them

### How was this PR tested?

tested manually. Added regression test


quickwit/quickwit-query/src/query_ast/tantivy_query_ast.rs
```diff
@@ -217,19 +217,27 @@ impl TantivyBoolQuery {
...
+            }
+            self.must.push(TantivyQueryAst::match_all());
+        } else {
+            let num_children =
+                self.must.len() + self.should.len() + self.must_not.len() + self.filter.len();
+            if num_children == 1 {
+                if let Some(ast) = self.must.pop().or(self.should.pop()) {
+                    return ast;
                 }
-                self.must.push(TantivyQueryAst::match_all());
```
fulmicoton, line 227: please add a unit test.

fulmicoton (approved):

PR #4833, opened 2024-04-03
Rebalance shards

guilload:
### Description
Rebalance shards periodically.

### How was this PR tested?
- Added a bunch of unit tests.
- Tested locally (1,2,3,4 nodes)
- ~~Tested on Cicada~~


guilload:
I need to write more unit tests for this change, but it is ready for review.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -724,9 +978,15 @@ fn summarize_shard_ids(shard_ids: &[ShardIds]) -> Vec<&str> {
         .collect()
 }
 
-/// Finds the shard with the highest ingestion rate on the ingester with the least number of open
-/// shards. If multiple shards have the same ingestion rate, the shard with the highest shard ID is
-/// chosen.
+#[derive(Debug)]
```
fulmicoton, line 981: Can you add doc here about what this message is about?

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -106,6 +121,7 @@ impl IngestController {
             metastore,
             ingester_pool,
             replication_factor,
+            rebalance_permits: Arc::new(Semaphore::new(1)),
```
fulmicoton, line 124: nitpick:

I suspect a Mutex and OwnedMutexGuard would have better captured your intent.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -83,6 +97,7 @@ pub struct IngestController {
     ingester_pool: IngesterPool,
     metastore: MetastoreServiceClient,
     replication_factor: usize,
+    rebalance_permits: Arc<Semaphore>,
```
fulmicoton, line 100:
```suggestion
    // We use this semaphore, with a number of capacity of 1, to ensure we do not start a rebalance while there is already an ongoing rebalance.
    rebalance_permits: Arc<Semaphore>,
```

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -706,9 +780,189 @@ impl IngestController {
...
 
-    pub fn rebalance_shards(&mut self) {
-        // TODO: As of now, it is only used for unit testing.
+    pub(crate) async fn rebalance_shards(
+        &mut self,
+        model: &mut ControlPlaneModel,
+        mailbox: &Mailbox<ControlPlane>,
+        progress: &Progress,
+    ) {
+        let Ok(_rebalance_permit) = self.rebalance_permits.clone().try_acquire_owned() else {
```
fulmicoton, line 789:
```suggestion
        let Ok(rebalance_permit) = self.rebalance_permits.clone().try_acquire_owned() else {
```

`_blabla` is  for stuff that is unused. Here it is not unused... Of course, rust 1.77 is a pain and forces you to have put _ on unused struct stuff but still.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -706,9 +780,189 @@ impl IngestController {
...
+                        error!("close shards request timed out");
+                    }
+                }
+            }
+            if closed_shards.is_empty() {
+                return;
+            }
+            let callback = RebalanceShardsCallback {
+                closed_shards,
+                _rebalance_permit,
```
fulmicoton, line 961:
```suggestion
                _rebalance_permit: rebalance_permit,
```

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -706,9 +780,189 @@ impl IngestController {
         }
     }
 
-    pub fn rebalance_shards(&mut self) {
-        // TODO: As of now, it is only used for unit testing.
+    pub(crate) async fn rebalance_shards(
```
fulmicoton, line 783: comment on this method please.

This is not that trivial as it seem for an outsider I think
rebalance shards could mean something about indexer rebalancing.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -706,9 +780,189 @@ impl IngestController {
...
+            if shard.is_open() {
+                num_open_shards += 1;
+
+                per_leader_open_shards
+                    .entry(&shard.leader_id)
+                    .or_default()
+                    .push(shard);
+            }
+        }
+        let num_open_shards_per_leader_target = num_open_shards / num_ingesters;
```
fulmicoton, line 812: just to be on the safe side, `ceil_div` would be nicer.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -706,9 +780,189 @@ impl IngestController {
...
+                    Duration::from_secs(3),
+                    ingester.close_shards(shards_to_close_request),
+                )
+                .await
+            };
+            close_shards_futures.push(close_shards_future);
+        }
+        let mailbox = mailbox.clone();
+        let close_shards_future = async move {
+            // We wait for a few seconds before closing the shards to give the ingesters some time
```
fulmicoton, line 937: good comment

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -706,9 +780,189 @@ impl IngestController {
...
+                tokio::time::timeout(
+                    Duration::from_secs(3),
+                    ingester.close_shards(shards_to_close_request),
+                )
+                .await
+            };
+            close_shards_futures.push(close_shards_future);
+        }
+        let mailbox = mailbox.clone();
+        let close_shards_future = async move {
```
fulmicoton, line 936: can we extract that in a different function?

one benefit is that it would make some important detail of the implementation clearer: 
- the shard opening happens sync to the control plane actor, but the closing happens on a separate task.
- `_rebalance_permit` is dropped after closed is done.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -706,9 +780,189 @@ impl IngestController {
...
+            // We wait for a few seconds before closing the shards to give the ingesters some time
+            // to learn about the ones we just opened via gossip.
+            tokio::time::sleep(Duration::from_secs(10)).await;
+
+            let mut closed_shards = Vec::new();
+
+            while let Some(close_shards_result) = close_shards_futures.next().await {
```
fulmicoton, line 943: You have an individual timeout on  the different close shard request, but no differentiated logging.

We should either have specific loggging information, or make the timeout global for the set of futures.

The benefit of the global timeout, is that you can move it closed to what's happening in the task, and help proofreading.

A reader wants to be able to answer questions like, "is it possible to hold the rebalance permit forever? Is it capped in time?"

fulmicoton (approved):

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -862,20 +859,46 @@ impl Handler<IndexerLeft> for ControlPlane {
...
+impl Handler<RebalanceShardsCallback> for ControlPlane {
+    type Reply = ();
+
+    async fn handle(
+        &mut self,
+        message: RebalanceShardsCallback,
+        _ctx: &ActorContext<Self>,
+    ) -> Result<Self::Reply, ActorExitStatus> {
+        info!(
+            "closing {} shards after rebalance",
```
fulmicoton, line 887: I think we already have a code path to learn about shards being close:
router will tell use after getting bounced.

I don't think closing proactively like is preventing the other path, so I am not sure this message/handle is useful.

guilload: Routers should not get bounced and ask the control plane for open shard alternatives because they should already know about the newly opened shards via gossip.

The other path that would let us know about those closed shards would be via ingesters broadcasting the status of their shards. However, this is propagated with some delay, which will cause the next rebalance operation to keep moving shards unnecessarily. We could also keep the rebalance lock for a few broadcast intervals, but I find that closing shards the current way is the simplest solution.

PR #4859, opened 2024-04-10
Load shed control plane and metastore

guilload:
### Description
- Implement load shed layer
- Place load shed layer in front of control plane and metastore
- Limit retries to 3 for the standard retry policy instead of 30

Ideally, we would have some per-client retry budget or adaptive throttling logic in the client to limit the number of retries when the control plane or metastore is shedding load.

### How was this PR tested?
- added unit test for `LoadShedLayer`
- test locally and manually by disabling retries on metastore client, setting `max_in_flight_requests` to 1, and adding delay to control plane RPCs.


quickwit/quickwit-common/src/tower/load_shed.rs
```diff
@@ -0,0 +1,160 @@
...
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = LoadShedFuture<S::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        if self.permit_opt.is_none() {
+            if let Ok(permit) = self.permits.clone().try_acquire_owned() {
+                self.permit_opt = Some(permit);
+            } else {
+                return Poll::Ready(Err(S::Error::make_load_shed_error()));
```
fulmicoton, line 68: I am not super familiar with tower services. Are we supposed to attach the waker in `cx` to the acquisition of the semaphore?

fulmicoton: https://github.com/tower-rs/tower/blob/39adf5c509a1b2141f679654d8317524ca96b58b/tower/src/limit/concurrency/service.rs#L65

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -321,14 +322,19 @@ async fn start_control_plane_if_needed(
             replication_factor,
         )
         .await?;
+        // These layers apply to all the RPCs of the control plane.
+        let shared_layers = ServiceBuilder::new()
+            .layer(CP_GRPC_SERVER_METRICS_LAYER.clone())
+            .layer(LoadShedLayer::new(100))
+            .into_inner();
```
fulmicoton, line 329: should/can we put those after in order to avoid spawning tasks if the load shed kicks in.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -321,14 +322,19 @@ async fn start_control_plane_if_needed(
...
         let control_plane_server = ControlPlaneServiceClient::tower()
-            .stack_create_index_layer(quickwit_common::tower::OneTaskPerCallLayer)
-            .stack_delete_index_layer(quickwit_common::tower::OneTaskPerCallLayer)
-            .stack_add_source_layer(quickwit_common::tower::OneTaskPerCallLayer)
-            .stack_toggle_source_layer(quickwit_common::tower::OneTaskPerCallLayer)
-            .stack_delete_source_layer(quickwit_common::tower::OneTaskPerCallLayer)
-            .stack_get_or_create_open_shards_layer(quickwit_common::tower::OneTaskPerCallLayer)
-            .stack_layer(CP_GRPC_SERVER_METRICS_LAYER.clone())
+            .stack_layer(shared_layers)
```
fulmicoton, line 331: same comment as for the control plane.

fulmicoton (approved):

quickwit/quickwit-serve/src/lib.rs
guilload: So [this](https://docs.rs/tower/latest/tower/builder/struct.ServiceBuilder.html#order) is the layering order for `ServiceBuilder`.

The layering order for code-generated `*TowerLayerStack` is the same... Since I couldn't remember it, I wrote a unit test in `codegen-example` (see `test_hello_codegen_tower_stack_layer_ordering`).

So in the end we want:
```rust
.layer(MetricsLayer)
.layer(LoadShedLayer)
.layer(LoadShedLayer)
.layer(OneTaskPerCallLayer)
```
which is what I did here and may have gotten wrong in the past.

PR #4861, opened 2024-04-11
store document length as fast field

### LINKED ISSUE #4852: Add document length feature for billing
Several multi-tenant user of quickwit need a way to measure the usage of their tenants.

We need  a sum of ingested bytes, grouped by a given field for a range of time.
Ideally it should be possible to make the range of time hourly.

# Metrics

We want to handle 1000 partitions, 100k tenants.
~100 tenants per partition.

# Solution

Add a property `document_length: true` in the docmapper.
When enabled, the schema would have an additional u64 meta field called `_doc_length`.
That field would receive the json payload length. It would be a fast field, non stored, not indexed.

It will require about 2 bytes per document.

This feature should be undocumented at the moment.

# Other considered solutions

- Simply have the indexer log the aggregate metrics and have another process do the billing.
- Add the aggregated metrics to splits.

trinity-1686a:
### Description

fix #4852
add a config option to store as fast field the length of the original document.
this field isn't shown when querying, as it is just  a fast field, but is available for aggregation (and sort/range query)

### How was this PR tested?

tested manually, plus added UT and IT


quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -215,7 +224,6 @@ impl TryFrom<DefaultDocMapperBuilder> for DefaultDocMapper {
                     default_search_field_name
                 )
             }
-            let dynamic_field = schema.get_field(DYNAMIC_FIELD_NAME).ok();
```
trinity-1686a, line 218: this value is already present in scope (created line 150/152)

quickwit/quickwit-doc-mapper/src/default_doc_mapper/default_mapper.rs
```diff
@@ -2014,6 +2029,20 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_lenght_field() {
```
fulmicoton, line 2033:
```suggestion
    fn test_length_field() {
```

fulmicoton (approved):

PR #4863, opened 2024-04-12
Issue/4009 shard locality

fulmicoton:
Added support for shard locality.


quickwit/quickwit-control-plane/src/model/shard_table.rs
```diff
@@ -111,6 +111,31 @@ impl ShardTableEntry {
     }
 }
 
+#[derive(Default)]
```
guilload, line 114: It's confusing that we are tracking the locations of shards on ingesters, i.e. where the data is ingested vs. indexers i.e. where the data is indexed, yet use the variables and names `indexer_id` or `indexer`.

quickwit/quickwit-control-plane/src/model/shard_table.rs
```diff
@@ -119,6 +144,7 @@ impl ShardTableEntry {
 pub(crate) struct ShardTable {
     table_entries: FnvHashMap<SourceUid, ShardTableEntry>,
     ingester_shards: FnvHashMap<NodeId, FnvHashMap<SourceUid, BTreeSet<ShardId>>>,
+    // This map is only used at the moment to ensure that shard_ids are unique.
```
guilload, line 147: ?

quickwit/quickwit-control-plane/src/model/shard_table.rs
```diff
@@ -146,6 +172,20 @@ fn remove_shard_from_ingesters_internal(
 }
 
 impl ShardTable {
+    /// Returns a ShardLocations object that maps each shard to the list of indexers hosting it.
+    /// All shards are considered regardless of their state (including unavailable).
+    pub fn shard_locations(&self) -> ShardLocations {
```
guilload, line 177: I think it's more natural to create the shard locations from iterating over the shards directly rather than iterating over the ingesters. It is because of the lifetime?

```python
for table_entry in self.shard_table {
  for shard in table_entry {
       shard_locations.add_location(shard.shard_id, shard.leader_id, shard.follower_id);
  }
}

guilload (approved):

quickwit/quickwit-control-plane/src/model/shard_table.rs
fulmicoton: it is tricky because ingesters are always indexers today, and in this code in particulary, we are trying to apply logic where we assign on indexers, depending on whether they host the shard or not in their "indexer role". 

We can rename stuff, we will still in some place pass a indexer_id to something calling the variable `ingester_id`.

I'll rename it here if you think that helps.

quickwit/quickwit-control-plane/src/model/shard_table.rs
fulmicoton: No good reason...
I thought it was more natural that way. 
A minor benefit is that I get reference to node ids maybe.

PR #4868, opened 2024-04-15
Update indexing settings

rdettai:
### Description

- Add the indexing settings to the update APIs
- Split the update APIs to one route per updatable field to avoid side effects (e.g not specifying the retention policy drops it)
- Simplify the CLI to be a very basic client for the APIs that read the input body from files

### How was this PR tested?

Unit tests and validated that swagger works fine.

### DIFF
--- a/docs/reference/rest-api.md
+++ b/docs/reference/rest-api.md
@@ -350,6 +350,7 @@ Calling the update endpoint with the following payload will remove the current r
     }
 }
 ```
+:::
 
 #### Response
 
--- a/docs/reference/rest-api.md
+++ b/docs/reference/rest-api.md
@@ -309,48 +309,60 @@ The response is the index metadata of the created index, and the content type is
 | `sources`          | List of the index sources configurations. | `Array<SourceConfig>` |
 
 
-### Update an index (search settings and retention policy only)
+### Update an index
 
 ```
-PUT api/v1/indexes/<index id>
+PUT api/v1/indexes/<index id>/indexing-settings
+PUT api/v1/indexes/<index id>/search-settings
+PUT api/v1/indexes/<index id>/retention-policy
 ```
 
-Updates the search settings and retention policy of an index. This endpoint follows PUT semantics (not PATCH), which means that all the updatable fields of the index configuration are replaced by the values specified in this request. In particular, omitting an optional field like retention_policy will delete the associated configuration. Unlike the create endpoint, this API only accepts JSON payloads.
+These endpoints follows PUT semantics (not PATCH), which means that all the fields of the updated configuration are replaced by the values specified in the request. Values that are not specified will be reset to their defaults. The API accepts JSON with `content-type: application/json` and YAML `content-type: application/yaml`.
+- The search settings update is automatically picked up by the janitor service on its next state refresh.
+- The retention policy update is automatically picked up when the next query is executed.
+- The indexing settings update is not automatically picked up by the indexer nodes, they need to be manually restarted.
 
-#### PUT payload
+#### PUT payloads
 
-| Variable            | Type               | Description                                                                                                           | Default value                         |
-|---------------------|--------------------|-----------------------------------------------------------------------------------------------------------------------|---------------------------------------|
-| `search_settings`   | `SearchSettings`   | Search settings object as specified in the [index config docs](../configuration/index-config.md#search-settings).     |                                       |
-| `retention`         | `Retention`        | Retention policy object as specified in the [index config docs](../configuration/index-config.md#retention-policy).   |                                       |
+| Endpoint            | Type               | Description                                                                              |
+|---------------------|--------------------|------------------------------------------------------------------------------------------|
+| `/search-settings`  | `SearchSettings`   | See [search settings config docs](../configuration/index-config.md#search-settings).     |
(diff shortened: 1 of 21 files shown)


quickwit/quickwit-control-plane/src/model/mod.rs
```diff
@@ -54,7 +54,7 @@ use tracing::{info, instrument, warn};
 #[derive(Default, Debug)]
 pub(crate) struct ControlPlaneModel {
     index_uid_table: FnvHashMap<IndexId, IndexUid>,
-    index_table: FnvHashMap<IndexUid, IndexMetadata>,
+    index_source_table: FnvHashMap<IndexUid, HashMap<SourceId, SourceConfig>>,
```
rdettai, line 57: - it is more explicit to only store the sources in the state as the rest of the IndexMetadata isn't relevant
- if we kept the entire `IndexMetadata`, we would need to also proxy the update calls through the control plane

quickwit/quickwit-metastore/src/tests/index.rs
```diff
@@ -146,6 +146,7 @@ pub async fn test_metastore_update_index<
             index_uid.clone(),
             &new_search_setting,
             &loop_retention_policy_opt,
+            &index_metadata.index_config.indexing_settings,
```
rdettai, line 149: TODO: also test indexing_settings mutation

quickwit/quickwit-integration-tests/src/tests/index_update_tests.rs
```diff
@@ -126,6 +126,7 @@ async fn test_update_on_multi_nodes_cluster() {
                     default_search_fields: vec!["title".to_string(), "body".to_string()],
                 },
                 retention_policy_opt: None,
+                indexing_settings: index_metadata.index_config.indexing_settings,
```
rdettai, line 129: TODO: also test indexing_settings mutation

rdettai (commented):
A few notes on this early draft

quickwit/quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -90,7 +92,10 @@ pub fn index_management_handlers(
     get_index_metadata_handler(index_service.metastore())
         .or(list_indexes_metadata_handler(index_service.metastore()))
         .or(create_index_handler(index_service.clone(), node_config))
-        .or(update_index_handler(index_service.metastore()))
+        .or(update_search_settings_handler(index_service.metastore()))
+        .or(update_indexing_setting_handler(index_service.metastore()))
+        .or(update_retention_policy_handler(index_service.metastore()))
+        .or(delete_retention_policy_handler(index_service.metastore()))
```
rdettai, line 98: This is not very DRY, but I didn't manage to do any better

quickwit/quickwit-control-plane/src/model/mod.rs
rdettai: Ok I'll revert. Don't you think it would be better to have a smaller struct that only contains the data that is really useful to the control plane? It would make it easier to read what the control plane does.

quickwit/quickwit-proto/protos/quickwit/metastore.proto
```diff
@@ -202,10 +202,16 @@ message CreateIndexResponse {
...
+  RETENTION_POLICY = 2;
+}
+
 message UpdateIndexRequest {
   quickwit.common.IndexUid index_uid = 1;
-  string search_settings_json = 2;
-  optional string retention_policy_json = 3;
+  UpdatedIndexConfig target_config = 2; 
+  optional string config_json = 3;
 }
```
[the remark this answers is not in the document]
rdettai: Option 3: it's a bit clumsy as `Option<RetentionPolicy> = None` already means "Remove the retention policy.

I will switch to 2, I like that it's less verbose.

docs/reference/rest-api.md
```diff
@@ -309,45 +309,58 @@ The response is the index metadata of the created index, and the content type is
...
-PUT api/v1/indexes/<index id>
+PUT api/v1/indexes/<index id>/indexing-settings
+PUT api/v1/indexes/<index id>/search-settings
+PUT api/v1/indexes/<index id>/retention-policy
 ```
 
-Updates the search settings and retention policy of an index. This endpoint follows PUT semantics (not PATCH), which means that all the updatable fields of the index configuration are replaced by the values specified in this request. In particular, omitting an optional field like retention_policy will delete the associated configuration. Unlike the create endpoint, this API only accepts JSON payloads.
+These endpoints follows PUT semantics (not PATCH), which means that all the fields of the updated configuration are replaced by the values specified in the request. Values that are not specified will be reset to their defaults. The API accepts JSON with `content-type: application/json` and YAML `content-type: application/yaml`.
+- The search settings update is automatically picked up by the janitor service on its next state refresh.
+- The retention policy update is automatically picked up when the next query is executed.
```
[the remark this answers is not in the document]
rdettai: good catch, I inverted "search settings" and "retention policy" between these two sentences 😊

fmassot:
what do we do with this PR @rdettai @guilload ?

rdettai:
It's the next thing in my backlog after SQS. It needs a pretty big rebase and quite a few changes we agreed upon.

rdettai:
This is superseeded by https://github.com/quickwit-oss/quickwit/pull/5078

PR #4867, opened 2024-04-12
Generalize debug endpoint

guilload:
### Description
One endpoint to get all the debug/diagnostic information.

### How was this PR tested?
- added some unit tests
- with 3 local nodes:
```bash
curl 'http://localhost:7280/api/developer/debug' | jq -r .[].node_config.node_id | sort
node-1
node-2
node-3
```

PR #4874, opened 2024-04-17
Apply the one task per call layer to the ingester on the local (non-grpc) path as well.

fulmicoton:
# Test

Tested locally. It was very easy to reproduce the bug using a single instance + simian. When killing simian, the task was getting cancelled just like for grpc.

# Implementation detail

In order to add a debug rpc, I suggested trinity to node add this debug rpc in the service to avoid all of the needless serialization boilerplate, and instead started storing the `Ingester` in QuickwitService (which kind of makes sense in my opinion).

Unfortunately, this lead to this bug. I tried to keep the code dry by keep the list of layers in a method.
We can consider some refactoring later, but this should happen after #4867.

### DIFF
--- a/quickwit/quickwit-serve/src/lib.rs
+++ b/quickwit/quickwit-serve/src/lib.rs
@@ -97,7 +97,9 @@ use quickwit_metastore::{
 use quickwit_opentelemetry::otlp::{OtlpGrpcLogsService, OtlpGrpcTracesService};
 use quickwit_proto::control_plane::ControlPlaneServiceClient;
 use quickwit_proto::indexing::{IndexingServiceClient, ShardPositionsUpdate};
-use quickwit_proto::ingest::ingester::{IngesterService, IngesterServiceClient, IngesterStatus};
+use quickwit_proto::ingest::ingester::{
+    IngesterService, IngesterServiceClient, IngesterServiceTowerLayerStack, IngesterStatus,
+};
 use quickwit_proto::ingest::router::IngestRouterServiceClient;
 use quickwit_proto::metastore::{
     EntityKind, ListIndexesMetadataRequest, MetastoreError, MetastoreService,
@@ -187,7 +189,8 @@ struct QuickwitServices {
     pub ingest_service: IngestServiceClient,
     // Ingest v2
     pub ingest_router_service: IngestRouterServiceClient,
-    pub ingester_opt: Option<Ingester>,
+    ingester_opt: Option<Ingester>,
+
     pub janitor_service_opt: Option<Mailbox<JanitorService>>,
     pub jaeger_service_opt: Option<JaegerService>,
     pub otlp_logs_service_opt: Option<OtlpGrpcLogsService>,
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-serve/src/lib.rs
```diff
@@ -851,12 +878,8 @@ async fn setup_ingest_v2(
                         // metrics, so we use both metrics layers.
                         let ingester = ingester_opt_clone_clone
                             .expect("ingester service should be initialized");
-                        let layers = ServiceBuilder::new()
-                            .layer(INGEST_GRPC_CLIENT_METRICS_LAYER.clone())
-                            .layer(INGEST_GRPC_SERVER_METRICS_LAYER.clone())
```
fulmicoton, line 856: the server metrics is already part of the stack.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -789,6 +803,19 @@ pub async fn serve_quickwit(
...
+fn ingester_service_layer_stack() -> IngesterServiceTowerLayerStack {
+    IngesterServiceClient::tower()
+        .stack_persist_layer(quickwit_common::tower::OneTaskPerCallLayer)
+        .stack_open_replication_stream_layer(quickwit_common::tower::OneTaskPerCallLayer)
+        .stack_init_shards_layer(quickwit_common::tower::OneTaskPerCallLayer)
+        .stack_retain_shards_layer(quickwit_common::tower::OneTaskPerCallLayer)
+        .stack_truncate_shards_layer(quickwit_common::tower::OneTaskPerCallLayer)
+        .stack_close_shards_layer(quickwit_common::tower::OneTaskPerCallLayer)
+        .stack_decommission_layer(quickwit_common::tower::OneTaskPerCallLayer)
+        .stack_layer(INGEST_GRPC_SERVER_METRICS_LAYER.clone())
```
guilload, line 816: As I explained in another PR, we want `.stack_layer(INGEST_GRPC_SERVER_METRICS_LAYER.clone())` first.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -878,12 +905,8 @@ async fn setup_ingest_v2(
...
                         let ingester = ingester_opt_clone_clone
                             .expect("ingester service should be initialized");
-                        let shared_layers = ServiceBuilder::new()
-                            .layer(INGEST_GRPC_CLIENT_METRICS_LAYER.clone())
-                            .layer(INGEST_GRPC_SERVER_METRICS_LAYER.clone())
-                            .into_inner();
-                        let ingester_service = IngesterServiceClient::tower()
-                            .stack_layer(shared_layers)
+                        let ingester_service = ingester_service_layer_stack()
+                            .stack_layer(INGEST_GRPC_CLIENT_METRICS_LAYER.clone())
```
guilload, line 909: Ideally, this layer would go first instead of last.

guilload (approved):

PR #4878, opened 2024-04-17
change dashboards 's timezone to null

hoangphuocbk:
### Description

Change Grafana dashboard timezone to browser from utc, users can see dashboard with the local time.

### DIFF
--- a/monitoring/grafana/dashboards/indexers.json
+++ b/monitoring/grafana/dashboards/indexers.json
@@ -1242,7 +1242,7 @@
     "to": "now"
   },
   "timepicker": {},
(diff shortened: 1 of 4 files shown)


guilload:
What's the benefit of having the default being "" vs. "utc"?

hoangphuocbk:
> What's the benefit of having the default being "" vs. "utc"?
with "utc" the timezone is fixed, with "" the timezone can be changed in user's profile. Sometimes when I investigate problem, it's little bit confusing what time it is :).
image

guilload (approved):

PR #4879, opened 2024-04-17
External Postgres metastore for ECS distrib

rdettai:
### Description

Add the possibility to use an external postgres for the metastore.

Also used the opportunity to add:
- extra ECR policies to enable using pul through cache
- configurable storage size

### How was this PR tested?

Tested in internal CI

### DIFF
--- a/distribution/ecs/quickwit/service/variables.tf
+++ b/distribution/ecs/quickwit/service/variables.tf
@@ -52,6 +52,8 @@ variable "quickwit_peer_list" {
 
 variable "s3_access_policy_arn" {}
 
+variable "task_execution_policy_arn" {}
+
 variable "quickwit_cpu_architecture" {}
 
 variable "module_id" {}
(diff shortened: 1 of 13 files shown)


distribution/ecs/quickwit/service/variables.tf
```diff
@@ -41,6 +41,7 @@ variable "service_config" {
     desired_count = optional(number, 1)
     memory        = number
     cpu           = number
+    storage       = optional(number, 21)
```
guilload, line 44:
```suggestion
    ephemeral_storage_size_gib       = optional(number, 21)
```

I had a use case yesterday for also attaching some form of persistent storage (ebs or efs), so I think "ephemeral" is less ambiguous.

distribution/ecs/quickwit/variables.tf
```diff
@@ -127,3 +129,8 @@ variable "rds_config" {
   })
   default = {}
 }
+
+variable "external_postgres_uri_parameter_arn" {
```
guilload, line 133:
```suggestion
variable "external_postgres_uri_ssm_parameter_arn" {
```

guilload (approved):

PR #4881, opened 2024-04-18
Making the source in flight component lazy.

fulmicoton:
Right now we pollute our metrics with unused sources (like pulsar, kafka etc.).
After this PR, only sources that are actually used will show up.

### DIFF
--- a/quickwit/quickwit-common/src/metrics.rs
+++ b/quickwit/quickwit-common/src/metrics.rs
@@ -18,6 +18,7 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::collections::HashMap;
+use std::sync::OnceLock;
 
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-common/src/metrics.rs
```diff
@@ -19,7 +19,7 @@
 
 use std::collections::HashMap;
 
-use once_cell::sync::Lazy;
+use once_cell::sync::{Lazy, OnceCell};
```
guilload, line 22: You could have kept using `Lazy`, but if you prefer the `OnceCell` API, there's an equivalent in the std now since 1.70: [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html)

guilload (approved):

PR #4884, opened 2024-04-19
Added a build_info metric

fulmicoton:
Here is the line:
quickwit_build_info{build_date="2024-04-19T01:24:49Z",commit_hash="d2d5809922a0ef04ec7586048d66db819528c128",target="aarch64-apple-darwin",version="0.8.0-nightly"} 1

### DIFF
--- a/quickwit/quickwit-cli/src/main.rs
+++ b/quickwit/quickwit-cli/src/main.rs
@@ -19,6 +19,8 @@
 
 #![recursion_limit = "256"]
 
+use std::collections::BTreeMap;
+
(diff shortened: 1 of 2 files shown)


guilload (approved):
LGTM.

Inspired by `cida_instance_up{app="...", host="...", instance="...", region="..."}`, I was about to add a `quickwit_node_up` metric with at least `node_id` and maybe `roles`. Should we make it a single metric?

fulmicoton:
@guilload Either is fine I think.

PR #4887, opened 2024-04-22
Adding a concurrency limit to rest requests.

fulmicoton:
This prevents the explosion of memory observed in presence of 100+ clients, sending a lot of small batches.


fulmicoton:
image

After on the right
Before on the left

Both with  `chico(rate=100kb,batch=20kb)*300`

fulmicoton:
I did not managed to wrangle the type system  in a way that would allow a LoadShedLayer :(

PR #4897, opened 2024-04-23
Configurable IAM policies for ECS distrib indexers

rdettai:
### Description

To be able to use AWS services as sources (e.g Kinesis), it is useful to be able to provide custom roles to the indexer task.

To keep the user facing configuration as simple as possible, only indexers can have extra policies (other nodes are not supposed to interact with external AWS services).

### How was this PR tested?

### DIFF
--- a/distribution/ecs/README.md
+++ b/distribution/ecs/README.md
@@ -55,6 +55,10 @@ You can use sidecars to inject additional secrets as files. This can be
 useful for configuring sources such as Kafka. See `./example/kafka.tf` for an
 example.
 
+To access external AWS services like the Kinesis source, use the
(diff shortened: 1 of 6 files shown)


distribution/ecs/README.md
```diff
@@ -55,6 +55,10 @@ You can use sidecars to inject additional secrets as files. This can be
 useful for configuring sources such as Kafka. See `./example/kafka.tf` for an
 example.
 
+To access external AWS services like the Kinesis source, use the
+`quickwit_indexer.extra_task_policy_arns` variable to attach the necessary
+IAM policies to the nodes.
```
guilload, line 60:
```suggestion
IAM policies to indexers.
```

guilload (approved):

PR #4900, opened 2024-04-23
Populate `items` field from Elastic bulk error

guilload:
### Description
Per title.

### How was this PR tested?
- added unit test
- added integration tests

### DIFF
--- a/quickwit/rest-api-tests/run_tests.py
+++ b/quickwit/rest-api-tests/run_tests.py
@@ -1,18 +1,20 @@
 #!/usr/bin/env python3
 
-import subprocess
-import requests
 import glob
-import yaml
-import sys
-from os import path as osp
-from os import mkdir
 import gzip
 import http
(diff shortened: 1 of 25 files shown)


quickwit/rest-api-tests/run_tests.py
```diff
@@ -85,7 +85,7 @@ def run_request_step(method, step, previous_result):
         step["headers"] = {'user-agent': 'my-app/0.0.1'}
     method_req = getattr(requests, method.lower())
     endpoint = step.get("endpoint", "")
-    url = step["api_root"] + endpoint
+    url = os.path.join(step["api_root"], endpoint)
```
fulmicoton, line 88: hm... this is a url here. not a filepath. For instance, I think os.path.join uses "\" as a delimiter on windows.

fulmicoton (approved):

PR #4918, opened 2024-04-26
upgrade aws-sdk

### LINKED ISSUE #4822: Upgrade aws
We need to upgrade AWS's sdk. It was suggested that #4799 is caused by a bug in the version we are using.

It includes a bunch of API change so this might be fairly annoying.

```
aws-config = "1.1"
aws-credential-types = { version = "1.1", features = [
  "hardcoded-credentials",
] }
aws-sdk-kinesis = "1.18"
aws-sdk-s3 = "1.21"
aws-smithy-async = "1.2"
# aws-smithy-client = "0.55.0" <- not sure what these two should become.
# aws-smithy-http = "0.55.0"
aws-smithy-types = "1.1.8"
aws-types = "1.1.8"
```

trinity-1686a:
### Description

fix #4822

### How was this PR tested?

updated unit tests


guilload:
AMAZING

tuziben:
AMAZING++

fulmicoton:
no wayyyyyyyyyy!

quickwit/quickwit-cli/src/source.rs
```diff
@@ -574,8 +574,8 @@ mod tests {
                 json!({"foo": {"bar": JsonValue::Bool(true)}, "baz": JsonValue::Bool(false)})
             ),
             vec![
-                ("foo.bar".to_string(), JsonValue::Bool(true)),
                 ("baz".to_string(), JsonValue::Bool(false)),
+                ("foo.bar".to_string(), JsonValue::Bool(true)),
```
fulmicoton, line 578: what cause the change?

fulmicoton: ok so aws forces the `preserve_order` feature from `serde_json`... Hmmm we need to make sure this does not lead to a massive regression before shipping this.

fulmicoton: At first sight, it seems preserve_order is only enabled for tests. Do you happen to know a way to make sure this is the case?

fulmicoton: or even better, see if there is a set of optjions to prevent aws from relying on `serde_json`   at all or with `preserve_order`

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -595,11 +599,26 @@ impl S3CompatibleObjectStorage {
...
                 .map(|path| ObjectIdentifier::builder().key(self.key(path)).build())
-                .collect();
-            let delete = Delete::builder().set_objects(Some(objects)).build();
+                .collect()
+            else {
+                error = Some(
+                    StorageErrorKind::Internal
+                        .with_error(anyhow!("failed to build object identifier")),
+                );
+                unattempted.extend(chunk.iter().map(|path| path.to_path_buf()));
```
fulmicoton, line 611: that code logic is not good:
we batch things in chunk... If one of the object is wrong in the chunk we ditch the entire chunk, but not the entire request, even though at this point not request happened.

This does not make much sense. Chunking should happen after creating the object identifier. 
The code woudl be simpler and the spec would make way more sense.

quickwit/quickwit-storage/src/split.rs
```diff
@@ -59,11 +61,34 @@ async fn range_byte_stream_from_payloads(
...
 
-    let body = Body::wrap_stream(stream::iter(bytestreams).flatten());
-    let concat_stream = ByteStream::new(body.into());
+    let body = Body::wrap_stream(stream::iter(bytestreams).map(StreamAdaptor).flatten());
+    let concat_stream = ByteStream::new(SdkBody::from_body_0_4(body));
     Ok(concat_stream)
 }
 
+// With sdk 1.0, ByteStream no longer implement Stream, despite having analogous functions
+// this adaptor is just meant to make it implmeent Stream for places where we really need it
```
fulmicoton, line 70:
```suggestion
// this adaptor is just meant to make it implement Stream for places where we really need it
```

quickwit/quickwit-cli/src/source.rs
trinity-1686a: > At first sight, it seems preserve_order is only enabled for tests.

from what i can tell, this is indeed the case

> Do you happen to know a way to make sure this is the case?

no, but i can try to find one. An option (which i don't like) would be to override our dependencies with fork of `serde_json` for which that feature does nothing

quickwit/quickwit-storage/src/split.rs
```diff
@@ -59,11 +61,34 @@ async fn range_byte_stream_from_payloads(
...
+    fn poll_next(
+        self: Pin<&mut Self>,
+        ctx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        self.project().0.poll_next(ctx)
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (min_u64, max_u64) = self.0.size_hint();
+        let min = min_u64.try_into().unwrap_or(usize::MAX);
```
fulmicoton, line 86:
```suggestion
        let min: usize = min_u64.try_into().unwrap_or(0);
```

trinity-1686a: I think MAX is the more correct value. If we fail the conversion, min_u64 was larger than what an usize can handle on that platform. zero is an always valid value to return, but in this case MAX is also valid, and more informative

quickwit/quickwit-storage/src/split.rs
```diff
@@ -59,11 +61,34 @@ async fn range_byte_stream_from_payloads(
...
+
+    fn poll_next(
+        self: Pin<&mut Self>,
+        ctx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        self.project().0.poll_next(ctx)
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (min_u64, max_u64) = self.0.size_hint();
```
fulmicoton, line 85: Nitpick... I think lower_bound / upper_bound is slightly better naming than min/max.

```suggestion
        let (lower_bound_u64, upper_bound_u64_opt) = self.0.size_hint();
```

fulmicoton (changes requested):
Awesome job. Please see comments

quickwit/quickwit-storage/src/split.rs
fulmicoton: oh I see! Yes you are right!

fulmicoton: Can you add a comment so future readers don't make the same mistake as me?

fulmicoton (approved):

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -643,7 +675,7 @@ impl S3CompatibleObjectStorage {
                 }
                 Err(delete_objects_error) => {
                     error = Some(delete_objects_error.into());
-                    unattempted.extend(chunk.iter().map(|path| path.to_path_buf()));
+                    unattempted.extend(chunk.iter().map(|(path, _oid)| path.to_path_buf()));
```
fulmicoton, line 678: This is technically incorrect. These path have been attempted. 
For instance, at this point, they might be deleted.

fulmicoton (approved):

fulmicoton:
I did not include this PR in the 0501 airmail test to avoid risk, but we can try that next time.

PR #4921, opened 2024-04-27
Batch `index_metadata` calls in indexing service

guilload:
### Description
Add a batch `indexes_metadata` API to the metastore and use it in the indexing service when spawning new indexing pipelines upon applying a new indexing plan.

### How was this PR tested?
Added unit test

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/file_backed/mod.rs
+++ b/quickwit/quickwit-metastore/src/metastore/file_backed/mod.rs
@@ -39,23 +39,27 @@ use std::time::Duration;
 
 use async_trait::async_trait;
 use futures::future::try_join_all;
+use futures::stream::FuturesUnordered;
+use futures::StreamExt;
 use itertools::Itertools;
 use quickwit_common::ServiceStream;
 use quickwit_config::IndexTemplate;
 use quickwit_proto::metastore::{
     serde_utils, AcquireShardsRequest, AcquireShardsResponse, AddSourceRequest, CreateIndexRequest,
     CreateIndexResponse, CreateIndexTemplateRequest, DeleteIndexRequest,
-    DeleteIndexTemplatesRequest, DeleteQuery, DeleteShardsRequest, DeleteSourceRequest,
-    DeleteSplitsRequest, DeleteTask, EmptyResponse, EntityKind, FindIndexTemplateMatchesRequest,
-    FindIndexTemplateMatchesResponse, GetIndexTemplateRequest, GetIndexTemplateResponse,
-    IndexMetadataRequest, IndexMetadataResponse, IndexTemplateMatch, LastDeleteOpstampRequest,
-    LastDeleteOpstampResponse, ListDeleteTasksRequest, ListDeleteTasksResponse,
-    ListIndexTemplatesRequest, ListIndexTemplatesResponse, ListIndexesMetadataRequest,
-    ListIndexesMetadataResponse, ListShardsRequest, ListShardsResponse, ListSplitsRequest,
-    ListSplitsResponse, ListStaleSplitsRequest, MarkSplitsForDeletionRequest, MetastoreError,
-    MetastoreResult, MetastoreService, MetastoreServiceStream, OpenShardSubrequest,
-    OpenShardsRequest, OpenShardsResponse, PublishSplitsRequest, ResetSourceCheckpointRequest,
-    StageSplitsRequest, ToggleSourceRequest, UpdateIndexRequest, UpdateSplitsDeleteOpstampRequest,
+    DeleteIndexTemplatesRequest, DeleteQuery, DeleteShardsRequest, DeleteShardsResponse,
+    DeleteSourceRequest, DeleteSplitsRequest, DeleteTask, EmptyResponse, EntityKind,
+    FindIndexTemplateMatchesRequest, FindIndexTemplateMatchesResponse, GetIndexTemplateRequest,
+    GetIndexTemplateResponse, IndexMetadataFailure, IndexMetadataFailureReason,
+    IndexMetadataRequest, IndexMetadataResponse, IndexTemplateMatch, IndexesMetadataRequest,
+    IndexesMetadataResponse, LastDeleteOpstampRequest, LastDeleteOpstampResponse,
+    ListDeleteTasksRequest, ListDeleteTasksResponse, ListIndexTemplatesRequest,
+    ListIndexTemplatesResponse, ListIndexesMetadataRequest, ListIndexesMetadataResponse,
+    ListShardsRequest, ListShardsResponse, ListSplitsRequest, ListSplitsResponse,
+    ListStaleSplitsRequest, MarkSplitsForDeletionRequest, MetastoreError, MetastoreResult,
+    MetastoreService, MetastoreServiceStream, OpenShardSubrequest, OpenShardsRequest,
+    OpenShardsResponse, PublishSplitsRequest, ResetSourceCheckpointRequest, StageSplitsRequest,
+    ToggleSourceRequest, UpdateIndexRequest, UpdateSplitsDeleteOpstampRequest,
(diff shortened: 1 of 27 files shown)


quickwit/quickwit-metastore/src/metastore/file_backed/mod.rs
```diff
@@ -715,26 +758,65 @@ impl MetastoreService for FileBackedMetastore {
...
-                    index_id: index_id.to_string(),
-                }));
+            .index_metadata_inner(request.index_id, request.index_uid)
+            .await
+            .map_err(|(metastore_error, _index_id_opt, _index_uid_opt)| metastore_error)?;
+        let response = IndexMetadataResponse::try_from_index_metadata(&index_metadata)?;
+        Ok(response)
+    }
+
+    async fn indexes_metadata(
```
fulmicoton, line 777: This implementation should probably be the default implementation in the metastore trait.

We should also probably have a TODO about making something specific for thefile based metastore. There is no benefit to the batching right now.

quickwit/quickwit-metastore/src/metastore/file_backed/mod.rs
```diff
@@ -715,26 +758,65 @@ impl MetastoreService for FileBackedMetastore {
...
+        &mut self,
+        request: IndexesMetadataRequest,
+    ) -> MetastoreResult<IndexesMetadataResponse> {
+        let mut indexes_metadata: Vec<IndexMetadata> =
+            Vec::with_capacity(request.subrequests.len());
+        let mut failures: Vec<IndexMetadataFailure> = Vec::new();
+
+        let mut index_metadata_futures = FuturesUnordered::new();
```
fulmicoton, line 785: `FuturesUnordered` is useless in the case of the file backed metastore. The inner logic takes a mutex right away.

quickwit/quickwit-metastore/src/metastore/index_metadata/mod.rs
```diff
@@ -99,6 +101,24 @@ impl IndexMetadata {
...
     }
 
+    /// Replaces or removes the current retention policy, returning whether a mutation occurred.
+    pub fn set_retention_policy(&mut self, retention_policy_opt: Option<RetentionPolicy>) -> bool {
+        if self.index_config.retention_policy_opt != retention_policy_opt {
+            self.index_config.retention_policy_opt = retention_policy_opt;
+            return true;
+        }
+        false
+    }
```
fulmicoton, line 111: nitpick (as it does not matter much here)

Early exit style goes:

```suggestion
        if self.index_config.retention_policy_opt == retention_policy_opt {
            return false;
        }
        self.index_config.retention_policy_opt = retention_policy_opt;
        true
    }
```

Because the function is so short, the benefit of early exit is very limited... but then I personally prefer having a else statement.


```
 if self.index_config.retention_policy_opt != retention_policy_opt {
     self.index_config.retention_policy_opt = retention_policy_opt;
     true
} else {
     false
}
```

quickwit/quickwit-metastore/src/metastore/index_metadata/mod.rs
```diff
@@ -99,6 +101,24 @@ impl IndexMetadata {
...
+
+    /// Replaces or removes the current search settings, returning whether a mutation occurred.
+    pub fn set_search_settings(&mut self, search_settings: SearchSettings) -> bool {
+        if self.index_config.search_settings != search_settings {
```
fulmicoton, line 115: same nitpick as above.

quickwit/quickwit-proto/protos/quickwit/metastore.proto
```diff
@@ -240,6 +243,33 @@ message IndexMetadataResponse {
...
+  repeated IndexMetadataSubrequest subrequests = 1;
+}
+
+message IndexMetadataSubrequest {
+  optional string index_id = 1;
+  optional quickwit.common.IndexUid index_uid = 2;
+}
+
+message IndexesMetadataResponse {
+  // A JSON serialized then ZSTD compressed list of `IndexMetadata`: `Vec<IndexMetadata> | JSON | ZSTD`.
```
fulmicoton, line 256: We don't use repeated here to make sure we get a better compression rate?
(it could be worth commenting)

fulmicoton: I am not sure about Tonic but I believe gRPC  supports per method compression. Should we see if we can move zstd compression to the higher layer maybe?

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
```diff
@@ -1310,16 +1389,21 @@ impl MetastoreService for PostgresqlMetastore {
...
+            return Ok(EmptyResponse {});
+        }
+        if let Some(pg_shard) = pg_shards
+            .into_iter()
+            .find(|pg_shard| Position::is_eof_str(&pg_shard.publish_position_inclusive))
         {
-            let message = "failed to delete shard ``: shard is not fully indexed".to_string();
+            let message = format!(
+                "failed to delete shard `{}`: shard is not fully indexed",
+                pg_shard.shard_id
```
fulmicoton, line 1402: I don't quite understand what is the `pg_shards`  varialbe about and even more what this `if let` statement is about.

Isn't the DELETE statement without `RETURNING` returning the number of modified rows?

Assuming this was returning actual deleted rows, why do we check for them to be EOF and log an error message saying otherwise?

fulmicoton (changes requested):
See comments... the PR can be tested in this week airmail test but I think the DELETE statement has something fishy (which was already here though)

fulmicoton:
This PR showed a regression on simian so I pulled it off of today's test with airmail.

With `- chico(rate=300kb)*1000` on 8 indexers and with postgres, I cannot go above 200MB/s.

This was working on qw-airmail-20240423

guilload:
Would you happen to have a clue as to what this PR is breaking?

quickwit/quickwit-metastore/src/metastore/file_backed/mod.rs
guilload: The trait is code-generated, so we can't have a default implementation. I actually don't know how I'd go about implementing this feature in `quickwit-codegen`.

quickwit/quickwit-metastore/src/metastore/file_backed/mod.rs
guilload: `FuturesUnordered` is useless compared to? a for loop? Yes, each future acquires the outer lock but should work on a separate per-index inner lock. We want some level of concurrency there, don't we?

quickwit/quickwit-proto/protos/quickwit/metastore.proto
guilload: - Yes. I'll add a comment.
- Unfortunatley, tonic does not support per method compression. It's all or nothing. We could have a custom codec based on prost codec where compression kicks once the size of a message is larger than some predefined threshold.

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
guilload: Yes, this is trash.

quickwit/quickwit-proto/protos/quickwit/metastore.proto
fulmicoton: thanks for the info.

quickwit/quickwit-metastore/src/metastore/file_backed/mod.rs
fulmicoton: You are right.

fulmicoton:
@guilload no clue. I did not dig further. I can redo the test if you want.

fulmicoton (approved):

fulmicoton:
I confirm it now works normally.

PR #4928, opened 2024-04-30
allow updating of doc mapper

trinity-1686a:
### Description

this pr aims at laying the foundation work for updating the doc mapping of an index

- [x] store a doc mapper version in the doc mapper
- [x] store a doc mapper version in splits meta
- [ ] allow updating the doc mapper, auto-incrementing its version (metastore level)
- [ ] allow updating the doc mapper (api level)
- [x] block merges between multiple version of a doc mapper
- [ ] restart ingestion pipeline so they notice the change in doc mapper decently quickly (stretch goal, maybe put in a 2nd pull request)

### How was this PR tested?

- [x] test we properly block merges (unit tests)
- [ ] test we properly block merges (integration tests)
- [x] test modifying different properties (tokenizer, what is stored...), and how it breaks (or not) search


rdettai:
We should probably merge https://github.com/quickwit-oss/quickwit/pull/4868 before addressing:
- `allow updating the doc mapper (api level)`

trinity-1686a:
we are going to ship this in multiple smaller PR, which means this should be ready for review

trinity-1686a:
made hard to work on by addition of doc_mapping_uid in a different PR. Remade from scratch in #5218

PR #4942, opened 2024-05-06
give tantivy a shared thread pool to run fetch doc on

### LINKED ISSUE #4940: Searcher gets deconnected on query
Observed on Airmail...

Some searcher does not answer its health probe on time and gets killed by k8s.

The current theory is that we have been using spawn_blocking a little bit too liberally in the code...
We use it  in many places to avoid CPU load on the runtime threads.

The trouble with this is that the associated pool thread is not bounded, because it is designed to deal with blocking IO.
500 threads trying to decompress some docstore block + 1 runtime thread  is likely to stall the runtime thread for a large amount of time.

We can confirm that idea, and generally speaking, it is worth auditing all of our usage of spawn_blocking.

Other possibilities would be something holding the runtime threads, or some variants of the lifo slot bug, but the latter is very unlikely.

trinity-1686a:
### Description

likely to fix #4940 
useless without upgrading tantivy post https://github.com/quickwit-oss/tantivy/pull/2386

### DIFF
--- a/quickwit/quickwit-search/src/thread_pool.rs
+++ b/quickwit/quickwit-search/src/thread_pool.rs
@@ -38,17 +38,6 @@ fn build_executor() -> Arc<Executor> {
     Arc::new(Executor::ThreadPool(rayon_pool))
 }
 
-fn search_thread_pool() -> &'static rayon::ThreadPool {
-    let executor_ptr = Arc::as_ptr(SEARCH_THREAD_POOL.get_or_init(build_executor));
-    // ptr is valid for as long as the Arc has strong count, and there is one reference in a
-    // global: ptr is valid for 'static.
-    let executor: &'static Executor = unsafe { &*executor_ptr };
-    match executor {
-        Executor::ThreadPool(pool) => pool,
-        Executor::SingleThread => unreachable!(),
-    }
-}
-
 pub(crate) fn search_executor() -> Arc<Executor> {
     SEARCH_THREAD_POOL.get_or_init(build_executor).clone()
 }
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-search/src/thread_pool.rs
```diff
@@ -18,22 +18,39 @@
...
-    SEARCH_THREAD_POOL.get_or_init(|| {
-        rayon::ThreadPoolBuilder::new()
-            .thread_name(|thread_id| format!("quickwit-search-{thread_id}"))
-            .panic_handler(|_my_panic| {
-                error!("task running in the quickwit search pool panicked");
-            })
-            .build()
-            .expect("Failed to spawn the spawning pool")
-    })
+    let executor_ptr = Arc::as_ptr(SEARCH_THREAD_POOL.get_or_init(build_executor));
```
fulmicoton, line 42: This code is correct, but could break if we modified a different part of the code.

It is sometimes acceptable (especially because it is in the same file, and private stuff), but still, I suspect this is unnecessary here.

Doesn't returning an `Arc` in search_thread_pool work?

fulmicoton:
```rust
pub async fn run_cpu_intensive<F, R>(cpu_heavy_task: F) -> Result<R, Panicked>
where
    F: FnOnce() -> R + Send + 'static,
    R: Send + 'static,
{
    let span = tracing::Span::current();
    search_executor().spawn_blocking(move || {
        let _guard = span.enter();
        let mut active_thread_guard =
            GaugeGuard::from_gauge(&crate::SEARCH_METRICS.active_search_threads_count);
        active_thread_guard.add(1i64);
        cpu_heavy_task()
    }).await.map_err(|_| Panicked)
}
```

trinity-1686a:
(marking as wip until it depends on tantivy/main and not an unmerged branch)

fulmicoton (approved):

PR #4947, opened 2024-05-07
Removing dependency to num_cpus (as the info is available in the std

fulmicoton:
nowadays).


trinity-1686a (approved):
the documentation isn't very verbose on why can `available_parallelism()` fail. I'd be in favor of issuing a warn and returning a small default value instead of panicking

PR #4951, opened 2024-05-07
Improve search split cache docs

rdettai:
### Description

Improve the search split cache UX by:
- being strict on the number of split in the cache instead of letting it exceed the configured limit by one
- making the docs a bit more explicit

### How was this PR tested?

### DIFF
--- a/docs/configuration/node-config.md
+++ b/docs/configuration/node-config.md
@@ -208,16 +208,16 @@ This section contains the configuration options for a Searcher.
 | `partial_request_cache_capacity` | Partial request in memory cache capacity on a Searcher. Cache intermediate state for a request, possibly making subsequent requests faster. It can be disabled by setting the size to `0`. | `64M` |
 | `max_num_concurrent_split_searches` | Maximum number of concurrent split search requests running on a Searcher. | `100` |
 | `max_num_concurrent_split_streams` | Maximum number of concurrent split stream requests running on a Searcher. | `100` |
-| `split_cache` | Searcher split cache configuration options defined in the section below. | |
+| `split_cache` | Searcher split cache configuration options defined in the section below. | _disabled_ |
 
 
 ### Searcher split cache configuration
 
-This section contains the configuration options for the searcher split cache.
+This section contains the configuration options for the on disk searcher split cache.
 
 | Property | Description | Default value |
 | --- | --- | --- |
-| `max_num_bytes` | Maximum size in bytes allowed in the split cache. | `1G` |
+| `max_num_bytes` | Maximum disk size in bytes allowed in the split cache. Can be exceeded by the size of one split. | |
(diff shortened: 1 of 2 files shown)


rdettai:
Was there a strong rational behind letting `max_num_splits` be exceedable by 1?

docs/configuration/node-config.md
```diff
@@ -208,16 +208,16 @@ This section contains the configuration options for a Searcher.
 | `partial_request_cache_capacity` | Partial request in memory cache capacity on a Searcher. Cache intermediate state for a request, possibly making subsequent requests faster. It can be disabled by setting the size to `0`. | `64M` |
 | `max_num_concurrent_split_searches` | Maximum number of concurrent split search requests running on a Searcher. | `100` |
 | `max_num_concurrent_split_streams` | Maximum number of concurrent split stream requests running on a Searcher. | `100` |
-| `split_cache` | Searcher split cache configuration options defined in the section below. | |
+| `split_cache` | Searcher split cache configuration options defined in the section below. | _disabled_ |
```
fulmicoton, line 211: I'm afraid people will interpret that as

```
split_cache: _disabled_
```

is a valid conf.

fulmicoton (approved):
approved for convenience but please have a look at the comment

docs/configuration/node-config.md
```diff
@@ -208,16 +208,16 @@ This section contains the configuration options for a Searcher.
...
 
 ### Searcher split cache configuration
 
-This section contains the configuration options for the searcher split cache.
+This section contains the configuration options for the on disk searcher split cache.
 
 | Property | Description | Default value |
 | --- | --- | --- |
-| `max_num_bytes` | Maximum size in bytes allowed in the split cache. | `1G` |
+| `max_num_bytes` | Maximum disk size in bytes allowed in the split cache. Can be exceeded by the size of one split. | |
```
fulmicoton, line 220:
```suggestion
| `max_num_bytes` | Maximum disk size in bytes allowed in the split cache. Can be exceeded by the size of one split. | |
```

I wonder if readers will understand what this is, and whether it really matters.
Isn't this just noise?

rdettai: Wouldn't you run into errors if you assigned `X` GB of disk to an searcher instance and the same `X` GB to max_num_bytes ? FYI, the context is that I'm trying to summarize all the resource configurations to put together a guide for node sizing (CPU, RAM and disk)

docs/configuration/node-config.md
rdettai: Good point. I'll do that and explain what it does in the description.

fulmicoton (approved):

docs/configuration/node-config.md
rdettai: Agreed about this during a discussion.

PR #4956, opened 2024-05-08
Issue on query per list shards request instead of one per subrequest

guilload:
### Description
Per title.

### How was this PR tested?
Updated unit tests.

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
+++ b/quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
@@ -17,6 +17,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+use std::collections::HashMap;
 use std::fmt::{self, Write};
 
 use async_trait::async_trait;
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
```diff
@@ -1329,40 +1331,77 @@ impl MetastoreService for PostgresqlMetastore {
...
-            let pg_shards: Vec<PgShard> = sqlx::query_as(LIST_SHARDS_QUERY)
-                .bind(subrequest.index_uid().to_string())
-                .bind(&subrequest.source_id)
-                .bind(shard_state)
-                .fetch_all(&self.connection_pool)
-                .await?;
+            sql_subquery_builder
+                .column(Asterisk)
+                .from(Shards::Table)
+                .and_where(Expr::col(Shards::IndexUid).eq(subrequest.index_uid().to_string()))
```
fulmicoton, line 1349: do we still need the `.to_string()`?

fulmicoton (approved):

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
guilload: Yeah, `sea-query` uses a different `Value` type than `sqlx`, which only supports owned types.

PR #4962, opened 2024-05-09
multi index leaf requests

PSeitz:
execute one leaf search request per node, instead one leaf search request per index.
A leaf search can span now multiple indices.

Especially in cases with search requests on many indices this may save a
lot of request overhead to other nodes. It also allows better control of
the aggregation memory limit of a request to a node, if it is not split up
over multiple requests, as well as finer control over concurrency.

The new `LeafSearchRequest` contains deduplication, so that if a request spans 1000 indices with the same doc_mapper, we send the doc_mapper only once.

TODO:
This version has no special optimizations for multi index early exit.
e.g. CanSplitDoBetter should be altered to handle multi indices.

### DIFF
--- a/quickwit/quickwit-search/src/leaf.rs
+++ b/quickwit/quickwit-search/src/leaf.rs
@@ -860,9 +860,25 @@ pub async fn multi_leaf_search(
 
     for leaf_search_request_ref in leaf_search_request.leaf_requests.into_iter() {
         let index_uri = quickwit_common::uri::Uri::from_str(
-            &leaf_search_request.index_uris[leaf_search_request_ref.index_uri_ord as usize],
+            &leaf_search_request
+                .index_uris
+                .get(leaf_search_request_ref.index_uri_ord as usize)
+                .ok_or_else(|| {
+                    SearchError::Internal(format!(
+                        "Received incorrect request, index_uri_ord out of bounds: {}",
+                        leaf_search_request_ref.index_uri_ord
+                    ))
+                })?,
         )?;
-        let doc_mapper = doc_mappers[leaf_search_request_ref.doc_mapper_ord as usize].clone();
+        let doc_mapper = doc_mappers
+            .get(leaf_search_request_ref.doc_mapper_ord as usize)
+            .ok_or_else(|| {
+                SearchError::Internal(format!(
+                    "Received incorrect request, doc_mapper_ord out of bounds: {}",
+                    leaf_search_request_ref.doc_mapper_ord
+                ))
+            })?
+            .clone();
 
         let leaf_request_future = tokio::spawn(
             resolve_storage_and_leaf_search(
@@ -897,14 +913,15 @@ pub async fn multi_leaf_search(
         }
     }
 
-    crate::run_cpu_intensive(|| incremental_merge_collector.finalize().map_err(Into::into))
+    crate::search_thread_pool()
+        .run_cpu_intensive(|| incremental_merge_collector.finalize().map_err(Into::into))
         .instrument(info_span!("incremental_merge_finalize"))
         .await
         .context("failed to merge split search responses")?
 }
 
 /// Resolves storage and calls leaf_search
-pub async fn resolve_storage_and_leaf_search(
+async fn resolve_storage_and_leaf_search(
     searcher_context: Arc<SearcherContext>,
     search_request: Arc<SearchRequest>,
     index_uri: quickwit_common::uri::Uri,
@@ -912,10 +929,10 @@ pub async fn resolve_storage_and_leaf_search(
     splits: Vec<SplitIdAndFooterOffsets>,
     doc_mapper: Arc<dyn DocMapper>,
     aggregations_limits: AggregationLimits,
-) -> Result<LeafSearchResponse, SearchError> {
+) -> crate::Result<LeafSearchResponse> {
     let storage = storage_resolver.resolve(&index_uri).await?;
 
-    let resp = leaf_search(
+    leaf_search(
         searcher_context.clone(),
         search_request.clone(),
         storage.clone(),
@@ -923,9 +940,7 @@ pub async fn resolve_storage_and_leaf_search(
         doc_mapper,
         aggregations_limits,
     )
-    .await?;
-
-    Ok(resp)
+    .await
 }
 
 /// `leaf` step of search.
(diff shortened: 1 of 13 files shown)


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -829,6 +829,104 @@ impl CanSplitDoBetter {
...
+    //
+    // It is a little bit tricky how to handle which is now the incremental_merge_collector, one
+    // per index, e.g. when to merge results and how to avoid lock contention.
+    let mut leaf_request_tasks = Vec::new();
+
+    for leaf_search_request_ref in leaf_search_request.leaf_requests.into_iter() {
+        let index_uri = quickwit_common::uri::Uri::from_str(
+            &leaf_search_request.index_uris[leaf_search_request_ref.index_uri_ord as usize],
+        )?;
+        let doc_mapper = doc_mappers[leaf_search_request_ref.doc_mapper_ord as usize].clone();
```
fulmicoton, line 864:
```suggestion
        let doc_mapper = doc_mappers[leaf_search_request_ref.doc_mapper_ord as usize].clone();
```
```suggestion
        let doc_mapper = doc_mappers.get(leaf_search_request_ref.doc_mapper_ord as usize).map_err(...)?.clone();
```

The point is that here the contract of having doc_mapper_ord referring to a properly enforced doc_mappers array is enforced by the grpc client that ran the request. This is code from a different file, running on a different server, possibly from a different version of quickwit, possibly not even coming from quickwit.

It is not something we want to panic on. We should return a proper error here.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -829,6 +829,104 @@ impl CanSplitDoBetter {
...
+    // This will lower resource consumption due to less in-flight futures and avoid contention.
+    // It also allows passing early exit conditions between indices.
+    //
+    // It is a little bit tricky how to handle which is now the incremental_merge_collector, one
+    // per index, e.g. when to merge results and how to avoid lock contention.
+    let mut leaf_request_tasks = Vec::new();
+
+    for leaf_search_request_ref in leaf_search_request.leaf_requests.into_iter() {
+        let index_uri = quickwit_common::uri::Uri::from_str(
+            &leaf_search_request.index_uris[leaf_search_request_ref.index_uri_ord as usize],
```
fulmicoton, line 862: same as below.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -829,6 +829,104 @@ impl CanSplitDoBetter {
...
+            )
+            .in_current_span(),
+        );
+        leaf_request_tasks.push(leaf_request_future);
+    }
+
+    let leaf_responses = try_join_all(leaf_request_tasks).await?;
+    let merge_collector = make_merge_collector(&search_request, &aggregation_limits)?;
+    let mut incremental_merge_collector = IncrementalCollector::new(merge_collector);
+    for result in leaf_responses {
```
fulmicoton, line 884: shouldn't we merge as they come here with a `FuturesUnordered`?

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -829,6 +829,104 @@ impl CanSplitDoBetter {
...
+        match result {
+            Ok(result) => {
+                incremental_merge_collector.add_result(result)?;
```
fulmicoton, line 887: ah add_result is light anyway... maybe it does not matter that much as long as we don't have any lazy optimization.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -829,6 +829,104 @@ impl CanSplitDoBetter {
...
+    }
+
+    crate::run_cpu_intensive(|| incremental_merge_collector.finalize().map_err(Into::into))
+        .instrument(info_span!("incremental_merge_finalize"))
+        .await
+        .context("failed to merge split search responses")?
+}
+
+/// Resolves storage and calls leaf_search
+pub async fn resolve_storage_and_leaf_search(
```
fulmicoton, line 906: does this need to be public.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -829,6 +829,104 @@ impl CanSplitDoBetter {
...
+    searcher_context: Arc<SearcherContext>,
+    search_request: Arc<SearchRequest>,
+    index_uri: quickwit_common::uri::Uri,
+    storage_resolver: StorageResolver,
+    splits: Vec<SplitIdAndFooterOffsets>,
+    doc_mapper: Arc<dyn DocMapper>,
+    aggregations_limits: AggregationLimits,
+) -> Result<LeafSearchResponse, SearchError> {
```
fulmicoton, line 914:
```suggestion
) -> crate::Result<LeafSearchResponse> {
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -829,6 +829,104 @@ impl CanSplitDoBetter {
...
+        searcher_context.clone(),
+        search_request.clone(),
+        storage.clone(),
+        splits,
+        doc_mapper,
+        aggregations_limits,
+    )
+    .await?;
+
+    Ok(resp)
```
fulmicoton, line 927:
```suggestion
    leaf_search(
        searcher_context.clone(),
        search_request.clone(),
        storage.clone(),
        splits,
        doc_mapper,
        aggregations_limits,
    )
    .await
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -976,6 +1080,8 @@ async fn leaf_search_single_split_wrapper(
         }),
     }
     if let Some(last_hit) = locked_incremental_merge_collector.peek_worst_hit() {
+        // TODO: we could use a RWLock instead and read the value instead of updateing it
+        // unconditionally.
```
fulmicoton, line 1084: does it matter? we only update it once per split don't we?

fulmicoton (approved):
see inline comments

quickwit/quickwit-search/src/leaf.rs
PSeitz: I planned this for later when moving the lower loop of splits per index to the top level

quickwit/quickwit-search/src/leaf.rs
PSeitz: I saw some contention on `incremental_merge_collector` when pushing results, which is also once per split.

It's more an issue when we start to share the threshold between indices search (which we don't currently)

PR #4968, opened 2024-05-09
Optimize and batch merge pipelines list splits queries

guilload:
### Description
- Add `node_id` column to `splits` table
- Populate `node_id` column from `splits.split_medata_json` column
- Add hash index on `splits.node_id` column
- Batch list splits query when new spawning new indexing pipelines

### How was this PR tested?
- Updated unit tests
- Ran this SQL query on Cicada which did not return null or empty values

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexing_service.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexing_service.rs
@@ -18,12 +18,13 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::collections::{HashMap, HashSet};
-use std::fmt::{Debug, Display, Formatter};
+use std::fmt::{Debug, Formatter};
 use std::path::PathBuf;
 use std::sync::Arc;
 
 use anyhow::Context;
 use async_trait::async_trait;
+use futures::TryStreamExt;
 use itertools::Itertools;
 use quickwit_actors::{
     Actor, ActorContext, ActorExitStatus, ActorHandle, ActorState, Handler, Healthz, Mailbox,
@@ -43,19 +44,22 @@ use quickwit_ingest::{
 };
 use quickwit_metastore::{
     IndexMetadata, IndexMetadataResponseExt, IndexesMetadataResponseExt,
-    ListIndexesMetadataResponseExt,
+    ListIndexesMetadataResponseExt, ListSplitsQuery, ListSplitsRequestExt, ListSplitsResponseExt,
+    SplitMetadata, SplitState,
 };
 use quickwit_proto::indexing::{
     ApplyIndexingPlanRequest, ApplyIndexingPlanResponse, IndexingError, IndexingPipelineId,
-    IndexingTask, PipelineMetrics,
+    IndexingTask, MergePipelineId, PipelineMetrics,
 };
 use quickwit_proto::metastore::{
     IndexMetadataRequest, IndexMetadataSubrequest, IndexesMetadataRequest,
-    ListIndexesMetadataRequest, MetastoreService, MetastoreServiceClient,
+    ListIndexesMetadataRequest, ListSplitsRequest, MetastoreResult, MetastoreService,
+    MetastoreServiceClient,
 };
-use quickwit_proto::types::{IndexId, IndexUid, PipelineUid, ShardId};
+use quickwit_proto::types::{IndexId, IndexUid, NodeId, PipelineUid, ShardId};
 use quickwit_storage::StorageResolver;
 use serde::{Deserialize, Serialize};
+use time::OffsetDateTime;
 use tokio::sync::Semaphore;
 use tracing::{debug, error, info, warn};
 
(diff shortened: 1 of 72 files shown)


guilload:
@fulmicoton, I still have to add at least three unit tests in `indexing_service.rs` and `merge_pipeline.rs`, but this PR is ready for review.

quickwit/quickwit-indexing/src/actors/merge_pipeline.rs
```diff
@@ -418,6 +415,36 @@ impl MergePipeline {
         }
         Ok(())
     }
+
+    async fn fetch_immature_splits(
+        &mut self,
+        ctx: &ActorContext<Self>,
+    ) -> MetastoreResult<Vec<quickwit_metastore::SplitMetadata>> {
+        if let Some(immature_splits) = self.initial_immature_splits_opt.take() {
```
fulmicoton, line 423:
```suggestion
        // In order to lighten the burden of the metastore, when a node gets assign
        // a set of pipelines, we group the metastore calls listing immature splits into one.
        //
        // This list of splits is stored in `initial_immature_splits_opt` and is only used when
        // we first spawn the pipeline. 
        // If in the future, the pipeline is restarted, we will need to refetch this list of immature splits.
        if let Some(immature_splits) = self.initial_immature_splits_opt.take() {
```

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -410,6 +392,67 @@ impl IndexingService {
         Ok(indexes_metadata)
     }
 
+    /// Fetches the immature splits candidates for merge for all the indexing pipelines for which a
+    /// merge pipeline is not running.
+    async fn fetch_immature_splits(
```
fulmicoton, line 397: can we change the name to express the fact this is target all new pipelines?

(it has the same name as the individual method in the merge pipeline here)

fulmicoton, line 397: I might be missing something: The way the code is made, I suspect we still can hammer the metastore when these index pipelines are new for instance (which is a common case).

In that case, the hashmap returned will not contain any entry for the `MergePipelineId`. 
The initialize_immature_merge will be set to `None` and as a result, we will redo the individual fetch_immature_split.

quickwit/quickwit-indexing/src/actors/merge_pipeline.rs
```diff
@@ -120,7 +123,11 @@ impl Actor for MergePipeline {
 impl MergePipeline {
     // TODO improve API. Maybe it could take a spawnbuilder as argument, hence removing the need
     // for a public create_mailbox / MessageCount.
-    pub fn new(params: MergePipelineParams, spawn_ctx: &SpawnContext) -> Self {
+    pub fn new(
+        params: MergePipelineParams,
+        initial_immature_splits_opt: Option<Vec<SplitMetadata>>,
```
fulmicoton, line 128: Is there a case where we want to call this with `None`? What is it?

quickwit/quickwit-indexing/src/actors/merge_pipeline.rs
```diff
@@ -418,6 +415,36 @@ impl MergePipeline {
...
+    async fn fetch_immature_splits(
+        &mut self,
+        ctx: &ActorContext<Self>,
+    ) -> MetastoreResult<Vec<quickwit_metastore::SplitMetadata>> {
+        if let Some(immature_splits) = self.initial_immature_splits_opt.take() {
+            return Ok(immature_splits);
+        }
+        let index_uid = self.params.pipeline_id.index_uid.clone();
+        let node_id = self.params.pipeline_id.node_id.clone();
+        let list_splits_query = ListSplitsQuery::for_index(index_uid)
```
fulmicoton, line 428: it does not change much tbh, but I wonder if the code after the first if statement should have been calling the other (batched) `fetch_immature_splits` function.

quickwit/quickwit-metastore/migrations/postgresql/19_add-split-node-id-field.up.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE splits
+    ADD COLUMN node_id VARCHAR(253);
+
+UPDATE
```
fulmicoton, line 4:
```suggestion
// Split metadata have been stable for quite a while, so we 
// allow ourselves to do this, but please, reader of the future, 
// do not blindly reapply this pattern.
UPDATE
```

fulmicoton (approved):

fulmicoton (approved):

quickwit/quickwit-indexing/src/actors/indexing_service.rs
guilload: It won't behave the way you described because of how the return `per_merge_pipeline_immature_splits` is initialized.
```
let mut per_merge_pipeline_immature_splits: HashMap<MergePipelineId, Vec<SplitMetadata>> =
            indexing_pipeline_ids
                .iter()
                .map(|indexing_pipeline_id| (indexing_pipeline_id.merge_pipeline_id(), Vec::new()))
                .collect();
```

guilload: This is actually unit-tested.

quickwit/quickwit-indexing/src/actors/merge_pipeline.rs
guilload: So I covered the `spawn_pipelines` branch, which is the one `apply_indexing_plan` uses, but I did not handle`spawn_pipeline`, which I think is used by the local CLI ingest. This is the reason for using an option.

I will refactor the code a little bit to eliminate the option. With some tweaking, `spawn_pipeline` can use `spawn_pipelines` under the hood.

PR #4973, opened 2024-05-10
Stop marking shards as `Unavailable`

guilload:
### Description
Stop marking shards as `Unavailable` + more debug info from router and ingester

### How was this PR tested?
- Added unit tests


quickwit/quickwit-serve/src/lib.rs
```diff
@@ -937,7 +940,7 @@ async fn setup_ingest_v2(
         })
     });
     ingester_pool.listen_for_changes(ingester_change_stream);
-    Ok((ingest_router_service, ingester_opt))
+    Ok((ingest_router, ingest_router_service, ingester_opt))
```
fulmicoton, line 943: nitpick / opinion.

I think I would have prefered to just return the `ingest_router` object here, and would have added an accessor on QuickwitServices to get the service.

In our case, the only layer we use is metrics, so that there is no benefit to making sure there we don't clone several times the same instance, as opposed to building different clients.
THAT BEING SAID... if we were to add a concurrency limit that would lead to different behavior.

We coudl have however have a static OnceLock in the accessor itself to make sure the result is the same as it is today.

fulmicoton (approved):

fulmicoton:
I have no idea what are the long term implication of this PR. Let's keep an eye on how it behaves.

quickwit/quickwit-serve/src/lib.rs
guilload: I'm not a fan either, TBH. I'll clean this up the next time I work on this file.

PR #4974, opened 2024-05-11
add support for aggregations on quickwit ui

### LINKED ISSUE #3223: date-picker only half responsive
**Describe the bug**
the date-picker goes left and right to accommodate screen width, but it does not always move when it should. More precisely, when the window gets bigger than it was when the page was first loaded, the picker moves alright, but when the window get smaller than when the page was loaded, the picker gets stuck and end up out of the viewport

1.  open `/ui/search` on a browser window taking half the screen width. 
2. select an index
3. resize the window. When it get wider, the date picker follows, but when it gets thinner than a half screen, the picker get stuck and end up out of the viewport

**Expected behavior**
The picker should always move to accommodate the window getting resized

**Configuration:**
Please provide:

1. Output of `quickwit --version`: `Quickwit 0.5.0-nightly (fe651b7 2023-04-24T10:20:16Z)`
2. The index_config.yaml: not relevant

trinity-1686a:
### Description

fix #3223
add support for aggregation on quickwit ui

image
image

### How was this PR tested?

manually. I have no idea how to properly test an ui

### Disclaimer

this is my first time interacting with React, and I'm far from being an UI/UX person. What I've done might have huge pitfalls I can't comprehend.


trinity-1686a:
first bug: the `API URL` link at the bottom is very broken with aggregations

trinity-1686a:
term aggregation crash if no matching document

quickwit/quickwit-ui/src/utils/models.ts
```diff
@@ -86,6 +86,100 @@ export type SearchRequest = {
...
+
+export type TermResult = {term: string, value: number}[];
+
+export type HistogramResult = {
+  timestamps: Date[],
+  data: {name: string | undefined, value: number[]}[],
+}
+
+export function extractAggregationResults(aggregation: any): ParsedAggregationResult {
+  console.log("doing the conversion");
```
ddelemeny, line 123: To be removed ?

quickwit/quickwit-ui/src/utils/models.ts
```diff
@@ -86,6 +86,100 @@ export type SearchRequest = {
...
+  const extract_value = (entry: any) => {
+    if (Object.prototype.hasOwnProperty.call(entry, "metric")) {
```
ddelemeny, line 125: nit : `"metric" in entry` is less verbose.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
```diff
@@ -0,0 +1,305 @@
...
+    const updatedMetric = Object.assign({}, metricRef.current, {field: value});
+    const updatedAggregation = Object.assign({}, props.searchRequest.aggregationConfig, {metric: updatedMetric});
+    const updatedSearchRequest = Object.assign({}, props.searchRequest, {aggregationConfig: updatedAggregation});
+    props.onSearchRequestUpdate(updatedSearchRequest);
+    metricRef.current = updatedMetric;
+  };
+
+  return (
+    <Box>
+      <FormControl variant="standard" sx={{ m: 1, minWidth: 120, display: 'flex', flexDirection: 'row', }}>
```
ddelemeny, line 75: FormControl is meant to hold only one input component. Multiple inputs will throw errors in console.

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
```diff
@@ -0,0 +1,305 @@
...
+        <Select
+          value={getAggregationKind(aggregations[0])}
+          onChange={(e) => handleAggregationChange(0, e)}
+          sx={{ "min-height": "44px", width: "190px" }}
+        >
+          { makeOptions(0, aggregations) }
+        </Select>
+        {drawAdditional(0, aggregations)}
```
ddelemeny, line 291: FormControl is meant to hold only one input component. Multiple inputs will throw errors in console.

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
```diff
@@ -0,0 +1,305 @@
...
+          <MenuItem value="1m">1 minute</MenuItem>
+          <MenuItem value="5m">5 minutes</MenuItem>
+          <MenuItem value="10m">10 minutes</MenuItem>
+          <MenuItem value="1h">1 hour</MenuItem>
+          <MenuItem value="1d">1 day</MenuItem>
+        </Select>
+      );
+    }
+    if (isTerm(agg)) {
+      return [(
```
ddelemeny, line 260: You can wrap Elements in a react fragment `<>…</>` instead of returning an array.

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
```diff
@@ -0,0 +1,305 @@
...
+    if (isHistogram(agg)) {
+      return "histogram";
+    }
+    if (isTerm(agg)) {
+      return "term";
+    }
+    return "new";
+  };
+
+  const makeOptions = (pos: number, agg: ({term: TermAgg} | {histogram: HistogramAgg})[]) => {
```
ddelemeny, line 204: Returning an array of elements requires each element to have a `key` prop or react will complain.
It is usually more practical to return the elements wrapped within a fragment `<>…</>.

ddelemeny (changes requested):
Nice job ! Spotted a few things to improve to avoid React/MUI complaints in console + a resource leak to fix.

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
trinity-1686a: oh, I saw some warning regarding FormControl, but couldn't figure out their meaning. Thanks!

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
trinity-1686a: in this case, it seemed easier to use key (no need for code change, and each element can have a reasonable key in form of its value)

quickwit/quickwit-ui/src/components/QueryEditor/QueryEditor.tsx
```diff
@@ -55,6 +55,10 @@ export function QueryEditor(props: SearchComponentProps) {
         runSearchRef.current(searchRequestRef.current);
       },
     })
+    const resize = () => {
+      editor.layout({width: Math.max(window.innerWidth - (260+180+2*24), 200), height: 84});
+    }
+    window.addEventListener('resize', resize);
```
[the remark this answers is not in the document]
trinity-1686a: i couldn't find how to do that with useEffect in this context. I used `handleEditorWillUnmount` instead. I don't know if it is right, it *seems* to work

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
```diff
@@ -0,0 +1,316 @@
...
+            />
+          </FormControl>
+          <FormControl variant="standard">
+            <TextField
+              variant="standard"
+              label="Return top"
+              type="number"
+              onChange={(e) => handleTermCountChange(pos, e)}
+              value={agg.term.size}
+              sx={{ "margin-left": "10px" }}
```
ddelemeny, line 280: Should be camelCased.

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
```diff
@@ -0,0 +1,316 @@
...
+      );
+    }
+    if (isTerm(agg)) {
+      return (<>
+          <FormControl variant="standard">
+            <TextField
+              variant="standard"
+              label="Field"
+              onChange={(e) => handleTermFieldChange(pos, e)}
+              sx={{ "margin-left": "10px" }}
```
ddelemeny, line 270: Should be camelCased.

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
```diff
@@ -0,0 +1,316 @@
...
+          newAggregations[pos] = {
+            term: {
+              field: "",
+              size: 10,
+            }
+          };
+          break;
+        }
+        case "rm": {
+          console.log("av", newAggregations);
```
ddelemeny, line 140: rm console logs

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
```diff
@@ -0,0 +1,316 @@
...
+    }
+    return options;
+  }
+
+  // do the initial filling of parameters
+  const aggregationConfig = props.searchRequest.aggregationConfig;
+  if (aggregationConfig.histogram === null && aggregationConfig.term === null) {
+    const initialAggregation = Object.assign({}, ...aggregations);
+    const initialSearchRequest = {...props.searchRequest, aggregationConfig: initialAggregation};
+    props.onSearchRequestUpdate(initialSearchRequest);
```
ddelemeny, line 239: This will cause a setState in the parent component from the render fn of the current one, which triggers a render somewhere else in the tree. Undesired side-effect inside a render, not allowed (react will warn in the console).

You can wrap this init block in a `useEffect(()=>{/*init here*/},[/*deps*/])` call. Leave the deps array empty if the effect is to be executed once.

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  const drawAdditional = (pos: number, aggs: ({term: TermAgg} | {histogram: HistogramAgg})[]) => {
+    const agg = aggs[pos]
+    if (isHistogram(agg)) {
```
ddelemeny, line 244: This may very well crash the tab of the unsuspecting user, trying to render thousands of buckets.
Search params being serialized in the urls make it even worse (refresh won't help).

I'd favor computing a safe-to-render resolution based on the selected time range.

ddelemeny (changes requested):
Second pass : some minor changes, some react arcane stuff and some safety boundaries required.

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
trinity-1686a: so i should remove that Select? What should i do when no time range is selected? Send 2 requests for the lowest/highest timestamp matching the query?

trinity-1686a: I think the 2nd option would be more friendly. Tutorials are made with "historical" data, where a default range would likely match nothing. 3 is a worst version of 2 imo. 4 sounds bad to use given the simple search allows not providing bounds

fulmicoton:
@fmassot can you take over the review?

quickwit/quickwit-ui/src/components/QueryEditor/AggregationEditor.tsx
JulesVautier-io: I started to implement the 2nd option on a private branch but it seems to add a lot of complexity, potention fetch errors, and lots of lignes of code to the merge requests. So went with the "removing the option here is probably the best course" solution. I did put 1day interval by default to reduce to probability of errors due to the number of buckets selected.

fmassot (approved):

PR #4977, opened 2024-05-13
Added latency metrics for REST endpoint.

fulmicoton:
This PR also makes it possible to set specific buckets for histograms.

Cloes #4932

### DIFF
--- a/quickwit/quickwit-search/src/metrics.rs
+++ b/quickwit/quickwit-search/src/metrics.rs
@@ -20,7 +20,9 @@
 // See https://prometheus.io/docs/practices/naming/
 
 use once_cell::sync::Lazy;
-use quickwit_common::metrics::{new_counter, new_histogram, Histogram, IntCounter};
+use quickwit_common::metrics::{
+    exponential_buckets, new_counter, new_histogram, Histogram, IntCounter,
+};
 
 pub struct SearchMetrics {
     pub leaf_searches_splits_total: IntCounter,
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-search/src/metrics.rs
```diff
@@ -40,6 +42,7 @@ impl Default for SearchMetrics {
                 "Number of seconds required to run a leaf search over a single split. The timer \
                  starts after the semaphore is obtained.",
                 "search",
+                exponential_buckets(0.002, 2.0, 8).unwrap(),
```
trinity-1686a, line 45: last bucket ends at half a second. while i understand smaller steps for leaf_search_split, in my experience having some split being longer than that isn't necessarily rare. imo either the start bucket or the number of bucket needs to be increased

quickwit/quickwit-serve/src/metrics.rs
```diff
@@ -34,6 +35,14 @@ impl Default for RestMetrics {
                 &[],
                 ["method", "status_code"],
             ),
+            request_duration_seconds: new_histogram_vec(
+                "request_duration_seconds",
+                "Response time in millisecs",
```
trinity-1686a, line 40: we use `secs`, not seconds, elsewhere.
the comment gives the wrong unit too

trinity-1686a (approved):
lgtm, just a few small changes

quickwit/quickwit-serve/src/metrics.rs
fulmicoton: good point. Yeah this is tricky. We used secs in all place but one (apart from here) and prometheus docs recommends seconds

fulmicoton: Ill switch to secs.

quickwit/quickwit-search/src/metrics.rs
fulmicoton: did both

PR #4983, opened 2024-05-14
use correct tokenizer when building termsetquery

### LINKED ISSUE #4945: Dynamic fields term set query not working
I am able to get
```
data.event_type:StandardMgaQuoteIssued OR data.event_type:QuoteIssued
```

query to work, but I cannot get a term set query to work, like
```
data.event_type:IN [StandardMgaQuoteIssued QuoteIssued]
```

This is my kafka ingestion configuration
```
version: 0.6
index_id: events

doc_mapping:
  field_mappings:
    - name: id
      type: text
      tokenizer: raw
...
    - name: data
      type: json
      tokenizer: default
    - name: occurred_at
      type: datetime
      fast: true
      input_formats:
        - rfc3339
        - "%Y-%m-%dT%H:%M:%S.%f"
      precision: seconds
  timestamp_field: occurred_at

indexing_settings:
  commit_timeout_secs: 10
```

```
{
  "build": {
    "build_date": "2024-03-29T16:35:13Z",
    "build_profile": "release",
    "build_target": "x86_64-unknown-linux-gnu",
    "cargo_pkg_version": "0.8.1",
    "commit_date": "2024-03-29T14:09:41Z",
    "commit_hash": "e6c53967f8e57401d93bcc555d361dad69bd4ece",
    "commit_short_hash": "e6c5396",
    "commit_tags": [
      "v0.8.1"
    ],
    "version": "v0.8.1"
  },
  "runtime": {
    "num_cpus_logical": 4,
    "num_cpus_physical": 2,
    "num_threads_blocking": 3,
    "num_threads_non_blocking": 1
  }
}
```

----
Im not sure if it may be a feature or a bug, based on the dynamic json fields that it is trying to use.

trinity-1686a:
### Description

fix #4945

### How was this PR tested?

added integration test (also fix an integration test not cleaning up behind)

### DIFF
--- a/quickwit/quickwit-query/src/query_ast/term_query.rs
+++ b/quickwit/quickwit-query/src/query_ast/term_query.rs
@@ -20,9 +20,10 @@
 use std::collections::HashMap;
 
 use serde::{Deserialize, Serialize};
-use tantivy::schema::Schema as TantivySchema;
+use tantivy::schema::{FieldType, Schema as TantivySchema};
 
 use super::{BuildTantivyAst, QueryAst};
+use crate::query_ast::utils::DYNAMIC_FIELD_NAME;
 use crate::query_ast::{FullTextParams, TantivyQueryAst};
 use crate::tokenizers::TokenizerManager;
 use crate::{BooleanOperand, InvalidQuery};
@@ -42,6 +43,55 @@ impl From<TermQuery> for QueryAst {
 }
 
 impl TermQuery {
+    fn get_tokenizer<'a>(&self, schema: &'a TantivySchema) -> Option<&'a str> {
+        let field = schema
+            .find_field(&self.field)
+            .or_else(|| schema.find_field(DYNAMIC_FIELD_NAME))?
+            .0;
+        match schema.get_field_entry(field).field_type() {
+            FieldType::Str(text_options) => Some(text_options.get_indexing_options()?.tokenizer()),
+            FieldType::JsonObject(json_options) => {
+                Some(json_options.get_text_indexing_options()?.tokenizer())
+            }
+            _ => None,
+        }
+    }
+
+    pub(crate) fn ast_for_term_extraction(
+        &self,
+        schema: &TantivySchema,
+        tokenizer_manager: &TokenizerManager,
+    ) -> Result<TantivyQueryAst, InvalidQuery> {
+        self.build_ast(
+            schema,
+            tokenizer_manager,
+            self.get_tokenizer(schema)
+                .or(Some("raw"))
+                .map(ToString::to_string),
+        )
+    }
+
+    fn build_ast(
+        &self,
+        schema: &TantivySchema,
+        tokenizer_manager: &TokenizerManager,
+        tokenizer: Option<String>,
+    ) -> Result<TantivyQueryAst, InvalidQuery> {
+        let full_text_params = FullTextParams {
+            tokenizer,
+            // This parameter should be an Or to handle integers on json fields
+            mode: BooleanOperand::Or.into(),
+            zero_terms_query: Default::default(),
+        };
+        crate::query_ast::utils::full_text_query(
+            &self.field,
+            &self.value,
+            &full_text_params,
+            schema,
+            tokenizer_manager,
+        )
+    }
+
     #[cfg(test)]
     pub fn from_field_value(field: impl ToString, value: impl ToString) -> Self {
         Self {
@@ -59,19 +109,7 @@ impl BuildTantivyAst for TermQuery {
         _search_fields: &[String],
         _with_validation: bool,
     ) -> Result<TantivyQueryAst, InvalidQuery> {
-        let full_text_params = FullTextParams {
-            tokenizer: Some("raw".to_string()),
-            // The parameter below won't matter, since we will have only one term
-            mode: BooleanOperand::Or.into(),
-            zero_terms_query: Default::default(),
-        };
-        crate::query_ast::utils::full_text_query(
-            &self.field,
-            &self.value,
-            &full_text_params,
-            schema,
-            tokenizer_manager,
-        )
+        self.build_ast(schema, tokenizer_manager, Some("raw".to_string()))
     }
 }
 
--- a/quickwit/quickwit-query/src/query_ast/term_set_query.rs
+++ b/quickwit/quickwit-query/src/query_ast/term_set_query.rs
@@ -56,8 +56,7 @@ impl TermSetQuery {
                     field: full_path.to_string(),
                     value: value.to_string(),
                 };
-                let ast =
-                    term_query.build_tantivy_ast_call(schema, tokenizer_manager, &[], false)?;
+                let ast = term_query.ast_for_term_extraction(schema, tokenizer_manager)?;
                 let tantivy_query: Box<dyn crate::TantivyQuery> = ast.simplify().into();
                 tantivy_query.query_terms(&mut |term, _| {
                     terms.insert(term.clone());
--- a/quickwit/quickwit-query/src/query_ast/utils.rs
+++ b/quickwit/quickwit-query/src/query_ast/utils.rs
@@ -31,7 +31,7 @@ use crate::query_ast::tantivy_query_ast::{TantivyBoolQuery, TantivyQueryAst};
 use crate::tokenizers::TokenizerManager;
 use crate::InvalidQuery;
 
-const DYNAMIC_FIELD_NAME: &str = "_dynamic";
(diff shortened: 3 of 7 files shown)


quickwit/quickwit-query/src/query_ast/term_query.rs
```diff
@@ -42,26 +43,43 @@ impl From<TermQuery> for QueryAst {
...
+        )
+    }
+
+    fn build_ast(
+        &self,
+        schema: &TantivySchema,
+        tokenizer_manager: &TokenizerManager,
+        tokenizer: Option<String>,
     ) -> Result<TantivyQueryAst, InvalidQuery> {
         let full_text_params = FullTextParams {
```
fulmicoton, line 80: I don't think this works.

What happens if your tokenizer generates two tokens?
You will end up making it into a disjunction.

so `roller-coaster` will match `roller-blade`... I'm not sure what the right spec is but this one is too naive.

trinity-1686a: we could use `get_normalizer()` instead, which would just lowercase (if required), and never split tokens. That would make  `roller-coaster` match nothing instead of things it shouldn't. Properly matching would require not creating a TermSetQuery, but a bool query of what works in a termset on one side, and what doesn't on the other, and i'm not sure that's worth the extra trouble

fulmicoton:
I think we want this to behave such as:

`field IN [xxxx yyyy]`

should behave exactly like 

`field:xxxx OR field:yyyy`

quickwit/quickwit-query/src/query_ast/term_query.rs
```diff
@@ -42,26 +43,43 @@ impl From<TermQuery> for QueryAst {
...
-    #[cfg(test)]
-    pub fn from_field_value(field: impl ToString, value: impl ToString) -> Self {
-        Self {
-            field: field.to_string(),
-            value: value.to_string(),
+    fn get_tokenizer<'a>(&self, schema: &'a TantivySchema) -> Option<&'a str> {
+        let field = schema
+            .find_field(&self.field)
+            .or_else(|| schema.find_field(DYNAMIC_FIELD_NAME))?
+            .0;
```
fulmicoton, line 50:
```suggestion
        let (field, _json_path) = schema
            .find_field(&self.field)
            .or_else(|| schema.find_field(DYNAMIC_FIELD_NAME))?;
```

quickwit/quickwit-query/src/query_ast/term_query.rs
```diff
@@ -42,26 +43,43 @@ impl From<TermQuery> for QueryAst {
 }
 
 impl TermQuery {
-    #[cfg(test)]
-    pub fn from_field_value(field: impl ToString, value: impl ToString) -> Self {
-        Self {
-            field: field.to_string(),
-            value: value.to_string(),
+    fn get_tokenizer<'a>(&self, schema: &'a TantivySchema) -> Option<&'a str> {
+        let field = schema
```
fulmicoton, line 47: let field = get_field(&self.field, schema)?;

quickwit/quickwit-query/src/query_ast/term_query.rs
fulmicoton, line 50:
```suggestion
        let field = get_field(&self.field, schema)?;
```

quickwit/quickwit-query/src/query_ast/term_query.rs
```diff
@@ -42,26 +43,43 @@ impl From<TermQuery> for QueryAst {
...
-        _search_fields: &[String],
-        _with_validation: bool,
+    ) -> Result<TantivyQueryAst, InvalidQuery> {
+        self.build_ast(
+            schema,
+            tokenizer_manager,
+            self.get_tokenizer(schema)
+                .or(Some("raw"))
+                .map(ToString::to_string),
+        )
```
fulmicoton, line 71:
```suggestion
        let tokenizer: String = self.get_tokenizer(schema)
            .unwrap_or("raw")
            .to_string();
        self.build_ast(
            schema,
            tokenizer_manager,
             Some(tokenizer)
        )
```

quickwit/quickwit-query/src/query_ast/term_query.rs
```diff
@@ -73,6 +91,26 @@ impl BuildTantivyAst for TermQuery {
...
+
+impl BuildTantivyAst for TermQuery {
+    fn build_tantivy_ast_impl(
+        &self,
+        schema: &TantivySchema,
+        tokenizer_manager: &TokenizerManager,
+        _search_fields: &[String],
+        _with_validation: bool,
+    ) -> Result<TantivyQueryAst, InvalidQuery> {
+        self.build_ast(schema, tokenizer_manager, Some("raw".to_string()))
```
fulmicoton, line 112:
```suggestion
        self.build_ast(schema, tokenizer_manager, "raw".to_string())
```

quickwit/quickwit-query/src/query_ast/term_query.rs
```diff
@@ -20,9 +20,10 @@
 use std::collections::HashMap;
 
 use serde::{Deserialize, Serialize};
-use tantivy::schema::Schema as TantivySchema;
+use tantivy::schema::{FieldType, Schema as TantivySchema};
 
 use super::{BuildTantivyAst, QueryAst};
+use crate::query_ast::utils::DYNAMIC_FIELD_NAME;
```
fulmicoton, line 26: `term_query.rs` should not have been modified at all.

This PR is obfuscating what it does with logic that only concerns term_set_query.

quickwit/quickwit-query/src/query_ast/term_set_query.rs
```diff
@@ -56,15 +63,34 @@ impl TermSetQuery {
                     field: full_path.to_string(),
                     value: value.to_string(),
                 };
-                let ast =
-                    term_query.build_tantivy_ast_call(schema, tokenizer_manager, &[], false)?;
+                let ast = term_query.ast_for_term_extraction(schema, tokenizer_manager)?;
```
fulmicoton, line 66: Let's not build a `TermQuery` at all and add 
a function 

`build_ast_for_term(field_path: &str, value: &str, schema: &Schema, tokenizer_manager: &TokenizerManager)`

quickwit/quickwit-query/src/query_ast/term_set_query.rs
```diff
@@ -56,15 +63,34 @@ impl TermSetQuery {
...
+                    let mut phrase = Vec::with_capacity(terms.len());
+                    for term in terms {
+                        if is_term_str(&term) {
+                            phrase.push(term);
+                        } else {
+                            all_terms.insert(term);
+                        }
+                    }
+                    intersections.push(phrase);
+                }
```
fulmicoton, line 90: The logic is rather odd here. 

You implicitly rely on the fact that you know that the query ast is a disjunction over different field:

e.g.
`(text query OR  numerical query)`

If you want to take that kind of shortcuts (after having considered other solutions), add comments to explain why it works.

Ideally, assert that the query is indeed what you think it is.

quickwit/quickwit-query/src/query_ast/term_set_query.rs
```diff
@@ -76,9 +102,26 @@ impl BuildTantivyAst for TermSetQuery {
         _search_fields: &[String],
         _with_validation: bool,
     ) -> Result<TantivyQueryAst, InvalidQuery> {
-        let terms_it = self.make_term_iterator(schema, tokenizer_manager)?;
-        let term_set_query = tantivy::query::TermSetQuery::new(terms_it);
-        Ok(term_set_query.into())
+        use tantivy::query::{BooleanQuery, Query, TermQuery, TermSetQuery};
+
+        let (terms_it, intersections) = self.make_term_iterator(schema, tokenizer_manager)?;
```
fulmicoton, line 107: we probably need a comment here.

quickwit/quickwit-query/src/query_ast/term_set_query.rs
```diff
@@ -36,13 +36,20 @@ pub struct TermSetQuery {
...
 }
 
+fn is_term_str(term: &Term) -> bool {
+    let val = term.value();
+    let typ = val.json_path_type().unwrap_or_else(|| val.typ());
+    typ == Type::Str
+}
+
 impl TermSetQuery {
     fn make_term_iterator(
```
fulmicoton, line 46: This function has a very non trivial purpose and a retrun type. It needs a comment.

fulmicoton (changes requested):
See comments in PR.
As far as I can tell, there are no reason to put code in the `term_query.rs` file.

PR #4988, opened 2024-05-15
Fix polling interval docs

rdettai:
### Description

Fixes some inacuracies regarding the polling interval configurations for file backed metastores.

### How was this PR tested?

### DIFF
--- a/docs/configuration/metastore-config.md
+++ b/docs/configuration/metastore-config.md
@@ -68,7 +68,7 @@ You can also configure it to poll the File-Backed Metastore periodically to keep
 To configure the polling interval (in seconds), add a URI fragment to the storage URI as follows: `s3://quickwit/my-indexes#polling_interval=30s`
 
 :::note
-The polling interval can be configured in seconds only, other units like minutes or hours are not supported yet.
+The polling interval can be configured in seconds only; other units, such as minutes or hours, are not supported.
 :::
 
 :::tip
--- a/docs/configuration/metastore-config.md
+++ b/docs/configuration/metastore-config.md
@@ -65,7 +65,7 @@ By default, the File-Backed Metastore is only read once when you start a Quickwi
 
 You can also configure it to poll the File-Backed Metastore periodically to keep a fresh view of it. This is useful for a Searcher instance that needs to be aware of new splits published by an Indexer running in parallel.
 
-To configure the polling interval (in seconds), add a URI fragment to the storage URI like this: `s3://quickwit/my-indexes#polling_interval=30s`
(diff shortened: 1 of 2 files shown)


docs/configuration/metastore-config.md
```diff
@@ -65,7 +65,11 @@ By default, the File-Backed Metastore is only read once when you start a Quickwi
 
 You can also configure it to poll the File-Backed Metastore periodically to keep a fresh view of it. This is useful for a Searcher instance that needs to be aware of new splits published by an Indexer running in parallel.
 
-To configure the polling interval (in seconds only), add a URI fragment to the storage URI like this: `s3://quickwit/my-indexes#polling_interval=30s`
+To configure the polling interval (in seconds), add a URI fragment to the storage URI like this: `s3://quickwit/my-indexes#polling_interval=30s`
```
guilload, line 68:
```suggestion
To configure the polling interval (in seconds), add a URI fragment to the storage URI as follows: `s3://quickwit/my-indexes#polling_interval=30s`
```

docs/configuration/metastore-config.md
```diff
@@ -65,7 +65,11 @@ By default, the File-Backed Metastore is only read once when you start a Quickwi
...
+
+:::note
+The polling interval can be configured in seconds only, other units like minutes or hours are not supported yet.
```
guilload, line 71:
```suggestion
The polling interval can be configured in seconds only; other units, such as minutes or hours, are not supported.
```

docs/configuration/node-config.md
```diff
@@ -123,17 +123,7 @@ This section may contain one configuration subsection per available metastore im
...
-| `polling_interval` | Time interval between successive polling attempts to detect metastore changes. | `30s` |
-
-Example of a metastore configuration for a file-backed implementation in YAML format:
-
-```yaml
-metastore:
-  file:
-    polling_interval: 1m
-```
+File-backed metastore doesn't have any node level configuration. You can configure the poll interval [at the index level](./metastore-config.md#polling-configuration).
```
guilload, line 126: Can you quickly add the feature instead? There's already a `FileMetastoreConfig` in `metastore_config.rs`, and most of the boilerplate code is already set up. I suspect it won't take much work.

guilload (approved):

docs/configuration/node-config.md
rdettai: Right, I was wondering if maybe it was implemented at some point and removed because of some perverse side effect.

rdettai: I'm merging this now to fix the docs and I'll add the feature request to my backlog. If nobody complained about this so far, the feature is likely not urgent.

PR #4989, opened 2024-05-15
rewrite agg extended bounds

### LINKED ISSUE #4926: partial request cache still not working when deriving metrics from documents
when using quickwit to derive metrics, you end up with queries looking like this (emitted by grafana datasource):
```json5
{
  "aggs": {
    "2": {
      "date_histogram": {
        "field": "timestamp",
        "fixed_interval": "1m",
        "min_doc_count": 0,
        "extended_bounds": {  // these dates prevent caching
          "min": 1672570377874,
          "max": 1672644841715
        }
      }
    }
  },
  "query": {
    "bool": {
      "filter": {
        "range": {
          "timestamp": { // these dates are correctly handled
            "gte": "2023-01-01T10:52:57.874Z",
            "lte": "2023-01-02T07:34:01.715Z"
          }
        }
      }
    }
  },
  "size": 0
}
```
sadly the `extended_bounds` changes at each query, which we don't handle, and causes the partial request cache to never hit on this kind of aggregation. An improvement could be to remove extended bounds in `leaf_search_single_split` so that the cache can hit, and verify we properly fill with zeroes when merging results from multiple leafs

trinity-1686a:
### Description

fix #4926
remove extended bounds from histogram aggregations at the leaf_search_single_split level.
This should make it possible for the partial request cache to hit on dashboards using that feature, such as grafana.
cc @PSeitz 

### How was this PR tested?

added integration test to verify the behavior doesn't change
added unit test to verify we rewrite the bounds at both top and leaf level
manually tested a few (manually wrote) aggregation and verify cache metrics record hits when expected.

### DIFF
--- a/quickwit/quickwit-search/src/leaf.rs
+++ b/quickwit/quickwit-search/src/leaf.rs
@@ -438,13 +438,10 @@ fn rewrite_aggregation(search_request: &mut SearchRequest) {
             }
         });
         if modified_something {
-            if let Ok(serialized_aggregation) = serde_json::to_string(&aggregations) {
-                // it's fine to put a (Tantivy)Aggregations and not a QuickwitAggregations because
-                // the former is an serde-untagged variant of the later
-                search_request.aggregation_request = Some(serialized_aggregation);
-            } else {
-                warn!("we failed to reserialize aggregation, this shouldn't happen");
-            }
+            // it's fine to put a (Tantivy)Aggregations and not a QuickwitAggregations because
+            // the former is an serde-untagged variant of the later
+            search_request.aggregation_request =
+                Some(serde_json::to_string(&aggregations).expect("serializing should never fail"));
         }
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -410,6 +411,55 @@ fn rewrite_request(
...
+                    histogram.extended_bounds.take().is_some()
+                }
+                AggregationVariants::DateHistogram(histogram) => {
+                    histogram.extended_bounds.take().is_some()
+                }
+                _ => false,
+            }
+        });
+        if modified_something {
+            if let Ok(serialized_aggregation) = serde_json::to_string(&aggregations) {
```
PSeitz, line 441: I think we can use `expect` here

PSeitz (approved):

PR #4997, opened 2024-05-17
support bulk ingest over PUT

### LINKED ISSUE #4995: Log spamming
As observed on airmail.

Quickwit ended up spamming the following logs:
```
insertion rejected due to {\"message\":\"index ID pattern `_bulk` is invalid: patterns must match the following regular expression: `^[a-zA-Z\\\\*][a-zA-Z0-9-_\\\\.\\\\*]{0,254}$`\"}`
```

The indexes are using ingest v2 and are created via the index template.

trinity-1686a:
### Description

likely fix #4995

### How was this PR tested?

added integration test, passing on both quickwit and elastic


trinity-1686a:
note: this is still useful for compatibility with ES

fulmicoton (approved):

PR #4998, opened 2024-05-17
add metric for search split affinity

### LINKED ISSUE #4996: Add metrics to track split affinity ratio
It would be interesting to know how often we end up picking the searcher with the best affinity.

trinity-1686a:
### Description

fix #4996

### How was this PR tested?

verified the metric exists by running a small cluster

### DIFF
--- a/quickwit/quickwit-search/src/metrics.rs
+++ b/quickwit/quickwit-search/src/metrics.rs
@@ -27,6 +27,8 @@ use quickwit_common::metrics::{
 pub struct SearchMetrics {
     pub leaf_searches_splits_total: IntCounter,
     pub leaf_search_split_duration_secs: Histogram,
+    pub job_assigned_total: IntCounter,
+    pub job_assigned_to_affinity_searcher: IntCounter,
 }
 
 impl Default for SearchMetrics {
@@ -44,6 +46,18 @@ impl Default for SearchMetrics {
                  starts after the semaphore is obtained.",
                 "search",
                 exponential_buckets(0.005, 2.0, 10).unwrap(),
+            ),
+            job_assigned_total: new_counter(
+                "job_assigned_total",
+                "Number of job assigned to searchers",
+                "search",
+                &[],
+            ),
+            job_assigned_to_affinity_searcher: new_counter(
+                "job_assigned_to_affinity_searcher",
+                "Number of job assigned to the searcher of highest affinity",
+                "search",
+                &[],
             ),
         }
     }
--- a/quickwit/quickwit-search/src/search_job_placer.rs
+++ b/quickwit/quickwit-search/src/search_job_placer.rs
@@ -29,7 +29,7 @@ use quickwit_common::pubsub::EventSubscriber;
 use quickwit_common::rendezvous_hasher::{node_affinity, sort_by_rendez_vous_hash};
 use quickwit_proto::search::{ReportSplit, ReportSplitsRequest};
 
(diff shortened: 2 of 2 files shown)


quickwit/quickwit-search/src/search_job_placer.rs
```diff
@@ -181,10 +181,22 @@ impl SearchJobPlacer {
             sort_by_rendez_vous_hash(&mut candidate_nodes, job.split_id());
             // Select the least loaded node.
             let chosen_node_idx = if candidate_nodes.len() >= 2 {
+                // TODO: we currently give little flexibility. By allowing node 0 to be slighly
+                // more loaded, we could improve the chance of picking the same node for multiple
+                // queries, without impacting load balancing much.
+                // for instance allowing node 0 to have sum(job.cost)/count(searcher) * 0.05 of
+                // load more than node 1 seems to improve choosing the 1st node in simulations,
+                // while not overloading a searcher by more than 5% over what a simple power of two
+                // choices would
```
trinity-1686a, line 190: some simulations i ran:
|config|1st min/max|1st hit|pow2 min/max|pow2 hit|biased min/max|biased hit|
|-|-|-|-|-|-|-|
|30 nodes, 20k splits| 597/705 | 1.0 | 665/668 | 0.68 | 643/681 | 0.97 |
|3 nodes, 200 splits| 48/52 | 1.0 | 50/50 | 0.78 | 49/52 | 0.98 |

1st is always take 1st searcher, pow2 is power of 2 choices, biased is power of 2 choices biased toward the 1st.
min/max: number min and max of jobs assigned on a searcher (closer is better)
hit: ratio of jobs that ended up picking their 1st searcher (higher is better)

fulmicoton: I am not sure what you called biased... Is it the thing that we do today?

quickwit/quickwit-search/src/metrics.rs
```diff
@@ -45,6 +47,18 @@ impl Default for SearchMetrics {
...
                 exponential_buckets(0.005, 2.0, 10).unwrap(),
             ),
+            job_assigned_total: new_counter(
+                "job_assigned_total",
+                "Number of job assigned to searchers",
+                "search",
+                &[],
+            ),
+            job_assigned_to_affinity_searcher: new_counter(
+                "job_assigned_to_affinity_searcher",
```
fulmicoton, line 57: _total

fulmicoton, line 57:
```suggestion
                "job_assigned_to_affinity_searcher_total",
```

quickwit/quickwit-search/src/search_job_placer.rs
trinity-1686a: biased is the thing i propose, where instead of comparing `node0.load > node1.load`, we would do `node0.load > node1.load + bias`. I haven't done simulations with splits of varying size (I don't think we currently apply a weight to that). I will do more testing before actually doing any changes, this was just food for thoughts regarding how we could improve the chance of the same node being picked for a split, without impacting too much load balancing

fulmicoton: Ah sorry I had missed the comment. This is an interesting idea. 

I'm surprised by how low `pow2 hit` is. Can you share your simulation code?

fulmicoton (approved):

PR #5000, opened 2024-05-17
Warp rejection handling

fulmicoton:


fulmicoton:
@trinity-1686a The main problem is that we don't catch request and keep testing other route. 

The following make it work for that route, but all of error are probably wrong.

Issue #5003, opened 2024-05-18
Add support for `time_series_dimension` flag in `_elastic/{index}/_field_caps` API method


kuzaxak:
Opensearch Dashboards calling `_field_caps` to build an index pattern. To use quickwit natively as an external data source for it dashboards need to determine a timestamp field to render histogram, such fields listed on [index pattern settings page][2]. For that it relies on a `time_series_dimension` flag in `<field>.<type>.time_series_dimension`.

Link to the [doc][1].

**Describe the solution you'd like**

Filed chosen as `timestamp_field` in index should be marked as `time_series_dimension` in `_field_caps` output.

**Describe alternatives you've considered**

I've made a little proxy to query `/api/v1/indexes?index_id_patterns=` and patch response of `_field_caps` method.

**Additional context**

I'm working on integrating quickwit as an Opensearch Dashboards backend to replace existing setup which is using regular Opensearch.

[1]: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-field-caps.html#search-field-caps-api-response-body
[2]: https://opensearch.org/docs/latest/dashboards/management/index-patterns/

fulmicoton:
@kuzaxak have you started experimenting with adding opendashboard support by any chance? I'd love to exchange if you have more info on other missing pieces.

kuzaxak:
> @kuzaxak have you started experimenting with adding opendashboard support by any chance? I'd love to exchange if you have more info on other missing pieces.

Yes, I was able to migrate our Open search dashboard completely to quickwit backend. But I'm using a proxy to monkey patch some responses of quickwit.

I can share a proxy code, it is a simple python flask app.

kuzaxak:
@fulmicoton [here](https://gist.github.com/kuzaxak/ce6178c185073f69f9836c30e0ea08d9) you can find a proxy server code. 

I'm using quickwit as an external source not to deal with auth, dashboards storage, etc.

1. Enable this feature for OSD:
```
data_source.enabled: true
```

2. Add proxy as an additional datasource:
<img width="1217" alt="image" src="https://github.com/quickwit-oss/quickwit/assets/5330267/14be19be-2e2f-43fc-ade4-96bf991e36b8">

3. Add Index pattern using external source:
<img width="1217" alt="image" src="https://github.com/quickwit-oss/quickwit/assets/5330267/e2775c73-a9e1-4a05-9efe-53fb5984d9e5">

[...]

PR #5004, opened 2024-05-18
Add time_series_dimension property into _field_caps API method

kuzaxak:
### Description

`time_series_dimension` determine whether field will be used by Opensearch Dashboards to build histogram or not.

Fixing issue: https://github.com/quickwit-oss/quickwit/issues/5003

### How was this PR tested?

Added test to cover my change.

### DIFF
--- a/quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
+++ b/quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
@@ -112,11 +112,13 @@ pub fn es_compat_search_handler(
 /// GET or POST _elastic/{index}/_field_caps
 pub fn es_compat_index_field_capabilities_handler(
     search_service: Arc<dyn SearchService>,
+    metastore_service: MetastoreServiceClient,
 ) -> impl Filter<Extract = (impl warp::Reply,), Error = Rejection> + Clone {
     elastic_index_field_capabilities_filter()
         .or(elastic_field_capabilities_filter())
         .unify()
         .and(with_arg(search_service))
+        .and(with_arg(metastore_service))
         .then(es_compat_index_field_capabilities)
         .map(|result| make_elastic_api_response(result, BodyFormat::default()))
 }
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
```diff
@@ -547,13 +549,16 @@ async fn es_compat_index_field_capabilities(
     search_params: FieldCapabilityQueryParams,
     search_body: FieldCapabilityRequestBody,
     search_service: Arc<dyn SearchService>,
+    mut metastore: MetastoreServiceClient,
 ) -> Result<FieldCapabilityResponse, ElasticsearchError> {
+    let indexes_metadata = resolve_index_patterns(&index_id_patterns, &mut metastore).await?;
```
fulmicoton, line 554:
```suggestion
    let indexes_metadata = resolve_index_patterns(&index_id_patterns, &mut metastore).await?;
```

I'm really not fond of doubling the calls to `resolve_index_patterns`.
We happen to do one in `root_list_fields` too.

It seems like we pre-build /cache the `ListFieldsEntryResponse` in the packager (and indexing time) so that `root_list_fields(..)` is a light operation.

It would make sense to add the bool property there, in the `ListFieldsEntryResponse`.

```
message ListFieldsEntryResponse {
  ...
  // True means that the field is the timestamp field of all of the requested indexes.
  bool is_timestamp = 8;
}
```

We then need to know how and when we populate it.


Solution 1:

The logical place would be the packager.
Pros: that's where we build the rest. 
Cons: Right now the packager does not have access to the docmapper and does not know what the timestamp field is. We would have to pass that information. It is reasonable however. 
Cons2: the data is not available for existing data.
-----------------

Solution 2:
Keep it set to false in the packager, and actually compute it in the root_list_fields function.
At this point we have access to the index_metadatas already.
Pros: easy
Cons: ugly.

fulmicoton (changes requested):
See comment inline. I don;'t think we want an extra call to the metastore here.

kuzaxak:
> See comment inline. I don;'t think we want an extra call to the metastore here.

Clear, will have a loot on weekend into the architecture of the service more deeply. Thanks for suggested solutions, will try to implement the nice one.

kuzaxak:
Not needed in reality, missed ES and OpenSearch spec

PR #5006, opened 2024-05-20
Adds recover_fn to many routes.

fulmicoton:
If a route matches, we want to run its handler, and upon failure, the last thing we want is for another overlapping route to be executed as a fallback.

To avoid this, we call recover on each individual routes that have no overlap with routes of lesser priority.

### DIFF
--- a/quickwit/quickwit-serve/src/otlp_api/rest_handler.rs
+++ b/quickwit/quickwit-serve/src/otlp_api/rest_handler.rs
@@ -45,10 +45,9 @@ pub(crate) fn otlp_ingest_api_handlers(
     otlp_traces_service: Option<OtlpGrpcTracesService>,
 ) -> impl Filter<Extract = (impl warp::Reply,), Error = Rejection> + Clone {
     otlp_default_logs_handler(otlp_logs_service.clone())
-        .or(otlp_default_traces_handler(otlp_traces_service.clone()))
-        .or(otlp_logs_handler(otlp_logs_service))
-        .or(otlp_ingest_traces_handler(otlp_traces_service))
-        .recover(recover_fn)
+        .or(otlp_default_traces_handler(otlp_traces_service.clone()).recover(recover_fn))
+        .or(otlp_logs_handler(otlp_logs_service).recover(recover_fn))
+        .or(otlp_ingest_traces_handler(otlp_traces_service).recover(recover_fn))
 }
 
 pub(crate) fn otlp_default_logs_handler(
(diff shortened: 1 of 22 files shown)


quickwit/quickwit-serve/src/otlp_api/rest_handler.rs
```diff
@@ -47,6 +48,7 @@ pub(crate) fn otlp_ingest_api_handlers(
         .or(otlp_default_traces_handler(otlp_traces_service.clone()))
         .or(otlp_logs_handler(otlp_logs_service))
         .or(otlp_ingest_traces_handler(otlp_traces_service))
+        .recover(recover_fn)
```
trinity-1686a, line 51: not related to this PR: this group of routes isn't present in the openapi definition

trinity-1686a, line 51: the otlp routes could be handled individually, they don't look to me like they overlap

quickwit/quickwit-serve/src/template_api/rest_handler.rs
```diff
@@ -55,6 +56,7 @@ pub(crate) fn index_template_api_handlers(
         .or(update_index_template_handler(metastore.clone()))
         .or(delete_index_template_handler(metastore.clone()))
         .or(list_index_templates_handler(metastore.clone()))
+        .recover(recover_fn)
 }
 
 fn create_index_template_handler(
```
trinity-1686a, line 62: this and `list_index_templates_handler` might be missing a `warp::path::end` ?

trinity-1686a (commented):
I think the following routes could have a recover_fn too:
```
es_compat_search_handler
es_compat_index_field_capabilities_handler
es_compat_index_search_handler
es_compat_index_count_handler
es_compat_index_multi_search_handler // also its comment uses the wrong route, should be _msearch
es_compat_scroll_handler
readiness_handler
```

quickwit/quickwit-serve/src/template_api/rest_handler.rs
fulmicoton: It is implicit in the warp::path! macro.

> The path! macro automatically assumes the path should include an end() filter. To build up a path filter prefix, such that the end() isn’t included, use the / .. syntax.

quickwit/quickwit-serve/src/otlp_api/rest_handler.rs
fulmicoton: agreed. I moved it inside.

fulmicoton: added routes to openapi..

trinity-1686a (approved):

PR #5012, opened 2024-05-21
Removing one task per call layer on control plane.

fulmicoton:
They are useless. The control plane is already an actor.


quickwit/quickwit-serve/src/lib.rs
```diff
@@ -331,20 +331,9 @@ async fn start_control_plane_if_needed(
...
         let control_plane_client = ControlPlaneServiceClient::tower()
-            .stack_layer(shared_layers)
-            .stack_create_index_layer(OneTaskPerCallLayer)
-            .stack_delete_index_layer(OneTaskPerCallLayer)
-            .stack_add_source_layer(OneTaskPerCallLayer)
-            .stack_toggle_source_layer(OneTaskPerCallLayer)
-            .stack_delete_source_layer(OneTaskPerCallLayer)
-            .stack_get_or_create_open_shards_layer(OneTaskPerCallLayer)
+            .stack_layer(CP_GRPC_SERVER_METRICS_LAYER.clone())
+            .stack_layer(LoadShedLayer::new(100))
```
guilload, line 336: Now that we have optimized a bunch of metastore queries, I wonder whether we should bump up that value. Thoughts?

guilload (approved):

quickwit/quickwit-serve/src/lib.rs
fulmicoton: The load shedder? I think 100 is plenty for the control plane.

guilload: Maybe not at midnight UTC when 3000+ new shards need to be opened.

Issue #5032, opened 2024-05-25
 Add optimization for pure count and count aggregation


fulmicoton:
as observed on airmail

We are slower than Opensearch on

- `_count` on AllQuery
- `_count` queries on termqueries
- aggregation on `AllQuery` like the one as follows

```
{
    size: 0
  , aggs: {
      hosts: {
        terms: {
          field: '_host'
        , size: 50
        }
      }
    , apps: {
        terms: {
          field: '_app'
        , size: 50
        }
      }
    , ingesters: {
        terms: {
          field: '_ingester'
        , size: 50
        }
      }
    , tags: {
        terms: {
          field: '_tag'
        , size: 50
        }
      }
    }
  }
  ```

All of the above can be addressed just looking at the split metadata, and/or the termdictionary.

**Comment by fulmicoton** [comment]
@PSeitz can you take over this issue.

You can ditch or use #5034 if you want.
For the last part: 'term aggregations on all query', it would be nice to detect that this can be addressed via the termdictionary alone..

Problem 1) 
It will require creative software engineering between tantivy and quickwit as the warm up is totally different.

Problem 2)
multiplicity of the same value will actually cause different different results if we have multivalued columns (and the same value can appear more than once)

Problem 3)
Technicallly, doing it using the column is faster if the term dictionary has a lot of distinct values, and the size of the aggregation is limited. In practise, we can probably just do a simple rule on the size of the term dictionary.
(e.g.: if term dictionary < 500KB or < column).


========

Other optim opportunity
it would be nice to have an optim on aggregation over columns for all queries.

We could directly operate over values, instead of filling a doc id buffer, and fetching those.

**Comment by PSeitz** [comment]
## Count All

`_count` with AllQuery **works**.  Uses only metadata.

## Count All + Top N

`_count` with AllQuery + top_n: Does a full search currently.

To support `_count` + top n, we could go for a hybrid approach and get the counts from metadata split and the top n from a selected split.
Question would be which top n we get should get for the `AllQuery`

## Count All + Top N + Sort By Date

This optimization could also be extended
`_count` with AllQuery + top_n, sort by timestamp.

Identify splits which can't contain data from the top n, with the time_range metadata:
```rust
/// If a timestamp field is available, the min / max timestamp in
/// the split, expressed in seconds.
pub time_range: Option<RangeInclusive<i64>>,
```

## Count All + Top N + Sort By Other Field

We don't have metadata for other fields on the split metadata, so there are two options:
[...]

PR #5034, opened 2024-05-26
Added a count-only optimisation to QuickwitCollector

fulmicoton:
Related to #5032


PSeitz:
I think we can reuse the same code we use for the `_field_caps` api.
The check in `is_metadata_count_request` is probably returning false but shouldn't

fulmicoton:
@PSeitz can we close this? Are all of the optimization suggested in this PR shipped in other PRs?

PSeitz:
Everything except the term query count special case, which I'm not sure we need

PR #5038, opened 2024-05-28
Add behavior version to Kinesis client builder

rdettai:
### Description

`BehaviorVersion` is now required when building any AWS client's config

### How was this PR tested?


rdettai:
@trinity-1686a looks like this slipped through during the last SDK upgrade because all tests on the Kinesis SDK are disabled

trinity-1686a (approved):

Issue #5010, opened 2024-05-21
Change the algorithm that place split search on searcher


fulmicoton:
Right now we  use the following logic.

```rust
pub async fn assign_jobs<J: Job>(
        &self,
        mut jobs: Vec<J>,
        excluded_addrs: &HashSet<SocketAddr>,
    ) -> anyhow::Result<impl Iterator<Item = (SearchServiceClient, Vec<J>)>> {
        let num_nodes = self.searcher_pool.len();

        let mut candidate_nodes: Vec<CandidateNodes> = self
            .searcher_pool
            .pairs()
            .into_iter()
            .filter(|(grpc_addr, _)| {
                excluded_addrs.is_empty()
                    || excluded_addrs.len() == num_nodes
                    || !excluded_addrs.contains(grpc_addr)
            })
            .map(|(grpc_addr, client)| CandidateNodes {
                grpc_addr,
                client,
                load: 0,
            })
            .collect();

        if candidate_nodes.is_empty() {
            bail!(
                "failed to assign search jobs. there are no available searcher nodes in the pool"
            );
        }
        jobs.sort_unstable_by(Job::compare_cost);

        let mut job_assignments: HashMap<SocketAddr, (SearchServiceClient, Vec<J>)> =
            HashMap::with_capacity(num_nodes);

        for job in jobs {
            sort_by_rendez_vous_hash(&mut candidate_nodes, job.split_id());
            // Select the least loaded node.
            let chosen_node_idx = if candidate_nodes.len() >= 2 {
                usize::from(candidate_nodes[0].load > candidate_nodes[1].load)
            } else {
                0
            };
            let chosen_node = &mut candidate_nodes[chosen_node_idx];
            chosen_node.load += job.cost();

            job_assignments
                .entry(chosen_node.grpc_addr)
                .or_insert_with(|| (chosen_node.client.clone(), Vec::new()))
                .1
                .push(job);
        }
        Ok(job_assignments.into_values())
```

This is rather suboptimal. I think we can come up with something considerably more efficient. For instance, @fmassot came up with the following logic:
```rust
let compute_target_load = total_load / num_searchers + SOME_MARGIN;
for job in jobs {
    sort_by_rendez_vous_hash(&mut candidate_nodes, job.split_id());
    let chosen_node_idx = candidates.iter().find(|candidate| candidate.load < compute_target_load).unwrap();
    let chosen_node = &mut candidate_nodes[chosen_node_idx];
[...]

PR #5051, opened 2024-05-30
improve placing algorithm

trinity-1686a:
### Description

fix https://github.com/quickwit-oss/quickwit/issues/5010
improve the job placing algorithm, to increase the chance of hitting the searcher with the highest affinity.
Also add a primitive cost evaluation of splits, based only on the count of documentation.

### How was this PR tested?

updated tests
added tests with many splits
tested manually with #4998 applied to verify the number of jobs run with affinity=0 improved (it did)

i had to modify some things to keep some root_search tests working, as now it is possible that searching 2 splits with 2 searchers ends up querying only one searcher.
(with N the job cost, assuming all jobs equals,`target_load = ⌈2 * N * (1+ ε) / 2⌉ = N+1`, after the 1st split, load is only N so the 2nd split can use the same searcher).
I've documented multiple fixes, and implemented the one that seemed most sensible to me


quickwit/quickwit-search/src/root.rs
```diff
@@ -1393,9 +1393,10 @@ async fn assign_client_fetch_docs_jobs(
 }
 
 // Measure the cost associated to searching in a given split metadata.
-fn compute_split_cost(_split_metadata: &SplitMetadata) -> usize {
+fn compute_split_cost(split_metadata: &SplitMetadata) -> usize {
     // TODO: Have a smarter cost, by smoothing the number of docs.
-    1
+    // +1 because a split of 1 doc would have an ilog2 of 0 otherwise
+    split_metadata.num_docs.ilog2() as usize + 1
```
fulmicoton, line 1399: 2 points...

First point: 
I know `num_docs` is not supposed to be 0 today, but there is really no reason to not be a tiny bit defensive here and make sure for `i = 0`.
The above would be better as :
`split_metadata.num_docs.max(2).ilog2() as usize`

Second point:
I don't the function is good here.
For instance, on both side of the range it behaves wrongly.
In particular, I feel like:
- a split of size 1's cost is way too low compared to a split of size 1024.
- a split of size 10M cost is way too low compared to a split of size 10M.

Of course, as we discussed offline, this is very dependant on the query type and what is in cache or not. It feels uncomfortable to even think about a one size fits all solution.

(it almost make sense to prefer diversity in size for all of the leaf in the algorithm...)

I suggest we do something like 
`K + num_docs / 100_000` ?

With some value of K somewhere between 1 and 5.

quickwit/quickwit-search/src/search_job_placer.rs
```diff
@@ -177,13 +178,30 @@ impl SearchJobPlacer {
...
             sort_by_rendez_vous_hash(&mut candidate_nodes, job.split_id());
-            // Select the least loaded node.
-            let chosen_node_idx = if candidate_nodes.len() >= 2 {
-                usize::from(candidate_nodes[0].load > candidate_nodes[1].load)
-            } else {
-                0
+            let mut candidate_nodes_iter = candidate_nodes.iter().enumerate();
+            // Select a node which hasn't reached the target load
+            let chosen_node_idx = loop {
+                let Some((id, candidate_node)) = candidate_nodes_iter.next() else {
```
fulmicoton, line 197: I think `.iter_mut().find(||)` would be clearer than this loop with break?


```
let Some(chosen_node) = candidate_nodes.iter_mut().find(|candidate| **candidate.load < target_load) else { ... };

**Review by fulmicoton — CHANGES_REQUESTED** [review]
see comments

File: quickwit/quickwit-search/src/search_job_placer.rs [hunk 3]
```diff
@@ -177,13 +178,30 @@ impl SearchJobPlacer {
...
+        let total_load: usize = jobs.iter().map(|job| job.cost()).sum();
+
+        // allow arround 1% disparity. Round up so we never end up in a case where
+        // target_load * num_nodes < total_load
+        // some of our tests needs 2 splits to be put on 2 different searchers. It makes sens for
+        // these tests to keep doing so (testing root merge). Either we can make the allowed
+        // difference stricter, find the right split names ("split6" instead of "split2" works).
+        // or modify mock_split_meta() so that not all splits have the same job cost
+        // for now i went with the mock_split_meta() changes.
+        const ALLOWED_DIFFERENCE: usize = 101;
```
fulmicoton, line 190: 1% is a. bit tight? Is this coming from your simulations?

```suggestion
        const ALLOWED_DIFFERENCE: usize = 101;
```

quickwit/quickwit-search/src/search_job_placer.rs
```diff
@@ -177,13 +178,30 @@ impl SearchJobPlacer {
...
+        let target_load = (total_load * ALLOWED_DIFFERENCE).div_ceil(num_nodes * 100);
```
fulmicoton, line 191: note this is not sufficient to make sure you can always pick a node.

I think the following works:

(total_load * ALLOWED_DIFFERENCE).div_ceil(num_nodes * 100).max(max_job_load + total_load.ceil_div(num_nodes));

fulmicoton: We know $J <= J_{max}$ by definition, so this implies.

$\forall n < N, L_n > {T-{J_{max}}} \over {N}$

fulmicoton: We know that $\sum L_n <= T - {J_max}$  (otherwise the algorithm would have terminated).
Therefore this in turn imply,

$T - J_{max} > T - J_{max}$, which is a contradiction.

fulmicoton: Do we need to use necessarily this bound? No not really, but we are certain this is "good enough" which is a comfortable guarantee. 

Is the current solution good enough? Actually no it is not.
Think about the case where we have K large splits with the same node affinity, and S searchers, with K << S.

quickwit/quickwit-search/src/search_job_placer.rs
trinity-1686a: it seems to be good when splits are roughly equal, but a bit low when there are wide variation. I'm going to increase it a bit

quickwit/quickwit-search/src/search_job_placer.rs
trinity-1686a: after discussion a bit, the current implementation actually seems to work, because it checks `current_load < target_load`, and not `current_load + job_weight < target_load`

fulmicoton (approved):

PR #5055, opened 2024-05-31
Fixing the oneshot compilation problem.

fulmicoton:
Relying on a hotfix tag of tantivy.


PSeitz:
I think this should also work
https://github.com/quickwit-oss/quickwit/pull/5054

fulmicoton:
I merged #5054 so we can close this.

PR #5057, opened 2024-05-31
Adding some detail on how to set the number of pipelines

fulmicoton:


fmassot:
Ideas for a follow-up PR:
- talked about the caveats in the case where someone has 2 kafka datasources with very different indexing throughput
- I think we should have a dedicated page to explain how to size a quickwit cluster with a kafka datasource like here: https://quickwit.io/docs/deployment/cluster-sizing

PR #5075, opened 2024-06-04
optimize topn requests

PSeitz:
add logic to detect which splits will deliver the top n results for
requests. This is only supported for match_all requests, with optional
sort_by on timestamp sorting.

The change extends the python tests to distribute ndjson to random splits

start_timestamp, end_timestamp as well as a filter on the timestamp field
is not supported currently but could be.
search_after is also not supported currently

https://qw-benchmarks.104.155.161.122.nip.io/?run_ids=1861,1862&search_metric=engine_duration

Compare S3 Fetch Requests:
https://qw-benchmarks.104.155.161.122.nip.io/?run_ids=1799,1864&search_metric=object_storage_fetch_requests

Addresses https://github.com/quickwit-oss/quickwit/issues/5032


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -879,6 +877,130 @@ impl CanSplitDoBetter {
...
 
+    /// This function tries to detect upfront which splits contain the top n hits and convert other
+    /// split searches to count only searches. It also optimizes split order.
+    ///
+    /// Returns the search_requests with their split.
+    fn optimize(
+        &self,
+        request: Arc<SearchRequest>,
+        mut splits: Vec<SplitIdAndFooterOffsets>,
+    ) -> Result<Vec<(SplitIdAndFooterOffsets, SearchRequest)>, SearchError> {
```
fulmicoton, line 888: In most queries we end up not modifying the request. It feels silly to clone the object so many times when we in fact have an Arc at disposal.

```suggestion
    ) -> Result<Vec<(SplitIdAndFooterOffsets, Arc<SearchRequest>)>, SearchError> {
```

PSeitz: Yes, but the behavior was the same before, but in the loop instead

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -879,6 +877,130 @@ impl CanSplitDoBetter {
...
+                        need_more_docs
+                    })
+                    .count()
+            };
+
+        // reuse the detected sort order in split_filter
+        // we want to detect cases where we can convert some split queries to count only queries
+        let num_requested_docs = request.start_offset + request.max_hits;
+        match self {
+            CanSplitDoBetter::SplitIdHigher(_) => {
```
fulmicoton, line 930: why do we apply the logic on `CanSplitDoBetter` instead of the actual sort order here?
Can't we have CanSplitDoBetter set to informative even though docs are requested to be ordered by timestamp?

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -879,6 +877,130 @@ impl CanSplitDoBetter {
...
+            .map_err(|err| SearchError::InvalidQuery(err.to_string()))?;
+
+        // TODO: Update the logic to handle start_timestamp end_timestamp ranges
+        if request.start_timestamp.is_some() || request.end_timestamp.is_some() {
+            return Ok(split_with_req);
+        }
+
+        if !matches!(query_ast, QueryAst::MatchAll) {
+            return Ok(split_with_req);
+        }
```
fulmicoton, line 909: this code is needlessly complicated.

I suggest we add a function outside of this 

```
/// Returns true if the query is a all query, with no range query,
/// nor search_after, nor aggregation.
fn is_simple_all_query(search_request: &SearchRequest) -> bool {
    if search_request.aggregation_request.is_some() {
        return false;
    }

   if search_request.search_after.is_some() {
       return false;
   }

    // TODO: Update the logic to handle start_timestamp end_timestamp ranges
    if search_request.start_timestamp.is_some() || search_request.end_timestamp.is_some() {
        return false;
    }

    let Ok(query_ast) = serde_json::from_str(&search_request.query_ast) else {
        return false;
    };

    matches!(query_ast, QueryAst::MatchAll)
}
```

We can then replace this block by

```
        if !is_simple_all_query(&*request) {
            // no optimization opportunity here.
            return Ok(splits
                .into_iter()
                .map(|split| (split, (*request).clone()))
                .collect::<Vec<_>>());
        }
```

The `count_required_splits` can then operate directly on `splits` rather than `split_with_req`.

quickwit/quickwit-search/src/leaf.rs
fulmicoton: ah answering my own question. if sorted by timestmap, but we don't have a bound, then `CanSplitDoBetter` is set to Higher or Lower but with `None`

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -879,6 +877,130 @@ impl CanSplitDoBetter {
...
+                // We order by timestamp asc. split_with_req is sorted by timestamp_start.
+                //
+                // Calculate the number of splits which are guaranteed to deliver enough documents.
+                let num_splits = count_required_splits(&split_with_req, num_requested_docs);
+                assert!(
+                    num_splits > 0,
+                    "We should always have at least one split to search"
+                );
+                //
+                // If we know that some splits will deliver enough documents, we can convert the
```
fulmicoton, line 950: great explanation

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -879,6 +877,130 @@ impl CanSplitDoBetter {
...
+        if request.start_timestamp.is_some() || request.end_timestamp.is_some() {
+            return Ok(split_with_req);
+        }
+
+        if !matches!(query_ast, QueryAst::MatchAll) {
+            return Ok(split_with_req);
+        }
+
+        // Count the number of splits which contain enough documents
+        let count_required_splits =
```
fulmicoton, line 912: I think we can afford computing it all of the time here and not introduce a closure.

also (but that's a nitpick)... 

I would prefer to have either an iterator style with `scan`:

```rust
        // Count the number of splits which contain enough documents
        let min_required_splits =
            splits.iter()
                .map(|split| split.num_docs as u64)
                // computing the partial sum
                .scan(0u64, |partial_sum: &mut u64, val| {
                    *partial_sum += val;
                    Some(*partial_sum)
                })
                .take_while(|cumulated_count| *cumulated_count < num_requested_docs)
                .count() + 1;
```

or a loop style

```rust
     // Count the number of splits which contain enough documents
        let min_required_splits = {
            let mut cumsum = 0;
            let mut count = 0;
            for split in splits {
                count += 1;
                cumsum += split.num_docs as u64;
                if cumsum >= num_requested_docs {
                    break;
                }
            }
            count
        };
```

The iterator style that mutates stuff captures by the closure is a bit tricky.
That part is a nitpick so feel free to ignore.

fulmicoton (changes requested):
see comments

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -879,6 +877,130 @@ impl CanSplitDoBetter {
...
+                // Let's get the smallest timestamp_start of the first num_splits splits
+                let smallest_start_timestamp = split_with_req
+                    .iter()
+                    .take(num_splits)
+                    .map(|(split, _)| split.timestamp_start())
+                    .min()
+                    .unwrap();
+                for (split, ref mut request) in split_with_req.iter_mut().skip(num_splits) {
+                    if split.timestamp_end() < smallest_start_timestamp {
+                        disable_search_request_hits(request);
```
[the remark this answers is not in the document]
PSeitz: strangely currently we don't have a no count parameter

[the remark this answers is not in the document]
fulmicoton: You already have an CountHits enum. We could add that or use the Option?

quickwit/quickwit-search/src/leaf.rs
PSeitz: yes that's better

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -879,6 +877,130 @@ impl CanSplitDoBetter {
...
+                    disable_search_request_hits(request);
+                }
+            }
+            CanSplitDoBetter::Uninformative => {}
+            CanSplitDoBetter::SplitTimestampLower(_) => {
+                // We order by timestamp asc. split_with_req is sorted by timestamp_start.
+                //
+                // Calculate the number of splits which are guaranteed to deliver enough documents.
+                let num_splits = count_required_splits(&split_with_req, num_requested_docs);
+                assert!(
```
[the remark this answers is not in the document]
PSeitz: It was there to safeguard the code below for changes later, but I changed the algorithm to handle num_splits=0

quickwit/quickwit-search/src/leaf.rs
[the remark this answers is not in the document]
PSeitz: > if we don't request count, the resulting disable_search_request does nothing.

No, it disables the returning of hits from that split. That leaves only the count, and in this simple query case the count can be served from metadata and is basically free. 
Initially we don't know which split may contain the best splits so we return hits from all splits.

> You already have an CountHits enum. We could add that or use the Option?

Yes, but that would be relevant for cases where we can't serve counts from metadata. But in that case we have already the `run_all_splits` optimization, which means we probably run num_cpu splits before early exiting.

fulmicoton (approved):

PR #5081, opened 2024-06-05
enable pprof cpu profiling

PSeitz:
This adds two routes to control pprof
Enable CPU profiling (100hz)
http://localhost:7280/pprof/start
Get the profile as flamegraph
http://localhost:7280/pprof/flamegraph

The routes are behind the `pprof` feature flag

`cargo install --locked --path . --features pprof`

### DIFF
--- a/quickwit/quickwit-serve/src/pprof.rs
+++ b/quickwit/quickwit-serve/src/pprof.rs
@@ -19,8 +19,8 @@
 
 use warp::Filter;
 
-/// pprof/start to start cpu profiling
-/// pprof/stop to stop cpu profiling and return a flamegraph
+/// pprof/start disabled
+/// pprof/flamegraph disabled
 #[cfg(not(feature = "pprof"))]
 pub fn pprof_routes() -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
     let start_profiler = {
@@ -33,7 +33,7 @@ pub fn pprof_routes() -> impl Filter<Extract = impl warp::Reply, Error = warp::R
     };
 
     let stop_profiler = {
-        warp::path!("pprof" / "stop").map(move || {
+        warp::path!("pprof" / "flamegraph").map(move || {
             warp::reply::with_status(
                 "not compiled with pprof feature",
                 warp::http::StatusCode::BAD_REQUEST,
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-serve/src/pprof.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    let stop_profiler = {
+        let profiler_guard = Arc::clone(&profiler_guard);
+        warp::path!("pprof" / "stop").map(move || {
+            let profiler_guard = Arc::clone(&profiler_guard);
+            get_flamegraph(profiler_guard)
+        })
+    };
+
+    fn get_flamegraph(profiler_guard: Arc<Mutex<Option<ProfilerGuard>>>) -> impl warp::Reply {
```
fulmicoton, line 81: wouldn't the following have worked too?

```suggestion
    fn get_flamegraph(profiler_guard: &Mutex<Option<ProfilerGuard>>) -> impl warp::Reply {
```

quickwit/quickwit-serve/src/pprof.rs
```diff
@@ -0,0 +1,106 @@
...
+        let mut guard = profiler_guard.lock().unwrap();
+        if let Some(profiler) = guard.take() {
+            if let Ok(report) = profiler.report().build() {
+                let mut buffer = Vec::new();
+                if report.flamegraph(&mut buffer).is_ok() {
```
fulmicoton, line 86: do you know if generating the report is cpu heavy? if so we should probably run it on the thread pool for cpu intensive stuff.

`run_cpu_intensive` in `quickwit_common`

fulmicoton (changes requested):
That looks good for experimentations but this is quite the footgun. Are statistics gathered really bounded? 

What would happen if someone called start and never called stop?
Can we guard ourselves against that?

I assume your point is that it feels silly to hope for a GET request to run over 30s.
Could it be two endpoints like you did, but 

`pprof/run` would start profiling for 30s, and
`pprof/flamegraph` would return the last computed profile, or something like that?

In your test were 100Hz sufficient? Should we make the frequency and the duration configurable maybe?

quickwit/quickwit-serve/src/pprof.rs
PSeitz: in the old version yes, in the new version we need a static lifetime

quickwit/quickwit-serve/src/pprof.rs
PSeitz: yes they can be cpu intensive, I moved it to a new thread

quickwit/quickwit-serve/src/pprof.rs
```diff
@@ -0,0 +1,159 @@
...
+                warp::http::StatusCode::BAD_REQUEST,
+            )
+            .into_response())
+        }
+    }
+
+    async fn save_flamegraph(
+        profiler_state: Arc<Mutex<ProfilerState>>,
+    ) -> tokio::task::JoinHandle<()> {
+        spawn_blocking(move || {
```
fulmicoton, line 145: Technically it is better to use `quickwit_common:run_cpu_intensive`
because spawn blocking is an infinite pool.

That being said, here, I think we don't care much. 
Fix if you want.

quickwit/quickwit-serve/src/pprof.rs
```diff
@@ -0,0 +1,159 @@
...
+            ))
+        } else {
+            Ok(warp::reply::with_status(
+                "CPU profiling is already running",
+                warp::http::StatusCode::BAD_REQUEST,
+            ))
+        }
+    }
+
+    async fn stop_profiler_handler(
```
fulmicoton, line 125: rename to `flamegraph_handler`?

fulmicoton (approved):

PSeitz:
I didn't see memory issues while collecting, but there's a performance hit usually, so we should avoid to run it indefinitely by accident.

100hz is usually fine, except very short running queries.

guilload:
I'm going to move this endpoint to `api/developer/pprof*`.

PR #5087, opened 2024-06-05
Add doc mapping UID attribute to `DocMapping`

guilload:
### Description
Per title.

### How was this PR tested?
Fixed and added backward compatibility tests.

### DIFF
--- a/quickwit/quickwit-config/src/index_config/serialize.rs
+++ b/quickwit/quickwit-config/src/index_config/serialize.rs
@@ -80,7 +80,7 @@ pub fn load_index_config_update(
     let current_index_parent_dir = &current_index_config
         .index_uri
         .parent()
-        .context("Unexpected `index_uri` format on current configuration")?;
+        .expect("index URI should have a parent");
     let new_index_config = load_index_config_from_user_config(
         config_format,
         index_config_bytes,
@@ -99,7 +99,9 @@ pub fn load_index_config_update(
         new_index_config.index_uri
     );
     ensure!(
-        current_index_config.doc_mapping == new_index_config.doc_mapping,
+        current_index_config
+            .doc_mapping
+            .eq_ignore_doc_mapping_uid(&new_index_config.doc_mapping),
         "`doc_mapping` cannot be updated"
     );
     Ok(new_index_config)
--- a/quickwit/quickwit-config/src/index_config/serialize.rs
+++ b/quickwit/quickwit-config/src/index_config/serialize.rs
@@ -19,7 +19,7 @@
 
 use anyhow::{ensure, Context};
 use quickwit_common::uri::Uri;
-use quickwit_proto::types::IndexId;
+use quickwit_proto::types::{DocMappingUid, IndexId};
 use serde::{Deserialize, Serialize};
 use tracing::info;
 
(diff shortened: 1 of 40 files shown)


quickwit/quickwit-doc-mapper/src/doc_mapping.rs
```diff
@@ -99,6 +100,10 @@ impl Default for Mode {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, utoipa::ToSchema)]
 #[serde(deny_unknown_fields)]
 pub struct DocMapping {
+    /// UID of the doc mapping.
+    #[serde(default = "DocMappingUid::new")]
```
guilload, line 104: This is going to cause an issue when deserializing doc mapping without an UID (we want to them to be 0). I'm working on a fix.

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -1681,10 +1681,12 @@ mod tests {
             .unwrap();
         assert!(indexing_tasks.is_empty());
 
-        let results: Vec<ApplyIndexingPlanRequest> =
-            client_inbox.drain_for_test_typed::<ApplyIndexingPlanRequest>();
-        assert_eq!(results.len(), 1);
-        assert!(results[0].indexing_tasks.is_empty());
+        let apply_plan_requests = client_inbox.drain_for_test_typed::<ApplyIndexingPlanRequest>();
+        assert!(!apply_plan_requests.is_empty());
```
trinity-1686a, line 1685: was the test too strict in some instances?

quickwit/quickwit-doc-mapper/src/doc_mapping.rs
```diff
@@ -99,6 +100,13 @@ impl Default for Mode {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize, utoipa::ToSchema)]
 #[serde(deny_unknown_fields)]
 pub struct DocMapping {
+    /// Doc mapping UID.
+    ///
+    /// Splits with the same doc mapping UID share the same schema and should use the same doc
+    /// mapper during indexing and querying.
```
trinity-1686a, line 106: am i right understanding that splits actually do not contain a doc mapping uid yet?

trinity-1686a (approved):
lgtm, just small questions

quickwit/quickwit-doc-mapper/src/doc_mapping.rs
guilload: Yes

quickwit/quickwit-control-plane/src/control_plane.rs
guilload: Yes, sometimes the actor has time to spin more than once, and two messages are emitted, which causes the test to be flaky.

quickwit/quickwit-config/src/index_config/serialize.rs
```diff
@@ -75,7 +80,7 @@ pub fn load_index_config_update(
     let current_index_parent_dir = &current_index_config
         .index_uri
         .parent()
-        .context("Unexpected `index_uri` format on current configuration")?;
+        .expect("index URI should have a parent");
```
rdettai, line 83: we should not panic here

PR #5094, opened 2024-06-07
docs: Vector fixes

jpds:
### Description

Without this change, Quickwit returns:

```
WARN quickwit_indexing::actors::doc_processor: doc mapper parse error: the document contains a field that is not declared in the schema: "service" index_id="otel-logs-v0_7" source_id="_ingest-api-source"
```

### How was this PR tested?

Integrating into the NixOS integration tests: https://github.com/NixOS/nixpkgs/blob/master/nixos/tests/vector/syslog-quickwit.nix


fulmicoton:
thank you!

PR #5095, opened 2024-06-07
Take bool env var values into account instead of presence

guilload:
### Description
When the Airmail team modifies boolean environment variables, I don't know whether they actually remove them or update the values, so I'd be more comfortable shipping that PR before the next Airmail releases.

### How was this PR tested?
- Added unit test

### DIFF
--- a/quickwit/Cargo.lock
+++ b/quickwit/Cargo.lock
@@ -6521,6 +6521,7 @@ dependencies = [
  "hostname",
  "md5",
  "once_cell",
+ "quickwit-common",
  "reqwest",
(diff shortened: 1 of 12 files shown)


trinity-1686a (approved):
i went through all `std::env::var` calls i could find. These 3 (lambda related) still parse boolean directly: `ENABLE_VERBOSE_JSON_LOGS` `DISABLE_MERGE` `DISABLE_JANITOR`

guilload:
Well spotted. I will fix those and merge the PR.

PR #5096, opened 2024-06-08
Add ES Resolve index API handler

kuzaxak:
### Description

This [method][1] is used by Opensearch Dashboards when user typing index
 name in pattern creation UI.

Case that is not handled within this change is `*:*` pattern which should be converted to `*` as QuickWit do not support `:` in the index name.

[1]: https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-resolve-index-api.html

### How was this PR tested?

Tested on own cluster. Couldn't figure out how to add integration tests to cover ES functionality.

image
image


kuzaxak:
Recreated to feature branch [here](https://github.com/quickwit-oss/quickwit/pull/5113)

Issue #1065, opened 2022-01-11
FileSourceParams validation mess


fulmicoton:
Right now the model for the FileSourceParams allows for an empty filepath. This 
is interpreted as StdIn.

This piggy back is useful in the ingest CLI, but harmful when we deserialize a config.
We have a bunch of checks to void this.

We should probably make this Option a filepath, and 
create a specific source type for StdIn. We also want to somehow prevent the serialization of this source type.

# Solution?

We could introduce an extra `SourceParams` called stdin.
We can prevent its deserialization, serialization or not (the story holds for Void as well).

The StdIn filesource would get a different factory from the file source, but that new factory would produce the same source type.

fulmicoton:
Move to quickwit 0.5

rdettai:
Note: currently the validation is happening in 2 places:
- https://github.com/quickwit-oss/quickwit/blob/97889f61dbc3f72ce468506870792f833eeb482b/quickwit/quickwit-serve/src/index_api/rest_handler.rs#L710-L716
- https://github.com/quickwit-oss/quickwit/blob/97889f61dbc3f72ce468506870792f833eeb482b/quickwit/quickwit-config/src/source_config/serialize.rs#L82-L89

PR #5148, opened 2024-06-21
Add File source with SQS notifications

rdettai:
### Description

This PR proposes the generic implementation of a "queue" source. For now, only an implementation for AWS SQS with its data backed by AWS S3 is exposed to the users. Google Pubsub as the queue implementation or inlined data (i.e messages containing the data itself and not the link to the object store) will come next.

We use the shard API to provide deduplication of messages. For the current implementation where the source data is stored on S3, the deduplication is made on the object URI.

High level summary of the abstractions that are part of the generic implementation:
- `Processor` exposes the exact same methods as the `Source` trait but does not implement it directly. Instead, the concrete queue sources (e.g. `SqsSource`) wrap the `Processor`.
- A pipeline of message states:
  - `RawMessage`: the message as received from the Queue
  - `PreProcessedPayload`: the message went through the minimal transformation to discover its partition id
  - `CheckpointedMessage`: the message was checked against the shared state (shard API), it is now ready to be processed
  - `InProgressMessage`: the message that is actively being read
- `QueueSharedState` is an abstraction over shard API. By calling `open_shard` upon reception of the messages we avoid costly redundant processing when receiving a duplicate message. 
- `QueueLocalState` represents the state machine of the messages as they are processed by the indexing pipeline
- `VisibilityTaskHandle` a task that extends the message visibility when required _(needs to be reworked)_

TODO:
- [x] Generic implementation of the queue source based on the `Processor` abstraction
- [x] SQS implementation of the `Queue` trait
- [x] Adapt the shard `open_shard` API to accept `publish_token` as a field. This gives upsert semantics to the API which makes it possible to acquire the shard upon creation.
- [x] Change the switching logic in the metastore implementation to allow any source to use the shard API for checkpointing (adding `SourceConfig.use_shard_api()`)
- [x] Unit tests the `Processor` abstraction
- [x] Unit test the `ShardState` abstraction 
- [x] LocalStack tests for the `SqsSource` (with some small refactoring to reuse the `setup_index` helper from the `KafkaSource`)
- [x] Implement the observable state updates
- [x] Change the source config to be a file source with SQS notification instead of an SQS source backed by a file.
- [x] Error handling (transient can be retried, bad messages should be discarded(?)...) (see [comment](https://github.com/quickwit-oss/quickwit/pull/5148#issuecomment-2188322448))
- [x] Deduplicate  within a received message batch
- [x] Resume the indexing of an unfinished file after a failure
- [x] Use 2x commit timeout as last visibility extension after InProgress status
- [x] End to end test the validate the queue behaviour after commiting the shard progress in the `Publisher` actor

TODO in subsequent PRs:
- GCP Pubsub (small)
- data within the queue payload (small)
- shard garbage collection (medium)
- improve the visibility extension logic (medium)

### How was this PR tested?

This PR contains unit tests and higher level tests that use LocalStack.


quickwit/quickwit-indexing/src/source/doc_file_reader.rs
```diff
@@ -48,8 +48,8 @@ impl SkipReader {
     }
 
     async fn skip(&mut self) -> io::Result<()> {
-        // Allocating 64KB once on the stack should be fine (<1% of the Linux stack size)
-        let mut buf = [0u8; 64000];
+        // allocate on the heap to avoid stack overflows
+        let mut buf = vec![0u8; 64000];
```
fulmicoton, line 52: I don't think this is helpful :-/

quickwit/quickwit-metastore/src/metastore/postgres/queries/shards/open.sql
```diff
@@ -1,5 +1,5 @@
-INSERT INTO shards(index_uid, source_id, shard_id, leader_id, follower_id)
-    VALUES ($1, $2, $3, $4, $5)
+INSERT INTO shards(index_uid, source_id, shard_id, leader_id, follower_id, publish_token)
+    VALUES ($1, $2, $3, $4, $5, $6)
```
fulmicoton, line 2: how is the publish token handled for ingest v2 today?

quickwit/quickwit-config/src/source_config/mod.rs
```diff
@@ -245,6 +248,22 @@ impl SourceParams {
     pub fn void() -> Self {
         Self::Void(VoidSourceParams)
     }
+
+    pub fn use_shard_api(&self) -> bool {
```
fulmicoton, line 252: This need some documentation explaining what this is about.

quickwit/quickwit-indexing/src/source/queue_sources/visibility.rs
```diff
@@ -0,0 +1,103 @@
...
+    ));
+    VisibilityTaskHandle {
+        _task_handle: task_handle,
+        ack_id,
+    }
+}
+
+/// This is still work in progress. Using the `PublishLock` isn't enough to
+/// communicate visibility extension failures:
+/// - we don't want to fail the pipeline if we fail to extend the visibility of a message that is
```
fulmicoton, line 64: if this is a rare enough event we don't care much.

quickwit/quickwit-indexing/src/source/queue_sources/visibility.rs
```diff
@@ -0,0 +1,103 @@
...
+///   still waiting for processing
+/// - the Processor must also be notified that it shouldn't process this message anymore
+async fn handle_visibility(
```
fulmicoton, line 67:
```suggestion
async fn handle_visibility_loop(
```

quickwit/quickwit-indexing/src/source/queue_sources/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+use std::fmt;
+use std::time::{Duration, Instant};
+
+use async_trait::async_trait;
+use message::RawMessage;
+
+/// The queue abstraction is based on the AWS SQS and Google Pubsub APIs. The
+/// only requirement of the underlying implementation is that messages exposed
+/// to a given consumer are hidden to other consumers for a configurable period
+/// of time. Retries are
```
fulmicoton, line 41: sentence unfinished.

quickwit/quickwit-indexing/src/source/queue_sources/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+#[async_trait]
+pub trait Queue: fmt::Debug + Send + Sync + 'static {
+    /// Poll the queue to receive messages.
+    ///
+    /// The implementation is in charge of choosing the wait strategy when there
+    /// are no messages in the queue. It will typically use long polling to do
+    /// this efficiently. On the other hand, when there is a message in the
```
fulmicoton, line 48: I didn't understand what was the meaning of the second setnence. If we don't long polling, we don't know when there is a message in the queue right away do we?

quickwit/quickwit-indexing/src/source/queue_sources/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// queue, it should be returned as quickly as possible.
+    async fn receive(&self) -> anyhow::Result<Vec<RawMessage>>;
+
+    /// Try to acknowledge the messages, effectively deleting them from the queue.
+    ///
+    /// The call might return `Ok(())` yet fail partially:
+    /// - if it's a transient failure? -> TODO check
```
fulmicoton, line 55: why would we return `Ok(())` for a transient failure?

quickwit/quickwit-indexing/src/source/queue_sources/processor.rs
```diff
@@ -0,0 +1,464 @@
...
+    num_bytes_processed: u64,
+    /// Number of messages processed by the source.
+    num_messages_processed: u64,
+    // Number of invalid messages, i.e., that were empty or could not be parsed.
+    num_invalid_messages: u64,
+    /// Number of time we looped without getting a single message
+    num_consecutive_empty_batches: u64,
+}
+
+pub struct QueueProcessor {
```
fulmicoton, line 55: Processor is not a great name.... we already use it do describe the thing that transforms json into a document.

quickwit/quickwit-metastore/src/metastore/postgres/queries/shards/open.sql
rdettai: from what I understood, the token is not acquired when openeing the shard in ingest v2 because this operation is performed by the control plane. The ingest sources later call acquire to assign a publish token.

quickwit/quickwit-indexing/src/source/queue_sources/local_state.rs
```diff
@@ -0,0 +1,150 @@
...
+/// recorded as `in_flight`, where the visibility handle is kept alive. Messages
+/// are then moved to `in_progress` to track the reader's progress. It's only
+/// when the events are published that the in-flight visiblity handles are
+/// dropped and the partition is marked as `completed``.
+#[derive(Default)]
+pub struct QueueLocalState {
+    in_flight: BTreeMap<PartitionId, VisibilityTaskHandle>,
+    ready: VecDeque<CheckpointedMessage>,
+    in_progress: Option<InProgressMessage>,
+    completed: BTreeSet<PartitionId>,
```
fulmicoton, line 41: we need comment explaining what all of these are about...
the naming is incredibly confusing.

I suspect better names will emerge from commenting.

quickwit/quickwit-indexing/src/source/queue_sources/local_state.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+impl QueueLocalState {
+    /// Split the message list into:
+    /// - processable (returned)
+    /// - already processed (returned)
+    /// - tracked as in progress (dropped)
+    pub fn filter_completed(
```
fulmicoton, line 49:
```suggestion
    fn filter_completed(
```

quickwit/quickwit-indexing/src/source/queue_sources/processor.rs
```diff
@@ -0,0 +1,464 @@
...
+                &self.publish_token,
+                categorized_using_local_state.processable,
+            )
+            .await?;
+
+        // Drop visibility tasks for messages that have been processed by another pipeline
+        let completed_visibility_tasks = categorized_using_shared_state
+            .already_processed
+            .iter()
+            .filter_map(|m| self.local_state.mark_completed(m.partition_id()))
```
fulmicoton, line 170:
```suggestion
            .filter_map(|m| self.local_state.mark_completed(m.partition_id()))
```

please avoid one letter variable.

quickwit/quickwit-config/src/source_config/mod.rs
rdettai: I think we could use a more explicit name and semantic. What do you think about replacing this with:

```
pub fn checkpoint_store() -> CheckpointStore
```

where 

```
pub enum CheckpointStore {
    ShardApi,
    IndexMetadata,
}
```

quickwit/quickwit-indexing/src/source/queue_sources/processor.rs
```diff
@@ -0,0 +1,464 @@
...
+
+        Ok(())
+    }
+
+    pub async fn emit_batches(
+        &mut self,
+        doc_processor_mailbox: &Mailbox<DocProcessor>,
+        ctx: &SourceContext,
+    ) -> Result<Duration, ActorExitStatus> {
+        if let Some(msg) = self.local_state.in_progress_mut() {
```
fulmicoton, line 192: I think this flow is not a great idea, also I agree this is tricky.

I would have preferred to see a stream with different tasks feeding into a channel:

```
- get message task
- download task
```

This method would then return whatever is available is in the channel.

quickwit/quickwit-indexing/src/source/queue_sources/processor.rs
```diff
@@ -0,0 +1,464 @@
...
+
+    pub async fn suggest_truncate(
+        &mut self,
+        checkpoint: SourceCheckpoint,
+        _ctx: &SourceContext,
+    ) -> anyhow::Result<()> {
+        let completed_visibility_handles = checkpoint
+            .iter()
+            .filter(|(_, pos)| pos.is_eof())
+            .filter_map(|(pid, _)| self.local_state.mark_completed(pid))
```
fulmicoton, line 220: Using the fluent iterator style to mutate stuff make things confusing
(here `mark_completed`)  and is error prone.

side effects like logging etc are ok of course. But modifying an object etc. is to avoid.
Prefer a simple look here.

quickwit/quickwit-indexing/src/source/queue_sources/mod.rs
fulmicoton: I understood what you meant after reading other bits of code.

quickwit/quickwit-indexing/src/source/queue_sources/local_state.rs
rdettai: I added some comments but better names are still to be found

fulmicoton:
we need a different handling of transient vs non-transient error.
e.g.
in the message parsing -> non-transient
disconnection while streaming file -> transient...
gzip corruption -> non-transient.

quickwit/quickwit-indexing/src/source/doc_file_reader.rs
fulmicoton: discussed offline, (I hallucinated and I thought it was the opposite)

quickwit/quickwit-config/src/source_config/mod.rs
fulmicoton, line 252: nitpick - I would put that logic as a funciton in the metastore project.

quickwit/quickwit-indexing/src/source/queue_sources/local_state.rs
rdettai: I made the names significantly more explicit.

quickwit/quickwit-indexing/src/source/queue_sources/local_state.rs
rdettai: I removed `pub` everywhere I could

quickwit/quickwit-indexing/src/source/queue_sources/processor.rs
rdettai: It is now a `QueueCoordinator`

quickwit/quickwit-indexing/src/source/queue_sources/mod.rs
rdettai: I think this ties back to https://github.com/quickwit-oss/quickwit/pull/5148#issuecomment-2188322448. I added an item list in the description for this.

quickwit/quickwit-indexing/src/source/doc_file_reader.rs
```diff
@@ -129,6 +135,142 @@ impl DocFileReader {
...
+                Position::offset(new_offset)
+            };
+            batch_builder
+                .checkpoint_delta
+                .record_partition_delta(
+                    self.partition_id.clone(),
+                    Position::offset(self.current_offset),
+                    to_position,
+                )
+                .unwrap();
```
fulmicoton, line 219: unwrap?

quickwit/quickwit-indexing/src/source/doc_file_reader.rs
```diff
@@ -129,6 +135,142 @@ impl DocFileReader {
...
+        while new_offset < limit_num_bytes {
+            if let Some(record) = source_ctx.protect_future(self.reader.next_record()).await? {
+                new_offset = record.next_offset as usize;
+                batch_builder.add_doc(record.doc);
+            } else {
+                self.is_eof = true;
+                break;
+            }
+        }
+        if new_offset > self.current_offset {
```
fulmicoton, line 206: So if we cut exactly at the limit, we will never have the EOF position?

quickwit/quickwit-indexing/src/source/doc_file_reader.rs
```diff
@@ -129,6 +135,142 @@ impl DocFileReader {
...
+#[async_trait]
+impl BatchReader for StdinBatchReader {
+    async fn read_batch(
+        &mut self,
+        source_ctx: &SourceContext,
+        source_type: SourceType,
+    ) -> anyhow::Result<BatchBuilder> {
+        let mut batch_builder = BatchBuilder::new(source_type);
+        while batch_builder.num_bytes < BATCH_NUM_BYTES_LIMIT {
+            let mut buf = String::new();
```
fulmicoton, line 254: why allocate at every single loop iteration?

quickwit/quickwit-indexing/src/source/queue_sources/visibility.rs
```diff
@@ -0,0 +1,226 @@
...
+    fn observable_state(&self) -> Self::ObservableState {
+        json!({
+            "ack_id": self.ack_id,
+            "extension_count": self.extension_count,
+        })
+    }
+}
+
+#[derive(Debug)]
+pub(super) struct Loop;
```
fulmicoton, line 181:
```suggestion
struct Loop;
```

It is super important to make it private for correctness reasons.

(`Loop` messages are extremely dangerous, it is very important to make their proofreading as trivial as possible)

quickwit/quickwit-indexing/src/source/queue_sources/visibility.rs
```diff
@@ -0,0 +1,226 @@
...
+struct VisibilityTask {
+    queue: Arc<dyn Queue>,
+    ack_id: String,
+    extension_count: u64,
+    current_deadline: Instant,
+    stop_extension_loop: bool,
+    visibility_settings: VisibilitySettings,
+}
+
+// A handle to the visibility actor. When dropped, the actor exits and the
```
fulmicoton, line 74: How does that work? is this tested?

quickwit/quickwit-indexing/src/source/queue_sources/visibility.rs
```diff
@@ -0,0 +1,226 @@
...
+            .await?;
+        ctx.schedule_self_msg(self.next_extension(), Loop);
+        Ok(())
+    }
+}
+
+/// Ensures that the visibility of the message is extended until the given
+/// deadline and then stops the extension loop.
+#[derive(Debug)]
+pub(super) struct RequestLastExtension {
```
fulmicoton, line 205:
```suggestion
struct RequestLastExtension {
```

fulmicoton, line 205:
```suggestion
RequestLastExtension {
```

quickwit/quickwit-indexing/src/source/queue_sources/local_state.rs
```diff
@@ -0,0 +1,117 @@
...
+            .any(|msg| &msg.partition_id() == partition_id)
+    }
+
+    pub fn is_read_in_progress(&self, partition_id: &PartitionId) -> bool {
+        self.read_in_progress
+            .as_ref()
+            .map_or(false, |msg| &msg.partition_id == partition_id)
+    }
+
+    pub fn is_awating_commit(&self, partition_id: &PartitionId) -> bool {
```
fulmicoton, line 60:
```suggestion
    pub fn is_awaiting_commit(&self, partition_id: &PartitionId) -> bool {
```

quickwit/quickwit-indexing/src/source/queue_sources/memory_queue.rs
```diff
@@ -0,0 +1,185 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Queue")
+            .field("in_queue_count", &self.in_queue.len())
+            .field("in_flight_count", &self.in_flight.len())
+            .field("acked_count", &self.acked.len())
+            .finish()
+    }
+}
+
+/// A simple in-memory queue
```
fulmicoton, line 49: is this for test?

quickwit/quickwit-indexing/src/source/queue_sources/memory_queue.rs
```diff
@@ -0,0 +1,185 @@
...
+#[derive(Clone, Default, Debug)]
+pub struct MemoryQueue {
```
fulmicoton, line 51: is this just for test mocking? If so can you make that clear? Not compile it on release? 
Rename it MemoryQueueForTest or whatever?

quickwit/quickwit-metastore/src/metastore/mod.rs
```diff
@@ -891,6 +892,26 @@ impl<T> Default for FilterRange<T> {
...
+        SourceParams::File(FileSourceParams::Sqs(_)) => true,
+        SourceParams::File(FileSourceParams::Stdin) => panic!("stdin cannot be checkpointed"),
+        SourceParams::Ingest => true,
+        SourceParams::IngestApi => false,
+        SourceParams::IngestCli => false,
+        SourceParams::Kafka(_) => false,
+        SourceParams::Kinesis(_) => false,
+        SourceParams::PubSub(_) => false,
+        SourceParams::Pulsar(_) => false,
+        SourceParams::Sqs(_) => true,
```
fulmicoton, line 909: I don't understand. You ended up doing two source params? One via the File source and one via an adhoc `SourceParams`?

quickwit/quickwit-proto/protos/quickwit/metastore.proto
```diff
@@ -41,6 +41,8 @@ enum SourceType {
   SOURCE_TYPE_PULSAR = 9;
   SOURCE_TYPE_VEC = 10;
   SOURCE_TYPE_VOID = 11;
+  /// Amazon SQS
+  SOURCE_TYPE_SQS = 12;
```
fulmicoton, line 45: This is super confusing...

Is it really SQS? Is it the Queue source? There is pubsub in this list too.
Do we have anything to hint that one is holding message, and hte other is holding urls to files?

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -121,35 +160,51 @@ impl TypedSourceFactory for FileSourceFactory {
...
+                    num_lines_processed: 0,
+                }
+            }
+            FileSourceParams::Stdin => FileSourceInner::Stream {
+                reader: Box::new(StdinBatchReader::new()),
+                num_bytes_processed: 0,
+                num_lines_processed: 0,
+            },
+            #[cfg(feature = "sqs")]
+            FileSourceParams::Sqs(sqs_config) => {
```
fulmicoton, line 193: for pubsub what will to look like?

quickwit/quickwit-indexing/src/source/queue_sources/memory_queue.rs
rdettai: > Not compile it on release? 

It's not compiled on release: https://github.com/quickwit-oss/quickwit/blob/37836be97f59e7420298dd6bbc668c3ba6d3327f/quickwit/quickwit-indexing/src/source/queue_sources/mod.rs#L22-L23

> Rename it MemoryQueueForTest or whatever?

:+1:

quickwit/quickwit-proto/protos/quickwit/metastore.proto
rdettai: answered in https://github.com/quickwit-oss/quickwit/pull/5148#discussion_r1662158060

quickwit/quickwit-indexing/src/source/file_source.rs
rdettai: https://github.com/quickwit-oss/quickwit/pull/5148#discussion_r1662158060

quickwit/quickwit-metastore/src/metastore/mod.rs
rdettai: I'm answering also https://github.com/quickwit-oss/quickwit/pull/5148#discussion_r1662168105 and https://github.com/quickwit-oss/quickwit/pull/5148#discussion_r1662163358 here to provide 1 consistent answer:

This approach is the result our discussion with @guilload:
- The file source reads data from files. The queue (SQS or Pubsub) is just a notification system and is configured as part of the `SourceParams::File`
- If requested by users, we might also provide queue sources (SQS and Pubsub) that read the data directly from the messages. I left the boilerplate for `SourceParams::Sqs` because I find it convenient to keep an eye on how that would look like while I still move the abstractions around. But we can definitively remove it in the name of YAGNI.

Both will use the same infrastructure for message tracking (`QueueCoordinator`), but in terms of API they will be completely separate sources.

quickwit/quickwit-indexing/src/source/doc_file_reader.rs
rdettai: Took it from current main (https://github.com/quickwit-oss/quickwit/blob/84572ea5955ecbe2b0d57ae5a1fbe5d5600ce716/quickwit/quickwit-indexing/src/source/file_source.rs#L102-L109). Changine to a `?` instead.

quickwit/quickwit-indexing/src/source/doc_file_reader.rs
rdettai: Copied from main https://github.com/quickwit-oss/quickwit/blob/84572ea5955ecbe2b0d57ae5a1fbe5d5600ce716/quickwit/quickwit-indexing/src/source/file_source.rs#L80. I guess we could allocate one big memory region that is `BATCH_NUM_BYTES_LIMIT`, load data into it and slice it by row. I think it's a bit far from this PR's concern though. Do you think these allocations are costly enough for opening an issue?

quickwit/quickwit-indexing/src/source/queue_sources/visibility.rs
rdettai: My understanding is that dropping the last mailbox of an actor exits it: https://github.com/quickwit-oss/quickwit/blob/207626185a5327d2d6fd84181f3e383fd274fe57/quickwit/quickwit-actors/src/mailbox.rs#L49-L50

But after adding some tests in https://github.com/quickwit-oss/quickwit/pull/5148/commits/95beb16ce36efae2bea5ae36e97b3495796e39d4 I realized that wasn't the case.

rdettai: added an `Arc<()>` to perform ref counting instead of relying on the mailbox

rdettai:
Note: I just stumbled upon https://github.com/quickwit-oss/quickwit/issues/1065 which could be addressed as part of the reorganization of the `FileSource` that is happening here.

quickwit/quickwit-aws/src/error.rs
```diff
@@ -36,6 +36,11 @@ use aws_sdk_s3::operation::get_object::GetObjectError;
 use aws_sdk_s3::operation::head_object::HeadObjectError;
 use aws_sdk_s3::operation::put_object::PutObjectError;
 use aws_sdk_s3::operation::upload_part::UploadPartError;
+#[cfg(feature = "sqs")]
+use aws_sdk_sqs::operation::{
+    change_message_visibility::ChangeMessageVisibilityError,
```
fulmicoton, line 41: nitpick I think you can group all of the sqs conditional complication like this.

```rust
#[cfg(feature = "sqs")]
{
    #[cfg(feature = "sqs")]
use aws_sdk_sqs::operation::{
    change_message_visibility::ChangeMessageVisibilityError,
    delete_message_batch::DeleteMessageBatchError, receive_message::ReceiveMessageError,
};

impl AwsRetryable for ReceiveMessageError {
    fn is_retryable(&self) -> bool {
        false
    }
}

#[cfg(feature = "sqs")]
impl AwsRetryable for DeleteMessageBatchError {
    fn is_retryable(&self) -> bool {
        false
    }
}

#[cfg(feature = "sqs")]
impl AwsRetryable for ChangeMessageVisibilityError {
    fn is_retryable(&self) -> bool {
        false
    }
}
}
```

If you had exports, it was still possible by introducing a module.

```
mod sqs {
 //..
}
pub use sqs::*;

File: quickwit/quickwit-config/src/source_config/mod.rs [hunk 28]
```diff
@@ -225,59 +230,113 @@ pub enum SourceParams {
...
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize, utoipa::ToSchema)]
 #[serde(deny_unknown_fields)]
-pub struct FileSourceParams {
-    /// Path of the file to read. Assume stdin if None.
-    #[schema(value_type = String)]
-    #[serde(skip_serializing_if = "Option::is_none")]
-    #[serde(default)]
-    #[serde(deserialize_with = "absolute_filepath_from_str")]
-    pub filepath: Option<PathBuf>, //< If None read from stdin.
+pub struct FileSourceParamsInner {
```
fulmicoton, line 280: Does this need to be `pub`?

fulmicoton, line 280: I don't think this needs to be pub.
The name is wrong too.

`FileSourceParamsInner` -> `FileSourceParamsForSerialization`

The member also do not need to be pub.

quickwit/quickwit-indexing/src/actors/publisher.rs
```diff
@@ -147,7 +147,7 @@ impl Handler<SplitsUpdate> for Publisher {
             );
             return Ok(());
         }
-        info!(new_splits=?split_ids, checkpoint_delta=?checkpoint_delta_opt, "publish-new-splits");
+        info!("publish-new-splits");
```
fulmicoton, line 150: Why?

It has actually extremely useful in the past to be able to search a split id to track its lifecycle.
I assume you did this because checkpoints  are getting too large?

Can you just fix their Debug implementation to make sure it gets shortened if it is too large.

quickwit/quickwit-indexing/src/source/doc_file_reader.rs
```diff
@@ -0,0 +1,549 @@
...
+    fn new(reader: Box<dyn AsyncRead + Send + Unpin>, num_bytes_to_skip: usize) -> Self {
+        Self {
+            reader: BufReader::new(reader),
+            num_bytes_to_skip,
+        }
+    }
+
+    async fn skip(&mut self) -> io::Result<()> {
+        // allocate on the heap to avoid stack overflows
+        let mut buf = vec![0u8; 64000];
```
fulmicoton, line 59:
```suggestion
        let mut buf = vec![0u8; 64_000];
```

quickwit/quickwit-cli/tests/cli.rs
```diff
@@ -252,14 +257,17 @@ async fn test_ingest_docs_cli() {
 
     // Ensure cache directory is empty.
     let cache_directory_path = get_cache_directory_path(&test_env.data_dir_path);
-
     assert!(cache_directory_path.read_dir().unwrap().next().is_none());
 
+    let does_not_exit_uri = uri_from_path(&test_env.data_dir_path)
```
fulmicoton, line 262:
```suggestion
    let does_not_exist_uri = uri_from_path(&test_env.data_dir_path)
```

quickwit/quickwit-indexing/src/source/doc_file_reader.rs
```diff
@@ -0,0 +1,549 @@
...
+            }
+        };
+        Ok(reader)
+    }
+
+    /// Reads the next record from the underlying file. Returns `None` when EOF
+    /// is reached.
+    pub async fn next_record(&mut self) -> anyhow::Result<Option<FileRecord>> {
+        let mut buf = String::new();
+        // TODO add retry if read stream is broken because of a transient error
```
fulmicoton, line 133: can you open a ticket if this is important? We tend to forget TODOs.

quickwit/quickwit-indexing/src/source/doc_file_reader.rs
```diff
@@ -0,0 +1,549 @@
...
+                if record.is_last {
+                    self.is_eof = true;
+                    break;
+                }
+            } else {
+                self.is_eof = true;
+                break;
+            }
+        }
+        if batch_builder.num_bytes > 0 {
```
fulmicoton, line 209: Is the `batch_builder.num_bytes > 0` necessary here?

fulmicoton: Won't this be a problem if the file is empty?

quickwit/quickwit-indexing/src/source/file_source.rs
```diff
@@ -17,45 +17,36 @@
...
 
-/// Number of bytes after which a new batch is cut.
-pub(crate) const BATCH_NUM_BYTES_LIMIT: u64 = 500_000u64;
-
-#[derive(Default, Clone, Debug, Eq, PartialEq, Serialize)]
-pub struct FileSourceCounters {
-    pub previous_offset: u64,
-    pub current_offset: u64,
-    pub num_lines_processed: u64,
+enum FileSourceInner {
```
fulmicoton, line 36: `FileSourceInner` is not a good name. 

We use the `Inner` suffix for `Pimpl` like patterns for instance, or when we need to wrap struct with a `Mutex`.

quickwit/quickwit-indexing/src/source/queue_sources/coordinator.rs
```diff
@@ -0,0 +1,523 @@
...
+    pub num_lines_processed: u64,
+    /// Number of messages processed by the source.
+    pub num_messages_processed: u64,
+    /// Number of messages that could not be pre-processed.
+    pub num_messages_failed_preprocessing: u64,
+    /// Number of messages that could not be moved to in-progress.
+    pub num_messages_failed_opening: u64,
+}
+
+pub struct QueueCoordinator {
```
fulmicoton, line 66: Can we have Docstring here explaining the basis about what the queue coordinator is for.

quickwit/quickwit-indexing/src/source/queue_sources/coordinator.rs
```diff
@@ -0,0 +1,523 @@
...
+        }
+        if preprocessed_messages.is_empty() {
+            self.queue.acknowledge(&discardable_ack_ids).await?;
+            return Ok(());
+        }
+
+        // in rare situations, there might be duplicates within a batch
+        let deduplicated_messages = preprocessed_messages
+            .into_iter()
+            .dedup_by(|x, y| x.partition_id() == y.partition_id());
```
fulmicoton, line 190: dedup only works with consecutive messages.

fulmicoton: you can use `unique_by` I think.

quickwit/quickwit-indexing/src/source/queue_sources/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+mod visibility;
+
+use std::fmt;
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+use async_trait::async_trait;
+use message::RawMessage;
+
+type ReceiveResult = anyhow::Result<Vec<RawMessage>>;
```
fulmicoton, line 40: I suspect this typedef makes the code harder to read.

quickwit/quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -1640,28 +1633,6 @@ mod tests {
         assert!(indexes.is_empty());
     }
 
-    #[tokio::test]
-    async fn test_create_file_source_returns_403() {
-        let metastore = metastore_for_test();
-        let index_service = IndexService::new(metastore.clone(), StorageResolver::unconfigured());
-        let mut node_config = NodeConfig::for_test();
-        node_config.default_index_root_uri = Uri::for_test("file:///default-index-root-uri");
```
fulmicoton, line 1648: can we get a test checking that a uri targetting a local file is forbidden?

(I see you added stdin)

quickwit/quickwit-proto/src/metastore/mod.rs
```diff
@@ -288,6 +290,8 @@ impl fmt::Display for SourceType {
             SourceType::Nats => "NATS",
             SourceType::PubSub => "Google Cloud Pub/Sub",
             SourceType::Pulsar => "Apache Pulsar",
+            SourceType::Stdin => "stdin",
```
fulmicoton, line 293:
```suggestion
            SourceType::Stdin => "Stdin",
```

quickwit/quickwit-lambda/src/indexer/model.rs
```diff
@@ -31,17 +32,17 @@ pub enum IndexerEvent {
 }
 
 impl IndexerEvent {
-    pub fn uri(&self) -> PathBuf {
-        match &self {
-            IndexerEvent::Custom { source_uri } => PathBuf::from(source_uri),
+    pub fn uri(&self) -> anyhow::Result<Uri> {
+        let path = match self {
```
fulmicoton, line 36:
```suggestion
        let path: String = match self {
```

quickwit/quickwit-aws/src/error.rs
rdettai: The approach with the plain `{}` block doesn´t seem to work. It was added for blocks within function [here](https://github.com/rust-lang/rfcs/pull/16/files) but not for top level blocks. 

Using `mod` is a good idea here, also applying this to the Kinesis part

quickwit/quickwit-config/src/source_config/mod.rs
rdettai: I was mimicking `KinesisSourceParamsInner` nearby, but I agree with all these comments!

quickwit/quickwit-indexing/src/actors/publisher.rs
rdettai: actually, I just removed these fields because they are also at the span level:
https://github.com/quickwit-oss/quickwit/blob/996849f60a98a54fa49dfa6721f965f62750d4fc/quickwit/quickwit-indexing/src/actors/publisher.rs#L104-L109

quickwit/quickwit-indexing/src/source/doc_file_reader.rs
rdettai: this is what's in the current `FileSource`:
https://github.com/quickwit-oss/quickwit/blob/69f700aa72930b9a395467886ce79b8f7eeabcb6/quickwit/quickwit-indexing/src/source/file_source.rs#L95-L114

I was wondering if it was fine sending empty batches through the pipeline. I'll take a deeper look to that and remove the condition if it's OK.

quickwit/quickwit-indexing/src/source/queue_sources/mod.rs
rdettai: Yes, I'm also sceptical about this. I needed something because the linter was complaining that `Option<BoxFuture<'static, ReceiveResult>>anyhow::Result<Vec<RawMessage>>` was too long. I'll go for a more local typedef.

quickwit/quickwit-indexing/src/source/doc_file_reader.rs
rdettai: https://github.com/quickwit-oss/quickwit/issues/5243

quickwit/quickwit-serve/src/index_api/rest_handler.rs
rdettai: I removed that validation as stated [here](https://github.com/quickwit-oss/quickwit/pull/5148#issuecomment-2210625199). But I actually agree it's better not to have this kind of side effects, so I'll add the check back.

quickwit/quickwit-metastore/src/metastore/mod.rs
rdettai: I removed the `SqsQueue` boilerplate for now. I created a branch to save the boilerplate for reference: https://github.com/quickwit-oss/quickwit/tree/with-sqs-src

docs/ingest-data/sqs-files.md
```diff
@@ -0,0 +1,235 @@
...
+In another terminal, in the Quickwit install directory, create an index:
+
+```bash
+cat << EOF > tutorial-sqs-file-index.yaml
+version: 0.7
+index_id: tutorial-sqs-file
+doc_mapping:
+  mode: dynamic
+indexing_settings:
+  commit_timeout_secs: 5
```
fulmicoton, line 186:
```suggestion
  commit_timeout_secs: 30
```

docs/ingest-data/sqs-files.md
```diff
@@ -0,0 +1,235 @@
...
+
+Replacing `<notification_queue_url>` with the corresponding Terraform output
+value, create a file source for that index:
+
+```bash
+cat << EOF > tutorial-sqs-file-source.yaml
+version: 0.8
+source_id: sqs-filesource
+source_type: file
+num_pipelines: 2
```
fulmicoton, line 200: It is worth adding a sentence or two explaining the point/effect of num_pipelines

quickwit/quickwit-indexing/src/source/queue_sources/message.rs
```diff
@@ -0,0 +1,295 @@
...
+    pub payload: PreProcessedPayload,
+}
+
+impl PreProcessedMessage {
+    pub fn partition_id(&self) -> PartitionId {
+        self.payload.partition_id()
+    }
+}
+
+fn uri_from_s3_notification(message: &OwnedBytes, ack_id: &str) -> Result<Uri, PreProcessingError> {
```
fulmicoton, line 137:
```suggestion
fn uri_from_s3_notification(message: &[u8], ack_id: &str) -> Result<Uri, PreProcessingError> {
```

quickwit/quickwit-indexing/src/source/queue_sources/message.rs
```diff
@@ -0,0 +1,295 @@
...
+    let value: Value =
+        serde_json::from_slice(message.as_slice()).context("invalid JSON message")?;
+    if matches!(value["Event"].as_str(), Some("s3:TestEvent")) {
+        return Err(PreProcessingError::Discardable {
+            reason: "S3 test event",
+            ack_id: ack_id.to_string(),
+        });
+    }
+    let key = value["Records"][0]["s3"]["object"]["key"]
```
fulmicoton, line 146: What if the event is not a PUT? 
I think we need to check the event type.

quickwit/quickwit-indexing/src/source/queue_sources/message.rs
```diff
@@ -0,0 +1,295 @@
...
+                        "eTag": "d41d8cd98f00b204e9800998ecf8427e",
+                        "versionId": "096fKKXTRTtl3on89fVO.nfljtsv6qko",
+                        "sequencer": "0055AED6DCD90281E5"
+                    }
+                }
+                }
+            ]
+        }"#;
+        let actual_uri =
+            uri_from_s3_notification(&OwnedBytes::new(test_message.as_bytes()), "myackid").unwrap();
```
fulmicoton, line 250:
```suggestion
            uri_from_s3_notification(test_message.as_bytes(), "myackid").unwrap();
```

docs/configuration/source-config.md
```diff
@@ -29,13 +29,48 @@ The source type designates the kind of source being configured. As of version 0.
...
 ```bash
-./quickwit tool local-ingest --input-path <INPUT_PATH>
+./quickwit tool local-ingest --index <index> --input-path <input-path>
+```
+
+Both local and object files are supported, provided that the environment is configured with the appropriate permissions. A tutorial is available [here](/docs/ingest-data/ingest-local-file.md).
+
+#### Notification based file ingestion
+
+You can setup a source that uses notifications to continuously index new files. A complete tutorial can be found [here](/docs/ingest-data/sqs-files.md).
```
fulmicoton, line 48:
```suggestion
Quickwit can automatically ingest all new files that are uploaded on a S3 bucket.
This requires creating and configuring a [SQS notification queue](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ways-to-add-notification-config-to-bucket.html).

A complete tutorial can be found [here](/docs/ingest-data/sqs-files.md).
```

fulmicoton: (right now we only support S3, and I suspect some users are not aware of this SQS notification system)

docs/configuration/source-config.md
```diff
@@ -29,13 +29,48 @@ The source type designates the kind of source being configured. As of version 0.
...
+source_id: my-sqs-file-source
+source_type: file
+num_pipelines: 2
+params:
+  notifications:
+    - type: sqs
+      queue_url: https://sqs.us-east-1.amazonaws.com/123456789012/queue-name
+      message_type: s3_notification
+EOF
+./quickwit source create --index my-index --source-config source-config.yaml
```
fulmicoton, line 73: can you add a disclaimer explaining that quickwit will NOT delete files after their indexing, and that this is the responsability of the user.

quickwit/quickwit-indexing/src/source/queue_sources/sqs_queue.rs
```diff
@@ -0,0 +1,471 @@
...
+
+        receive_output
+            .messages
+            .unwrap_or_default()
+            .into_iter()
+            .map(|msg| {
+                let delivery_attempts: usize = msg
+                    .attributes
+                    .as_ref()
+                    .and_then(|attrs| {
```
fulmicoton, line 96: nitpick

Generally speaking we try to avoid leaning too much into this iterator/functional style.
When dealing with errors. It rapidly makes the code unreadable.

quickwit/quickwit-indexing/src/source/queue_sources/message.rs
```diff
@@ -0,0 +1,295 @@
...
+        Ok(PreProcessedMessage {
+            metadata: self.metadata,
+            payload,
+        })
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum PreProcessedPayload {
+    ObjectUri(Uri),
```
fulmicoton, line 110: This is overengineering preparing fro what?

quickwit/quickwit-indexing/src/source/queue_sources/message.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+/// A message that went through the minimal transformation to discover its
+/// partition id. Indeed, the message might be discarded if the partition was
+/// already processed, so it's better to avoid doing unnecessary work at this
+/// stage.
+#[derive(Debug, PartialEq, Eq)]
+pub struct PreProcessedMessage {
+    pub metadata: MessageMetadata,
+    pub payload: PreProcessedPayload,
```
fulmicoton, line 128: payload is too generic. What is this payload about?
Right now I'd rather have a simple `Uri`

quickwit/quickwit-indexing/src/source/queue_sources/message.rs
```diff
@@ -0,0 +1,295 @@
...
+
+impl ReadyMessage {
+    pub async fn start_processing(
+        self,
+        storage_resolver: &StorageResolver,
+    ) -> anyhow::Result<Option<InProgressMessage>> {
+        let partition_id = self.partition_id();
+        match self.content.payload {
+            PreProcessedPayload::ObjectUri(uri) => {
+                let batch_reader = ObjectUriBatchReader::try_new(
```
fulmicoton, line 171: What is happening here if the object is not present (was deleted for instance).

I think we need to log a warning and acknowledge the object to remove it from the queue.

Feel free to discuss the specification, but we need something that makes sense + a unit test checking that?

quickwit/quickwit-indexing/src/source/queue_sources/shared_state.rs
```diff
@@ -0,0 +1,372 @@
...
+        partitions: Vec<PartitionId>,
+    ) -> anyhow::Result<Vec<(PartitionId, Position)>> {
+        let open_shard_subrequests = partitions
+            .iter()
+            .enumerate()
+            .map(|(idx, partition_id)| OpenShardSubrequest {
+                subrequest_id: idx as u32,
+                index_uid: Some(self.index_uid.clone()),
+                source_id: self.source_id.clone(),
+                // TODO: make this optional?
```
fulmicoton, line 58:
```suggestion
```

docs/configuration/source-config.md
fulmicoton, line 48: Can you mark the feature a beta in some way?

quickwit/quickwit-indexing/src/actors/publisher.rs
fulmicoton: Good point!

quickwit/quickwit-indexing/src/source/queue_sources/message.rs
rdettai: I agree that we can apply YAGNI here. But at the same time, I would like to quickly use this to re-implement the Google Pub/Sub source in a follow up PR.

quickwit/quickwit-indexing/src/source/queue_sources/message.rs
rdettai: I think a deleted message **is** an error. It should really never happen, and if it does we are in a situation where we want to let the user determine manually what to do. The safe way to this is to let the message go into a dead letter queue after a few retries.

quickwit/quickwit-indexing/src/source/queue_sources/message.rs
rdettai: > Taking &OwnedBytes is never a good idea 

`&[u8]` is much better here, thanks

> it is possible for a notification to not be about a PUT (I think?). In that case, I don't think this is a great idea to return it as an error
We cannot really do much more on the caller's side by knowing that the message has a wrong event type vs. any other kind of malformation, so I'll leave the return type as is for now.

fulmicoton: I don't understand your point.

quickwit/quickwit-indexing/src/source/queue_sources/message.rs
fulmicoton: This seems needlessly rigid but ok.

rdettai: You are right, this is excessively rigid, these can be acknowledged. I'll leave a warn! because this user configuration generates unnecessary traffic on the queue.

fulmicoton: ok

fulmicoton (approved):

PR #5156, opened 2024-06-25
More aggressive shard scale up settings.

fulmicoton:
1 shard every 2 seconds vs 5 seconds before.


guilload:
FYI, I had already increased the limit from 5 shards/minute to 12.

PR #5163, opened 2024-06-26
Faster document validation

fulmicoton:
This PR just removes the need to build documents when validating
documents.


guilload (approved):
LGTM. The PR could use some unit tests, though.

PR #5181, opened 2024-06-28
Optimize json parsing for validation

fulmicoton:


quickwit/quickwit-doc-mapper/src/default_doc_mapper/date_time_type.rs
```diff
@@ -71,6 +71,38 @@ impl Default for QuickwitDateTimeOptions {
...
+                // first.
+                if let Some(timestamp_i64) = timestamp.as_i64() {
+                    quickwit_datetime::parse_timestamp_int(timestamp_i64, &self.input_formats.0)?;
+                    Ok(())
+                } else if let Some(timestamp_f64) = timestamp.as_f64() {
+                    quickwit_datetime::parse_timestamp_float(timestamp_f64, &self.input_formats.0)?;
+                    Ok(())
+                } else {
+                    Err(format!(
+                        "failed to parse datetime `{:?}`: value is larger than i64::MAX",
```
PSeitz, line 90: this code is unreachable since `as_f64` always succeeds

quickwit/quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
```diff
@@ -225,6 +217,20 @@ impl BinaryFormat {
                 format!("expected hex string, got `{byte_str}`: {hex_decode_err}")
             })?,
         };
+        Ok(payload)
+    }
+
+    /// Parses the `serde_json::Value` into `tantivy::schema::Value`.
+    pub fn parse_json(&self, json_val: &JsonValue) -> Result<TantivyValue, String> {
```
PSeitz, line 224: Comment and name suggest the method is more generic than it is
```suggestion
    /// Parses the `serde_json::Value::String` into `TantivyValue::Bytes`.
    pub fn parse_json_str(&self, json_val: &JsonValue) -> Result<TantivyValue, String> {
```

PSeitz: ah I just saw it's a method in BinaryFormat

quickwit/quickwit-doc-mapper/src/default_doc_mapper/date_time_type.rs
fulmicoton: I can't unwrap though, always returning `Some` is an implementation detail of serde_borrow_json. I'll change the error message.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/field_mapping_entry.rs
fulmicoton: I did not understand

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -691,6 +700,47 @@ fn insert_json_val(
...
+                    Err(format!(
+                        "expected JSON number, got string `\"{str_val}\"`. enable coercion to {} \
+                         with the `coerce` parameter in the field mapping",
+                        type_name::<Self>()
+                    ))
+                }
+            }
+            _ => {
+                let message = if coerce {
+                    format!("expected JSON number or string, got `{json_val}`")
```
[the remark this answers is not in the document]
fulmicoton: Maybe... but that's a different issue, and probably not worth the effort.

PR #5202, opened 2024-07-08
Emitting both a short term average and a long term average of shard

fulmicoton:
throughput.

Scaling up relies on the short term average in order to rapidly react to a change in throughput, while scaling down and the indexing scheduler relies on the long term average.


quickwit/quickwit-ingest/src/ingest_v2/broadcast.rs
```diff
@@ -148,24 +160,119 @@ impl LocalShardsSnapshot {
...
+                .entry(source_uid.clone())
+                .or_default()
+                .insert(shard_info);
+        }
+        per_source_shard_infos
+    }
+}
+
+#[derive(Default)]
+struct ShardThroughputTimeSerie {
```
guilload, line 225: The singular of time-series is time-series.

guilload (approved):

PR #5206, opened 2024-07-09
Removing invalid docs from persistence.

### LINKED ISSUE #5205: Ditching invalid documents upon ingestion

fulmicoton:
Removing invalid docs from persistence.

    This PR filters out invalid docs from batches before persisting the
    batch.
    That way, they won't be persisted in the mrecordlog and the
    indexing pipeline will not have to parse them again (and log an error).

    This PR also improves a little bit on the metric for ingested docs.
    The num_docs and bytes metric now have a validity `label`, and the
    point at which they are measured is documented in the metric description.

    Closes #5205

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/doc_mapper.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/doc_mapper.rs
@@ -18,7 +18,7 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::collections::hash_map::Entry;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, Weak};
 
 use once_cell::sync::OnceCell;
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-ingest/src/ingest_v2/doc_mapper.rs
```diff
@@ -122,7 +148,7 @@ pub(super) async fn validate_doc_batch(
     doc_mapper: Arc<dyn DocMapper>,
 ) -> IngestV2Result<(DocBatchV2, Vec<ParseFailure>)> {
     if is_document_validation_enabled() {
```
trinity-1686a, line 150: out of scope for this PR: it might be worth thinking about disabling validation when VRL is present, or to do VRL validation early. I can imagine a document-doc_mapper-VRL combo where not having VRL here causes a document to be rejected, even though it would have been accepted during actual indexation

trinity-1686a (approved):

quickwit/quickwit-ingest/src/ingest_v2/doc_mapper.rs
fulmicoton: excellent point! I'll create an issue.

Issue #5109, opened 2024-06-12
Actually support the format option in the ES API's RangeQuery


fulmicoton:
the parameter is introduced in #5098, but is noop at the moment.

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html
It makes it possible for the client to pass dates in a specific format 
https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html

fulmicoton:
It can be done by doing the conversion in the elastic search dsl range query (`quickwit-query/src/elastic_query_dsl/range_query.rs`)
and reserializing it into rfc3339 before creating the QueryASt's range query.

PR #5207, opened 2024-07-09
feat(es): add format support for RangeQuery

etolbakov:
### Description

Add `format` parameter to the `RangeQuery`

### How was this PR tested?

TBD

### Related Issue
https://github.com/quickwit-oss/quickwit/issues/5109

### DIFF
--- a/quickwit/quickwit-query/src/elastic_query_dsl/range_query.rs
+++ b/quickwit/quickwit-query/src/elastic_query_dsl/range_query.rs
@@ -40,7 +40,6 @@ pub struct RangeQueryParams {
     lte: Option<JsonLiteral>,
     #[serde(default)]
     boost: Option<NotNaNf32>,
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-query/src/elastic_query_dsl/range_query.rs
```diff
@@ -76,6 +75,15 @@ impl ConvertableToQueryAst for RangeQuery {
...
             },
+            format: match format {
+                None => None,
+                Some(v) => {
+                    match v {
+                        JsonLiteral::String(s) => Some(s),
+                        _ => None
+                    }
+                }
+            },
```
etolbakov, line 86: at the moment I came up with the most straightforward solution, I wonder if that's something that you more or less expected to see @fulmicoton?

PR #5212, opened 2024-07-10
add api endpoint to get query ast

### LINKED ISSUE #5190: add an endpoint to get the query ast

trinity-1686a:
### Description

fix #5190
add api endpoint to get the quickwit and tantivy ast for a query, as well as what splits would be hit (taking into account time range and tags), and how many/what kind of storage request are to be expected for each split

### How was this PR tested?

tested manually plus added basic unit test

### DIFF
--- a/quickwit/quickwit-search/src/root.rs
+++ b/quickwit/quickwit-search/src/root.rs
@@ -35,8 +35,8 @@ use quickwit_proto::metastore::{
 };
 use quickwit_proto::search::{
     FetchDocsRequest, FetchDocsResponse, Hit, LeafHit, LeafRequestRef, LeafSearchRequest,
-    LeafSearchResponse, PartialHit, SearchRequest, SearchResponse, SnippetRequest,
-    SortDatetimeFormat, SortField, SortValue, SplitIdAndFooterOffsets,
+    LeafSearchResponse, PartialHit, SearchPlanResponse, SearchRequest, SearchResponse,
+    SnippetRequest, SortDatetimeFormat, SortField, SortValue, SplitIdAndFooterOffsets,
 };
 use quickwit_proto::types::{IndexUid, SplitId};
 use quickwit_query::query_ast::{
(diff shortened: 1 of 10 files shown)


quickwit/quickwit-search/src/root.rs
```diff
@@ -1103,6 +1121,104 @@ pub async fn root_search(
...
+                .filter(|load_position| **load_position)
+                .count()
+        })
+        .sum();
+    Ok(SearchPlanResponse {
+        result: serde_json::to_string(&SearchPlanResponseRest {
+            quickwit_ast: request_metadata.query_ast_resolved,
+            tantivy_ast: format!("{query:#?}"),
+            searched_splits: split_ids,
+            storage_requests: StorageRequestCount {
```
fulmicoton, line 1210: that's cool you extended the thing to a warmup info etc! can we actually get the number of bytes as well?

quickwit/quickwit-search/src/root.rs
```diff
@@ -1103,6 +1121,104 @@ pub async fn root_search(
...
+    let sstable_query_count = warmup_info.term_dict_fields.len()
+        + warmup_info
+            .terms_grouped_by_field
+            .values()
+            .map(|terms| terms.len())
+            .sum::<usize>();
+    let position_query_count = warmup_info
+        .terms_grouped_by_field
+        .values()
+        .map(|terms| {
```
fulmicoton, line 1198:
```rust
        .map(|terms: <PLEASE ADD THE TYPE HERE>| {
```

quickwit/quickwit-search/src/search_response_rest.rs
```diff
@@ -100,3 +101,37 @@ impl TryFrom<SearchResponse> for SearchResponseRest {
...
+    /// It's possible older splits actually resolve to a different ast.
+    pub tantivy_ast: String,
+    /// List of splits that would be searched by this query
+    pub searched_splits: Vec<String>,
+    /// Requests expected for each split
+    #[schema(value_type = Object)]
+    pub storage_requests: StorageRequestCount,
+}
+
+/// Number of expected storage requests, per request kind
```
fulmicoton, line 122:
```suggestion
/// Number of expected storage requests, per request kind.
/// 
/// This figures do not take in account whether the data is already cached or not.
```

fulmicoton (approved):
see comments inline.

quickwit/quickwit-search/src/root.rs
trinity-1686a: that's something i would find interesting too, but right now it's not as easy to get. All metadata we get here, we get only through the metastore, with zero call to the storage. To get more, we would need to actually download things from storage, which starts having more of a `EXPLAIN ANALYSE` than an `EXPLAIN` semantic. I think that should be left out of scope of this PR, but let's open a ticket so eventually people can pass `?analyze=true` or something to get that information as well

PR #5226, opened 2024-07-16
Make the split deletion grace period configurable.

fulmicoton:
... with the environment variable QW_SPLIT_DELETION_GRACE_PERIOD_SECS.

The default is sensible for most case, but for project airmail, we need to be able to recover from a metastore backup that could be several hours old.

### DIFF
--- a/quickwit/quickwit-search/src/root.rs
+++ b/quickwit/quickwit-search/src/root.rs
@@ -18,13 +18,14 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::collections::{HashMap, HashSet};
+use std::sync::OnceLock;
 use std::time::Duration;
 
 use anyhow::Context;
 use futures::future::try_join_all;
 use itertools::Itertools;
 use quickwit_common::pretty::PrettySample;
-use quickwit_common::shared_consts::{DELETION_GRACE_PERIOD, SCROLL_BATCH_LEN};
+use quickwit_common::shared_consts;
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-search/src/root.rs
```diff
@@ -63,7 +64,23 @@ use crate::{
...
+        assert!(
+            split_deletion_grace_period_secs
+                >= shared_consts::MINIMUM_DELETION_GRACE_PERIOD.as_secs(),
+            "The split deletion grace period is too short ({}s). This should not happen.",
+            split_deletion_grace_period_secs
+        );
+        // We remove an extra margin of 2minutes from the split deletion grace period.
+        Duration::from_secs(
+            quickwit_common::shared_consts::split_deletion_grace_period().as_secs() - 60 * 2,
+        )
```
trinity-1686a, line 81: `Duration`  implements Sub/Ord. I think using that makes the result easier to read
```suggestion
        let split_deletion_grace_period = shared_consts::split_deletion_grace_period();
        assert!(
            split_deletion_grace_period
                >= shared_consts::MINIMUM_DELETION_GRACE_PERIOD,
            "The split deletion grace period is too short ({split_deletion_grace_period:?}). This should not happen."
        );
        // We remove an extra margin of 2minutes from the split deletion grace period.
        split_deletion_grace_period - Duration::from_secs(60 * 2)
```

trinity-1686a (approved):

PR #5233, opened 2024-07-17
Pick ingester at random when adding shards.

fulmicoton:
# Removing bias from shard allocation.

We have noticed after running for a long time, in airmail the distribution of shard count amongst ingester seems uniform, but one or two indexers are getting most of the throughput.

This could be caused by an indirect bias in the allocation of shard to
ingester. For instance, in airmail, most indexes are very small, but a
few of them are much larger. Small indexes have 1 shard with a very low
throughput. Large indexes on the other hands have several shards with
typically >2MB of throughut.

Larger indexes are also more subject to scale up/down, since other
indexes tend to stick to having 1 shard (we don't scale down to 0).

This PR tries to remove any possible bias when assigning / removing
shards in
- scale up
- scale down
- rebalance.

Scale up
---------------------------

This is the most important change/bias.
In presence of a tie, we were picking the ingester with the
lowest ingester id.

Also, on replication, the logic picking a follower was broken
(for a given leader, we were always picking the same follower).

The `max_num_shards_to_allocate_per_node` was also wrong (division
instead of ceil div) (probably minor).

Scale down
----------------------------

The code was relying on the long term ingestion rate, and then
ties were solved by the hashmap iterator. The Hashmap iterator
is quite random so this was probably not a problem.

Rebalance
----------------------------

Arithmetic used to compute the target number of shards was a
little bit inaccurate.

The shard that are rebalanced are now picked at random (instead of
picking the oldest shards in the model).

### DIFF
--- a/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
+++ b/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
@@ -17,7 +17,8 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
-use std::collections::{BTreeSet, HashMap};
+use std::collections::btree_map::Entry;
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::fmt;
 use std::future::Future;
 use std::sync::Arc;
@@ -26,7 +27,6 @@ use std::time::Duration;
 use fnv::FnvHashSet;
 use futures::stream::FuturesUnordered;
 use futures::StreamExt;
-use itertools::Itertools;
 use quickwit_actors::Mailbox;
 use quickwit_common::pretty::PrettySample;
 use quickwit_common::Progress;
@@ -48,7 +48,10 @@ use quickwit_proto::metastore::{
     serde_utils, MetastoreResult, MetastoreService, MetastoreServiceClient, OpenShardSubrequest,
     OpenShardsRequest, OpenShardsResponse,
 };
-use quickwit_proto::types::{IndexUid, NodeId, Position, ShardId, SourceUid};
+use quickwit_proto::types::{IndexUid, NodeId, NodeIdRef, Position, ShardId, SourceUid};
+use rand::rngs::ThreadRng;
+use rand::seq::SliceRandom;
+use rand::{thread_rng, Rng, RngCore};
 use serde::{Deserialize, Serialize};
 use tokio::sync::{Mutex, OwnedMutexGuard};
 use tokio::task::JoinHandle;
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -444,67 +545,45 @@ impl IngestController {
...
+                {
+                    *shard_count += 1;
+                }
+                if let Some(follower_id) = shard.follower_id.as_ref() {
+                    if let Some(shard_count) =
+                        per_node_num_open_shards.get_mut(follower_id.as_str())
+                    {
+                        *shard_count += 1;
+                    }
                 }
```
fmassot, line 569: is it possible to have a shard not present in `per_node_num_open_shards`?
Should we log something when it happens?

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -444,67 +545,45 @@ impl IngestController {
...
-            let leader = leader_id.clone();
-            let mut follower_opt = None;
-
-            if self.replication_factor > 1 {
-                follower_opt = Some(follower_id.clone());
-            }
-            leader_follower_pairs.push((leader, follower_opt));
-        }
-        Some(leader_follower_pairs)
+        assert!(self.replication_factor == 1 || self.replication_factor == 2);
```
fmassot, line 573: why do we do this assert now? can we ensure this before? like when we build the controller?
Adding a comment on the variable to mention that would be nice too.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,6 +94,105 @@ fn fire_and_forget(
     });
 }
 
+// Returns a random position of the els `slice`, such that the element in this array is NOT
+// `except_el`.
+fn pick_position(els: &[&str], except_el_opt: Option<&str>, rng: &mut ThreadRng) -> Option<usize> {
+    let except_pos_opt =
+        except_el_opt.and_then(|except_el| els.iter().position(|el| *el == except_el));
+    if let Some(expect_pos) = except_pos_opt {
```
fmassot, line 102: Replace `expect_pos` by `except_pos`

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -1103,39 +1183,36 @@ pub(crate) struct RebalanceShardsCallback {
     pub rebalance_guard: OwnedMutexGuard<()>,
 }
 
-/// Finds the shard with the highest ingestion rate on the ingester with the most number of open
-/// shards. If multiple shards have the same ingestion rate, the shard with the lowest (oldest)
+/// Finds a shard on the ingester with the highest number of open
```
fmassot, line 1186: +1 for choossing randomly a shard instead of choosing the one with the highest ingestion rate.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
fulmicoton: It is normal if it is not there, and it does not require a log. I added a comment about that.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
fulmicoton: It is enforced before. The assert is here to be defensive and as documentation. An enum or a bool would have been nice.

Issue #3918, opened 2023-10-05
Mishandling of wildcard in query_string queries


esatterwhite:
**Describe the bug**
When using the `query_string` query in the elasticsearch compatible DSL, it seems the handling of trailing wild cards is different, or missing

Given the query
```json
        {
          "query_string": {
            "fields": [
              "app"
            ],
            "query": "plushies*",
            "lenient": true
          }
        }
```

Elasticsearch will return documents where the value of `app` is specifically `plushies` as it matches the `plushies*` query. Quick wit does not. I've tried the wildcard in several character positions, and only when omitted entirely will the expected results return

**Steps to reproduce (if applicable)**

1. insert a document when a field `app` with value `plushies`
2. attempt to match the document using a `query_string` with a query of `plushies*`

**Expected behavior**
Wild cards should be supported and match all documents that include the root value preceding the wildcard.

**Index Settings**
```json
{
  "field_mappings": [
    {
      "name": "_account",
      "type": "text",
      "fast": {
        "normalizer": "raw"
      },
      "fieldnorms": false,
      "indexed": true,
      "record": "basic",
      "stored": true,
      "tokenizer": "raw"
    },
    {
      "name": "_app",
      "type": "text",
      "fast": {
        "normalizer": "lowercase"
      },
      "fieldnorms": false,
      "indexed": true,
      "record": "basic",
      "stored": true,
      "tokenizer": "raw"
    },
    {
      "name": "_bid",
      "type": "text",
      "fast": {
        "normalizer": "lowercase"
      },
      "fieldnorms": false,
      "indexed": true,
      "record": "basic",
      "stored": true,
      "tokenizer": "raw"
    },
    {
      "name": "_cluster",
      "type": "text",
      "fast": {
        "normalizer": "raw"
      },
      "fieldnorms": false,
      "indexed": true,
      "record": "basic",
      "stored": true,
      "tokenizer": "raw"
    },
    {
      "name": "_env",
      "type": "text",
      "fast": {
        "normalizer": "lowercase"
      },
      "fieldnorms": false,
      "indexed": true,
      "record": "basic",
      "stored": true,
      "tokenizer": "raw"
    },
    {
      "name": "_file",
      "type": "text",
      "fast": {
        "normalizer": "lowercase"
      },
      "fieldnorms": false,
      "indexed": true,
      "record": "basic",
      "stored": true,
      "tokenizer": "raw"
    },
    {
      "name": "_hash",
      "type": "text",
[...]

PR #5251, opened 2024-07-24
wildcard casing

trinity-1686a:
fix normalizer used for the `raw_lowercase` tokenanizer

related to #3918

### DIFF
--- a/quickwit/quickwit-query/src/tokenizers/tokenizer_manager.rs
+++ b/quickwit/quickwit-query/src/tokenizers/tokenizer_manager.rs
@@ -29,6 +29,7 @@ use crate::DEFAULT_REMOVE_TOKEN_LENGTH;
 
 const RAW_TOKENIZER_NAME: &str = "raw";
 const LOWERCASE_TOKENIZER_NAME: &str = "lowercase";
+const RAW_LOWERCASE_TOKENIZER_NAME: &str = "raw_lowercase";
 
 #[derive(Clone)]
 pub struct TokenizerManager {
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-query/src/tokenizers/tokenizer_manager.rs
```diff
@@ -83,7 +89,7 @@ impl TokenizerManager {
             .get(tokenizer_name)
             .copied()?;
         let analyzer = if use_lowercaser {
-            LOWERCASE_TOKENIZER_NAME
+            RAW_LOWERCASE_TOKENIZER_NAME
```
trinity-1686a, line 92: the tokenizer here isn't really supposed to generate multiple tokens

quickwit/quickwit-search/src/root.rs
```diff
@@ -1207,6 +1207,11 @@ pub async fn search_plan(
             .terms_grouped_by_field
             .values()
             .map(|terms: &HashMap<tantivy::Term, bool>| terms.len())
+            .sum::<usize>()
+        + warmup_info
+            .term_ranges_grouped_by_field
```
trinity-1686a, line 1212: range queries (including wildcards) were not properly counted

fulmicoton:
Can we have a unit test somewhere?

fulmicoton (changes requested):
Please add a unit test. If it is difficult, an integration test will do.
Make sure it runs on elasticsearch if you go for the integration test.

fulmicoton (approved):

PR #5252, opened 2024-07-24
fix endless loop in ingest_v2

### LINKED ISSUE #5240: Infinite Loop in ingest_v2 When Record Payload Exceeds batch_num_bytes
**Describe the bug**

When attempting to use `ingest_v2`, I noticed that sometimes the CPU usage spikes, with one core being utilized at 100%. It appears to trigger some sort of infinite loop.

Upon debugging, I found that the issue occurs when a record payload slightly exceeds `batch_num_bytes` (I used the default configuration `DEFAULT_BATCH_NUM_BYTES=1 MiB`). 

Specifically, this issue is in `quickwit/quickwit-ingest/src/ingest_v2/fetch.rs`:
https://github.com/quickwit-oss/quickwit/blob/8e6dc179c6abc3f105e84637a07d159aa6470a36/quickwit/quickwit-ingest/src/ingest_v2/fetch.rs#L148-L151

Here, since the payload length exceeds `batch_num_bytes`, the `if` condition is met, and `has_drained_queue` is set to `false`. The program then loops back to the start:
https://github.com/quickwit-oss/quickwit/blob/8e6dc179c6abc3f105e84637a07d159aa6470a36/quickwit/quickwit-ingest/src/ingest_v2/fetch.rs#L127

Since `has_drained_queue` is `false`, the `if` condition fails, and the program proceeds to the subsequent code block, looping back again to:
https://github.com/quickwit-oss/quickwit/blob/8e6dc179c6abc3f105e84637a07d159aa6470a36/quickwit/quickwit-ingest/src/ingest_v2/fetch.rs#L148-L151

`has_drained_queue` is repeatedly set to `false`, resulting in an infinite loop.

**Steps to Reproduce (if applicable)**

    When the record payload is greater than `batch_num_bytes`, this issue is triggered.

I wrote a test to reproduce this issue:
   https://github.com/Stool233/quickwit/commit/33ea6145f89e79d7540a67c53063d50a32903d06
Here are the action run results:
   https://github.com/Stool233/quickwit/actions/runs/10021273109/job/27699744997#step:10:2061

**Expected behavior**

When encountering a record payload greater than `batch_num_bytes`, either temporarily increase the size of `batch_num_bytes` to handle the record, or reject the record instead of entering an infinite loop.

I wrote a patch that temporarily increases the size of `batch_num_bytes` to handle such records, which works in my scenario:
- Patch: https://github.com/Stool233/quickwit/commit/6ddfe56f2b0e1807a3028c34d9fcc7514586437b
- Action run results: https://github.com/Stool233/quickwit/actions/runs/10021272530

If the community finds it appropriate, I can submit a related PR.

**Configuration:**

Please provide:

   - Compiled from the latest main (8e6dc179c6abc3f105e84637a07d159aa6470a36)

2. The `index_config.yaml`

   QW_ENABLE_INGEST_V2=true

   ```yaml
   indexer:
     enable_cooperative_indexing: true
   ```

PSeitz:
Fix endless loop by temporarily increasing the buffer size.
Alternative would be to reject the message, but I'm not sure this should be done
here.

Based on proposed solution of @Stool233

Fixes #5240

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/fetch.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/fetch.rs
@@ -1226,7 +1226,7 @@ pub(super) mod tests {
             open_fetch_stream_request,
             mrecordlog.clone(),
             shard_status_rx,
-            10,
+            10, //< we request batch larger than 10 bytes.
         );
 
         let mut mrecordlog_guard = mrecordlog.write().await;
@@ -1244,6 +1244,7 @@ pub(super) mod tests {
             .append_records(
                 &queue_id,
                 None,
+                // This doc is longer than 10 bytes.
                 std::iter::once(MRecord::new_doc("test-doc-foo").encode()),
             )
             .await
--- a/quickwit/quickwit-ingest/src/ingest_v2/fetch.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/fetch.rs
@@ -145,7 +145,10 @@ impl FetchStreamTask {
                 break;
             };
             for Record { payload, .. } in mrecords {
-                if mrecord_buffer.len() + payload.len() > mrecord_buffer.capacity() {
+                // Accept at least one message
+                if !mrecord_buffer.is_empty()
+                    && (mrecord_buffer.len() + payload.len() > mrecord_buffer.capacity())
+                {
                     has_drained_queue = false;
                     break;
                 }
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-ingest/src/ingest_v2/fetch.rs
```diff
@@ -1199,6 +1202,80 @@ pub(super) mod tests {
...
+            source_id: source_id.clone(),
+            shard_id: Some(shard_id.clone()),
+            from_position_exclusive: Some(Position::Beginning),
+        };
+        let (shard_status_tx, shard_status_rx) = watch::channel(ShardStatus::default());
+        let (mut fetch_stream, _fetch_task_handle) = FetchStreamTask::spawn(
+            open_fetch_stream_request,
+            mrecordlog.clone(),
+            shard_status_rx,
+            10,
```
fulmicoton, line 1229:
```suggestion
            10, //< we request batch larger than 10 bytes.
```

quickwit/quickwit-ingest/src/ingest_v2/fetch.rs
```diff
@@ -1199,6 +1202,80 @@ pub(super) mod tests {
...
+            .await
+            .unwrap();
+
+        mrecordlog_guard
+            .as_mut()
+            .unwrap()
+            .append_records(
+                &queue_id,
+                None,
+                std::iter::once(MRecord::new_doc("test-doc-foo").encode()),
```
fulmicoton, line 1247:
```suggestion
                // This doc is longer than 10 bytes.
                std::iter::once(MRecord::new_doc("test-doc-foo").encode()),
```

PR #5257, opened 2024-07-25
update aws-sdks

PSeitz:
### How was this PR tested?

run rest-api-tests locally with minio

### DIFF
--- a/quickwit/quickwit-storage/src/lib.rs
+++ b/quickwit/quickwit-storage/src/lib.rs
@@ -113,17 +113,6 @@ pub async fn load_file(
     Ok(bytes)
 }
 
-// this function isn't meant to be called, just to break compilation if
-// serde_json::Map is an ordered map and not a btree map
-#[allow(dead_code)]
-#[cfg(not(any(test, feature = "testsuite", feature = "integration-testsuite")))]
-unsafe fn serde_json_preserve_order_canary(
-    val: serde_json::Map<String, serde_json::Value>,
-) -> std::collections::BTreeMap<String, serde_json::Value> {
-    use std::mem::transmute as assert_serde_json__preserve_order__disabled;
-    assert_serde_json__preserve_order__disabled(val)
-}
-
 #[cfg(any(test, feature = "testsuite", feature = "integration-testsuite"))]
 mod for_test {
     use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-storage/src/lib.rs
```diff
@@ -113,17 +113,6 @@ pub async fn load_file(
     Ok(bytes)
 }
 
-// this function isn't meant to be called, just to break compilation if
-// serde_json::Map is an ordered map and not a btree map
-#[allow(dead_code)]
-#[cfg(not(any(test, feature = "testsuite", feature = "integration-testsuite")))]
-unsafe fn serde_json_preserve_order_canary(
```
fulmicoton, line 120: is this a good thing?

PSeitz: This was added here https://github.com/quickwit-oss/quickwit/pull/4918/files#diff-ae5e0a4b8f9a4cd70442abaa503b12f78bf86fd614dcffd215f740e10678be3dR117. The update preserve order seems to have been added to our serde_json due to feature unification, which caused a test to fail.

I don't understand why there's a compilation time check for this. @trinity-1686a

fulmicoton: I think the point was to make sure that no dependency would break this.

trinity-1686a: this was added after [this discussion](https://github.com/quickwit-oss/quickwit/pull/4918#discussion_r1582692530). in particular, this was a way to make sure this statement wouldn't become false without us making a deliberate choice
> At first sight, it seems preserve_order is only enabled for tests. Do you happen to know a way to make sure this is the case?

PSeitz: It's ensuring that a dependency enables the preserve order feature, a feature which we don't enable ourselves. I don't get it, if we want the feature we would just enable it.

PSeitz: Ah I understood it incorrectly. We could pick a serde_json version which isn't affected by feature unification ...

fulmicoton: So if I understood you correctly, they used to enable it for test only, and now it is for everything?

PSeitz: hmm I re-enabled the check and the issue disappeared

rdettai:
works with the lambda distribution: https://github.com/quickwit-inc/terraform/actions/runs/10107393824/job/27951188416

rdettai (approved):
also tested with the Kinesis integration and works as well!

Issue #5084, opened 2024-06-05
support updating doc mapping


trinity-1686a:
this is a meta-ticket to group pull requests related to each others

- [x] add doc mapping version to mapper (https://github.com/quickwit-oss/quickwit/pull/5087)
- [x] add doc mapping version to split (https://github.com/quickwit-oss/quickwit/pull/5218)
- [x] don't merge incompatible splits (https://github.com/quickwit-oss/quickwit/pull/5218)
- [x] add api to update the doc mapping (https://github.com/quickwit-oss/quickwit/pull/5253)
- [x] plug doc mapping change to control-plan and restart indexing pipellines (https://github.com/quickwit-oss/quickwit/pull/5265)
- [x] fix fields disappearing if their type change int vs string (https://github.com/quickwit-oss/quickwit/pull/5076)
- [x] fix fields disappearing if their type change obj vs json (https://github.com/quickwit-oss/quickwit/pull/5281)
- [x] fix getting errors about missing positions when updating what's recorded (https://github.com/quickwit-oss/quickwit/pull/5289)
  - when done, remove the corresponding `#[ignore]` annotations in the integration tests (https://github.com/quickwit-oss/quickwit/pull/5266)
- [x] documentation (https://github.com/quickwit-oss/quickwit/pull/5290)

fulmicoton:
@trinity-1686a My understanding is that the only thing missing is the doc.Can you work on this (for the 0.9 release)

trinity-1686a:
> Just for confirmation, there aren't any CLI command for this, right?

there is a `quickwit index update` command. i haven't tested it, i think it was developed by @rdettai

PR #5265, opened 2024-07-26
restart indexing pipeline on index update

trinity-1686a:
### Description

restart indexing pipelines on index update
part of #5084 

### How was this PR tested?

todo


trinity-1686a:
this still needs automated tests, but should otherwise be ready for a first round of review

rdettai:
It would be great to add a test this in https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-integration-tests/src/tests/update_tests/doc_mapping_tests.rs

trinity-1686a:
thinking about it, as we only restart indexing pipelines, but don't do anything about shards, shards from ingest-v2 could have a stall doc_mapping, and if validation is enabled, they could reject documents that are valid under the new mapping. Ideally we'd want to also close all shards for the index and open as many new ones.
We would then end up with a race condition where we said "this document is valid" at validation, run it against the new mapper, and throw it away as invalid then, without notifying the writer.

fulmicoton (approved):

PR #5076, opened 2024-06-04
cast updated fields

trinity-1686a:
### Description

Following a docmapper update, we might end up with documents with a one type stored, while the docmapper expecting a different type, attempt to cast what we have to what we want.

### How was this PR tested?

completed unit tests, tested manually after updating an index

PR #5266, opened 2024-07-26
Add integration test for doc mapper update

rdettai:
### Description

Add some high level validation that searches still work after doc mapper update. 

It would be interesting to do the same with aggregations.

Part of https://github.com/quickwit-oss/quickwit/issues/5084

### How was this PR tested?

NA (these are tests)


rdettai (commented):
Any juicy scenario suggestion is welcome

trinity-1686a:
ideas of test i've run with various level of success:
- u64 -> string, verify the field is still present after update (initially broken, fixed by #5076)
- with dynamic mode: create index without field1, ingest data with field1, create that field1, ingest some more. Check we can search for both before and after.
- with strict mode: create index without a field1, ingest docs without field1, create field1, ingest data with this field. Check we don't get error from old splits because of searching on unknown field
- index string with `raw` tokenizer no position; ingest document containing `-` (dash) in value, such as `roller-coaster`. update tokenizer to default, with positions. ingest more documents. make a phrase query for that doc (this is expected to work)
- index string with `default` tokenizer no position; ingest document with `-` (dash) in value. update to enable positions. ingest more documents. make a phrase query for that doc. this one should work in the future, but is expected to fail currently

rdettai:
> u64 -> string, verify the field is still present after update 

Not sure I get this one. If the search on the field goes through, it necessarily means that the field is still present, no?

rdettai:
@trinity-1686a I added all your test scnarii

Because of your suggestion to test "u64 -> string, verify the field is still present after update", I changed the assert from just the number of hits to the whole hit content. I got 2 surprising results:
- `test_update_doc_mapping_json_to_object`: I update a json type to an explicit object type, and the fields are dropped in the resulting hit.
- `test_update_doc_mapping_strict_to_dynamic`: I switch from an explicit (strict) schema to a dynamic one, and the fields are dropped in the resulting hit.

trinity-1686a:
that's interesting, can you add these test as disabled, and i'll try to make them pass

rdettai:
done

trinity-1686a (approved):

PR #5269, opened 2024-07-27
Upgrade hyper and tonic except for aws-sdk and server.

fmassot:


rdettai:
https://github.com/quickwit-oss/quickwit/pull/5777

PR #5272, opened 2024-07-27
Avoid retrying right away on shards that have been rate limited in the

### LINKED ISSUE #5271: Smarter retry on ingest v2
right now in the router/workbench we retry 3 times regardless of the number of available shard / number of available target node.

fulmicoton:
same workbench.

Closes #5271

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
@@ -27,7 +27,7 @@ use quickwit_proto::ingest::router::{
     IngestFailure, IngestFailureReason, IngestResponseV2, IngestSubrequest, IngestSuccess,
 };
 use quickwit_proto::ingest::{IngestV2Error, IngestV2Result, RateLimitingCause};
-use quickwit_proto::types::{NodeId, SubrequestId};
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -51,6 +52,28 @@ pub(super) struct IngestWorkbench {
...
+        } else {
+            None
+        }
+    });
+    // We sort the output in tests, to make the output deterministic.
+    #[cfg(test)]
+    {
+        use itertools::Itertools;
+        pending_subrequests_iterator.sorted_by_key(|subrequest| subrequest.subrequest_id)
+    }
```
trinity-1686a, line 70: I'm not sure this is required (or rather, don't understand why it would be), the iterator should already be sorted because of the BTreeMap.

fwiw i ran the tests a few time without that, and didn't get any error related to this

trinity-1686a (approved):

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
fulmicoton: oh yes you are right!

PR #5276, opened 2024-07-29
lambda: Allow setting postgres as the metastore

tontinton:
Give the user the full control over the lambda metastore uri.

Backwards compatible.

### DIFF
--- a/distribution/lambda/README.md
+++ b/distribution/lambda/README.md
@@ -85,7 +85,7 @@ simplify the setup and avoid unstable deployments.
 | Variable | Description | Default |
 |---|---|---|
 | QW_LAMBDA_INDEX_ID | the index this Lambda interacts with (one and only one) | required |
-| QW_LAMBDA_METASTORE_BUCKET | bucket name for metastore files | required |
+| QW_LAMBDA_METASTORE_URI | [Metastore URI](https://quickwit.io/docs/configuration/metastore-config) | required |
 | QW_LAMBDA_INDEX_BUCKET | bucket name for split files | required |
 | QW_LAMBDA_OPENTELEMETRY_URL | HTTP OTEL tracing collector endpoint | none, OTEL disabled |
 | QW_LAMBDA_OPENTELEMETRY_AUTHORIZATION | Authorization header value for HTTP OTEL calls | none, OTEL disabled |
@@ -106,7 +106,6 @@ Indexer only:
 Searcher only:
 | Variable | Description | Default |
 |---|---|---|
-| QW_LAMBDA_SEARCHER_METASTORE_POLLING_INTERVAL_SECONDS | refresh interval of the metastore | 60 |
 | QW_LAMBDA_PARTIAL_REQUEST_CACHE_CAPACITY | `searcher.partial_request_cache_capacity` node config | 64M |
 
 
--- a/distribution/lambda/cdk/stacks/services/indexer_service.py
+++ b/distribution/lambda/cdk/stacks/services/indexer_service.py
@@ -20,6 +20,12 @@ class IndexerService(Construct):
     ) -> None:
         super().__init__(scope, construct_id, **kwargs)
 
+        metastore_uri = environment.get("QW_LAMBDA_METASTORE_URI")
+        if not metastore_uri:
+            # Backwards compatibility.
+            metastore_bucket = environment.get("QW_LAMBDA_METASTORE_BUCKET", store.bucket)
+            metastore_uri = f"s3://${metastore_bucket}/index"
+
         self.lambda_function = aws_lambda.Function(
             self,
             id="Lambda",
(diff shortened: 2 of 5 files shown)


rdettai:
Thanks @tontinton for your contribution!

The reason why we forced the metastore to be S3 for now is because:
- a Posgres instance is pretty much the absolute opposite of Lambda in terms of scalability. In particular, it doesn´t scale well to 0
-  setting up Lambda to work properly with long lived and VPC based Postgres connections requires a bit of extra work

Before offering this feature, I think we would need to:
- implement and test a CDK script that sets up Quickwit Lambda functions that can connect to a database on VPC (e.g RDS)
  - maybe we can skip this step if we consider that publicly exposed databases  over SSL are now a "common enough" thing (Neon, Supabase Postgres...)
- test how the Lambda function warm start behaves after a period of inactivity greater than the Postgres timeout
- assess whether we should recommend the [RDS proxy](https://docs.aws.amazon.com/lambda/latest/dg/services-rds.html)

What kind of tests have you performed on your end? Have you tried to run your change on a long running example? Can you tell us a bit more about your usecase?

tontinton:
I don't care about having a Postgres that scales to 0 (although I might migrate to [neon](https://neon.tech/) in the future).

I just want to be able to have multiple of these lambdas running concurrently and not have data races.

I'm currently trying to have a setup where I use this lambda to run on every new S3 file, but also to have it run on already stored files in S3 (a bootstrap indexing pipeline of sorts), basically running a pool of these lambdas on old files, meaning I can't use the file backed metastore.

About RDS and proxies, I currently use AWS Aurora. I can easily point to it from my lambda, I just need this change of course to be able to do it.

tontinton:
Also, I just remembered Aurora also enables scaling to 0: https://aws.amazon.com/rds/aurora/serverless/

So why not allow for this lambda to run with postgres?

rdettai:
> I'm currently trying to have a setup where I use this lambda to run on every new S3 file, but also to have it run on already stored files in S3 (a bootstrap indexing pipeline of sorts), basically running a pool of these lambdas on old files, meaning I can't use the file backed metastore.

Interesting use case! One small remark: I don't know how big your Lambda pool will be, but keep in mind that the Lambda compute is very expensive (10x an EC2 instance). So if you want to process 100s of GBs, you would probably get away for a fraction of the price by spinning up a cluster for your backfill. We are about to publish a new file source that uses SQS based S3 notifications. We could very well imagine a hybrid setup where your backfill is performed by an ECS cluster but your continuous low volume ingestion is done by AWS Lambda (this would still require Postgres to enable the concurrency between the writers)

> About RDS and proxies, I currently use AWS Aurora. I can easily point to it from my lambda, I just need this change of course to be able to do it.

If I understand correctly, you are connecting to a publicly exposed Aurora instance without RDS Proxy. Have you performed more sustained tests on this? My concern is about Postgres connections with Lambda containers that were paused for a couple of minutes (warm starts), and how the metastore client recovers in that case.

rdettai:
Also, the Lambda feature is in beta, so I would rather avoid retro compatibility dis-junctions and make a breaking change if we decide to go in this direction.

tontinton:
I see, you say I should create a temporary cluster for the backfill, and that in the near future we'll be able to simply point quickwit to an SQS stream of events instead of using this lambda.

Very nice! Looking forward to the release.

I haven't tested my setup yet, still in the design stage, but your concern on connections being kept alive is valid. Couldn't we set a connection timeout to the metastore with quickwit to solve this? Maybe this is also a change that should be made?

Also is it really a problem? I'm fine with having multiple connections alive to my postgres currently.

Lastly, I want to ask you: What do you think is the simplest way to solve the backfill problrm until the SQS release? I thought the lambda pool was simple (albeit a bit expensive), maybe the ECR solution is the simplest? How do I configure it to index the old files?

rdettai:
> Shouldn't Quickwit reopen the connection on a cold start?

On cold start, yes. On a warm start the behavior should be tested.

> What do you think is the simplest way to solve the backfill problem until the SQS release?

This is a bit off topic for this issue, but right now we don´t have a straightforward way to do it. I think the easiest would be to use the [local ingest](https://quickwit.io/docs/ingest-data/ingest-local-file)  (using an S3 URI as file is not well documented but should work). You would need to come up with a simple script that runs a controlled number of them in parallel using a thread pool (a few lines of Python with `concurrent.futures.ThreadPoolExecutor` and `subprocess` should do the trick). Running one ingest process every 4 cores should give you the best hardware usage (provided that you are in the same region as your S3 files).

tontinton:
Ok thanks, I'll close the PR for now, and if I'll get to test the warm start, I'll reopen without backwards compatibility.

rdettai:
note: I tested it and the support for S3 URIs from the CLI is not supported in the current release. To ingest files from s3 you would need to copy them to your local filesystem in the script first. It will be supported in the new release!

Issue #5209, opened 2024-07-09
Plug OTLP to ingest v2


guilload:
Currently using ingest v1.

Issue #5068, opened 2024-06-03
decommissioning ingest_v2 hangs


PSeitz:
When trying to quit quickwit, it occasionally hangs at decommissioning ingester. Only `kill -9` works
```
^C2024-06-03T09:35:15.935Z  INFO quickwit_ingest::ingest_v2::ingester: decommissioning ingester
^C
```

I couldn't reproduce it, seems to happen randomly.

guilload:
If there's data inflight, have you waited long enough for the next commit? It can take up to `commit_timeout_secs + ε.`

PSeitz:
In some cases I didn't ingest data, so there shouldn't be any data inflight. I also didn't enable ingestv2 via `QW_ENABLE_INGEST_V2`.

guilload:
Here is the bug.

Executing the REST API tests against a running node that does not have ingest V2 enabled still creates a shard and drops a few records in it. The shard does not get indexed and is not cleaned up. The next time you try to decommission the node without v2 enabled, we wait for the shard to be drained, which will never happen.

rdettai:
> Executing the REST API tests against a running node that does not have ingest V2 enabled still creates a shard and drops a few records in it. The shard does not get indexed and is not cleaned up. The next time you try to decommission the node without v2 enabled, we wait for the shard to be drained, which will never happen.

I don´t think this is the only issue. See analysis in https://github.com/quickwit-oss/quickwit/pull/5283.

rdettai:
In https://github.com/quickwit-oss/quickwit/pull/5283 I added 2 integration tests covering the two issues mentioned above:
- `ingest_tests::test_shutdown_control_plane_early_shutdown`
- `test_shutdown_separate_indexer`

Both pass on ingest V1 and fail if we enable ingest V2

rdettai:
Some extra docs: https://github.com/quickwit-oss/quickwit/pull/5418

PR #5283, opened 2024-07-31
Switch OTLP endpoints to ingest v2

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/5209

### How was this PR tested?

Integration tests


rdettai:
Integration tests are hanging for OTLP are hanging on shutdown 
```
[INFO  quickwit_ingest::ingest_v2::ingester] decommissioning ingester
```

Likely related to https://github.com/quickwit-oss/quickwit/issues/5068

rdettai:
After some further analysis, it seems that the ingester decomissioning freeze happens when we stop the cluster right after the first ingestions. Likely race condition:
- indexer has no shard, no ingest v2 indexing pipeline is scheduled
- indexer receives events, creates a shard
- the control plane is shut down before rebuilding the indexing plan, so the pipeline is never scheduled for the shard
- the indexer hangs on shutdown because the shard is never indexed

rdettai:
Adding a wait after the ingest to give time to the control plane to schedule an indexing pipeline works but:
- the indexing pipeline sometimes becomes irresponsive (no extra logs)
- when the indexing goes through, the decommissioning of the indexer still fails. Tries to create a workbench right in the middle of the teardown:
```
[INFO  quickwit-indexing] index-doc-batches; index_id=otel-logs-v0_7 source_id=_ingest-source pipeline_uid=01J47CHTZPH4SX2J4D6HSVY5RH workbench_id=01J47CJDJEAQBP1NHP0GJTGDMP
```

rdettai:
Note: the issue mentioned above
> on some runs, the indexing pipeline becomes irresponsive (no extra logs)

was not addressed. I didn't manage to consistently reproduce or get more logs.

quickwit/quickwit-jaeger/src/lib.rs
```diff
@@ -60,8 +60,8 @@ use tracing::{debug, error, instrument, warn, Span as RuntimeSpan};
 
 use crate::metrics::JAEGER_SERVICE_METRICS;
 
-#[cfg(test)]
-mod integration_tests;
```
rdettai, line 64: just remembered I disabled this test. This needs to be fixed before merging

quickwit/quickwit-integration-tests/src/tests/otlp_tests.rs
```diff
@@ -17,100 +17,245 @@
...
+    let mut sandbox = ClusterSandbox::start_cluster_with_otlp_service(&nodes_services)
+        .await
+        .unwrap();
+    // Wait fo the pipelines to start (one for logs and one for traces)
+    sandbox.wait_for_indexing_pipelines(2).await.unwrap();
+
+    let export_trace_request = ExportTraceServiceRequest {
+        resource_spans: make_resource_spans_for_test(),
+    };
+    // let span_count = export_trace_request.resource_spans.len();
```
fmassot, line 237: to remove

quickwit/quickwit-opentelemetry/src/otlp/logs.rs
```diff
@@ -301,21 +305,27 @@ impl OtlpGrpcLogsService {
...
         request: ExportLogsServiceRequest,
         parent_span: RuntimeSpan,
-        index_id: IndexId,
     ) -> Result<ParsedLogRecords, Status> {
         let (log_records, mut num_parse_errors) = parse_otlp_logs(request)?;
         let num_log_records = log_records.len() as u64;
         let mut error_message = String::new();
 
-        let mut doc_batch = DocBatchBuilder::new(index_id).json_writer();
+        let mut doc_batch_builder = DocBatchV2Builder::default().json_writer();
```
fmassot, line 313: Getting a `JsonDocBatchV2Builder` from `DocBatchV2Builder::default().json_writer()` is not straightforward.

Why not having `let json_doc_batch_builder = JsonDocBatchV2Builder::default();`?

quickwit/quickwit-ingest/src/ingest_v2/mod.rs
```diff
@@ -139,6 +141,41 @@ impl DocBatchV2Builder {
...
+        let old_len = self.doc_buffer.get_ref().len();
+        serde_json::to_writer(&mut self.doc_buffer, &payload)?;
+        let new_len = self.doc_buffer.get_ref().len();
+        let written_len = new_len - old_len;
+        self.doc_uids.push(doc_uid);
+        self.doc_lengths.push(written_len as u32);
+        Ok(())
+    }
+
+    pub fn into_inner(self) -> DocBatchV2Builder {
```
fmassot, line 172: I would not call that `into_inner`. We could use `DocBatchV2Builder::from` (and implement the trait)

quickwit/quickwit-opentelemetry/src/otlp/mod.rs
```diff
@@ -220,6 +222,21 @@ pub(crate) fn extract_otel_index_id_from_metadata(
     Ok(index_id.to_string())
 }
 
+fn ingest_failures_to_error(mut response: IngestResponseV2) -> Result<(), IngestServiceError> {
```
fmassot, line 225: The contract of this function is weird. Why should I expect it to fail if there is one single failure/success?

I'm ok for implementing a helper but it should be clear for the user that it is handling only one subrequest. Adding docs will help too.

quickwit/quickwit-opentelemetry/src/otlp/traces.rs
```diff
@@ -753,22 +760,28 @@ impl OtlpGrpcTracesService {
...
         let mut num_parse_errors = 0;
         let mut error_message = String::new();
 
-        let mut doc_batch_builder = DocBatchBuilder::new(index_id).json_writer();
+        let mut doc_batch_builder = DocBatchV2Builder::default().json_writer();
+        let mut doc_uid_generator = DocUidGenerator::default();
         for span in spans {
-            if let Err(error) = doc_batch_builder.ingest_doc(&span.0) {
+            let doc_uid = doc_uid_generator.next_doc_uid();
+            if let Err(error) = doc_batch_builder.add_doc(doc_uid, span.0) {
```
fmassot, line 773: idea: we could have a `add_json_doc` function on the `DocBatchV2Builder`.

fmassot (changes requested):
See comments.

quickwit/quickwit-opentelemetry/src/otlp/logs.rs
rdettai: This was a 1 to 1 port of the V1 `DocBatchBuilder`, but I agree it could be simpler.

quickwit/quickwit-opentelemetry/src/otlp/traces.rs
rdettai: I copied over the `JsonDocBatchBuilder` logic from V1. To use `DocBatchV2Builder` directly, we would either:
- need to first serialize the struct then add thoses bytes to the `BytesMut`, but that means an extra copy
- also use a `Writer<BytesMut>` in `DocBatchV2Builder`, but I assumed it was not 0 cost otherwise it would have been done that way in V1

rdettai: Following the other review comments, I simplified the `JsonDocBatchV2Builder` as much as possible by avoiding the unnecessary forth and back with `DocBatchV2Builder`. I would prefer not to unify the two structures without ensuring that there isn't a performance impact as this is on the critical write path. What do you think?

quickwit/quickwit-opentelemetry/src/otlp/mod.rs
rdettai: I factorized a bit more code to get rid of this weird function.

quickwit/quickwit-opentelemetry/src/otlp/traces.rs
fmassot: fine for me!

quickwit/quickwit-opentelemetry/src/otlp/mod.rs
```diff
@@ -220,6 +225,40 @@ pub(crate) fn extract_otel_index_id_from_metadata(
     Ok(index_id.to_string())
 }
 
+async fn store_helper(
```
fmassot, line 228: I would call it `ingest_doc_batch`, `store_helper` is confusing

fmassot (approved):

PR #5288, opened 2024-08-01
Minor refactoring in leaf search.

fulmicoton:
Recording failed splits even if task panicked.
Avoiding to use `join_all` on a Vec of task's JoinHandle.

### DIFF
--- a/quickwit/quickwit-search/src/leaf.rs
+++ b/quickwit/quickwit-search/src/leaf.rs
@@ -44,6 +44,7 @@ use tantivy::directory::FileSlice;
 use tantivy::fastfield::FastFieldReaders;
 use tantivy::schema::Field;
 use tantivy::{DateTime, Index, ReloadPolicy, Searcher, Term};
+use tokio::task::JoinError;
 use tracing::*;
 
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1236,44 +1238,58 @@ pub async fn leaf_search(
...
+            if join_error.is_cancelled() {
+                // An explicit task cancellation is not an error.
+                continue;
+            }
+            if join_error.is_panic() {
+                error!(split=%split, "leaf search task panicked");
+            } else {
+                error!(split=%split, "please report: leaf search was not cancelled, and could not extract panic. this should never happen");
+            }
+            split_search_join_errors.push((split, join_error));
```
trinity-1686a, line 1277: is there a reason to not call `incremental_merge_collector.add_failed_split()` directly from here?

fulmicoton: yes. We get the incremental_merge_collector by unwrapping the Arc.
The unwrap will only work if all the other references have been dropped. 

So this is loop is both to wait for all of the task to terminate (successfully or not), and to gather errors.

trinity-1686a (approved):

PR #5289, opened 2024-08-01
don't error on phrase query schema error for old splits

trinity-1686a:
### Description

part of #5084 
don't execute phrase queries on old splits if they miss positions

### How was this PR tested?

re-enabled tests


rdettai (approved):
This fixes the bug (so approving the PR), but we should add unit tests to cover these scenarii and more. In the various  files of the `query_ast` module there are some unit tests that could be extended. For instance in `full_text_query.rs` you have tests for some successful TantivyQueryAst builds. We should extend these against "all" possible schema field types (expected to be valid or not).

PR #5290, opened 2024-08-01
start adding documentation on updating mapping

trinity-1686a:
### Description

add documentation on the known limitations when updating a doc mapping

@rdettai i know you were surprised by a few things, please step in if you think I missed anything important
@fmassot you might have opinions on things to add too.

this is based on the expect stat of updates once all currently existing PR on the subject are merged


docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,42 @@
+# Updating the doc mapping of an index
+
+Quickwit allows updating the mapping it uses, to add more fields to an existing index, or change how they are indexed. In doing so,
+it does not reindex existing data, still lets you search through older documents where possible.
+
+## Indexing
+
+When receiving a new doc mapping for an index, Quickwit will restart indexing pipelines to take into account the changes. This operation
```
fmassot, line 8: You mean "When you update a doc mapping"?

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,42 @@
...
+this usually happens very fast, possibly while some documents sent before the udpate haven't been commited yet. This means you can't
+assume documents sent before the update won't be indexed using the new doc mapping.
+
+## Querying
+
+When receiving a query, Quickwit always validate that query against the most recent mapping. If a query was valid under a previous
+mapping but is not compatible with the newer mapping, that query will be rejected. For instance if a field which was indexed no longer
+is, or a field was removed, any query that uses it will become invalid. On the other hand, if a query was not valid for a previous doc
+mapping, but is valid under the new doc mapping, Quickwit will process the query. When querying newer splits, it will behave normally,
+when querying older splits, it will try ot execute the query as correctly as possible. If you find a situation where older splits
```
fmassot, line 19: Nitpick: I would remove the sentence `If you find a situation where older splits...`.

I would insist in the previous sentence that quickwit "will try ot execute the query as correctly as possible without returning errors"

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,42 @@
...
+causes a valid request to return an error, please open a bug report.
+
+For instance, if a field was added or newly marked as indexed, quickwit will behave as if no older document had the right value. This
```
fmassot, line 22: It would be nice to have a table to describe the change in one column and the behavior in a second column.

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,42 @@
...
+## Indexing
+
+When receiving a new doc mapping for an index, Quickwit will restart indexing pipelines to take into account the changes. This operation
+is asynchronous. You can't assume documents send immediately after the update will necessarily use the new mapping. On the other hand,
+this usually happens very fast, possibly while some documents sent before the udpate haven't been commited yet. This means you can't
+assume documents sent before the update won't be indexed using the new doc mapping.
+
+## Querying
+
+When receiving a query, Quickwit always validate that query against the most recent mapping. If a query was valid under a previous
```
rdettai, line 15:
```suggestion
When receiving a query, Quickwit always validates it against the most recent mapping. If a query was valid under a previous
```

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,42 @@
...
+mapping but is not compatible with the newer mapping, that query will be rejected. For instance if a field which was indexed no longer
+is, or a field was removed, any query that uses it will become invalid. On the other hand, if a query was not valid for a previous doc
```
rdettai, line 17: I think a concrete example with mapping snippets might help the reader here

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,42 @@
...
+to an object. If a field goes from being an array, to being single-valued, only the first value will be returned, as a direct value.
+When converting an array field to be a part of a json field, Quickwit will decide per document and per field to show it as a direct
+value (if it contained exactly one element for this document), or an array if it contained any other number of values.
+
+## Reversibility
+
+Quickwit doesn't store old doc mappings, it can't revert an update all by itself. However it also does not modify existing data when
+receiving a new doc mapping. If you realize you updated the mapping in a way that's very wrong, you can re-update using the previous
+mapping. Documents indexed while the mapping was wrong will be impacted, but any document that was commited before the change will be
+back to its original state.
```
rdettai, line 42:
```suggestion
Quickwit does not modify existing data when receiving a new doc mapping. If you realize that you updated the mapping in a wrong way, you can re-update your index using the previous mapping. Documents indexed while the mapping was wrong will be impacted, but any document that was committed before the change will be queryable as if nothing happened.
```

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,42 @@
...
+documents, and will behave normally for newer.
+
+If you remove stored fields, the documents returned from a query will instantly have that field hidden, like if that field never was
+stored. If you change the type of a field, Quickwit will attempt to convert the old document format into the newer one. If it can't,
+it will show the field as absent. Some conversions are trivially always correct, for instance converting from integer to text. Some are
+dependant on your data, for instance converting text to integer. Finally some will always hide the fields, for instance converting text
+to an object. If a field goes from being an array, to being single-valued, only the first value will be returned, as a direct value.
+When converting an array field to be a part of a json field, Quickwit will decide per document and per field to show it as a direct
+value (if it contained exactly one element for this document), or an array if it contained any other number of values.
+
```
rdettai, line 36: We should also specify how tokenizer changes are handled.

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
```diff
@@ -640,6 +640,11 @@ fn value_to_bool(value: TantivyValue) -> Result<JsonValue, TantivyValue> {
             1 => Some(true),
             _ => None,
         },
+        TantivyValue::F64(number) => match number {
+            0.0 => Some(false),
+            1.0 => Some(true),
+            _ => None,
+        },
```
trinity-1686a, line 647: i realized while writing the doc that missing this conversion was odd given u64/i64 had it, and bool->f64 was a thing

docs/reference/rest-api.md
```diff
@@ -318,12 +318,10 @@ Updates the configurations of an index. This endpoint follows PUT semantics, whi
 
 - The retention policy update is automatically picked up by the janitor service on its next state refresh.
 - The search settings update is automatically picked up by searcher nodes when the next query is executed.
-- The indexing settings update is not automatically picked up by the indexer nodes, they need to be manually restarted.
-- The doc mapping update is not automatically picked up by the indexer nodes, they have to be manually restarted.
+- The indexing settings update is automatically picked up by the indexer nodes once the control plane emit a new indexing plan.
```
guilload, line 321:
```suggestion
- The indexing settings update is automatically picked up by the indexer nodes once the control plane emits a new indexing plan.
```

docs/reference/rest-api.md
```diff
@@ -318,12 +318,10 @@ Updates the configurations of an index. This endpoint follows PUT semantics, whi
...
+- The doc mapping update is automatically picked up by the indexer nodes once the control plane emit a new indexing plan.
 
-Updating the doc mapping doesn't reindex existing data. Queries and answers are mapped on a best effort basis when querying older splits.
-It is also not possible to update the timestamp field, or to modify/remove existing non-default tokenizers (but it is possible to change
-which tokenizer is used for a field).
+Updating the doc mapping doesn't reindex existing data. Queries and answers are mapped on a best effort basis when querying older splits. For more details, check [the reference](updating-mapper.md)
```
guilload, line 324:
```suggestion
Updating the doc mapping doesn't reindex existing data. Queries and results are mapped on a best-effort basis when querying older splits. For more details, check [the reference](updating-mapper.md) out.
```

quickwit/quickwit-doc-mapper/src/default_doc_mapper/mapping_tree.rs
guilload: Good catch.

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,179 @@
+# Updating the doc mapping of an index
+
+Quickwit allows updating the mapping it uses, to add more fields to an existing index, or change how they are indexed. In doing so,
+it does not reindex existing data, still lets you search through older documents where possible.
```
guilload, line 4:
```suggestion
Quickwit allows updating the mapping it uses to add more fields to an existing index or change how they are indexed. In doing so, it does not reindex existing data but still lets you search through older documents where possible.
```

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Indexing
+
+When you update a doc mapping for an index, Quickwit will restart indexing pipelines to take the changes into account. As both this operation and the document ingestion are asynchronous, you can't assume documents sent immediately after the update will necessarily use the new mapping nor that documents sent immediately before the update won't be indexed using the new doc mapping.
```
guilload, line 8:
```suggestion
When you update a doc mapping for an index, Quickwit will restart indexing pipelines to take the changes into account. As both this operation and the document ingestion are asynchronous, you can't assume documents sent immediately after the update will necessarily use the new mapping or documents sent immediately before the update will be indexed using the new doc mapping.
```
Removing the double negative.

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Querying
+
+When receiving a query, Quickwit always validate it against the most recent mapping.
```
guilload, line 12:
```suggestion
Quickwit always validate queries against the most recent mapping.
```

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,179 @@
...
+If a query was valid under a previous mapping but is not compatible with the newer mapping, that query will be rejected.
+For instance if a field which was indexed no longer is, any query that uses it will become invalid.
+On the other hand, if a query was not valid for a previous doc mapping, but is valid under the new doc mapping, Quickwit will process the query.
+When querying newer splits, it will behave normally, when querying older splits, it will try ot execute the query as correctly as possible.
```
guilload, line 16:
```suggestion
When querying newer splits, it will behave normally, when querying older splits, it will try to execute the query as correctly as possible.
```

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,179 @@
...
+If you find a situation where older splits causes a valid request to return an error, please open a bug report.
```
guilload, line 17:
```suggestion
If you find yourself in a situation where older splits causes a valid request to return an error, please open a bug report.
```

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,179 @@
...
+See example 1 and 2 below for clarification.
```
guilload, line 18:
```suggestion
See examples 1 and 2 below for clarification.
```

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,179 @@
...
+
+Change in tokenizer affect only newer splits, older splits keep using the tokenizers they were created with.
+
+Document retrieved are mapped from Quickwit internal format to JSON based on the latest doc mapping. This means if fields are deleted,
+they will stop appearing (see also Reversibility below) unless mapper mode is Dynamic. If the type of some field changed, it will be converted on a best effort basis:
```
guilload, line 23:
```suggestion
they will stop appearing (see also Reversibility below) unless mapper mode is Dynamic. If the type of some field changed, it will be converted on a best-effort basis:
```

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,179 @@
...
+integers will get turned into text, text will get turned into string when it is possible, otherwise, the field is omited.
+See example 3 for clarification.
+
+## Reversibility
+
+Quickwit does not modify existing data when receiving a new doc mapping. If you realize that you updated the mapping in a wrong way, you can re-update your index using the previous mapping. Documents indexed while the mapping was wrong will be impacted, but any document that was committed before the change will be queryable as if nothing happened.
+
+## Type update reference
+
+Conversion from a type to itself is omited. Conversion which never succeed and always omit the field are omited too.
```
guilload, line 33:
```suggestion
Conversion from a type to itself is omitted. Conversions that never succeed and always omit the field are omitted, too.
```

docs/reference/updating-mapper.md
```diff
@@ -0,0 +1,179 @@
...
+| f64 | datetime | parse according to current input\_format, otherwise omit |
+| array\<T\> | array\<U\> | convert individual elements, skipping over those which can't be converted |
+| T | array\<U\> | convert element, emiting array of a single element, or empty array if it can't be converted |
+| array\<T\> | U | convert individual elements, keeping the first which can be converted |
+| json | object | try convert individual elements if they exists inside object, omit individual elements which can't be |
+| object | json | convert individual elements. Previous lists of one element are converted to a single element not in an array.
+
+## Examples
+
+In all exemples, fields which are not relevant are removed for conciseness, you will not be able to use these index config as is.
```
guilload, line 67:
```suggestion
In the below examples, fields which are not relevant are removed for conciseness, you will not be able to use these index config as is.
```

docs/reference/updating-mapper.md
trinity-1686a: i think won't was what i meant, if you send a document while there is a large queue in the wal, and then update the mapping, it is possible for the pipeline to be restarted before reaching that document, so it may be indexed with the newer mapping (or it may not, if the wal was short/empty).

PR #5296, opened 2024-08-02
improve ingest lock

trinity-1686a:
- add metric for how long a lock was held
- when a log is held for too long, log which purpose it had
- when waiting for too long, log why we wanted to acquire the lock
- await inside the macro to force awaiting at the right time (currently, we sometime await outside of the macro, meaning we record an instant access to the lock, and then actually proceed to acquire the lock). I'm not a fan of awaiting inside a macro as it's hidden control flow, but i consider it worth the easy-to-miss mistake

### DIFF
--- a/quickwit/quickwit-ingest/src/lib.rs
+++ b/quickwit/quickwit-ingest/src/lib.rs
@@ -125,6 +125,49 @@ impl CommitType {
     }
 }
 
+struct TimedMutexGuard<T> {
+    guard: T,
+    acquired_at: std::time::Instant,
+    purpose: [&'static str; 2],
+}
+
+use std::ops::{Deref, DerefMut};
+
+impl<T> Deref for TimedMutexGuard<T> {
+    type Target = T;
+
+    fn deref(&self) -> &T {
+        &self.guard
+    }
+}
+
+impl<T> DerefMut for TimedMutexGuard<T> {
+    fn deref_mut(&mut self) -> &mut T {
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-ingest/src/lib.rs
```diff
@@ -125,6 +125,49 @@ impl CommitType {
     }
 }
 
+struct TimedMutexGuard<T> {
+    guard: T,
+    acquired_at: std::time::Instant,
+    purpose: [&'static str; 2],
```
fulmicoton, line 131: one of the sad things with relying on dynamic labels like this, is that you end up doing one hashmap lookup very time you need to access the counter (here once per lock).

It is probably not catastrophic in this case, but here is the alternative.

have a histogram: directly on the `TimedMutexGuard`

```
histogram: &'static Histogram,
```

You can then cache the Histogram counter in different ways. In ingest v2 you can see the pattern done here https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-ingest/src/ingest_v2/metrics.rs#L59-L74.

trinity-1686a: i like this pattern of statically defining things, however it's a kind of a pain, and a large typo hazard to implement. Having some sort of macro to derive "Build a static (histogram|gauge|counter) vec from a dynamic vec" would be nice, and would likely make us use this more

PR #5297, opened 2024-08-02
Fix OTLP timestamps

rdettai:
### Description

OTLP timestamp are multiplied by 10^9 before being stored, even though they should already be in nano format.

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-opentelemetry/src/otlp/logs.rs
+++ b/quickwit/quickwit-opentelemetry/src/otlp/logs.rs
@@ -60,6 +60,8 @@ doc_mapping:
     - name: timestamp_nanos
       type: datetime
       input_formats: [unix_timestamp]
+      # should be this but not supported
+      # input_formats: [unix_timestamp_nanos]
       output_format: unix_timestamp_nanos
       indexed: false
       fast: true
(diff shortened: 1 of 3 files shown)


rdettai (commented):
The first commit just shows the problem. The integration test I added for showcasing the issue returns:
```
thread 'tests::index_tests::test_ingest_logs_with_otlp_grpc_api' panicked at quickwit-integration-tests/src/tests/index_tests.rs:794:9:
assertion `left == right` failed
  left: 1000000001000000000
 right: 1000000001
```

quickwit/quickwit-opentelemetry/src/otlp/logs.rs
```diff
@@ -60,13 +60,17 @@ doc_mapping:
     - name: timestamp_nanos
       type: datetime
       input_formats: [unix_timestamp]
+      # should be this but not supported
```
fmassot, line 63: Not relevant. `unix_timestamp` means quickwit detect automatically if it's in seconds or something else.

rdettai:
ok, the problem was in the test, the dummy value provided was not detected as a nano timestamp

PR #5304, opened 2024-08-08
limit search thread pool size

PSeitz:
Don't take all threads for search, reserve one thread for incoming requests. Otherwise there may not be enough CPU left to answer health checks
This behavior can be overruled by manually setting the search thread pool size via `QW_SEARCH_THREAD_POOL_SIZE`.

improve logging:
log actual indices that are being queried on leaf nodes instead of the patterns from the user request
reduce env var spam

Related to https://github.com/quickwit-oss/quickwit/issues/5305


fmassot:
I suggest closing this PR as we are not reproducing the health checks issue with heavy CPU queries

PSeitz:
Yes, I'll move the logging improvements to a new PR

PR #5281, opened 2024-07-31
fix conversion from old to new mapping for json and object

trinity-1686a:
### Description

progress on #5084
fix conversions of old documents where some fields changed from json to object or the other way

### How was this PR tested?

enabled broken integration tests, added a few more unit tests.

PR #5308, opened 2024-08-08
allow querying non existing fields

### LINKED ISSUE #5302: Quickwit Should Ignore Field queries with a mismatch type
**Describe the bug**
When querying against multiple fields, when field has a defined index mapping type, querying for a field that does not exist creates an error. For example, querying for a nested field against a text field will result in an error

```txl
invalid query: field does not exist: `<field>`
```

1. define an index with a text field named `_app` as a text field
```json5
    , {
        name: '_app'
      , type: 'text'
      , tokenizer: 'raw_lowercase'
      , fast: {
          normalizer: 'lowercase'
        }
      }
```
2. insert a document with an object field`{o_app: {name: 'test-app'}}`
3. query for `test-app` against `_app.name` or `o_app.name`

```json5
{
  query: {
    bool: {
      must: [
        { range: { _ts: { gte: 1722876383510 } } },
        { range: { _ts: { lt: 1722879983744 } } },
        {
          query_string: {
            fields: [ '_app.name', 'o_app.name' ],
            query: 'test-app',
            lenient: true
          }
        }
      ],
      should: null
    }
  }
}
```

**Expected behavior**
Quickwit should ignore the case when a field doesn't exist and query against the fields that do.

**Configuration:**

```json5
    "build_info": {
      "build_date": "2024-07-31T06:15:28Z",
      "build_profile": "release",
      "build_target": "x86_64-unknown-linux-gnu",
      "cargo_pkg_version": "0.8.0",
      "commit_date": "2024-07-31T05:41:46Z",
      "commit_hash": "c263c8d8cfe4c154f8b21a22c5d2840b928369b8",
      "commit_short_hash": "c263c8d",
      "commit_tags": [
        "qw-airmail-20240731"
      ],
      "version": "0.8.0-nightly"
    },
```

trinity-1686a:
### Description

fix #5302
the spec around update is a bit odd: we can search inside deleted fields (validation is okay with inexistant field, and mapping the query per-split actually find the field for older splits), but we don't return these fields

### How was this PR tested?
updated ut, added it


quickwit/quickwit-integration-tests/src/tests/update_tests/doc_mapping_tests.rs
```diff
@@ -220,9 +220,9 @@ async fn test_update_doc_mapping_json_to_text() {
         ingest_after_update,
         &[
             ("body:hello", Ok(&[json!({"body": "hello world"})])),
-            // error expected because the validation is performed
-            // by latest doc mapping
-            ("body.field1:hello", Err(())),
+            // ok because we can query for non existent field and they are ignored, but we don't
+            // get the field because it was deleted. This spec is very odd :-/
```
rdettai, line 224: I agree this behavior is very weird! Where is this spec described? In this test, we use the dynamic (default) mode. Shouldn´t we get all the fields when that's the case?

trinity-1686a: ho, i thought the tests where on a lenient doc mapping, but no, it's just the absence of #5281 showing itself

rdettai: ah, so we should definitively merge that and this should work!

trinity-1686a: hum, nevermind, body didn't disappear, it was transformed from json to text, which we currently make return nothing (there is no good way to put a complex object in a string)

trinity-1686a:
we may make so we accept query on non-existing field only when lenient is true. This is different from what ES does, but there seems to be a consensus among us that what ES does here is not a behaviour we like (being too lenient by default)

quickwit/quickwit-query/src/elastic_query_dsl/match_query.rs
```diff
@@ -45,8 +45,8 @@ pub(crate) struct MatchQueryParams {
     // Regardless of this option Quickwit behaves in elasticsearch definition of
```
fulmicoton, line 45: Can you edit the comment?

quickwit/quickwit-query/src/elastic_query_dsl/multi_match.rs
```diff
@@ -51,8 +51,8 @@ struct MultiMatchQueryForDeserialization {
     // Regardless of this option Quickwit behaves in elasticsearch definition of
     // lenient. We include this property here just to accept user queries containing
     // this option.
-    #[serde(default, rename = "lenient")]
-    _lenient: bool,
+    #[serde(default)]
+    lenient: bool,
```
fulmicoton, line 55: same as for match query. we need to edit the comment.

quickwit/quickwit-query/src/query_ast/user_input_query.rs
```diff
@@ -99,6 +105,7 @@ fn convert_user_input_ast_to_query_ast(
     user_input_ast: UserInputAst,
```
fulmicoton, line 105: can you add some short documentation for this method?

fulmicoton (approved):

PR #5316, opened 2024-08-14
lift clauses

trinity-1686a:
### Description

this is akin to https://github.com/quickwit-oss/tantivy/pull/2449 (which runs optimization inside the query parser), but for ast not built through the query parser. Inspired by queries seen on airmail

### How was this PR tested?

added unit test to verify the optimizations are run, added prop-test to verify the optimizations never change the result of a query (which happened to be the case before)

### DIFF
--- a/quickwit/quickwit-query/src/query_ast/tantivy_query_ast.rs
+++ b/quickwit/quickwit-query/src/query_ast/tantivy_query_ast.rs
@@ -202,7 +202,69 @@ impl TantivyBoolQuery {
             return TantivyQueryAst::match_all();
         }
 
-        // TODO lift clauses
+        let mut new_must = Vec::with_capacity(self.must.len());
+        for must in self.must {
+            let mut must_bool = match must {
+                TantivyQueryAst::Bool(bool_query) => bool_query,
+                _ => {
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-query/src/query_ast/tantivy_query_ast.rs
```diff
@@ -188,6 +201,72 @@ impl TantivyBoolQuery {
...
+                TantivyQueryAst::Bool(bool_query) => bool_query,
+                _ => {
+                    new_must.push(must);
+                    continue;
+                }
+            };
+            if must_bool.should.is_empty() {
+                new_must.append(&mut must_bool.must);
+                self.filter.append(&mut must_bool.filter);
+                self.must_not.append(&mut must_bool.must_not);
```
PSeitz, line 217: shouldn't we call simplify here, so the optimization is recursive?

trinity-1686a: `simplify_asts` on everything at the start of the function

PSeitz (approved):

Issue #5091, opened 2024-06-07
Range queries on fast str field and a dynamic schema are silently ignored


RaphaelMarinier:
**Describe the bug**
Range queries on an str fast field that is dynamically created based on indexed documents (dynamic schema) seem to be silently ignored.
See also #5090 where such queries fail when the str fast field is explicit in the schema.

**Steps to reproduce (if applicable)**

1. Create a dynamic schema, e.g. https://github.com/quickwit-oss/benchmarks/blob/main/tracks/generated-logs/index-config.quickwit.yaml
2. Ingest documents with text fields (other than `message` explicit in the schema above), e.g. the `generated-logs` dataset of the benchmark above.
3. Query that field:
```
quickwit index search --index generated-logs  --query 'trace_id:[abc TO def]' | grep trace_id
```
>     "trace_id": "3579345088981071944"
>      "trace_id": "3099098393060003496"
>      "trace_id": "7455985833741068554"
>      "trace_id": "8397828520949179503"
>      "trace_id": "452393755932269140"
>      "trace_id": "4372017533496144022"
>      "trace_id": "3838521724700584677"
>      "trace_id": "8729677587882026026"

**Expected behavior**
Similarly to #5090, I expected such queries to be either supported, or disallowed in documentation and returning an error instead of being silently ignored.

**Configuration:**

1. Output of `quickwit --version`
Quickwit 0.8.0 (x86_64-unknown-linux-gnu 2024-06-06T20:01:05Z 1fd7d0b)
3. The index_config.yaml
https://github.com/quickwit-oss/benchmarks/blob/main/tracks/generated-logs/index-config.quickwit.yaml

Issue #5090, opened 2024-06-07
Range Query on str field not supported (contrary to what docs imply)


RaphaelMarinier:
**Describe the bug**

Range queries on fast str fields are not supported. However, the documentation seems to imply so:

https://quickwit.io/docs/reference/query-language#range-fieldlow_bound-to-high_bound
> Order
> For text fields, the ranges are defined by lexicographic order on uft-8 encoded byte arrays. It means for a text field, 100 is between 1 and 2.

https://quickwit.io/docs/get-started/query-language-intro includes the following range query example:
> last_name:[banner TO miller]

**Steps to reproduce (if applicable)**
1. Have an index with a `trace_id` text fast field:
```
doc_mapping:
  mode: lenient
  field_mappings:
    - name: timestamp
      type: datetime
      precision: milliseconds
      fast: true
      input_formats:
        - rfc3339
      output_format: unix_timestamp_secs
    - name: trace_id
      type: text
      tokenizer: raw
      fast: true
  timestamp_field: timestamp
```
(Having an explicit schema instead of a dynamic one is important - another bug report is coming)

2. Query that field:
```
quickwit tool local-search --index generated-logs-tiny-fast  --config quickwit.yaml --query 'trace_id:[abc TO def]'
```
> ❯ Searching directly on the index storage (without calling REST API)...
> 2024-06-07T09:55:19.575Z ERROR quickwit: command failed error=invalid query: range query on `str` field (`trace_id`) forbidden
> ✘ command failed: invalid query: range query on `str` field (`trace_id`) forbidden

Note: using `quickwit index search` does not change the outcome:
```
quickwit index search --index generated-logs-tiny-fast  --query 'trace_id:[abc TO def]'
```
> 2024-06-07T10:04:24.052Z ERROR quickwit: command failed error=API error: (code=400 Bad Request, message=invalid query: range query on `str` field (`trace_id`) forbidden)
> ✘ command failed: API error: (code=400 Bad Request, message=invalid query: range query on `str` field (`trace_id`) forbidden)
>
> Caused by:
>    (code=400 Bad Request, message=invalid query: range query on `str` field (`trace_id`) forbidden)

**Expected behavior**
[...]

PR #5324, opened 2024-08-19
enable str fast field range queries

PSeitz:
enable range queries on str based fast field queries
update tantivy

closes https://github.com/quickwit-oss/quickwit/issues/5091
closes https://github.com/quickwit-oss/quickwit/issues/5090

### DIFF
--- a/quickwit/quickwit-query/src/query_ast/range_query.rs
+++ b/quickwit/quickwit-query/src/query_ast/range_query.rs
@@ -245,6 +245,15 @@ impl BuildTantivyAst for RangeQuery {
                     .into();
                     sub_queries.push(str_query);
                 }
+                if sub_queries.is_empty() {
+                    return Err(InvalidQuery::InvalidBoundary {
+                        expected_value_type: "i64, u64, f64, str",
+                        field_name: field_entry.name().to_string(),
+                    });
+                }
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-query/src/query_ast/range_query.rs
```diff
@@ -191,13 +220,33 @@ impl BuildTantivyAst for RangeQuery {
...
+                        term
+                    }),
+                    upper_bound.map(|val| {
+                        let val = get_normalized_text(&mut normalizer, val);
+                        let mut term = empty_term.clone();
+                        term.append_type_and_str(&val);
+                        term
+                    }),
+                )
+                .into()
```
trinity-1686a, line 244: this has a funky behavior in case the bounds are numbers, but the column contains at least some strings. For instance if have `{"abc": "1"}`, `abc:1` returns it, but `abc:[0 TO 2]` returns nothing, while `abc:[0 TO :]` does actually returns something
I think for json fields, it would be safer to build an OR of integer/string if the bounds can be parsed to an integer

trinity-1686a (changes requested):

trinity-1686a (approved):
some tests needs fixing, but the code looks okay

PR #5327, opened 2024-08-19
Fix source path in Lambda distrib

rdettai:
### Description

Bug introduced by https://github.com/quickwit-oss/quickwit/pull/5148

Also removed the overwrite option which was hardcoded and buggy anyway.

### How was this PR tested?

Added an end 2 end integration test exposed in the Makefile. Note that the test is currently failing until https://github.com/quickwit-oss/quickwit/pull/5330 is also merged.

### DIFF
--- a/quickwit/quickwit-lambda/src/indexer/model.rs
+++ b/quickwit/quickwit-lambda/src/indexer/model.rs
@@ -38,6 +38,7 @@ impl IndexerEvent {
             IndexerEvent::S3(event) => [
                 "s3://",
                 event.records[0].s3.bucket.name.as_ref().unwrap(),
+                "/",
                 event.records[0].s3.object.key.as_ref().unwrap(),
             ]
             .join(""),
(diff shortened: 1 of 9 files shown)


rdettai:
Internal CI run: https://github.com/quickwit-inc/terraform/actions/runs/10455107621

quickwit/quickwit-lambda/src/indexer/model.rs
```diff
@@ -38,6 +38,7 @@ impl IndexerEvent {
             IndexerEvent::S3(event) => [
                 "s3://",
                 event.records[0].s3.bucket.name.as_ref().unwrap(),
+                "/",
```
trinity-1686a, line 41: it feels like the kind of things `test_s3_event_uri` should have caught. What gives?

trinity-1686a (approved):

quickwit/quickwit-lambda/src/indexer/model.rs
rdettai: I added a make command to run the Lambda tests, but these tests are not running in CI (for performance reasons)

trinity-1686a: okay. then we should make sure to run them manually when they may be relevant

PR #5335, opened 2024-08-22
Add shard pruning endpoint

rdettai:
### Description

Add a PruneShards endpoint to the Shard API. This will be used by the FileSource for garbage collecting old checkpoints

### How was this PR tested?

Unit tests


quickwit/quickwit-metastore/src/metastore/postgres/queries/shards/prune_count.sql
```diff
@@ -0,0 +1,12 @@
...
+    FROM shards
+    WHERE index_uid = $1
+        AND source_id = $2
+    ORDER BY update_timestamp DESC
+    LIMIT $3
+)
+DELETE FROM shards
+WHERE index_uid = $1
+    AND source_id = $2
+    AND shard_id NOT IN (SELECT shard_id FROM recent_shards)
```
fulmicoton, line 12: this does not work?
```suggestion
    AND shard_id NOT IN  recent_shards
```

fulmicoton (approved):

PR #5336, opened 2024-08-22
fix warming-up position incorrectly when 2 phrase prefix query are issued for the same final term

trinity-1686a:
### Description

fix a bug where `field:"short"* field:"not short"*` would succeed, but `field:"not short"* field:"short"*` would fail to load position, and return an error

### How was this PR tested?

tested manually, added unit test for the faulty code


fmassot (approved):
Good job!!!!!

PR #5337, opened 2024-08-22
rename index-id to hopefully fix coverage

trinity-1686a:
two test had the same index-id, which is my best guess as to why CI is currently red on main


rdettai (approved):

rdettai:
:pray:

rdettai:
I didn´t find the CI run where this was an issue, but fixing the typo doesn´t hurt :smile:

PR #5339, opened 2024-08-23
Garbage collect shards in SQS Filesource

rdettai:
### Description

Trigger the prune_shards endpoint from the SQS file sources.

### How was this PR tested?

Added integration test.

### DIFF
--- a/quickwit/quickwit-indexing/src/source/queue_sources/shared_state.rs
+++ b/quickwit/quickwit-indexing/src/source/queue_sources/shared_state.rs
@@ -27,7 +27,7 @@ use quickwit_proto::metastore::{
     AcquireShardsRequest, MetastoreService, MetastoreServiceClient, OpenShardSubrequest,
     OpenShardsRequest, PruneShardsRequest,
 };
-use quickwit_proto::types::{DocMappingUid, IndexUid, Position, ShardId};
+use quickwit_proto::types::{DocMappingUid, Position, ShardId, SourceUid};
 use time::OffsetDateTime;
 use tracing::{error, info};
 
@@ -36,8 +36,7 @@ use super::message::PreProcessedMessage;
 #[derive(Clone)]
 pub struct QueueSharedState {
     metastore: MetastoreServiceClient,
-    pub index_uid: IndexUid,
-    pub source_id: String,
+    pub source_uid: SourceUid,
     /// Duration after which the processing of a shard is considered stale and
     /// should be reacquired
     reacquire_grace_period: Duration,
@@ -49,18 +48,16 @@ impl QueueSharedState {
     /// in the background
     pub fn new(
         metastore: MetastoreServiceClient,
-        index_uid: IndexUid,
-        source_id: String,
+        source_uid: SourceUid,
         reacquire_grace_period: Duration,
-        max_age: Option<u32>,
+        max_age: Option<Duration>,
         max_count: Option<u32>,
         pruning_interval: Duration,
     ) -> Self {
         let cleanup_handle = Arc::new(());
         tokio::spawn(Self::run_cleanup_task(
             metastore.clone(),
-            index_uid.clone(),
-            source_id.clone(),
+            source_uid.clone(),
             max_age,
             max_count,
             pruning_interval,
@@ -68,8 +65,7 @@ impl QueueSharedState {
         ));
         Self {
             metastore,
-            index_uid,
-            source_id,
+            source_uid,
             reacquire_grace_period,
             _cleanup_handle: cleanup_handle,
         }
@@ -77,9 +73,8 @@ impl QueueSharedState {
 
     async fn run_cleanup_task(
         metastore: MetastoreServiceClient,
-        index_uid: IndexUid,
-        source_id: String,
-        max_age: Option<u32>,
+        source_uid: SourceUid,
+        max_age: Option<Duration>,
         max_count: Option<u32>,
         pruning_interval: Duration,
         owner_handle: Arc<()>,
@@ -87,6 +82,11 @@ impl QueueSharedState {
         if max_count.is_none() && max_age.is_none() {
             return;
         }
+        let max_age_secs = max_age.map(|duration| duration.as_secs() as u32);
+        let SourceUid {
+            index_uid,
+            source_id,
+        } = source_uid;
         tokio::spawn(async move {
             let mut interval = tokio::time::interval(pruning_interval);
             loop {
@@ -98,7 +98,7 @@ impl QueueSharedState {
                     .prune_shards(PruneShardsRequest {
                         index_uid: Some(index_uid.clone()),
                         source_id: source_id.clone(),
-                        max_age,
+                        max_age_secs,
                         max_count,
                     })
                     .await;
@@ -124,8 +124,8 @@ impl QueueSharedState {
             .enumerate()
             .map(|(idx, partition_id)| OpenShardSubrequest {
                 subrequest_id: idx as u32,
-                index_uid: Some(self.index_uid.clone()),
-                source_id: self.source_id.clone(),
+                index_uid: Some(self.source_uid.index_uid.clone()),
+                source_id: self.source_uid.source_id.clone(),
                 leader_id: String::new(),
                 follower_id: None,
                 shard_id: Some(ShardId::from(partition_id.as_str())),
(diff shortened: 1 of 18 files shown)


rdettai:
There is one footgun here that would like to avoid: it might happen that a shard is garbage collected right before it's published (e.g `deduplication_window_max_messages` is small compared to processed message rate). Possible solutions:
- GC only EOF shards, the drawback is that if a lot of junk shards that couldn't be published accumulate, we might end up overwhelming the metastore with a huge shard table
- set a grace period of `x*commit_timeout`, reducing the situations where the shard could be garbage collected while it was still being worked on. This is not bulletproof though as the indexing might be interrupted for some reason and resume after the grace period, but failing the commit in those rare cases should be supportable

quickwit/quickwit-indexing/src/source/queue_sources/shared_state.rs
```diff
@@ -18,38 +18,104 @@
...
+            loop {
+                interval.tick().await;
+                if Arc::strong_count(&owner_handle) == 1 {
+                    break;
+                }
+                let result: Result<_, _> = metastore
+                    .prune_shards(PruneShardsRequest {
+                        index_uid: Some(index_uid.clone()),
+                        source_id: source_id.clone(),
+                        max_age,
```
fulmicoton, line 101: we should probably rename that into max_age_seconds

quickwit/quickwit-indexing/src/source/queue_sources/shared_state.rs
```diff
@@ -18,38 +18,104 @@
...
 
 impl QueueSharedState {
+    /// Create a shared state service and runs a cleanup task that prunes shards
+    /// in the background
+    pub fn new(
+        metastore: MetastoreServiceClient,
+        index_uid: IndexUid,
+        source_id: String,
+        reacquire_grace_period: Duration,
+        max_age: Option<u32>,
```
fulmicoton, line 55:
```suggestion
        max_age: Option<Duration>,
```

quickwit/quickwit-indexing/src/source/queue_sources/shared_state.rs
```diff
@@ -18,38 +18,104 @@
...
+            source_id,
+            reacquire_grace_period,
+            _cleanup_handle: cleanup_handle,
+        }
+    }
+
+    async fn run_cleanup_task(
+        metastore: MetastoreServiceClient,
+        index_uid: IndexUid,
+        source_id: String,
```
fulmicoton, line 81: source_uid: SourceUid

fulmicoton, line 81:
```suggestion
     source_uid: SourceUid,
```

quickwit/quickwit-indexing/src/source/queue_sources/shared_state.rs
```diff
@@ -18,38 +18,104 @@
...
+        owner_handle: Arc<()>,
+    ) {
+        if max_count.is_none() && max_age.is_none() {
+            return;
+        }
+        tokio::spawn(async move {
+            let mut interval = tokio::time::interval(pruning_interval);
+            loop {
+                interval.tick().await;
+                if Arc::strong_count(&owner_handle) == 1 {
```
fulmicoton, line 94: Is there a reason why you rely on strong_count as opposed to just using a `std::sync::Weak`?

quickwit/quickwit-serve/src/index_api/rest_handler.rs
```diff
@@ -919,6 +921,42 @@ async fn delete_source(
...
+        .list_shards(ListShardsRequest {
+            subrequests: vec![ListShardsSubrequest {
+                index_uid: Some(index_uid),
+                source_id: source_id.to_string(),
+                ..Default::default()
+            }],
+        })
+        .await?;
+
+    Ok(shards.subresponses[0].clone().shards)
```
fulmicoton, line 957:
```suggestion
    Ok(shards.subresponses[0].clone().shards)
```
the clone makes no sense here.

quickwit/quickwit-indexing/src/source/queue_sources/shared_state.rs
```diff
@@ -18,38 +18,104 @@
...
+        metastore: MetastoreServiceClient,
+        index_uid: IndexUid,
+        source_id: String,
+        reacquire_grace_period: Duration,
+        max_age: Option<u32>,
+        max_count: Option<u32>,
+        pruning_interval: Duration,
+    ) -> Self {
+        let cleanup_handle = Arc::new(());
+        tokio::spawn(Self::run_cleanup_task(
```
fulmicoton, line 60: I don't understand. This runs on every node?

fulmicoton: Can you add some developer documentation in the source/queue_sources/design.md, or point me to some other place if that exists somewhere.

quickwit/quickwit-indexing/src/source/queue_sources/shared_state.rs
rdettai: Not really. I'll switch to `Weak` as it's what's currently used everywhere else in the codebase.

quickwit/quickwit-metastore/src/metastore/file_backed/file_backed_index/shards.rs
```diff
@@ -245,9 +245,10 @@ impl Shards {
     ) -> MetastoreResult<MutationOccurred<PruneShardsResponse>> {
         let initial_shard_count = self.shards.len();
 
-        if let Some(max_age) = request.max_age {
+        if let Some(max_age_secs) = request.max_age_secs {
             self.shards.retain(|_, shard| {
-                let limit_timestamp = OffsetDateTime::now_utc().unix_timestamp() - max_age as i64;
+                let limit_timestamp =
+                    OffsetDateTime::now_utc().unix_timestamp() - max_age_secs as i64;
                 shard.update_timestamp >= limit_timestamp
```
fulmicoton, line 252: nitpick
```suggestion
                let now = OffsetDateTime::now_utc().unix_timestamp();
                shard.update_timestamp + max_secs as i64 >= now
```

or if you want the extra name
```suggestion
                let gc_deadline = shard.update_timestamp + max_secs as i64;
                let now = OffsetDateTime::now_utc().unix_timestamp();
                gc_deadline >= now
```

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
```diff
@@ -1495,8 +1495,9 @@ impl MetastoreService for PostgresqlMetastore {
         const PRUNE_AGE_SHARDS_QUERY: &str = include_str!("queries/shards/prune_age.sql");
         const PRUNE_COUNT_SHARDS_QUERY: &str = include_str!("queries/shards/prune_count.sql");
 
-        if let Some(max_age) = request.max_age {
-            let limit_datetime = OffsetDateTime::now_utc() - Duration::from_secs(max_age as u64);
+        if let Some(max_age_secs) = request.max_age_secs {
+            let limit_datetime =
```
fulmicoton, line 1499: same nitpick as above.

quickwit/quickwit-indexing/src/source/queue_sources/shared_state.rs
```diff
@@ -125,7 +191,7 @@ impl QueueSharedState {
 /// Acquires shards from the shared state for the provided list of messages and
 /// maps results to that same list
 pub async fn checkpoint_messages(
```
fulmicoton, line 193: to be discussed: I don't understand what this function does, nor its naming.

fulmicoton (approved):

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
rdettai: In the comment above, it was possible to rewrite the calculations to use `Add` instead of `Sub`. Can we do something similar here or should we just explicit why we are confident that no underflow can happen?

quickwit/quickwit-indexing/src/source/queue_sources/shared_state.rs
rdettai: > also I am not fond of the iterator style with a non pure function

I got rid of those.

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
fulmicoton: ah right. no pb.

fulmicoton (approved):

fulmicoton (approved):

PR #5340, opened 2024-08-23
Fix flaky postgres tests

rdettai:
### Description

Postgres tests tend to fail in coverage CI runs.

### How was this PR tested?


trinity-1686a (approved):
looks reasonable

PR #5350, opened 2024-08-27
Fix force commit on ingest V2

rdettai:
### Description

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-integration-tests/src/tests/ingest_tests.rs
+++ b/quickwit/quickwit-integration-tests/src/tests/ingest_tests.rs
@@ -317,10 +317,103 @@ async fn test_ingest_v2_happy_path() {
 }
 
 #[tokio::test]
-async fn test_commit_modes() {
+async fn test_commit_force() {
     initialize_tests();
-    let sandbox = ClusterSandboxBuilder::build_and_start_standalone().await;
-    let index_id = "test_commit_modes";
+    let mut sandbox = ClusterSandboxBuilder::build_and_start_standalone().await;
+    let index_id = "test_commit_force";
+    let index_config = format!(
+        r#"
+        version: 0.8
+        index_id: {index_id}
+        doc_mapping:
+            field_mappings:
+            - name: body
+              type: text
+        indexing_settings:
+            commit_timeout_secs: 20
+        "#
+    );
+
+    // Create index
+    sandbox
+        .indexer_rest_client
+        .indexes()
+        .create(index_config, ConfigFormat::Yaml, false)
+        .await
+        .unwrap();
+
+    sandbox.enable_ingest_v2();
+
+    ingest_with_retry(
+        &sandbox.indexer_rest_client,
+        index_id,
+        ingest_json!({"body": "force"}),
+        CommitType::Force,
+    )
+    .await
+    .unwrap();
+
+    // commit_timeout_secs is set to a large value, so this will timeout if
+    // CommitType::Force is not working
+    sandbox
+        .wait_for_splits(index_id, Some(vec![SplitState::Published]), 1)
+        .await
+        .unwrap();
+
+    sandbox.assert_hit_count(index_id, "body:force", 1).await;
+
+    sandbox.shutdown().await.unwrap();
+}
+
+// #[tokio::test]
+// async fn test_commit_wait_for() {
+//     initialize_tests();
+//     let mut sandbox = ClusterSandboxBuilder::build_and_start_standalone().await;
+//     let index_id = "test_commit_wait_for";
+//     let index_config = format!(
+//         r#"
+//         version: 0.8
+//         index_id: {index_id}
+//         doc_mapping:
+//             field_mappings:
+//             - name: body type: text
+//         indexing_settings:
+//             commit_timeout_secs: 2
+//         "#
+//     );
+
+//     // Create index
+//     sandbox
+//         .indexer_rest_client
+//         .indexes()
+//         .create(index_config, ConfigFormat::Yaml, false)
+//         .await
+//         .unwrap();
+
+//     sandbox.enable_ingest_v2();
+
+//     sandbox
+//         .indexer_rest_client
+//         .ingest(
+//             index_id,
+//             ingest_json!({"body": "wait"}),
+//             None,
+//             None,
+//             CommitType::WaitFor,
+//         )
+//         .await
+//         .unwrap();
+
+//     sandbox.assert_hit_count(index_id, "body:wait", 1).await;
+
+//     sandbox.shutdown().await.unwrap();
+// }
+
+#[tokio::test]
+async fn test_commit_auto() {
+    initialize_tests();
+    let mut sandbox = ClusterSandboxBuilder::build_and_start_standalone().await;
+    let index_id = "test_commit_auto";
     let index_config = format!(
         r#"
         version: 0.8
@@ -334,7 +427,6 @@ async fn test_commit_modes() {
         "#
     );
 
-    // Create index
     sandbox
         .indexer_rest_client
         .indexes()
@@ -342,37 +434,8 @@ async fn test_commit_modes() {
         .await
         .unwrap();
 
-    // TODO: make this test work with ingest v2 (#4438)
-    // sandbox.enable_ingest_v2();
+    sandbox.enable_ingest_v2();
 
-    // Test force commit
-    ingest_with_retry(
-        &sandbox.indexer_rest_client,
-        index_id,
-        ingest_json!({"body": "force"}),
-        CommitType::Force,
-    )
-    .await
-    .unwrap();
-
-    sandbox.assert_hit_count(index_id, "body:force", 1).await;
-
-    // Test wait_for commit
-    sandbox
-        .indexer_rest_client
-        .ingest(
-            index_id,
-            ingest_json!({"body": "wait"}),
-            None,
-            None,
-            CommitType::WaitFor,
-        )
-        .await
-        .unwrap();
-
-    sandbox.assert_hit_count(index_id, "body:wait", 1).await;
-
-    // Test auto commit
     sandbox
         .indexer_rest_client
         .ingest(
@@ -388,13 +451,12 @@ async fn test_commit_modes() {
     sandbox.assert_hit_count(index_id, "body:auto", 0).await;
 
     sandbox
-        .wait_for_splits(index_id, Some(vec![SplitState::Published]), 3)
+        .wait_for_splits(index_id, Some(vec![SplitState::Published]), 1)
         .await
         .unwrap();
 
     sandbox.assert_hit_count(index_id, "body:auto", 1).await;
 
-    // Clean up
     sandbox.shutdown().await.unwrap();
 }
 


quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -17,22 +17,108 @@
...
+/// `wait_for` commit mode.
+///
+/// Registers a set of shard positions and listens to [`ShardPositionsUpdate`]
+/// events to assert when all the persisted events have been published. To make
+/// sure that no events are missed:
+/// - the tracker should be created before the persist requests are sent
+/// - the `shard_persisted` method should for all successful persist subrequests
+struct PublishTracker {
+    state: Arc<Mutex<PublishState>>,
+    publish_complete: Arc<Notify>,
```
fulmicoton, line 56: I recommend using a `oneshot::Send<()>` here.

Notify works too of course, but oneshot assymetry (Receiver / Sender) makes it less ambiguous for the reader.

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -17,22 +17,108 @@
...
+/// A helper for tracking the progress of the publish events when running in
+/// `wait_for` commit mode.
+///
+/// Registers a set of shard positions and listens to [`ShardPositionsUpdate`]
+/// events to assert when all the persisted events have been published. To make
+/// sure that no events are missed:
+/// - the tracker should be created before the persist requests are sent
+/// - the `shard_persisted` method should for all successful persist subrequests
+struct PublishTracker {
+    state: Arc<Mutex<PublishState>>,
```
fulmicoton, line 55: Do we need the `state` in the publish tracker? I suspect we can remove it from this object, have the callback be the only owner, and remove the lock as a result.

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
fulmicoton: Ah I see. I guess your problem is that you want to be able to populate the already_published stuff after listening Is this correct?

quickwit/rest-api-tests/scenarii/es_compatibility/_setup.quickwit.yaml
```diff
@@ -66,4 +66,3 @@ params:
   refresh: "true"
 headers: {"Content-Type": "application/json", "content-encoding": "gzip"}
 body_from_file: gharchive-bulk.json.gz
-sleep_after: 3
```
fulmicoton, line 69: nice! you remembered to clean this up.

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -138,6 +246,12 @@ impl IngestWorkbench {
             );
             return;
         };
+        if let Some(publish_tracker) = &mut self.publish_tracker {
```
fulmicoton, line 249: I don't understand the mut here.
```suggestion
        if let Some(publish_tracker) = &self.publish_tracker {
```

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
fulmicoton: Can you write a small story about that there in the PublishTracker doc?

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -17,22 +17,108 @@
...
+                }
+            });
+        Self {
+            state,
+            _publish_listen_handle,
+            publish_complete,
+        }
+    }
+
+    fn shard_persisted(&self, shard_id: ShardId, new_position: Position) {
```
fulmicoton, line 97: this method name is not so great. 

Can we rename it?

`set_shard_pesisted_position` maybe?

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -17,22 +17,108 @@
...
+use quickwit_proto::types::{NodeId, Position, ShardId, SubrequestId};
+use tokio::sync::Notify;
 use tracing::warn;
 
 use super::router::PersistRequestSummary;
 
+#[derive(Default)]
+struct PublishState {
+    awaiting_publish: HashMap<ShardId, Position>,
+    already_published: HashMap<ShardId, Position>,
```
fulmicoton, line 43: could we create the publish state with the list of ShardId we have an interest in?
The point would be to avoid cloning and saving all of the shards (thousands) getting published in Quickwit.

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -17,22 +17,108 @@
...
+    state: Arc<Mutex<PublishState>>,
+    publish_complete: Arc<Notify>,
+    _publish_listen_handle: EventSubscriptionHandle,
+}
+
+impl PublishTracker {
+    fn new(event_tracker: EventBroker) -> Self {
+        let state = Arc::new(Mutex::new(PublishState::default()));
+        let state_clone = state.clone();
+        let publish_complete = Arc::new(Notify::new());
```
fulmicoton, line 64: I would have preferred a oneshot here.

The Sender woudl be moved in the closure, while the receiver would be in the PublishTracker object.

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -17,22 +17,108 @@
...
+            _publish_listen_handle,
+            publish_complete,
+        }
+    }
+
+    fn shard_persisted(&self, shard_id: ShardId, new_position: Position) {
+        let mut state_handle = self.state.lock().unwrap();
+        match state_handle.already_published.get(&shard_id) {
+            Some(already_published_position) if new_position <= *already_published_position => {
+                // already published, no need to track this shard's position updates
```
fulmicoton, line 101: I think this is a bug...
Who will be notify everyone if that happens?

fulmicoton (changes requested):
@rdettai I think there is a bug if the required position arrives after the position notification.

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
rdettai: If I understand correctly, the sequence you are flagging here as a potential bug is the following:
1. the publish event is recorded
2. the persisted shard position is recorded

In that case, the shard position is never recorded in `awaiting_publish`. You are right that this means that the notification is never tirggered. But this situation is handled in the waiting method:
https://github.com/quickwit-oss/quickwit/blob/fbbdc0f442738692ded82841c9764b747bf72194/quickwit/quickwit-ingest/src/ingest_v2/workbench.rs#L111-L119

Do you think I missed something in the correctness argument?

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
rdettai: I hesitated doing so, but I wasn't quite sure that maintaining that extra filter would be cheaper than recording all publish events. If the published shard event bus can be very busy, it's definitively worth it.

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
rdettai: good point

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
rdettai: leftover from a previous mutable design. Quite surprised clippy didn't catch this!

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
rdettai: actually, after some extra thoughts, there is a bug if:
1. a persisted position is recorded for shard1
2. the publish event for shard1 is recorded
3. at this point the notifications is triggered and stored (notify_one)
4. a persisted shard position is recorded
5. `wait_publish_complete` is called and doesn't block because the notification was recorded

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
fulmicoton: I think we want to support at least 10k indexes, and commit every 30s.

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -109,13 +114,15 @@ impl PublishTracker {
...
     async fn wait_publish_complete(self) {
-        // correctness:
-        // - `state.awaiting_publish` can only grow before this (`self` is consumed)
-        // - `publish_complete.notify()` is called as soon as the last shard is published
+        // correctness: new shards cannot be added to `state.awaiting_publish`
+        // at this point because `self` is consumed. By subscribing to
+        // `publish_complete` before checking `awaiting_publish`, we make sure we
+        // don't miss the moment when it becomes empty.
+        let notified = self.publish_complete.notified();
         if self.state.lock().unwrap().awaiting_publish.is_empty() {
```
fulmicoton, line 122: there is still a race condition, no?

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
rdettai: changed the state machine to track only requested shards

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -17,22 +17,156 @@
...
-use quickwit_proto::types::{NodeId, ShardId, SubrequestId};
-use tracing::warn;
+use quickwit_proto::types::{NodeId, Position, ShardId, SubrequestId};
+use tokio::sync::Notify;
+use tracing::{error, warn};
 
 use super::router::PersistRequestSummary;
 
+enum PublishState {
+    Tracked,
```
fulmicoton, line 43: can you put a short comment describing the different states?

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -17,22 +17,156 @@
...
+                    state_handle.awaiting_count += 1;
+                }
+            }
+        } else {
+            error!(%shard_id, "requested shards should be registered before their position is tracked")
+        }
+    }
+
+    async fn wait_publish_complete(self) {
+        // correctness: `awaiting_count` cannot be increased after this point
```
fulmicoton, line 158: ok i get the argument

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -17,22 +17,156 @@
...
+    fn new(event_tracker: EventBroker) -> Self {
+        let state = Arc::new(Mutex::new(ShardPublishStates::default()));
+        let state_clone = state.clone();
+        let publish_complete = Arc::new(Notify::new());
+        let publish_complete_notifier = publish_complete.clone();
+        let _publish_listen_handle =
+            event_tracker.subscribe(move |update: ShardPositionsUpdate| {
+                let mut state_handle = state_clone.lock().unwrap();
+                for (updated_shard_id, updated_position) in &update.updated_shard_positions {
+                    if let Some(publish_state) = state_handle.states.get_mut(updated_shard_id) {
```
fulmicoton, line 81: I think this code should be in a method of `ShardPublishStates`?

fulmicoton: why name it `state_handle`? I would suggest `shard_publish_states`, no?

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
```diff
@@ -358,11 +536,117 @@ impl IngestSubworkbench {
...
-    use quickwit_proto::ingest::ingester::PersistFailureReason;
-    use quickwit_proto::types::ShardId;
+    use std::time::Duration;
+
+    use quickwit_proto::ingest::ingester::{PersistFailureReason, PersistSubrequest};
+    use quickwit_proto::types::{IndexUid, ShardId, SourceUid};
 
     use super::*;
 
+    #[tokio::test]
```
fulmicoton, line 546: nitpick: I would have tested the `ShardPublishState` object independently.
(Note this is considered bad practise by some company however)

fulmicoton (approved):

quickwit/quickwit-ingest/src/ingest_v2/workbench.rs
rdettai: I refactored the whole thing to a separate module for more clarity

rdettai:
sorry, I meant to squash these commits before merging, sorry for flooding the main branch!

PR #5357, opened 2024-08-28
share aggregation limit on node

PSeitz:
Changes how aggregation limits are shared. Previously they were per
request, now they are per node.
The counters are only updated after an allocation in tantivy, which
should keep the contention risk relatively low.

fix flaky tests

### DIFF
--- a/docs/configuration/node-config.md
+++ b/docs/configuration/node-config.md
@@ -191,7 +191,7 @@ This section contains the configuration options for a Searcher.
 
 | Property | Description | Default value |
 | --- | --- | --- |
-| `aggregation_memory_limit` | Controls the maximum amount of memory that can be used for aggregations before aborting. This limit is per request and single leaf query (a leaf query is querying one or multiple splits concurrently). It is used to prevent excessive memory usage during the aggregation phase, which can lead to performance degradation or crashes. Since it is per request, concurrent requests can exceed the limit. | `500M`|
+| `aggregation_memory_limit` | Controls the maximum amount of memory that can be used for aggregations before aborting. This limit is per searcher node. A node may run concurrent queries, which share the limit. The first query that will hit the limit will be aborted and frees its memory. It is used to prevent excessive memory usage during the aggregation phase, which can lead to performance degradation or crashes. | `500M`|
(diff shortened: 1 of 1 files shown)


quickwit/rest-api-tests/scenarii/sort_orders/0001-sort-elasticapi.yaml
```diff
@@ -9,17 +9,17 @@ json:
         match:
           count: 10
   sort:
-    - count: {"order" : "desc"}
+    - id: {"order" : "desc"}
```
fulmicoton, line 12: what is this change about?

quickwit/rest-api-tests/scenarii/sort_orders/0001-sort-elasticapi.yaml
```diff
@@ -29,16 +29,16 @@ json:
         match:
           count: 10
   sort:
-    - count: {"order" : "asc"}
+    - id: {"order" : "asc"}
```
fulmicoton, line 32: same question as above.

fulmicoton (changes requested):
If I understand correctly, this PR makes it so that a new query will fail if running the query would cause the allocated memory over the whole node to exceed `aggregation_memory_limit`, correct?

Can we document that on the docstring of `aggregation_memory_limit` and in the documentation of Quickwit?

quickwit/rest-api-tests/scenarii/sort_orders/0001-sort-elasticapi.yaml
PSeitz: this fixes the flakiness in the test

PSeitz:
> If I understand correctly, this PR makes it so that a new query will fail if running the query would cause the allocated memory over the whole node to exceed `aggregation_memory_limit`, correct?
> 

Yes, except we need to execute the query and will abort it once it exceeds the memory, as we don't know up front how much memory it needs. (would sounds like we could do that up front)

> Can we document that on the docstring of `aggregation_memory_limit` and in the documentation of Quickwit?

Updated the docs

fulmicoton (approved):

Issue #5351, opened 2024-08-27
Ingest V2 misses commit `wait_for`


rdettai:
**Is your feature request related to a problem? Please describe.**
Similar to https://github.com/quickwit-oss/quickwit/issues/4438 that tracks the behavior of `CommitType::Force` in ingest V2, this issue is about `CommitType::wait` in ingest V2. See the [docs](https://quickwit.io/docs/reference/rest-api#controlling-when-the-indexed-documents-will-be-available-for-search) for the specs.

**Describe the solution you'd like**
Ingest V2 should behave like ingest V1 when using `?commit=wait_for`

**Describe alternatives you've considered**
Remove the feature for ingest V2.

**Additional context**
With ingest V1, indexing was happening on the same node as the one receiving the query. This is not true anymore with ingest V2, so the notification system that makes the waiting possible should happen at the cluster level. Possible solutions:
- track commit progress and push notifications from the control plane
- communicate back the commits directly from the indexer to the original ingester

rdettai:
https://github.com/quickwit-oss/quickwit/issues/4438#issuecomment-2312860539

PR #5359, opened 2024-08-28
Fix wait_for commit on ingest V2

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/5351

### How was this PR tested?


rdettai:
Merging this to upstream branch because `commit=force` also requires to wait (see [docs](https://quickwit.io/docs/reference/rest-api#controlling-when-the-indexed-documents-will-be-available-for-search)

PR #5360, opened 2024-08-28
Disable control plane check for searcher.

fmassot:


guilload:
I'd get rid of the env var. We never need the control plane check for searchers anyway.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -338,7 +338,10 @@ async fn start_control_plane_if_needed(
             balance_channel_for_service(cluster, QuickwitService::ControlPlane).await;
 
         // If the node is a metastore, we skip this check in order to avoid a deadlock.
-        if !node_config.is_service_enabled(QuickwitService::Metastore) {
+        // If the node is a searcher, we skip this check because the searcher does not need to.
+        if !(node_config.is_service_enabled(QuickwitService::Metastore)
+            || node_config.is_service_enabled(QuickwitService::Searcher))
```
trinity-1686a, line 343: I think this is wrong: unless i'm missing something it's telling "if metastore is disabled AND searcher is disabled", connect to control plane. This means if this node is searcher+indexer, don't connect to the metastore

What we want is `if !(is_service_enabled(Metastore) OR enabled_services.filter(service != Searcher).is_empty)`

fulmicoton: I agree with @trinity-1686a . (except I believe the expression can simply be

`enabled_services.filter(service != Searcher).is_empty()` (as it includes the metastore)

PR #5361, opened 2024-08-28
add jitter to retention schedule

### LINKED ISSUE #5353: Make sure the janitor does not spam the metastore too much
In https://github.com/quickwit-oss/quickwit/pull/5346 we have spotted that our implementation of delete index was too aggressive.

For airmail, their internal job deleting a large number of indexes ended up hammering the metastore, hence disrupting indexing.

We want to make sure that we don't have a similar pattern in the janitor. In particular, when running the retention policy.

trinity-1686a:
### Description

add optional jitter to retention policy schedule, so that numerous indexes with the same schedule don't hammer down the metastore when they all execute their retention policy
fix #5353

### How was this PR tested?

unit test


guilload:
I wonder if we could avoid exposing users to one more knob by jittering by default. We would have to pick a value between [next evaluation..next evaluation+1], maybe biased towards the beginning of the interval. WDYT? Would that be confusing for users?

trinity-1686a:
i think it would be fine for policies which happen often enough (once every day or more often), but could get confusing if policies put in more delays, especially when entered in cron format. For instance if i put `0 0 * * 0` (every sunday at midnight), I wouldn't be too surprised if the task starts a few minutes or maybe an hour late, but i would be surprised if it started doing something on tuesday (or any other day). We could also run into issues where if we do something daily (0utc assumed), and for some reason it's company policy to restart servers at 1am, only 1/24 indexes would actually be cleaned each day (the other 23/24 are scheduled after 1am, then we restart, look at what the next time to run would be, it's midnight + some delay, sleep for the rest of the day)

We could something where the jitter_secs is `min(1h, time_between_2_schedules)`, which sounds generally less surprising, and maybe (or maybe not) allow manual configuration to something higher than 1h?

guilload:
Yes, `min(1h, time_between_2_schedules)` is perfect. I meant something along those lines with "biased".

quickwit/quickwit-config/src/index_config/mod.rs
```diff
@@ -207,6 +180,16 @@ pub struct RetentionPolicy {
...
+
+    /// A jitter to apply to the schedule. The policy will be evaluated [0..jitter_second] seconds
+    /// after the scheduled time. When many indexes use the same schedule, this can be used to
+    /// spread the load instead of causing a very bursty load.o
+    ///
+    /// If unset, a default jitter of `min(1 hour, next_next_evaluation - next_evaluation)` is
+    /// applied. Said otherwise, an operation may start any time between the next time it's
+    /// scheduled, and the time after that, but no later than 1h after the scheduled time.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub jitter_secs: Option<u64>,
```
guilload: nit: human time is nicer

quickwit/quickwit-config/src/index_config/mod.rs
```diff
@@ -236,13 +219,24 @@ impl RetentionPolicy {
...
-        let duration = (future_date - Utc::now())
+        let mut duration = (future_date - Utc::now())
             .to_std()
             .map_err(|err| anyhow::anyhow!(err.to_string()))?;
+        let jitter_secs = self.jitter_secs.unwrap_or_else(|| {
+            if let Some(next_next_date) = schedule_iter.next() {
+                let time_between_schedules = next_next_date - future_date;
+                time_between_schedules.num_seconds().clamp(0, 3600) as u64
+            } else {
+                // we don't know when the schedule is. That's odd. Let's allow no jitter
```
guilload: warn!

guilload (approved):
This was a nice one :) Let's merge it.

quickwit/quickwit-config/src/index_config/mod.rs
```diff
@@ -320,6 +321,18 @@ pub struct RetentionPolicy {
...
+    /// after the scheduled time. When many indexes use the same schedule, this can be used to
+    /// spread the load instead of causing a very bursty load.o
+    ///
+    /// If unset, a default jitter of `min(1 hour, next_next_evaluation - next_evaluation)` is
+    /// applied. Said otherwise, an operation may start any time between the next time it's
+    /// scheduled, and the time after that, but no later than 1h after the scheduled time.
+    #[serde(default, deserialize_with = "parse_human_duration_opt")]
+    #[serde(serialize_with = "serialize_duration_opt")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub jitter: Option<Duration>,
```
guilload:
```suggestion
    pub evaluation_schedule_jitter: Option<Duration>,
```
:x

quickwit/quickwit-config/src/index_config/mod.rs
```diff
@@ -366,6 +394,27 @@ impl RetentionPolicy {
     }
 }
 
+fn parse_human_duration_opt<'de, D>(deserializer: D) -> Result<Option<Duration>, D::Error>
```
guilload: I think we have a helper or a proper type for that now.

guilload (approved):

PR #5365, opened 2024-08-29
add more env var config for postgres

trinity-1686a:
### Description

add environment variables related to postgres metastore:
- `QW_POSTGRES_SKIP_MIGRATIONS`: don't run migrations, but verifies all migrations were already run
- `QW_POSTGRES_SKIP_MIGRATION_LOCKING`: prevent locking database for migrations. This might make us more compatible with cockroachdb. We might not be entirely compatible still, as it [doesn't support DDL in transaction](https://www.cockroachlabs.com/docs/stable/online-schema-changes).
- `QW_POSTGRES_READ_ONLY`: make all connections read-only by default. This isn't really a security feature as we can do `BEGIN READ WRITE` to get a rw transaction, but it should prevent involuntary writes. While this doesn't imply `QW_POSTGRES_SKIP_MIGRATIONS`, being read only and running migrations always fail (even if migrations are up to date)

### How was this PR tested?

tested manually against existing databases with missing and all migrations, and added unit test.


docs/reference/cli.md
```diff
@@ -838,6 +838,15 @@ Disables [telemetry](../telemetry.md) when set to any non-empty value.
 
 `QW_DISABLE_TELEMETRY=1 quickwit help`
 
+### QW_POSTGRES_SKIP_MIGRATIONS
+
+Don't run database migrations (but verify migrations were run succesfully before, and no unknown migration was run).
```
guilload, line 843:
```suggestion
Don't run database migrations (but verify that migrations were run successfully before and that no unknown migration was run).
```

docs/reference/cli.md
```diff
@@ -838,6 +838,15 @@ Disables [telemetry](../telemetry.md) when set to any non-empty value.
...
+
+### QW_POSTGRES_SKIP_MIGRATION_LOCKING
+
+Don't lock the database during migration. This may increase compatiblity with altenrative databases using the postgres wire protocol. It is dangerous
```
guilload, line 847:
```suggestion
Don't lock the database during migration. This may increase compatibility with alternative databases using the PostgreSQL wire protocol. However, it is dangerous to use this if you can't guarantee that only one node will run the migrations.
```

quickwit/quickwit-metastore/migrations/postgresql/21_add-shard-update-timestamp-field.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE shards
+    DROP IF EXISTS update_timestamps;
```
guilload, line 2:
```suggestion
    DROP IF EXISTS update_timestamp;
```

guilload (approved):

Issue #5311, opened 2024-08-13
[otel log] Handle logs without `Timestamp` field better.


c0per:
**Describe the bug**

Currently, quickwit is ignoring otel logs with a `null` (or `0`) `Timestamp`.
```rust
// quickwit-opentelemetry::otlp::logs.rs line 441
for log_record in scope_log.log_records {
    if log_record.time_unix_nano == 0 {
        rate_limited_error!(limit_per_min = 10, "skipping record");
        num_parse_errors += 1;
        continue;
    }
    // ...
}
```

Some instrumentation only set `ObservedTimestamp` and leave `Timestamp` null.
Maybe we should default to `ObservedTimestamp` when `Timestamp` is not set.

**Steps to reproduce**
Use any Otel logs exporter which doesn't set the timestamp field.
In my case, I'm using rust tracing + `opentelemetry-appender-tracing`.

**Expected behavior**

Other people have encountered this problem using the same setup with the official otel collector (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/11752#issuecomment-1895383712).

The official otel collector exporter use `ObservedTimestamp` when `Timestamp` is not set.
The PR: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31844

Maybe Quickwit should also do this. Default to `ObservedTimestamp` when `Timestamp` is not set, instead of throwing away logs without a timestamp.

spetz:
I've just encountered the same issue - traces work great, but the logs are silently ignored with the same error.

mfaerevaag:
Experiencing the same issue as @spetz. Fingers crossed for https://github.com/quickwit-oss/quickwit/pull/5366

PR #5366, opened 2024-08-29
Change default timestamps in OTEL logs

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/5311

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-opentelemetry/src/otlp/logs.rs
+++ b/quickwit/quickwit-opentelemetry/src/otlp/logs.rs
@@ -22,7 +22,6 @@ use std::collections::{btree_set, BTreeSet, HashMap};
 
 use async_trait::async_trait;
 use prost::Message;
-use quickwit_common::rate_limited_error;
 use quickwit_common::thread_pool::run_cpu_intensive;
 use quickwit_common::uri::Uri;
 use quickwit_config::{load_index_config_from_user_config, ConfigFormat, IndexConfig};
@@ -36,6 +35,7 @@ use quickwit_proto::opentelemetry::proto::collector::logs::v1::{
 use quickwit_proto::types::{DocUidGenerator, IndexId};
 use serde::{Deserialize, Serialize};
 use serde_json::Value as JsonValue;
+use time::OffsetDateTime;
 use tonic::{Request, Response, Status};
 use tracing::field::Empty;
 use tracing::{error, instrument, warn, Span as RuntimeSpan};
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-opentelemetry/src/otlp/logs.rs
```diff
@@ -446,16 +443,27 @@ fn parse_otlp_logs(
...
 
             for log_record in scope_log.log_records {
-                if log_record.time_unix_nano == 0 {
-                    rate_limited_error!(limit_per_min = 10, "skipping record");
-                    num_parse_errors += 1;
-                    continue;
-                }
-                let observed_timestamp_nanos = if log_record.observed_time_unix_nano != 0 {
-                    Some(log_record.observed_time_unix_nano)
+                let observed_timestamp_nanos = if log_record.observed_time_unix_nano == 0 {
```
fmassot, line 446: I understand what you meant (great comment by the way). But I don't want to fill a field that we don't really use on quickwit side. It will just take more space.

I would just set `timestamp_nanos` to `observed_timestamp_nanos` if there is one. I consider the user responsible for setting at least one timestamp info. If there is nothing in the record, then we should reject it so the user realizes he's doing bad things.

quickwit/quickwit-opentelemetry/src/otlp/logs.rs
```diff
@@ -47,7 +47,7 @@ use super::{
 use crate::otlp::extract_attributes;
 use crate::otlp::metrics::OTLP_SERVICE_METRICS;
 
-pub const OTEL_LOGS_INDEX_ID: &str = "otel-logs-v0_7";
+pub const OTEL_LOGS_INDEX_ID: &str = "otel-logs-v0_9";
```
fmassot, line 50: I would not change the index ID as there is not change in the schema.

rdettai: Yes, I hesitated. My rational was mostly that it would be less confusing for users if logs and traces have the same version number.

quickwit/quickwit-opentelemetry/src/otlp/logs.rs
rdettai: 1. I think it should never happen, sending OTEL without ObservedTimestamp seems like be a pretty big disrespect for the standard :smile: 
2. If it did happen, it would be hard for the user to fix the exporter side as the issue would likely come from some tooling they don't really control. I really think that preventing the user from using Quickwit in that (unlikely) case is a bad idea.

fmassot: Let's keep what you did then.

quickwit/quickwit-opentelemetry/src/otlp/logs.rs
fmassot: ok

fmassot (approved):

PR #5374, opened 2024-09-02
Debounce shard pruning request

rdettai:
### Description

Instead of running a separate process to clean up the shard table, we initiate the cleanup requests from the files sources. To avoid an increasing number of requests when the number of pipelines grows the requests are debounced in the control plane.

### How was this PR tested?

Integration tests


quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -772,6 +778,49 @@ impl Handler<DeleteSourceRequest> for ControlPlane {
...
+#[async_trait]
+impl Handler<DebouncedPruneShardsRequest> for ControlPlane {
+    type Reply = ControlPlaneResult<EmptyResponse>;
+
+    async fn handle(
+        &mut self,
+        request: DebouncedPruneShardsRequest,
+        ctx: &ActorContext<Self>,
+    ) -> Result<ControlPlaneResult<EmptyResponse>, ActorExitStatus> {
+        let metastore_request = request.request.unwrap();
```
fulmicoton, line 790:
```suggestion
        let metastore_request = request.request.unwrap();
```
no unwrap here.

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -71,13 +72,16 @@ pub(crate) const CONTROL_PLAN_LOOP_INTERVAL: Duration = if cfg!(any(test, featur
...
+const PRUNE_SHARDS_DEFAULT_COOLDOWN_PERIOD: Duration = Duration::from_secs(120);
+
 /// Minimum period between two rebuild plan operations.
 const REBUILD_PLAN_COOLDOWN_PERIOD: Duration = Duration::from_secs(2);
 
 #[derive(Debug)]
 struct ControlPlanLoop;
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
```
fulmicoton, line 84:
```suggestion
#[derive(Debug, Default, Clone, Copy)]
```

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -772,6 +778,49 @@ impl Handler<DeleteSourceRequest> for ControlPlane {
...
+
+        // Update the cooldown period in case the config changes. If multiple
+        // indexers run with different configs, the control plane ends up
+        // running one of them in a seemingly arbitrary way.
+        if let Some(interval) = metastore_request.interval {
+            debouncer.set_next_cooldown_period(Duration::from_secs(interval as u64));
+        }
+        debouncer.self_send_with_cooldown(
+            ctx,
+            DebouncedPruneShardsRequest {
```
fulmicoton, line 815: Shouldn't this be two messages? DebouncedPruneShardRequests and PruneShardsRequest ?
(without a execute_now boolean)

quickwit/quickwit-control-plane/src/debouncer.rs
```diff
@@ -119,18 +119,23 @@ impl Debouncer {
     pub fn self_send_with_cooldown<M>(
         &self,
         ctx: &ActorContext<impl Handler<M> + DeferableReplyHandler<M>>,
+        message: M,
```
fulmicoton, line 122:
```suggestion
        message: M,
        ctx: &ActorContext<impl Handler<M> + DeferableReplyHandler<M>>,   
```

we tend to make ctx the last argument

quickwit/quickwit-control-plane/src/debouncer.rs
```diff
@@ -91,25 +91,25 @@ impl Debouncer {
...
         A: Actor + Handler<M> + DeferableReplyHandler<M>,
-        M: Default + std::fmt::Debug + Send + Sync + 'static,
+        M: std::fmt::Debug + Send + Sync + 'static,
     {
-        let _ = ctx.mailbox().send_message_with_high_priority(M::default());
+        let _ = ctx.mailbox().send_message_with_high_priority(message);
     }
 
-    fn schedule_post_cooldown_callback<A, M>(&self, ctx: &ActorContext<A>)
+    fn schedule_post_cooldown_callback<A, M>(&self, ctx: &ActorContext<A>, message: M)
```
fulmicoton, line 102: I think there are a couple of ugly things here, but maybe this is ok.

The reason why I did not put `message` in the API is because this suppose a bit of a footgun.
If someone sends different messages, the message that will be shadowed through deboucning will be quite random.

Another thing I don't like is that `schedule_post_cooldown_callback` taking message makes no sense at all today. It is an artefact of the implementation.

Can you see if you can remove it by storing the message within the `DebouncerState::CooldownScheduled(Message)` maybe?

quickwit/quickwit-control-plane/src/control_plane.rs
rdettai: I don't see any harm in implementing Copy here, but what is the actual benefit?

quickwit/quickwit-control-plane/src/control_plane.rs
rdettai: I hesitated. 2 messages are probably a bit more understandable.

quickwit/quickwit-control-plane/src/control_plane.rs
rdettai: nevermind, irrelevant if considering [this comment](https://github.com/quickwit-oss/quickwit/pull/5374#discussion_r1764588527)

quickwit/quickwit-control-plane/src/debouncer.rs
rdettai, line 1: No real change here, just simplifying the trait bounds for readability

quickwit/quickwit-control-plane/src/debouncer.rs
rdettai: I decided to go with a much simpler debouncing as we don't really care if we miss out on some pruning requests

quickwit/quickwit-control-plane/src/debouncer.rs
fulmicoton: thx

quickwit/quickwit-proto/protos/quickwit/metastore.proto
```diff
@@ -458,14 +458,11 @@ message PruneShardsRequest {
   quickwit.common.IndexUid index_uid = 1;
   string source_id = 2;
   // The maximum age of the shards to keep, in seconds.
-  optional uint32 max_age_secs = 5;
+  optional uint32 max_age_secs = 3;
```
fulmicoton, line 461: do not remap ids in protobuf.

fulmicoton: https://protobuf.dev/programming-guides/dos-donts/#reuse-number

fulmicoton (approved):

PR #5375, opened 2024-09-02
ingestv2 doc validation improvements

### LINKED ISSUE #5164: The ingest API document validation does not check for the presence of the timestamp field.

### LINKED ISSUE #5229: Disable document validation if VRL is here
(as spotted by @trinity-1686a)

We recently introduced document validation in ingest v2.

However, if a VRL script is configured, it could well modify an invalid document into a valid one.

trinity-1686a:
### Description

fix #5164
fix #5229

### How was this PR tested?

added unit tests


quickwit/quickwit-ingest/src/ingest_v2/models.rs
```diff
@@ -55,6 +55,9 @@ pub(super) struct IngesterShard {
     pub is_advertisable: bool,
     /// Document mapper for the shard. Replica shards and closed solo shards do not have one.
     pub doc_mapper_opt: Option<Arc<dyn DocMapper>>,
+    /// Whether to validate documents in this shard. True if no preprocessing (VLR) happends before
```
fulmicoton, line 58:
```suggestion
    /// Whether to validate documents in this shard. True if no preprocessing (VRL) happened before
```

fulmicoton (approved):

guilload:
I don't think the part of the PR disabling doc validation if a VRL transform exists for the ingest source was necessary since adding a VRL transform for that source via a supported API is impossible. If users do so by manually editing the source, it's their problem anyway.

In the long run, we also want to perform VRL transforms during the ingest phase so users can get immediate feedback.

guilload:
Don't bother. We can revert it when we also perform VRL transforms during ingestion.

PR #5382, opened 2024-09-04
Add timeout to ingester decommissioning

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/5068

### How was this PR tested?

Integration tests

### DIFF
--- a/quickwit/quickwit-serve/src/lib.rs
+++ b/quickwit/quickwit-serve/src/lib.rs
@@ -760,7 +760,9 @@ pub async fn serve_quickwit(
         // We must decommission the ingester first before terminating the indexing pipelines that
         // may consume from it. We also need to keep the gRPC server running while doing so.
         if let Some(ingester) = ingester_opt {
-            if let Err(error) = wait_for_ingester_decommission(ingester).await {
+            if let Err(error) =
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-serve/src/lib.rs
```diff
@@ -760,7 +760,9 @@ pub async fn serve_quickwit(
         // We must decommission the ingester first before terminating the indexing pipelines that
         // may consume from it. We also need to keep the gRPC server running while doing so.
         if let Some(ingester) = ingester_opt {
-            if let Err(error) = wait_for_ingester_decommission(ingester).await {
+            if let Err(error) =
+                wait_for_ingester_decommission(ingester, Duration::from_secs(5)).await
```
rdettai, line 764: 5s is arbitrary and probably a bit short. We could also only timeout when we fail to reach the CP/metastore and keep retrying for other kinds of errors

guilload:
I'm not comfortable with this change. This is fine when an ingester is decommissioned as part of a restart but can be dangerous and lead to data loss when a node is fully decommissioned, i.e., shutdown, and removed from the cluster without further access to disks. This is why I did not add a timeout in the first place when implementing the decommissioning.

Issue #5211, opened 2024-07-10
GCS allows only 1 write/s per key


trinity-1686a:
Per [GCS documentation](https://cloud.google.com/storage/quotas#objects), we may not write an object more than once a second. For most objects, we never write more than once, but for the metastore, we may update it more frequently than that, which causes errors

trinity-1686a:
retry is now implemented. Keeping the ticket opened there for a future where we may batch metastore writes

PR #5384, opened 2024-09-04
make some s3 error retryable

trinity-1686a:
### Description

mark some S3 errors as retry-able. The code in qw-storage already uses `aws_retry()` on write requests. The actual error we get from GCS is `SlowDown`, but aws sdk contains a list of errors, so we that instead of hard-coding a single value;
improvement for #5211

### How was this PR tested?

added unit test to Storage::put for retry.

### DIFF
--- a/quickwit/quickwit-aws/src/error.rs
+++ b/quickwit/quickwit-aws/src/error.rs
@@ -19,6 +19,7 @@
 
 #![allow(clippy::match_like_matches_macro)]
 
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-aws/src/error.rs
```diff
@@ -19,6 +19,7 @@
 
 #![allow(clippy::match_like_matches_macro)]
 
+use aws_runtime::retries::classifiers::{THROTTLING_ERRORS, TRANSIENT_ERRORS};
```
guilload, line 22: Nice find.

guilload (approved):

PR #5377, opened 2024-09-03
memoize s3 client

trinity-1686a:
### Description

memoize s3 client, keeping them until they seem unused. I don't think we ever use more than a single S3StorageConfig, so we could store a single value, but a map we gc seems less confusing if we start having multiple config at once
fix #4933 
fix #5236

### How was this PR tested?

ran a cluster and looked at searcher logs. Didn't look at janitor logs, though there is no reason for it to behave differently

PR #5387, opened 2024-09-05
improve logging

PSeitz:
log actual index_id patterns that are queried on leaf nodes
reduce log spam

### DIFF
--- a/quickwit/quickwit-config/src/storage_config.rs
+++ b/quickwit/quickwit-config/src/storage_config.rs
@@ -18,6 +18,7 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use std::ops::Deref;
+use std::sync::OnceLock;
 use std::{env, fmt};
 
 use anyhow::ensure;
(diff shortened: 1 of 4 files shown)


fulmicoton (approved):

quickwit/quickwit-config/src/storage_config.rs
```diff
@@ -370,11 +371,14 @@ impl S3StorageConfig {
     }
 
     pub fn force_path_style_access(&self) -> Option<bool> {
-        let force_path_style_access = get_bool_from_env(
-            "QW_S3_FORCE_PATH_STYLE_ACCESS",
-            self.force_path_style_access,
-        );
-        Some(force_path_style_access)
+        static FORCE_PATH_STYLE: OnceLock<Option<bool>> = OnceLock::new();
```
fulmicoton, line 374: Have you discovered if all that logging was justified? Is this normal that we created that many S3Storage?

PSeitz: It was not unreasonable, but could be improved with caching, which is covered by https://github.com/quickwit-oss/quickwit/pull/5377 (not sure if it completely resolves the issue)

PR #5389, opened 2024-09-05
Fix actor timeout in ingest integration tests

rdettai:
### Description

Integration tests were sometimes failing because actor heartbeat is smaller in integ tests.

### How was this PR tested?

Running the OTLP integ tests in a loop never fails anymore.

### DIFF
--- a/quickwit/quickwit-indexing/src/source/mod.rs
+++ b/quickwit/quickwit-indexing/src/source/mod.rs
@@ -90,7 +90,7 @@ pub use gcp_pubsub_source::{GcpPubSubSource, GcpPubSubSourceFactory};
 pub use kafka_source::{KafkaSource, KafkaSourceFactory};
 #[cfg(feature = "kinesis")]
 pub use kinesis::kinesis_source::{KinesisSource, KinesisSourceFactory};
-use once_cell::sync::OnceCell;
+use once_cell::sync::{Lazy, OnceCell};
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-indexing/src/source/mod.rs
```diff
@@ -138,7 +138,19 @@ use crate::source::ingest_api_source::IngestApiSourceFactory;
 /// 5MB seems like a good one size fits all value.
 const BATCH_NUM_BYTES_LIMIT: u64 = ByteSize::mib(5).as_u64();
 
-const EMIT_BATCHES_TIMEOUT: Duration = Duration::from_millis(if cfg!(test) { 100 } else { 1_000 });
+static EMIT_BATCHES_TIMEOUT: Lazy<Duration> = Lazy::new(|| {
+    if cfg!(any(test, feature = "testsuite")) {
```
trinity-1686a, line 142: so the underlying issue was `feature = "testsuite"` missing here, and the rest is just to warn about bad environments?

rdettai: exactly!

rdettai: being able to tweak the actor heartbeat is a pretty big footgun, so I think it doesn´t hurt to warn against very bad settings

PR #5390, opened 2024-09-05
Trigger early commit when shutting down ingest V2

rdettai:
### Description

Currently the shutdown waits for the shards to be indexed without triggering an early commit, so the shutdown can take as long as the longest commit timeout.

### How was this PR tested?

Integ tests


rdettai:
Gravitates around https://github.com/quickwit-oss/quickwit/issues/5068

rdettai:
As discussed internally, this new mechanism adds too much complexity. We will favor simpler workarounds for now:
- increase termination grace periods in Helm charts and ECS configs
- add a listener for second ctrl-c when running nodes in experimentation/tutorial contexts
- document shutdown behavior

PR #5392, opened 2024-09-05
fix lints from 1.79-1.81

trinity-1686a:
### Description

fix lints from rust/clippy 1.79-1.81

### DIFF
--- a/quickwit/quickwit-indexing/Cargo.toml
+++ b/quickwit/quickwit-indexing/Cargo.toml
@@ -101,6 +101,7 @@ testsuite = [
   "quickwit-storage/testsuite"
 ]
 vrl = ["dep:vrl", "quickwit-config/vrl"]
+ci-test = []
 
 [dev-dependencies]
 bytes = { workspace = true }
--- a/quickwit/quickwit-ingest/src/metrics.rs
+++ b/quickwit/quickwit-ingest/src/metrics.rs
@@ -28,6 +28,7 @@ pub struct IngestMetrics {
 
     pub replicated_num_bytes_total: IntCounter,
     pub replicated_num_docs_total: IntCounter,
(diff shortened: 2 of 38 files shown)


quickwit/quickwit-indexing/Cargo.toml
```diff
@@ -101,6 +101,7 @@ testsuite = [
   "quickwit-storage/testsuite"
 ]
 vrl = ["dep:vrl", "quickwit-config/vrl"]
+ci-test = []
```
trinity-1686a, line 104: I don't think this feature is ever used, but some tests depends on it. We might want to enable it somewhere

quickwit/quickwit-ingest/src/metrics.rs
```diff
@@ -28,6 +28,7 @@ pub struct IngestMetrics {
 
     pub replicated_num_bytes_total: IntCounter,
     pub replicated_num_docs_total: IntCounter,
+    #[allow(dead_code)] // this really shouldn't be dead, it needs to be used somewhere
     pub queue_count: IntGauge,
```
trinity-1686a, line 32: this ought to be used somewhere

quickwit/quickwit-search/Cargo.toml
```diff
@@ -60,3 +60,4 @@ quickwit-storage = { workspace = true, features = ["testsuite"] }
 
 [features]
 testsuite = []
+ci-test = []
```
trinity-1686a, line 63: I don't think this feature is ever used, but some tests depends on it. We might want to enable it somewhere

rdettai (approved):
we should probably update the Rust version in `rust-toolchain.toml` quickly to prevent the formatting from drifting again

PR #5395, opened 2024-09-06
shorten delete_splits_marked_for_deletion log

trinity-1686a:
### Description

make `delete_splits_marked_for_deletion` shorter. Before it used to show only a single index id, but with gc batching, it can now show thousands at once when that function returns an error.

### DIFF
--- a/quickwit/quickwit-index-management/src/garbage_collection.rs
+++ b/quickwit/quickwit-index-management/src/garbage_collection.rs
@@ -166,7 +166,7 @@ pub async fn run_garbage_collect(
     )
     .await
 }
-#[instrument(skip(storages, metastore, progress_opt))]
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-index-management/src/garbage_collection.rs
```diff
@@ -166,7 +166,7 @@ pub async fn run_garbage_collect(
     )
     .await
 }
-#[instrument(skip(storages, metastore, progress_opt))]
+#[instrument(skip(index_uids, storages, metastore, progress_opt), fields(num_index=%index_uids.len()))]
```
rdettai, line 169:
```suggestion
#[instrument(skip(index_uids, storages, metastore, progress_opt), fields(num_indexes=%index_uids.len()))]
```

rdettai (approved):

Issue #5241, opened 2024-07-21
Abort/timeout long running requests


PSeitz:
Currently long running requests may continue to consume a large amount resources, while the original request already timed-out in a http-layer above. We should add a configurable timeout to abort requests in quickwit

There are two different timeouts configured:
### Tonic  Timeout
tonic timeouts return errors like:
```
{
  "message": "internal error: `Timeout expired`"
}
```
This timeout is originating here:
```rust
let timeout_channel = Timeout::new(node.channel(), Duration::from_secs(30));
let search_client = create_search_client_from_channel(
    grpc_addr,
    timeout_channel,
    max_message_size,
);
Some(Change::Insert(grpc_addr, search_client))
```

### Tower Timeout
There's another similar timeout, from tower
```
{
  "message": "internal error: `request timed out`"
{
```

The tower timeout is defined here:
```rust
/// Creates a channel from a socket address.
///
/// The function is marked as `async` because it requires an executor (`connect_lazy`).
pub async fn make_channel(socket_addr: SocketAddr) -> Channel {
    let uri = Uri::builder()
        .scheme("http")
        .authority(socket_addr.to_string())
        .path_and_query("/")
        .build()
        .expect("provided arguments should be valid");
    Endpoint::from(uri)
        .connect_timeout(Duration::from_secs(5))
        .timeout(Duration::from_secs(30))
        .connect_lazy()
}
```

The behavior of timeouts also differs currently on the dispatch type, local dispatch never times out.

### Retries

The current behavior is to retry directly after a timeout (`cluster_client.rs::leaf_search`). This could add additional load on a overloaded node.

The first step would be to properly transform the error type to keep the semantics

PSeitz:
Fixed by https://github.com/quickwit-oss/quickwit/pull/5402

PR #5402, opened 2024-09-09
add request_timeout_secs config to searcher config

PSeitz:
All GRPC get a 30s timeout, but the search RPCs, that have their own configurable timeout.

### Current behavior before this PR
We have two fixed timeouts of when sending search requests between nodes, 30s on the channel and 60s on the search client. On very large datasets these fixed timeouts are too low for some queries. Local search does not have any timeout, which leads to inconsistent behavior. When converting between the errors, timeouts are converted to internal errors, loosing the semantics. The retry logic does not take timeout error into account, which could further overload the node.
This PR addresses these issues as outlined below.

Related: https://github.com/quickwit-oss/quickwit/issues/5241

### Configurable Timeout

This PR adds a new setting:
- `request_timeout_secs` on the searcher config (Defaults to 30sec)

The timeout on the channel in the cluster is removed and we introduce a new tower timeout layer. That layer is similar to the timeout layer provided by tower, but requires to have a conversion from the `TimeoutExceeded` error to the services error. This allows us to forward the timeout semantics.

### Cancel Search

This PR will add some logic to cancel the search on the leaf node if it exceeds the configured timeout. Previously it would have needlessly continued to consume resources.
It also removes the retry if we timeout, to not further overload the node.

### How was this PR tested?

Manually tested the node config and the timeout on a distributed setting and singel node search.

### Future work

Convert timeout errors to Quickwit errors for the search case


fulmicoton (changes requested):
Can we put the request timeout on search request only, and put that conf on the searcher config?

fulmicoton (changes requested):
Can we put the request timeout on search request only, and put that conf on the searcher config?

PSeitz:
> Can we put the request timeout on search request only, and put that conf on the searcher config?

It's what I started with, but the `Channel` in `make_channel`, which is used in quickwit-cluster, is used for multiple services e.g. search and indexing. So we can't have a search specific timeout there.

We could remove the timeout there and try to move to other layers, or reopen another channel without the timeout in the search case.

PSeitz:
As discussed offline, the generic channel timeout is moved to a tower timeout

fulmicoton (changes requested):
Why not use?
https://tower-rs.github.io/tower/tower/timeout/struct.TimeoutLayer.html

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -1145,6 +1147,7 @@ fn setup_indexer_pool(
                     } else {
                         let client = IndexingServiceClient::tower()
                             .stack_layer(INDEXING_GRPC_CLIENT_METRICS_LAYER.clone())
+                            .stack_layer(TimeoutLayer::new(Duration::from_secs(30)))
```
fulmicoton, line 1150: if you use the layer in tower, I think here is a helper function that goes 
`.timeout(duration)`

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1054,13 +1055,43 @@ impl CanSplitDoBetter {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct TimeoutHandle {
```
fulmicoton, line 1059:
```suggestion
struct TimeoutHandle {
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1226,6 +1263,7 @@ pub async fn leaf_search(
     let incremental_merge_collector = Arc::new(Mutex::new(incremental_merge_collector));
 
     for (split, mut request) in split_with_req {
+        timeout.err_on_timeout()?;
```
fulmicoton, line 1266: I think if you timeout on the server side using a tower layer, it will cancel the future and do what you want without the TimeoutHandle.

PSeitz: I tested the current behavior and the client side was unaffected from the server timeouts and kept running. There's no special logic in the tower timeout layer to forward the timeout to the client side, so not sure how that could be the case.

But there's another problem: If we dispatch locally, it doesn't use the grpc stuff, so we need to have some timeout for that case.

I replaced `TimeoutHandle` with a tokio timeout, that should be much nicer overall.

PSeitz:
> Why not use? https://tower-rs.github.io/tower/tower/timeout/struct.TimeoutLayer.html

It's the same, but the main difference is:

```
/// The error of the service must implement `From<TimeoutExceeded>`.
```

So we keep the semantics of the timeout, this let's us e.g. easily disable retries.
The tower timeout just uses a `Box<Error>`.

quickwit/quickwit-search/src/leaf.rs
PSeitz: replaced `TimeoutHandle` with a tokio timeout

quickwit/quickwit-serve/src/lib.rs
PSeitz: Yes, with the manual TimeoutLayer we don't have access to that, but internally it does the same

```rust
self.layer(crate::timeout::TimeoutLayer::new(timeout))
```

fulmicoton, line 1150:
```suggestion
                            .stack_layer(TimeoutLayer::new(GRPC_TIMEOUT))
```

quickwit/quickwit-common/src/tower/timeout.rs
```diff
@@ -0,0 +1,145 @@
...
+        match this.sleep.poll(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(_) => Poll::Ready(Err(TimeoutExceeded.into())),
+        }
+    }
+}
+
+/// Applies timeouts on the inner service.
+///
+/// If the inner service does not complete within the specified duration,
```
fulmicoton, line 100:
```suggestion
/// This is similar to tower's Timeout Layer except it gives us more 
/// over the error emitted. If the inner service does not complete within the specified
/// duration, the response will be aborted with the error `TimeoutExceeded`.
///
/// The error of the service must implement `From<TimeoutExceeded>`.
```

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -946,6 +946,7 @@ async fn setup_ingest_v2(
                     } else {
                         let ingester_service = IngesterServiceClient::tower()
                             .stack_layer(INGEST_GRPC_CLIENT_METRICS_LAYER.clone())
+                            .stack_layer(TimeoutLayer::new(Duration::from_secs(30)))
```
fulmicoton, line 949:
```suggestion
                            .stack_layer(GRPC_TIMEOUT)
```

fulmicoton (approved):
see minor comments

PR #5403, opened 2024-09-09
Fix configuration interpolation

rdettai:
### Description

The regex for env variable interpolation in configs doesn't capture properly multiple variables in the same line.

Closes https://github.com/quickwit-oss/quickwit/pull/5403

### How was this PR tested?

Unit test + tests in the lambda package


quickwit/quickwit-config/src/templating.rs
```diff
@@ -30,7 +30,7 @@ use tracing::debug;
 // `ENV_VAR` or `ENV_VAR:DEFAULT`
 // Ignores whitespaces in curly braces
 static TEMPLATE_ENV_VAR_CAPTURE: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(r"\$\{\s*([A-Za-z0-9_]+)\s*(?::\-\s*([\S]+)\s*)?}")
```
rdettai, line 33: this is the only code change of this PR, the rest are tests

PR #5404, opened 2024-09-09
batch delete from gc

trinity-1686a:
### Description

follow up to #5380
if a cluster generate less than 1k split per 10 minutes, everything was fine, but in case it would generate more than that, we have a slow down for 2 reasons:
- we no longer run 10 tasks in parallel
- we may get a batch of 1k splits from many different indexes, and get a 2nd batch from the same set of indexes, meaning more than one delete per index, despite having few splits to delete for each indexes. This can in theory cause us to do up to n_index time more delete calls to the metastore than necessary (caped to 1k)

Changes made:
- we run up to 10 deletion from storage + metastore concurrently
- we retrieve splits sorted by index_uid from metastore, so we don't do many delete per index
- we query for 10k splits instead of 1k: before we would query for 10 * 1k, but as this is now batched, we can increase this number to keep a similar effect to before

### How was this PR tested?

tested on a small cluster to not be utterly broken, for perf improvement, we'll need a way bigger cluster

### DIFF
--- a/quickwit/quickwit-index-management/src/garbage_collection.rs
+++ b/quickwit/quickwit-index-management/src/garbage_collection.rs
@@ -41,7 +41,7 @@ use time::OffsetDateTime;
 use tracing::{error, instrument};
 
 /// The maximum number of splits that the GC should delete per attempt.
-const DELETE_SPLITS_BATCH_SIZE: usize = 1000;
+const DELETE_SPLITS_BATCH_SIZE: usize = 10_000;
 
 /// [`DeleteSplitsError`] describes the errors that occurred during the deletion of splits from
 /// storage and metastore.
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-metastore/src/metastore/postgres/utils.rs
```diff
@@ -179,6 +179,39 @@ pub(super) fn append_query_filters(sql: &mut SelectStatement, query: &ListSplits
...
+            // truncate.
+            //
+            // Or we could just add a btree index to splits.index_uid. That might be the better
+            // long term solution.
+            sql.join(
+                JoinType::Join,
+                Indexes::Table,
+                Expr::col((Splits::Table, Splits::IndexUid))
+                    .equals((Indexes::Table, Indexes::IndexUid)),
+            )
```
trinity-1686a, line 209: note: i think this is a bad idea to do this *long term*, but i'd rather this fix be made without a migration for now, so we can easily revert clusters affected to previous versions if we find things are somehow still not good. If you disagree, we can either create the required index manually, or make a migration

quickwit/quickwit-index-management/src/garbage_collection.rs
```diff
@@ -182,12 +182,12 @@ async fn delete_splits_marked_for_deletion(
     let mut removed_splits = Vec::new();
     let mut failed_splits = Vec::new();
 
-    'outer: loop {
-        let mut exit = false;
+    loop {
         let query = ListSplitsQuery::try_from_index_uids(index_uids.clone())?
```
trinity-1686a, line 186: todo this is the only Err code path, and it shouldn't ever happen, we should just return an empty success as there was nothing to do

fulmicoton (approved):

PR #5421, opened 2024-09-13
allow specifying index in otel http protobuf api

### LINKED ISSUE #5413: Can not specify an index id on the http otel api
I am building a Java service to index Opentelemetry Logs into quickwit. The Java service is receiving logs from the OTel collector. (We need that extra Java layer to add business logic to it)

I use HttpUrlConnection to send Protobuf over HTTP. 
I found via the discord that the endpoint to use was : `/api/v1/otlp/v1/logs` 
Unfortunately, the logs are trying to be indexed in the default otel index: `otel-logs-v0_7`

When setting the header `qw-otel-logs-index:my_custom_index`, quickwit should use my_custom_index when sending logs with protobuf over http

I try to define a custom index using `qw-otel-logs-index` but to no avail.

trinity-1686a:
### Description

fix #5413

### How was this PR tested?

updated tests

### DIFF
--- a/quickwit/quickwit-serve/src/otlp_api/rest_handler.rs
+++ b/quickwit/quickwit-serve/src/otlp_api/rest_handler.rs
@@ -270,11 +270,40 @@ mod tests {
         let mut mock_ingest_router = MockIngestRouterService::new();
         mock_ingest_router
             .expect_ingest()
+            .times(2)
             .withf(|request| {
-                request.subrequests.len() == 1
-                    && request.subrequests[0].doc_batch.is_some()
+                if request.subrequests.len() == 1 {
+                    let subrequest = &request.subrequests[0];
+                    subrequest.doc_batch.is_some()
                     // && request.commit == CommitType::Auto as i32
-                    && request.subrequests[0].doc_batch.as_ref().unwrap().doc_lengths.len() == 1
+                    && subrequest.doc_batch.as_ref().unwrap().doc_lengths.len() == 1
+                    && subrequest.index_id == quickwit_opentelemetry::otlp::OTEL_LOGS_INDEX_ID
+                } else {
+                    false
+                }
+            })
+            .returning(|_| {
+                Ok(IngestResponseV2 {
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-serve/src/otlp_api/rest_handler.rs
```diff
@@ -126,7 +141,16 @@ pub(crate) fn otlp_default_traces_handler(
         .and(with_arg(BodyFormat::default()))
         .map(into_rest_api_response)
 }
-
+/// Open Telemetry REST/Protobuf traces ingest endpoint.
+#[utoipa::path(
+    post,
+    tag = "Open Telemetry",
+    path = "/{index}/otlp/v1/traces",
```
rdettai, line 148: I'm a bit surprised that we use a path param here and not a header (also what was suggested in https://github.com/quickwit-oss/quickwit/issues/5413). It would be more in line with the way it works in GRPC. But of course in GRPC you don't have a path, so the comparison doesn't really hold. I guess both are fine!

rdettai (approved):
An entry in the docs (https://quickwit.io/docs/main-branch/log-management/otel-service and https://quickwit.io/docs/main-branch/distributed-tracing/otel-service) would be cool! I remember struggling to figure out the exact endpoint when setting up OTEL ingestion with Grafana Tempo :smile:

quickwit/quickwit-serve/src/otlp_api/rest_handler.rs
trinity-1686a: That api endpoint was already there (but didn't work properly and was absent from openapi definiton). @fmassot, looks like that code was yours, what do you think (index in path vs in header)?

guilload: Customizing the endpoint or adding headers with the OTEL collectors is easy, so let's support both, but this PR is great for a first pass. @trinity-1686a, please go ahead and merge this.

PR #5425, opened 2024-09-13
Fix ancient test that recently re-surfaced

rdettai:
### Description

Kind of like the balrog that was awaken by the dwarfes mining mithril, this test was re-activated in CI by https://github.com/quickwit-oss/quickwit/pull/5392.

The default sort order was inverted in https://github.com/quickwit-oss/quickwit/pull/3726/files and this dormant test was not updated accordingly.

### How was this PR tested?

Test now passes.

### DIFF
--- a/docs/internals/ingest-v2.md
+++ b/docs/internals/ingest-v2.md
@@ -18,7 +18,3 @@ indexer:
 See [full configuration example](https://github.com/quickwit-oss/quickwit/blob/main/config/quickwit.yaml).
 
 The only way to use the ingest API V2 is to use the [bulk endpoint](../reference/es_compatible_api.md#_bulk--batch-ingestion-endpoint) of the Elasticsearch-compatible API. The native Quickwit API is not yet compatible with the ingest V2 API.
-
-## Caveats
-
-The `refresh` parameter is not yet supported on the ingest V2 API.
--- a/docs/overview/concepts/querying.md
+++ b/docs/overview/concepts/querying.md
@@ -108,4 +108,4 @@ Quickwit supports sorting docs by their BM25 scores. In order to query by score,
 
 ### Document ID
 
-Each document in Quickwit is assigned a unique document ID which is a combination of the split ID and the Tantivy DocId within the split. You cannot assign a custom ID. This ID is used for every search query as sort order (after the explicitly specified sort values) to make the results deterministic.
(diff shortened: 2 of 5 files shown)


docs/internals/ingest-v2.md
```diff
@@ -18,7 +18,3 @@ indexer:
 See [full configuration example](https://github.com/quickwit-oss/quickwit/blob/main/config/quickwit.yaml).
 
 The only way to use the ingest API V2 is to use the [bulk endpoint](../reference/es_compatible_api.md#_bulk--batch-ingestion-endpoint) of the Elasticsearch-compatible API. The native Quickwit API is not yet compatible with the ingest V2 API.
-
-## Caveats
-
-The `refresh` parameter is not yet supported on the ingest V2 API.
```
rdettai, line 24: This has nothing to do with this PR but was left out of https://github.com/quickwit-oss/quickwit/pull/5350

docs/overview/concepts/querying.md
```diff
@@ -105,3 +105,7 @@ Quickwit does caching in many places to deliver a highly performing query engine
 ### Scoring
 
 Quickwit supports sorting docs by their BM25 scores. In order to query by score, [fieldnorms](../../configuration/index-config.md#text-type) must be enabled for the field. By default, BM25 scoring is disabled to improve query latencies but it can be opt-in by setting the `sort_by` option to `_score` in queries.
+
+### Document ID
+
+Each document in Quickwit is assigned a unique document ID which is a combination of the split ID and the Tantivy DocId within the split. You cannot assign a custom ID. This ID is used for every search query as sort order (after the explicitly specified sort values) to make the results deterministic.
```
rdettai, line 111: I think it is worth defining the doc id here, especially for seasoned ES users who might be looking into how this works in Quickwit. I'm open to suggestions on the level of details we should go into, as this is _mostly_ an internal concept.

trinity-1686a: i guess you could add that this id is not stable across merges, current wording might suggest it is (i don't think it's necessarily obvious that split id changes after merges).
I don't know if saying it's a combination of split id and tantivy docid is actually useful

trinity-1686a (approved):

rdettai:
https://github.com/quickwit-oss/quickwit/actions/runs/10852041432

docs/overview/concepts/querying.md
rdettai: > I don't know if saying it's a combination of split id and tantivy docid is actually useful

yeah it's kind of an implementation detail, but I think it helps figuring out how they behave

PR #5427, opened 2024-09-13
Downgrade actions/download-artifact from 4.1.8 to 4.1.7

rdettai:
### Description

Looks like the bump from 4.1.7 to 4.1.8 (https://github.com/quickwit-oss/quickwit/pull/5410) broke the docker CI (https://github.com/quickwit-oss/quickwit/actions/workflows/publish_docker_images.yml). I didn't investigate why.

### How was this PR tested?

Test now passes.

### DIFF
--- a/.github/workflows/publish_docker_images.yml
+++ b/.github/workflows/publish_docker_images.yml
@@ -91,7 +91,7 @@ jobs:
     needs: [docker]
     steps:
       - name: Download digests
-        uses: actions/download-artifact@v4.1.8
+        uses: actions/download-artifact@v4.1.7
         with:
           name: digest
           path: /tmp/digests


trinity-1686a:
the pipeline on b61ba35253509a6a5bd2875313b12449b36118bb failed too, that was just after 3->4.1.7, but before 4.1.7->4.1.8 , so I don't think this actually fixes the issue.
I see a `uses: actions/upload-artifact@v3` a few line before. I'm not familiar with actions, but i would guess a missmatch of major version between upload-artifact and download-artifact could easily break things

rdettai:
You are right, I didn't notice there was a bump 3->4 right before! Indeed, uploads from 3 are not compatible with dowloads from 4 (https://github.com/actions/download-artifact?tab=readme-ov-file#breaking-changes)

rdettai:
Superseeded by https://github.com/quickwit-oss/quickwit/pull/5428

PR #5428, opened 2024-09-16
Upgrade actions/upload-artifact to match actions/download-artifact

rdettai:
### Description

Looks like the bump from actions/download-artifact 3 to 4 broke the docker CI (https://github.com/quickwit-oss/quickwit/actions/workflows/publish_docker_images.yml). This is an expected breaking change (https://github.com/actions/download-artifact?tab=readme-ov-file#breaking-changes))

### How was this PR tested?

### DIFF
--- a/.github/workflows/publish_docker_images.yml
+++ b/.github/workflows/publish_docker_images.yml
@@ -95,6 +95,7 @@ jobs:
         with:
           pattern: digest-*
           path: /tmp/digests
+          merge-multiple: true
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
--- a/.github/workflows/publish_docker_images.yml
+++ b/.github/workflows/publish_docker_images.yml
@@ -79,7 +79,7 @@ jobs:
           touch "/tmp/digests/${digest#sha256:}"
 
       - name: Upload digest
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4.4.0
         with:
           name: digest
           path: /tmp/digests/*
--- a/.github/workflows/publish_docker_images.yml
+++ b/.github/workflows/publish_docker_images.yml
@@ -81,7 +81,7 @@ jobs:
       - name: Upload digest
         uses: actions/upload-artifact@v4.4.0
         with:
-          name: digest
+          name: digest-${{ matrix.platform }}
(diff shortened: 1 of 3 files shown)


rdettai (commented):
In the [breaking change] section of this new version there is the following item:

> Uploading to the same named Artifact multiple times. Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once. Otherwise you will encounter an error.

I think this will likely be a problem as we repeatedly use the same artifact names. But it's not completely clear from the docs whether this restriction applies only within a job run or in general.

trinity-1686a:
https://github.com/quickwit-oss/quickwit/actions/runs/10879677075

trinity-1686a:
> Run actions/upload-artifact@v4.4.0
With the provided path, there will be 1 file uploaded
Artifact name is valid!
Root directory input is valid!
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run

.github/workflows/publish_docker_images.yml
```diff
@@ -79,7 +79,7 @@ jobs:
           touch "/tmp/digests/${digest#sha256:}"
 
       - name: Upload digest
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4.4.0
         with:
           name: digest
```
trinity-1686a, line 84: looks like we hit the breaking change you mentioned 
```suggestion
          name: digest-${{ matrix.platform }}
```
and line 96
```diff
-          name: digest
+          name: digest-*
```

trinity-1686a (changes requested):

rdettai:
https://github.com/quickwit-oss/quickwit/actions/runs/10881279988

rdettai:
https://github.com/quickwit-oss/quickwit/actions/runs/10881445673

rdettai:
https://github.com/quickwit-oss/quickwit/actions/runs/10883228939

trinity-1686a (approved):
> ERROR: can't push with no tags specified, please set --tag or --dry-run

looks like the kind of error we'd expect from running that way, we can ignore it

PR #5434, opened 2024-09-17
Fix bulk api response order

rdettai:
### Description

The bulk API should have its responses in the same order as the submitted actions (https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-api-response-body)

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-serve/src/elasticsearch_api/bulk_v2.rs
+++ b/quickwit/quickwit-serve/src/elasticsearch_api/bulk_v2.rs
@@ -77,6 +77,7 @@ type ElasticDocId = String;
 
 #[derive(Debug)]
 struct DocHandle {
+    doc_position: usize,
     doc_uid: DocUid,
     es_doc_id: Option<ElasticDocId>,
     // Whether the document failed to parse. When the struct is instantiated, this value is set to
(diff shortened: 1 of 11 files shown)


quickwit/quickwit-serve/src/elasticsearch_api/bulk_v2.rs
```diff
@@ -293,13 +300,25 @@ fn make_elastic_bulk_response_v2(
...
         per_subrequest_doc_handles.is_empty(),
         "doc handles should be empty"
     );
+
+    assert_eq!(
+        positioned_actions.len(),
+        action_count,
+        "request and response action count should match"
+    );
+    positioned_actions.sort_unstable_by_key(|(idx, _)| *idx);
```
trinity-1686a, line 316: instead of sorting, we could have an array of `Option<Action>` where we write the result at the correct position directly. that would be faster, I'm not sure how much we care though, here is probably not performance critical

rdettai: Yes, I went that way originally, but the code was a bit confused and as you say, I doubt the performance gain is significative.

Issue #5437, opened 2024-09-18
Missing index on indexes table to support wildcard expansion


esatterwhite:
**Describe the bug**
The query generated for index name wildcard (`*`) expansion doesn't have an index suitable to satisfy the query resulting in a lot of full table scans. In environments with a high search volume over a large number of indexes, this can be a bit of a bottleneck

```sql
SELECT * FROM indexes
WHERE (index_id LIKE $1) -- where $1 would be something like logline.abc123.2024-01-01*
```

The only index for `index_id` is a unique constraint, which uses the default btree index. BTREE by default cannot satisfy `LIKE` only `=`, `>`, `>=` , `<`, `<=`, and `<>`

```sql
ALTER TABLE indexes ADD CONSTRAINT indexes_index_id_unique UNIQUE (index_id);
```

**Steps to reproduce (if applicable)**

1. Insert 50K records into the indexes table with non common index_id prefix value
2. insert 3 records in the indexes table with a common index_id prefix value
3. Explain a query using the `WHERE index_id LIKE <prefix>*`
4. Observe the full table scan

**Expected behavior**
The operation of expanding an index name wildcard needs to be rather fast as to not be a bottle neck in a search query. There should be an index, that satisifes the `LIKE` clause. This can be a GIST index using trigrams which will support wildcards in multiple positions in the index id

```sql
CREATE EXTENSION pg_trgm;
CREATE INDEX CONCURRENTLY index_id_gist_idx
    ON public.indexes USING gist
    (index_id gist_trgm_ops)
    WITH (buffering=auto);
```

or if a wildcard in the terminal position is the only concern, `varchar_pattern_ops` can be used with a regular btree index

```sql
ALTER TABLE indexes
DROP CONSTRAINT indexes_index_id_unique on indexes;
CREATE UNIQUE INDEX indexes_index_id_unique
  ON indexes USING btree ("index_id" varchar_pattern_ops);
```
**Build Info**

```javascript
{
  "build": {
    "build_date": "2024-09-10T15:37:54Z",
    "build_profile": "release",
    "build_target": "aarch64-unknown-linux-gnu",
    "cargo_pkg_version": "0.8.0",
    "commit_date": "2024-09-10T15:16:01Z",
    "commit_hash": "9e02f6c145c5aee6be5697105cd5a8cf86c40f46",
    "commit_short_hash": "9e02f6c",
    "commit_tags": [
      "qw-airmail-20240910"
    ],
[...]

PR #5438, opened 2024-09-18
fix(metastore): add index to support index name wildcard expansion

esatterwhite:
### Description
The sql that is generated to expand an index name (index_id) into all of the matching index names uses the LIKE clause to do partial text matches. However, the only index on the index_id column utilizes the default btree configuration which can only do exact value comparisons. In environments with a large search volume over many thousands of indexes, this can be a bottle neck as the query will always do an full table scan. This drops and recreates the unique index on index_id to include the `varchar_pattern_ops` which allows for wildcard matching on a varchar column using the LIKE operator

Fixes: #5437

### How was this PR tested?
* Quickwit release binary built
* Docker docker compose started
* started quickwit metastore `quickwit run --service metastore
* Verify the unique index of the same name created with the `varchar_pattern_ops` operator class exists on `index_id`
* Insert a number of records into the table, and confirm a `LIKE` clause w/ a terminal wildcard produces an index scan.

```txl
\d+ indexes
                                                                   Table "public.indexes"
       Column        |            Type             | Collation | Nullable |                 Default                  | Storage  | Stats target | Description 
---------------------+-----------------------------+-----------+----------+------------------------------------------+----------+--------------+-------------
 index_uid           | character varying(282)      |           | not null |                                          | extended |              | 
 index_metadata_json | text                        |           | not null |                                          | extended |              | 
 create_timestamp    | timestamp without time zone |           | not null | timezone('UTC'::text, CURRENT_TIMESTAMP) | plain    |              | 
 index_id            | character varying(255)      |           | not null | ''::character varying                    | extended |              | 
Indexes:
    "indexes_pkey" PRIMARY KEY, btree (index_uid)
    "indexes_index_id_unique" UNIQUE, btree (index_id varchar_pattern_ops)
Referenced by:
    TABLE "delete_tasks" CONSTRAINT "delete_tasks_index_uid_fkey" FOREIGN KEY (index_uid) REFERENCES indexes(index_uid) ON DELETE CASCADE
    TABLE "shards" CONSTRAINT "shards_index_uid_fkey" FOREIGN KEY (index_uid) REFERENCES indexes(index_uid) ON DELETE CASCADE
    TABLE "splits" CONSTRAINT "splits_index_uid_fkey" FOREIGN KEY (index_uid) REFERENCES indexes(index_uid) ON DELETE CASCADE
```

```txl
EXPLAIN(ANALYZE, BUFFERS, COSTS, TIMING) SELECT * FROM indexes WHERE (index_id LIKE 'logline.d1f92cd2db.2024-08-18%');

                                                                   QUERY PLAN                                                                   
------------------------------------------------------------------------------------------------------------------------------------------------
 Index Scan using indexes_index_id_unique_bis on indexes  (cost=0.41..8.43 rows=3 width=1945) (actual time=0.012..0.012 rows=1 loops=1)
   Index Cond: (((index_id)::text ~>=~ 'logline.d1f92cd2db.2024-08-18'::text) AND ((index_id)::text ~<~ 'logline.d1f92cd2db.2024-08-19'::text))
   Filter: ((index_id)::text ~~ 'logline.d1f92cd2db.2024-08-18%'::text)
   Buffers: shared hit=4
 Planning:
   Buffers: shared hit=41
 Planning Time: 0.214 ms
 Execution Time: 0.025 ms
```

### DIFF
--- a/quickwit/quickwit-metastore/migrations/postgresql/23_change-indexes-unique-index.down.sql
+++ b/quickwit/quickwit-metastore/migrations/postgresql/23_change-indexes-unique-index.down.sql
@@ -1,0 +1,2 @@
+DROP INDEX IF EXISTS indexes_index_id_unique;
+ALTER TABLE indexes ADD CONSTRAINT indexes_index_id_unique UNIQUE (index_id);
--- a/quickwit/quickwit-metastore/migrations/postgresql/23_change-indexes-unique-index.up.sql
+++ b/quickwit/quickwit-metastore/migrations/postgresql/23_change-indexes-unique-index.up.sql
@@ -1,0 +1,4 @@
+ALTER TABLE indexes DROP CONSTRAINT IF EXISTS indexes_index_id_unique;
+
+CREATE UNIQUE INDEX IF NOT EXISTS indexes_index_id_unique
+  ON indexes USING btree ("index_id" varchar_pattern_ops);


quickwit/quickwit-metastore/migrations/postgresql/23_change-indexes-unique-index.down.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX IF EXISTS IF EXISTS indexes_index_id_unique;
```
trinity-1686a, line 1: `IF EXISTS` is repeated 2 times

esatterwhite: Fixed

trinity-1686a (approved):

Issue #5443, opened 2024-09-24
Additional Search Related Metrics


esatterwhite:
**Is your feature request related to a problem? Please describe.**
We have been experiencing periodic slowness and spikes in resource usage on search node, and We can't seem to figure out why.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
Add any other context or information about the feature request here.

PR #5447, opened 2024-09-24
Add some additional search metrics

guilload:
### Description
Add some basic search metrics. Will add some more in follow up PR.

### How was this PR tested?
`make test-all`

### DIFF
--- a/quickwit/quickwit-search/src/metrics.rs
+++ b/quickwit/quickwit-search/src/metrics.rs
@@ -21,11 +21,15 @@
 
 use once_cell::sync::Lazy;
 use quickwit_common::metrics::{
-    exponential_buckets, new_counter, new_counter_vec, new_histogram, Histogram, IntCounter,
-    IntCounterVec,
+    exponential_buckets, new_counter, new_counter_vec, new_histogram, new_histogram_vec, Histogram,
+    HistogramVec, IntCounter, IntCounterVec,
 };
 
 pub struct SearchMetrics {
+    pub root_search_requests_total: IntCounterVec<1>,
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-search/src/metrics.rs
```diff
@@ -34,6 +38,36 @@ pub struct SearchMetrics {
...
+                &[("kind", "server")],
+                ["status"],
+            ),
+            leaf_search_request_duration_seconds: new_histogram_vec(
+                "leaf_search_request_duration_seconds",
+                "Duration of request in seconds.",
+                "search",
+                &[("kind", "server")],
+                ["status"],
+                exponential_buckets(0.001, 2.0, 12).unwrap(),
```
trinity-1686a, line 69: line 82 (for leaf_search_split_duration_secs), the parameters are  `(0.005, 2.0, 10)`. We should use the same parameters (no big opinion on which is better, but having 5ms precision on split, and 1ms on leaf search seems the wrong way around)

trinity-1686a (approved):

quickwit/quickwit-search/src/metrics.rs
trinity-1686a: on second though, the number of buckets should probably be higher. https://github.com/quickwit-oss/quickwit/issues/5443 requests bucket up to 30s, so `(0.005, 2.0, 13)` or `(0.001, 2.0, 15)`

guilload: Thanks for spotting this. I used `(0.001, 2.0, 15)` for both RPCs.

Issue #5460, opened 2024-09-28
ES API date format `strict_date_optional_time` is not supported


kuzaxak:
**Describe the bug**

After PR https://github.com/quickwit-oss/quickwit/pull/5208 ES API responds with error when OSD sends a request like:

```
"query": {
  "bool": {
    "must": [],
    "filter": [
      {
        "match_all": {}
      },
      {
        "range": {
          "time": {
            "gte": "2024-09-28T10:07:55.797Z",
            "lte": "2024-09-28T10:22:55.797Z",
            "format": "strict_date_optional_time"
          }
        }
      }
    ],
    "should": [],
    "must_not": []
  }
}
```

to `_search` endpoint returns an error:

```
{
    "status": 400,
    "error": {
        "caused_by": null,
        "reason": "Failed to parse date time: a character literal was not valid",
        "stack_trace": null,
        "type": null
    }
}
```

**Expected behavior**

[Formats from ES](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html#built-in-date-formats) should be supported.

**Configuration:**
Please provide:

1. Output of `quickwit --version`
`Quickwit 0.8.0 (aarch64-unknown-linux-gnu 2024-09-20T09:00:32Z 634a1bc)`
2. The index_config.yaml
Default OTEL traces index.

kuzaxak:
It looks like a duplicate for already reported:
https://github.com/quickwit-oss/quickwit/issues/5228
https://github.com/quickwit-oss/quickwit/issues/5109

And seems it will be fixed in [PR](https://github.com/quickwit-oss/quickwit/pull/5248)

kuzaxak:
Implemented in https://github.com/quickwit-oss/quickwit/pull/5462

PR #5248, opened 2024-07-23
fix(es): add java date format support

etolbakov:
### Description
add conversion from elasticsearch/opensearch date format to quickwit expected strptime

### How was this PR tested?
Provide curl output below 

### Related Issue
https://github.com/quickwit-oss/quickwit/issues/5228

The code is take from https://github.com/quickwit-oss/quickwit/pull/5237

PR #5461, opened 2024-09-28
Add ES API compatible date time parsers

kuzaxak:
### Description

ES uses [own names][1] for date formats. I implemented the ones that I was able to find in OSD.

Test cases taken from [doc][1] and [OS test][2].

This PR is fixing [issue](https://github.com/quickwit-oss/quickwit/issues/5460) and restore compatibility with OSD.

[1]: https://opensearch.org/docs/latest/field-types/supported-field-types/date/
[2]: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/test/java/org/opensearch/common/time/DateFormattersTests.java

### How was this PR tested?

Added unit tests to cover new parsers.

### DIFF
--- a/quickwit/quickwit-datetime/src/date_time_format.rs
+++ b/quickwit/quickwit-datetime/src/date_time_format.rs
@@ -162,6 +162,8 @@ pub enum DateTimeInputFormat {
     Rfc3339,
     Strptime(StrptimeParser),
     Timestamp,
+    ESStrictDateOptionalTime,
+    ESStrictDateTimeNoMillis,
 }
 
 impl DateTimeInputFormat {
@@ -172,6 +174,8 @@ impl DateTimeInputFormat {
             DateTimeInputFormat::Rfc3339 => "rfc3339",
             DateTimeInputFormat::Strptime(parser) => parser.borrow_strptime_format(),
             DateTimeInputFormat::Timestamp => "unix_timestamp",
+            DateTimeInputFormat::ESStrictDateOptionalTime => "strict_date_optional_time",
+            DateTimeInputFormat::ESStrictDateTimeNoMillis => "strict_date_time_no_millis",
         }
     }
 }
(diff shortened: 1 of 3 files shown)


etolbakov:
@kuzaxak you might be interested in checking this related PR. As far as I understand Paul’s idea was to enable the support using a regexset 
https://github.com/quickwit-oss/quickwit/pull/5248

Maybe we can finalize it 
Cc @fulmicoton

kuzaxak:
> @kuzaxak you might be interested in checking this related PR. As far as I understand Paul’s idea was to enable the support using a regexset #5248
> 
> Maybe we can finalize it Cc @fulmicoton

Okay, I missed it. Do you have plans to finish it? Or maybe I can help you?
As of now, this issue is blocking onboarding QW as OpenSearch Dashboards constantly send `strict_date_optional_time` in a format field. 

I can patch requests on the fly if implementation takes longer than a week.

@etolbakov

kuzaxak:
Closing as https://github.com/quickwit-oss/quickwit/pull/5248 should be a generic solution instead of a partial one presented in this PR.

etolbakov:
I was curious to work on that just for sake of learning, so don’t have any sticky timelines.
IIRC the latest status for that ticket: I was a bit stuck and needed Paul’s input. 

If you can proceed/ take a lead on this PR I think it will be great!

kuzaxak:
> If you can proceed/ take a lead on this PR I think it will be great!

I can. What is left to be done there? 

I see that you added some commented-out tests. Does it mean we need to implement Logix to make them work?

And I think `strict_date_optional_time` should be extended to accept only year as required param.

@etolbakov

etolbakov:
I added a bit more tests for `week` patterns which should be ok. 
The commented failing tests are a question mark - do we need them at all (I mean, if Quickwit team would like to support the functionality)
My initial idea was to port all "format"-related tests from here:
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/test/java/org/opensearch/common/joda/JavaJodaTimeDuellingTests.java#L260
but maybe it’s an overkill.

I also see that "cargo nexttest" step is failing.
However, local execution of "clippy" is working fine, haven’t solved that.
@kuzaxak

PR #5467, opened 2024-10-01
Introduce a searcher config option to timeout get requests.

### LINKED ISSUE #5466: Retry S3 GET request more agressively.
As spotted on airmail...

Some requests are strongly affected by lagging GET requests observed on a few nodes.

AWS recommends agressively retrying requests that are taking too long.
https://docs.aws.amazon.com/whitepapers/latest/s3-optimizing-performance-best-practices/timeouts-and-retries-for-latency-sensitive-applications.html

The best layer on which to insert ourselves is probably in the warmup IO onion

fulmicoton:
This is done at a storage instead of the s3 client as the s3 client only had a global operation timeout parameter.

Here the timeout is proportional to the length (in byte) of the slice being downloaded.

By default, if the searcher config does not define a timeout explicitly, this PR is no-op.

Closes #5466

### DIFF
--- a/quickwit/quickwit-config/src/node_config/mod.rs
+++ b/quickwit/quickwit-config/src/node_config/mod.rs
@@ -243,7 +243,9 @@ pub struct SearcherConfig {
 pub struct StorageTimeoutPolicy {
     pub min_throughtput_bytes_per_secs: u64,
     pub timeout_millis: u64,
-    pub repeat: usize,
+    // Disclaimer: this is a number of retry, so the overall max number of
+    // attempts is `max_num_retries + 1``.
+    pub max_num_retries: usize,
 }
 
 impl StorageTimeoutPolicy {
@@ -255,7 +257,7 @@ impl StorageTimeoutPolicy {
         };
         let timeout = Duration::from_millis(self.timeout_millis)
             + Duration::from_secs_f64(min_download_time_secs);
-        std::iter::repeat(timeout).take(self.repeat)
+        std::iter::repeat(timeout).take(self.max_num_retries + 1)
     }
 }
 
--- a/quickwit/quickwit-config/src/node_config/mod.rs
+++ b/quickwit/quickwit-config/src/node_config/mod.rs
@@ -223,6 +223,40 @@ pub struct SearcherConfig {
     pub split_cache: Option<SplitCacheLimits>,
     #[serde(default = "SearcherConfig::default_request_timeout_secs")]
     request_timeout_secs: NonZeroU64,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub storage_timeout_policy: Option<StorageTimeoutPolicy>,
+}
+
+/// Configuration controlling how fast a searcher should timeout a `get_slice`
+/// request to retry it.
+///
+/// [Amazon's best practise](https://docs.aws.amazon.com/whitepapers/latest/s3-optimizing-performance-best-practices/timeouts-and-retries-for-latency-sensitive-applications.html)
+/// suggests that to ensure low latency, it is best to:
+/// - retry small GET request after 2s
+/// - retry large GET request when the throughput is below some percentile.
+///
+/// This policy is inspired by this guidance. It does not track instanteneous throughput, but
+/// computes an overall timeout using the following formula:
+/// `timeout_offset + num_bytes_get_request / min_throughtput`
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct StorageTimeoutPolicy {
+    pub min_throughtput_bytes_per_secs: u64,
+    pub timeout_millis: u64,
+    pub repeat: usize,
+}
+
+impl StorageTimeoutPolicy {
+    pub fn compute_timeout(&self, num_bytes: usize) -> impl Iterator<Item = Duration> {
+        let min_download_time_secs: f64 = if self.min_throughtput_bytes_per_secs == 0 {
+            0.0f64
+        } else {
+            num_bytes as f64 / self.min_throughtput_bytes_per_secs as f64
+        };
+        let timeout = Duration::from_millis(self.timeout_millis)
+            + Duration::from_secs_f64(min_download_time_secs);
+        std::iter::repeat(timeout).take(self.repeat)
+    }
 }
 
 impl Default for SearcherConfig {
@@ -237,6 +271,7 @@ impl Default for SearcherConfig {
             aggregation_bucket_limit: 65000,
             split_cache: None,
             request_timeout_secs: Self::default_request_timeout_secs(),
+            storage_timeout_policy: None,
         }
     }
 }
--- a/quickwit/quickwit-config/src/node_config/serialize.rs
+++ b/quickwit/quickwit-config/src/node_config/serialize.rs
@@ -615,7 +615,7 @@ mod tests {
                 storage_timeout_policy: Some(crate::StorageTimeoutPolicy {
                     min_throughtput_bytes_per_secs: 100_000,
                     timeout_millis: 2_000,
-                    repeat: 2
+                    max_num_retries: 2
                 })
             }
         );
--- a/quickwit/quickwit-config/src/node_config/serialize.rs
+++ b/quickwit/quickwit-config/src/node_config/serialize.rs
@@ -612,6 +612,11 @@ mod tests {
                 max_num_concurrent_split_streams: 120,
                 split_cache: None,
                 request_timeout_secs: NonZeroU64::new(30).unwrap(),
+                storage_timeout_policy: Some(crate::StorageTimeoutPolicy {
+                    min_throughtput_bytes_per_secs: 100_000,
+                    timeout_millis: 2_000,
+                    repeat: 2
+                })
             }
         );
         assert_eq!(
--- a/quickwit/quickwit-search/src/leaf.rs
+++ b/quickwit/quickwit-search/src/leaf.rs
@@ -36,7 +36,7 @@ use quickwit_query::query_ast::{BoolQuery, QueryAst, QueryAstTransformer, RangeQ
 use quickwit_query::tokenizers::TokenizerManager;
 use quickwit_storage::{
     wrap_storage_with_cache, BundleStorage, MemorySizedCache, OwnedBytes, SplitCache, Storage,
-    StorageResolver,
+    StorageResolver, TimeoutAndRetryStorage,
 };
 use tantivy::aggregation::agg_req::{AggregationVariants, Aggregations};
 use tantivy::aggregation::AggregationLimitsGuard;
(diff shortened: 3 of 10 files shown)


quickwit/quickwit-storage/src/timeout_and_retry_storage.rs
```diff
@@ -0,0 +1,270 @@
...
+        {
+            let storage_with_delay =
+                StorageWithDelay::new(vec![Duration::from_secs(5), Duration::from_secs(3)]);
+            let storage =
+                TimeoutAndRetryStorage::new(Arc::new(storage_with_delay), timeout_policy.clone());
+            assert_eq!(
+                storage.get_slice(path, 10..100).await.unwrap_err().kind,
+                StorageErrorKind::Timeout
+            );
+        }
```
trinity-1686a, line 262: it could be interesting to assert on the time of the simulation

quickwit/quickwit-storage/src/timeout_and_retry_storage.rs
```diff
@@ -0,0 +1,270 @@
...
+        {
+            let get_slice_fut = self.underlying.get_slice(path, range.clone());
+            match tokio::time::timeout(timeout_duration, get_slice_fut).await {
+                Ok(result) => return result,
+                Err(_elapsed) => {
+                    if let Some(get_slice_timeout_count) = crate::STORAGE_METRICS
+                        .get_slice_timeout_total_by_attempts
+                        .get(attempt_id)
+                    {
+                        get_slice_timeout_count.inc();
```
trinity-1686a, line 108: > Number of `get_slice` operations that timed out

the update of metric feels odd. Nothing is incremented if we succeed on first try (it feels like "0" should be), if we fail the first try, and succeed at 2nd, "0" is incremented (despite there be a timeout), and "1" is not. If we time out on 3rd try, "0" and "1" are incremented, but "2" isn't.
There is an off by one, and I don't think the metric should be updated on the sad path, only on the happy path, or after the last try

quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -223,6 +223,40 @@ pub struct SearcherConfig {
...
+/// - retry large GET request when the throughput is below some percentile.
+///
+/// This policy is inspired by this guidance. It does not track instanteneous throughput, but
+/// computes an overall timeout using the following formula:
+/// `timeout_offset + num_bytes_get_request / min_throughtput`
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct StorageTimeoutPolicy {
+    pub min_throughtput_bytes_per_secs: u64,
+    pub timeout_millis: u64,
+    pub repeat: usize,
```
trinity-1686a, line 246: this should be validated to be non zero, a NonZeroUsize might do the trick

quickwit/quickwit-storage/src/timeout_and_retry_storage.rs
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    async fn copy_to_file(&self, path: &Path, output_path: &Path) -> StorageResult<u64> {
+        self.underlying.copy_to_file(path, output_path).await
+    }
+
+    /// Downloads a slice of a file from the storage, and returns an in memory buffer
+    async fn get_slice(&self, path: &Path, range: Range<usize>) -> StorageResult<OwnedBytes> {
+        let num_bytes = range.len();
+        for (attempt_id, timeout_duration) in self
```
esatterwhite, line 95: One concern I would have here is that this will likely chew into rate limits much faster.
I think we're pretty well with in the 5500/s range. But we do see spikes / surges from the searcher upwards of 3K+/s currently.

Are storage failures recorded in a metric or logged so that we can identify failures / rate limits?

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -134,13 +134,34 @@ pub(crate) async fn open_index_with_caches(
     tokenizer_manager: Option<&TokenizerManager>,
     ephemeral_unbounded_cache: bool,
 ) -> anyhow::Result<Index> {
-    let (hotcache_bytes, bundle_storage) =
-        open_split_bundle(searcher_context, index_storage, split_and_footer_offsets).await?;
+    // Let's a layer to retry `get_slice` requests if they are taking too long,
```
fulmicoton, line 137:
```suggestion
    // Let's add a storage proxy to retry `get_slice` requests if they are taking too long,
```

quickwit/quickwit-storage/src/timeout_and_retry_storage.rs
fulmicoton: yeah, it is counting timeouts and the position where they occur.
One benefit is that without filtering on the label, you get the total number of timeout.

I thought it was natural, but if you find it confusing, I can think of another solution.

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -612,6 +612,11 @@ mod tests {
                 max_num_concurrent_split_streams: 120,
                 split_cache: None,
                 request_timeout_secs: NonZeroU64::new(30).unwrap(),
+                storage_timeout_policy: Some(crate::StorageTimeoutPolicy {
```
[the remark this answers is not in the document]
fulmicoton: I had forgotten to commit `quickwit.{json,toml,yaml}`.

quickwit/quickwit-storage/src/timeout_and_retry_storage.rs
fulmicoton: The rate limiting is per prefix and overall.

We are working on improving your number of splits per day considerably.
This should improve your figure by a lot.

We were also planning to add a random prefix to split's ULID to avoid creating some hotspot... Unfortunately, your one-index-per-day scheme will make it useless.

quickwit/quickwit-storage/src/timeout_and_retry_storage.rs
```diff
@@ -0,0 +1,270 @@
...
+            .storage_timeout_policy
+            .compute_timeout(num_bytes)
+            .enumerate()
+        {
+            let get_slice_fut = self.underlying.get_slice(path, range.clone());
+            match tokio::time::timeout(timeout_duration, get_slice_fut).await {
```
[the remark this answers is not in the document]
fulmicoton: The code complexity is not too bad.

```
+        let mut futures_unordered = FuturesUnordered::new();
         let num_bytes = range.len();
+
         for (attempt_id, timeout_duration) in self
             .storage_timeout_policy
             .compute_timeout(num_bytes)
             .enumerate()
         {
             let get_slice_fut = self.underlying.get_slice(path, range.clone());
-            match tokio::time::timeout(timeout_duration, get_slice_fut).await {
-                Ok(result) => {
+            futures_unordered.push(get_slice_fut);
+            match tokio::time::timeout(timeout_duration, futures_unordered.next()).await {
+                Ok(Some(result)) => {
                     crate::STORAGE_METRICS
                         .get_slice_timeout_successes
                         .get(attempt_id)
@@ -110,6 +113,9 @@ impl Storage for TimeoutAndRetryStorage {
                         .inc();
                     return result;
                 }
+                Ok(None) => {
+                    // ..
+                }
                 Err(_elapsed) => {
                     rate_limited_info!(limit_per_min=60, num_bytes=num_bytes, path=%path.display(), timeout_secs=timeout_duration.as_secs_f32(), "get timeout elapsed");
                     continue;
(END)
```

but I am a tad scared of the extra code complexity involved by all solutions (shared tasks or independent tasks).

I'll add a todo to run proper tests.

quickwit/quickwit-config/src/node_config/mod.rs
fulmicoton: good point. I switched to `max_num_retries`

quickwit/quickwit-storage/src/timeout_and_retry_storage.rs
fulmicoton: `Does the quickwit_storage_object_storage_gets_total metric also get incremented on a retry?

yes

`Are storage failures recorded in a metric or logged so that we can identify failures / rate limits?`

Storage timeout have a metric yes. Rate limiting will show up in logs.

fulmicoton:
(Tested on simian)

quickwit/quickwit-storage/src/metrics.rs
```diff
@@ -41,14 +43,30 @@ pub struct StorageMetrics {
...
+            "Outcome of get_slice operations. success_after_1_timeout means the operation \
+             succeeded after a retry caused by a timeout.",
+            "storage",
+            &[],
+            ["outcome"],
+        );
+        let get_slice_timeout_successes = [
+            get_slice_timeout_outcome_total_vec.with_label_values(["success_after_0_timeout"]),
+            get_slice_timeout_outcome_total_vec.with_label_values(["success_after_1_timeout"]),
+            get_slice_timeout_outcome_total_vec.with_label_values(["success_after_2+_timeout"]),
```
trinity-1686a, line 57: i think these labels look odd, but don't have concrete improvement proposal to make

trinity-1686a (approved):

quickwit/quickwit-storage/src/metrics.rs
esatterwhite: If it's a metric, it should be 1 label. Maybe `attempts`. 
But I am not sure how valuable there is there. 

If I can see number of retries as a metric, total attempts as a metric, and failed attempts as a metric, I have enough to tell me storage is having a problem. 

These labels are just adding noise and bloat to the metrics that I would have to remove anyway

quickwit/quickwit-storage/src/timeout_and_retry_storage.rs
esatterwhite: How is an error like that handled? Does the entire request fail, or does it surface as a failed shard with the recent change you made?

fulmicoton: There will be one retry happening on a different node. If that retry fails too, then the split will show up as a failed shards.

quickwit/quickwit-storage/src/metrics.rs
fulmicoton: @esatterwhite I don't understand your comment.

> If it's a metric, it should be 1 label

If you mean a single label key, this is the case right now. 
`success_after_0_timeout` in here is a label value.
If you mean a single label value, this is like have no label at all.

I am open to change the metrics again, but I would prefer an accurate 
description, that prometheus users won't be at risk of misinterpretation.

esatterwhite: maybe a single label of `success_after_timeouts=N`. 
It seems like you're wanting a histogram that observes how many timeouts occurred before there was a success.

It feels like this is really 2 different metrics. But it doesn't need to hold up getting the fix out.

fulmicoton: Merging without change. We can revisit this later if it is not practical.

PR #5470, opened 2024-10-02
Fix pulsar uri scheme

miton18:
### Description

Allow to use a Pulsar instance which enable TLS

### How was this PR tested?

On our internal Pulsar cluster

### DIFF
--- a/quickwit/quickwit-config/src/source_config/mod.rs
+++ b/quickwit/quickwit-config/src/source_config/mod.rs
@@ -28,6 +28,7 @@ use quickwit_common::is_false;
 use quickwit_common::uri::Uri;
 use quickwit_proto::metastore::SourceType;
 use quickwit_proto::types::SourceId;
+use regex::Regex;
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-config/src/source_config/mod.rs
```diff
@@ -498,8 +499,9 @@ pub enum PulsarSourceAuth {
 fn pulsar_uri<'de, D>(deserializer: D) -> Result<String, D::Error>
 where D: Deserializer<'de> {
     let uri: String = Deserialize::deserialize(deserializer)?;
+    let re: Regex = Regex::new(r"pulsar(\+ssl)?://.*").unwrap();
```
guilload, line 502:
```suggestion
    let re: Regex = Regex::new(r"pulsar(\+ssl)?://.*").expect("regular expression should compile");
```

guilload (approved):

PR #5472, opened 2024-10-02
Add number of splits per root/leaf search histograms

guilload:
### Description
Per title.

### How was this PR tested?
`make test-all`

### DIFF
--- a/quickwit/quickwit-search/src/leaf.rs
+++ b/quickwit/quickwit-search/src/leaf.rs
@@ -48,6 +48,7 @@ use tokio::task::JoinError;
 use tracing::*;
 
 use crate::collector::{make_collector_for_split, make_merge_collector, IncrementalCollector};
+use crate::metrics::SEARCH_METRICS;
 use crate::root::is_metadata_count_request_with_ast;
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1304,13 +1305,23 @@ pub async fn leaf_search(
         });
     }
 
-    let result = crate::search_thread_pool()
+    let leaf_search_response_result = crate::search_thread_pool()
         .run_cpu_intensive(|| incremental_merge_collector.finalize())
         .instrument(info_span!("incremental_merge_intermediate"))
         .await
-        .context("failed to merge split search responses")??;
+        .context("failed to merge split search responses")?;
```
trinity-1686a, line 1312: hopefully the collector doesn't panic, but if it does, this should probably count as an error. This is currently not accounted for in metrics

trinity-1686a (approved):

quickwit/quickwit-search/src/leaf.rs
guilload: Good point. I was sloppy :/ Will fix and merge.

PR #5475, opened 2024-10-04
Issue/5474 merge pipeline finalization

fulmicoton:
Before this PR, when the last indexing pipeline
is shutdown, the merge pipeline associated to an index
would be abruptly shutdown too.

This PR makes two changes to this behavior.
First, we wait for ongoing or pending merges to
be executed before shutting down the pipeline.

Second, merge policy get the opportunity to offer
a list of extra merges to run.

This functionality is introduced to help users
who migrated from elasticsearch and rely on a daily
indexes.

Closes https://github.com/quickwit-oss/quickwit/issues/5474


guilload (approved):
The abundant comments made the PR super easy to review. Thanks!

PR #5489, opened 2024-10-14
don't enumerate index_uid when requesting splits to gc

trinity-1686a:


quickwit/quickwit-metastore/src/metastore/file_backed/mod.rs
```diff
@@ -379,22 +379,54 @@ impl FileBackedMetastore {
     /// Returns the list of splits for the given request.
     /// No error is returned if any of the requested `index_uid` does not exist.
```
fulmicoton, line 380: there is no mention about the expected order, yet it seems like the code takes great care to respect some very specific order.

quickwit/quickwit-index-management/src/garbage_collection.rs
```diff
@@ -317,10 +313,12 @@ async fn delete_splits_marked_for_deletion_several_indexes(
...
-    let Some(list_splits_query) = ListSplitsQuery::try_from_index_uids(index_uids) else {
-        error!("failed to create list splits query. this should never happen");
-        return split_removal_info;
-    };
+    // we ask for all indexes because the query is more efficient and we almost always want all
+    // indexes anyway. The exception is when garbage collecting a single index from the commandline.
+    // In this case, we will log a bunch of warn. i (trinity) consider it worth the more generic
+    // code which needs fewer special case while testing, but we could check index_uids len if we
+    // think it's a better idea.
+    let list_splits_query = ListSplitsQuery::for_all_indexes();
```
fulmicoton, line 321: isn't the post filtering missing here?

fulmicoton (approved):

quickwit/quickwit-index-management/src/garbage_collection.rs
```diff
@@ -340,19 +341,33 @@ async fn delete_splits_marked_for_deletion_several_indexes(
...
-                .into_group_map();
+        for meta in splits_metadata_to_delete {
+            if !storages.contains_key(&meta.index_uid) {
+                info!(index_uid=?meta.index_uid, "split not listed in storage map: skipping");
+                continue;
+            }
+            if let Some(splits) = splits_metadata_to_delete_per_index.get_mut(&meta.index_uid) {
+                splits.push(meta);
+            } else {
+                splits_metadata_to_delete_per_index.insert(meta.index_uid.clone(), vec![meta]);
```
trinity-1686a, line 368: this can be expressed with the [entry](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry) api, which imo is easier to read

quickwit/quickwit-index-management/src/garbage_collection.rs
```diff
@@ -340,19 +341,33 @@ async fn delete_splits_marked_for_deletion_several_indexes(
...
+            HashMap::with_capacity(storages.len());
 
-        let splits_metadata_to_delete_per_index: HashMap<IndexUid, Vec<SplitMetadata>> =
-            splits_metadata_to_delete
-                .into_iter()
-                .map(|meta| (meta.index_uid.clone(), meta))
-                .into_group_map();
+        for meta in splits_metadata_to_delete {
+            if !storages.contains_key(&meta.index_uid) {
+                info!(index_uid=?meta.index_uid, "split not listed in storage map: skipping");
```
trinity-1686a, line 362: this won't log often, but when it does, it may log *a lot*. this should be rate-limited

quickwit/quickwit-index-management/src/garbage_collection.rs
fulmicoton: Yeah. I was avoiding the needless allocation. I can remove that.

quickwit/quickwit-index-management/src/garbage_collection.rs
fulmicoton: good point

fulmicoton (approved):

PR #5490, opened 2024-10-15
Bugfix sort order query string in elastic search api.

fulmicoton:


trinity-1686a:
fix https://github.com/quickwit-oss/quickwit/issues/5491

PR #5492, opened 2024-10-15
fix(query-documentation): typo range query

idrissneumann:
### Description

Fix a typo in this documentation: https://quickwit.io/docs/get-started/query-language-intro

`duration:[0 TO 1000]` instead of `duration:[0 TO 1000}`


guilload:
That's not necessarily a typo: `[` is for inclusive, `{` is for exclusive. I'd be better to leave the examples as is and enhance the description.

PR #5509, opened 2024-10-22
Prevents single split searches from different leaf_search from interleaving

fulmicoton:
Before this PR, we just used a semaphore to acquire a permit and start a new tokio task to run our single split search.

In pseudo code, a leaf_search would look like:
```
for split in splits {
    let permit = semaphore.acquire().await;
    tokio::spawn(async move {
       single_split_search(split);
       drop(permit)
    })
}
```

The problem with this is that it allows interleaving split search from one search request with another one.

This interleaving strongly impacts search latency. For instance, one can imagine two queries A and B with 3 splits each. Executing as follows
| A | A | A | B | B | B |
offers a much short latency for A than
| A | B | B | A | B | A |

This PR also adds two metrics to monitor the number of queue single split search.

### DIFF
--- a/quickwit/quickwit-search/src/search_permits.rs
+++ b/quickwit/quickwit-search/src/search_permits.rs
@@ -1,0 +1,213 @@
+// Copyright (C) 2024 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+use std::collections::VecDeque;
+use std::sync::{Arc, Mutex, Weak};
+
+use quickwit_common::metrics::GaugeGuard;
+use tokio::sync::oneshot;
+
+/// `SearchPermits` is a distributor of permits to perform single split
+/// search operation.
+///
+/// Requests are served in order.
+#[derive(Clone)]
+pub struct SearchPermits {
+    inner: Arc<Mutex<InnerSearchPermits>>,
+}
+
+impl SearchPermits {
+    pub fn new(num_permits: usize) -> SearchPermits {
+        SearchPermits {
+            inner: Arc::new(Mutex::new(InnerSearchPermits {
+                num_permits_available: num_permits,
+                permits_requests: VecDeque::new(),
+            })),
+        }
+    }
+
+    /// Returns a list of future permits in the form of a Receiver channel.
+    pub fn get_one_permit(&self) -> oneshot::Receiver<SearchPermit> {
+        let mut permits_lock = self.inner.lock().unwrap();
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-search/src/search_permits.rs
```diff
@@ -0,0 +1,213 @@
...
+
+use quickwit_common::metrics::GaugeGuard;
+use tokio::sync::oneshot;
+
+/// `SearchPermits` is a distributor of permits to perform single split
+/// search operation.
+///
+/// Requests are served in order.
+#[derive(Clone)]
+pub struct SearchPermits {
```
trinity-1686a, line 31: this is a misnomer, it sounds like this is a akin to a tokio `SemaphorePermit` with num_permits > 1, but this is actually a semaphore, or a PermitProvider, or something of the like

quickwit/quickwit-search/src/search_permits.rs
```diff
@@ -0,0 +1,213 @@
...
+        }
+        crate::SEARCH_METRICS
+            .leaf_search_single_split_tasks_pending
+            .set(self.permits_requests.len() as i64);
+    }
+}
+
+pub struct SearchPermit {
+    _ongoing_gauge_guard: GaugeGuard<'static>,
+    inner_opt: Option<Weak<Mutex<InnerSearchPermits>>>,
```
trinity-1686a, line 127: I don't believe there is need for Weak here: `InnerSearchPermits` doesn't hold any `SearchPermit` (a oneshot sender doesn't hold the thing to transmit, the oneshot receiver do)

i find this type, `Option<Weak<Mutex<InnerSearchPermits>>>` rather confusing, in part because SearchPermit**s** isn't the right name imo, in part because the `Option<>` very much looks like a way to store a bool alongside the rest of the structure to reduce the size of the struct by one usize. I'm not against using an Option that way, but if that's the reason, it ought to be written down, and if it isn't the goal, then i'm missing what the goal of this Option is, rather than a `disable_drop: bool`

trinity-1686a (commented):
I think the changes make sense, but some struct/params names make the code more confusing than it could be

quickwit/quickwit-search/src/search_permits.rs
fulmicoton: I prefer the `Option`. Setting it to None seems easier to "proofread". I agree that it makes the code less readable however, as the intent is way less clearer than with the boolean solution.

I'll add some comments.

quickwit/quickwit-search/src/search_permits.rs
fulmicoton: Yeah you are right.

quickwit/quickwit-search/src/search_permits.rs
```diff
@@ -0,0 +1,213 @@
...
+struct InnerSearchPermits {
+    num_permits_available: usize,
+    permits_requests: VecDeque<oneshot::Sender<SearchPermit>>,
+}
+
+impl InnerSearchPermits {
+    fn get_permits(
+        &mut self,
+        num_permits: usize,
+        inner: &Arc<Mutex<InnerSearchPermits>>,
```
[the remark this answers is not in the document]
fulmicoton: I don't quite get the confusion/comment.
The "outter" is SearchPermits.

I have not found a common practise on how to name the object of this pattern in Rust (in C++ `pimpl` is common, even if this is not the common usage of the pimpl pattern).

 What would be a better name? would inner_arc help a bit?

quickwit/quickwit-search/src/search_permits.rs
trinity-1686a: you ended up changing it to a bool. If you think an Option is better, you can put it back, just add a 1-2 line comment on the purpose of None

trinity-1686a (approved):

quickwit/quickwit-search/src/search_permits.rs
fulmicoton: Yeah, I thought after packing everything as `drop_without_recycling_permit` it was fine like that.

PR #5514, opened 2024-10-22
Improve search job placer logging

### LINKED ISSUE #5507: Improve logging on job placer "no available searcher nodes in the pool" error
Is the pool empty? Are we running out of candidates after retrying on all the nodes (can happen on 2-3 node cluster)

guilload:
### Description
Per title. Fixes #5507.

### How was this PR tested?
Ran `quickwit-search` unit tests.

### DIFF
--- a/quickwit/quickwit-search/src/search_job_placer.rs
+++ b/quickwit/quickwit-search/src/search_job_placer.rs
@@ -28,7 +28,7 @@ use async_trait::async_trait;
 use quickwit_common::pubsub::EventSubscriber;
 use quickwit_common::rendezvous_hasher::{node_affinity, sort_by_rendez_vous_hash};
 use quickwit_proto::search::{ReportSplit, ReportSplitsRequest};
-use tracing::warn;
+use tracing::{info, warn};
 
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-search/src/search_job_placer.rs
```diff
@@ -150,39 +150,51 @@ impl SearchJobPlacer {
...
-        let mut candidate_nodes: Vec<CandidateNodes> = self
-            .searcher_pool
-            .pairs()
+        if all_nodes.is_empty() {
+            bail!(
+                "failed to assign search jobs: there are no available searcher nodes in the \
+                 cluster"
+            );
+        }
+        if !excluded_addrs.is_empty() && excluded_addrs.len() < all_nodes.len() {
```
PSeitz, line 161: The condition `excluded_addrs.len() < all_nodes.len()` seems broken.
if there are more excluded addresses than nodes, we don't filter anything

The tests below don't cover `excluded_addrs`

trinity-1686a: this seems to be on purpose. Per function doc:
> /// When exclude_addresses filters all clients it is ignored.

trinity-1686a (approved):

PR #5521, opened 2024-10-23
Record object storage request latencies

guilload:
### Description
Per title.

### How was this PR tested?
Ran `quickwit-storage` test suite


trinity-1686a (approved):

quickwit/quickwit-storage/src/metrics.rs
```diff
@@ -41,6 +41,15 @@ pub struct StorageMetrics {
...
     pub object_storage_upload_num_bytes: IntCounter,
+
+    pub object_storage_delete_requests_total: IntCounter,
+    pub object_storage_bulk_delete_requests_total: IntCounter,
+
+    pub object_storage_get_request_duration: Histogram,
+    pub object_storage_put_request_duration: Histogram,
+    pub object_storage_upload_part_request_duration: Histogram,
+    pub object_storage_delete_request_duration: Histogram,
+    pub object_storage_bulk_delete_request_duration: Histogram,
```
fulmicoton, line 52: These will create a lot of metrics (one per bucket) and I cannot think of a use case for them.
For instance GET and PUT request duration are generally interesting, but they strongly depend on the size of the payload.

Do we need them?

guilload: Users want storage latencies. Delete request latencies would be useful for understanding why GC'ing splits is so slow on Fly.

You make some good points, though. Latencies for PUT requests are not useful and will be captured by the upload rate.

For GET requests, we could add a size dimension, so we could build a heatmap, but at the end of the day, time to first byte is what we really want.

I'll remove the GET/PUT histograms. Keep the delete stuff and reduce the number of buckets since we're only interested in knowing "are delete requests fast, slow, or very slow".

Issue #5312, opened 2024-08-13
Support X-Correlation-ID in header


PSeitz:
Add support for `X-Correlation-ID` in the header.

The `X-Correlation-ID` is an id that gets passed by a client and is getting printed in the logs. 
This allows end to end tracing of requests across systems.

rdettai:
could be interesting to use trace propagation with https://github.com/tokio-rs/tracing-opentelemetry

fulmicoton:
So we do have trace propagation using opentelemetry. I think Pascal means (I am not sure) putting the trace in logs, so that we can look at logs, get a request id, and check at the trace.

I wrote code to extract the trace id a few years ago. I ended up trashing it assuming we would never use it.

PR #5524, opened 2024-10-25
Limit warmup memory usage

fulmicoton:
We also measure the amount of memory taken by a split search, and log this.

### DIFF
--- a/quickwit/quickwit-search/src/root.rs
+++ b/quickwit/quickwit-search/src/root.rs
@@ -694,7 +694,8 @@ pub fn get_count_from_metadata(split_metadatas: &[SplitMetadata]) -> Vec<LeafSea
 /// This function only considers the memory usage associated to the input data
 /// and does not take in account aggregations (intermediary or not) results for instance.
 ///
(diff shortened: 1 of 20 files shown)


PSeitz:
Related: https://github.com/quickwit-oss/quickwit/issues/5312

quickwit/quickwit-search/src/root.rs
```diff
@@ -744,9 +781,21 @@ pub(crate) async fn search_partial_hits_phase(
...
     );
+
+    if let Some(resource_stats) = &leaf_search_response.resource_stats {
+        if is_top_1pct_memory_intensive(
+            resource_stats.short_lived_cache_num_bytes,
+            resource_stats.split_num_docs,
+        ) {
+            // We log at most 5 times per minute.
+            quickwit_common::rate_limited_info!(limit_per_min=5, split_num_docs=resource_stats.split_num_docs, %search_request.query_ast, short_lived_cached_num_bytes=resource_stats.short_lived_cache_num_bytes, query=%search_request.query_ast, "memory intensive query");
+        }
```
rdettai, line 792: should we create a metric as well?

PR #5529, opened 2024-10-28
Added support for Quickwit enterprise license.

fulmicoton:


quickwit/quickwit-license/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+
+#[cfg(test)]
+const QUICKWIT_PUBLIC_KEY_HEX: &'static str = "5a81acfc945c7c64c3e95a0c8ce2caafec7c496e76db9ff528755eca7fc16c25";
+
+#[cfg(not(test))]
+const QUICKWIT_PUBLIC_KEY_HEX: &'static str = "c8a5f551d1744547498c97638ce5c5520755cb9de0f052dcee07f90dc9663042";
+
+#[derive(Debug)]
+pub struct License {
+    pub license_level: LicenseLevel,
```
PSeitz, line 33: If we want to be able to enable features separately, we would need something more here

fulmicoton: I think we can stick to license level, at least for the moment.

quickwit/scripts/.ee.license_header.txt
```diff
@@ -0,0 +1,20 @@
+// The Quickwit Enterprise Edition (EE) license
+// Copyright (c) {\d+}-present Quickwit Inc.
+//
+// With regard to the Quickwit Software:
+//
+// This software and associated documentation files (the "Software") may only be
+// used in production, if you (and any entity that you represent) hold a valid
```
PSeitz, line 7: I assume the production part should allow test use cases, but the wording is not clear, e.g. it can be understand as it can only be used in production and nowhere else.

Maybe we can remove the production part, and it can be covered by "Quickwit Enterprise license corresponding to your usage". We would need to emit access tokens anyway.

fulmicoton: This is the wording used in several other famous license. It should be ok.

quickwit/quickwit-license/src/lib.rs
fulmicoton: I did not pick that from a specific license, and I am not sure of the origin of this formulation: you can google "license" "may only be used in production" to see a bunch of examples.

PSeitz (approved):

altanozlu:
Hey @fulmicoton, to be clear after this change, is an enterprise license required for internal production use cases of Quickwit?

fulmicoton:
No. It is just for specific new features.

PR #5532, opened 2024-10-30
make tokenizer emit at least one empty token on empty strings

trinity-1686a:
### Description

always emit at least one token when indexing/querying with an empty field

### How was this PR tested?

tested manually. todo: add unit and integration test

### DIFF
--- a/quickwit/quickwit-query/src/query_ast/full_text_query.rs
+++ b/quickwit/quickwit-query/src/query_ast/full_text_query.rs
@@ -316,6 +316,7 @@ mod tests {
     use crate::{create_default_quickwit_tokenizer_manager, BooleanOperand};
 
     #[test]
+    #[ignore]
     fn test_zero_terms() {
         let full_text_query = FullTextQuery {
             field: "body".to_string(),
--- a/quickwit/quickwit-query/src/tokenizers/mod.rs
+++ b/quickwit/quickwit-query/src/tokenizers/mod.rs
@@ -26,7 +26,7 @@ mod tokenizer_manager;
 use once_cell::sync::Lazy;
 use tantivy::tokenizer::{
     AsciiFoldingFilter, Language, LowerCaser, RawTokenizer, RemoveLongFilter, SimpleTokenizer,
-    Stemmer, TextAnalyzer, WhitespaceTokenizer,
+    Stemmer, TextAnalyzer, Token, TokenStream, Tokenizer, WhitespaceTokenizer,
 };
 
 use self::chinese_compatible::ChineseTokenizer;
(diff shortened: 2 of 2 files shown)


quickwit/quickwit-query/src/query_ast/full_text_query.rs
```diff
@@ -316,6 +316,7 @@ mod tests {
     use crate::{create_default_quickwit_tokenizer_manager, BooleanOperand};
 
     #[test]
+    #[ignore]
```
trinity-1686a, line 319: the behavior this test tries for changed, I don't think it make sense to keep it

PSeitz:
Why do we need this?

quickwit/quickwit-query/src/tokenizers/mod.rs
```diff
@@ -128,6 +132,69 @@ pub fn get_quickwit_fastfield_normalizer_manager() -> &'static TokenizerManager
     &QUICKWIT_FAST_FIELD_NORMALIZER_MANAGER
 }
 
+#[derive(Debug, Clone)]
+struct EmitEmptyTokenizer<T>(T);
```
fulmicoton, line 136: We need a comment to express the intent on that kind of tokenizer. It is very difficult for a future readers to infer the point of this.

quickwit/quickwit-query/src/tokenizers/mod.rs
```diff
@@ -128,6 +132,69 @@ pub fn get_quickwit_fastfield_normalizer_manager() -> &'static TokenizerManager
...
+                true
+            }
+            EmitEmptyState::UsingInner => self.inner.advance(),
+            EmitEmptyState::EmitEmpty(_) => false,
+        }
+    }
+
+    fn token(&self) -> &Token {
+        match self.state {
+            EmitEmptyState::Start => unreachable!(),
```
fulmicoton, line 183: It is reachable.

```suggestion
            EmitEmptyState::Start => panic!("token() should not be called before a first call to advance"),
```

fulmicoton (changes requested):
We need to explain the intent in the commit message and in the comment of the tokenizer.

A unit test checking for the behavior end to end (indexing + query side) should be added too.

We also need to think about what the behavior should be for a multivalued field
`["hello", ""]` (does it match or not), and add a unit test to validate this behavior.

trinity-1686a:
> Why do we need this?

Today `{"field": ""}` and `{}` get the same tokens for most tokenizers (except raw), which is no token at all. It's not possible to search for empty strings only. This allows doing just that

PSeitz:
> > Why do we need this?
> 
> Today `{"field": ""}` and `{}` get the same tokens for most tokenizers (except raw), which is no token at all. It's not possible to search for empty strings only. This allows doing just that

Tokenization is already quite expensive during indexing, I think this change may add some non-negligable overhead there.

quickwit/quickwit-query/src/tokenizers/mod.rs
fulmicoton: This tokenizer emits the empty token whenever the underlying tokenizer emits no token.
I am not sure this is the behavior we want.

For instance, a stop word would result in the emission of the empty token.

A stricter behavior, emitting an empty token iff the text is effectively empty might actually make more sense. I suspect the code would be simpler too.

PR #5535, opened 2024-11-01
Parallelize S3 delete objects requests

guilload:
### Description
Parallelize S3 delete objects requests

### How was this PR tested?
Ran `quickwit-storage` test suite locally.


trinity-1686a (approved):
some non-blocking observations on lifetime&async, and atomic ordering

guilload:
This PR is no longer necessary and since the patch is complex, I'd rather close.

Issue #5539, opened 2024-11-04
Compile Times + Warp Types


PSeitz:
Quickwit compile times are pretty long, this could by some part be caused by overly complex warp types. 
Below is an example of such a type. We probably should break these type chains by boxing them.

[...]

PR #5540, opened 2024-11-05
box warp routes

PSeitz:
box warp routes to avoid super heavy types.
Performance impact should be okay, since these heavy types
are not cache friendly.

This reduces type complexity significantly, but some really long type chains remain.
I've identified two, but they may be more
1. Boxed or chains
2. Response type chains

We can replace the boxed or chains with a Vec, since they all contain the same
box type now. Such a type doesn't seem to exist in warp yet.

`cargo install --path .`
Compile time before
4m49s
Compile time now
4m14s

Tool to list heavy functions:
`
CARGO_PROFILE_RELEASE_LTO=fat cargo llvm-lines --release --bin quickwit > llvm_lines
`
Size down from 32MB to 22MB

update rust tool chain to 1.81 to fix compilation error (https://github.com/servo/rust-url/issues/992)

addresses #5539

### DIFF
--- a/quickwit/quickwit-codegen/example/src/lib.rs
+++ b/quickwit/quickwit-codegen/example/src/lib.rs
@@ -62,6 +62,7 @@ where S: Service<R>
 }
 
 #[derive(Debug, Clone, Default)]
+#[allow(dead_code)]
 struct CounterLayer {
     counter: Arc<AtomicUsize>,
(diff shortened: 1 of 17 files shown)


fulmicoton (approved):

fulmicoton:
nice!

Issue #5431, opened 2024-09-17
precondition failed for splits


sergeyignatov:
Hi, 
We have 4 indexes in quickwit, all indexes are in GCS, quickwit version 0.8.2. 
Recently started issue with the biggest index fluentbit-logs-2024-09-12 when splits older 1h is automatically removed.  Other indexes are ok. The issue is not gone after the index recreation.

The affected index has way more Staged splits than other indexes. 
```
select count(*), split_state, index_uid from splits  group by index_uid, split_state order by index_uid;
 count |    split_state    |                          index_uid
-------+-------------------+--------------------------------------------------------------

  1941 | Staged            | fluentbit-logs-2024-09-12:01J7MNHZMJSJJ80B2JN40CE12J
   485 | Published         | fluentbit-logs-2024-09-12:01J7MNHZMJSJJ80B2JN40CE12J
  9851 | MarkedForDeletion | fluentbit-logs-2024-09-12:01J7MNHZMJSJJ80B2JN40CE12J
  8349 | MarkedForDeletion | fluentbit-logs-connect-2024-02-05:01HSA1Z6D9N74GPZ0DX9B3RVYV
   902 | Published         | fluentbit-logs-connect-2024-02-05:01HSA1Z6D9N74GPZ0DX9B3RVYV
    16 | Staged            | fluentbit-logs-connect-2024-02-05:01HSA1Z6D9N74GPZ0DX9B3RVYV
  9156 | MarkedForDeletion | fluentbit-logs-sk-2024-01-30:01HNDX8JAKT5A6A06BFVZF8TC2
  1020 | Published         | fluentbit-logs-sk-2024-01-30:01HNDX8JAKT5A6A06BFVZF8TC2
    15 | Staged            | fluentbit-logs-safedk-2024-01-30:01HNDX8JAKT5A6A06BFVZF8TC2
```
In the indexer logs the only errors:
```
[...]

PR #5543, opened 2024-11-06
restrict maturity period to retention

trinity-1686a:
### Description

reduce time before maturity of a split based on time to end of retention
reduce likelihood of encountering #5431 . It can technically still happen if a split is subject to merge at time t, its merge start, and then subject deletion at time t+1 while the merge is still running, but that should be a much rarer occurrence. 

### How was this PR tested?

added unit test for new code

### DIFF
--- a/quickwit/quickwit-indexing/src/models/split_attrs.rs
+++ b/quickwit/quickwit-indexing/src/models/split_attrs.rs
@@ -21,8 +21,9 @@ use std::collections::BTreeSet;
 use std::fmt;
 use std::ops::{Range, RangeInclusive};
 use std::sync::Arc;
+use std::time::Duration;
 
-use quickwit_metastore::SplitMetadata;
+use quickwit_metastore::{SplitMaturity, SplitMetadata};
 use quickwit_proto::types::{DocMappingUid, IndexUid, NodeId, SourceId, SplitId};
 use tantivy::DateTime;
 use time::OffsetDateTime;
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-indexing/src/models/split_attrs.rs
```diff
@@ -120,3 +133,92 @@ pub fn create_split_metadata(
         num_merge_ops: split_attrs.num_merge_ops,
     }
 }
+
+/// reduce the maturity period of a split based on retention policy, so that it doesn't get merged
+/// after it expires.
+fn max_maturity_before_end_of_retention(
```
fulmicoton, line 139: I feel like the code would be clearer and easier to reason with if you defined an order relationship on Maturity.

The relationship would be built upon the idea of "the lower" = "the sooner you reach maturity".
So mature < immature.
immature(x) < immature(y) if x < y.

We could then imply a SplitMaturity based on the retention period. Store that as the "max_maturity" instead of carrying around a RetentionPolicy object, and just compute a `min` here.

trinity-1686a: i agree with implementing Ord for maturity, but I don't think it would remove the need for having the retention policy (or period) passed here: we can't compute a maximum maturity from just the retention period, we also need the split timestamp_range.end

fulmicoton: You are right.

fulmicoton (approved):

PR #5544, opened 2024-11-06
Parse datetimes and timestamps with leading and/or trailing whitespace

guilload:
### Description
Per title. Request by Airmail. Supported by ES.

### How was this PR tested?
- Updated unit tests
- `make test-all`


trinity-1686a (approved):
I wonder if for `Strptime` patterns we might not want to trim the parser spec: if the parser is `  %Y  ` (this pattern only parse years, but that's not the point), it won't parse `   2024   ` as valid because it will expect those spaces we just removed

guilload:
That makes sense. Let me add a commit on top of this PR.

quickwit/quickwit-datetime/src/date_time_parsing.rs
```diff
@@ -36,22 +36,24 @@ pub fn parse_date_time_str(
     date_time_str: &str,
     date_time_formats: &[DateTimeInputFormat],
 ) -> Result<TantivyDateTime, String> {
+    let trimmed_date_time_str = date_time_str.trim();
```
fulmicoton, line 39:
```suggestion
    let trimmed_date_time_str = date_time_str.trim_ascii();
```

fulmicoton: It is probably minor, but triming unicode is considerably more expensive than trimming ascii. 
(It requires decoding utf-8, and check if each individual char is a whitespace or not. )

I'd feel safer if we restricted ourselves to ascii. It will just prevent us from trimming weird whitespace like the japanese "　".

fulmicoton:
@guilload can you merge this?

trinity-1686a:
it seems the strptime is already lenient to whitespaces inside the pattern (but not the parsed string, which we trim ourselves anyway). I updated the tests in consequence 

the java-compatible parser used for range requests doesn't have access to that trim code I think, but it's not going to reject documents, so I don't think we necessarily care as much

fulmicoton:
@guilload can we merge this?

guilload:
Yes, we should be good.

PR #5568, opened 2024-11-28
 Limit and monitor warmup memory usage

rdettai:
### Description

To better control the memory usage during search, we add some tooling around the warmup cache.

### How was this PR tested?

### TODO
- [x] validate initial allocation rules
- [ ] ~docs~ hide this feature for now


quickwit/quickwit-search/src/root.rs
```diff
@@ -744,9 +781,21 @@ pub(crate) async fn search_partial_hits_phase(
...
     );
+
+    if let Some(resource_stats) = &leaf_search_response.resource_stats {
+        if is_top_5pct_memory_intensive(
+            resource_stats.short_lived_cache_num_bytes,
+            resource_stats.split_num_docs,
+        ) {
+            // We log at most 5 times per minute.
+            quickwit_common::rate_limited_info!(limit_per_min=5, split_num_docs=resource_stats.split_num_docs, %search_request.query_ast, short_lived_cached_num_bytes=resource_stats.short_lived_cache_num_bytes, query=%search_request.query_ast, "memory intensive query");
+        }
```
rdettai, line 792: should we create a metric as well?

quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -226,6 +226,8 @@ pub struct SearcherConfig {
     #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub storage_timeout_policy: Option<StorageTimeoutPolicy>,
+    pub warmup_memory_budget: ByteSize,
```
fulmicoton, line 229: Can we define a default value for this?

And reuse the default value function in SearcherConfig::default for consistency

quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -274,6 +276,8 @@ impl Default for SearcherConfig {
             split_cache: None,
             request_timeout_secs: Self::default_request_timeout_secs(),
             storage_timeout_policy: None,
+            warmup_memory_budget: ByteSize::gb(1),
```
fulmicoton, line 279: I think we want larger defaults here. 
Maybe memory budget of 10GB
and single split alloc to 1GB or something like this.

quickwit/quickwit-directories/src/caching_directory.rs
```diff
@@ -30,40 +30,44 @@ use tantivy::{Directory, HasLen};
...
 
-impl fmt::Debug for CachingDirectory {
+impl<T> fmt::Debug for CachingDirectory<T> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "CachingDirectory({:?})", self.underlying)
     }
 }
 
-struct CachingFileHandle {
+struct CachingFileHandle<C> {
```
fulmicoton, line 64: what is the extra complexity for?

rdettai: To avoid dropping the permits early or late compared to the actual memory dealocation, I want to attach the `SearchPermit`'s lifecycle to the `ByteRangeCache`. But this wasn't easy because both `ByteRangeCache` and the abstraction above `CachingDirectory` are used in multiple place, with and without a permit. Making the `CachingDirectory` support both a cache instance that is tied to a permit or not is the simplest way I found. Then remained the question of dynamic VS static dispatch. I opted for static dispatch because I found that its explicit display of the kind of cache used on the caller side is actually useful. It makes it clear whether we use a cache that holds a permit or not, and exhibits well which allocations are controlled by the memory pool and which aren't.

The specific type parameter on `CachingFileHandle<C>` that you highlight here is a collateral damage (required because file handles contain a reference to the directory), but it's pretty minor given that it is a private data structure.

quickwit/quickwit-config/src/node_config/mod.rs
rdettai: Should I also do the same for all the other fields of the same struct? https://github.com/quickwit-oss/quickwit/blob/c3de281064deb42dc19554291500957b314baf0c/quickwit/quickwit-config/src/node_config/mod.rs#L211-L218

quickwit/quickwit-directories/src/caching_directory.rs
fulmicoton: I don't think this is a good addition. I understand the point of tying search permit to the byte range, but:
- this is a bit overkill
- it gives the misleading impression that the memory permit is reflecting accurately the content of the byte range cache. This is not the case.

quickwit/quickwit-search/src/lib.rs
```diff
@@ -340,3 +343,126 @@ pub fn searcher_pool_for_test(
             }),
     )
 }
+
+pub(crate) fn merge_resource_stats_it<'a>(
+    stats_it: impl IntoIterator<Item = &'a Option<ResourceStats>>,
```
fulmicoton, line 348: I think this was my code originally, but I suspect 

```suggestion
    stats_it: impl IntoIterator<Item = &'a ResourceStats>,
```
would be ok too.

quickwit/quickwit-search/src/collector.rs
```diff
@@ -919,6 +921,11 @@ fn merge_leaf_responses(
         return Ok(leaf_responses.pop().unwrap());
     }
 
+    let resource_stats_it = leaf_responses
+        .iter()
+        .map(|leaf_response| &leaf_response.resource_stats);
```
fulmicoton, line 926:
```suggestion
        .flat_map(|leaf_response| leaf_response.resource_stats.as_ref());
```

quickwit/quickwit-search/src/root.rs
fulmicoton: I think an histogram with not too many bucket for general memory usage. might be nice. what do you think?

quickwit/quickwit-search/src/lib.rs
rdettai: I changed this for `Option` because otherwise the caller side is in charge of handling the `None` case. This way, the policy for handling `None` is defined once and for all and tested.

quickwit/quickwit-search/src/search_permit_provider.rs
```diff
@@ -18,151 +18,272 @@
...
+/// warmup (limit concurrent downloads) and a pessimistic amount of memory. Once
+/// the warmup is completed, the actual memory usage is set and the warmup slot
+/// is released. Once the search is completed and the permit is dropped, the
+/// remaining memory is also released.
 #[derive(Clone)]
 pub struct SearchPermitProvider {
-    inner_arc: Arc<Mutex<InnerSearchPermitProvider>>,
+    message_sender: mpsc::UnboundedSender<SearchPermitMessage>,
+    #[cfg(test)]
+    actor_stopped: watch::Receiver<bool>,
```
fulmicoton, line 43: why not just use `message_sender.is_closed()` ?

quickwit/quickwit-search/src/root.rs
rdettai: sounds good! The stats struct also contains other metrics: warmup_microsecs, cpu_thread_pool_wait_microsecs, cpu_microsecs. Did you have any plans for those?

quickwit/quickwit-search/src/search_permit_provider.rs
rdettai: of course...

rdettai: ah no, I remember, it's because it's precisely dropping the sender that is supposed to stop the actor

quickwit/quickwit-directories/src/caching_directory.rs
rdettai: Very good point. I'll remove it.

quickwit/quickwit-search/src/collector.rs
rdettai: same as https://github.com/quickwit-oss/quickwit/pull/5568#discussion_r1865520137

quickwit/quickwit-config/src/node_config/mod.rs
fulmicoton: Yes, but does it apply to the individual attributes? It really works that way?

fulmicoton: just tested. yep that works that way.

quickwit/quickwit-search/src/root.rs
rdettai: As discussed, we will not add timing metrics for now and just add a histogram about the final size of individual short lived caches (note that `shortlived_cache` was already added as global storage metric)

quickwit/quickwit-search/src/search_permit_provider.rs
```diff
@@ -18,151 +18,312 @@
...
+    let size_bytes = [
+        split_size,
+        proportional_allocation,
+        warmup_single_split_initial_allocation.as_u64(),
+    ]
+    .into_iter()
+    .min()
+    .unwrap();
+    ByteSize(size_bytes)
 }
```
rdettai, line 81: I'm not convinced by this estimate logic. The `proportional_allocation` makes it very hard to define what `warmup_single_split_initial_allocation` actually is.

rdettai: Ran some tests on the [hdfs dataset](https://raw.githubusercontent.com/quickwit-oss/quickwit/main/config/tutorials/hdfs-logs/index-config.yaml), and running a query with 3 aggregrations (severity, tenant_id and date) I get a warmup cache in the 50-100MB range already. It feels like the 1GB/10 million docs can be easily exceeded on datasets a tiny bit more pathological (which is probably the case when OOMs happen currently).

quickwit/quickwit-directories/src/hot_directory.rs
```diff
@@ -205,14 +205,12 @@ impl StaticDirectoryCache {
...
 
-    /// return the files and their cached lengths
-    pub fn get_stats(&self) -> Vec<(PathBuf, usize)> {
+    pub fn get_file_lengths(&self) -> Vec<(PathBuf, u64)> {
         let mut entries = self
-            .slices
+            .file_lengths
             .iter()
-            .map(|(path, cache)| (path.to_owned(), cache.len()))
+            .map(|(k, v)| (k.clone(), *v))
```
fulmicoton, line 212:
```suggestion
            .map(|(k, v)| (k.clone(), *v))
```

can we give a concrete name to k and v?

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -124,33 +126,39 @@ pub(crate) async fn open_split_bundle(
...
-            index_storage,
-            storage_timeout_policy.clone(),
-        ))
-    } else {
-        index_storage
-    };
+    ephemeral_unbounded_cache: Option<ByteRangeCache>,
+) -> anyhow::Result<(Index, HotDirectory)> {
+    let index_storage_with_retry_on_timeout =
+        configure_storage_retries(searcher_context, index_storage);
```
fulmicoton, line 161: good idea.

fulmicoton (approved):

quickwit/quickwit-search/src/search_permit_provider.rs
rdettai: > What was the size of the split to get this warmup cache size?

Merged splits (~10 million docs) are ~750MB (with the store). A fast field with that many documents can be easily in the 50MB range if it has a poor compression ratio. I don't know the most extreme use cases for the other indexes.

PR #5572, opened 2024-12-05
Error when updating nested dynamic mapping to primitive explicit

rdettai:
### Description

Fix `invalid query: field does not exist: xxxx` when a field that used to be created dynamically and with nested fields is added to the doc mapping with a primitive type.

### How was this PR tested?


rdettai (commented):
Completes https://github.com/quickwit-oss/quickwit/issues/5302

rdettai:
The reproduction test isn't right because the query is not in lenient mode, so querying this field that does not exist is not supposed to work

PR #5574, opened 2024-12-07
make jaeger span attribute-to-tag conversion exhaustive

arlyon:
### Description

Right now, our quickwit span -> jaeger trace code doesn't support objects or null values and instead completely errors out which means that any trace where there is at least one span with an object with it does not work with Jaeger. I resolved this by adding support for Object and Null variants, encoding object keys delimited with `.`.

```
2024-12-07T11:42:07.655Z ERROR quickwit_serve::jaeger_api::rest_handler: failed to fetch traces error=Status { code: Internal, message: "Failed to serialize attributes: unexpected type `Object {}`", source: None }
```

https://github.com/quickwit-oss/quickwit/blob/53293792066aa7e5a0951d0be04fee94fc93cf0c/quickwit/quickwit-jaeger/src/lib.rs#L989

I opted to use iterools::Either because it was already available and avoids heap allocation except in the recursive 'flattening' case. I believe this will not impact performance of existing, functioning, use cases at all, in fact collect is shown to be faster in some cases, though I have not benchmarked it as it is a relatively cold part of the codebase. 

### How was this PR tested?

Docker build. Previously traces would error, now they don't.

### DIFF
--- a/quickwit/quickwit-jaeger/src/lib.rs
+++ b/quickwit/quickwit-jaeger/src/lib.rs
@@ -24,7 +24,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use async_trait::async_trait;
-use itertools::Itertools;
+use itertools::{Either, Itertools};
 use prost::Message;
 use prost_types::{Duration as WellKnownDuration, Timestamp as WellKnownTimestamp};
 use quickwit_config::JaegerConfig;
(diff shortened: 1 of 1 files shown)


arlyon:
Hello, any chance I can get CI approved at least? CLA signed.

guilload (commented):
Please update the test `test_otlp_attributes_to_jaeger_tags` accordingly.

guilload (approved):

PR #5577, opened 2024-12-09
Enable lenient mode for term set query and range query

rdettai:
### Description

This change is a follow up to https://github.com/quickwit-oss/quickwit/pull/5575 if we wanted to make it possible to have the same level of leniency as ES.

It makes range queries and term set queries potentially lenient.

### How was this PR tested?


trinity-1686a:
i don't think anyone uses this combination of features yet, but ig it makes sense to support it

Issue #5560, opened 2024-11-22
Existence queries unable to match nested object fields


esatterwhite:
**Describe the bug**
The field existence query doesn't work on nested field unless the field is terminal

**Steps to reproduce (if applicable)**

Indexing a document like the following

```json5
{
  "foo": {
    "bar": 1
  }
, "whiz": true
}
```

#### No Matches
```txl
foo:*
```

#### Matches 
```txl
foo:bar:*
```

#### Matches

```txl
whiz:*
```

**Expected behavior**
I would expect a document with a field of foo to match when using the existence query even if the field value is an object. That is the behavior seen by elasticsearch nested fields.

#### Version Information
```json5
{
  "build": {
    "build_date": "2024-10-29T15:18:16Z",
    "build_profile": "release",
    "build_target": "aarch64-unknown-linux-gnu",
    "cargo_pkg_version": "0.8.0",
    "commit_date": "2024-10-29T11:50:14Z",
    "commit_hash": "826f10f0d7ba1a448795890fe296941b84872c57",
    "commit_short_hash": "826f10f",
    "commit_tags": [
      "qw-airmail-20241029"
    ],
    "version": "0.8.0-nightly"
  },
  "runtime": {
    "num_cpus": 16,
    "num_threads_blocking": 14,
    "num_threads_non_blocking": 2
  }
}
```

fulmicoton:
@rdettai have you checked that it was working as described in elasticsearch? My understanding is that your plan si to implement in both 
- fast field
- indexed field,
correct?

rdettai:
> have you checked that it was working as described in elasticsearch?

yes (see [test](https://github.com/quickwit-oss/quickwit/pull/5581/files#diff-2b098a4f4073f28ea6f44774809068b84d54e5a8294f3a329b768c8ded5f07f4))

PR #5581, opened 2024-12-10
Fix existence queries for nested fields

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/5560

### How was this PR tested?

### DIFF
--- a/quickwit/rest-api-tests/scenarii/es_compatibility/0011-exists-query.yaml
+++ b/quickwit/rest-api-tests/scenarii/es_compatibility/0011-exists-query.yaml
@@ -25,6 +25,16 @@ expected:
     total:
       value: 60
 ---
+json:
+  query:
+    exists:
+      field: payload
+expected:
+  hits:
+    total:
+      # one of the docs contains `"payload":{}`
+      value: 99
+---
 # Fortunately, ES does not accept this quirky syntax in the
 # case of exists query.
 json:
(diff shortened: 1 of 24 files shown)


quickwit/rest-api-tests/scenarii/es_compatibility/0011-exists-query.yaml
```diff
@@ -25,6 +25,16 @@ expected:
...
       value: 60
 ---
+json:
+  query:
+    exists:
+      field: payload
+expected:
+  hits:
+    total:
+      # one of the docs contains `"payload":{}`
```
fulmicoton, line 35: did you run this test against elasticsearch too locally?

fulmicoton: in particular is `{}` considered not having the value? Is this true in elastic and in quickwit? Both for fast field and indexed field path?

rdettai: yes, I run this test against Elastic. That's how I got the 99 value and discovered that there is a `{}` and that it is considered has not having a value in Elastic

fulmicoton: I thnk you are the first dev to actually run those tests against elastic too :) thx

fulmicoton (approved):

Issue #5570, opened 2024-12-04
Test whether searchers work without control plane


rdettai:
**Is your feature request related to a problem? Please describe.**
We recently made some changes to enable searchers to work without control plane (https://github.com/quickwit-oss/quickwit/pull/5360). A few things should be validated though to ensure this works as expected:
- concerns regarding the correctness of the change https://github.com/quickwit-oss/quickwit/pull/5360#discussion_r1735870060
- the searchers seem to be configured with a metastore client that is proxied through the CP: https://github.com/quickwit-oss/quickwit/blob/c9f34df6507f6ac9ab5d66d667b186e8e1ede97a/quickwit/quickwit-serve/src/lib.rs#L604-L612

**Describe the solution you'd like**
Write an integration test to validate this.

rdettai:
cc @trinity-1686a

PR #5599, opened 2025-01-06
Fix and test clusters without control plane

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/5570

### How was this PR tested?

Integration tests

### DIFF
--- a/quickwit/quickwit-serve/src/lib.rs
+++ b/quickwit/quickwit-serve/src/lib.rs
@@ -493,7 +493,7 @@ pub async fn serve_quickwit(
     // Set up the "control plane proxy" for the metastore.
     let metastore_through_control_plane = MetastoreServiceClient::new(ControlPlaneMetastore::new(
         control_plane_client.clone(),
-        metastore_client.clone(),
+        metastore_client,
     ));
 
     // Setup ingest service v1.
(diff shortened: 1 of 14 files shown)


quickwit/quickwit-serve/src/lib.rs
```diff
@@ -604,7 +604,7 @@ pub async fn serve_quickwit(
     let (search_job_placer, search_service) = setup_searcher(
         &node_config,
         cluster.change_stream(),
-        metastore_through_control_plane.clone(),
+        metastore_client.clone(),
```
rdettai, line 607: this change is actually not _strictly_ necessary: even when using the proxied metastore client, the searchers don't get proxied because they only use metastore actions that go directly to the metastore (the proxied metastore client selectively decides which metastore queries to proxy or not)

fulmicoton:
@guilload can you review?

quickwit/quickwit-serve/src/lib.rs
guilload: Sounds good. Can you update the doc of `ControlPlaneMetastore` and state which services use a proxied metastore and which don't.

guilload (approved):

quickwit/quickwit-serve/src/lib.rs
rdettai: After a second thought, I actually reverted to using the `ControlPlaneMetastore` and just added a comment when providing ti to the the searcher initialization method. It doesn't make any difference today, but I prefer that the proxying rules are applied everywhere consistently, because an incomplete state in the control plane would be very hard to debug.

Issue #5417, opened 2024-09-12
Force commit might not work with v2 ingest CLI command


rdettai:
**Describe the bug**

Given that ingest V2 might send different requests to `api/v1/<index id>/ingest/ingest-v2` to different shards, it isn't guarantied that the various requests will be picked up by the same indexer or in order. But the `quickwit index ingest` CLI command breaks down its input data stream into batches and only assigns the `force` and `wait_for`  `commit-type` to the last batch.

**Steps to reproduce (if applicable)**
This is  pretty hard to reproduce, it might happen when ingesting large files (batch size limit is 10MB) on a busy multi-indexer cluster (otherwise it is unlikely for shards to be processed out of order)

**Expected behavior**
After returning, all documents processed by `quickwit index ingest` should be indexed and committed.

PR #5600, opened 2025-01-07
Use ingest v2 by default

guilload:


quickwit/quickwit-cli/src/lib.rs
rdettai, line 1: also remove https://github.com/quickwit-oss/quickwit/blob/65dbf92085f532451e22160a261491fc823703f9/quickwit/quickwit-cli/src/index.rs#L146-L151

quickwit/quickwit-integration-tests/src/tests/ingest_tests.rs
rdettai, line 1: nit, a small test with a cluster running with ingest V1 would be nice

quickwit/quickwit-control-plane/src/model/mod.rs
```diff
@@ -392,12 +402,157 @@ impl ControlPlaneModel {
         self.shard_table
             .release_scaling_permits(source_uid, scaling_mode)
     }
+
+    // Quickwit 0.9 uses the ingest v2 source by default. For indexes created prior to 0.9, we need
+    // to ensure that the ingest v2 source is created and enabled if necessary.
```
rdettai, line 407: I think this deserves a few lines in an [upgrade docs page](https://quickwit.io/docs/main-branch/operating/upgrades)

rdettai (commented):
It is probably the right time to adjust the docs as well, no? Whether we do it here or in a separate PR, here are some places where a doc update is necessary:
- [data dir structure](https://github.com/quickwit-oss/quickwit/blob/main/docs/operating/data-directory.md), add `wal/`
- [ingest API source](https://github.com/quickwit-oss/quickwit/blob/main/docs/ingest-data/ingest-api.md), a small section on the 2 ingest API versions would be very valuable I think
- [CLI docs](https://github.com/quickwit-oss/quickwit/blob/main/docs/reference/cli.md) add a ref to [this issue](https://github.com/quickwit-oss/quickwit/issues/5417) 
- [cluster sizing](https://github.com/quickwit-oss/quickwit/blob/main/docs/deployment/cluster-sizing.md), update `<!-- TODO: change this note when releasing ingest v2 -->`
- internal [doc](https://github.com/quickwit-oss/quickwit/blob/main/docs/internals/template-index.md) and [doc](https://github.com/quickwit-oss/quickwit/blob/main/docs/internals/ingest-v2.md)
- create a 0.9 [upgrade doc](https://github.com/quickwit-oss/quickwit/blob/main/docs/operating/upgrades.md)
- [metrics](https://github.com/quickwit-oss/quickwit/blob/main/docs/reference/metrics.md), but those are so outdated that it is a completely different task to update them

quickwit/quickwit-integration-tests/src/tests/ingest_tests.rs
guilload: Fair.

quickwit/quickwit-serve/src/ingest_api/rest_handler.rs
```diff
@@ -90,46 +105,90 @@ fn ingest_filter(
...
+    enable_ingest_v2: bool,
+) -> Result<IngestResponse, IngestServiceError> {
+    if enable_ingest_v2 {
+        return ingest_v2(index_id, body, ingest_options, ingest_router).await;
+    }
+    if !enable_ingest_v1 {
+        let message = "ingest v1 is disabled: environment variable `QW_DISABLE_INGEST_V1` is set";
+        return Err(IngestServiceError::Internal(message.to_string()));
+    }
+    ingest_v1(index_id, body, ingest_options, ingest_service).await
```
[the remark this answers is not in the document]
guilload: I like the sound of `QW_USE_LEGACY_INGEST` but I don't want to break anything for users who rely on `QW_DISABLE_INGEST_V1` or `QW_ENABLE_INGEST_V2`. Especially because both should go away in the 0.10 release.

rdettai (approved):

quickwit/quickwit-actors/src/actor_handle.rs
```diff
@@ -105,7 +105,7 @@ impl<A: Actor> Supervisable for ActorHandle<A> {
         {
             Health::Healthy
         } else {
-            error!(actor = self.name(), "actor-timeout");
+            error!(actor = self.name(), "actor timed out");
```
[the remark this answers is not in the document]
guilload: Reverted.

quickwit/quickwit-config/src/lib.rs
rdettai, line 1: After some thought, I think we should disable ingest V1 by default. My understanding is that leaving it enabled is only intended to purge the existing V1 WALs. But that's only useful when migrating from 0.8, and it will leave V1 running for all users  un-necessarily by default until the next release (which might be in a long time). If we document it in a migration guide and disable it by default:
- users upgrading from 0.8 will just have to set `QW_ENABLE_INGEST_V1=true` when first running 0.9
- new deployments will be optimal out of the box
- users who upgrade without reading the guide will loose there existing V1 WAL's worth of data, but apart from that everything will keep working just fine

fulmicoton (approved):

quickwit/quickwit-config/src/lib.rs
fulmicoton: What you are saying also makes sense, but let's stick to Adrien's solution here. Polling ingest v1  has very little impact. We can tell our users how to get rid of that source once it is consumed.

mamushe993 (approved):

Issue #5061, opened 2024-06-02
Support regex queries


fmassot:
This has been asked several times on Discord either explicitly or through better wildcard support such as `*abcdef` (suffix queries) or `abc*def`.

Currently, only prefix queries are supported.

vinhjaxt:
+1 really need this feature, please

recursionbane:
Regex support is super-important, and tantivy [already](https://docs.rs/tantivy/latest/tantivy/query/struct.RegexQuery.html) supports this.

guilload:
Hello,

We've been delaying this feature because achieving good performance for regex queries on object storage is actually non-trivial. tantivy regex queries are fast in memory or on SSD because its term dictionary is implemented using an FST, which is a great data structure for this purpose but requires performing a lot of small fetches from memory or disk. Quickwit uses a different term dictionary, which plays nicely with object storage but does not allow for efficient regex queries.

PR #5606, opened 2025-01-09
support arbitrary wildcard

### LINKED ISSUE #5306: support wildcard in non-suffix position
currently we allow `field:value*`, but not `field:val*ue`

trinity-1686a:
### Description

fix #5306 #5061
add support for arbitrary wildcard through running them as automatons
also add support for regexes

### How was this PR tested?

added some tests, and tested manually

### DIFF
--- a/quickwit/quickwit-query/src/query_ast/wildcard_query.rs
+++ b/quickwit/quickwit-query/src/query_ast/wildcard_query.rs
@@ -206,7 +206,7 @@ impl BuildTantivyAst for WildcardQuery {
             tantivy_fst::Regex::new(&regex).context("failed to parse regex built from wildcard")?;
         let regex_automaton_with_path = prefix::JsonPathPrefix {
             prefix: path.unwrap_or_default(),
-            automaton: regex,
+            automaton: regex.into(),
         };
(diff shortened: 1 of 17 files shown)


quickwit/quickwit-query/src/query_ast/wildcard_query.rs
```diff
@@ -44,72 +45,78 @@ impl From<WildcardQuery> for QueryAst {
...
-    let mut saw_wildcard = false;
-    let saw_wildcard = &mut saw_wildcard;
+fn sub_query_parts_to_regex(
+    sub_query_parts: Vec<SubQuery>,
+    tokenizer_name: &str,
+    tokenizer_manager: &TokenizerManager,
+) -> anyhow::Result<String> {
+    let mut normalizer = tokenizer_manager
+        .get_normalizer(tokenizer_name)
+        .with_context(|| format!("no tokenizer named `{}` is registered", tokenizer_name))?;
```
guilload, line 91:
```suggestion
        .with_context(|| format!("no tokenizer named `{tokenizer_name}` is registered"))?;
```

guilload (approved):
Awesome end-to-end job in Quickwit and tantivy 🎉

PR #5635, opened 2025-01-15
Detailed ingest response

rdettai:
### Description

Currently ingest v2 only reports docs that couldn't be parsed correctly in metrics. This is even worse than ingest v1 that was just generating logs.

### How was this PR tested?

Unit and integration tests.

### DIFF
--- a/quickwit/quickwit-serve/src/ingest_api/response.rs
+++ b/quickwit/quickwit-serve/src/ingest_api/response.rs
@@ -1,0 +1,258 @@
+// Copyright (C) 2024 Quickwit, Inc.
+//
+// Quickwit is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at hello@quickwit.io.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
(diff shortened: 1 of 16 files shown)


quickwit/quickwit-serve/src/ingest_api/response.rs
```diff
@@ -0,0 +1,258 @@
...
+                    reason: ParseFailureReason::from_i32(failure.reason).unwrap_or_default(),
+                });
+            }
+            resp.parse_failures = Some(parse_failures);
+        }
+        Ok(resp)
+    }
+}
+
+impl Add for RestIngestResponse {
```
guilload, line 108: Ok, this explains the `response + response` seen earlier. Isn't this more readable and explicit?

```rust
impl RestIngestResponse {
  /// Merges the `other` response into `self`.
  fn merge(&mut self, response: Self) {
    ...
  }
}
```

guilload (approved):

quickwit/quickwit-serve/src/ingest_api/response.rs
rdettai: agreed, I usually don't like overloading operators, I don't what's gotten into me :-\

rdettai:
@guilload FYI I added the `detailed-response` flag to the CLI

Issue #5634, opened 2025-01-15
Update sources from the API


rdettai:
**Is your feature request related to a problem? Please describe.**
Updating a source is currently not possible. It would be very valuable because some paramters like the `transform` or the `num_pipelines` might need to evolve through time.

**Describe the solution you'd like**
A REST endpoint 
```
PUT api/v1/indexes/<index id>/sources/<source id>
```

Like the index update, the body would be a full source config file (json or yaml):
- it would only be valid to update some specific fields: `num_pipelines`, `transform`, `params`
  - updating `params` is a footgun for some source types, but it's still valuable (e.g update kafka ssl settings)
- `source_id` and `source_type` cannot be updated.

**Describe alternatives you've considered**
Recreating a source with the new params is usually not a good idea because you would loose the checkpoints.

**Additional context**
NA

PR #5636, opened 2025-01-16
Add update source API

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/5634

### How was this PR tested?


quickwit/quickwit-proto/protos/quickwit/indexing.proto
```diff
@@ -47,7 +47,7 @@ message IndexingTask {
   // The shards assigned to the indexer.
   repeated quickwit.ingest.ShardId shard_ids = 3;
   // Fingerprint of the pipeline parameters. Anything that should cause a pipeline restart (such
-  // as updating indexing settings or doc mapping) should influence this value.
+  // as updating indexing settings, the doc mapping or the source) should influence this value.
   uint64 params_fingerprint = 6;
```
rdettai, line 51: It would probably be cleaner to create a dedicated type here to avoid all ambiguity. But it's not worth the trouble of updating the proto schema just for that.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -315,7 +324,16 @@ impl IndexingService {
...
-        let params_fingerprint = index_config.indexing_params_fingerprint();
+        let params_fingerprint = indexing_params_fingerprint(&index_config, &source_config);
+        if let Some(expected_params_fingerprint) = expected_params_fingerprint {
+            if params_fingerprint != expected_params_fingerprint {
+                warn!(
+                    expected_fingerprint = expected_params_fingerprint,
+                    actual_fingerprint = params_fingerprint,
+                    "params fingerprint mismatch"
+                );
+            }
```
rdettai, line 335: This will happen when the params got updated after the plan was calculated but before it was applied. We now the pipeline is very likely to be short lived in that case because it should be restarted by the control plane on the next recalculation. Should we take another action here other than warn?

quickwit/quickwit-control-plane/src/model/mod.rs
```diff
@@ -254,6 +253,20 @@ impl ControlPlaneModel {
         Ok(())
     }
 
+    pub(crate) fn update_source(
```
guilload, line 256: Unit test reminder.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
guilload: This is expected so I don't think it's necessary. I think it's better to log:

1. on the control plane:
- when we detected a index or source config change
- when we issue a new (changed) indexing plan

2. on indexers:
- when we received a new (changed) indexing plan
- when we spawn new pipelines
- when we shutdown pipelines

If we do that correctly, we should be able to observe correctly the impact of updating an index or source config.

Thoughts?

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -636,20 +654,14 @@ impl IndexingService {
     /// Identifies the pipelines to spawn and shutdown by comparing the scheduled plan with the
     /// current running plan.
     fn compute_pipeline_diff(&self, tasks: &[IndexingTask]) -> IndexingPipelineDiff {
-        let mut pipelines_to_spawn: Vec<IndexingPipelineId> = Vec::new();
+        let mut pipelines_to_spawn: Vec<IndexingTask> = Vec::new();
```
guilload, line 657: Why going from `IndexingPipelineId` to `IndexingTask` here?

quickwit/quickwit-metastore/src/metastore/index_metadata/mod.rs
```diff
@@ -155,6 +155,24 @@ impl IndexMetadata {
...
 
+    /// Adds a source to the index. Returns whether a mutation occurred and an
+    /// error if the source doesn't exist.
+    pub fn update_source(&mut self, source_config: SourceConfig) -> MetastoreResult<bool> {
+        match self.sources.entry(source_config.source_id.clone()) {
+            Entry::Occupied(mut entry) => {
+                if entry.get() == &source_config {
+                    return Ok(false);
+                }
+                entry.insert(source_config.clone());
```
guilload, line 166: Can we remove the `clone`? Or is the compiler not "smart enough"?

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
```diff
@@ -1080,6 +1082,21 @@ impl MetastoreService for PostgresqlMetastore {
         Ok(EmptyResponse {})
     }
 
+    #[instrument(skip(self))]
+    async fn update_source(&self, request: UpdateSourceRequest) -> MetastoreResult<EmptyResponse> {
+        let source_config = request.deserialize_source_config()?;
+        let index_uid: IndexUid = request.index_uid().clone();
+        run_with_tx!(self.connection_pool, tx, "add source", {
```
guilload, line 1089:
```suggestion
        run_with_tx!(self.connection_pool, tx, "update source", {
```

quickwit/quickwit-proto/protos/quickwit/metastore.proto
```diff
@@ -127,9 +127,12 @@ service MetastoreService {
   // Deletes splits.
   rpc DeleteSplits(DeleteSplitsRequest) returns (EmptyResponse);
 
-  // Adds source.
+  // Adds a source.
   rpc AddSource(AddSourceRequest) returns (EmptyResponse);
 
+  // Update a source.
```
guilload, line 133:
```suggestion
  // Updates a source.
```

quickwit/quickwit-proto/protos/quickwit/metastore.proto
```diff
@@ -127,9 +127,12 @@ service MetastoreService {
...
+  rpc UpdateSource(UpdateSourceRequest) returns (EmptyResponse);
+
   // Toggles source.
```
guilload, line 136:
```suggestion
  // Toggles (turns on or off) a source.
```

guilload (commented):
In good shape so far. I have yet to review the `quickwit-serve` part.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
rdettai: The goal is to make the fingerprint carried by the `IndexingTask` available when spawning the new pipeline and enable the `warn` [discussed here](https://github.com/quickwit-oss/quickwit/pull/5636#discussion_r1918737913).

The thing that made me inspect this part of the code is that I was confused that even though the RCP `IndexingTask` message carries the fingerprint, it isn't used anywhere in the indexer (it is used only locally to the control plane). Instead, we reload the config from the metastore and re-calculate a fingerprint. This leads to a bit of a transient inconsistency:
- the control plane detects a change from `fingerprint_a` to `fingerprint_b`, so it generates a new pipeline_uid (say `pipline_uid_b`) and sends an indexing task with `(fingerprint_b, pipline_uid_b)`
- the indexing service receives `(fingerprint_b, pipline_uid_b)`, detects that the current `pipline_uid_b` is not the current pipeline_uid so it restarts the pipeline. But the config might have changed in between, so when the indexer loads the new configuration from the metastore and recomputes the fingerprint, the new pipeline might be configured with `(fingerprint_c, pipline_uid_b)`. This is a bit surprising and I thought it deserved a warning (useful both for someone reading the logs and someone reading the code).
- soon after, the control plane detects the config change from `fingerprint_b` to `fingerprint_c`, so it generates a new pipeline_uid and sends out a new new task `(fingerprint_c, pipline_uid_c)`
- at that point, the indexer restarts the pipeline again, possibly with the same config as before.

quickwit/quickwit-config/src/index_config/mod.rs
```diff
@@ -262,13 +262,22 @@ pub struct IndexConfig {
...
+    /// fingerprint.
+    pub(crate) fn indexing_params_fingerprint(&self) -> u64 {
         let mut hasher = SipHasher::new();
         self.doc_mapping.doc_mapping_uid.hash(&mut hasher);
         self.indexing_settings.hash(&mut hasher);
         hasher.finish()
     }
 
+    /// Compare IndexConfig config level fingerprints
+    pub fn equals_fingerprint(&self, other: &Self) -> bool {
```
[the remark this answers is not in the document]
rdettai: The complete explanation of why we have this method here is carried by the crate method above https://github.com/quickwit-oss/quickwit/blob/6f7bcd267bb6040c25fb4c6a2ff5e07640aca35f/quickwit/quickwit-config/src/index_config/mod.rs#L264-L269

This is mostly a workaround for the lack of explicit typing for the fingerprints (they are just raw `u64`) that I raised in https://github.com/quickwit-oss/quickwit/pull/5636#discussion_r1918686344. If we had an explicit type for the fingerprint, like we do for `PipelineUid` or `ShardId`, it would help making it more explicit what the fingerprint really is and avoid potential confusion between the `IndexConfig` fingerprint and the pipeline (`IndexConfig`+`SourceConfig`) fingerprint.

docs/reference/rest-api.md
```diff
@@ -595,10 +595,58 @@ Create source by posting a source config JSON payload.
...
+| `source_id`     | `String` | Source ID, must be the same source as in the request URL.                                | _required_    |
+| `source_type`   | `String` | Source type: `kafka`, `kinesis` or `pulsar`. Cannot be updated.                          | _required_    |
+| `num_pipelines` | `usize`  | Number of running indexing pipelines per node for this source.                           | `1`           |
+| `transform`     | `object` | A [VRL](https://vector.dev/docs/reference/vrl/) transformation applied to incoming documents, as defined in [source config docs](../configuration/source-config.md#transform-parameters).                          | `null`         |
+| `params`        | `object` | Source parameters as defined in [source config docs](../configuration/source-config.md). | _required_    |
+
+:::warning
+
+While updating `num_pipelines` and `transform` is generally safe and reversible, updating `params` has consequences specific to the source type and might have side effects such as loosing the source's checkpoints. Perform such updates with great care.
```
rdettai, line 646: I wonder whether we should add a "per source type" update validation to avoid footguns. For instance changing the kafka topic is likely to break the source as the checkpoints will make no sense in the new topic.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -1200,7 +1221,9 @@ mod tests {
 
     #[tokio::test]
     async fn test_indexing_service_apply_plan() {
-        const PARAMS_FINGERPRINT: u64 = 3865067856550546352;
+        const PARAMS_FINGERPRINT_INGEST_API: u64 = 1637744865450232394;
+        const PARAMS_FINGERPRINT_SOURCE_1: u64 = 1705211905504908791;
+        const PARAMS_FINGERPRINT_SOURCE_2: u64 = 8706667372658059428;
```
rdettai, line 1226: the change of fingerprints will force all pipelines to restart on existing clusters

quickwit/quickwit-config/src/index_config/mod.rs
```diff
@@ -262,13 +262,22 @@ pub struct IndexConfig {
...
+    /// full indexing pipeline fingerprint that also includes the source's
+    /// fingerprint.
+    pub(crate) fn indexing_params_fingerprint(&self) -> u64 {
         let mut hasher = SipHasher::new();
         self.doc_mapping.doc_mapping_uid.hash(&mut hasher);
         self.indexing_settings.hash(&mut hasher);
         hasher.finish()
     }
 
+    /// Compare IndexConfig config level fingerprints
```
guilload, line 276:
```suggestion
    /// Compares IndexConfig config level fingerprints.
```

quickwit/quickwit-config/src/index_config/mod.rs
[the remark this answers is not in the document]
guilload: Ok, I understand your intent now. I'm still not super convinced, but why not.

quickwit/quickwit-config/src/lib.rs
```diff
@@ -286,6 +288,17 @@ pub trait TestableForRegression: Serialize + DeserializeOwned {
     fn assert_equality(&self, other: &Self);
 }
 
+/// Return a fingerprint of all parameters that should trigger an indexing pipeline restart.
```
guilload, line 291:
```suggestion
/// Returns a fingerprint (a hash) of all the parameters that should trigger an indexing pipeline to restart upon index or source config updates.
```

quickwit/quickwit-config/src/source_config/mod.rs
```diff
@@ -143,6 +145,20 @@ impl SourceConfig {
         }
     }
 
+    /// Return a fingerprint of parameters relevant for indexers
```
guilload, line 148:
```suggestion
    /// Returns a fingerprint of parameters relevant for indexers.
```

quickwit/quickwit-index-management/src/index.rs
```diff
@@ -481,6 +481,40 @@ impl IndexService {
...
+    ) -> Result<SourceConfig, IndexServiceError> {
+        let source_id = source_config.source_id.clone();
+        check_source_connectivity(&self.storage_resolver, &source_config)
+            .await
+            .map_err(IndexServiceError::InvalidConfig)?;
+        let update_source_request =
+            UpdateSourceRequest::try_from_source_config(index_uid.clone(), &source_config)?;
+        self.metastore.update_source(update_source_request).await?;
+        info!(
+            "source `{}` successfully updated for index `{}`",
```
guilload, line 498:
```suggestion
            "source `{source_id}` successfully updated for index `{}`",
```

quickwit/quickwit-index-management/src/index.rs
```diff
@@ -481,6 +481,40 @@ impl IndexService {
...
+        index_uid: IndexUid,
+        source_config: SourceConfig,
+    ) -> Result<SourceConfig, IndexServiceError> {
+        let source_id = source_config.source_id.clone();
+        check_source_connectivity(&self.storage_resolver, &source_config)
+            .await
+            .map_err(IndexServiceError::InvalidConfig)?;
+        let update_source_request =
+            UpdateSourceRequest::try_from_source_config(index_uid.clone(), &source_config)?;
+        self.metastore.update_source(update_source_request).await?;
```
guilload, line 496: nit: The `AddSource` and `UpdateSource` source should return the added/updated source. I suspect I forgot to do so for `AddSource`. No need to tackle that in this PR, though.

quickwit/quickwit-index-management/src/index.rs
```diff
@@ -481,6 +481,40 @@ impl IndexService {
...
+            source_id, index_uid.index_id,
+        );
+        let index_metadata_request = IndexMetadataRequest::for_index_id(index_uid.index_id);
+        let source = self
+            .metastore
+            .index_metadata(index_metadata_request)
+            .await?
+            .deserialize_index_metadata()?
+            .sources
+            .get(&source_id)
```
guilload, line 508: nit: Calling `expect` would also be reasonable here. The gRPC/REST server should catch the panic and return an internal error.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
guilload: Ok, good point. Then I think we should not spawn the pipeline when that happens and wait for the next plan sent by the cp.

quickwit/quickwit-indexing/src/actors/indexing_service.rs
rdettai: updated the migration guide to recommend scaling down indexers

guilload (approved):

quickwit/quickwit-indexing/Cargo.toml
```diff
@@ -47,6 +47,7 @@ tracing = { workspace = true }
 ulid = { workspace = true }
 utoipa = { workspace = true }
 vrl = { workspace = true, optional = true }
+warp = { workspace = true, optional = true }
```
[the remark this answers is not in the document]
rdettai: Unfortunately no. It is re-exposed by a feature.

PR #5644, opened 2025-01-23
Retry 503 in CLI

rdettai:
### Description

When ingesting the hdfs dataset the ingest CLI fails with 503. This PR adds a retry when that happens to make the CLI more robust (and make the hdfs tutorial work :-) )

### How was this PR tested?

Manual test with the gzipped hdfs dataset + integration test


rdettai:
https://github.com/quickwit-oss/quickwit/pull/5651 fixes transformed a lot of 503 back to 429, so retrying on 503 is probably not useful anymore. The tests, docs and more verbose output for the ingest CLI are still relevant.

PR #5651, opened 2025-01-26
Return 429 instead of 503 when all the shards of an index are rate limited

guilload:
### Description
Per title.

### How was this PR tested?
Added unit test


rdettai (approved):
You might want to revert this: https://github.com/quickwit-oss/quickwit/pull/5566

Issue #5270, opened 2024-07-27
Increase ramp up of ingest through quickwit cli


fulmicoton:
Right now the creation of the shard is a bit slow.

sahilarora535:
Hi @fulmicoton and team, thank you for the great work on Quickwit. Are there any ETAs when can we expect to see 0.9? I've been tracking the project, and it looks like the last piece for 0.9 as per https://github.com/orgs/quickwit-oss/projects/24. Thank you.

rdettai:
https://github.com/quickwit-oss/quickwit/pull/5679 starts addressing this issue. It is not a silver bullet, you can still expect a bit of throttling because the scaling decisions take 5-10secs to be taken. I don't think any other action on this front before 0.9. We just have to also get a few other fixes out and perform some extra tests and we should be good to go!

PR #5655, opened 2025-01-28
Add number of throttled query in CLI response

rdettai:
### Description

This PR is a reuse of the tests and docs proposed in https://github.com/quickwit-oss/quickwit/pull/5644, which itself is not necessary anymore after the the status code was fixed to be 429 when shards need scaling up (https://github.com/quickwit-oss/quickwit/pull/5651).

It adds a small indication of the number of retries that occurred to the CLI ingest command. This is handy for troubleshooting and shows concretely to users that retries are often necessary.

Related to https://github.com/quickwit-oss/quickwit/issues/5270

### How was this PR tested?

Integration tests and running the CLI ingest command on the [HDFS dataset](https://quickwit.io/docs/get-started/tutorials/tutorial-hdfs-logs).

### DIFF
--- a/quickwit/quickwit-integration-tests/src/tests/ingest_v2_tests.rs
+++ b/quickwit/quickwit-integration-tests/src/tests/ingest_v2_tests.rs
@@ -359,14 +359,20 @@ async fn test_ingest_v2_high_throughput() {
     let line = json!({"body": "my dummy repeated payload"}).to_string();
     let num_docs = body_size / line.len();
     let body = std::iter::repeat_n(&line, num_docs).join("\n");
-    let ingest_resp = ingest(
-        &sandbox.rest_client(QuickwitService::Indexer),
-        index_id,
-        IngestSource::Str(body),
-        CommitType::Auto,
-    )
-    .await
-    .unwrap();
+    let ingest_resp = sandbox
+        .rest_client(QuickwitService::Indexer)
+        .ingest(
+            index_id,
+            IngestSource::Str(body),
+            // TODO: when using the default 10MiB batch size, we get persist
+            // timeouts with code 500 on some lower performance machines (e.g.
+            // Github runners). We should investigate why this happens exactly.
+            Some(5_000_000),
+            None,
+            CommitType::Auto,
+        )
+        .await
+        .unwrap();
     assert_eq!(ingest_resp.num_docs_for_processing, num_docs as u64);
     assert_eq!(ingest_resp.num_ingested_docs, Some(num_docs as u64));
     assert_eq!(ingest_resp.num_rejected_docs, Some(0));
--- a/quickwit/quickwit-integration-tests/src/tests/ingest_v2_tests.rs
+++ b/quickwit/quickwit-integration-tests/src/tests/ingest_v2_tests.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use futures_util::FutureExt;
 use itertools::Itertools;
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-rest-client/src/models.rs
```diff
@@ -78,6 +79,43 @@ pub enum IngestSource {
...
+            }),
+            num_rejected_docs: apply_op(self.num_rejected_docs, other.num_rejected_docs, |a, b| {
+                a + b
+            }),
+            parse_failures: apply_op(self.parse_failures, other.parse_failures, |a, b| {
+                a.into_iter().chain(b).collect()
+            }),
+            num_too_many_requests: self.num_too_many_requests,
+        }
+    }
```
rdettai, line 107: I moved this back here as it makes more sense than in the API model because accumulating responses is quite specific to the rest client.

quickwit/quickwit-integration-tests/src/tests/ingest_v2_tests.rs
```diff
@@ -359,14 +359,17 @@ async fn test_ingest_v2_high_throughput() {
...
-        CommitType::Auto,
-    )
-    .await
-    .unwrap();
+    let ingest_resp = sandbox
+        .rest_client(QuickwitService::Indexer)
+        .ingest(
+            index_id,
+            IngestSource::Str(body),
+            Some(5_000_000),
```
esatterwhite, line 367: Casually observing. 

Our batches tend to have 200 - 1000 items during bulk request and target 1 - 10 different indexes. The higher end is about 3mb. Here the typical rate of ingest is about 900mb/s 

Our retry work load is capped to 500 items per batch and sits around 1mb. These would appear to touch 1 index. We are trying to determine if they are all the same index

rdettai: I'm playing with the batch size in this test because I had a surprising 500 (internal timeout) error. We want the system to behave consistently regardless of the batch size (obviously with very small batches adding a bit of overhead).

esatterwhite: What do we consider very small?

rdettai: One overhead of the top of my head is the http frame overhead. If you have more http headers than document payload, that's a bit of unnecessary extra work.

rdettai: the 1MB granularity should be pretty good (negligeable framing overhead in network exchanges but small enough to be nicely balanced between shards)

rdettai:
TODO BEFORE MERGING: figure out why the high throughput test fails with a big payload but not with smaller ones

quickwit/quickwit-integration-tests/src/tests/ingest_v2_tests.rs
```diff
@@ -332,6 +332,83 @@ async fn test_ingest_v2_happy_path() {
...
+    let body = std::iter::repeat_n(&line, num_docs).join("\n");
+    let ingest_resp = sandbox
+        .rest_client(QuickwitService::Indexer)
+        .ingest(
+            index_id,
+            IngestSource::Str(body),
+            // TODO: when using the default 10MiB batch size, we get persist
+            // timeouts with code 500 on some lower performance machines (e.g.
+            // Github runners). We should investigate why this happens exactly.
+            Some(5_000_000),
```
rdettai, line 370: @guilload I didn't find a good explanation for why this timeout occur here in the persist https://github.com/quickwit-oss/quickwit/blob/ce4501fc9cde184273a17a09cac1bc6cf55173b4/quickwit/quickwit-ingest/src/ingest_v2/router.rs#L432-L443

Persisting 10MB should not take 6 sec, even on a slow system, should it?

PR #5658, opened 2025-01-29
Restore rate limit metric distinction

rdettai:
### Description

In https://github.com/quickwit-oss/quickwit/pull/5651 errors where all shards are rate limited were re-qualified from no_shards_error (503) to a rate_limiting_error (429). But rate limiting errors also cover the situation where we unfortunately tried to write to rate limited shards during all 5 workbench retries. We actually need to keep the distinction between the two for troubleshooting reasons, especially in the metrics.

### How was this PR tested?


rdettai:
Moving this back to draft, because this distinction is actually not bulletproof. Having made 5 persist attempts that ended up on rate limited shards we can still not conclude whether all shards were limited or not. It's still very much function of the number of shards.

PR #5661, opened 2025-01-30
Reduce aggregation result footprint

rdettai:
### Description

Profiling showed that parsing the json serialized aggregation string returned by the root searcher into a `serde_json::Value` can take 5x time the memory size of the returned payload. On a loaded system, this can represent a lot of memory that is consumed even if results are served from cache.

By using `serde_json_borrowed::OwnedValue`, we can reduce this tp 3x the size of the returned payload.

### How was this PR tested?

Heaptrack profiling.

### DIFF
--- a/quickwit/quickwit-rest-client/src/models.rs
+++ b/quickwit/quickwit-rest-client/src/models.rs
@@ -17,6 +17,8 @@ use std::time::Duration;
 
 use reqwest::StatusCode;
 use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
+use serde_json::Value as JsonValue;
 
 use crate::error::{ApiError, Error, ErrorResponsePayload};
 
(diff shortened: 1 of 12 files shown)


quickwit/quickwit-rest-client/src/models.rs
```diff
@@ -71,6 +73,20 @@ impl ApiResponse {
     }
 }
 
+/// A cousin of [`quickwit_search::SearchResponseRest`] that implements [`Deserialize`]
+///
+/// This version of the response is necessary because
+/// `serde_json_borrow::OwnedValue` is not deserializeable.
```
fulmicoton, line 79: ah we can ask @PSeitz to make it Deserializable.

fulmicoton (approved):

quickwit/quickwit-rest-client/src/models.rs
PSeitz: `OwnedValue` is a little bit special, because it needs to to take ownership of the original `String`

fulmicoton: couldn't it be possible to still make it deserializable by keeping a state in the deserializer, and building a (non-original) `String`

PSeitz: Yes, that should be possible, but it would require to have a special deserialization logic. Currently it's just a thin wrapper

rdettai:
TODO BEFORE MERGING: add the same thing on the ES API search path

rdettai:
@fulmicoton I made some changes to also use the same borrowed deserialization on the ES API code path. If you could take a look that would be great!

Issue #5664, opened 2025-02-04
Add `drop_on_abort` option to VRL processing


cometkim:
**Is your feature request related to a problem? Please describe.**

As reported on the Discord https://discord.com/channels/908281611840282624/915644818988863558/1335798024382971924

Quickwit always record a warn log on explicitly aborted transform with `abort` expression. But users may want to drop it.

**Describe the solution you'd like**

Originally there is `drop_on_abort` (default: true) option in Vector configuration.

https://vector.dev/highlights/2021-04-21-vrl-abort/

I'd like to suggest the same config.

**Describe alternatives you've considered**

User could set `RUST_LOG` to such as `quickwit_indexing::actors::vrl_processing=error` to avoid the warning.

But it's far from ideal solution because it'll not inform user what is real problems or intentionally aborted ones.

fulmicoton-dd:
thank you for the clear report.

PR #5665, opened 2025-02-04
Add drop_on_abort option to transform config

cometkim:
### Description

See #5664 

### How was this PR tested?

TBD, should verify aborted logs will not be counted


cometkim:
I can add the [`drop_on_error`](https://vector.dev/docs/reference/configuration/transforms/remap/#drop_on_error) option too, might be useful in some cases.

PR #5672, opened 2025-02-07
Output logs in JSON with `QW_LOG_FORMAT=json`

guilload:
### Description
Per title.

### How was this PR tested?
Ran locally and inspected logs manually

### DIFF
--- a/quickwit/quickwit-cli/src/logger.rs
+++ b/quickwit/quickwit-cli/src/logger.rs
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::env;
 use std::sync::Arc;
+use std::{env, fmt};
 
 use anyhow::Context;
 use opentelemetry::trace::TracerProvider;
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-cli/src/logger.rs
```diff
@@ -116,3 +119,75 @@ pub fn setup_logging_and_tracing(
         Ok(())
     }))
 }
+
+enum EventFormat<'a> {
+    Full(Format<Full, UtcTime<Vec<BorrowedFormatItem<'a>>>>),
+    Json(tracing_subscriber::fmt::format::Format<tracing_subscriber::fmt::format::Json>),
```
rdettai, line 125:
```suggestion
    Json(Format<tracing_subscriber::fmt::format::Json>),
```

quickwit/quickwit-cli/src/logger.rs
```diff
@@ -116,3 +119,75 @@ pub fn setup_logging_and_tracing(
...
+}
```
rdettai, line 126: Is there any strong reason for using a custom event format rather than a [boxed layer](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/trait.Layer.html#method.boxed)? Maybe from a performance point of view enum is a tiny bit better, but boxing would be substantially simpler.
https://github.com/quickwit-oss/quickwit/blob/f9466262c559220cbef0c172cef07bc79cad731d/quickwit/quickwit-lambda/src/logger.rs#L92-L102

guilload: I assumed we didn't go for a boxed implementation in the first place for a good reason, so I also added unboxed code. I'm open to a refactor that uses the same code in `cli` and `lambda` and potentially uses boxed layers.

PR #5679, opened 2025-02-13
Faster shard scaling

rdettai:
### Description

Indexers return a lot of 429 when ingest rates fluctuate, even if the overall indexing capacity (~5MB/core) is note reached. This PR makes some small parameter adjustments to decrease the overall error rates on typical workloads. It will be the topic for another issue/PR to introduce the structual changes that are necessary to improve the routing strategy.

- Customizable shard burst limit with default increased to 50MB: helps absorbing the load until the next scale up
- Shard scaling factor: helps keeping up with load fluctuations on workloads with high number of shards (e.g having +20MB/s on an index where the average ingest rate is 100MB/s is more likely than +20MB/s on an index where the average rate is 1MB/s)

Related to https://github.com/quickwit-oss/quickwit/issues/5270

### How was this PR tested?

Adhoc python load testing framework: https://github.com/rdettai/qw-ingest-tests

### DIFF
--- a/quickwit/quickwit-config/src/cluster_config/mod.rs
+++ b/quickwit/quickwit-config/src/cluster_config/mod.rs
@@ -24,6 +24,7 @@ pub struct ClusterConfig {
     pub default_index_root_uri: Uri,
     pub replication_factor: usize,
     pub shard_throughput_limit: ByteSize,
+    pub shard_scaling_factor: f32,
 }
 
 impl ClusterConfig {
@@ -35,6 +36,7 @@ impl ClusterConfig {
             default_index_root_uri: Uri::for_test("ram:///indexes"),
             replication_factor: 1,
             shard_throughput_limit: quickwit_common::shared_consts::DEFAULT_SHARD_THROUGHPUT_LIMIT,
+            shard_scaling_factor: 1.01,
         }
     }
 }
--- a/quickwit/quickwit-config/src/cluster_config/mod.rs
+++ b/quickwit/quickwit-config/src/cluster_config/mod.rs
@@ -24,7 +24,7 @@ pub struct ClusterConfig {
     pub default_index_root_uri: Uri,
     pub replication_factor: usize,
     pub shard_throughput_limit: ByteSize,
-    pub shard_scaling_factor: f32,
+    pub shard_scale_up_factor: f32,
 }
 
 impl ClusterConfig {
@@ -36,7 +36,7 @@ impl ClusterConfig {
             default_index_root_uri: Uri::for_test("ram:///indexes"),
             replication_factor: 1,
             shard_throughput_limit: quickwit_common::shared_consts::DEFAULT_SHARD_THROUGHPUT_LIMIT,
-            shard_scaling_factor: 1.01,
+            shard_scale_up_factor: 1.01,
         }
     }
 }
--- a/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
+++ b/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
@@ -285,6 +285,7 @@ impl IngestController {
         ingester_pool: IngesterPool,
         replication_factor: usize,
         max_shard_ingestion_throughput_mib_per_sec: f32,
+        shard_scaling_factor: f32,
     ) -> Self {
         IngestController {
             metastore,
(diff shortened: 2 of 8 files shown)


rdettai:
Observations so far:
- there is a high latency between a scaling action and the entry being added to the router (5+s)
- even with a high `shard_burst_limit` (100MB), a fairly low ingest rate (20MB/s) runs into throttling before the shards are scaled an the routers know about the new shards
- TODO: check when the extra shard is taken into account for the average shard throughput calculation
  - if it is already accounted for while it isn't known by routers, that's very bad because it means the arbiter will think the load has lowered when we are actually rejecting some ingest that could have gone to the un-registered shards
  - ideally, it should be only accounted for once it has been broadcasted (even though that wouldn't work well if we add other shard knowledge propagation techniques in the future)

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -670,18 +673,19 @@ impl IngestController {
         shard_stats: ShardStats,
         model: &mut ControlPlaneModel,
         progress: &Progress,
+        shards_to_create: usize,
```
guilload, line 676:
```suggestion
        num_shards_to_open: usize,
```

guilload: nit

quickwit/quickwit-config/src/cluster_config/mod.rs
```diff
@@ -24,6 +24,7 @@ pub struct ClusterConfig {
     pub default_index_root_uri: Uri,
     pub replication_factor: usize,
     pub shard_throughput_limit: ByteSize,
+    pub shard_scaling_factor: f32,
```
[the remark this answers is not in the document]
rdettai: I applied the renaming as it is more explicit and future proof. For the scale down

I would rather be conservative on the scale down for now because removing shards might reset indexing pipelines and removing multiple shards at once is likely to impact multiple pipelines. I think we should observe on real world workloads if speeding up scale down is really worth it.

quickwit/quickwit-control-plane/src/ingest/scaling_arbiter.rs
```diff
@@ -29,38 +30,67 @@ pub(crate) struct ScalingArbiter {
...
+            shard_scale_up_factor,
         }
     }
 
+    /// Computes the maximum number of shards we can have without going below
+    /// the long term scale up threshold
+    fn long_term_scale_up_threshold_max_shards(&self, shard_stats: ShardStats) -> usize {
+        (shard_stats.avg_long_term_ingestion_rate * shard_stats.num_open_shards as f32
+            / self.scale_up_shards_long_term_threshold_mib_per_sec)
+            .floor() as usize
```
fulmicoton-dd, line 57: I'm not sure about the floor?

quickwit/quickwit-control-plane/src/ingest/scaling_arbiter.rs
```diff
@@ -29,38 +30,67 @@ pub(crate) struct ScalingArbiter {
...
             >= self.scale_up_shards_short_term_threshold_mib_per_sec
         {
-            let long_term_ingestion_rate_after_scale_up = shard_stats.avg_long_term_ingestion_rate
-                * (shard_stats.num_open_shards as f32)
-                / (shard_stats.num_open_shards as f32 + 1.0f32);
-            if long_term_ingestion_rate_after_scale_up
-                >= self.scale_up_shards_long_term_threshold_mib_per_sec
-            {
-                return Some(ScalingMode::Up);
+            let new_calculated_num_shards = usize::min(
```
fulmicoton-dd, line 78: nitpick: the symmetry is actually a tad misleading in the call to min here.
My understanding is that we want scale_up_factor_target_shards to curb the value obtained by long_term_scale_up.
`a.min(b)` is a tad better maybe?

fulmicoton-dd (approved):

quickwit/quickwit-control-plane/src/ingest/scaling_arbiter.rs
rdettai: If it was a `ceil`, it would mean that scaling up by this number of shards would get us slightly below the `scale_up_shards_long_term_threshold_mib_per_sec`. For small shard counts it might get us a bit to close to `scale_down_shards_threshold_mib_per_sec`

quickwit/quickwit-control-plane/src/ingest/scaling_arbiter.rs
rdettai: I naturally wouldn't really read `a.min(b)` as "a curbed by b", but it makes sense. I prefer this writing because with these awfully long variable names the `.min` kind of gets lost in the line wrappings

Issue #5682, opened 2025-02-13
dynamic mode default tokenizer does not align with documentation


fulmicoton-dd:
In https://github.com/quickwit-oss/quickwit/pull/5653 we fixed the default value of the tokenizer of the dynamic mapping as it was not aligned with the documentation.

We need to revert that change and resend a PR.

Some companies (cutomer, but also OSS users) are relyign on `qw-image` that do pull code from 0.9, so we need to be very careful about change of behavior.

Versioning is here to allow that kind fo change. We can add a 0.9.1 version that changes the default value for instance.
Other solutions are ok too of course.

Also, the PR gives the right link to the documentation, but does not explain when the change was introduced. Has it always been like this?
Did we make that change by mistake or on purpose? There is a possibility that this change was on purpose and the correct fix is a documentation fix.

PR #5684, opened 2025-02-17
Fix the docs for json and dynamic fields

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/5682

### How was this PR tested?


docs/configuration/index-config.md
```diff
@@ -497,7 +497,7 @@ doc_mapping:
   dynamic_mapping:
     indexed: true
     stored: true
-    tokenizer: default
+    tokenizer: raw
     record: basic
     expand_dots: true
     fast: true
```
trinity-1686a, line 503: i don't think this is right given other changes
```suggestion
    fast: false
```

trinity-1686a (approved):

fulmicoton-dd (approved):

PR #5690, opened 2025-02-20
Fix delete task not respecting default fields

rdettai:
### Description

This seems to have been broken in https://github.com/quickwit-oss/quickwit/pull/3148

According to the docs, delete tasks should support default fields https://github.com/quickwit-oss/quickwit/blob/722cb877a611972f29167b11191aa0f6acb9c213/docs/reference/rest-api.md#L759

### How was this PR tested?

Unit tests.

### DIFF
--- a/quickwit/quickwit-serve/src/delete_task_api/handler.rs
+++ b/quickwit/quickwit-serve/src/delete_task_api/handler.rs
@@ -205,11 +205,33 @@ mod tests {
         let metastore = test_sandbox.metastore();
         let delete_query_api_handlers =
             super::delete_task_api_handlers(metastore).recover(recover_fn);
+
+        // POST a delete query
         let resp = warp::test::request()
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-serve/src/delete_task_api/handler.rs
```diff
@@ -45,8 +45,9 @@ pub struct DeleteQueryRequest {
     /// Query text. The query language is that of tantivy.
     pub query: String,
     // Fields to search on
+    #[serde(rename(deserialize = "search_field"))]
```
rdettai, line 48: this is necessary to be consistent with the docs and the search API https://github.com/quickwit-oss/quickwit/blob/ce3ae7769932f3efbe84970a44073966fcc6fd88/quickwit/quickwit-serve/src/search_api/rest_handler.rs#L190-L197

fulmicoton (approved):

PR #5711, opened 2025-03-11
Fix index update endpoint panic when index_uri is the bucket root

rdettai:
### Description

Users can set the the index_uri to be the root of a bucket, and this panics when calling the index update endpoint. This PR:
- replaces the panic with an Internal Server error
- makes it possible to update an index even if the index_uri is the root of a bucket

### How was this PR tested?

### DIFF
--- a/docs/reference/rest-api.md
+++ b/docs/reference/rest-api.md
@@ -348,7 +348,7 @@ Updating the doc mapping doesn't reindex existing data. Queries and results are 
 |---------------------|--------------------|-----------------------------------------------------------------------------------------------------------------------|---------------------------------------|
 | `version`           | `String`           | Config format version, use the same as your Quickwit version.                                                         | _required_                            |
 | `index_id`          | `String`           | Index ID, must be the same index as in the request URI.                                                               | _required_                            |
-| `index_uri`         | `String`           | Defines where the index files are stored. Cannot be updated.                                                          | `{current_index_uri}`                 |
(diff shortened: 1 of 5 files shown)


guilload:
The benefit of your fix only lies in not panicking in `load_index_config_update`. The new `Uri::parent_unchecked` function is only used in `update_index` in `quickwit-serve` and its logic is irrelevant. In your unit test, the value returned by `parent_unchecked` is ignored because the update carries an index URI. If you remove it,  `parent_unchecked` returns `s3://` to which the index ID is joined. Now we have `s3://<index ID>` instead of `s3://<bucket name>` for the index URI and we still have an error. So `parent_unchecked` is useless.

The best fix I can come up with is to deserialize the new config as a `JsonValue`, and extract `index_uri` from it. If it exists, verify that we have `current_index_config.index_uri == new_index_config.index_uri,` if not override `new_index_config.index_uri` with the old one, passing a dummy `default_index_root_uri` to `load_index_config_from_user_config` which will be ignored.

rdettai:
Thanks for your comment. I think the root of the problem here is that I initially over-complexified the behavior of the update endpoint regarding the `index_uri`. I wanted to avoid that an update of the `default_index_root_uri` breaks updates that use that default. I think it was over engineering. I propose to simplify that to let the update endpoint use the same default URI as the create endpoint. 

It think it's fine to make this change as the update endpoint isn't released yet and anyway, this has no consequence unless the `default_index_root_uri` root URI is used along with the update endpoint.

guilload (approved):

PR #5715, opened 2025-03-17
non-ambiguous internal aggregations

trinity-1686a:
use a non-ambiguous format for aggregation results internally

the api is unchanged, but it makes it easier to manipulate results from inside quickwit

before that, trying to re-parse the aggregation results (json) may lead to a different ast than expected, which made it hard to apply any transformation without also traversing the aggregation request ast at the same time


quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -289,8 +289,11 @@ message SearchResponse {
   // The searcherrors that occurred formatted as string.
   repeated string errors = 4;
 
-  // Serialized aggregation response
-  optional string aggregation = 5;
+  // used to be json-encoded aggregation
+  reserved 5;
+
+  // Postcard-encoded aggregation response
+  optional bytes aggregation_postcard = 9;
```
rdettai, line 296: this is not retro-compatible, right? it means we cannot perform a rolling upgrade of searchers with this change

quickwit/quickwit-search/src/search_response_rest.rs
```diff
@@ -13,28 +13,27 @@
...
 use crate::error::SearchError;
 
-/// A lightweight serializable representation of aggregation results.
-///
-/// We use `serde_json_borrow` here to avoid unnecessary
-/// allocations. On large aggregation results with tens of thousands of
-/// entries this has a significant impact compared to `serde_json`.
+/// A classic ES aggregation result ast
+// TODO previously, we were using zero-copy when possible, which we are no longer doing:
+// is that problematic? How can we return to zero/low-copy without it being painful?
```
rdettai, line 28: The reason I switched to `serde_json_borrow` was a very large allocation for this intermediate representation ([up to 5x the actual response payload](https://github.com/quickwit-oss/quickwit/pull/5661)). I assume postcard should be a lot better than `serde_json::Value`, even if it's not borrowing. A quick check with heaptrack should confirm this.

quickwit/quickwit-query/src/aggregations.rs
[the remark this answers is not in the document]
trinity-1686a: postcard also doesn't like `skip_serialize_if`, but yeah, that's the main reason

when parsing back something like a [`BucketResult`](https://docs.rs/tantivy/latest/tantivy/aggregation/agg_result/enum.BucketResult.html), you end up parsing the wrong variant most of the time. That's an issue when converting the aggregation result to anything that's not strictly an ES lookalike

quickwit/quickwit-proto/protos/quickwit/search.proto
rdettai: > you need the node which serves the rest api and the root-searcher for the request to be the same

From what I read in the code, this is always the case: the REST API uses `SearchService` that only has one concrete implementation (`SearchServiceImpl`).

Now, this raises a new question. We could also solve this by getting rid of the serialization entirely. If the `SearchService` trait was never re-implemented so far, we can assume it was a premature extra abstraction layer.

quickwit/quickwit-query/src/aggregations.rs
```diff
@@ -0,0 +1,410 @@
...
+}
+
+/// This is the wrapper of buckets entries, which can be vector or hashmap
+/// depending on if it's keyed or not.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum BucketEntries<T> {
+    /// Vector format bucket entries
+    Vec(Vec<T>),
+    /// HashMap format bucket entries
+    HashMap(FxHashMap<String, T>),
```
fulmicoton-dd, line 225: Could it have been `Vec<(String, T)>`?

quickwit/quickwit-search/src/search_response_rest.rs
fulmicoton-dd: the largest footprint was probably the BTreeMap in serde_json. Here at least we are using hashmap (which I believe could be `Vec<(K,V)>` or something more frugal).

fulmicoton-dd (approved):

quickwit/quickwit-proto/protos/quickwit/search.proto
fulmicoton-dd: let's keep that abstraction

quickwit/quickwit-query/src/aggregations.rs
trinity-1686a: i was about to say that we'd need to use something like [`serde_with::Map`](https://docs.rs/serde_with/latest/serde_with/struct.Map.html), but actually, we could just switch all hashmaps to Vec of tuples, this is our own format, and as long as we can convert losslessly from/to tantivy aggregations, we can do anything we want, including not storing things in the most evident format

PR #5719, opened 2025-03-20
Failure to create an index with ingest v2 returns 429

rdettai:
### Description of the issue

When using index templates, specifying and index name that matches the pattern but has illegal characters results in a 429 response code instead of a 400.

Description of the problem:
- the index_id validity is not validated until the metastore is called
- calling `GetOrCreateOpenShards` with an invalid index_id on the metastore fails with MetastoreError::JsonDeserializeError
- a failure to call the metastore is logged but not repported to the ingest workbench
- because the router is not populated, the workbench fails with "shard not found errors" for the failing index but also other targetted indexes batched with it

### Proposed solution

This PR solves the problem in two places:
- the index id is now validated in `quickwit-serve`, before calling the ingest router (ES bulk and native APIs)
- control plane request errors are now recorded to the workbench so that they can properly be surfaced to the ingest requests. Transient control plane / metastore errors are surfaced as 503 (unavailable) and errors that can't be retried as 500 (internal)

### How was this PR tested?

Added unit and integration (python) tests


quickwit/quickwit-ingest/src/ingest_v2/debouncing.rs
rdettai, line 1: Here we add the capability to write errors back to the barrier so that control plane errors can be shared back with all ingest routing requests that are waiting for shards.

quickwit/quickwit-proto/protos/quickwit/router.proto
```diff
@@ -68,6 +68,7 @@ enum IngestFailureReason {
   INGEST_FAILURE_REASON_ROUTER_LOAD_SHEDDING = 8;
   INGEST_FAILURE_REASON_LOAD_SHEDDING = 9;
   INGEST_FAILURE_REASON_CIRCUIT_BREAKER = 10;
+  INGEST_FAILURE_REASON_UNAVAILABLE = 11;
```
rdettai, line 71: until all servers are upgraded, this will be converted to `IngestServiceError::Internal`

quickwit/quickwit-ingest/src/ingest_v2/router.rs
```diff
@@ -169,7 +170,9 @@ impl IngestRouter {
 
         let mut state_guard = self.state.lock().await;
 
-        for subrequest in pending_subrequests(&workbench.subworkbenches) {
+        for subrequest in
+            pending_subrequests_for_attempt(&workbench.subworkbenches, workbench.num_attempts)
```
rdettai, line 174: we added the logic to not process further subrequests for which we already oberved an error when trying to create the shards. Otherwise control plane errors are overriden as "no shard available".

rdettai:
Splitting this PR into https://github.com/quickwit-oss/quickwit/pull/5721 and https://github.com/quickwit-oss/quickwit/pull/5722

PR #5721, opened 2025-03-25
Invalid index_id with ingest v2 returns 429 

rdettai:
### Description of the issue

When using index templates, specifying and index name that matches the pattern but has illegal characters results in a 429 response code instead of a 400.

Description of the problem:
- the index_id validity is not validated until the metastore is called
- calling `GetOrCreateOpenShards` with an invalid index_id on the metastore fails with MetastoreError::JsonDeserializeError
- a failure to call the metastore is logged but not repported to the ingest workbench
- because the router is not populated, the workbench fails with "shard not found errors" for the failing index but also other targetted indexes batched with it

### Proposed solution

The index id is now validated in `quickwit-serve`, before calling the ingest router (ES bulk and native APIs)

### How was this PR tested?

Added unit and integration (python) tests

### DIFF
--- a/quickwit/quickwit-serve/src/elasticsearch_api/bulk_v2.rs
+++ b/quickwit/quickwit-serve/src/elasticsearch_api/bulk_v2.rs
@@ -123,7 +123,7 @@ pub(crate) async fn elastic_bulk_ingest_v2(
                 )
             })?;
 
-        // Validate index id early because propagating back the right error (400)
+        // Validate index ID early because propagating back the right error (400)
         // from deeper ingest layers is harder
         if validate_identifier("", &index_id).is_err() {
             let invalid_item = make_invalid_index_id_item(index_id.clone(), meta.es_doc_id);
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-serve/src/elasticsearch_api/bulk_v2.rs
```diff
@@ -121,6 +122,16 @@ pub(crate) async fn elastic_bulk_ingest_v2(
                     Some(ElasticException::ActionRequestValidation),
                 )
             })?;
+
+        // Validate index id early because propagating back the right error (400)
```
guilload, line 126:
```suggestion
        // Validate index ID early because propagating back the right error (400)
```

quickwit/quickwit-serve/src/ingest_api/rest_handler.rs
```diff
@@ -215,6 +215,14 @@ async fn ingest_v2(
         None
     };
 
+    // Validate index id early because propagating back the right error (400)
+    // from deeper ingest layers is harder
+    if validate_identifier("", &index_id).is_err() {
+        return Err(IngestServiceError::BadRequest(
+            "invalid index_id".to_string(),
```
guilload, line 222:
```suggestion
            "invalid index ID".to_string(),
```

guilload (approved):

PR #5728, opened 2025-04-03
Implement gRPC health service

guilload:
### Description
Implement basic gRPC health

### How was this PR tested?
Added unit test

### DIFF
--- a/quickwit/Cargo.toml
+++ b/quickwit/Cargo.toml
@@ -246,6 +246,7 @@ tokio-util = { version = "0.7", features = ["full"] }
 toml = "0.7.6"
 tonic = { version = "0.9.0", features = ["gzip"] }
 tonic-build = "0.9.0"
+tonic-health = "0.9.0"
 tonic-reflection = "0.9" 
 tower = { version = "0.4.13", features = [
   "balance",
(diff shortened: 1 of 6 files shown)


quickwit/Cargo.toml
```diff
@@ -246,6 +246,7 @@ tokio-util = { version = "0.7", features = ["full"] }
 toml = "0.7.6"
 tonic = { version = "0.9.0", features = ["gzip"] }
 tonic-build = "0.9.0"
+tonic-health = "0.9.0"
```
trinity-1686a, line 249: you'll need to update the BOSM file

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -1205,7 +1210,15 @@ async fn node_readiness_reporting_task(
     ingester_opt: Option<impl IngesterService>,
     grpc_readiness_signal_rx: oneshot::Receiver<()>,
     rest_readiness_signal_rx: oneshot::Receiver<()>,
+    mut health_reporter: HealthReporter,
 ) {
+    let mut node_ready = false;
+    cluster.set_self_node_readiness(node_ready).await;
+    // Set the initial health status to not serving with "" meaning all services.
```
trinity-1686a, line 1217: is that a convention we choose, or a standard? tonic_health doesn't seem to say anything about empty service names being special

trinity-1686a (approved):

quickwit/quickwit-serve/src/lib.rs
guilload: This is standard:
> The server should use an empty string as the key for server's overall health status

https://github.com/grpc/grpc/blob/master/doc/health-checking.md

guilload: (I updated the comment)

Issue #5731, opened 2025-04-05
`_elastic/_msearch` can return results in a wrong order


kuzaxak:
**Describe the bug**
When using QW with the Grafana plugin, I noticed that requests with multiple metrics may respond with wrong labels. This issue was reported in pthe lugin repo:
* https://github.com/quickwit-oss/quickwit-datasource/issues/139
* https://github.com/quickwit-oss/quickwit-datasource/issues/143

I've added dubug logs to the plugin to see if it's a problem on the plugin side and noticed that the API endpoint `_elastic/_msearch` can return results in a wrong order.

Query 
```
{"ignore_unavailable":true,"index":["otel-traces-v*"]}
{"aggs":{"2":{"date_histogram":{"field":"span_start_timestamp_nanos","fixed_interval":"1m","min_doc_count":0,"extended_bounds":{"min":1743847765408,"max":1743847889194}}}},"query":{"bool":{"filter":[{"range":{"span_start_timestamp_nanos":{"gte":"2025-04-05T10:09:25.408Z","lte":"2025-04-05T10:11:29.194Z"}}},{"query_string":{"default_operator":"AND","query":"service_name:\"master-service\"\n -is_root:\"false\""}}]}},"size":0}
{"ignore_unavailable":true,"index":["otel-traces-v*"]}
{"aggs":{"2":{"date_histogram":{"field":"span_start_timestamp_nanos","fixed_interval":"5m","min_doc_count":0,"extended_bounds":{"min":1743847765408,"max":1743847889194}}}},"query":{"bool":{"filter":[{"range":{"span_start_timestamp_nanos":{"gte":"2025-04-05T10:09:25.408Z","lte":"2025-04-05T10:11:29.194Z"}}},{"query_string":{"default_operator":"AND","query":"service_name:\"ledger\"\n -is_root:\"false\""}}]}},"size":0}
```

Response 1 (correct):
```
{
  "responses": [
    {
      "status": 200,
      "took": 2,
      "timed_out": false,
      "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
      },
      "hits": {
        "total": {
          "value": 41,
          "relation": "eq"
        },
        "hits": []
      },
      "aggregations": {
        "2": {
          "buckets": [
            {
              "doc_count": 0,
              "key": 1743847740000.0,
              "key_as_string": "2025-04-05T10:09:00Z"
            },
            {
              "doc_count": 26,
[...]

PR #5732, opened 2025-04-05
Fix order mismatch in elastic compatible `msearch` api

kuzaxak:
### Description

`buffer_unordered` will collect records in the order in which they complete. But we need to return data in the same order as user requested otherwise user can't match response with requests.

Fixes issue: https://github.com/quickwit-oss/quickwit/issues/5731

### How was this PR tested?

Manually using Grafana plugin


trinity-1686a (approved):
thanks!

PR #5734, opened 2025-04-07
Updating github actions plugins version

fulmicoton-dd:


rdettai (approved):
I triggered the CI to check that it still works because IIRC we already had issues when bumping these actions

PR #5735, opened 2025-04-07
Fix coverage tests

rdettai:
### Description

`test_merge_executor_controlled_directory_kill_switch` is failing since [this change](https://github.com/quickwit-oss/quickwit/pull/5704/files#diff-3e039cd511306f2cec6416a24e964d09696c0cb6d07092ec4683936c940ffd8cR118)

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-indexing/failpoints/mod.rs
+++ b/quickwit/quickwit-indexing/failpoints/mod.rs
@@ -35,7 +35,6 @@ use std::sync::{Arc, Barrier, Mutex};
 use std::time::Duration;
 
 use fail::FailScenario;
-use quickwit_actors::ActorExitStatus;
 use quickwit_common::io::IoControls;
 use quickwit_common::rand::append_random_suffix;
 use quickwit_common::split_file;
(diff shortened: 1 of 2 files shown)


rdettai:
Codecov run: https://github.com/quickwit-oss/quickwit/actions/runs/14309012485/job/40099315322

quickwit/quickwit-indexing/failpoints/mod.rs
```diff
@@ -334,14 +333,15 @@ async fn test_merge_executor_controlled_directory_kill_switch() -> anyhow::Resul
...
+    )
+    .unwrap();
     merge_executor_mailbox.send_message(merge_scratch).await?;
     before_universe_kill.wait();
     universe.kill();
     after_universe_kill.wait();
-    fail::cfg("before-merge-split", "off").unwrap();
-
-    let (exit_status, _) = merge_executor_handle.join().await;
-    assert!(matches!(exit_status, ActorExitStatus::Failure(_)));
```
rdettai, line 344: @fulmicoton-dd I know you wrote this a long time ago, but do you happen to remember why you were testing the `merge_executor_handle.join()` result instead of using a panic failpoint?

Note: the actor exit status is now `Killed` instead of `Failure` because since https://github.com/quickwit-oss/quickwit/pull/5704, merge failures return `Ok()` with an `error!`. log instead of failing the actor. But this is not what this test is about, hence the proposed change.

trinity-1686a (approved):

PR #5742, opened 2025-04-08
Add total size bytes gauge for search after cache

rdettai:
### Description

The search after cache might get pretty big on a system with many scroll / search after queries. This PR adds a way to track its size.

This is part of the global effort to understand memory consumption on searchers.

### How was this PR tested?

Would add a test if we could merge https://github.com/quickwit-oss/quickwit/pull/5718 first.

### DIFF
--- a/quickwit/quickwit-search/src/cluster_client.rs
+++ b/quickwit/quickwit-search/src/cluster_client.rs
@@ -193,7 +193,10 @@ impl ClusterClient {
 
     /// Attempts to store a given key value pair within the cluster.
     ///
-    /// This function may fail silently, if no clients was available.
+    /// Tries to replicate the pair to [`TARGET_NUM_REPLICATION`] nodes, but this function may fail
+    /// silently (e.g if no client was available). Even in case of success, this storage is not
+    /// persistent. For instance during a rolling upgrade, all replicas will be lost as there is no
+    /// mechanism to maintain the replication count.
     pub async fn put_kv(&self, key: &[u8], payload: &[u8], ttl: Duration) {
         let clients: Vec<SearchServiceClient> = self
             .search_job_placer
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-search/src/metrics.rs
```diff
@@ -146,6 +147,12 @@ impl Default for SearchMetrics {
                 &[],
                 ["affinity"],
             ),
+            search_after_cache_size_bytes: new_gauge(
```
trinity-1686a, line 150: that's not a search_after cache, search_after leverage partial request cache when it can (which is not as often as it could)
it should be scrol_context or something

rdettai: thanks for the clarification, I just took the name from here: https://github.com/quickwit-oss/quickwit/blob/9d12159868d7441437f211864e0cff4d44d17867/quickwit/quickwit-search/src/service.rs#L55-L61

Just to be sure that my understanding is correct: "search after" is the name of the search functionality that skip results. To use it, you have two APIs
- `_search/scroll` endpoint that is stateful because it also freezes a list of splits while going through the pages
- the `search_after` field in `_search` that is kind of a stateless equivalent of scroll

I'm going to rename some things here and there to make this clear.

trinity-1686a: classic pagination is the rough equivalent of an SQL `LIMIT $num_res OFFSET $num_res*$page_count` (and has the exact same inefficiency!!), search after is roughly adding a `WHERE sort_key > $last_res_sort_key ORDER BY sort_key ASC`.
scroll are the equivalent to sql stateful cursors, the first implementation was based on pagination (and i believe ES scroll still is), it now rely on search after as that's more efficient. It relying on search after is an implementation detail, and search after does not in anyway need scrolls to exist

quickwit/quickwit-search/src/cluster_client.rs
```diff
@@ -191,7 +191,7 @@ impl ClusterClient {
         client.leaf_list_terms(request.clone()).await
     }
 
-    /// Attempts to store a given search context within the cluster.
+    /// Attempts to store a given key value pair within the cluster.
```
trinity-1686a, line 194: i don't think this is ever used for anything else than a search/scroll context, nor should it be (that's not a persistent storage)

trinity-1686a (approved):

quickwit/quickwit-search/src/cluster_client.rs
rdettai: Agreed. Unfortunately, everything in the code, including core abstractions and RPCs, wants to make this generic. I just tried to add more consistency to that to help the reader. 

I left hints in `quickwit-search/src/scroll_context.rs` and `quickwit-search/src/metrics.rs` about how this store is currently used. I also added some extra clarifications to the function contract [here](https://github.com/quickwit-oss/quickwit/pull/5742/commits/bd570a4162def97de495fbba0bc750cbaba88dac).

PR #5743, opened 2025-04-10
Record search metrics on cancelation

rdettai:
### Description

This PRs moves `[root|leaf]_search_requests_total` `[root|leaf]_search_request_duration_seconds` `[root|leaf]_search_targeted_splits` into tasks so that they are recorded even on future cancelation. This will help having a more accurate view on queries that timeout.

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-search/src/leaf.rs
+++ b/quickwit/quickwit-search/src/leaf.rs
@@ -45,7 +45,6 @@ use tokio::task::JoinError;
 use tracing::*;
 
 use crate::collector::{IncrementalCollector, make_collector_for_split, make_merge_collector};
-use crate::metrics::SEARCH_METRICS;
 use crate::root::is_metadata_count_request_with_ast;
 use crate::search_permit_provider::{SearchPermit, compute_initial_memory_allocation};
 use crate::service::{SearcherContext, deserialize_doc_mapper};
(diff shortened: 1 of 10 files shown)


guilload:
Writing a custom future would be much cleaner:
https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-common/src/tower/metrics.rs

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1447,15 +1446,6 @@ pub async fn leaf_search(
...
 
-    let label_values = match leaf_search_response_reresult {
-        Ok(Ok(_)) => ["success"],
-        _ => ["error"],
-    };
-    SEARCH_METRICS
-        .leaf_search_targeted_splits
-        .with_label_values(label_values)
-        .observe(num_splits as f64);
-
```
rdettai, line 1458: Was there any specific reason why this was measured per index here?

rdettai:
> Writing a custom future would be much cleaner

Good point, I updated the PR to use a future for the leaf search. 

For the root search the state machine is a bit more complicated because the request can also fail/cancel during planning (fetch of the split metadata). I did some refactoring to have a clearer split between the `plan` and `exec` phases, and applied a tracking future to both.

quickwit/quickwit-search/src/root.rs
```diff
@@ -1765,6 +1764,69 @@ pub fn jobs_to_fetch_docs_requests(
...
+}
+
+#[pinned_drop]
+impl<F> PinnedDrop for RootSearchMetricsFuture<F> {
+    fn drop(self: Pin<&mut Self>) {
+        let (targeted_splits, status) = match (&self.step, self.is_success) {
+            // is is a partial success, actual success is recorded during the search step
+            (RootSearchMetricsStep::Plan, Some(true)) => return,
+            (RootSearchMetricsStep::Plan, Some(false)) => (0, "plan-error"),
+            (RootSearchMetricsStep::Plan, None) => (0, "plan-cancelled"),
```
rdettai, line 1789: These extra statuses seem valuable, but we should also try to avoid creating too many series. No strong opinion on whether we should have these or not.

(On that topic, I think `root_search_requests_total` is actually redundant because `root_search_request_duration_seconds` is a histogram and that creates a `_count` series)

guilload (approved):

PR #5748, opened 2025-04-15
Add `--retries` parameter to CLI commands

guilload:
### Description
- Add `--retries` parameter to CLI commands
- Update OpenDAL

I got pulled into updating OpenDAL in order to use `reqwest-middleware` and `reqwest-retry`.

### How was this PR tested?
- Added unit test
- Ran command locally with 10 retries against server that was not up yet. When I finally launched the server, the command succeeded

### DIFF
--- a/quickwit/quickwit-rest-client/src/rest_client.rs
+++ b/quickwit/quickwit-rest-client/src/rest_client.rs
@@ -27,7 +27,10 @@ use quickwit_serve::{
 };
 use reqwest::header::{CONTENT_TYPE, HeaderMap, HeaderValue};
 use reqwest::tls::Certificate;
-use reqwest::{Client, ClientBuilder, Method, StatusCode, Url};
+use reqwest::{ClientBuilder as ReqwestClientBuilder, Method, StatusCode, Url};
+use reqwest_middleware::{ClientBuilder as ReqwestMiddlewareClientBuilder, ClientWithMiddleware};
+use reqwest_retry::RetryTransientMiddleware;
+use reqwest_retry::policies::ExponentialBackoff;
 use serde::Serialize;
 use serde_json::json;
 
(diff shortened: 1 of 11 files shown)


quickwit/quickwit-rest-client/src/rest_client.rs
```diff
@@ -47,28 +50,41 @@ pub const DEFAULT_CLIENT_COMMIT_TIMEOUT: Timeout = Timeout::from_mins(30);
...
-            client_builder = client_builder.connect_timeout(duration);
+            reqwest_client_builder = reqwest_client_builder.connect_timeout(duration);
         }
         if let Some(ca_cert) = ca_cert {
-            client_builder = client_builder
+            reqwest_client_builder = reqwest_client_builder
                 .tls_built_in_root_certs(false)
                 .add_root_certificate(ca_cert);
         }
+        let retry_policy = ExponentialBackoff::builder().build_with_max_retries(num_retries);
```
trinity-1686a, line 76: the default max backoff seems to be 30m, that seems overly long for a cli. i'd be in favor of capping that to a few minutes

quickwit/quickwit-rest-client/src/rest_client.rs
```diff
@@ -771,9 +801,7 @@ mod test {
         let server_url = Url::parse(&format!("http://127.0.0.1:{port}")).unwrap();
         let qw_client = QuickwitClientBuilder::new(server_url).build();
         let error = qw_client.indexes().list().await.unwrap_err();
-
-        assert!(matches!(error, Error::Client(_)));
-        assert!(error.to_string().contains("tcp connect error"));
+        assert!(matches!(error, Error::Middleware(_)));
```
trinity-1686a, line 804: can we assert that we still say something about a tcp connect error? i don't know how this error stringify itself, but if it says nothing about the underlying situation, it's not really useful to the (human) consumer

trinity-1686a (approved):

quickwit/quickwit-rest-client/src/rest_client.rs
guilload: Fair, let me tweak the retry policy.

Issue #5747, opened 2025-04-15
Indexing scheduling broken for Kafka source


rdettai:
**Describe the bug**
When using a `SourceToScheduleType::NonSharded`, the current implementation of the the indexing scheduler seems systematically collocates all pipeline of a given source into the same indexer. For the Kafka source, this prevents distributing the indexing load of a given topic across indexers.

**Steps to reproduce (if applicable)**
The easiest is probably to setup an integration test.

**Expected behavior**
Pipelines should more or less evenly distributed across indexers.

**Configuration:**
Main

rdettai:
Rational behind the current behavior:
- Colocating pipelines for a source on the same node is important to improve merge quality
- We favor a constant default for the estimated pipeline resource because we'd rather have a stable indexing plan than take the risk of entering rebalancing ping pong between the control plane and Kafka

rdettai:
Here is an interesting behavior of the scheduler. 

Let's assign one source with four pipelines to a cluster with 4 nodes (8CPU each):
```rust
let mut problem: SchedulingProblem = SchedulingProblem::with_indexer_cpu_capacities(vec![
    mcpu(8000),
    mcpu(8000),
    mcpu(8000),
    mcpu(8000),
]);
problem.add_source(
    4,
    NonZeroU32::new(PIPELINE_FULL_CAPACITY.cpu_millis()).unwrap(),
);
let old_solution = problem.new_solution();
let solution = solve(problem, old_solution);
for assignement in &solution.indexer_assignments {
    println!("{:?}", assignement);
}
```
```
IndexerAssignment { indexer_ord: 0, num_shards_per_source: {0: 1} }
IndexerAssignment { indexer_ord: 1, num_shards_per_source: {0: 1} }
IndexerAssignment { indexer_ord: 2, num_shards_per_source: {0: 1} }
IndexerAssignment { indexer_ord: 3, num_shards_per_source: {0: 1} }
```
This shows that when the capacity is not exceeded, pipelines are evenly balanced across nodes. We should have colocated at least 2 pipelines per indexer according to the principle stated above:
> Colocating pipelines for a source on the same node is important to improve merge quality

Now if we increase the number of sources to 4 (same cluster capacity):
```rust
let mut problem: SchedulingProblem = SchedulingProblem::with_indexer_cpu_capacities(vec![
    mcpu(8000),
    mcpu(8000),
    mcpu(8000),
    mcpu(8000),
]);
for _ in 0..4 {
[...]

PR #5755, opened 2025-04-17
Balance pipelines when unscaled capacity of individual nodes is exceeded

rdettai:
### Description

Implementation of proposed solution 2:
> for each source, try to first limit the max number of pipelines that can be assigned to each node according to its unscaled original capacity.

### How was this PR tested?

Added unit test.


rdettai:
Superseeded by https://github.com/quickwit-oss/quickwit/pull/5808

PR #5756, opened 2025-04-17
increase sleep time to make test not/less flaky

trinity-1686a:
### Description

increase timeout to make test not flaky (or way less, i wasn't able to make it fail again, but it's possible it would still do, just less often)

### How was this PR tested?

looped `cargo nextest` for a while

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexing_service.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexing_service.rs
@@ -1032,6 +1032,7 @@ mod tests {
     };
 
     use super::*;
+    use crate::actors::merge_pipeline::SUPERVISE_LOOP_INTERVAL;
 
     async fn spawn_indexing_service_for_test(
         data_dir_path: &Path,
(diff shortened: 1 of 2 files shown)


guilload (approved):

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -1605,7 +1605,7 @@ mod tests {
         let observation = indexing_server_handle.process_pending_and_observe().await;
         assert_eq!(observation.num_running_pipelines, 0);
         assert_eq!(observation.num_running_merge_pipelines, 0);
-        universe.sleep(*HEARTBEAT).await;
+        universe.sleep(2 * *HEARTBEAT).await;
```
rdettai, line 1608: ah ok I think I understand why, the supervise loop [actually runs every 1s ](https://github.com/quickwit-oss/quickwit/blob/f843b8858a34de5d6941a476d30bc95828f51e29/quickwit/quickwit-indexing/src/actors/merge_pipeline.rs#L483)

It happens to be that `2 * *HEARTBEAT` is also 1s, but I think it would be more explicit to create a `SUPERVISE_LOOP_INTERVAL` const and use it here and in the `SuperviseLoop` handler.

rdettai (approved):

PR #5758, opened 2025-04-19
Get count from split metadata on simple query with timerange

tontinton:
I think I found a nice optimization here while starting to research the search query code.

Count queries can return much faster by not downloading splits, and I couldn't think of a good reason to always download split files on timerange queries.

### DIFF
--- a/quickwit/quickwit-search/src/root.rs
+++ b/quickwit/quickwit-search/src/root.rs
@@ -636,9 +636,18 @@ async fn search_partial_hits_phase_with_scroll(
 /// metadata count.
 ///
 /// This is done by exclusion, so we will need to keep it up to date if fields are added.
-pub fn is_metadata_count_request(request: &SearchRequest) -> bool {
+pub fn is_metadata_count_request(
+    request: &SearchRequest,
+    split_start_timestamp: Option<i64>,
+    split_end_timestamp: Option<i64>,
+) -> bool {
     let query_ast: QueryAst = serde_json::from_str(&request.query_ast).unwrap();
-    is_metadata_count_request_with_ast(&query_ast, request)
+    is_metadata_count_request_with_ast(
+        &query_ast,
+        request,
+        split_start_timestamp,
+        split_end_timestamp,
+    )
 }
 
 /// Check if the request is a count request without any filters, so we can just return the split
@@ -647,7 +656,12 @@ pub fn is_metadata_count_request(request: &SearchRequest) -> bool {
 /// This is done by exclusion, so we will need to keep it up to date if fields are added.
 ///
 /// The passed query_ast should match the serialized on in request.
-pub fn is_metadata_count_request_with_ast(query_ast: &QueryAst, request: &SearchRequest) -> bool {
+pub fn is_metadata_count_request_with_ast(
+    query_ast: &QueryAst,
+    request: &SearchRequest,
+    split_start_timestamp: Option<i64>,
+    split_end_timestamp: Option<i64>,
+) -> bool {
     if query_ast != &QueryAst::MatchAll {
         return false;
     }
@@ -655,14 +669,23 @@ pub fn is_metadata_count_request_with_ast(query_ast: &QueryAst, request: &Search
         return false;
     }
 
-    // If the start and end timestamp encompass the whole split, it is still a count query.
-    // We remove this currently on the leaf level, but not yet on the root level.
-    // There's a small advantage when we would do this on the root level, since we have the
-    // counts available on the split. On the leaf it is currently required to open the split
-    // to get the count.
-    if request.start_timestamp.is_some() || request.end_timestamp.is_some() {
-        return false;
+    if let Some(request_start_timestamp) = request.start_timestamp {
+        let Some(split_start_timestamp) = split_start_timestamp else {
+            return false;
+        };
+        if split_start_timestamp < request_start_timestamp {
+            return false;
+        }
     }
+    if let Some(request_end_timestamp) = request.end_timestamp {
+        let Some(split_end_timestamp) = split_end_timestamp else {
+            return false;
+        };
+        if split_end_timestamp >= request_end_timestamp {
+            return false;
+        }
+    }
+
     if request.aggregation_request.is_some() || !request.snippet_fields.is_empty() {
         return false;
     }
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-search/src/root.rs
```diff
@@ -730,26 +750,32 @@ pub(crate) async fn search_partial_hits_phase(
...
-    let leaf_search_responses: Vec<LeafSearchResponse> =
-        if is_metadata_count_request(search_request) {
-            get_count_from_metadata(split_metadatas)
+    let mut leaf_search_responses: Vec<LeafSearchResponse> =
+        Vec::with_capacity(split_metadatas.len());
+    let mut leaf_search_jobs = Vec::new();
+    for split in split_metadatas {
+        let start_time = split.time_range.as_ref().map(|x| x.start()).copied();
+        let end_time = split.time_range.as_ref().map(|x| x.end()).copied();
+        if is_metadata_count_request(search_request, start_time, end_time) {
```
rdettai, line 759: Nit, to make `search_partial_hits_phase` a bit more readable, I would pass the `time_range` directly to `is_metadata_count_request`

quickwit/quickwit-search/src/root.rs
```diff
@@ -730,26 +750,32 @@ pub(crate) async fn search_partial_hits_phase(
...
+            .search_job_placer
+            .assign_jobs(leaf_search_jobs, &HashSet::default())
+            .await?;
+        let mut leaf_request_tasks = Vec::new();
+        for (client, client_jobs) in assigned_leaf_search_jobs {
+            let leaf_request =
+                jobs_to_leaf_request(search_request, indexes_metas_for_leaf_search, client_jobs)?;
+            leaf_request_tasks.push(cluster_client.leaf_search(leaf_request, client.clone()));
+        }
+        leaf_search_responses.extend(try_join_all(leaf_request_tasks).await?);
```
rdettai, line 777: Nit, moving it to a separate line would help see a bit more clearly how errors are resolved.

```suggestion
        let executed_leaf_search_responses =         try_join_all(leaf_request_tasks).await?;
leaf_search_responses.extend(executed_leaf_search_responses);
```

rdettai (commented):
Nice optimization. I think it deserves a unit test. E.g the following test scenario:
- The mock metastore returns 4 splits (1 outside the range, 1 overlapping the star, 1 overlapping the end, 1 overlapping the whole range and 1 within the range. 
- The mock search service expects only 1 call.

tontinton:
> Nice optimization. I think it deserves a unit test. E.g the following test scenario:
> 
>     * The mock metastore returns 4 splits (1 outside the range, 1 overlapping the star, 1 overlapping the end, 1 overlapping the whole range and 1 within the range.
> 
>     * The mock search service expects only 1 call.

Will add tests when I have a bit more time.

quickwit/quickwit-search/src/root.rs
```diff
@@ -5056,4 +5076,89 @@ mod tests {
...
+        let mut mock_search = MockSearchService::new();
+        mock_search.expect_leaf_search().times(1).returning(|_req| {
+            Ok(quickwit_proto::search::LeafSearchResponse {
+                num_hits: 1,
+                partial_hits: vec![mock_partial_hit("split_inside", 1, 1)],
+                failed_splits: Vec::new(),
+                num_attempted_splits: 1,
+                ..Default::default()
+            })
+        });
```
rdettai, line 5139: - asserting that `_req` contains the right splits is the important part of this test
- the mock response should be all splits except "split_inside" which is the one we resolve at the root level 🙃

quickwit/quickwit-search/src/root.rs
```diff
@@ -5056,4 +5076,89 @@ mod tests {
...
+        let resp = root_search(
+            &ctx,
+            search_request,
+            MetastoreServiceClient::from_mock(mock_metastore),
+            &cluster_client,
+        )
+        .await?;
+
+        assert_eq!(resp.num_hits, 1);
+        assert_eq!(resp.hits.len(), 1);
```
rdettai, line 5160: This is not the response I would have expected here. `num_hits` should be 10 (num docs in spit inside) + whatever you decide `leaf_search` returns for the overlapping splits.

quickwit/quickwit-search/src/root.rs
```diff
@@ -5056,4 +5076,89 @@ mod tests {
...
     }
+
+    #[tokio::test]
+    async fn test_search_partial_hits_phase_filters_by_time_range() -> anyhow::Result<()> {
+        let search_request = quickwit_proto::search::SearchRequest {
+            start_timestamp: Some(122_000),
+            end_timestamp: Some(129_000),
+            index_id_patterns: vec!["test-index".to_string()],
+            query_ast: qast_json_helper("test", &["body"]),
+            max_hits: 10,
```
rdettai, line 5087: your change is not expected to run if `max_hits!=0`

rdettai (changes requested):
Thanks for adding a test. Unfortunately I think it is flawed (so was my example of test earlier, I should have written the mock search service expects to be called for 3 splits)

tontinton:
> Thanks for adding a test. Unfortunately I think it is flawed (so was my example of test earlier, I should have written "the mock search service expects to be called for 3 splits")

You're totally right, I was supposed to be making a count query, fixed now.

quickwit/quickwit-search/src/root.rs
```diff
@@ -5056,4 +5076,112 @@ mod tests {
...
+
+        let mut mock_search = MockSearchService::new();
+        mock_search
+            .expect_leaf_search()
+            .withf(|leaf_search_req| {
+                let mut expected = HashSet::new();
+                expected.insert("split_before");
+                expected.insert("split_covering_whole");
+                expected.insert("split_overlap_end");
+                expected.insert("split_overlap_start");
```
tontinton, line 5139: This is bad, will look into why all these 4 get searched on...

quickwit/quickwit-search/src/root.rs
```diff
@@ -5056,4 +5076,114 @@ mod tests {
...
+                        == expected
+            })
+            .times(1)
+            .returning(|req| {
+                Ok(quickwit_proto::search::LeafSearchResponse {
+                    num_hits: req.leaf_requests[0]
+                        .split_offsets
+                        .iter()
+                        .map(|s| s.num_docs)
+                        .sum(),
```
rdettai, line 5158: We could make this test more powerful by setting a smaller number here.

quickwit/quickwit-search/src/root.rs
```diff
@@ -5056,4 +5076,114 @@ mod tests {
...
+        let resp = root_search(
+            &ctx,
+            search_request,
+            MetastoreServiceClient::from_mock(mock_metastore),
+            &cluster_client,
+        )
+        .await?;
+
+        assert_eq!(resp.num_hits, 50);
+        assert_eq!(resp.hits.len(), 0);
```
rdettai, line 5185: it is missing from the other tests, but would be nice to also assert `resp.num_successful_splits` here

rdettai (approved):

rdettai:
last minute thought, you could also add some tests in https://github.com/tontinton/quickwit/blob/926a2f33d7b35a5cee064adc457c4503f96dc725/quickwit/rest-api-tests/scenarii/qw_search_api/0001_ts_range.yaml to double check the limit conditions, e.g:
```
endpoint: simple/search
params:
  query: "*"
  start_timestamp: 1684993000
  end_timestamp: 1684993004
expected:
  num_hits: 3
```
should confirm that the upper bound condition is correct

PR #5759, opened 2025-04-19
Optimize time ranged search queries

tontinton:
When the search request contains a time range, we aborted the optimization of converting unneeded split searches into count queries.

Removed the `TODO` that was in the code for this.


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1049,6 +1075,8 @@ impl CanSplitDoBetter {
         // Calculate the number of splits which are guaranteed to deliver enough documents.
         let min_required_splits = splits
             .iter()
+            // splits are sorted by whether they are contained in the request time range
+            .filter(|split| Self::is_contained(split, &request))
             .map(|split| split.num_docs)
             // computing the partial sum
             .scan(0u64, |partial_sum: &mut u64, num_docs_in_split: u64| {
```
rdettai, line 1082: Can you explain why this works? Say you have 5 splits:
- split_1 is contained in the request 
- split_2 to split_5 are only overlapping the request

In that case `min_required_splits` would be 1 here, but there might be better results in any of the other splits.

rdettai (commented):
Thanks for your PR. The way the logic is split here between `optimize_split_order` and `optimize` makes it very hard to proof read. Currently it's already a bit confusing because the `optimize()` logic for each variant depends on the sort order which is different for each variant. It would be good if we could avoid tightening the coupling between the two `match` statements even further by making the sort orders more complex.

quickwit/quickwit-search/src/leaf.rs
tontinton: My thoughts were that there's a

```rust
.take_while(|partial_sum| *partial_sum < num_requested_docs)
```

so `min_required_splits` would not get to be 1, but now I see that I should validate that we actually reached this condition, and only if we reached it to set `min_required_splits` and optimize, otherwise return early.

tontinton: I pushed a quick fix, I need to test it, but I think it answers the problem.

tontinton:
> Thanks for your PR. The way the logic is split here between `optimize_split_order` and `optimize` makes it very hard to proof read. Currently it's already a bit confusing because the `optimize()` logic for each variant depends on the sort order which is different for each variant. It would be good if we could avoid tightening the coupling between the two `match` statements even further by making the sort orders more complex.

It is a bit confusing, I'll try to think of a better way to do this when I have some more time.

quickwit/quickwit-search/src/leaf.rs
rdettai: I didn't know of `std::ops::ControlFlow` 😄. Nevertheless, this reaches the limit when iterators stop being practical. A `for` loop is much more readable here.

rdettai:
I would try refactoring this into:
```
fn optimize()
  match self {
    CandSplitDoBetter::SplitIdHigher(_) => optimize_split_id_higher()
    CandSplitDoBetter::SplitTimestampLower(_) => optimize_split_timesamp_lower()
    ...
    CanSplitDoBetter::Uninformative => {}
  }
```

where `optimize_split_xxx`:
- sorts
- returns early if !is_simple_all_query
- applies its variant specific optimization

This would regroup the optimizations logics with the sorts they depend on to be correct.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1000,134 +995,236 @@ impl CanSplitDoBetter {
...
+
+            partial_sum += split.num_docs;
+
+            if partial_sum >= num_requested_docs {
+                return Some(min_required_splits + 1);
             }
-            CanSplitDoBetter::SplitTimestampLower(_) => {
-                splits.sort_unstable_by_key(|split| split.timestamp_start())
+
+            min_required_splits += 1;
```
rdettai, line 1055: nit, this is more readable
```
min_required_splits += 1;
if partial_sum >= num_requested_docs {
    return Some(min_required_splits);
}
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1000,134 +995,236 @@ impl CanSplitDoBetter {
...
-    fn optimize_split_order(&self, splits: &mut [SplitIdAndFooterOffsets]) {
-        match self {
-            CanSplitDoBetter::SplitIdHigher(_) => {
-                splits.sort_unstable_by(|a, b| b.split_id.cmp(&a.split_id))
+    /// If there's a time range and not enough splits contain at least the number of requested
+    /// documents, return None.
+    fn get_min_required_splits(
+        splits: &[SplitIdAndFooterOffsets],
+        request: &SearchRequest,
+    ) -> Option<usize> {
```
rdettai, line 1038: nice, this is now really straightforward to understand and verify 😄

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1000,134 +995,236 @@ impl CanSplitDoBetter {
...
+    }
+
+    fn optimize_split_timestamp_higher(
+        &self,
+        request: Arc<SearchRequest>,
+        mut splits: Vec<SplitIdAndFooterOffsets>,
+    ) -> Result<Vec<(SplitIdAndFooterOffsets, SearchRequest)>, SearchError> {
+        splits.sort_unstable_by_key(|split| {
+            let contained = Self::is_split_contained_in_search_time_range(split, &request);
+            (!contained, std::cmp::Reverse(split.timestamp_end()))
```
rdettai, line 1097: It is really hard to know here if contained splits come first or second (the ordering of boolean is not obvious, the `!` negation doesn't help 😄 ). Could we maybe explicitly map true/false to something that is more clearly ordered:
```
let contained_first = match Self::is_split_contained_in_search_time_range(split, &request) {
  true => 0u8,
  false => 1u8,
};
(contained_first, std::cmp::Reverse(split.timestamp_end()))
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1000,134 +995,236 @@ impl CanSplitDoBetter {
...
+    }
+
+    fn to_splits_with_request(
+        splits: Vec<SplitIdAndFooterOffsets>,
+        request: Arc<SearchRequest>,
+    ) -> Vec<(SplitIdAndFooterOffsets, SearchRequest)> {
+        splits
+            .into_iter()
+            .map(|split| (split, (*request).clone()))
+            .collect::<Vec<_>>()
```
rdettai, line 1028: keep the comment:
```
// TODO: we maybe want here some deduplication + Cow logic
```

quickwit/quickwit-search/src/leaf.rs
rdettai: MORE IMPORTANT: now that I think about it, I thing it's not good idea to have contained splits processed first in the general case (`!is_simple_all_query` or get_min_required_splits().is_none()`). You miss an optimization opportunity, because once you'll have processed them, you will still need to check the splits overlapping the higher end of the range for better matches.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1000,134 +995,236 @@ impl CanSplitDoBetter {
...
+        }
+
+        None
+    }
+
+    fn optimize_split_id_higher(
+        &self,
+        request: Arc<SearchRequest>,
+        mut splits: Vec<SplitIdAndFooterOffsets>,
+    ) -> Result<Vec<(SplitIdAndFooterOffsets, SearchRequest)>, SearchError> {
```
rdettai, line 1065: we could actually call `to_splits_with_request` before (inside `optimize`) and simplify this signature to something like:
```suggestion
    fn optimize_split_id_higher(
       &mut [(SplitIdAndFooterOffsets, SearchRequest)]
    ) -> Result<(), SearchError> {
```
This would prevent quite a lot of repetition and is also more explicit because it shows that we don't drop splits (note that `&self` is also dropped because not used)
(applies to all `optimzie_split_xxx`)

rdettai (changes requested):
Thanks! This is a bit more verbose, but I find it easier to read!

Check the comment marked "MORE IMPORTANT" first 😉

quickwit/quickwit-search/src/leaf.rs
tontinton: hmm, we can maybe instead of sorting by contained, keep the same sort mechanism as it was.

but still keep the new implementation of `get_min_required_splits`, meaning we won't optimize the first few splits which might not be contained in the search query's range.

tontinton: made the change to not sort, tell me what you think.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1000,134 +995,220 @@ impl CanSplitDoBetter {
...
+            // not enough splits contained in time range.
+            return Ok(Self::to_splits_with_request(splits, request));
+        };
+
+        let mut split_with_req = Self::to_splits_with_request(splits, request);
+
+        // In this case there is no sort order, we order by split id.
+        // If the the first split has enough documents, we can convert the other queries to
+        // count only queries.
+        for (_split, request) in split_with_req.iter_mut().skip(min_required_splits) {
```
trinity-1686a, line 1082: i think this is now wrong due to `get_min_required_splits` skipping over some splits.
Imagine a 2 split scenario, split A contains 2 documents, one in time range and one not, split B contains 10 docs all in range, we want 10 docs. `to_splits_with_request` isn't fully contained in the time range, so it's ignored, B is in range, so it is counted, we get that one split is enough to get the docs we want. Now here B get disabled, because it isn't the 1st split, so only a single doc from A gets returned

tontinton: oh nice catch, that's an easy fix, just `min_required_splits += 1` when `!Self::is_split_contained_in_search_time_range(split, request) {`

PR #5763, opened 2025-04-29
Heap profiler with leak tracking

rdettai:
Description
This PR adds a very lightweight, feature gated, heap profiling endpoint. It superseeds the previous attempt in https://github.com/quickwit-oss/quickwit/pull/5724

Using the jemalloc native profiling is complicated, it's hard to asses its performance impact, and the resulting traces are hard to interpret.

This PR proposes an custom and simple heap profiler:

- written so that the profiler has almost no performance impact when it's not being used actively. Every allocation/deallocation simply checks for an atomic ENABLED flag with relaxed ordering and does no extra work if that flag is false.
- the profiler can be activated with a webservice. It builds allocation statistics by call site and prints a backtrace when they exceed their last highest by a given threshold.

When started, the profiling allocates ~30MB of symbols.

How was this PR tested?
Python scripts.


rdettai:
I ran some tests both during indexing (ingest V2) and search (float aggregation). On a pretty system (10 indexes with 1 milion docs ingested to each index):

Without the feature flag:
- indexing duration: 96.81
- avg search duration: 0.0058

With the feature flag but without starting the profiling:
- indexing duration: 96.80
- avg search duration: 0.0060

With the profiling activated with min_alloc_size=64KiB
- indexing duration: 96.77
- avg search duration: 0.0059

With the profiling activated for search only and min_alloc_size=1KiB
- indexing duration: 96.80
- avg search duration: 0.014

With the profiling activeted for indexing and min_alloc_size=1KiB
- heap profiling stopped, memory_locations full

rdettai:
New measurement with the tokio tracing:

Without the feature flag:
indexing duration: 96.82606220245361
avg search duration: 0.004787445068359375

With the profiling activated with min_alloc_size=64KiB
indexing duration: 96.7932481765747
avg search duration: 0.005725836753845215

quickwit/quickwit-common/src/alloc_tracker.rs
```diff
@@ -0,0 +1,437 @@
...
+        if new_threshold_exceeded {
+            let reported_statistic = *entry;
+            entry.last_report = entry.size;
+            AllocRecordingResponse::ThresholdExceeded(reported_statistic)
+        } else {
+            AllocRecordingResponse::ThresholdNotExceeded
+        }
+    }
+
+    /// Updates the the memory location and size of an existing allocation. Only
```
fulmicoton-dd, line 145:
```suggestion
    /// Updates the memory location and size of an existing allocation. Only
```

quickwit/quickwit-common/src/alloc_tracker.rs
```diff
@@ -0,0 +1,437 @@
...
+        new_size_bytes: u64,
+        old_ptr: *mut u8,
+        new_ptr: *mut u8,
+    ) -> ReallocRecordingResponse {
+        let TrackerStatus::Started { reporting_interval } = self.status else {
+            return ReallocRecordingResponse::NotStarted;
+        };
+        let (callsite_hash, old_size_bytes) = if old_ptr != new_ptr {
+            let Some(old_alloc) = self.memory_locations.remove(&(old_ptr as usize)) else {
+                return ReallocRecordingResponse::ThresholdNotExceeded;
```
fulmicoton-dd, line 160: is this supposed to never happen or something like that?

quickwit/quickwit-common/src/alloc_tracker.rs
```diff
@@ -0,0 +1,437 @@
...
+            };
+            self.memory_locations.insert(
```
fulmicoton-dd, line 162: how are we certain that this insert does not allocate?

fulmicoton-dd (approved):
cool job

quickwit/quickwit-common/src/alloc_tracker.rs
rdettai: Allocations that exist before the profiler is started are not tracked in `self.memory_locations`. For those untracked allocations we decide not to track reallocations either. It's a small blind spot in the profiler's capability. For instance we are not able to pick up the growth of a long lived memory arena.

quickwit/quickwit-common/src/alloc_tracker.rs
rdettai: Indeed, this insert does not grow the map because of the remove right before. At a load factor of 0.5 I'm pretty confident no resize will be triggered.

PR #5771, opened 2025-05-20
Increase target shard usage

rdettai:
### Description

Currently the target shard usage before scaling down the number of shards is 20%. This is quite low which results in a high number of shards.

### How was this PR tested?

Added unit tests.

### DIFF
--- a/quickwit/quickwit-control-plane/src/ingest/scaling_arbiter.rs
+++ b/quickwit/quickwit-control-plane/src/ingest/scaling_arbiter.rs
@@ -43,8 +43,8 @@ impl ScalingArbiter {
             scale_up_shards_short_term_threshold_mib_per_sec: max_shard_throughput_mib_per_sec
                 * 0.8f32,
             scale_up_shards_long_term_threshold_mib_per_sec: max_shard_throughput_mib_per_sec
-                * 0.3f32,
-            scale_down_shards_threshold_mib_per_sec: max_shard_throughput_mib_per_sec * 0.2f32,
+                * 0.375f32,
+            scale_down_shards_threshold_mib_per_sec: max_shard_throughput_mib_per_sec * 0.25f32,
             shard_scale_up_factor,
         }
     }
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-control-plane/src/ingest/scaling_arbiter.rs
```diff
@@ -43,8 +43,8 @@ impl ScalingArbiter {
             scale_up_shards_short_term_threshold_mib_per_sec: max_shard_throughput_mib_per_sec
                 * 0.8f32,
             scale_up_shards_long_term_threshold_mib_per_sec: max_shard_throughput_mib_per_sec
-                * 0.3f32,
-            scale_down_shards_threshold_mib_per_sec: max_shard_throughput_mib_per_sec * 0.2f32,
+                * 0.375f32,
+            scale_down_shards_threshold_mib_per_sec: max_shard_throughput_mib_per_sec * 0.25f32,
             shard_scale_up_factor,
```
rdettai, line 48: Increase the target load by 25% only. Increasing it more would prevent the scaling from 1 shard to 2 shards to happen properly. Take `max_shard_throughput_mib_per_sec=5MB` and `num_open_shards=1`:
- with 0.3 as scale up load factor, `scale_up_shards_long_term_threshold_mib_per_sec=1.5MB`. This means that `avg_long_term_ingestion_rate` must be >3MB for scaling event to occur.
- with 0.375 as scale up load factor, `scale_up_shards_long_term_threshold_mib_per_sec=1.875MB`. This means that `avg_long_term_ingestion_rate` must be >3.75MB for scaling event to occur.
- with 0.45 as scale up load factor, `scale_up_shards_long_term_threshold_mib_per_sec=2.25MB`. This means that `avg_long_term_ingestion_rate` must be >4.5MB for scaling event to occur.

This is due to how `long_term_scale_up_threshold_max_shards` avoids getting into the interval [scale_down, scale_up]

PR #5780, opened 2025-05-27
Update a bunch of dependencies

guilload:
### Description
Per title.

### How was this PR tested?
Ran `make test-all`


tandr:
Folks,  

I was looking at the diffs, and puzzled as why there are tens of dependencies have what feels like a version downgrades? 

For example
>coarsetime = "0.1.33"

became 

>coarsetime = "0.1"

Am I missing something here, or cargo will just pull "latest for that version" automatically?

Thank you in advance.

PR #5781, opened 2025-05-30
Add `min_shards` parameter to index config

guilload:
### Description
Add `min_shards` parameter to index config. When the control place receives an `GetOrCreateOpenShards` request, it opens `min_shards` shards instead of 1. When scaling down, it ensures that we do not scale down below `min_shards`. When scaling up, it ensures that there is at least `min_shards` open if the index is actively receiving data (not idling).

### How was this PR tested?
Added unit tests

### DIFF
--- a/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
+++ b/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
@@ -16,6 +16,7 @@ use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::fmt;
 use std::future::Future;
+use std::num::NonZeroUsize;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -393,18 +394,24 @@ impl IngestController {
             &local_shards_update.source_uid,
             &local_shards_update.shard_infos,
         );
-        let Some(scaling_mode) = self.scaling_arbiter.should_scale(shard_stats) else {
+        let min_shards = model
+            .index_metadata(&local_shards_update.source_uid.index_uid)
+            .expect("index should exist")
+            .index_config
+            .ingest_settings
+            .min_shards;
+
+        let Some(scaling_mode) = self.scaling_arbiter.should_scale(shard_stats, min_shards) else {
             return Ok(());
         };
-
         match scaling_mode {
-            ScalingMode::Up(shards) => {
+            ScalingMode::Up(num_shards) => {
                 self.try_scale_up_shards(
                     local_shards_update.source_uid,
                     shard_stats,
                     model,
                     progress,
-                    shards,
+                    num_shards,
                 )
                 .await?;
             }
@@ -412,6 +419,7 @@ impl IngestController {
                 self.try_scale_down_shards(
                     local_shards_update.source_uid,
                     shard_stats,
+                    min_shards,
                     model,
                     progress,
                 )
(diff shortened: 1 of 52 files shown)


esatterwhite:
If idle shards are an expensive thing to keep, I don't think it a bad thing if it goes below. Especially once the finalized runs after data stops flowing. 

The initial create is kind of the bigger win. In our case at least

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -464,13 +472,20 @@ impl IngestController {
...
+                    .min_shards
+                    .get();
                 let source_uid = SourceUid {
                     index_uid,
                     source_id: get_open_shards_subrequest.source_id.clone(),
                 };
                 *num_missing_shards_per_source_uids
                     .entry(source_uid)
-                    .or_default() += 1;
+                    .or_default() += min_shards;
```
fulmicoton-dd: now I am loss with the code an the previous one.

I was expecting...

`num_missing_shards_per_source_uids.insert(source_uid, min_shards);`

Why do we count subrequests targeting the same source_uid as several? (Does this case even exist?)

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -854,13 +869,15 @@ impl IngestController {
...
         source_uid: SourceUid,
         shard_stats: ShardStats,
+        min_shards: NonZeroUsize,
         model: &mut ControlPlaneModel,
         progress: &Progress,
     ) -> MetastoreResult<()> {
-        if shard_stats.num_open_shards == 0 {
+        // The scaling arbiter should not suggest scaling down if the number of shards is already
+        // below the minimum, but we're just being defensive here.
+        if shard_stats.num_open_shards <= min_shards.get() {
```
fulmicoton-dd, line 876: what will happen for indexes that are "sunset"? In the case of mezmo, indexes from yesterday.

fulmicoton-dd (approved):

quickwit/quickwit-control-plane/src/ingest/scaling_arbiter.rs
```diff
@@ -66,10 +68,20 @@ impl ScalingArbiter {
...
     /// make the [`quickwit_proto::control_plane::GetOrCreateOpenShardsRequest`]
-    pub(crate) fn should_scale(&self, shard_stats: ShardStats) -> Option<ScalingMode> {
-        if shard_stats.num_open_shards == 0 {
+    pub(crate) fn should_scale(
+        &self,
+        shard_stats: ShardStats,
+        min_shards: NonZeroUsize,
+    ) -> Option<ScalingMode> {
+        // If ingest is idle, there is nothing to do.
+        if shard_stats.num_open_shards == 0 || shard_stats.avg_long_term_ingestion_rate == 0.0 {
```
rdettai, line 78: If an index with hundreds of shards stops receiving data suddenly, avg_long_term_ingestion_rate is likely to become 0 before the number of shards is scaled down. In that case it would become idle with a large number of shards.

guilload: Not a problem:
https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-ingest/src/ingest_v2/idle.rs

rdettai: @guilload could we maybe add a small hint about this in the comment?

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
guilload: Yes, the client should send at most one subrequest per source UID so using `insert` makes more sense.

quickwit/quickwit-control-plane/src/ingest/scaling_arbiter.rs
guilload: Sounds good.

PR #5784, opened 2025-06-04
Fix Jemalloc not used in release build

rdettai:
### Description

This PR merges the hotfix qw-airmail-20250522-hotfix.

Jemalloc was disabled in the release build since https://github.com/quickwit-oss/quickwit/pull/5763

### How was this PR tested?

Checked whether the jemalloc metrics are now ok (if jemalloc is not the global alloc their value max out at ~100KB)


rdettai:
good point, I'll do that in a separate PR

Issue #4630, opened 2024-02-26
Improve assignement of kafka source pipelines on several indexers 


fmassot:
A user reported the following issue (I checked the cluster state):

- He has around 50 Kafka sources. The biggest Kafka source has 4 desired pipelines, most of which have one.
- He has 10 indexers, each with 4-6 pipelines. 1 indexer has all the pipelines of the biggest Kafka source. 

This is bad because, as you can imagine, only 1 indexer is responsible for indexing the biggest Kafka source. And all other indexers are doing less work.

fulmicoton:
Replied offline by @fmassot : no the problem is an unbalanced load on the different nodes.

axzsd:
Is this fixed now?

Issue #4470, opened 2024-01-27
Sequential merge prevent provoke back pressure on large index


fulmicoton:
When working with large hardware it is possible to  schedule any number of a pipeline from the same index.
A single merge pipeline is then shared an might not be able to keep up with the indexing pipelines.

We need to make it possible to execute several merges concurrently.

PR #5792, opened 2025-06-10
Limit to 3 pipelines per node per source

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/4470
Closes https://github.com/quickwit-oss/quickwit/issues/4630

This addresses both the limitation that only 1 merge pipelines can run per indexer at any given time and the fact that nodes systematically end up with all pipelines of a source when using Kafka, even if the number of pipelines for that source is rather large.

### How was this PR tested?

Added unit test, should add more

### DIFF
--- a/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
+++ b/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
@@ -417,17 +417,14 @@ fn place_unassigned_shards_single_source(
     for PlacementCandidate {
         indexer_ord,
         available_capacity,
+        current_num_shards,
         ..
     } in sorted_candidates
     {
-        let current_num_shards_for_indexer_and_source = *solution.indexer_assignments[*indexer_ord]
-            .num_shards_per_source
-            .get(&source.source_ord)
-            .unwrap_or(&0);
         let num_placable_shards_for_available_capacity =
             available_capacity.cpu_millis() / source.load_per_shard;
-        let num_placable_shards_for_limit = limit_num_shards_per_indexer_per_source
-            .saturating_sub(current_num_shards_for_indexer_and_source);
+        let num_placable_shards_for_limit =
+            limit_num_shards_per_indexer_per_source.saturating_sub(*current_num_shards);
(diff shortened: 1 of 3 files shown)


fmassot:
why 3?

rdettai:
> why 3?

It's a ratio proposed by Paul [here](https://github.com/quickwit-oss/quickwit/issues/5747). The rational is that if you you start saturating the systems, merges being 3x faster than indexing, with this ratio merges would be able to keep up.

rdettai:
Actually this is wrong for now, I thought shards in the simplified problem were indexing pipelines in the physical plan, but it's not true. Shards in the simplified problem are physical shards, and they are mapped to pipelines in https://github.com/quickwit-oss/quickwit/blob/5853b730e7c4bae827d617eb3582cb71ca216283/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs#L204

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -349,15 +352,31 @@ struct NotEnoughCapacity;
...
     source: &Source,
     mut indexer_with_capacities: impl Iterator<Item = (IndexerOrd, CpuCapacity)>,
+    num_indexers: usize,
     solution: &mut SchedulingSolution,
 ) -> Result<(), NotEnoughCapacity> {
     let mut num_shards = source.num_shards;
+    // To ensure that merges can keep up, we try not to assign more than 3
+    // pipelines per indexer for a source (except if there aren't enough nodes).
+    let target_limit_num_shards_per_indexer_per_source =
+        3 * MAX_LOAD_PER_PIPELINE.cpu_millis() / source.load_per_shard.get();
```
rdettai, line 414: This is creating some undesired coupling with the rest of the code:
- we rely on convert_scheduling_solution_to_physical_plan to use the same MAX_LOAD_PER_PIPELINE to create the right amount of pipelines
- we rely on the default load_per_pipeline for non-ingest sources (e.g Kafka) to also use MAX_LOAD_PER_PIPELINE.

PR #5793, opened 2025-06-11
Test scheduling solution stability

rdettai:
### Description

I noticed that sometime the scheduling solution changes when the same problem is solved multiple times. This PR adds a test to ensure that this happens only once and we don't end up with an unstable solution.

### How was this PR tested?

This PR doesn't change the logic, only adds tests.


quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -208,6 +208,27 @@ fn assert_enforce_nodes_cpu_capacity_post_condition(
     );
 }
 
+// ----------------------------------------------------
```
rdettai, line 211: not related to the PR, just moved this up because it was wrongly placed.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic_model.rs
```diff
@@ -209,7 +209,12 @@ impl IndexerAssignment {
             .unwrap_or(0u32)
     }
 
+    /// Add shards to a source (noop of `num_shards` is 0).
     pub fn add_shards(&mut self, source_ord: u32, num_shards: u32) {
+        // No need to fill indexer_assignments with empty assignments.
+        if num_shards == 0 {
+            return;
+        }
```
rdettai, line 217: this doesn't change the logic, just ensures a more consistent representation of the state so that we can compare it better in tests.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -358,10 +359,6 @@ fn place_unassigned_shards_single_source(
         let num_placable_shards = available_capacity.cpu_millis() / source.load_per_shard;
         let num_shards_to_place = num_placable_shards.min(num_shards);
         // Update the solution, the shard load, and the number of shards to place.
-        if num_shards_to_place == 0u32 {
```
fulmicoton-dd, line 361: just to be sure, you removed that statement because it is useless correct?
This is not exactly no-op. It will create an entry in indexer_assignement. Do we want this?

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic_model.rs
fulmicoton-dd: ah ok you moved it here.

fulmicoton-dd (approved):

PR #5797, opened 2025-06-12
Clarify how to perform exact match searches in Quickwit

rdettai:
### Description

This PR adds some docs to explain the Quickwit equivalent of the Elasticsearch `keyword` type. It also adds some examples of textual exact match strategies in the integration tests. 

I did not document the workaround where we use range queries to perform fast field based exact queries. It's kind of hacky and given that the parser doesn't support whitespaces in range queries, it's hardly usable in practice.

### How was this PR tested?

Integ tests

### DIFF
--- a/quickwit/rest-api-tests/scenarii/qw_search_api/0004_exact_string.yaml
+++ b/quickwit/rest-api-tests/scenarii/qw_search_api/0004_exact_string.yaml
@@ -60,7 +60,7 @@ expected:
 endpoint: nested/search
 method: POST
 json:
-  query: "text_fast:[fast_text_value TO fast_text_value]"
+  query: "text_fast:[fast-text-value TO fast-text-value]"
 expected:
   num_hits: 0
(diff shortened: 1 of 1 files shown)


quickwit/rest-api-tests/scenarii/qw_search_api/0004_exact_string.yaml
```diff
@@ -0,0 +1,85 @@
...
+method: POST
+json:
+  query: "text_fast:[fast_text_value_dashes TO fast_text_value_dashes]"
+expected:
+  num_hits: 0
+---
+endpoint: nested/search
+method: POST
+json:
+  query: "text_fast:[fast_text_value TO fast_text_value]"
```
trinity-1686a: what is this testcase about? maybe it should be with dashes instead of underscores?

trinity-1686a (approved):

PR #5804, opened 2025-06-18
Update scheduling README

rdettai:
### Description

Update the README to try to better match the implementation.

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/README.md
+++ b/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/README.md
@@ -3,13 +3,10 @@
 Quickwit needs to assign indexing tasks to a set of indexers nodes.
 We call the result of this decision the indexing physical plan.
 
-This needs to be done under the constraints of:
-- not exceeding the maximum load of each node. (O)
-
 We also want to observe some interesting properties such as:
 - (A) we want to avoid moving indexing tasks from one indexer to another one needlessly.
 - (B) we want a source to be spread amongst as few nodes as possible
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/README.md
```diff
@@ -3,13 +3,10 @@
...
 We call the result of this decision the indexing physical plan.
 
-This needs to be done under the constraints of:
-- not exceeding the maximum load of each node. (O)
-
 We also want to observe some interesting properties such as:
 - (A) we want to avoid moving indexing tasks from one indexer to another one needlessly.
 - (B) we want a source to be spread amongst as few nodes as possible
-- (C) we prefer to respect some margin on the capacity of all nodes.
+- (C) we want to balance the load between nodes as soon as the load is significative (>30%)
```
fulmicoton-dd:
```suggestion
- (C) we want to balance the load between nodes as soon as the load is significatively (>30%) higher than the average (target) load
```

fulmicoton-dd: I forgot if we discussed it but I wonder if the cause of the imbalance we observe on mezmo, espeically after restarts, could be due to the fact that we only rebalance if a node is above the >30%: we do not do anything about nodes not getting enough work.

One could have 30 nodes running at 110% of the average value, and 3 nodes doing nothing.
The scheduler would be happy with this.

fulmicoton-dd (approved):

PR #5808, opened 2025-06-19
Improve shard collocation while limiting to 3 pipelines per node

### LINKED ISSUE #4470: Sequential merge prevent provoke back pressure on large index

### LINKED ISSUE #4630: Improve assignement of kafka source pipelines on several indexers

### LINKED ISSUE #5747: Indexing scheduling broken for Kafka source

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/4470
Closes https://github.com/quickwit-oss/quickwit/issues/4630

#### Improve shard collocation

When placing new shards for an existing source, when affinity is not defined (e.g Kafka source), we currently have no affinity for nodes where we have existing indexing pipelines. This can create a fragmented distribution of shards.

This solution is not perfect, ideally we should split the assignation in two steps (just as we do for the shard affinity):
- first assign shards to nodes that already have shards for all the source
- second for sources where we still have shards to assign, assign them to new nodes

#### Limit to 3 pipelines per node

See https://github.com/quickwit-oss/quickwit/pull/5792

### How was this PR tested?

Added unit test

### DIFF
--- a/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
+++ b/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
@@ -217,14 +217,15 @@ fn assert_enforce_nodes_cpu_capacity_post_condition(
 // We go through the sources in decreasing order of their load,
 // in two passes.
 //
-// In the first pass, we have a look at
-// the nodes with which there is an affinity.
-//
-// If one of them has room for all of the shards, then we assign all
-// of the shards to it.
+// In the first pass, we have a look at the nodes with which there is an
+// affinity.
 //
 // In the second pass, we just put as many shards as possible on the node
 // with the highest available capacity.
+//
+// Each time node is considered for placing shards of a given source, we place
+// as many of them as possible, but without exceeding the current node capacity
+// and the maximum capacity of 3 indexing pipelines.
 //
 // If this algorithm fails to place all remaining shards, we inflate
 // the node capacities by 20% in the scheduling problem and start from the beginning.
--- a/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
+++ b/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
@@ -12,14 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::cmp::Reverse;
+use std::cmp::{Ordering, Reverse};
 use std::collections::BTreeMap;
 use std::collections::btree_map::Entry;
 
-use itertools::Itertools;
 use quickwit_proto::indexing::CpuCapacity;
 
 use super::scheduling_logic_model::*;
+use crate::indexing_scheduler::MAX_LOAD_PER_PIPELINE;
 use crate::indexing_scheduler::scheduling::inflate_node_capacities_if_necessary;
 
 // ------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -229,21 +228,61 @@ fn assert_enforce_nodes_cpu_capacity_post_condition(
...
+
+impl PartialOrd for PlacementCandidate {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        // Higher affinity is better
+        match self.affinity.cmp(&other.affinity) {
+            Ordering::Equal => {}
+            ordering => return Some(ordering.reverse()),
+        }
+        // If tie, pick the node with shards already assigned first
+        match self.current_num_shards.cmp(&other.current_num_shards) {
```
fulmicoton-dd: I think what matters here is whether current_num_shards is non null, no?

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -229,21 +228,61 @@ fn assert_enforce_nodes_cpu_capacity_post_condition(
...
-            source,
-            indexers_with_most_available_capacity,
-            &mut solution,
-        )?;
+        let mut placements: Vec<PlacementCandidate> = solution
+            .indexer_assignments
+            .iter()
+            .map(|indexer_assignment: &IndexerAssignment| {
+                let available_capacity = indexer_assignment.indexer_available_capacity(problem);
+                assert!(available_capacity >= 0i32);
```
fulmicoton-dd, line 280: indexer with no cap have been filtered upstream?

fulmicoton-dd (approved):

rdettai:
Note: when using ingest V2, the effect of this PR should be moderate. Indeed, ingest V2 creates shards on nodes randomly: https://github.com/quickwit-oss/quickwit/blob/52ac2e18c8d372b205c031aad1faaf94a33f0d1a/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs#L114-L125

This means that in most cases it's the affinity placement stage that is going to scatter shards across indexers. It's only when nodes don't have enough (artificially scaled) capacity during that stage that the placement change from this PR will kick in. In practice this happens fairly frequently.

rdettai:
Hi @daniele-br you should definitely test this branch, it has helped a lot for our setup, and our load distribution is a bit similar to yours (probably a bit less skewed). There are still distribution issues, especially when adding new nodes and I'm planning to work on it in the coming weeks.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -229,19 +229,72 @@ fn assert_enforce_nodes_cpu_capacity_post_condition(
 // If this algorithm fails to place all remaining shards, we inflate
 // the node capacities by 20% in the scheduling problem and start from the beginning.
```
fulmicoton-dd, line 231: @rdettai should we update the comment above?

rdettai-sk: well spotted

rdettai-sk:
@earlbread this is not up to me. We are running a fork with this change.

earlbread:
@rdettai-sk Got it, thanks.

minhyuk-playground:
Hi @fulmicoton-dd , This PR looks approved but is currently behind main. Could I ask if there are any plans to rebase and merge it?

paragor:
Guys, any updates? This PR makes the delivery path kafka -> quickwit much more reliable and easier to maintenance

earlbread:
@rdettai-sk 
We've been running this branch in production since June (16 indexers, Kafka-only sources, 1GB/s), big improvement over main.
One issue we experienced: when a node is added or restarted, it gets no tasks assigned and stays idle, since the scheduler keeps the previous assignment and never rebalances.
Our workaround: disable -> enable a large source, which gets its pipelines re-placed onto the idle node.
```
quickwit source disable --index my-index --source my-source
quickwit source enable --index my-index --source my-source
```

PR #5811, opened 2025-06-19
Record end-to-end indexing duration in S3 file notification source

tontinton:


quickwit/quickwit-common/src/metrics.rs
```diff
@@ -429,16 +429,28 @@ impl InFlightDataGauges {
...
 }
 
+/// This function returns `index_name-source_name` or projects it to `<any>` if per-index metrics are disabled.
+pub fn source_label(index_name: &str, source_name: &str) -> String {
+    if is_per_index_metrics_enabled() {
+        format!("{index_name}-{source_name}")
+    } else {
+        "__any__".to_string()
+    }
+}
```
rdettai, line 454: Having the source and the index in separate labels is better in all ways.

quickwit/quickwit-indexing/src/metrics.rs
```diff
@@ -98,6 +100,18 @@ impl Default for IndexerMetrics {
...
+            queue_source_index_duration_seconds: Lazy::new(|| {
+                new_histogram_vec(
+                    "queue_source_index_duration_seconds",
+                    "Number of seconds it took since the message was generated until it was sent \
+                     to be acknowledged (deleted).",
+                    "indexing",
+                    &[],
+                    ["source"],
+                    // 15 seconds up to 3 minutes
+                    linear_buckets(15.0, 15.0, 12).unwrap(),
```
rdettai, line 112: Is 3 min really enough in general? I think you might be interested to see much higher value when the system starts to behave badly. But 12 buckets is already a lot, I would then switch to an exponential scale.

quickwit/quickwit-indexing/src/metrics.rs
```diff
@@ -98,6 +100,18 @@ impl Default for IndexerMetrics {
                     &[],
                 )
             }),
+            queue_source_index_duration_seconds: Lazy::new(|| {
+                new_histogram_vec(
+                    "queue_source_index_duration_seconds",
+                    "Number of seconds it took since the message was generated until it was sent \
```
rdettai, line 106:
```suggestion
                    "Duration (seconds) between the queue message event time (parsed from its content if available) and its acknowledgment"
```

quickwit/quickwit-indexing/src/source/queue_sources/local_state.rs
```diff
@@ -34,8 +35,8 @@ pub struct QueueLocalState {
     /// Message that is currently being read and sent to the `DocProcessor`
     read_in_progress: Option<InProgressMessage>,
     /// Partitions that were read and are still being indexed, with their
-    /// associated ack_id
-    awaiting_commit: BTreeMap<PartitionId, String>,
+    /// associated ack_id and optional creation timestamp
+    awaiting_commit: BTreeMap<PartitionId, (String, Option<OffsetDateTime>)>,
```
rdettai, line 39: We could refactor `(String, Option<OffsetDateTime>)` to 

```rust
pub struct AwaitingCommitMessage {
  ack_id: String,
  source_event_time_opt: Option<OffsetDateTime>
}
```

It would help readability I think.

quickwit/quickwit-indexing/src/source/queue_sources/message.rs
```diff
@@ -122,6 +128,7 @@ impl PreProcessedPayload {
 pub struct PreProcessedMessage {
     pub metadata: MessageMetadata,
     pub payload: PreProcessedPayload,
+    pub timestamp_opt: Option<OffsetDateTime>,
```
rdettai, line 131:
```suggestion
    pub source_event_time_opt: Option<OffsetDateTime>,
```

quickwit/quickwit-indexing/src/source/queue_sources/coordinator.rs
```diff
@@ -302,8 +304,20 @@ impl QueueCoordinator {
             .collect::<Vec<_>>();
         let mut completed = Vec::new();
         for partition_id in committed_partition_ids {
-            let ack_id_opt = self.local_state.mark_completed(partition_id);
-            if let Some(ack_id) = ack_id_opt {
+            let completed_opt = self.local_state.mark_completed(partition_id);
+            if let Some((ack_id, timestamp_opt)) = completed_opt {
+                if let Some(timestamp) = timestamp_opt {
+                    let duration = OffsetDateTime::now_utc() - timestamp;
```
rdettai, line 310: This can panic if something is off with the event time. It's unlikely, but it's unfortunate to panic because of a metric. Refactor this as:
```
fn record_index_duration_metric(index_id: &str, source_id: &str, source_event_time_opt: Option<OffsetDateTime>) {
  let Some(source_event_time) = source_event_time_opt else {
     return
  };
  let now = OffsetDateTime::now_utc();
  if now < source_event_time {
    error!("Event time smaller than current time");
    return;
  }
  let duration = now - source_event_time;
  let index_label = index_label(
      index_id,
  );
  let source_label = source_label(
      source_id,
  );
  crate::metrics::INDEXER_METRICS
      .queue_source_index_duration_seconds
      .with_label_values([index_label, source_label])
      .observe(duration.as_seconds_f64());
}
```

rdettai (commented):
Thanks for the contribution. It's very specific, but I think it can be valuable to any user. Could you try also running the coverage tests (which also contain SQS tests) on your fork repo? It seems I can't do that from this repo.

rdettai:
Did you check out the SQS ApproximateAgeOfOldestMessage metric? I'm not sure what you want to monitor exactly, but usually it's what you would use to monitor that queues are processed through in a timely manner.

Issue #5787, opened 2025-06-05
Control Plane not exiting on panic


UN0wen:
**Describe the bug**
We're deploying quickwit in cluster mode through helm (chart version 7.15, quickwit version 0.8.2). We noticed today that some new pods are spinning up but not connected to the existing cluster, instead only peering between themselves and crashing since there are no metastore.

Upon inspection, we found this traceback in the control plane logs (the pod is healthy)

```
thread 'tokio-runtime-worker' panicked at /usr/local/cargo/git/checkouts/chitchat-22cf90d3696646d6/d039699/chitchat/src/delta.rs:413:9:
assertion failed: mtu >= 100
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It seems like after encountering this error, the control plane does not exit and continues to run, seemingly doing nothing (?). 

**Expected behavior**
Control plane should exit with an error if an internal component panics.

**Configuration:**
Please provide:

1. Output of `quickwit --version`

quickwit version: 0.8.2 (aarch64-unknown-linux-gnu 2024-06-17T16:36:47Z 42766b8)

rachel-mj-park:
@earlbread and I are working together, so I attached additional info.

- Quickwit version: `0.8.2`
- Component: 12 indexer, 1 control plane, 1 janitor, 3 metastore, 10 searcher
- I am not sure, but there were likely hundreds of zombie nodes in metastore at that time

So we first restarted metastore, and then sequentially restarted indexer, searcher and janitor to resolve the issue.

rdettai:
About the root cause: Zombie nodes in the the chitchat state is a [known issue](https://github.com/quickwit-oss/quickwit/issues/5135). It was solved a few months ago.

About the panic issue: too bad you don't have a stack trace 😢 . 

If you are adventurous, until we cut a release (long overdue, I know), you could try using the tag `qw-airmail-20250522`. It includes all recent changes and is pretty stable.

earlbread:
Thank you for the response.
If I upgrade from version 0.8.2 to `qw-airmail-20250522`, are there any breaking changes?

fulmicoton:
> It seems like after encountering this error, the control plane does not exit and continues to run, seemingly doing nothing (?).

This is a very valid point. Let's fix this.

earlbread:
@rdettai Hi, I see that there are two branches: qw-airmail-20250522 and qw-airmail-20250522-hotfix.

It looks like this commit has been added:
[...]

PR #5819, opened 2025-06-25
Exit when chitchat server fails

rdettai:
### Description

Closes https://github.com/quickwit-oss/quickwit/issues/5787

Currently, Quickwit node would keep running if chitchat server stops (panic or error). This PR adds a watcher for this and stops the nodes if something goes wrong in chitchat.

This PR depends on https://github.com/quickwit-oss/chitchat/pull/162

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-serve/src/lib.rs
+++ b/quickwit/quickwit-serve/src/lib.rs
@@ -395,7 +395,12 @@ fn start_shard_positions_service(
     });
 }
 
-async fn shutdown_handler(
+/// Waits for the shutdown signal and notifies all other services when it
+/// occurs.
+///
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-serve/src/lib.rs
```diff
@@ -395,6 +395,36 @@ fn start_shard_positions_service(
     });
 }
 
+async fn shutdown_handler(
```
fulmicoton-dd:
```suggestion
/// This handler is typically called when we receive a SIGTERM signal.
/// For instance, if k8s is trying to decomission a pod.
async fn shutdown_handler(
```

fulmicoton-dd (approved):
approved. let's merge the chitchat PR and update the rev accordingly though

PR #5821, opened 2025-06-26
Reorganize object store metrics

rdettai:
### Description

Rationalize object store metrics. Be more exhaustive in request and error recording.

Important: This is a breaking change as some metrics are renamed.

### How was this PR tested?
TODO: add tests to the metrics wrappers

### DIFF
--- a/quickwit/quickwit-storage/src/metrics.rs
+++ b/quickwit/quickwit-storage/src/metrics.rs
@@ -16,7 +16,7 @@
 
 use once_cell::sync::Lazy;
 use quickwit_common::metrics::{
-    GaugeGuard, Histogram, IntCounter, IntCounterVec, IntGauge, new_counter, new_counter_vec,
+    GaugeGuard, HistogramVec, IntCounter, IntCounterVec, IntGauge, new_counter, new_counter_vec,
     new_gauge, new_histogram_vec,
 };
 
@@ -30,19 +30,13 @@ pub struct StorageMetrics {
     pub searcher_split_cache: CacheMetrics,
     pub get_slice_timeout_successes: [IntCounter; 3],
     pub get_slice_timeout_all_timeouts: IntCounter,
-    pub object_storage_get_total: IntCounter,
-    pub object_storage_get_errors_total: IntCounterVec<1>,
+    pub object_storage_requests_total: IntCounterVec<2>,
+    pub object_storage_request_duration: HistogramVec<2>,
     pub object_storage_get_slice_in_flight_count: IntGauge,
     pub object_storage_get_slice_in_flight_num_bytes: IntGauge,
-    pub object_storage_put_total: IntCounter,
-    pub object_storage_put_parts: IntCounter,
     pub object_storage_download_num_bytes: IntCounter,
+    pub object_storage_download_errors: IntCounterVec<1>,
     pub object_storage_upload_num_bytes: IntCounter,
-
-    pub object_storage_delete_requests_total: IntCounter,
-    pub object_storage_bulk_delete_requests_total: IntCounter,
-    pub object_storage_delete_request_duration: Histogram,
-    pub object_storage_bulk_delete_request_duration: Histogram,
 }
 
 impl Default for StorageMetrics {
@@ -63,31 +57,6 @@ impl Default for StorageMetrics {
         let get_slice_timeout_all_timeouts =
             get_slice_timeout_outcome_total_vec.with_label_values(["all_timeouts"]);
 
-        let object_storage_requests_total = new_counter_vec(
-            "object_storage_requests_total",
-            "Total number of object storage requests performed.",
-            "storage",
-            &[],
-            ["action"],
-        );
-        let object_storage_delete_requests_total =
-            object_storage_requests_total.with_label_values(["delete_object"]);
-        let object_storage_bulk_delete_requests_total =
-            object_storage_requests_total.with_label_values(["delete_objects"]);
-
-        let object_storage_request_duration = new_histogram_vec(
-            "object_storage_request_duration_seconds",
-            "Duration of object storage requests in seconds.",
-            "storage",
-            &[],
-            ["action"],
-            vec![0.1, 0.5, 1.0, 5.0, 10.0, 30.0, 60.0],
-        );
-        let object_storage_delete_request_duration =
-            object_storage_request_duration.with_label_values(["delete_object"]);
-        let object_storage_bulk_delete_request_duration =
-            object_storage_request_duration.with_label_values(["delete_objects"]);
-
         StorageMetrics {
             fast_field_cache: CacheMetrics::for_component("fastfields"),
             fd_cache_metrics: CacheMetrics::for_component("fd"),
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-storage/src/metrics.rs
```diff
@@ -124,35 +97,26 @@ impl Default for StorageMetrics {
...
-            ),
             object_storage_download_num_bytes: new_counter(
                 "object_storage_download_num_bytes",
                 "Amount of data downloaded from an object storage.",
                 "storage",
                 &[],
             ),
+            object_storage_download_errors: new_counter_vec(
+                "object_storage_download_errors",
+                "Number of download requests that received successfull response headers but \
```
fulmicoton-dd:
```suggestion
                "Number of download requests that received successful response headers but \
```

quickwit/quickwit-storage/src/object_storage/metrics_wrappers.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+#[pinned_drop]
+impl<F> PinnedDrop for DownloadMetricsWrapper<F>
+where F: Future<Output = io::Result<u64>>
+{
+    fn drop(self: Pin<&mut Self>) {
+        let status = match &self.result {
+            None => Cow::Borrowed("cancelled"),
+            Some(Err(e)) => Cow::Owned(format!("{e:?}")),
```
fulmicoton-dd: that seems a bit crazy to have a formatted error as a label value. The cardinality and length could be mad.
Of course, here your error is not an Error but an ErrorKind.

But 
1) this is not locally obvious
2) using a Cow suggests otherwise

So this is confusing to readers and it makes it easy for people to make mistakes when they modify your code.

Unfortunately, as_str is crate private.
Can we add a minimal implementation that match on the most importants error kind? A copy past of std is ok.

quickwit/quickwit-storage/src/metrics.rs
```diff
@@ -124,35 +97,26 @@ impl Default for StorageMetrics {
...
-            ),
-            object_storage_put_parts: new_counter(
-                "object_storage_puts_parts",
-                "Number of object parts uploaded.",
-                "",
-                &[],
-            ),
             object_storage_download_num_bytes: new_counter(
                 "object_storage_download_num_bytes",
                 "Amount of data downloaded from an object storage.",
```
fulmicoton-dd:
```suggestion
                // This does not include request that failed or were cancelled.
                "Amount of data successfully downloaded from an object storage.",
```

quickwit/quickwit-storage/src/object_storage/azure_blob_storage.rs
```diff
@@ -276,6 +276,7 @@ impl AzureBlobStorage {
                             .put_block(block_id.clone(), data)
                             .hash(hash)
                             .into_future()
+                            .with_count_metric("put_block")
```
fulmicoton-dd: is there a semantic difference with the s3 equivalents? I would rather have the same metric names. We may want to ship dashboards for instance.

quickwit/quickwit-storage/src/metrics_wrappers.rs
```diff
@@ -0,0 +1,282 @@
...
+use std::time::Instant;
+
+use pin_project::{pin_project, pinned_drop};
+use tokio::io::{AsyncBufRead, AsyncWrite};
+
+use crate::STORAGE_METRICS;
+
+pub enum Status {
+    Pending,
+    #[allow(dead_code)]
```
fulmicoton-dd, line 60:
```suggestion
    // only useful on feature="azure"
    #[allow(dead_code)]
```

quickwit/quickwit-storage/src/metrics_wrappers.rs
```diff
@@ -0,0 +1,427 @@
...
+impl ActionLabel {
+    fn as_str(&self) -> &'static str {
+        match self {
+            ActionLabel::AbortMultipartUpload => "abort_multipart_upload",
+            ActionLabel::CompleteMultipartUpload => "complete_multipart_upload",
+            ActionLabel::CreateMultipartUpload => "create_multipart_upload",
+            ActionLabel::DeleteObject => "delete_object",
+            ActionLabel::DeleteObjects => "delete_objects",
+            ActionLabel::GetObject => "get_object",
+            ActionLabel::HeadObject => "head_object",
```
rdettai, line 49: Azure: get_properties
S3: head_object
Opendal: stat

quickwit/quickwit-storage/src/metrics_wrappers.rs
```diff
@@ -0,0 +1,426 @@
...
+                let i = ready!(Pin::new(&mut *me.writer).poll_write(cx, buffer))?;
+                if i == 0 {
+                    return Poll::Ready(Err(std::io::ErrorKind::WriteZero.into()));
+                }
+                self.amt += i as u64;
+                Pin::new(&mut *self.reader).consume(i);
+            }
+        }
+    }
+}
```
rdettai, line 404: It was actually pretty simple to have an estimate for the number of bytes already downloaded when the error occurred, so I added it.

quickwit/quickwit-storage/src/metrics.rs
```diff
@@ -97,62 +66,63 @@ impl Default for StorageMetrics {
...
+                 recorded when the response headers are returned, download failures will not \
+                 appear as errors.",
                 "storage",
                 &[],
+                ["action", "status"],
             ),
-            object_storage_get_errors_total: new_counter_vec::<1>(
-                "object_storage_get_errors_total",
-                "Number of GetObject errors.",
+            object_storage_request_duration: new_histogram_vec(
```
fulmicoton-dd, line 78: if it is ttfb it should be not be called duration!?

(technically I don't think this is ttfb btw :-/ )

rdettai-sk: Yes, I agree the terminology TTFB is a bit stretched here as it also applies to other types of queries (deletes). And it's probably not a good idea to call it request duration for GET requests when it doesn't measure the whole download duration.

I still think it's valuable to measure the time it takes for get requests to get to the byte stream. It's valuable when trying to configure the StorageTimeoutPolicy, and it's also a way to measure the read performance of the object store without having to normalize by the downloaded size. Would you be fine with calling that separate metric `object_storage_get_ttfb` or do you still think it's not appropriate?

fmassot: @rdettai what are we measuring exactly?

fulmicoton-dd: @fmassot The storage SDK returning a ByteStream.
We can hope most of the meta data layer stuff is done at this point.

@rdettai-sk 

> it's also a way to measure the read performance of the object store without having to normalize by the downloaded size

My comment had two parts.
- the first one is: you called the metric duration. This needs to be renamed.
- the ttfb part it does not matter much and I agree your metric is useful. I just don't want people to start comparing hyperscalers and drawing wrong conclusions due to minor implementation differences. I would edit the comment to `ttfb-like` for instance.

I think we need to rename that metric and keep it...
And then, count + duration total + num bytes total.

With that, at least we have enough to compute a model 
`TTFB + bytes * throughput`.

rdettai-sk: To summarise the changes on existing metrics:
- I'll add a new metric called object_storage_get_slice_latency -> `help: "TTFB-like latency of the object store get requests in seconds. This doesn't include the time to download the full response payload."`
- object_storage_request_duration is used only for **other** object store requests (only delete for now) with an updated description -> `help: "Duration in seconds of the object store request"`

I can add a metric to measure the total duration (download included). Should it be a counter or a histogram? 
- A histogram would be hard to leverage because you loose the connection to the actual download size. If we want the download throughput distribution, I think a histogram of the download throughput (wrapping the download itself with a timer) would be easier to use. 
- If we opt for a counter of the total duration, how would you use it to estimate the throughput?
- We have retries when receiving an HTTP error response (e.g when receiving 427 slow down). I think failed attempts should not be included in the duration.

PR #5827, opened 2025-07-02
Print memory profiling logs even when OTLP exports are enabled

rdettai:
Tested 
- [x] that logs are properly printed with and without the profiler feature
- [x] reload works without the profiler feature and returns an error with the profiler feature
   - fixed the reload error message which was incorrect, invalid filters are usually silent and the error might be something else
- [x] otlp export with and without the profiler feature

### DIFF
--- a/quickwit/quickwit-cli/src/logger.rs
+++ b/quickwit/quickwit-cli/src/logger.rs
@@ -231,7 +231,7 @@ pub(super) mod jemalloc_profiled {
 
     /// An event formatter specific to the memory profiler output.
     ///
-    /// Also displays a backtrace after spans and the fields of the tracing
+    /// Also displays a backtrace after the spans and fields of the tracing
     /// event (into separate lines).
     struct ProfilingFormat {
         time_formatter: UtcTime<Vec<BorrowedFormatItem<'static>>>,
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-cli/src/logger.rs
```diff
@@ -63,9 +65,31 @@ pub fn setup_logging_and_tracing(
...
+
+    let event_format = EventFormat::get_from_env();
+    let fmt_fields = event_format.format_fields();
     let registry = tracing_subscriber::registry();
+
+    let (reloadable_env_filter, reload_handle) = ReloadLayer::new(startup_env_filter(level)?);
+
+    #[cfg(not(feature = "jemalloc-profiled"))]
+    let registry = registry.with(reloadable_env_filter).with(
+        tracing_subscriber::fmt::layer()
```
fulmicoton-dd, line 77: I don't understand why there is a difference here.
Can you comment? dynamic reload does not work with jemalloc?

rdettai: This is commented a few lines below, when the reload is actually ignored: https://github.com/quickwit-oss/quickwit/blob/876ce2a0b2b723f0d1ece39e8f33944bd0cf2596/quickwit/quickwit-cli/src/logger.rs#L306-L317

I improved the comment a bit.

fulmicoton-dd: ah ok thank you

fulmicoton-dd (approved):

PR #5829, opened 2025-07-04
Attempt to migrate warp to axum.

fmassot:
I started the migration work to test cursor capabilities. As the result was not so bad, I iterated a bit o clean up some horrible pieces of code. And there is the PR.

I expect there is still a lot of refactor to do as I did not have the time to read thoroughtly the axum documentation and find the "right" axum to implement a given handler. I'm notably thinking about how decompression is done.

I did not check lambda and I suggest to remove the module and we don't really want to maintain it for 0.9.

All tests are passing.


scMarkus:
@fmassot is this still progerssing? I came here from #5842

PR #5843, opened 2025-07-15
Add a hard limit on field list size

rdettai:
### Description

Searchers sometime OOM when returning too many fields (~millions). This PR adds a hard limit on the number of fields that can be returned.
- configured through an env variable QW_FIELD_LIST_SIZE_LIMIT
- if there are more fields than the limit, a 500 error is returned

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-search/src/list_fields.rs
+++ b/quickwit/quickwit-search/src/list_fields.rs
@@ -21,6 +21,7 @@ use anyhow::Context;
 use futures::future;
 use futures::future::try_join_all;
 use itertools::Itertools;
+use once_cell::sync::OnceCell;
 use quickwit_common::shared_consts::SPLIT_FIELDS_FILE_NAME;
 use quickwit_common::uri::Uri;
 use quickwit_metastore::SplitMetadata;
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -184,6 +198,12 @@ fn merge_leaf_list_fields(
                 flush_group(&mut responses, &mut current_group);
             }
         }
+        if responses.len() >= get_field_list_size_limit() {
+            return Err(SearchError::Internal(format!(
```
PSeitz, line 202: We could return a truncated list instead

rdettai: My concern with a truncated list is that it doesn't clearly indicate that something went wrong and that results are not deterministic. This is really a short term patch to prevent excessive memory usage, but I think we should work on a proper solution that correctly prunes the long tail of "rare" fields.

fulmicoton-dd: I think the part missing here, is that the call fails anyway today.

PR #5844, opened 2025-07-16
Make in-flight storage get requests a memory metric

rdettai:
### Description

Generate an updated metrics docs using copilot. Most metrics have been checked manually and comments have been added. Some of them where not verified, like OTLP and Jaeger metrics. In any case, it cannot be worse that the current doc that is completely outdated.

Also removed the `get_slice_in_flight_*` metrics to have these memory allocations measured as part of `quickwit_memory_in_flight`. This makes it more discoverable and usable.

### How was this PR tested?

Tested some of the metrics on a running instance.


quickwit/quickwit-ingest/src/metrics.rs
```diff
@@ -13,18 +13,14 @@
...
 pub struct IngestMetrics {
+    // With ingest V1 all ingested documents are considered valid
     pub ingested_docs_bytes_valid: IntCounter,
+    pub ingested_docs_valid: IntCounter,
     pub ingested_docs_bytes_invalid: IntCounter,
     pub ingested_docs_invalid: IntCounter,
-    pub ingested_docs_valid: IntCounter,
-
-    pub replicated_num_bytes_total: IntCounter,
-    pub replicated_num_docs_total: IntCounter,
```
rdettai, line 25: moved this to be part of of ingest_v2 metrics

quickwit/quickwit-ingest/src/metrics.rs
```diff
@@ -13,18 +13,14 @@
...
-    #[allow(dead_code)] // this really shouldn't be dead, it needs to be used somewhere
```
rdettai, line 26: this has been dead for a year

quickwit/quickwit-storage/src/metrics.rs
```diff
@@ -32,8 +32,6 @@ pub struct StorageMetrics {
     pub get_slice_timeout_all_timeouts: IntCounter,
     pub object_storage_requests_total: IntCounterVec<2>,
     pub object_storage_request_duration: HistogramVec<2>,
-    pub object_storage_get_slice_in_flight_count: IntGauge,
-    pub object_storage_get_slice_in_flight_num_bytes: IntGauge,
```
rdettai, line 36: - I created this metric to make sure that this wasn't the reason for OOMs on search and I confirmed it's not the case
- I think it makes sense to keep tracking the memory usage, but move it to quickwit_memory_in_flight where it is more discoverable and actionable

Issue #3142, opened 2023-04-05
Add sort_by_field examples in the docs


fmassot:
At least an example to sort on a timestamp field is just a big missing piece.

ph:
I am currently working on this and learning about it, I am currently running quickwit in a docker container and using minio as the storage layer to give some background.

I've found a few gotcha or interrogation?

 1. We cannot sort on a `string` yet based on #1454 and #1450
 2. To make sorting work, I have to explicitly mark them as `fast` in the index mapping? Noticed the warning when using an `i64` and a `datetime` field. Is that expected?
 ```json
{"message":"Invalid query: Sort by field must be a fast field | please add the fast property to your field `points`."} 
```
 3. Looking at the Swagger API from the quickwit-ui I should be able to use the following format `{ "sort_by_field" : { "name":  "points", "order" : "asc" }` but it complains see the following example: 
```sh
curl -XPOST "http://o.local:7280/api/v1/friends/search" -H 'Content-Type: application/json' -d '{
                                                                   "query": "name:caro",
                                                                   "sort_by_field": {
                                                                        "field_name": "name",
                                                                        "order": "ASC"
                                                                    }
                                                               }'
```
```json
{"message":"Request body deserialize error: invalid type: map | expected a string at line 3 column 84"}
```

Following seems to work:

```sh
curl -XPOST "http://o.local:7280/api/v1/friends/search" -H 'Content-Type: application/json' -d '{
                                                                    "query": "name:Caroline",
                                                                    "sort_by_field": "last_modified_at"
                                                               }'
```

I can make issue for 3, I just want to get a bit more information before jumping into this. Thanks

PR #5845, opened 2025-07-16
Add sorting order description to sort_by field in REST API docs

earlbread:
### Description

This updates the `sort_by` field description in the REST API documentation to clarify how sorting order is handled.
Related to #3142.

### How was this PR tested?

Documentation update only.


rdettai:
I was working on exactly the same 😸 https://github.com/quickwit-oss/quickwit/pull/5846

PR #5857, opened 2025-08-08
fix: Use appropriate Term method for date search

Darkheir:
### Description

Tantivy's PR https://github.com/quickwit-oss/tantivy/pull/2456 removed date truncation to the second in method `from_field_date` and introduced a new method `from_field_date_for_search` for this purpose.

> Fixes date truncation issue in Term, where DateTime was truncated to seconds precision unconditionally, but we only want to apply that truncation on the inverted index.

It seems that the method used by Quickwit to generate the Term struct during searches was never updated.

### How was this PR tested?

Added a unit test

### DIFF
--- a/quickwit/quickwit-query/src/query_ast/utils.rs
+++ b/quickwit/quickwit-query/src/query_ast/utils.rs
@@ -143,7 +143,7 @@ fn compute_query_with_field(
         }
         FieldType::Date(_) => {
             let dt = parse_value_from_user_text(value, field_entry.name())?;
-            let term = Term::from_field_date(field, dt);
+            let term = Term::from_field_date_for_search(field, dt);
             Ok(make_term_query(term))
         }
         FieldType::Str(text_options) => {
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-query/src/query_ast/utils.rs
```diff
@@ -143,7 +143,7 @@ fn compute_query_with_field(
         }
         FieldType::Date(_) => {
             let dt = parse_value_from_user_text(value, field_entry.name())?;
-            let term = Term::from_field_date(field, dt);
```
PSeitz-dd, line 146: @trinity-1686a Is this code path only used for the inverted index?

In the inverted index we want to truncate, but not in the columnar storage. Not sure where the best place to truncate is, when creating the query, or when executing it.

trinity-1686a: currently yes, term queries are always sent toward the inverted index

i'm thinking of also being able to do term queries on fast fields when the inverted index is absent, so people can drop the inverted index for workload were it sees almost no usage, but can still run point queries from time to time, but that doesn't exist yet

PSeitz-dd (approved):

PSeitz-dd:
@Darkheir Thanks for the PR!

PR #5874, opened 2025-09-03
feat: Support regex in user input queries

Darkheir:
### Description

Update tantivy to add support for regexes inside user input queries.

### How was this PR tested?

Added a unit test


quickwit/Cargo.toml
```diff
@@ -1,72 +1,72 @@
 [workspace]
```
guilload, line 1: Please remove the reformatting of this file from this PR. This is too noisy.

Darkheir: Done

Darkheir:
> your PR looks good, but it conflicts with current main. Can you update it so we can get it merged?

Thank you for the review, I just rebased my PR :)

PR #5886, opened 2025-09-09
Remove search stream endpoint

guilload:


quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -322,8 +324,8 @@ impl Default for SearcherConfig {
             fast_field_cache_capacity: ByteSize::gb(1),
             split_footer_cache_capacity: ByteSize::mb(500),
             partial_request_cache_capacity: ByteSize::mb(64),
-            max_num_concurrent_split_streams: 100,
             max_num_concurrent_split_searches: 100,
+            _max_num_concurrent_split_streams: None,
```
Abdul-Andha, line 328: Why do we keep this in `SearcherConfig` if search streams is no longer supported?

Abdul-Andha (approved):
looks good, I just left a question

quickwit/quickwit-config/src/node_config/mod.rs
guilload: `SearcherConfig` has the following `serde` attribute: `#[serde(deny_unknown_fields)]`. It means that when attempting to parse a searcher config, `serde` will error during deserialization when encountering unknown fields. If we remove the field totally, this won't be backward compatible, so we just ignore it.

dikuchan:
Hi! We are one of the users affected by this removal.

We use Quickwit alongside ClickHouse. Our backend already relies on `/search/stream`, and we had also started working on querying Quickwit directly from ClickHouse through `url()` tables, exactly the integration described in [Quickwit’s own documentation](https://quickwit.io/docs/0.8.1/guides/add-full-text-search-to-your-olap-db).

We discovered this while preparing to upgrade to 0.9. We initially missed this breaking change. While the removal is listed under "Removed" section in the changelog, it is absent from both the "Breaking / Migration" section and the [upgrade guide](https://quickwit.io/docs/get-started/upgrade). Neither provides migration guidance for this documented integration.

We process substantial data volumes, and this endpoint is a real dependency for us. Reading that “nobody is using this seriously as far as we know” is therefore quite frustrating. IMO users can depend on an advertised feature without regularly reporting that dependency on GitHub.

I understand that maintaining this endpoint has a cost. However, publishing an integration guide encourages people to make architectural decisions around that functionality. Removing it leaves those users with significant work to recover a capability the project explicitly promoted.

Search/scroll is not a drop-in replacement here when results span multiple pages: a ClickHouse URL table cannot follow scroll cursors on its own. That requires an additional adapter, with its own maintenance and performance implications.

What is the intended migration path for this documented integration?

PR #5888, opened 2025-09-10
Fixing the counter for indexing pipelines

fulmicoton:
It was stuck to 0 because I forgot to increment the guard upon creation.


quickwit/quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -159,7 +159,9 @@ impl IndexingPipeline {
         let indexing_pipelines_gauge = crate::metrics::INDEXER_METRICS
             .indexing_pipelines
             .with_label_values([&params.pipeline_id.index_uid.index_id]);
-        let indexing_pipelines_gauge_guard = OwnedGaugeGuard::from_gauge(indexing_pipelines_gauge);
+        let mut indexing_pipelines_gauge_guard =
+            OwnedGaugeGuard::from_gauge(indexing_pipelines_gauge);
+        indexing_pipelines_gauge_guard.add(1);
```
PSeitz-dd: nitpick: I think the guard should do the increment in the constructor

PSeitz-dd (approved):

quickwit/quickwit-indexing/src/actors/indexing_pipeline.rs
fulmicoton: agreed. i made the change

PR #5890, opened 2025-09-12
Added different ways for users to set the number of vcpus.

fulmicoton:
The following methods are used in order:
- from the `QW_NUM_CPUS` environment variable
- from the `KUBERNETES_LIMITS_CPU` environment variable
- from the operating system
-  default to 2.

### DIFF
--- a/quickwit/quickwit-common/src/cpus.rs
+++ b/quickwit/quickwit-common/src/cpus.rs
@@ -1,0 +1,137 @@
+// Copyright 2021-Present Datadog, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-common/src/cpus.rs
```diff
@@ -0,0 +1,128 @@
...
+        .ok();
+    let num_cpus_from_env_opt = get_num_cpus_from_env(QW_NUM_CPUS_ENV_KEY);
+    let num_cpus_from_k8s_limit = get_num_cpus_from_env(KUBERNETES_LIMITS_CPU);
+
+    if let Some(num_cpus) = num_cpus_from_env_opt {
+        return num_cpus;
+    }
+
+    if let Some(num_cpus_from_k8s_limit) = num_cpus_from_k8s_limit {
+        info!("num cpus from k8s limit: {},  possibly overriding os value {:?}", num_cpus_from_k8s_limit, num_cpus_from_env_opt);
```
PSeitz-dd: we could compare them and print a warning instead if num_cpus_from_os_opt is set and they are different

fulmicoton: hmm that seems overkill. We want an info statement anyway.

PR #5892, opened 2025-09-12
update tantivy

PSeitz-dd:


quickwit/Cargo.toml
```diff
@@ -346,7 +346,7 @@ quickwit-serve = { path = "quickwit-serve" }
 quickwit-storage = { path = "quickwit-storage" }
 quickwit-telemetry = { path = "quickwit-telemetry" }
 
-tantivy = { git = "https://github.com/quickwit-oss/tantivy/", rev = "80f5f1e", default-features = false, features = [
+tantivy = { git = "https://github.com/quickwit-oss/tantivy/", rev = "7963b0b", default-features = false, features = [
```
otetard, line 349: The `columnar-zstd-compression` feature should also be added (this is not the default anymore since https://github.com/quickwit-oss/tantivy/pull/2633)

PSeitz: No, this is still the default, but default features are disabled

fulmicoton: I think otetard is correct

fulmicoton: ah yes you added it

PR #5898, opened 2025-09-15
    Disabling the lifo slot and variable shard load by default

fulmicoton-dd:


PSeitz-dd (approved):

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -121,7 +123,30 @@ impl fmt::Debug for IndexingScheduler {
 fn enable_variable_shard_load() -> bool {
     static IS_SHARD_LOAD_CP_ENABLED: OnceCell<bool> = OnceCell::new();
     *IS_SHARD_LOAD_CP_ENABLED.get_or_init(|| {
-        !quickwit_common::get_bool_from_env("QW_DISABLE_VARIABLE_SHARD_LOAD", false)
```
PSeitz-dd, line 124: In the future I think this pattern would be a bit simpler
`QW_VARIABLE_SHARD_LOAD=true|enable`
`QW_VARIABLE_SHARD_LOAD=false|disable`

fulmicoton-dd: I kind of disagree. 

But I start thinking that the _DISABLE stuff was not a good idea. (The point was to suggest that the default was enabling of course, but it really sucks when we want to change the default)

PR #5899, opened 2025-09-15
Adding an environment to configure the default load per shard.

fulmicoton-dd:


quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -177,10 +177,21 @@ fn compute_load_per_shard(shard_entries: &[&ShardEntry]) -> NonZeroU32 {
...
-        NonZeroU32::new(PIPELINE_FULL_CAPACITY.cpu_millis() / 4).unwrap()
+        get_default_load_per_shard()
     }
 }
 
+fn get_default_load_per_shard() -> NonZeroU32 {
+    static DEFAULT_LOAD_PER_SHARD: OnceLock<NonZeroU32> = OnceLock::new();
+    *DEFAULT_LOAD_PER_SHARD.get_or_init(|| {
+        let default_load_per_shard = quickwit_common::get_from_env(
+            "QW_DEFAULT_LOAD_PER_SHARD",
```
PSeitz-dd, line 188: it's not super clear what load means here
`QW_DEFAULT_CPU_PER_SHARD` ?

PR #5905, opened 2025-09-17
add migration for in-metastore kv store

trinity-1686a:


quickwit/quickwit-metastore/migrations/postgresql/24_add-arbitrary-kv.down.sql
```diff
@@ -0,0 +1,4 @@
+DROP TABLE kv (
+    key VARCHAR(50) PRIMARY KEY,
```
fulmicoton-dd: it is the first time I see the schema joined in a drop table statement. Is this a thing?

trinity-1686a: no it's not, it's a copy past that was too lenient

fulmicoton:
closed for https://github.com/quickwit-oss/quickwit/pull/5920

PR #5906, opened 2025-09-17
handle fast field search

PSeitz-dd:
Warm up fast fields that are searched and not indexed
add fast field search test

### DIFF
--- a/quickwit/Cargo.lock
+++ b/quickwit/Cargo.lock
@@ -5595,7 +5595,7 @@ checksum = "1a80800c0488c3a21695ea981a54918fbb37abf04f4d0720c453632255e2ff0e"
 [[package]]
 name = "ownedbytes"
 version = "0.9.0"
-source = "git+https://github.com/quickwit-oss/tantivy/?rev=7963b0b#7963b0b4aa66dda2fcff83f8bc38f3ff5c4faa85"
+source = "git+https://github.com/quickwit-oss/tantivy/?rev=203751f#203751f2fe83488c3578fa84188fc10e55ef5006"
 dependencies = [
  "stable_deref_trait",
 ]
@@ -9536,7 +9536,7 @@ checksum = "7b2093cf4c8eb1e67749a6762251bc9cd836b6fc171623bd0a9d324d37af2417"
 [[package]]
 name = "tantivy"
 version = "0.25.0"
-source = "git+https://github.com/quickwit-oss/tantivy/?rev=7963b0b#7963b0b4aa66dda2fcff83f8bc38f3ff5c4faa85"
+source = "git+https://github.com/quickwit-oss/tantivy/?rev=203751f#203751f2fe83488c3578fa84188fc10e55ef5006"
 dependencies = [
  "aho-corasick",
  "arc-swap",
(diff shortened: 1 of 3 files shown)


quickwit/rest-api-tests/scenarii/qw_search_api/0005_fast_field_search.yaml
```diff
@@ -0,0 +1,15 @@
...
+expected:
+  num_hits: 1
+---
+# Non-matching exact value should return no hits
+endpoint: nested/search
+method: POST
+json:
+  query: "text_fast:fast_text_value_dashes"
+expected:
+  num_hits: 0
```
trinity-1686a, line 15: can you add a piece of doc somewhere that when we end up using the fastfield for term queries, it behave like if indexed with the raw tokenizer?

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -45,12 +45,62 @@ impl<'a> QueryAstVisitor<'a> for RangeQueryFields {
...
-    fields: HashSet<FastFieldWarmupInfo>,
+/// Term Queries on fields which are fast but not indexed.
+struct TermSearchOnColumnar<'f> {
+    fields: &'f mut HashSet<FastFieldWarmupInfo>,
     schema: Schema,
 }
+impl<'a, 'f> QueryAstVisitor<'a> for TermSearchOnColumnar<'f> {
+    type Err = Infallible;
+
+    fn visit_term_set(&mut self, term_set_query: &'a TermSetQuery) -> Result<(), Infallible> {
```
trinity-1686a, line 56: is there a test covering TermSetQuery on fastfield?

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -109,7 +152,18 @@ pub(crate) fn build_query(
...
+    }
+    .visit(query_ast);
+
+    let _: Result<(), Infallible> = ExistsQueryFastFields {
+        fields: &mut fast_fields,
+        schema: schema.clone(),
+    }
+    .visit(query_ast);
 
     let query = query_ast.build_tantivy_query(
```
trinity-1686a: `let Ok(_) = infallible_statement()` didn't work on old rust, but it does now, i think that's a better way to show the code is infallible than calling expect

trinity-1686a (approved):
some nits, lgtm overall

quickwit/quickwit-doc-mapper/src/query_builder.rs
PSeitz-dd: I added a test

quickwit/quickwit-doc-mapper/src/query_builder.rs
PSeitz-dd: updated to `Ok(_)`

quickwit/rest-api-tests/scenarii/qw_search_api/0005_fast_field_search.yaml
PSeitz-dd: I opened an issue for it

PR #5919, opened 2025-09-24
Fix azure multipart upload data corruption

shrijeet:
### Description

Blocks were added to BlockList in completion order rather than sequential order, causing data corruption. Zero pad block IDs and sort before committing to ensure correct blob reconstruction.

Also improve the existing test that was only checking file, the test would fail on main if integrity check was turned on.

### How was this PR tested?

This PR passes the unit test, also has been runnin in for last two days in production without data corruption, prior to this PR - split was getting corrupted every 45 minutes.

### DIFF
--- a/quickwit/quickwit-storage/src/lib.rs
+++ b/quickwit/quickwit-storage/src/lib.rs
@@ -383,9 +383,28 @@ pub(crate) mod test_suite {
     #[cfg(feature = "integration-testsuite")]
     pub async fn storage_test_multi_part_upload(storage: &mut dyn Storage) -> anyhow::Result<()> {
         let test_path = Path::new("hello_large.txt");
-        let test_buffer = vec![0u8; 15_000_000];
-        storage.put(test_path, Box::new(test_buffer)).await?;
+
+        let mut test_buffer = Vec::with_capacity(15_000_000);
+        for i in 0..15_000_000u32 {
+            test_buffer.push((i % 256) as u8);
+        }
+
+        storage
+            .put(test_path, Box::new(test_buffer.clone()))
+            .await?;
+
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-storage/src/lib.rs
```diff
@@ -383,9 +383,26 @@ pub(crate) mod test_suite {
...
     pub async fn storage_test_multi_part_upload(storage: &mut dyn Storage) -> anyhow::Result<()> {
         let test_path = Path::new("hello_large.txt");
-        let test_buffer = vec![0u8; 15_000_000];
+
+        let mut test_buffer = Vec::with_capacity(15_000_000);
+        for i in 0..15_000_000u32 {
+            test_buffer.push((i % 256) as u8);
+        }
+
+        let expected_hash = md5::compute(&test_buffer);
```
guilload: nit: I don't think we need to compute an MD5 digest, we can just clone `test_buffer` to keep it around and do a byte-compare  with the downloaded data.

quickwit/quickwit-storage/tests/azure_storage.rs
```diff
@@ -45,7 +45,7 @@ async fn azure_storage_test_suite() -> anyhow::Result<()> {
 
     object_storage.set_policy(MultiPartPolicy {
         // On azure, block size is limited between 64KB and 100MB.
-        target_part_num_bytes: 5 * 1_024 * 1_024, // 5MB
+        target_part_num_bytes: 2 * 1_024 * 1_024, // 2MB
```
guilload: What is informing this change?

guilload:
```suggestion
        target_part_num_bytes: 2 * 1_024 * 1_024, // 2MiB
```

PR #5922, opened 2025-09-28
source: preserve Position arg in ObjectUriBatchReader

noam-sol:
### Description
Before the fix, we lose the granularity between Beginning and offset 0 which
affects the output checkpoint_delta.

### How was this PR tested?
Added a UT

### DIFF
--- a/quickwit/quickwit-indexing/src/source/doc_file_reader.rs
+++ b/quickwit/quickwit-indexing/src/source/doc_file_reader.rs
@@ -191,6 +191,7 @@ impl ObjectUriBatchReader {
         }
         let limit_num_bytes = self.current_offset + BATCH_NUM_BYTES_LIMIT as usize;
         let mut new_offset = self.current_offset;
+        let mut eof_position: Option<Position> = None;
         while new_offset < limit_num_bytes {
             if let Some(record) = source_progress
                 .protect_future(self.reader.next_record())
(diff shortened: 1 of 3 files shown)


guilload:
This is not the right fix. It's a bug in the checkpoint delta logic in the source. If you print the batch checkpoint deltas in that unit test, you will see something like:
```rust
for batch in batches {
            println!("batch checkpoint delta: {:?}", batch.checkpoint_delta);
        }
```
```
batch checkpoint delta: ∆(file:///var/folders/cj/3nfybfc572b9s2qvd18qzh_m0000gp/T/.tmpDz99t1:(00000000000000000000..~00000000000000000350])
batch checkpoint delta: ∆(file:///var/folders/cj/3nfybfc572b9s2qvd18qzh_m0000gp/T/.tmp6o6nbz:(00000000000000000070..~00000000000000000350])
```

The first checkpoint delta should be `(beginning..~00000000000000000350`].

Do you want to take a stab at it?

noam-sol:
ahh I see, on this line - https://github.com/quickwit-oss/quickwit/blob/a58b6de9542db62455888dfc9de65da901c64831/quickwit/quickwit-indexing/src/source/doc_file_reader.rs#L161

we lose the granularity between Beginning and offset 0; however I don't feel like properly fixing it is worth the risk of breaking something

Issue #5937, opened 2025-10-20
Dependencies not pinned by hash


osyniakov:
Dependencies not pinned by hash pose a supply chain security risk.

Pinning by hash ensures that the build will only accept the exact, verified files.

This typically involves reviewing and updating:

- GitHub Actions: Pinning all third-party GitHub Actions to a full commit SHA (e.g., actions/checkout@v4 becomes actions/checkout@<FULL_SHA_HASH>).

- Container Images: Pinning images in Dockerfiles/Compose files to their digest (e.g., image: latest or image: v1 becomes image@sha256:<DIGEST>).
 
- Build/Runtime Dependencies: Ensuring all package manifest files (package.json, requirements.txt, go.mod, etc.) use a lockfile that includes dependency hashes, or updating the dependencies themselves to include a verifiable hash.

PR #5949, opened 2025-10-24
Pin Python packages in workflows. See #5937

osyniakov:
### Description

Pin Python packages in GitHub workflows by hash to ensure reliable production builds.

- Use actions/setup-python to set up a Python version in sync with Pipfile
- Use pipenv

### How was this PR tested?

n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,7 @@ jobs:
       - name: Install python packages
         if: always() && steps.modified.outputs.rust_src == 'true'
         run: |
-          pip install --user pipenv
+          pip install --user pipenv==2025.0.4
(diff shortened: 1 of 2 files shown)


osyniakov:
@guilload could you please check this fix as well?

guilload:
This is going to be a bit of work to maintain each time we add or upgrade a dependency. Can we have CI use pipenv instead and rely on the same Pipfile / Pipfile.lock files?

osyniakov:
@guilload I pinned the pipenv version, which is definitely an improvement, but it looks like the scorecard issue persists. It would be nice if you could take one more look and merge it.

guilload (approved):

PR #5965, opened 2025-11-04
Pin pipenv and it's deps by hash in workflows. See #5937

osyniakov:
### Description

Pinned pipenv and its dependencies by hash to make scorecard check pass while waiting for a reply regarding best practices https://github.com/ossf/scorecard/discussions/4834

### How was this PR tested?

n/a


osyniakov:
@guilload could you please check this one?

guilload:
Is there a public place to see the scorecard? Is there way to get notified if we break something and we are no longer compliant?
Would you quickly document what you did to secure the repo / software in some SECURITY.md file at the root of the repo for instance?

PR #5971, opened 2025-11-04
Support ignore_unavailable query param

rdettai-sk:
### Description

ignore_unavailable is an ES query param that enables skipping indexes that don't exist without throwing an error. It is false by default (aligns with the current behavior which is to throw an error).

This PR enables the support for this query param in both _search and _msearch

### How was this PR tested?

Integration tests, running on both ES and QW


quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -243,6 +243,10 @@ message SearchRequest {
   optional PartialHit search_after = 16;
 
   CountHits count_hits = 17;
+
+  // When an exact index_id is provided (not a pattern), the query fails if that
```
guilload:
```suggestion
  // When an exact index ID is provided (not a pattern), the query fails only if that index is not found and this parameter is set to `false`.
```

quickwit/quickwit-search/src/root.rs
```diff
@@ -1243,7 +1249,12 @@ pub async fn search_plan(
         .deserialize_indexes_metadata()
         .await?;
 
-    check_all_index_metadata_found(&indexes_metadata[..], &search_request.index_id_patterns[..])?;
+    if !search_request.ignore_missing_indexes {
+        check_all_index_metadata_found(
+            &indexes_metadata[..],
+            &search_request.index_id_patterns[..],
+        )?;
+    }
```
guilload, line 1257: Please add tests in the file. Tedious but doable :) You can also push the `ignore_missing_indexes` parameter to `check_all_index_metadata_found` and just unit test that function instead. That would make me happy too :)

quickwit/quickwit-serve/src/elasticsearch_api/model/multi_search.rs
```diff
@@ -50,6 +51,9 @@ pub struct MultiSearchQueryParams {
     pub ignore_throttled: Option<bool>,
     #[serde(default)]
     pub ignore_unavailable: Option<bool>,
+    #[serde_as(deserialize_as = "OneOrMany<_, PreferMany>")]
+    #[serde(default)]
+    pub index: Vec<String>,
```
guilload: Can you add a comment such as `List of indexes to search`, rename the variable to `indexes` and use `#[serde(rename = "index")]` (assuming it works with `serde_as`.

quickwit/rest-api-tests/scenarii/es_compatibility/0025-msearch.yaml
```diff
@@ -104,3 +104,25 @@ expected:
...
+method: POST
+ndjson:
+  - {"index":"idontexist", "ignore_unavailable": true}
+  - {"query" : {"match" : { "type": "PushEvent"}}, "size": 0}
+expected:
+  responses:
+    - hits:
+        total:
+          value: 0
+      status: 200
```
guilload: nit: please add missing new line

quickwit/rest-api-tests/scenarii/es_compatibility/multi-indices/0004-missing_index_query.yaml
```diff
@@ -0,0 +1,30 @@
...
+params:
+  q: "*"
+  ignore_unavailable: "true"
+expected:
+  hits:
+    total:
+      value: 4
+
+
+
```
guilload: Please, remove the extra empty lines and insert the missing new line. Your editor can do that automatically.

quickwit/quickwit-search/src/root.rs
rdettai-sk: I have some philosophical doubts about a `check_all_index_metadata_found()` function that takes an `ignore_missing_indexes` parameter 😅

guilload: I don't really care as long as the logic is tested in this file.

rdettai-sk: done with proper tests of the calling functions

PR #5973, opened 2025-11-05
Split cbench workflow to remediate security concern

osyniakov:
#### Problem

The original cbench.yml workflow used pull_request_target trigger and checked out untrusted code from pull requests, then executed it on a self-hosted runner. This is a dangerous pattern because:

- pull_request_target runs with write permissions in the context of the base repository
- Checking out PR code (github.event.pull_request.head.sha) allows execution of untrusted code
- Self-hosted runners are particularly sensitive as they may have access to internal resources

Even with authorization checks, this pattern is considered dangerous by OSSF because the check itself is in the untrusted workflow file.

#### Solution

The CBENCH workflow has been split into two separate workflows to address OSSF Scorecard's "Dangerous Workflow" security concern. See https://scorecard.dev/viewer/?uri=github.com/quickwit-oss/quickwit

### How was this PR tested?

Testing is pending due to the unavailability of the self-hosted runner and benchmarking code.


osyniakov:
see https://github.com/quickwit-oss/quickwit/pull/5975

PR #5981, opened 2025-11-13
format js files in quickwit-ui

Platane:
### Description

I noticed that some js files don't respect the formatting configured (prettier)

I run a pass of formatter, which unfortunately touch a lot of files

I also took the opportunity to replace prettier with biome which is a more mature tool.

### How was this PR tested?

na

### DIFF
--- a/.github/workflows/ui-ci.yml
+++ b/.github/workflows/ui-ci.yml
@@ -38,6 +38,8 @@ jobs:
               yarn --cwd quickwit-ui cypress run
           - name: Lint
             command: yarn --cwd quickwit-ui lint
+          - name: Check formatting
+            command: yarn --cwd quickwit-ui check-formatting
           - name: Unit Test
             command: yarn --cwd quickwit-ui test
     services:


.github/workflows/ui-ci.yml
```diff
@@ -38,6 +38,8 @@ jobs:
               yarn --cwd quickwit-ui cypress run
           - name: Lint
             command: yarn --cwd quickwit-ui lint
+          - name: Check formatting
+            command: yarn --cwd quickwit-ui biome format
```
guilload: I'm not familiar with biome but usually linter and formater have a dedicated `check` command or `--fail` parameter that make the process return a non-zero status when exiting and fail the job

Platane: biome does have a check command to do format + lint (that's for a future PR)

by default it will fail if the formatting or linting is wrong, fixing files is done with a `--write` flag

guilload (approved):

PR #5984, opened 2025-11-14
Add `validate_docs` parameter to ingest settings

guilload:
### Description
Add `validate_docs` parameter to ingest settings. Default to `true`. We have a use case for this parameter at Datadog. This could also be used for OTEL indexes and by users who know that their documents are always well-formed to shave a few CPU cycles.

### How was this PR tested?
Updated index config unit tests

### DIFF
--- a/quickwit/quickwit-config/src/index_config/mod.rs
+++ b/quickwit/quickwit-config/src/index_config/mod.rs
@@ -26,6 +26,7 @@ use chrono::Utc;
 use cron::Schedule;
 use humantime::parse_duration;
 use quickwit_common::uri::Uri;
+use quickwit_common::{is_true, true_fn};
 use quickwit_doc_mapper::{DocMapper, DocMapperBuilder, DocMapping};
 use quickwit_proto::types::IndexId;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-config/src/index_config/mod.rs
```diff
@@ -170,6 +171,16 @@ pub struct IngestSettings {
...
     pub min_shards: NonZeroUsize,
+    /// Whether to validate documents against the current doc mapping during ingestion.
+    /// Defaults to true. When false, documents will be written directly to the WAL without
+    /// validation, but might still be rejected during indexing when applying the doc mapping
+    /// in the doc processor, in that case the documents are dropped and a warning is logged.
+    ///
+    /// Note that when a source has a VRL transform configured, documents are not validated against
+    /// the doc mapping during ingestion either.
+    #[schema(default = true, value_type = bool)]
+    #[serde(default = "true_fn", skip_serializing_if = "is_true")]
```
fulmicoton-dd, line 182: why `skip_serializing_if`? In most case it is better to serialize defaults no?
Here this is index config, so the data is deserialized -> serialized and end up in the metastore.

Serializing the default ensure changing default won't affect existing indexes, which is usually a good thing.

guilload: I agree with you with one caveat: after adding a new field, if you have to rollback then the cluster is borked because of the `deny_unknown_fields` that we have. Maybe we should remove them and use `serde_ignored` instead to log a warning.

PR #5985, opened 2025-11-17
add node_id to emited spans

trinity-1686a:
### Description

add the node_id to emitted spans, making traces easier to interpret

### How was this PR tested?

spin a one node cluster, make it trace to itself, and verify the new field is shown in jaeger

### DIFF
--- a/quickwit/quickwit-cli/src/logger.rs
+++ b/quickwit/quickwit-cli/src/logger.rs
@@ -56,6 +56,7 @@ pub fn setup_logging_and_tracing(
     level: Level,
     ansi_colors: bool,
     build_info: &BuildInfo,
+    node_name: String,
 ) -> anyhow::Result<EnvFilterReloadFn> {
     #[cfg(feature = "tokio-console")]
     {
@@ -112,6 +113,7 @@ pub fn setup_logging_and_tracing(
             .with_resource(Resource::new([
                 KeyValue::new("service.name", "quickwit"),
                 KeyValue::new("service.version", build_info.version.clone()),
+                KeyValue::new("service.node", node_name),
             ]))
             .build();
         let tracer = provider.tracer("quickwit");
--- a/quickwit/quickwit-cli/src/main.rs
+++ b/quickwit/quickwit-cli/src/main.rs
@@ -19,12 +19,12 @@ use std::collections::BTreeMap;
 use anyhow::Context;
 use colored::Colorize;
 use opentelemetry::global;
-use quickwit_cli::busy_detector;
 use quickwit_cli::checklist::RED_COLOR;
 use quickwit_cli::cli::{CliCommand, build_cli};
 #[cfg(feature = "jemalloc")]
 use quickwit_cli::jemalloc::start_jemalloc_metrics_loop;
 use quickwit_cli::logger::setup_logging_and_tracing;
+use quickwit_cli::{busy_detector, load_node_config};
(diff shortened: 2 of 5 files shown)


quickwit/quickwit-cli/src/main.rs
```diff
@@ -101,8 +101,28 @@ async fn main_impl() -> anyhow::Result<()> {
...
+    // purpose
+    let node_id = {
+        async || {
+            let config_uri = command.config_uri()?;
+            let config = load_node_config(config_uri).await.ok()?;
+            Some(config.node_id.take())
+        }
+    }()
+    .await
+    .unwrap_or_else(|| "unknown".to_string());
```
fulmicoton-dd, line 118: should the default value contain a random number in it?

quickwit/quickwit-cli/src/main.rs
```diff
@@ -101,8 +101,28 @@ async fn main_impl() -> anyhow::Result<()> {
...
     let build_info = BuildInfo::get();
-    let env_filter_reload_fn =
-        setup_logging_and_tracing(command.default_log_level(), ansi_colors, build_info)?;
+
+    // we need this value very early to use it in otel, but this is so early we haven't setup
+    // logging yet. What we do is read the config, and if it fails, provide a default value.
+    // Except for race conditions, an error will get logged the 2nd time the config is read,
+    // inside the command. In case of race condition, we just don't know the node id for tracing
+    // purpose
+    let node_id = {
```
fulmicoton-dd, line 110: I see. I suggest we do simpler... but less correct.
Just use the `QW_NODE_ID` env variable here.

It won't capture people using the `node_id` config value, and it will break if we change the helm chart, but it has the merit of not adding the extra complexity in this PR (loading the config twice, etc.)

fulmicoton-dd: (your call though)

quickwit/quickwit-cli/src/main.rs
trinity-1686a: i think it won't be useful because the 2 situations where this is emitted should be:
- the command is not one that behave like a node at all
- we couldn't read the config file, and assuming that stays true, the node will crash shortly

but there is little harm at doing it, so why not

quickwit/quickwit-cli/src/main.rs
trinity-1686a: i prefer reading the config: not all deployments of quickwit uses an helm chart
i'd argue that our configuration loading process is complexe, but well encapsulated, so loading the config twice doesn't harm code maintainability

quickwit/quickwit-cli/src/logger.rs
```diff
@@ -112,6 +113,7 @@ pub fn setup_logging_and_tracing(
             .with_resource(Resource::new([
                 KeyValue::new("service.name", "quickwit"),
                 KeyValue::new("service.version", build_info.version.clone()),
+                KeyValue::new("service.node", node_name),
```
guilload:
```suggestion
                KeyValue::new("service.instance.id", node_id),
```
OTEL strongly discourages custom attributes that overlap with existing semantics like `service.node`:
https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/#service-instance-id

guilload:
We can also ensure the trace is set up with `EnvResourceDetector` and rely on `OTEL_RESOURCE_ATTRIBUTES` to achieve what we want for what we're doing at the kennel.

PR #5991, opened 2025-11-21
Track indexing lag

guilload:
### Description
Track indexing lag for each split and emit metric.

### How was this PR tested?
In progress.


guilload:
@rdettai-sk, any interest in reviewing this PR?

PR #6002, opened 2025-11-28
Propagate cancellation within leaf search

rdettai-sk:
### Description

We observed that query spikes create huge leaf search tasks backlogs that don't get cancelled when the queries time out.

This is caused by the timeout cancellation that isn't propagated to spawned tasks.

This implementation is based on JoinSet, a Tokio primitive that helps managing the lifecycle of a group of tasks. It is crucial to make sure all the tasks get cancelled when the leaf request times out.

### How was this PR tested?


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1218,33 +1218,17 @@ pub async fn multi_index_leaf_search(
...
         });
-        leaf_request_tasks.push(leaf_request_future);
     }
 
-    let leaf_responses: Vec<crate::Result<LeafSearchResponse>> = tokio::time::timeout(
-        searcher_context.searcher_config.request_timeout(),
-        try_join_all(leaf_request_tasks),
-    )
-    .await??;
+    let leaf_responses: Vec<LeafSearchResponse> = try_join_vec.try_join_all().await?;
```
guilload: I know this is how things are currently, but do we actually need the responses to come back in the same order. That seems odd.

rdettai-sk: I asked myself the same question. I got to the conclusion that it's likely for reproducibility reasons: same list of splits + same query => same result. But it only holds at the leaf level. Given that the split list doesn't seem to be deterministic on the root (no order for `list_indexes_metadata()`), I don't know how much we really win from this.

guilload: Probably nothing. Let's get rid of that constraint so we can use `JoinSet` directly.

rdettai-sk: Fine by me! It simplifies the code quite a bit. I'll apply the same orderless processing when gathering join errors from individual splits.

rdettai-sk: @guilload done! I simplified the code to use plain JoinSet and rebased the PR 😃

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1202,8 +1202,7 @@ pub async fn multi_index_leaf_search(
     //
     // It is a little bit tricky how to handle which is now the incremental_merge_collector, one
     // per index, e.g. when to merge results and how to avoid lock contention.
-    let mut leaf_request_tasks = Vec::new();
-
+    let mut join_set = JoinSet::new();
```
guilload:
```suggestion
    let mut leaf_request_futures = JoinSet::new();
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1241,33 +1240,16 @@ pub async fn multi_index_leaf_search(
...
-            }
-            Err(err) => {
-                incremental_merge_collector.add_failed_split(SplitSearchError {
-                    split_id: "unknown".to_string(),
-                    error: format!("{err}"),
-                    retryable_error: true,
-                });
-            }
-        }
+    while let Some(result) = join_set.join_next().await {
```
guilload:
```suggestion
    while let Some(leaf_response_join_result) = leaf_request_futures.join_next().await {
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1241,33 +1240,16 @@ pub async fn multi_index_leaf_search(
...
+        incremental_merge_collector.add_result(result??)?;
```
guilload: This not the original behavior, right? Before we would keep going, now we return an error right away.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1394,35 +1375,37 @@ pub async fn single_doc_mapping_leaf_search(
...
-    for (split, leaf_search_join_handle) in leaf_search_single_split_join_handles {
+    while let Some(leaf_search_join_result) = join_set.join_next().await {
         // splits that did not panic were already added to the collector
-        if let Err(join_error) = leaf_search_join_handle.await {
+        if let Err(join_error) = leaf_search_join_result {
             if join_error.is_cancelled() {
                 // An explicit task cancellation is not an error.
                 continue;
             }
+            let position = split_with_task_id
```
guilload: It would be easier to have the future return the split id.

quickwit/quickwit-search/src/service.rs
```diff
@@ -195,8 +195,9 @@ impl SearchService for SearchServiceImpl {
             start: Instant::now(),
             targeted_splits: num_splits,
             status: None,
-        }
-        .await
+        };
+        let timeout = self.searcher_context.searcher_config.request_timeout();
```
guilload, line 199: If the root search is wrapped with the same timeout value and we cancel all the in flight feature when we cancel the root search, why do we need this timeout? Belt and suspenders?

quickwit/quickwit-search/src/leaf.rs
rdettai-sk: that doesn't work in case of panic (JoinError)

quickwit/quickwit-search/src/leaf.rs
rdettai-sk: We were already erroring right away for JoinError. For regular errors we were continuing, but adding a SplitSearchError with "unknown" split_id to the list of failed splits. I think the most likely reason a child request might fail is a merge error. Given that the user doesn't know how many and which splits failed, it seems quite unlikely that the result can be reasonably used. I can revert this part if you think the partial result is valuable in this scenario.

quickwit/quickwit-search/src/service.rs
rdettai-sk: I would rather not trust the RPC for cancellation. In case of connection drop, that would rely on the network level timeouts to perform the cancelation, and those are often hidden defaults that are hard to figure out. It's also close to impossible to test against regressions (even though to be honest the current cancellation also lacks tests and I didn't find a good solution yet).

I also have [this other PR](https://github.com/quickwit-oss/quickwit/pull/6010/files#diff-d2ea9881caab67f230ee487690e5e378a9ec5b96d6d58a3999f727293cc44d9f) where I enable different timeouts for different search sizes. In that PR the root timeout is disabled because the actual timeout is chosen on the leaf (open to change that, see [PR description](https://github.com/quickwit-oss/quickwit/pull/6010) for details).

quickwit/quickwit-search/src/leaf.rs
fulmicoton-dd: @rdettai-sk you might be right and this 100% deserves diving into the detail and do a refactoring...

However, we do not have time to do this at the moment, and the rest of the PR is valuable. Is it possible to revert this part?

rdettai-sk: done ✅

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1394,35 +1388,37 @@ pub async fn single_doc_mapping_leaf_search(
...
-    for (split, leaf_search_join_handle) in leaf_search_single_split_join_handles {
+    while let Some(leaf_search_join_result) = split_search_futures.join_next().await {
         // splits that did not panic were already added to the collector
-        if let Err(join_error) = leaf_search_join_handle.await {
+        if let Err(join_error) = leaf_search_join_result {
             if join_error.is_cancelled() {
                 // An explicit task cancellation is not an error.
                 continue;
             }
+            let position = split_with_task_id
```
guilload: Can we also use a map from task ID to split ID here instead of having to do the find index + remove dance.

rdettai-sk: Done! I was using a vec because it is only looked up very rarely when you have a panic.

quickwit/quickwit-search/src/leaf.rs
rdettai-sk: I just stumbled upon the [leaf search retry mechanism](https://github.com/quickwit-oss/quickwit/blob/5f9f8f7c95f5150b8dd14fe53776a832cd117215/quickwit/quickwit-search/src/retry/search.rs#L46-L48), which happens to silently swallow this `unknown` split_id. I think this is a bug.

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1241,24 +1240,20 @@ pub async fn multi_index_leaf_search(
...
+                .in_current_span()
                 .await
             }
-            .in_current_span()
         });
-        leaf_request_tasks.push(leaf_request_future);
     }
 
-    let leaf_responses: Vec<crate::Result<LeafSearchResponse>> = tokio::time::timeout(
-        searcher_context.searcher_config.request_timeout(),
```
trinity-1686a, line 1282: ~~this timeout seems to have disappeared. Before it would shorten a query (but not actually release resources 😭). Can we bring it back?~~

never mind, it was added on the caller side

quickwit/quickwit-search/src/service.rs
trinity-1686a: from my observation, cancellation doesn't seem to be propagated properly from root search to leaf nodes

PR #6003, opened 2025-12-01
Improve control plane logging

guilload:
### Description
Adding some logging statements and updating the formatting to be less noisy

### How was this PR tested?
`make test-all`

### DIFF
--- a/quickwit/quickwit-common/src/pretty.rs
+++ b/quickwit/quickwit-common/src/pretty.rs
@@ -23,14 +23,15 @@ impl<I> PrettySample<I> {
     }
 }
 
-impl<I, T> fmt::Debug for PrettySample<I>
+impl<I> fmt::Debug for PrettySample<I>
 where
-    I: IntoIterator<Item = T> + Clone,
-    T: fmt::Debug,
+    I: IntoIterator + Clone,
+    I::Item: fmt::Debug,
 {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(formatter, "[")?;
-        // in general we will get passed a reference (&[...], &HashMap...) or a Map<_> of them.
+
+        // In general, we will receive a reference (&[...], &HashMap...) or a Map<_> of them.
         // So we either perform a Copy, or a cheap Clone of a simple struct
         let mut iter = self.0.clone().into_iter().enumerate();
         for (i, item) in &mut iter {
@@ -55,9 +56,9 @@ pub trait PrettyDisplay {
     fn pretty_display(&self) -> impl fmt::Display;
 }
 
-struct PrettyDurationDisplay<'a>(&'a Duration);
+struct DurationPrettyDisplay<'a>(&'a Duration);
 
-impl fmt::Display for PrettyDurationDisplay<'_> {
+impl fmt::Display for DurationPrettyDisplay<'_> {
     fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
         // This is enough for my current use cases. To be extended as you see fit.
         let duration_millis = self.0.as_millis();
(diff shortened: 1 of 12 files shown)


quickwit/quickwit-common/src/pretty.rs
```diff
@@ -76,7 +77,37 @@ impl fmt::Display for PrettyDurationDisplay<'_> {
...
 impl PrettyDisplay for Duration {
     fn pretty_display(&self) -> impl fmt::Display {
-        PrettyDurationDisplay(self)
+        DurationPrettyDisplay(self)
+    }
+}
+
+struct SequencePrettyDisplay<I>(I);
+
+impl<I> fmt::Display for SequencePrettyDisplay<I>
```
trinity-1686a, line 86: this code is similar enough to PrettySample that i wonder if SequencePrettyDisplay shouldn't just contain a `PrettySample(_, u64::MAX)`

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -1068,40 +1079,63 @@ impl IngestController {
...
+
         if shards_to_move.is_empty() {
+            crate::metrics::CONTROL_PLANE_METRICS
+                .rebalance_shards
+                .set(0);
             return Ok(None);
         }
+        crate::metrics::CONTROL_PLANE_METRICS
+            .rebalance_shards
+            .set(shards_to_move.len() as i64);
```
trinity-1686a:
```suggestion
        crate::metrics::CONTROL_PLANE_METRICS
            .rebalance_shards
            .set(shards_to_move.len() as i64);
        if shards_to_move.is_empty() {
            return Ok(None);
        }
```

quickwit/quickwit-control-plane/src/model/mod.rs
```diff
@@ -216,7 +216,6 @@ impl ControlPlaneModel {
         };
         let fp_changed = !index_model.index_config.equals_fingerprint(&index_config);
         index_model.index_config = index_config;
-        self.update_metrics();
```
trinity-1686a, line 219: what did this do that's no longer needed?

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -351,30 +367,34 @@ impl ControlPlane {
...
-            });
-            per_index_shards_json
-                .entry(index_uid.clone())
-                .or_default()
-                .extend(shards_json);
+                });
+                per_index_and_leader_shards_json
+                    .entry(source_uid.index_uid.clone())
+                    .or_default()
+                    .entry(leader_id.clone())
```
trinity-1686a: i don't think this needs clone (that or the definition of leader_id can be inlined here, which would be more coherent with the rest of the block)

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -422,7 +442,20 @@ impl Handler<ShardPositionsUpdate> for ControlPlane {
         shard_positions_update: ShardPositionsUpdate,
         ctx: &ActorContext<Self>,
     ) -> Result<(), ActorExitStatus> {
-        debug!(shard_positions_update=?shard_positions_update, "shard positions update");
+        // For logging purposes only.
+        let pretty_positions: Vec<String> = shard_positions_update
```
trinity-1686a: this could go inside a [`enabled!`](https://docs.rs/tracing/latest/tracing/macro.enabled.html) block (more for self-documenting this is used only for logging than for the actual performance gain, which i believe would be small)

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -1054,12 +1054,12 @@ impl Ingester {
             Err(_) => {
                 return json!({
                     "status": "initializing",
-                    "shards": [],
+                    "shards": {},
```
trinity-1686a, line 1057: (not for this PR): this kind  of change can easily be wrong (as in not all codepath were updated), in the future i think we want a debug_info_internal which returns a struct, and use `serde_json::to_value` to generate the type-erased value inside debug_info()

quickwit/quickwit-common/src/pretty.rs
guilload: `PrettySample` relies on `Debug` whereas for `PrettyDisplay` I really want to be able to use something other than `Debug` or `Display`.

quickwit/quickwit-control-plane/src/model/mod.rs
guilload: `update_metrics` updates the `indexes_total` metric, which obviously does not change when updating an index config.

PR #6005, opened 2025-12-03
feat: Add case_insensitive param to some Elastic DSL queries

Darkheir:
### Description

Add support for the `case_insensitive` parameter for some elasticsearch DSL queries:
* term
* prefix
* regex
* wildcard

To support case insensitive search on regex based ASTs a mode modifier `(?i)`, is added at the beginning of the regex string.

For term queries this PR fallback to using a `RegexQuery` AST when `case_insensitive` is set to True. 
It seems that elasticsearch uses a case insensitive Lucene automaton to cover this case.

### How was this PR tested?

Added integration tests

### DIFF
--- a/quickwit/quickwit-query/src/elastic_query_dsl/term_query.rs
+++ b/quickwit/quickwit-query/src/elastic_query_dsl/term_query.rs
@@ -55,9 +55,7 @@ enum TermValue {
 }
 
 fn deserialize_term_value<'de, D>(deserializer: D) -> Result<String, D::Error>
-where
-    D: Deserializer<'de>,
-{
+where D: Deserializer<'de> {
     let term_value = TermValue::deserialize(deserializer)?;
     match term_value {
         TermValue::I64(i64) => Ok(i64.to_string()),


quickwit/quickwit-query/src/elastic_query_dsl/regex_query.rs
```diff
@@ -22,16 +22,22 @@ use crate::query_ast::{QueryAst, RegexQuery as AstRegexQuery};
...
+    #[serde(default)]
+    case_insensitive: bool,
 }
 
 pub type RegexQuery = OneFieldMap<RegexQueryParams>;
 
 impl ConvertibleToQueryAst for RegexQuery {
     fn convert_to_query_ast(self) -> anyhow::Result<QueryAst> {
+        let regex = if self.value.case_insensitive {
+            format!("(?i){}", self.value.value)
```
trinity-1686a, line 34: til that syntax works in the flavour of regex we have

quickwit/quickwit-query/src/elastic_query_dsl/term_query.rs
```diff
@@ -54,7 +55,9 @@ enum TermValue {
 }
 
 fn deserialize_term_value<'de, D>(deserializer: D) -> Result<String, D::Error>
-where D: Deserializer<'de> {
+where
+    D: Deserializer<'de>,
+{
```
trinity-1686a: the CI disagree with this change, you might not be using the same rustfmt as it does

Darkheir: It seems to be the difference between nightly and stable :)

quickwit/quickwit-query/src/elastic_query_dsl/regex_query.rs
Darkheir: I would have preferred to pass the mode outside of the regex as a flag to the regex constructor, but it seems that the underlying lib requires us to pass it as a modifier.

PR #6007, opened 2025-12-05
Improve search error logs and metrics

rdettai-sk:
### Description

Improve search metrics to account for partial success and log errors.

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-search/src/metrics_trackers.rs
+++ b/quickwit/quickwit-search/src/metrics_trackers.rs
@@ -19,56 +19,78 @@ use std::task::{Context, Poll, ready};
 use std::time::Instant;
 
 use pin_project::{pin_project, pinned_drop};
-use quickwit_proto::search::LeafSearchResponse;
+use quickwit_proto::search::{LeafSearchResponse, SearchResponse};
 
 use crate::SearchError;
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-search/src/metrics_trackers.rs
```diff
@@ -19,56 +19,78 @@ use std::task::{Context, Poll, ready};
...
+            None => "plan-cancelled",
+        };
+
+        let label_values = [status];
+        SEARCH_METRICS
+            .root_search_requests_total
+            .with_label_values(label_values)
+            .inc();
+        SEARCH_METRICS
+            .root_search_request_duration_seconds
```
fulmicoton-dd, line 56: but this is just the planning part isn't it?

fulmicoton-dd: I don't understand, you are recording this in a metric that is used for something else?

rdettai-sk: this is recorded here only if the planning failed (error or cancellation), this function returns early in case of success
```rust
            // this is a partial success, actual status will be recorded during the search step
            Some(true) => return,
```

PR #6009, opened 2025-12-08
Add config to fail if search targets too many splits

rdettai-sk:
### Description

When a search needs to process too many splits compared to the cluster capacity, we know upfront that it is pretty much impossible for the query to succeed. Instead of trusting resources for the whole duration of the configured timeout, this PR adds a node config to cancel these requests early (after the planning but before sending the jobs to the leafs).

### How was this PR tested?

Unit test

### DIFF
--- a/quickwit/quickwit-config/src/node_config/mod.rs
+++ b/quickwit/quickwit-config/src/node_config/mod.rs
@@ -268,6 +268,7 @@ pub struct SearcherConfig {
     pub split_footer_cache_capacity: ByteSize,
     pub partial_request_cache_capacity: ByteSize,
     pub max_num_concurrent_split_searches: usize,
+    pub max_splits_per_search: Option<usize>,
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -325,6 +326,7 @@ impl Default for SearcherConfig {
             split_footer_cache_capacity: ByteSize::mb(500),
             partial_request_cache_capacity: ByteSize::mb(64),
             max_num_concurrent_split_searches: 100,
+            max_splits_per_search: None,
```
fulmicoton, line 329: I'm surprised serializing tests are passing without having to update json files. Have we broken something?

fulmicoton: ah no. we don't have tests like that for node_config. That's ok then

fulmicoton (approved):

PR #6010, opened 2025-12-08
Add secondary search queue

rdettai-sk:
### Description

Currently, the search permit provider handles only one query at a time. This is good for the overall query latency when queries are relatively short. It is catastrophic in any concurrent setup where some queries are slow, because all the other queries will be starved from resources until the large query completes.

This PR creates a secondary search permit provider that is independently configured. "Fast" leaf queries are routed to one provider and "slow" ones to the other. Query duration is naively inferred from the number of targetted splits, and the threshold for being routed to one provider or the other can be configured using `SearcherConfig.secondary_targeted_split_count_threshold`.

One limitation of the current implementation of this PR is that for a given root search, various leaf searches might fall on different sides of the threshold. We could fix this by applying the threashold on the root and add the targeted provider in the leaf request.

### How was this PR tested?

We ran this setup on our highly concurrent production environment. It helped stabilize query performance drastically.

### DIFF
--- a/quickwit/quickwit-search/src/leaf.rs
+++ b/quickwit/quickwit-search/src/leaf.rs
@@ -45,7 +45,7 @@ use tokio::task::JoinError;
 use tracing::*;
 
 use crate::collector::{IncrementalCollector, make_collector_for_split, make_merge_collector};
-use crate::metrics::SplitSearchOutcomeCounters;
+use crate::metrics::{SplitSearchOutcomeCounters, queue_label};
 use crate::root::is_metadata_count_request_with_ast;
 use crate::search_permit_provider::{SearchPermit, compute_initial_memory_allocation};
 use crate::service::{SearcherContext, deserialize_doc_mapper};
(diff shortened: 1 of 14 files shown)


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1366,10 +1380,12 @@ pub async fn single_doc_mapping_leaf_search(
                 .warmup_single_split_initial_allocation,
         )
     });
-    let permit_futures = searcher_context
-        .search_permit_provider
-        .get_permits(permit_sizes)
-        .await;
+    let permit_provider = if is_broad_search {
+        &searcher_context.secondary_search_permit_provider
```
fulmicoton, line 1384: with this logic, short request cannot use the broad search request permits. I don't think this is good.

(do you want a supermarket where people with less than 10 articles cannot go to the regular cashier)

rdettai-sk: Makes sense. We just don't want to queue them there if there is some broad search going on.

Before making this change, what do you think about the fact that the permit provider is chosen on the leafs? I would prefer that for a large search, all leaf searches were tagged as "broad" in the root. Currently, the timeout cannot be enforced on the root because we don't know yet what timeout will be applied to its leaf searches. This requires a change in the RPC so I wanted to be sure the the change will be merged upstream before doing it.

fulmicoton-dd:
@rdettai-sk I think we should not merge this PR. At least at the moment.
I understand that it is useful to you.

The trouble is that it is a hacky solution (the problem I pointed out, but also others: Do we really want to not use the entire permit pool for a long request if there are no request running? Would a priority system be more suited?), this adds complexity (it has many new property in the configuration for instance) to the project. After the PR is merged, moving away for a better solution would require coordination.

I understand this is a rather "unfair": we would happily merge a hack like this if it was scratching our customer's itch.

PR #6018, opened 2025-12-11
Improve rebalance shards algorithm

guilload:


quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -1008,62 +1009,93 @@ impl IngestController {
...
+            } else if per_leader_open_shards[last_idx].len() > max_num_open_shards_per_leader {
+                let shard = per_leader_open_shards[last_idx]
+                    .pop()
+                    .expect("shards should not be empty");
+                per_leader_open_shards[0].push(shard.clone());
+                shards_to_rebalance.push(shard);
+            } else {
+                break;
             }
+            per_leader_open_shards.sort_unstable_by_key(Vec::len);
```
fulmicoton-dd: I think the body of this loop only needs the leader with the largest/smallest number of shards.

sorting seems overkill and obfuscated that.
I think itertools has thing to get a min/max. We could use that over iter_mut() and get the mut ref over the two. (I cannot tell if there will be fighting involved with the borrow checker or not).

Another thing is, I would put that at the top of loop and remove the previous sort. (This is a pre-condition.)

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -1008,62 +1009,93 @@ impl IngestController {
...
                 num_available_ingesters = num_ingesters,
-                rebalance_threshold = num_open_shards_per_leader_threshold,
+                min_shards_threshold = min_num_open_shards_per_leader,
+                max_shards_threshold = max_num_open_shards_per_leader,
                 num_shards_to_rebalance,
                 "rebalancing shards"
             );
         }
-        shards_to_move
+        shards_to_rebalance
```
fulmicoton-dd: This function returns the list of shards to move... It does not move stuff itself.
This is nice because it means any sophisticated logic downstream will be applied to place the shard on the right server.

HOWEVER
The code downstream does not enforce your new condition.

This means that as a set : this function + the function that attributes your stuff does not necessarily ever converges. I'm thinking in particular of the code that tries to place shards from the same index on the same indexer.

quickwit/quickwit-search/src/tests.rs
```diff
@@ -325,7 +308,7 @@ async fn test_single_node_several_splits() -> anyhow::Result<()> {
             .as_ref()
             .map(|partial_hit| (partial_hit.split_id.as_str(), partial_hit.doc_id as i32))
     });
-    assert!(is_reverse_sorted(hit_keys));
+    assert!(hit_keys.is_sorted_by(|left, right| left.cmp(right) == Ordering::Greater));
```
fulmicoton-dd, line 311: Alternatively, I think the following works?

```suggestion
    assert!(hit_keys.is_sorted_by_key(std::cmp::Reverse);
```

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
guilload: TIL about the minmax stuff in `itertools`. I'll see what I can do. Initially, I had the following as the prelude to the loop and, understandably, the borrow checker was unhappy:
```rust
min_leader = .get_mut(0);
max_leader = .get_mut(-1);

if ....
```

quickwit/quickwit-search/src/tests.rs
guilload: TIL

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
guilload: This logic only exists in the indexing controller. In the ingest controller, the placement logic is much simpler: we allocate shards to the ingesters with the least number of open shards. Right?

fulmicoton: Got it!

PR #6023, opened 2025-12-16
Jaegerv2

nico34638:
### Description

Jaeger has a new specification for the GRPC remote storage API. They have removed the old version from release 2.6 onwards

https://github.com/jaegertracing/jaeger/tree/v2.13.0/internal/storage/v2/grpc

### How was this PR tested?

I add unit test for grpc endpoint and new helper function. And I tested the features with the latest version of Jaeger, and everything works fine.


quickwit/quickwit-jaeger/src/lib.rs
```diff
@@ -562,6 +566,441 @@ impl SpanReaderPlugin for JaegerService {
     }
 }
 
+// Implementation for Jaeger v2 storage API
+#[async_trait]
```
fulmicoton: Can we isolate this code in a v2 module file?

(and maybe do the same for v1)?

nico34638: I split between v1 and v2 and tried to factorize as much as possible in lib.rs the functions common to both

quickwit/quickwit-jaeger/src/v2.rs
```diff
@@ -0,0 +1,644 @@
...
+                        quickwit_proto::jaeger::storage::v2::any_value::Value::DoubleValue(d) => {
+                            d.to_string()
+                        }
+                        quickwit_proto::jaeger::storage::v2::any_value::Value::BoolValue(b) => {
+                            b.to_string()
+                        }
+                        _ => return None,
+                    };
+                    Some((kv.key, string_value))
+                })
```
fulmicoton: (this needs fixing indentation)
```suggestion
            let value = kv.value?.value?;
            let string_value = match value {
                        quickwit_proto::jaeger::storage::v2::any_value::Value::StringValue(s) => s,
                        quickwit_proto::jaeger::storage::v2::any_value::Value::IntValue(i) => {
                            i.to_string()
                        }
                        quickwit_proto::jaeger::storage::v2::any_value::Value::DoubleValue(d) => {
                            d.to_string()
                        }
                        quickwit_proto::jaeger::storage::v2::any_value::Value::BoolValue(b) => {
                            b.to_string()
                        }
                        _ => return None,
                    };
                    Some((kv.key, string_value))
           })
```

quickwit/quickwit-jaeger/src/v2.rs
```diff
@@ -0,0 +1,644 @@
...
+        let term_query = TermQuery {
+            field: "trace_id".to_string(),
+            value,
+        };
+        query.should.push(term_query.into());
+    }
+
+    if root_only {
+        let is_root = UserInputQuery {
+            user_text: "NOT is_root:false".to_string(),
```
fulmicoton, line 366: why not `is_root:true` ? Why not a term query? Is this in v1?

quickwit/quickwit-jaeger/src/v2.rs
```diff
@@ -0,0 +1,644 @@
...
+            default_fields: None,
+            default_operator: BooleanOperand::And,
+            lenient: true,
+        };
+        let mut new_query = BoolQuery::default();
```
fulmicoton, line 371: nitpick: BooleanQuery::intersection

nico34638: I can't find the BoolQuery::intersection method, are you sure it exists?

quickwit/quickwit-jaeger/src/v2.rs
nico34638: https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-jaeger/src/lib.rs#L322 same in v1 implementation 

I don't think a term query is appropriate because, if I understand correctly, a term query searches for fields where is_root is false, or in cases where there are child spans, the field does not exist

quickwit/quickwit-jaeger/src/v2.rs
fulmicoton: my bad. that's probably only in tantivy

fulmicoton (approved):

fulmicoton (approved):

PR #6024, opened 2025-12-16
add predicate cache

trinity-1686a:
### Description

Add a predicate cache which can store the list of documents matching a (sub-)query.
This can be useful when multiple requests share the same fragment of query or the same query with other part of the request being distinct (search after, successive aggregation on the same set of document...)

at the moment we only insert these nodes to speed up search_after, ideally we would let people suggest in their query what might be replayed a bunch (the top level of a query that's going to use pagination, or a part of a query added systematically for access control by some middleware), or detect automatically good candidates for being cached.

### DIFF
--- a/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
+++ b/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
@@ -23,6 +23,7 @@ use std::time::Duration;
 use fnv::FnvHashSet;
 use futures::StreamExt;
 use futures::stream::FuturesUnordered;
+use itertools::{Itertools, MinMaxResult};
 use quickwit_actors::Mailbox;
 use quickwit_common::Progress;
 use quickwit_common::pretty::PrettySample;
@@ -81,7 +82,7 @@ const FIRE_AND_FORGET_TIMEOUT: Duration = Duration::from_secs(3);
 /// All errors are ignored, and not even logged.
 fn fire_and_forget(
     fut: impl Future<Output = ()> + Send + 'static,
-    operation: impl std::fmt::Display + Send + Sync + 'static,
+    operation: impl std::fmt::Display + Send + 'static,
 ) {
     tokio::spawn(async move {
         if let Err(_timeout_elapsed) = tokio::time::timeout(FIRE_AND_FORGET_TIMEOUT, fut).await {
@@ -997,78 +998,9 @@ impl IngestController {
         }
     }
 
-    /// This method just "computes"" the number of shards to move for rebalance.
-    /// It does not run any side effect except logging.
-    ///
-    /// TODO we consider the number of alive ingesters for this computation,
-    /// but deal with entire number of shards here.
-    /// This could cause problems when dealing with a lot of unavailable ingesters.
-    ///
-    /// On the other hand it biases thing the "right way":
-    /// If we are missing some ingesters, their shards should still be in the model, but they should
-    /// be missing from the ingester pool.
-    ///
-    /// As a result `num_open_shards_per_leader_threshold` should be inflated.
-    ///
-    /// TODO this implementation does not consider replica.
-    fn rebalance_compute_shards_to_move(&self, model: &ControlPlaneModel) -> Vec<Shard> {
-        let num_ingesters = self.ingester_pool.len();
-        let mut num_open_shards: usize = 0;
-
-        if num_ingesters == 0 {
-            debug!("no ingester available");
-            return Vec::new();
-        }
-        let mut per_leader_open_shards: HashMap<&str, Vec<&ShardEntry>> = HashMap::new();
-
-        for shard in model.all_shards() {
-            if shard.is_open() {
-                num_open_shards += 1;
-                per_leader_open_shards
-                    .entry(&shard.leader_id)
-                    .or_default()
-                    .push(shard);
-            }
-        }
-        // We tolerate an ingester with 10% more shards than the average.
-        // Let's first identify the list of shards we want to "move".
-        let num_open_shards_per_leader_threshold =
-            (num_open_shards * 11).div_ceil(10 * num_ingesters);
-
-        let mut shards_to_move: Vec<Shard> = Vec::new();
-
-        let mut rng = thread_rng();
-        for open_shards in per_leader_open_shards.values() {
-            if let Some(num_shards_to_move) = open_shards
-                .len()
-                .checked_sub(num_open_shards_per_leader_threshold)
-            {
-                shards_to_move.extend(
-                    open_shards[..]
-                        .choose_multiple(&mut rng, num_shards_to_move)
-                        .map(|shard_entry| shard_entry.shard.clone()),
-                );
-            }
-        }
-        let num_shards_to_rebalance = shards_to_move.len();
-
-        if num_shards_to_rebalance == 0 {
-            info!("no shards to rebalance");
-        } else {
-            info!(
-                num_open_shards,
-                num_available_ingesters = num_ingesters,
-                rebalance_threshold = num_open_shards_per_leader_threshold,
-                num_shards_to_rebalance,
-                "rebalancing shards"
-            );
-        }
-        shards_to_move
-    }
-
-    /// Moves shards from ingesters with too many shards to ingesters with too few shards. Moving a
-    /// shard consists of closing the shard on the source ingester and opening a new one on the
-    /// target ingester.
+    /// Rebalances shards from ingesters with too many shards to ingesters with too few shards.
+    /// Moving a shard consists of closing the shard on the source ingester and opening a new
+    /// one on the target ingester.
     ///
     /// This method is guarded by a lock to ensure that only one rebalance operation is performed at
     /// a time.
@@ -1084,18 +1016,18 @@ impl IngestController {
         };
         self.stats.num_rebalance_shards_ops += 1;
 
-        let shards_to_move: Vec<Shard> = self.rebalance_compute_shards_to_move(model);
+        let shards_to_rebalance: Vec<Shard> = self.compute_shards_to_rebalance(model);
 
         crate::metrics::CONTROL_PLANE_METRICS
             .rebalance_shards
-            .set(shards_to_move.len() as i64);
+            .set(shards_to_rebalance.len() as i64);
 
-        if shards_to_move.is_empty() {
+        if shards_to_rebalance.is_empty() {
             return Ok(None);
         }
         let mut per_source_num_shards_to_open: HashMap<SourceUid, usize> = HashMap::new();
 
-        for shard in &shards_to_move {
+        for shard in &shards_to_rebalance {
             *per_source_num_shards_to_open
                 .entry(shard.source_uid())
                 .or_default() += 1;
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -154,17 +155,24 @@ impl<'a, 'f> QueryAstVisitor<'a> for ExistsQueryFastFields<'f> {
...
 /// Build a `Query` with field resolution & forbidding range clauses.
 pub(crate) fn build_query(
-    query_ast: &QueryAst,
-    schema: Schema,
-    tokenizer_manager: &TokenizerManager,
-    search_fields: &[String],
-    with_validation: bool,
+    query_ast: QueryAst,
+    context: &BuildTantivyAstContext,
+    cache_context: Option<(Arc<dyn quickwit_query::query_ast::PredicateCache>, String)>,
```
fulmicoton-dd: Not fond of the `(PredicateCache, String)`. One needs to travel a lot of get String is a split id.

quickwit/quickwit-query/src/query_ast/cache_node.rs
```diff
@@ -0,0 +1,861 @@
...
+        hit_set.boost = boost;
+        Ok(Box::new(hit_set))
+    }
+
+    fn explain(&self, reader: &SegmentReader, doc: DocId) -> tantivy::Result<Explanation> {
+        self.inner_weight.explain(reader, doc)
+    }
+}
+
+pub struct CachePreIgniter {
```
fulmicoton-dd: This needs a comment saying what this is about, and the name should say "PredicateCache", not cache.

quickwit/quickwit-doc-mapper/src/query_builder.rs
fulmicoton-dd: Should this pair be the `CachePreIgniter` object directly? (with a better name)

quickwit/quickwit-query/src/query_ast/cache_node.rs
```diff
@@ -0,0 +1,861 @@
...
+        } else {
+            self.block_pos += 1;
+        }
+        self.doc()
+    }
+
+    // fn seek(&mut self, target: DocId) -> DocId {
+    // }
+
+    fn doc(&self) -> DocId {
```
fulmicoton-dd, line 293:
```suggestion
    #[inline(always)]
    fn doc(&self) -> DocId {
```

quickwit/quickwit-query/src/query_ast/cache_node.rs
```diff
@@ -0,0 +1,861 @@
...
+    pub fn into_buffer(self) -> OwnedBytes {
+        self.buffer
+    }
+}
+
+impl DocSet for HitSet {
+    fn advance(&mut self) -> DocId {
+        if self.block_pos == BitPacker1x::BLOCK_LEN - 1 {
+            self.block_pos = 0;
+            let Some(num_bits) = self.buffer.get(self.buffer_pos) else {
```
fulmicoton-dd: I think it would be a tad nicer if you moved that code in a method called.

`fn load_new_block(&mut self) -> bool`

quickwit/quickwit-query/src/query_ast/cache_node.rs
```diff
@@ -0,0 +1,861 @@
...
+                self.finished = true;
+                return self.doc();
+            };
+            self.buffer_pos += 1;
+            if *num_bits == 0x80 {
+                // final block, decode as many ids as possible
+                let mut i = 0;
+                for chunk in self.buffer[self.buffer_pos..].as_chunks().0 {
+                    self.current_block[i] = u32::from_le_bytes(*chunk);
```
fulmicoton-dd: `from_ne_bytes` ?
it does not matter much but it could be nice to imply this serialization is just here for local/in ram usage.

quickwit/quickwit-query/src/query_ast/cache_node.rs
```diff
@@ -0,0 +1,861 @@
...
+    #[cfg(test)]
+    fn empty() -> Self {
+        Self::from_buffer(OwnedBytes::new(vec![0, 0, 0, 0]))
+    }
+
+    /// Build a HitSet from its serialized form.
+    ///
+    /// The provided buffer must come from `HitSet::into_buffer`
+    pub fn from_buffer(buffer: OwnedBytes) -> Self {
+        let mut this = Self {
```
fulmicoton-dd, line 224: Since you went for a posting list format (as opposed to a bit set for instance), it might be nice to reuse tantivy's format and make this a `TermScorer`. It would add the skip list and might enable some dynamic performance optimization.

(can be done later and requires code in tantivy)

quickwit/quickwit-query/src/query_ast/cache_node.rs
```diff
@@ -0,0 +1,861 @@
...
+    fn end_of_block(&self) -> bool {
+        self.in_block_pos() == (BitPacker1x::BLOCK_LEN - 1)
+    }
+
+    fn flush_block(&mut self) {
+        let num_bits =
+            BitPacker1x.num_bits_strictly_sorted(self.previous_last_val, &self.current_block);
+        self.buffer.push(num_bits);
+        let current_buffer_pos = self.buffer.len();
+        let new_end = current_buffer_pos + BitPacker1x::BLOCK_LEN / 8 * num_bits as usize;
```
fulmicoton-dd:
```suggestion
        let new_end = current_buffer_pos + (BitPacker1x::BLOCK_LEN * num_bits as usize) / 8 ;
```

quickwit/quickwit-query/src/query_ast/cache_node.rs
```diff
@@ -0,0 +1,861 @@
...
+        self.current_block[self.in_block_pos()] = value;
+        if self.end_of_block() {
+            self.flush_block();
+        }
+        self.count += 1;
+    }
+
+    pub fn build(mut self) -> HitSet {
+        if self.in_block_pos() != 0 {
+            self.buffer.push(0x80);
```
fulmicoton-dd: Can we make that a const with a nice name?

quickwit/quickwit-query/src/query_ast/mod.rs
```diff
@@ -115,6 +118,17 @@ impl QueryAst {
                     boost,
                 })
             }
+            QueryAst::Cache(cache_node) => {
+                let inner = cache_node.inner.parse_user_query(default_search_fields)?;
```
fulmicoton-dd, line 122: I wonder if we should return an error here or log a warning.

The thing I am scared of, is in the future, for whatever reasons, use running `parse_user_query` downstream, and our cache node to silently switched to uninitialized again.

(The dev or the reviewer adding this change would never be able to predict such a side effect)

quickwit/quickwit-query/src/query_ast/mod.rs
```diff
@@ -144,30 +158,53 @@ impl QueryAst {
...
+pub struct BuildTantivyAstContext<'a> {
+    pub schema: &'a TantivySchema,
+    pub tokenizer_manager: &'a TokenizerManager,
+    pub search_fields: &'a [String],
+    pub with_validation: bool,
+}
+
+/// Shorthand to build a context from a schema in tests.
+#[macro_export]
+macro_rules! test_context {
```
fulmicoton-dd: The macro is here just to get a variadic? Can we replace that by two functions?

quickwit/quickwit-query/src/query_ast/visitor.rs
```diff
@@ -236,4 +252,22 @@ pub trait QueryAstTransformer {
     fn transform_regex(&mut self, regex_query: RegexQuery) -> Result<Option<QueryAst>, Self::Err> {
         Ok(Some(QueryAst::Regex(regex_query)))
     }
+
+    fn transform_cache_node(
+        &mut self,
+        cache_node: CacheNode,
+    ) -> Result<Option<QueryAst>, Self::Err> {
+        if !matches!(cache_node.state, CacheState::CacheHit(_)) {
```
fulmicoton-dd: why the negation? can we swap the if and the else statement?
And maybe use an early-return?

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -608,6 +616,18 @@ fn rewrite_request(
         remove_redundant_timestamp_range(search_request, split, timestamp_field);
     }
     rewrite_aggregation(search_request);
+    if search_request.search_after.is_some() {
```
fulmicoton-dd, line 621: Can we add a comment explaining the logic here?
If I understand correctly, this will only kick in after the first scroll?

quickwit/quickwit-query/src/query_ast/cache_node.rs
trinity-1686a: i started with a naive bitset, but moved to a posting list for compactness. There didn't seem to be enough public code in tantivy to use its posting list format from the outside, but i agree with the idea

quickwit/quickwit-query/src/query_ast/mod.rs
trinity-1686a: it's technically not a valid function lifetime-wise (it would return a reference to the stack), but we can afford writing `&test_contex(..)` when calling it

quickwit/quickwit-search/src/leaf.rs
trinity-1686a: we start caching only after the 1st query with a search after (2nd query of a series), so the cache only starts helping at the 3rd

i'm not sure what rules to put in place exactly, maybe just caching the entire query every-time would be better. It would cause more cache eviction, but would start helping sooner too, and related queries are likely to be very close in time 🤔

quickwit/quickwit-query/src/query_ast/cache_node.rs
trinity-1686a: this can't be harmful, but i do wonder: is it really useful here? I would expect inlining to happen either way

quickwit/quickwit-doc-mapper/src/query_builder.rs
trinity-1686a: at the moment i went with `quickwit_proto::types::SplitId` (which is a type alias for String)
passing the transformer that fills the cache could make sens ig

quickwit/quickwit-query/src/query_ast/mod.rs
trinity-1686a: what do you think of a debug_assert, and possibly a warn? Wouldn't break in prod, but assuming good test coverage/manual testing with assertions, this should fail loudly enough to make someone notice

quickwit/quickwit-search/src/leaf.rs
fulmicoton-dd: I think that's ok.. But I am wondering if we should add a small short lived cache for posting lists.

quickwit/quickwit-query/src/query_ast/mod.rs
fulmicoton-dd: ok

quickwit/quickwit-query/src/query_ast/cache_node.rs
fulmicoton-dd: It is probably not useful, but I don't like bad surprises.

PR #6028, opened 2025-12-18
bump cargo compatible and some incompatible dependencies

congx4:
### Description

bump cargo compatible and some incompatible dependencies

### How was this PR tested?

```

make test-all 
```

### DIFF
--- a/quickwit/quickwit-config/src/node_config/mod.rs
+++ b/quickwit/quickwit-config/src/node_config/mod.rs
@@ -269,6 +269,7 @@ pub struct SearcherConfig {
     pub partial_request_cache_capacity: ByteSize,
     pub predicate_cache_capacity: ByteSize,
     pub max_num_concurrent_split_searches: usize,
+    pub max_splits_per_search: Option<usize>,
     // Deprecated: stream search requests are no longer supported.
     #[serde(alias = "max_num_concurrent_split_streams", default, skip_serializing)]
     pub _max_num_concurrent_split_streams: Option<serde::de::IgnoredAny>,
@@ -327,6 +328,7 @@ impl Default for SearcherConfig {
             partial_request_cache_capacity: ByteSize::mb(64),
             predicate_cache_capacity: ByteSize::mb(256),
             max_num_concurrent_split_searches: 100,
+            max_splits_per_search: None,
             _max_num_concurrent_split_streams: None,
             aggregation_memory_limit: ByteSize::mb(500),
             aggregation_bucket_limit: 65000,
--- a/quickwit/quickwit-config/src/node_config/serialize.rs
+++ b/quickwit/quickwit-config/src/node_config/serialize.rs
@@ -663,6 +663,7 @@ mod tests {
                 partial_request_cache_capacity: ByteSize::mb(64),
                 predicate_cache_capacity: ByteSize::mb(256),
                 max_num_concurrent_split_searches: 150,
+                max_splits_per_search: None,
                 _max_num_concurrent_split_streams: Some(serde::de::IgnoredAny),
                 split_cache: None,
                 request_timeout_secs: NonZeroU64::new(30).unwrap(),
--- a/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
+++ b/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
@@ -925,7 +925,10 @@ impl IngestController {
         request: AdviseResetShardsRequest,
         model: &ControlPlaneModel,
     ) -> AdviseResetShardsResponse {
-        info!("advise reset shards");
+        info!(
+            "received advise reset shards request from `{}`",
+            request.ingester_id
+        );
         debug!(shard_ids=?summarize_shard_ids(&request.shard_ids), "advise reset shards");
 
         let mut shards_to_delete: Vec<ShardIds> = Vec::new();
@@ -1011,7 +1014,7 @@ impl IngestController {
         progress: &Progress,
     ) -> MetastoreResult<Option<JoinHandle<()>>> {
         let Ok(rebalance_guard) = self.rebalance_lock.clone().try_lock_owned() else {
-            info!("skipping rebalance: another rebalance is already in progress");
+            debug!("skipping rebalance: another rebalance is already in progress");
             return Ok(None);
         };
         self.stats.num_rebalance_shards_ops += 1;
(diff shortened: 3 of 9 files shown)


quickwit/Cargo.lock
```diff
@@ -4247,7 +4273,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
```
fulmicoton-dd, line 4276: what happened here?

quickwit/Cargo.lock
```diff
@@ -4525,7 +4575,7 @@ dependencies = [
  "byteorder",
  "csv",
  "encoding",
- "env_logger",
+ "env_logger 0.10.2",
```
fulmicoton-dd, line 4578: Can you check why we end up shipping 2 versions of env_logger?
Which dep is causing that?

quickwit/Cargo.lock
```diff
@@ -5358,6 +5395,25 @@ dependencies = [
  "url",
 ]
 
+[[package]]
+name = "objc2-core-foundation"
```
fulmicoton-dd, line 5390: Can you check which deps require this and whether this is really needed?

quickwit/Cargo.lock
```diff
@@ -7866,6 +7951,15 @@ dependencies = [
  "bitflags 2.10.0",
 ]
 
+[[package]]
+name = "redox_syscall"
```
fulmicoton-dd, line 7946: Do we really need this? What is pulling redox_syscall?

quickwit/Cargo.lock
```diff
@@ -8302,7 +8396,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.11.0",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
```
fulmicoton-dd, line 8390: what happened here?

congx4: we total have 5 different version of windows-sys, rustix 1.1.2 has a flexible requirement like windows-sys = ">=0.52", so Cargo is free to pick any version ≥0.52. It picks whichever version minimizes the total number of duplicate crates in your dependency graph.

https://docs.rs/crate/rustix/latest

quickwit/Cargo.lock
congx4: yes, we need this: 
redox_syscall v0.6.0
└── libredox v0.1.11
    └── filetime v0.2.26
        └── tar v0.4.44

quickwit/Cargo.lock
congx4: we total have 5 different version of windows-sys, is-terminal 0.4.17 has a flexible requirement like windows-sys = ">=0.52", so Cargo is free to pick any version ≥0.52. It picks whichever version minimizes the total number of duplicate crates in your dependency graph.

https://docs.rs/crate/is-terminal/latest

quickwit/Cargo.lock
congx4: objc2-core-foundation v0.3.2
├── objc2-io-kit v0.3.2
│   └── sysinfo v0.37.2

quickwit/Cargo.lock
congx4: We use 2 different versions: 
env_logger v0.10.2
├── lindera-cc-cedict-builder v0.27.2
│   └── lindera-dictionary v0.27.2
│       ├── lindera-tokenizer v0.27.2
│       │   └── quickwit-query v0.8.0 

env_logger v0.11.8
└── quickwit-common

quickwit/Cargo.lock
fulmicoton-dd: but this means another deps has a strict requirement isn't it?

quickwit/Cargo.lock
fulmicoton-dd: If we update to lindera 1 I think the deps is removed.

quickwit/Cargo.lock
fulmicoton-dd: I think this will go away if remove default features, and only add the disk feature of sysinfo. Can you try this?

quickwit/Cargo.lock
fulmicoton-dd: ok this one does not matter. this is only enabled when compiling on redox.

quickwit/Cargo.lock
congx4: ok, updating lindera is going to be in the following pr.

quickwit/Cargo.lock
congx4: we don't have features for sysinfo = "0.37", is this what you meant?

quickwit/Cargo.lock
congx4: I’m unsure if this is related to the dependencies I updated and then reverted. Those updates required additional code fixes and also relied on windows-sys, which may have triggered Cargo to recalculate the version resolution.

I don't have any idea why it becomes 0.52.0. I will finish updating the rest deps to see if it can go back to 0.61.2. But let me know if you have any other thoughts.

congx4: It is updated back to 0.61.2 after the second pr. 
https://github.com/quickwit-oss/quickwit/pull/6029#discussion_r2632369124

quickwit/Cargo.lock
fulmicoton-dd: This is sysinfo Cargo.toml
https://github.com/GuillaumeGomez/sysinfo/blob/main/Cargo.toml

By default it enables the following features
```default = ["component", "disk", "network", "system", "user"]```

I think we only use disk.

If you disable default features and explicitly opt in for component, I think (?) these deps will go away

Comment (congx4) [reply quickwit/Cargo.lock:5390]: After just using disk feature from sysinfo, other packages were removed, but not this one. 
https://github.com/quickwit-oss/quickwit/pull/6028/changes/2b74fd8007a30a2741211e99335e5d7189e2e262

File: quickwit/Cargo.toml [hunk 6]
```diff
@@ -230,20 +230,20 @@ sqlx = { version = "0.8", features = [
 ] }
 syn = { version = "2.0", features = ["extra-traits", "full", "parsing"] }
 sync_wrapper = "1"
-sysinfo = "0.33"
+sysinfo = { version = "0.37", default-features = false, features = ["disk"] }
```
fulmicoton-dd, line 233: awesome thank you!

PR #6029, opened 2025-12-18
bump more incompatible deps

congx4:
### Description

 1. Removed deprecated GoogleToken/GoogleTokenLoad API: These types no longer exist in reqsign 0.18.x, so I removed the unused DummyTokenLoader struct and its implementation.
 2. Fixed opendal Scheme comparison: Changed storage_info.scheme() == &opendal::Scheme::Gcs to storage_info.scheme() == "gcs" since the OperatorInfo::scheme() now returns &str. 
 3. update the test to align with the latest apis from regsign. 
 

### How was this PR tested?

```

make test-all 
```

### DIFF
--- a/quickwit/quickwit-config/src/node_config/mod.rs
+++ b/quickwit/quickwit-config/src/node_config/mod.rs
@@ -269,6 +269,7 @@ pub struct SearcherConfig {
     pub partial_request_cache_capacity: ByteSize,
     pub predicate_cache_capacity: ByteSize,
     pub max_num_concurrent_split_searches: usize,
+    pub max_splits_per_search: Option<usize>,
     // Deprecated: stream search requests are no longer supported.
     #[serde(alias = "max_num_concurrent_split_streams", default, skip_serializing)]
     pub _max_num_concurrent_split_streams: Option<serde::de::IgnoredAny>,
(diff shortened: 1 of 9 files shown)


quickwit/Cargo.lock
```diff
@@ -8396,7 +8540,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.11.0",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
```
congx4, line 8534: It is back to 0.61.2 after the first pr. This should resolve the question here. 

https://github.com/quickwit-oss/quickwit/pull/6028#discussion_r2631613156

quickwit/quickwit-storage/tests/google_cloud_storage.rs
```diff
@@ -25,20 +25,34 @@ mod gcp_storage_test_suite {
...
-            .await?
-            .ok_or_else(|| anyhow::anyhow!("Failed to obtain authentication token"))?;
-
-        let signer = reqsign::GoogleSigner::new("storage");
-        signer.sign(req, &token)?;
-
+        let signer = reqsign::google::default_signer("storage");
+        
+        // Create http::Request and extract parts for signing
+        let http_req = http::Request::builder()
```
congx4, line 35: This test is updated by Claude, I just ensure it keeps the same behavior as before.

PR #6038, opened 2025-12-22
update otl and rand family deps

congx4:
### Description

cargo upgrade --incompatible \
  -p opentelemetry \
  -p opentelemetry_sdk \
  -p opentelemetry-otlp \
  -p tracing-opentelemetry \
  && cargo check --all-targets
  
cargo upgrade --incompatible \
  -p rand \
  -p rand_distr \
  && cargo check --all-targets && cargo test

### How was this PR tested?

make test-all

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -65,38 +65,37 @@ jobs:
           --health-retries 5
 
       kafka-broker:
-        image: confluentinc/cp-kafka:7.2.1
+        image: confluentinc/confluent-local:7.4.11
         ports:
           - "9092:9092"
           - "9101:9101"
         env:
-          KAFKA_BROKER_ID: 1
-          KAFKA_ZOOKEEPER_CONNECT: "zookeeper:2181"
-          KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,PLAINTEXT_HOST:PLAINTEXT
-          KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka-broker:29092,PLAINTEXT_HOST://localhost:9092
+          # Mode KRaft (Single Node)
+          KAFKA_NODE_ID: 1
+          KAFKA_PROCESS_ROLES: 'broker,controller'
+          KAFKA_CONTROLLER_QUORUM_VOTERS: '1@localhost:9093'
+          KAFKA_LOG4J_LOGGERS: "org.apache.kafka.image.loader.MetadataLoader=WARN"
+
+          # Listeners
+          KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: 'CONTROLLER:PLAINTEXT,EXTERNAL:PLAINTEXT'
+          KAFKA_LISTENERS: 'EXTERNAL://0.0.0.0:9092,CONTROLLER://0.0.0.0:9093'
+          KAFKA_ADVERTISED_LISTENERS: 'EXTERNAL://localhost:9092'
+          KAFKA_CONTROLLER_LISTENER_NAMES: 'CONTROLLER'
+          KAFKA_INTER_BROKER_LISTENER_NAME: 'EXTERNAL'
+
+          # Configuration simplifiée
           KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
           KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS: 0
-          KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-cli/src/logger.rs
```diff
@@ -94,25 +94,20 @@ pub fn setup_logging_and_tracing(
...
+                    // Quickwit can generate a lot of spans, especially in debug mode, and the
+                    // default queue size of 2048 is too small.
+                    .with_max_queue_size(32_768)
+                    .build(),
+            )
+            .build();
+        let provider = opentelemetry_sdk::trace::SdkTracerProvider::builder()
             .with_span_processor(batch_processor)
-            .with_resource(Resource::new([
-                KeyValue::new("service.name", "quickwit"),
```
guilload, line 113: There should be a new API for setting resource attributes.

quickwit/quickwit-cli/src/main.rs
```diff
@@ -114,7 +113,6 @@ async fn main_impl() -> anyhow::Result<()> {
         0
     };
 
-    global::shutdown_tracer_provider();
```
guilload, line 117: We need to call `.shutdown` on the processor in order to flush and send the in-flight spans. You will have to refactor `setup_logging_and_tracing` a little bit to do so.

quickwit/quickwit-proto/src/lib.rs
```diff
@@ -196,7 +196,7 @@ impl Extractor for MetadataMap<'_> {
 pub fn set_parent_span_from_request_metadata(request_metadata: &tonic::metadata::MetadataMap) {
     let parent_cx =
         global::get_text_map_propagator(|prop| prop.extract(&MetadataMap(request_metadata)));
-    Span::current().set_parent(parent_cx);
+    let _ = Span::current().set_parent(parent_cx);
```
guilload, line 199: When can this fail now?

quickwit/quickwit-cli/src/logger.rs
congx4: Yes, it still has, added it back.

quickwit/quickwit-cli/src/main.rs
congx4: Returned the trace_provider from `setup_logging_and_tracing`, so it can be shutdown from main.rs.

quickwit/quickwit-proto/src/lib.rs
congx4: when I run `cargo check --all-targets && cargo test `in local, it is just warning: 
```
warning: unused `Result` that must be used
   --> quickwit-proto/src/lib.rs:199:5
    |
199 |     Span::current().set_parent(parent_cx);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this `Result` may be an `Err` variant, which should be handled
    = note: `#[warn(unused_must_use)]` (part of `#[warn(unused)]`) on by default
help: use `let _ = ...` to ignore the resulting value
    |
199 |     let _ = Span::current().set_parent(parent_cx);
    |     +++++++
```

But during CI phase, it shows as an error.

quickwit/quickwit-cli/src/logger.rs
```diff
@@ -92,45 +92,52 @@ pub fn setup_logging_and_tracing(
 
     // Note on disabling ANSI characters: setting the ansi boolean on event format is insufficient.
     // It is thus set on layers, see https://github.com/tokio-rs/tracing/issues/1817
-    if get_bool_from_env(QW_ENABLE_OPENTELEMETRY_OTLP_EXPORTER_ENV_KEY, false) {
+    let provider = if get_bool_from_env(QW_ENABLE_OPENTELEMETRY_OTLP_EXPORTER_ENV_KEY, false) {
```
guilload:
```suggestion
    let provider_opt = if get_bool_from_env(QW_ENABLE_OPENTELEMETRY_OTLP_EXPORTER_ENV_KEY, false) {
```

quickwit/quickwit-cli/src/main.rs
```diff
@@ -99,7 +98,7 @@ async fn main_impl() -> anyhow::Result<()> {
     start_jemalloc_metrics_loop();
 
     let build_info = BuildInfo::get();
-    let env_filter_reload_fn =
+    let (env_filter_reload_fn, tracer_provider) =
```
guilload:
```suggestion
    let (env_filter_reload_fn, tracer_provider_opt) =
```

guilload (approved):
We use the `_opt` suffix (inconsistently) in this codebase for options.

PR #5369, opened 2024-08-30
add support for f64 for concat fields

PSeitz:

PR #6047, opened 2025-12-23
Improve documentation regarding json type and concatenated json

rdettai-sk:
### Description

Clarify how json types are stored and provide some pointers regarding concat field specificities.

### DIFF
--- a/quickwit/Cargo.lock
+++ b/quickwit/Cargo.lock
@@ -8111,9 +8111,9 @@ checksum = "6c20b6793b5c2fa6553b250154b78d6d0db37e72700ae35fad9387a46f487c97"
 
 [[package]]
 name = "rsa"
-version = "0.9.9"
+version = "0.9.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "40a0376c50d0358279d9d643e4bf7b7be212f1f4ff1da9070a7b54d22ef75c88"
+checksum = "b8573f03f5883dcaebdfcf4725caa1ecb9c15b2ef50c43a07b816e06799bb12d"
 dependencies = [
  "const-oid",
  "digest",


docs/configuration/index-config.md
```diff
@@ -435,8 +435,8 @@ record: basic
 Concatenate fields don't support fast fields, and are never stored. They uses their own tokenizer, independently of the
 tokenizer configured on the individual fields.
 At query time, concatenate fields don't support range queries.
-Only the following types are supported inside a concatenate field: text, bool, i64, u64, f64, json. Other types are rejected
-at index creation, or silently discarded during indexation if they are found inside a json field.
+Only the following types are supported inside a concatenate field: text, bool, i64, u64, json. Other types are rejected
+at index creation, or silently discarded during indexation if they are found inside a json field. f64 is treated specially because search on floats is not very reliable: it is allowed in explicit concatenate field definitions but silently discarded when found in json fields.
```
trinity-1686a: is there more discussion on that rational? While search on floating points definitely has its issues, if feels like "it's not reliable so lets make it purposefully broken" isn't a great argument either.

I guess there might be some concerns of increasing the cost of some queries if we may have one more type to search through at query time, but if that's the underlying reason, we really ought to write it somewhere (and we'd still have to pay that price because we support explicit f64?)

I wonder if f64 just didn't get forgotten because we simply don't see a lot of it

rdettai-sk: I fully agree. I got very disturbed when walking through this part of the code. This bit:
```rust
        JsonValue::Number(number) => {
            if let Some(val) = u64::from_json_number(&number) {
                Some((val).into())
            } else {
                i64::from_json_number(&number).map(|val| (val).into())
            }
        }
```
seems **extremely** deliberate to me (not just forgotten). If you want to fix the behavior, feel free to do so. If not, it deserves an explicit documentation.

trinity-1686a: doing some archeology, f64 were initially not [supported on purpose](https://github.com/quickwit-oss/quickwit/pull/4773#discussion_r1542307075). At this time, concatenate fields worked by essentially .to_string() everything, which didn't work well at all for f64.
Then concatenate fields were changed to [leverage json fields instead](https://github.com/quickwit-oss/quickwit/pull/4937), and shortly after, [f64 support was actually added](https://github.com/quickwit-oss/quickwit/pull/5369).
Lastly the documentation was later updated [to indicate f64 support](https://github.com/quickwit-oss/quickwit/pull/5370).

So afaict, it was purposefully not supported when f64 weren't supported, but got forgotten when f64 support was finally added. It's not consistent at all that a f64 in a dedicated field gets indexed, but a f64 inside a json field doesn't, so i think we should just fix that

rdettai-sk: I'll update this when https://github.com/quickwit-oss/quickwit/pull/6074 is merged 🙂

docs/configuration/index-config.md
```diff
@@ -552,6 +552,15 @@ src.port:53
 src.port:53 AND query_params.ctk:e42bb897d
 ```
 
+Stored primitive types are inferred from the JSON value types using the following rules:
+- a boolean value `true` of `false` is stored as `bool`
```
trinity-1686a:
```suggestion
- a boolean value `true` or `false` is stored as `bool`
```

docs/configuration/index-config.md
```diff
@@ -552,6 +552,15 @@ src.port:53
...
+- numeric values are cast to the first compatible format between `i64`, `u64` or
+  `f64` (in this order)
+- for string values (surrounded with quotes), Tantivy attempts to parse a date
+  in `rfc3339` format. If the parsing fails, the value is stored as `text` using
```
trinity-1686a: i'm not sure that's what is actually happening, following this thread https://github.com/quickwit-oss/quickwit/pull/6074#discussion_r2664891288

rdettai-sk: For regular json types (and dynamic mappings of course) it is. It's only concatenate fields where the JSON text value doesn't go through this parsing. I clarified this now.

PR #6053, opened 2025-12-24
Fix `test_readiness_updates` flaky test

### LINKED ISSUE #6052: Flaky test in quickwit-serve crate
## Summary
A test in the quickwit-serve crate showed flaky behavior in one of the test runs.

## Flaky Test
- `tests::test_readiness_updates`

## Behavior
This test appeared as flaky in run 2:
- Run 2: FLAKY 2/6 [0.069s]

## Test Environment
The test was run with `make test-all` and showed flaky behavior in 1 out of 4 runs.

## Impact
This flaky test may indicate:
- Race conditions in readiness update logic
- Timing-sensitive assertions
- Issues with service startup/shutdown sequencing
- Network or port binding conflicts

## Next Steps
1. Investigate the `test_readiness_updates` test implementation
2. Check for race conditions in readiness update mechanisms
3. Review timing assumptions in the test
4. Consider adding more robust synchronization or retry logic
5. Verify test isolation and cleanup procedures

guilload:
### Description
Introduce `assert_eventually` macro to make the test more robust to slowness. Fixes #6052.

### How was this PR tested?
`c t --manifest-path quickwit/Cargo.toml -p quickwit-serve -- test_readiness_updates`

### DIFF
--- a/quickwit/quickwit-common/src/lib.rs
+++ b/quickwit/quickwit-common/src/lib.rs
@@ -189,6 +189,32 @@ pub fn no_color() -> bool {
 }
 
 #[macro_export]
+macro_rules! assert_eventually {
+    ($cond:expr, $timeout:expr, $interval:expr) => {
+        let start = std::time::Instant::now();
+        loop {
+            if $cond {
+                break;
+            }
+            if start.elapsed() > $timeout {
+                panic!(
+                    "assertion failed: condition `{}` never became true within {} ms",
+                    stringify!($cond),
+                    $timeout.as_millis()
+                );
+            }
+            tokio::time::sleep($interval).await;
+        }
+    };
+    ($cond:expr, $timeout:expr) => {
+        assert_eventually!($cond, $timeout, std::time::Duration::from_millis(50));
+    };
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-common/src/lib.rs
```diff
@@ -188,6 +188,32 @@ pub fn no_color() -> bool {
...
+                    "assertion failed: condition `{}` never became true within {} ms",
+                    stringify!($cond),
+                    $timeout.as_millis()
+                );
+            }
+            tokio::time::sleep($interval).await;
+        }
+    };
+    ($cond:expr, $timeout:expr) => {
+        assert_eventually!($cond, $timeout, std::time::Duration::from_millis(100));
```
trinity-1686a: can we make the interval smaller? In the callsite the *timeout* was 1/4th of that already

guilload: Sure. 25 millis was probably too aggressive and we already have a bunch of tests with 100, but I guess since we are looping we can decrease to 50.

Issue #4699, opened 2024-03-07
`test_delete_shard_on_eof` is flaky


guilload:
It happens occasionally when running `make test-all` on my laptop.
```
---- control_plane::tests::test_delete_shard_on_eof stdout ----
thread 'control_plane::tests::test_delete_shard_on_eof' panicked at quickwit-control-plane/src/control_plane.rs:1513:9:
assertion `left == right` failed
  left: 2
 right: 1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #6064, opened 2025-12-29
Fix flaky test `test_delete_shard_on_eof`

Abdul-Andha:
### Description

See https://github.com/quickwit-oss/quickwit/issues/4699 

There is a small chance (~1%) extra requests are in `client_inbox` right after the sleep. They are irrelevant to this part of the test anyway, so I made an edit to only check the last `apply_plan_request`

### How was this PR tested?

- `make test-all`
- Ran the test many times in a loop


quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -1948,9 +1948,8 @@ mod tests {
         let apply_plan_requests = client_inbox.drain_for_test_typed::<ApplyIndexingPlanRequest>();
         assert!(!apply_plan_requests.is_empty());
```
guilload: This assert can be removed because of the following `last().unwrap()`

guilload (approved):

PR #6067, opened 2025-12-30
Introduce cargo machete to remove unused deps 

congx4:
### Description

1. Introduce cargo machete to remove unused deps
2. Enforce it in CI
3. Add it to `make fix` command.

### How was this PR tested?

cargo check --all-targets && cargo test && make test-all

running `make fix `before removing unused deps : 
Screenshot 2025-12-30 at 1 05 17 PM

running `make fix `after removing unused deps: 
Screenshot 2025-12-30 at 1 10 27 PM

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -19,7 +19,7 @@ env:
   AWS_ACCESS_KEY_ID: "placeholder"
   AWS_SECRET_ACCESS_KEY: "placeholder"
   CARGO_INCREMENTAL: 0
-  PUBSUB_EMULATOR_HOST: "localhost:9898"
+  PUBSUB_EMULATOR_HOST: "localhost:8681"
   QW_DISABLE_TELEMETRY: 1
   QW_S3_ENDPOINT: "http://localhost:4566" # Services are exposed as localhost because we are not running coverage in a container.
   QW_S3_FORCE_PATH_STYLE_ACCESS: 1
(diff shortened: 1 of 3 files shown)


quickwit/Makefile
```diff
@@ -21,6 +21,11 @@ fix:
 	@echo "Running cargo clippy --fix"
 	@cargo clippy --workspace --all-features --tests --fix --allow-dirty --allow-staged
 	@$(MAKE) fmt
+	@$(MAKE) unused-deps
+
+unused-deps:
+	@echo "Checking for unused dependencies"
+	@cargo machete
```
guilload, line 29: Check whether cargo machete is installed and display the command to install if not the same way it's done for the nightly toolchain in fmt

guilload: Should we run with `--fix`?

congx4: We can run this `--fix`, but it also removes false positives dependencies as well.

guilload: Ok. Let's not run with `--fix` if ti removes false positives as well then.

quickwit/Makefile
```diff
@@ -21,6 +21,11 @@ fix:
 	@echo "Running cargo clippy --fix"
 	@cargo clippy --workspace --all-features --tests --fix --allow-dirty --allow-staged
 	@$(MAKE) fmt
+	@$(MAKE) unused-deps
+
+unused-deps:
+	@echo "Removing unused dependencies"
+	@(cargo --list | ( ! grep -q machete && echo "cargo-machete is not installed. Please install using 'cargo install cargo-machete'.") ) || cargo machete --fix
```
guilload: Testing with `command -v cargo-machete` should also work

congx4: yes, I included it. But it only works when it is in your $PATH, so I keep the fallback check.

PR #6074, opened 2026-01-05
fix f64 not working properly with concat fields

trinity-1686a:
### Description

fix f64 in concat being handled only on f64 fields but not in json fields
see https://github.com/quickwit-oss/quickwit/pull/6047/changes#r2662159497

### How was this PR tested?

added integration test


quickwit/quickwit-doc-mapper/src/doc_mapper/mapping_tree.rs
```diff
@@ -147,8 +147,10 @@ pub(crate) fn map_primitive_json_to_tantivy(value: JsonValue) -> Option<TantivyV
         JsonValue::Number(number) => {
             if let Some(val) = u64::from_json_number(&number) {
```
PSeitz: the order should be i64, u64, f64

quickwit/rest-api-tests/scenarii/concat_fields/0001_concat_field.yaml
```diff
@@ -115,3 +115,39 @@ params:
 expected:
   num_hits: 1
 ---
+endpoint: concat/search
+params:
+  query: "concat_default:1.5"
```
PSeitz, line 120: we should also test values above i64::MAX

quickwit/quickwit-doc-mapper/src/doc_mapper/mapping_tree.rs
```diff
@@ -145,10 +145,12 @@ pub(crate) fn map_primitive_json_to_tantivy(value: JsonValue) -> Option<TantivyV
...
         JsonValue::Bool(val) => Some((val).into()),
         JsonValue::Number(number) => {
-            if let Some(val) = u64::from_json_number(&number) {
+            if let Some(val) = i64::from_json_number(&number) {
+                Some((val).into())
+            } else if let Some(val) = u64::from_json_number(&number) {
                 Some((val).into())
             } else {
-                i64::from_json_number(&number).map(|val| (val).into())
+                f64::from_json_number(&number).map(|val| (val).into())
```
rdettai-sk, line 153: I think that with this change, the conversion is now equivalent to the `From<serde_json::Value> for OwnedValue {}` implemented in Tantivy, except for the extra conversion in the Tantivy implem if the string is an RFC339 date 😉

trinity-1686a: i don't think we want to map to a Date, concatenated fields don't support fastfields (and range requests) so there isn't a clear advantage at doing the conversion, but that would prevent wildcard queries that can currently be used as a poor man's range

`impl From<serde_json::Value> for OwnedValue` also handle Object and Array, which shouldn't matter as it's handled by the caller already

rdettai-sk: I'll add a comment in https://github.com/quickwit-oss/quickwit/pull/6047 to explicit these specificity

PR #6078, opened 2026-01-07
Fix infinite loop OOM bug caused by rebalancing shards from inactive ingesters

nadav-govari:
Rebalances all shards from unavailable ingesters. Moves to only look at ingesters with more shards than allowed as candidates for rebalancing. Also moves to integer division to simplify target calculation.

### DIFF
--- a/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
+++ b/quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
@@ -3544,7 +3544,8 @@ mod tests {
     /// - `available_ingester_shards`: open shards per available ingester
     /// - `unavailable_ingester_shards`: open shards on unavailable ingesters
     fn test_compute_shards_to_rebalance_aux(
-        available_ingester_shards: &[usize], unavailable_ingester_shards: &[usize],
+        available_ingester_shards: &[usize],
+        unavailable_ingester_shards: &[usize],
     ) {
         let index_id = "test-index";
         let index_metadata = IndexMetadata::for_test(index_id, "ram://indexes/test-index");


quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -1103,96 +1102,78 @@ impl IngestController {
...
 
-        let mut rng = rng();
-        let mut per_leader_open_shard_shuffled: Vec<Vec<&Shard>> = per_leader_open_shards
-            .into_values()
-            .map(|mut shards| {
-                shards.shuffle(&mut rng);
-                shards
-            })
-            .collect();
+        let target_max = total_open_shards.div_ceil(num_available_ingesters);
```
fulmicoton: I think the point of using 0.9/1.1 was to add some tolerance to make sure rebalance do not happen too often.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -1103,96 +1102,78 @@ impl IngestController {
...
         }
-        if num_ingesters < 2 {
+        if num_available_ingesters < 2 {
             return Vec::new();
         }
-        let mut num_open_shards: usize = 0;
-        let mut per_leader_open_shards: HashMap<&str, Vec<&Shard>> = HashMap::new();
+
+        let mut total_open_shards: usize = 0;
+        let mut orphaned_shards: Vec<Shard> = Vec::new();
```
fulmicoton: counting orphaned shards as needing to be moved is dangerous. 

Right now moving a shard is done by opening a shard, and closing the shard moved.
Closing a shard requires contacting the node hosting it, so it can fail. In this exact case, it WILL fail because the ingester is not even in the pool.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -1103,56 +1103,42 @@ impl IngestController {
...
-            return Vec::new();
-        }
-        let mut num_open_shards: usize = 0;
-        let mut per_leader_open_shards: HashMap<&str, Vec<&Shard>> = HashMap::new();
+        let mut per_available_ingester_shards: HashMap<String, Vec<&Shard>> = self
+            .ingester_pool
+            .keys()
+            .into_iter()
+            .map(|ingester_id| (ingester_id.as_str().to_string(), Vec::new()))
+            .collect();
```
guilload, line 1122: nit: I'd rather deal with `NodeId` instead of String:
```rust
let mut per_available_ingester_shards: HashMap<NodeId, Vec<&Shard>> 
...
.map(|ingester_id| (ingester_id, Vec::new()))
...
let leader_id_ref = NodeIdRef::from_str(&shard.leader_id);
if let Some(shards) = per_available_ingester_shards.get_mut(leader_id_ref) {

File: quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs [hunk 4]
```diff
@@ -3544,7 +3537,14 @@ mod tests {
         );
     }
 
-    fn test_compute_shards_to_rebalance_aux(shard_allocation: &[usize]) {
+    /// Test helper for compute_shards_to_rebalance_v2.
+    ///
+    /// - `active_ingester_shards`: shards per active ingester (these are in the pool)
+    /// - `inactive_ingester_shards`: shards per inactive ingester (NOT in the pool - orphaned)
```
guilload:
```suggestion
    /// - `available_ingester_shards`: shards per available ingester (shards whose leader is in the pool)
    /// - `unavailable_ingester_shards`: shards per unavailable ingester (NOT in the pool - orphaned)
```

guilload: I don't understand the notion of orphaned. Aren't the shards whose leader is in NOT the pool the orphans already?

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -3544,7 +3537,14 @@ mod tests {
...
+    fn test_compute_shards_to_rebalance_aux(
+        active_ingester_shards: &[usize],
+        inactive_ingester_shards: &[usize],
```
guilload: `s/active/available/` everywhere in this function.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: Yeah, the comments got jumbled in between revisions. Let me fix.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: done

PR #6089, opened 2026-01-12
Make regex lenient to start and end anchors

rdettai-sk:
### Description

tantivy_fst returns an error if the regex starts with ^ or ends with $. This is a very strict behavior as strictly speaking it always behaves as if these were present. This PR strips these anchors. This is also how ES behaves.

### How was this PR tested?

ES compatibility integration tests executed on both QW and ES


rdettai-sk:
@trinity-1686a something like this: https://github.com/quickwit-inc/fst/pull/24/files ?

rdettai-sk:
For the record: ES does support regexes like the one suggested by Trinity, so this fix is incomplete.

PR #6092, opened 2026-01-13
Initial stab at AZ awareness

nadav-govari:
### Description
Adds an env var for availability zone, and makes it optional on the node config. It's propagated to node state and also gossiped. A follow-up PR will introduce AZ awareness at further levels.

### How was this PR tested?
Unit tests pass, but no other functionality exists than optional configs. No new tests.

### DIFF
--- a/quickwit/quickwit-cluster/src/member.rs
+++ b/quickwit/quickwit-cluster/src/member.rs
@@ -80,7 +80,7 @@ impl NodeStateExt for NodeState {
     }
 
     fn availability_zone(&self) -> Option<String> {
-        self.get(AVAILABILITY_ZONE_KEY).map(|x| x.to_string())
+        self.get(AVAILABILITY_ZONE_KEY).map(|az| az.to_string())
     }
 }
 
--- a/quickwit/quickwit-config/src/config_value.rs
+++ b/quickwit/quickwit-config/src/config_value.rs
@@ -41,18 +41,17 @@ where
         }
     }
 
-    #[cfg(test)]
-    pub(crate) fn for_test(value: T) -> Self {
+    pub(crate) fn none() -> Self {
         Self {
-            provided: Some(value),
+            provided: None,
             default: None,
         }
     }
 
     #[cfg(test)]
-    pub(crate) fn none() -> Self {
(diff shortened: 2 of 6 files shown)


quickwit/quickwit-cluster/src/member.rs
```diff
@@ -74,6 +78,10 @@ impl NodeStateExt for NodeState {
             .map(|(key, value)| key.len() + value.value.len() + SIZE_OF_VERSION + SIZE_OF_TOMBSTONE)
             .sum()
     }
+
+    fn availability_zone(&self) -> Option<String> {
+        self.get(AVAILABILITY_ZONE_KEY).map(|x| x.to_string())
```
guilload:
```suggestion
        self.get(AVAILABILITY_ZONE_KEY).map(|az| az.to_string())
```
We tend to avoid variable names such as `x`, `i`, `e`, etc.

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -169,6 +169,8 @@ struct NodeConfigBuilder {
     cluster_id: ConfigValue<String, QW_CLUSTER_ID>,
     #[serde(default = "default_node_id")]
     node_id: ConfigValue<String, QW_NODE_ID>,
+    #[serde(default)]
+    availability_zone: ConfigValue<String, QW_AVAILABILITY_ZONE>,
```
guilload, line 177: Populate the `availability_zone` attribute in the `quickwit-config/resources/tests/node_config/quickwit.{json,toml,yaml|` test files and update the corresponding unit test. The goal is to catch regressions and testing backward compatibility when modifying the node config object.

quickwit/quickwit-config/src/config_value.rs
```diff
@@ -41,18 +41,17 @@ where
         }
     }
 
-    #[cfg(test)]
```
guilload, line 44: What happened here? :D

nadav-govari: This is the first config that has a default value of none apparently, so I needed this little helper, since the struct fields aren't public!

PR #6099, opened 2026-01-16
Dependabot/docker/node 25

Platane:
### Description

Bumps node from 20 to 24.


.devcontainer/devcontainer.json
```diff
@@ -33,7 +33,7 @@
             "moby": true
         },
         "ghcr.io/devcontainers/features/node:1": {
-            "version": "18"
+            "version": "24"
```
Platane, line 36: I tried to launch the devcontainer, it works (with some tweaks https://github.com/quickwit-oss/quickwit/pull/6100)

Issue #5693, opened 2025-02-24
index_filter support for elastic search compatible endpoints


ruoliu2:
**Is your feature request related to a problem? Please describe.**
Need index filter support for large scale dataset, give good entry point for understanding data.

**Describe the solution you'd like**
support index_filter

**Describe alternatives you've considered**
NA

ruoliu2:
- yes I'm referring to the field capabilities API
- I just updated the description some more, do you think it's good now?

rdettai:
It's perfect thanks! Do you want to give it a try?

ruoliu2:
yeah I'd love to! I'll take a look this weekend

ruoliu2:
Hi, I took a stab at this, and to trigger CI I need approval from maintainer, can I get an approval for my pr?

PR #6102, opened 2026-01-17
feat(es-compat): add index_filter support for field capabilities API

ruoliu2:
## Summary

- Adds `index_filter` parameter support for the ES-compatible `_field_caps` endpoint
- Parses ES Query DSL (term, bool, range, match, etc.) and converts to internal QueryAst
- Passes query through to leaf nodes for future split-level filtering
- Adds unit tests and REST API integration tests

## Test plan

- [x] Unit tests for index_filter parsing (6 tests)
- [x] REST API integration tests for various query types
- [x] Verify invalid queries return 400 Bad Request
- [x] Existing field_caps tests still pass

## Notes

This implementation accepts and parses the `index_filter` parameter for API compatibility with Elasticsearch. Full split-level document filtering (executing queries to prune splits with no matching documents) can be added as a follow-up enhancement.

Closes https://github.com/quickwit-oss/quickwit/issues/5693

### DIFF
--- a/quickwit/quickwit-ui/e2e/homepage.spec.ts
+++ b/quickwit/quickwit-ui/e2e/homepage.spec.ts
@@ -16,7 +16,7 @@ import { expect, test } from "@playwright/test";
 
 test.describe("Home navigation", () => {
   test("Should display sidebar links", async ({ page }) => {
-    await page.goto("/");
+    await page.goto("/ui");
     await expect(page.locator("a")).toContainText([
       "Query editor",
       "Indexes",
@@ -25,7 +25,7 @@ test.describe("Home navigation", () => {
   });
 
   test("Should navigate to cluster state", async ({ page }) => {
-    await page.goto("/");
+    await page.goto("/ui");
     await page.getByRole("link", { name: "Cluster" }).click();
     await expect(page.getByLabel("breadcrumb")).toContainText("Cluster");
     await expect(page.getByText("cluster_id")).toBeVisible();


quickwit/quickwit-serve/src/elasticsearch_api/model/field_capability.rs
```diff
@@ -177,12 +180,182 @@ pub fn convert_to_es_field_capabilities_response(
 pub fn build_list_field_request_for_es_api(
     index_id_patterns: Vec<String>,
     search_params: FieldCapabilityQueryParams,
-    _search_body: FieldCapabilityRequestBody,
+    search_body: FieldCapabilityRequestBody,
 ) -> Result<quickwit_proto::search::ListFieldsRequest, ElasticsearchError> {
+    // Parse index_filter if provided
+    let query_ast_json: Option<String> = if search_body.index_filter.is_null()
```
fulmicoton-dd: Can you make a well-named function with the prototype, to implement this block of code?

`(serde_json::Value) -> anyhow::Result<Option<QueryAst>`

quickwit/rest-api-tests/scenarii/es_field_capabilities/0001-field-capabilities.yaml
```diff
@@ -334,3 +334,123 @@ engines:
...
+      field: "value"
+status_code: 400
+---
+# Test _field_caps API with empty index_filter (should work like no filter)
+method: [POST]
+engines:
+  - quickwit
+endpoint: fieldcaps/_field_caps?fields=name
+json_body:
+  index_filter: {}
```
fulmicoton-dd, line 415: kudos for checking these edge cases too!

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -310,13 +310,29 @@ impl FieldPattern {
...
     searcher_context: &SearcherContext,
     split_ids: &[SplitIdAndFooterOffsets],
     field_patterns_str: &[String],
+    query_ast: Option<&str>,
 ) -> crate::Result<ListFieldsResponse> {
+    // Log if index_filter was provided (for observability)
+    // Full query-based split filtering is a TODO for future enhancement
+    if query_ast.is_some() {
+        tracing::debug!(
+            "index_filter provided for field capabilities, but split-level filtering is not yet \
```
fulmicoton-dd: We should not merge this if it is not implemented.

quickwit/rest-api-tests/scenarii/es_field_capabilities/0001-field-capabilities.yaml
```diff
@@ -334,3 +334,123 @@ engines:
   - elasticsearch
 endpoint: doesno*texist/_field_caps?fields=date
 status_code: 200
+---
+# Test _field_caps API with index_filter (term query)
+method: [POST]
+engines:
+  - quickwit
```
rdettai-sk: why are you testing agains QW only? If this PR's goal is to increase the compatibility with ES it should be tested agains ES as well.

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -310,13 +310,29 @@ impl FieldPattern {
 }
 
 /// `leaf` step of list fields.
+///
+/// If `query_ast` is provided (from ES-compatible `index_filter`), it indicates that
+/// the caller wants to filter fields based on documents matching the query.
+///
+/// Note: Full query-based filtering is not yet implemented. When `query_ast` is provided,
+/// fields from all documents are currently returned. This matches the ES behavior of
+/// returning field capabilities for indices where at least one document matches.
```
rdettai-sk: The [ES spec](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-field-caps#operation-field-caps-body-application-json-index_filter) sais the following:
> Filter indices if the provided query rewrites to match_none on every shard.
> IMPORTANT: The filtering is done on a best-effort basis, it uses index statistics and mappings to rewrite queries to match_none instead of fully running the request. For instance a range query over a date field can rewrite to match_none if all documents within a shard (including deleted documents) are outside of the provided range. However, not all queries can rewrite to match_none so this API may return an index even if the provided filter matches no document.

Given that in QW we don't have the necessary info to know much about the fields in the metadata, we need to open the splits. I think nevertheless that we need to aim for a lightweight query execution, even if that returns too many fields.

quickwit/quickwit-serve/src/elasticsearch_api/model/field_capability.rs
```diff
@@ -177,12 +180,182 @@ pub fn convert_to_es_field_capabilities_response(
...
+
+        Some(serde_json::to_string(&query_ast).expect("QueryAst should be JSON serializable"))
+    };
+
     Ok(quickwit_proto::search::ListFieldsRequest {
         index_id_patterns,
         fields: search_params.fields.unwrap_or_default(),
         start_timestamp: search_params.start_timestamp,
         end_timestamp: search_params.end_timestamp,
+        query_ast: query_ast_json,
```
rdettai-sk: You need to leverage the two metastore pruning capabilities:
- refine start and end time using the query ast
- extract tags from query ast and provide them to split listing call

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -309,20 +314,127 @@ impl FieldPattern {
     }
 }
 
+/// Checks if any documents in the split match the query.
+/// Returns true if at least one document matches, false otherwise.
+///
+/// This is a lightweight query execution that only counts matches without
+/// materializing documents, used for split-level filtering in field capabilities.
```
rdettai-sk: I would not qualify "count" as lightweight here. It will actually involve downloading every chunk of the split (fast field, index including postings) that are necessary to solve the query. I don't think this is the original intent of the `index_filter` parameter on the endpoint. Can explain for which usecase it is worth going so deep in the split inspection?

ruoliu2: Hey after checking elastic search implementation I fixed it to best effort. Can you take a look with current best effort impl?

quickwit/quickwit-search/src/list_fields.rs
```diff
@@ -399,29 +410,63 @@ pub async fn root_list_fields(
...
+            .map_err(|err| SearchError::InvalidQuery(err.to_string()))?;
+
+        // Refine time range from query AST if timestamp field is available
+        if let Some(ref timestamp_field) = timestamp_field_opt {
+            crate::root::refine_start_end_timestamp_from_ast(
+                &query_ast,
+                timestamp_field,
+                &mut start_timestamp,
+                &mut end_timestamp,
+            );
```
rdettai-sk, line 457: now that we are extracting timestamps here I think we should deprecate start_timestamp and end_timestamp in FieldCapabilityQueryParams which are not compatible with ES

quickwit/rest-api-tests/scenarii/es_field_capabilities/0001-field-capabilities.yaml
rdettai-sk: Have you tried running this these tests agains ES? unfortunately that is not performed by the CI currently. Also a test with tags would be perfect!

rdettai-sk (approved):

quickwit/rest-api-tests/scenarii/es_field_capabilities/0001-field-capabilities.yaml
ruoliu2: rebased and ran tests against ES and added tests with tags, need approval for running ci

quickwit/quickwit-search/src/list_fields.rs
ruoliu2: do we want to include this in this pr or do we want it as a followup

ruoliu2:
it looks like merging is still blocked and Im unsure if i have access. Do you think you can help rebase branch, approve ci run workflow and squash and merge this pr? since I dont have permission to run the workflow 👀 @rdettai-sk

rdettai-sk:
@fulmicoton I think this is good to go!

quickwit/rest-api-tests/scenarii/es_field_capabilities/0001-field-capabilities.yaml
```diff
@@ -334,3 +334,140 @@ engines:
   - elasticsearch
 endpoint: doesno*texist/_field_caps?fields=date
 status_code: 200
+---
+# Test _field_caps API with index_filter (term query)
+# Note: term queries require exact token match; 'fritz' is lowercase due to default tokenizer
+method: [POST]
+engines:
+  - quickwit
+  - elasticsearch
```
fulmicoton-dd:
```suggestion
```

Issue #5445, opened 2024-09-24
Tunable Cache Eviction Policies


esatterwhite:
**Is your feature request related to a problem? Please describe.**

We ingest and search on petabytes of data reaching as far back as 90 days old. We maintain multiple terabytes of cache for quickwit. Our Search workload is very front heavy in relation to time. Meaning, the further a given document's timestamp is from `now` the less it is accessed. The data that has been ingested in the last 3 days is generally the most frequently accessed data. However, We have internal jobs that run once a day or once a week that may access older data. Our users may periodically run reports or expensive aggregations across months of data, etc. Because Quickwit is using LRU cache eviction policies, these expensive, one off, and infrequently executed queries can invalidate a large portion of cache resulting in depredated performance across the cluster

**Describe the solution you'd like**
It would be good for more workload to implement means to switch the eviction policy. Namely, switch from the default Least Recently Used (LRU) to a Least Frequently Used (LFU) eviction policy. This would allow quickwit to evict split entries that are much less likely to be accessed again rather than just entries that happen to have a slightly older age than others keeping hot / relevant data in cache.

fulmicoton:
thanks for the accurate description of your issue!

esatterwhite:
@fulmicoton out of curiosity does the split cache operate per index?

Or could it? how complicated would that be?

fulmicoton:
the split cache does not operate per index.

PR #6103, opened 2026-01-18
improve in memory cache

trinity-1686a:
### Description

fix https://github.com/quickwit-oss/quickwit/issues/5445

Add support for 2 new cache families which should be less sensitive to full scan
Add concept of "virtual caches" which don't cache anything, but behave exactly as if they did, and record metrics, to help with cache sizing. They can be of any size (including a size to big for values to fit in memory, though the keys still consume memory as usual) and policy. They default to the size and policy of the real cache if one option isn't provided.
Refactor configuration handling so it can support these new features, while keeping backward compatibility with the old config syntax.

fix some queries querying cache when they can be replied to from split metadata alone, which causes them to never be cached. This caused cache misses to be counted for queries for which touching the cache has no real advantage.

### How was this PR tested?

Tested by creating one lru cache and one of each virtual caches, and verifying that metrics looked accurate. Notably, that repeating one query twice, then sending random queries one by one, and as soon as all cache evicted something, run the first query, results in better results for non lru caches

at the moment there is no real test

### DIFF
--- a/LICENSE-3rdparty.csv
+++ b/LICENSE-3rdparty.csv
@@ -1,6 +1,7 @@
 Component,Origin,License,Copyright
 adler2,https://github.com/oyvindln/adler2,0BSD OR MIT OR Apache-2.0,"Jonas Schievink <jonasschievink@gmail.com>, oyvindln <oyvindln@users.noreply.github.com>"
 advapi32-sys,https://github.com/retep998/winapi-rs,MIT,Peter Atashian <retep998@gmail.com>
+ahash,https://github.com/tkaitchuck/ahash,MIT OR Apache-2.0,Tom Kaitchuck <Tom.Kaitchuck@gmail.com>
 aho-corasick,https://github.com/BurntSushi/aho-corasick,Unlicense OR MIT,Andrew Gallant <jamslam@gmail.com>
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -287,6 +302,89 @@ pub struct SearcherConfig {
...
+    }
+
+    fn deserialize_with_default<'de, D, const DEFAULT_CAPACITY: u64>(
+        deserializer: D,
+    ) -> Result<CacheConfig, D::Error>
+    where D: Deserializer<'de> {
+        use serde_with::{DeserializeAs, FromInto, PickFirst, Same};
+
+        let mut cache_config: CacheConfig =
+            PickFirst::<(Same, FromInto<ByteSize>)>::deserialize_as(deserializer)?;
```
fulmicoton-dd, line 355: TIL about PickFirst.

trinity-1686a: internally, it deserializes to a more exhaustive version of a serde_json::Value, and tries to deserialize that successively for each type param, so it's likely not something we ever want on hot codepath, but for config it's much more convenient than the alternative

PR #6109, opened 2026-01-21
Catch empty AZ env vars

nadav-govari:
### Description
If the env var exists but there was a problem parsing the version, it might return `Some("")`, which we don't want.

### DIFF
--- a/quickwit/quickwit-config/src/config_value.rs
+++ b/quickwit/quickwit-config/src/config_value.rs
@@ -19,6 +19,7 @@ use std::{any, fmt};
 use anyhow::{self, Context};
 use serde::{Deserialize, Deserializer};
 use tracing::log::warn;
+
 use crate::qw_env_vars::{QW_ENV_VARS, QW_NONE};
 
 #[derive(Debug, Clone, Eq, PartialEq)]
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -224,7 +224,11 @@ impl NodeConfigBuilder {
         env_vars: &HashMap<String, String>,
     ) -> anyhow::Result<NodeConfig> {
         let node_id = self.node_id.resolve(env_vars).map(NodeId::new)?;
-        let availability_zone = self.availability_zone.resolve_optional(env_vars)?;
+        let availability_zone = self
+            .availability_zone
+            .resolve_optional(env_vars)?
+            // if the env var is empty, it'll return Some(""), and we don't want that.
+            .filter(|str| !str.is_empty());
```
guilload: Let's push that logic into `resolve_optional`.

quickwit/quickwit-config/src/config_value.rs
```diff
@@ -68,7 +68,7 @@ where
         // QW env vars take precedence over the config file values.
         if E > QW_NONE
             && let Some(env_var_key) = QW_ENV_VARS.get(&E)
-            && let Some(env_var_value) = env_vars.get(*env_var_key)
+            && let Some(env_var_value) = env_vars.get(*env_var_key).filter(|val| !val.is_empty())
```
guilload:
```suggestion
            && let Some(env_var_value) = env_vars.get(*env_var_key).filter(|val| {
              if val.is_empty() {
                warn!("environment variable `{}` is set but value is empty", *env_var_key)
                false
              } else {
                true
              }
            })
```

PR #6117, opened 2026-01-23
[Draft] Rework routing to be node based rather than shard based

nadav-govari:
Today, a specific shard on a specific node is targeted for persisting a request. This is suboptimal; we can pick a node that we know has at least one shard on it and delegate the shard selection logic to it.

In order to do that, the routing table needs to be aware of per-node level shards, not just local/remote shards. This reworks the routing table to maintain a single map of <Node, Vec<Shard>>. A preference is given to the node with the most open shards. TODO: make it psuedorandom, but weigh nodes with more open shards higher.

As a result, some other shapes changes. I tried to be minimally intrusive when making the changes, but there's a lot that can go wrong here (shard updates, etc)

I'm not sure yet as to what the effects will be on retries and things of that nature. So that's also something to keep in mind.

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/publish_tracker.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/publish_tracker.rs
@@ -62,16 +62,6 @@ impl PublishTracker {
         }
     }
 
-    pub fn register_requested_shards<'a>(
-        &'a self,
-        shard_ids: impl IntoIterator<Item = &'a ShardId>,
-    ) {
-        let mut publish_states = self.state.lock().unwrap();
-        for shard_id in shard_ids {
-            publish_states.shard_tracked(shard_id.clone());
-        }
-    }
-
     pub fn track_persisted_shard_position(&self, shard_id: ShardId, new_position: Position) {
         let mut publish_states = self.state.lock().unwrap();
         publish_states.position_persisted(&shard_id, &new_position)
@@ -91,13 +81,11 @@ impl PublishTracker {
 }
 
 enum PublishState {
-    /// The persist request for this shard has been sent
-    Tracked,
     /// The persist request for this shard success response has been received
     /// but the position has not yet been published
     AwaitingPublish(Position),
-    ///  The shard has been published up to this position (might happen before
-    ///  the persist success is received)
+    /// The shard has been published up to this position (might happen before
+    /// the persist success is received via early publish event)
     Published(Position),
 }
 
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -545,25 +519,60 @@ impl Ingester {
...
+                // control plane, which confirms that the shard was correctly opened in the
+                // metastore.
+                shard.is_advertisable = true;
+                let doc_mapper = shard.doc_mapper_opt.clone().expect("shard should be open");
+                let validate_docs = shard.validate_docs;
+                let follower_id_opt = shard.follower_id_opt().cloned();
+                let from_position_exclusive = shard.replication_position_inclusive.clone();
+                let shard_id = match split_queue_id(&queue_id) {
+                    Some((_, _, shard_id)) => Some(shard_id),
+                    None => None,
```
nadav-govari, line 558: todo: helper for queue_id for this

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -545,25 +519,60 @@ impl Ingester {
...
+                // means that this node cannot persist this request, and it should be retried.
                 if !rate_limiter.acquire_bytes(requested_capacity) {
                     debug!("failed to persist records to shard `{queue_id}`: rate limited");
 
                     let persist_failure = PersistFailure {
                         subrequest_id: subrequest.subrequest_id,
                         index_uid: subrequest.index_uid,
                         source_id: subrequest.source_id,
-                        shard_id: subrequest.shard_id,
+                        shard_ids: vec![shard_id.unwrap()],
```
nadav-govari, line 575: same helper for queue_id as above should take care of this

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -741,23 +751,28 @@ impl Ingester {
...
                                 );
-                                shards_to_delete.insert(queue_id);
+                                shards_to_delete.insert(queue_id.clone());
                                 PersistFailureReason::ShardNotFound
                             }
                         };
+                        let shard_id = match split_queue_id(&queue_id) {
+                            Some((_, _, shard_id)) => Some(shard_id),
+                            None => None,
+                        };
```
nadav-govari, line 769: todo: same as above

PR #6119, opened 2026-01-26
grant permits in SRTF fashion

trinity-1686a:
### Description

When scheduling leaf splits, allow up to N queries to run concurrently so that a single large query doesn't hog all ressources on a searcher.

### How was this PR tested?

added unit test to verify it emits permits in the expected order

### DIFF
--- a/quickwit/quickwit-search/src/search_permit_provider.rs
+++ b/quickwit/quickwit-search/src/search_permit_provider.rs
@@ -201,15 +201,11 @@ impl LeafPermitRequest {
 
     fn pop_if_smaller_than(&mut self, max_size: u64) -> Option<SingleSplitPermitRequest> {
         // IntoIter::as_slice() allows us to peek at the next element without consuming it
-        if self
-            .single_split_permit_requests
-            .as_slice()
-            .first()
-            .is_some_and(|request| request.permit_size <= max_size)
-        {
-            self.single_split_permit_requests.next()
-        } else {
-            None
+        match self.single_split_permit_requests.as_slice().first() {
+            Some(request) if request.permit_size <= max_size => {
+                self.single_split_permit_requests.next()
+            }
+            _ => None,
         }
     }
 
@@ -275,12 +271,9 @@ impl SearchPermitActor {
         if self.num_warmup_slots_available == 0 {
             return None;
         }
-        let Some(available_memory) = self
+        let available_memory = self
             .total_memory_budget
-            .checked_sub(self.total_memory_allocated)
-        else {
-            return None;
-        };
+            .checked_sub(self.total_memory_allocated)?;
         let mut peeked = self.permits_requests.peek_mut()?;
 
         if let Some(permit_request) = peeked.pop_if_smaller_than(available_memory) {


quickwit/quickwit-search/src/search_permit_provider.rs
```diff
@@ -78,14 +78,20 @@ pub fn compute_initial_memory_allocation(
 }
 
 impl SearchPermitProvider {
-    pub fn new(num_download_slots: usize, memory_budget: ByteSize) -> Self {
+    pub fn new(
+        num_download_slots: usize,
+        num_concurrent_query: usize,
```
guilload:
```suggestion
        num_concurrent_queries: usize,
```

quickwit/quickwit-search/src/search_permit_provider.rs
```diff
@@ -196,13 +237,30 @@ impl SearchPermitActor {
...
             return None;
         }
-        if let Some((_, next_permit_size)) = self.permits_requests.front()
-            && self.total_memory_allocated + next_permit_size <= self.total_memory_budget
+        let (leaf, leaf_position) = if let Some(leaf_permit_request) =
+            self.permits_requests.get_mut(self.next_query_to_serve)
+        {
+            (leaf_permit_request, self.next_query_to_serve)
+        } else if let Some(leaf_permit_request) = self.permits_requests.front_mut() {
+            (leaf_permit_request, 0)
```
rdettai-sk: nit: for easier proof reading I think we can mutate `self.next_query_to_serve` instead of returning `leaf_position`

quickwit/quickwit-search/src/search_permit_provider.rs
```diff
@@ -134,11 +135,72 @@ struct SearchPermitActor {
...
+impl PartialEq for LeafPermitRequest {
+    fn eq(&self, other: &Self) -> bool {
+        self.cmp(other).is_eq()
+    }
+}
+
+impl Eq for LeafPermitRequest {}
+
+impl LeafPermitRequest {
+    fn from_estimated_costs(permit_sizes: Vec<u64>) -> (Self, Vec<SearchPermitFuture>) {
```
fulmicoton-dd, line 180: Can you add a comment highlighting that the search permit futures are ordered like the permits?

quickwit/quickwit-search/src/search_permit_provider.rs
```diff
@@ -134,11 +135,72 @@ struct SearchPermitActor {
...
     total_memory_budget: u64,
     total_memory_allocated: u64,
-    permits_requests: VecDeque<(oneshot::Sender<SearchPermit>, u64)>,
+    permits_requests: BinaryHeap<LeafPermitRequest>,
     #[cfg(test)]
     stopped: watch::Sender<bool>,
 }
 
+struct LeafPermitRequest {
+    single_split_permit_requests: Vec<(oneshot::Sender<SearchPermit>, u64)>,
```
fulmicoton-dd: Can you add a comment explaining that the `single_split_permit_requests` are sorted in reverse order of their priority to make it simpler to pop them? 

very arguable nitpick:
Another solution could have been to use `std::vec::IntoIter<T>` or `Peek<IntoIter<T>>`. 
(IntoIter is already peekable via as_slice())

fulmicoton-dd: The nature of the u64 was REALLY unclear to me. Can we transform that pair into a struct?

quickwit/quickwit-search/src/search_permit_provider.rs
```diff
@@ -196,13 +255,20 @@ impl SearchPermitActor {
     }
 
     fn pop_next_request_if_serviceable(&mut self) -> Option<(oneshot::Sender<SearchPermit>, u64)> {
-        if self.num_warmup_slots_available == 0 {
+        if self.num_warmup_slots_available == 0
+            || self.total_memory_budget <= self.total_memory_allocated
```
fulmicoton-dd: nitpick.

One way to get both the early exit, and the safe difference is

let Some(available_memory) = self.total_memory_budget.checked_sub(self.total_memory_allocated) else {
    return None;
};

The case where it is 0  will be caught by the pop_if_smaller_than.

fulmicoton-dd:
@trinity-1686a your impl is light and clean gg.

PR #6125, opened 2026-01-28
Remove unused features from 6 dependencies

fulmicoton-dd:
## Summary

- Disable unused default features from 6 dependencies to reduce compile times and binary size
- **hyper-util**: Changed from `full` to specific features (`client-legacy`, `server-auto`, `server-graceful`, `service`, `tokio`)
- **tokio-util**: Changed from `full` to specific features (`compat`, `io-util`)
- **prometheus**: Disabled `protobuf` default (only `TextEncoder` is used)
- **dialoguer**: Disabled `editor` and `password` defaults (only `Confirm` is used)
- **zstd**: Disabled `legacy`, `arrays`, and `zdict_builder` defaults
- **env_logger**: Disabled `humantime` and `regex` defaults (timestamps are disabled anyway)

## Test plan

- [x] `cargo check --workspace` passes

### DIFF
--- a/quickwit/Cargo.toml
+++ b/quickwit/Cargo.toml
@@ -102,11 +102,11 @@ colored = "3.0"
 console-subscriber = "0.5"
 criterion = { version = "0.8", features = ["async_tokio"] }
 cron = "0.15"
-dialoguer = "0.12"
+dialoguer = { version = "0.12", default-features = false }
 dotenvy = "0.15"
 dyn-clone = "1.0"
 enum-iterator = "2.3"
-env_logger = "0.11"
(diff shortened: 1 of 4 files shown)


quickwit/Cargo.toml
```diff
@@ -299,7 +308,7 @@ vrl = { version = "0.29", default-features = false, features = [
 warp = { version = "0.4", features = ["server", "test"] }
 whichlang = "0.1"
 wiremock = "0.6"
-zstd = "0.13"
+zstd = { version = "0.13", default-features = false }
```
fulmicoton, line 311: This removes legacy. I checked the legacy format is really old and predates Quickwit by a large margin.

PR #6127, opened 2026-01-29
Bump bytesize major version; fix backwards incompatible changes

nadav-govari:
### Description

The primary problem with this was that bytesize moved to read "50M" as 50MiB, whereas our code expects it to be 50MB frequently. To keep this consistent and to introduce minimal changes to our system, primarily for tests, we can ser/de bytesize into the number of bytes, and then read it in decimal or binary.

### How was this PR tested?

All existing tests pass.


quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -555,23 +555,23 @@ impl IngestApiConfig {
         ensure!(
             self.max_queue_disk_usage > ByteSize::mib(256),
             "max_queue_disk_usage must be at least 256 MiB, got `{}`",
-            self.max_queue_disk_usage
+            self.max_queue_disk_usage.display().si()
```
congx4, line 558: nice!

PR #6133, opened 2026-01-30
Always refill permits on rate limiter; use governor clock instead of std

nadav-govari:
Calling `available_permits()` didn't actually supply an accurate portrayal of capacity, as enough time may have elapsed that a refill is needed. So now a refill is automatically triggered on any call to `available_permits()` or `acquire()`.

Additionally, the `governor` crate has a custom implementation of a clock that is faster than std::time. I benchmarked it as follows:

```
    while benchmark_start.elapsed() < Duration::from_seconds(20) {
        let acquire_start = Instant::now();
        let acquired = rate_limiter.acquire(100_000);
        let latency = acquire_start.elapsed();

        latencies.push(latency);
        iterations += 1;

        if acquired {
            successful_acquires += 1;
        } else {
            failed_acquires += 1;
        }
    }
```
With the following results:

| Metric | code in main | with refill changes, Instant::now | governor::DefaultClock | governor::QuantaUpkeepClock updating every 1ms |                                                                                                                                                       
  |--------|------|---------------------|--------------|-------------------|                                                                                                                                                       
  | Min    | 25ns | 49ns                | 31ns         | 25ns              |                                                                                                                                                       
  | Avg    | 51ns | 51ns                | 32ns         | 27ns              |                                                                                                                                                       
  | P50    | 51ns | 50ns                | 32ns         | 27ns              |                                                                                                                                                       
  | P90    | 52ns | 51ns                | 33ns         | 28ns              |                                                                                                                                                       
  | P99    | 52ns | 52ns                | 34ns         | 28ns              |                                                                                                                                                       
  | P99.9  | 68ns | 72ns                | 53ns         | 42ns              |                                                                                                                                                       
  | Max    | 45.04µs | 147.62µs         | 94.50µs      | 26.45µs           |  

Seeing as the benchmark itself used Instant::now, this is probably not scientific, but at these latencies this doesn't really matter.

I went with the default governor clock seeing as an OS thread updating every millisecond is overkill, and the speed of this is not going to contribute to any meaningful bottlenecks enough to optimize this further.

### DIFF
--- a/quickwit/quickwit-common/src/rate_limiter.rs
+++ b/quickwit/quickwit-common/src/rate_limiter.rs
@@ -12,9 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::time::{Duration, Instant};
+use std::ops::Add;
+use std::time::Duration;
 
 use bytesize::ByteSize;
+use governor::clock::{Clock, DefaultClock, Reference};
+use governor::nanos::Nanos;
 
 use crate::tower::{ConstantRate, Rate};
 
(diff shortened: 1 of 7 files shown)


quickwit/quickwit-common/src/rate_limiter.rs
```diff
@@ -148,20 +140,22 @@ impl RateLimiter {
...
     }
 
-    fn refill(&mut self, now: Instant) {
-        if now < self.refill_at {
+    fn refill(&mut self, now: C::Instant) {
+        if now.lt(&self.refill_at) {
             return;
         }
-        let elapsed = (now - self.refill_at).as_micros() as u64;
+        let elapsed = now.duration_since(self.refill_at).as_u64();
```
guilload:
```suggestion
        let elapsed_nanos = now.duration_since(self.refill_at).as_u64();
```
Before it was clear those were micros, but that's not the case anymore.

quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -83,10 +83,13 @@ impl InnerIngesterState {
             .filter(|shard| {
                 shard.is_open() && shard.index_uid == *index_uid && shard.source_id == *source_id
             })
-            .max_by(|left, right| {
-                left.rate_limiter
-                    .available_permits()
-                    .cmp(&right.rate_limiter.available_permits())
+            .reduce(|best, current| {
```
guilload, line 88: nit: let's leave a comment explaining why we use `reduce` here.

PR #6137, opened 2026-02-02
Add shard-level ingestion metrics

nadav-govari:
### Description
Emit shard level metrics on ingestion rates, per index/source/shard. We can then derive per-index or per-source metrics as well.

The total cardinality here is actually just the number of shards, since each shard is uniquely tied to an index/source and the IDs are ULIDs. And since nodes re-use the same IDs (indexer-2, indexer-3, etc), the cardinality shouldn't ever be much higher than a little bit more than the max number of open shards.

Note that this is not accurate - it is **the view that the control plane has**, which is what goes into scaling decisions.

Also added the aggregate to the debug info for better visualizations.

### DIFF
--- a/quickwit/quickwit-control-plane/src/metrics.rs
+++ b/quickwit/quickwit-control-plane/src/metrics.rs
@@ -44,8 +44,8 @@ pub struct ControlPlaneMetrics {
     pub remote_shards: IntGauge,
 
     // Per-shard ingestion rate metrics.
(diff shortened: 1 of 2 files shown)


esatterwhite:
I think you need a way to disable this. We have many tens of thousands of indexes that ingest data into. This destroy our metrics cardinality limits

nadav-govari:
Yeah, even with one index, shards open and close too frequently. I'm just closing this, there are more effective ways of getting these metrics.

PR #6142, opened 2026-02-04
feat(obs): export internal logs with otlp exporter

Totodore:
### Description
Currently quickwit only export its traces, this change also export internal logs if the `QW_ENABLE_OPENTELEMETRY_OTLP_EXPORTER` is set to true.

This is particularly important in a production env when we want to have structured logs to monitor quickwit and find root cause for quickwit internal errors.

### How was this PR tested?
* Tested in local with dog fooding, the instance eat its logs and put it in `otel-logs-v0_9` as expected.

### Side note
We may want to put another env config to disable this behavior by default.
We may also want to configure custom batching behavior according to the log volume (something that I'm not particularly aware of).

### DIFF
--- a/quickwit/quickwit-lambda-client/src/invoker.rs
+++ b/quickwit/quickwit-lambda-client/src/invoker.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::time::Duration;
+
 use anyhow::Context as _;
 use async_trait::async_trait;
 use aws_sdk_lambda::Client as LambdaClient;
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-cli/src/main.rs
```diff
@@ -113,8 +113,11 @@ async fn main_impl() -> anyhow::Result<()> {
...
 
-    if let Some(provider) = tracer_provider_opt {
-        provider
+    if let Some((trace_provider, logs_provider)) = tracer_provider_opt {
+        trace_provider
+            .shutdown()
+            .context("failed to shutdown OpenTelemetry tracer provider")?;
+        logs_provider
             .shutdown()
             .context("failed to shutdown OpenTelemetry tracer provider")?;
```
ncoiffier-celonis: Minor typo: `failed to shutdown OpenTelemetry tracer provider` -> `failed to shutdown OpenTelemetry log provider`

PR #6145, opened 2026-02-06
feat(search): add skip_aggregation_finalization to SearchRequest

congx4:
## Summary

Add an optional `skip_aggregation_finalization` field to `SearchRequest` that, when set to `true`, skips the finalization of aggregation results in `root.rs` and returns the raw `IntermediateAggregationResults` bytes instead.

This enables callers (e.g., CloudPrem/Pomsky) to handle intermediate aggregation results directly for multi-step query merging.

## Changes

- **`search.proto`**: Added `optional bool skip_aggregation_finalization = 19` to `SearchRequest`.
- **`root.rs`**: Modified `finalize_aggregation_if_any()` to return intermediate bytes directly when the flag is `true`.
- **`rest_handler.rs`** (search API + elasticsearch API): Set `skip_aggregation_finalization: None` in `SearchRequest` construction.


quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -247,6 +247,10 @@ message SearchRequest {
   // When an exact index ID is provided (not a pattern), the query fails only if
   // that index is not found and this parameter is set to `false`.
   bool ignore_missing_indexes = 18;
+
+  // When true, skip finalization of aggregation results and return
+  // the raw IntermediateAggregationResults bytes instead.
+  optional bool skip_aggregation_finalization = 19;
```
guilload: I think you can remove the optional, because the default boolean value is `false` in protobuf.

PR #6151, opened 2026-02-11
feat: Make elastic TermsQuery use TermSetQuery internally

Darkheir:
### Description

Rely on TermSetQuery instead of a bool should query with many criteria on the same field as it should improve performances.

### How was this PR tested?

Integration tests should still pass


quickwit/quickwit-query/src/query_ast/term_set_query.rs
```diff
@@ -78,3 +127,67 @@ impl From<TermSetQuery> for QueryAst {
...
+
+    use tantivy::schema::{FAST, INDEXED, Schema};
+
+    use super::TermSetQuery;
+    use crate::query_ast::{BuildTantivyAst, BuildTantivyAstContext};
+
+    #[test]
+    fn test_term_set_query_with_fast_only_field_returns_bool_query() {
+        let mut schema_builder = Schema::builder();
+        schema_builder.add_u64_field("fast_field", FAST);
```
trinity-1686a, line 143: if we keep things that way, i think a test with mixed "one fast field and one indexed field" would be nice to have
given this is a workaround and we may update tantivy so it's no longer needed, it's not really needed though

PR #6152, opened 2026-02-11
Implement IngesterAffinity broadcast

nadav-govari:
### Background
Main idea: https://docs.google.com/document/d/1XUpBdMFnuX8d23erK-XwQkomRgbeRTJ0TJtve7RGW3k/edit?tab=t.0. 

All work on this feature will be merged PR by PR into the base branch **nadav/feature-node-based-routing**, which will then eventually be merged into main once it's fully ready.

### PR Description
Creates a new broadcast to prepare for node based routing. The idea is described more in depth in 

The primary thinking here is:
* Ingester affinity score for receiving new requests. This will be used in a weighted power of two choices comparison against other nodes. The node with the higher affinity score wins and receives the request for persistence.
* The number of open shards for the individual index can act as a tiebreaker.
  *  This isn't perfect but we can iterate on it
Ingesters will move away from keeping shard level data, and instead keep this node level data for routing requests. Routing tables will move to be node based and use the data from these broadcasts to update their routing tables.

### DIFF
--- a/quickwit/quickwit-common/src/ring_buffer.rs
+++ b/quickwit/quickwit-common/src/ring_buffer.rs
@@ -14,13 +14,23 @@
 
 use std::fmt::{Debug, Formatter};
 
-/// A fixed-capacity circular buffer that overwrites the oldest element when full.
+/// Fixed-size buffer that keeps the last N elements pushed into it.
 ///
-/// Elements are stored in a flat array of size `N` and rotated on each push.
-/// The newest element is always at position `N - 1` (the last slot), and the
-/// oldest is at position `N - len`.
+/// `head` is the write cursor. It advances by one on each push and wraps
+/// back to 0 when it reaches N, overwriting the oldest element.
+///
+/// ```text
+/// RingBuffer<u32, 4> after pushing 1, 2, 3, 4, 5, 6:
+///
+///   buffer = [5, 6, 3, 4]    head = 2    len = 4
+///                 ^
+///                 next write goes here
+///
+///   logical view (oldest → newest): [3, 4, 5, 6]
+/// ```
 pub struct RingBuffer<T: Copy + Default, const N: usize> {
     buffer: [T; N],
+    head: usize,
     len: usize,
 }
 
@@ -28,6 +38,7 @@ impl<T: Copy + Default, const N: usize> Default for RingBuffer<T, N> {
     fn default() -> Self {
         Self {
             buffer: [T::default(); N],
+            head: 0,
             len: 0,
         }
     }
@@ -40,11 +51,11 @@ impl<T: Copy + Default + Debug, const N: usize> Debug for RingBuffer<T, N> {
 }
 
 impl<T: Copy + Default, const N: usize> RingBuffer<T, N> {
-    pub fn push(&mut self, value: T) {
-        self.len = (self.len + 1).min(N);
-        self.buffer.rotate_left(1);
-        if let Some(last) = self.buffer.last_mut() {
-            *last = value;
+    pub fn push_back(&mut self, value: T) {
+        self.buffer[self.head] = value;
+        self.head = (self.head + 1) % N;
+        if self.len < N {
+            self.len += 1;
         }
     }
 
@@ -52,14 +63,14 @@ impl<T: Copy + Default, const N: usize> RingBuffer<T, N> {
         if self.len == 0 {
             return None;
         }
-        self.buffer.last().copied()
+        Some(self.buffer[(self.head + N - 1) % N])
     }
 
-    pub fn oldest(&self) -> Option<T> {
+    pub fn front(&self) -> Option<T> {
         if self.len == 0 {
             return None;
         }
-        Some(self.buffer[N - self.len])
+        Some(self.buffer[(self.head + N - self.len) % N])
     }
 
     pub fn len(&self) -> usize {
@@ -70,9 +81,9 @@ impl<T: Copy + Default, const N: usize> RingBuffer<T, N> {
         self.len == 0
     }
 
-    /// Iterates from oldest to newest over the recorded elements.
-    pub fn iter(&self) -> impl Iterator<Item = &T> {
-        self.buffer[N - self.len..].iter()
+    pub fn iter(&self) -> impl Iterator<Item = &T> + '_ {
+        let start = (self.head + N - self.len) % N;
+        (0..self.len).map(move |i| &self.buffer[(start + i) % N])
     }
 }
 
@@ -86,30 +97,30 @@ mod tests {
         assert!(rb.is_empty());
         assert_eq!(rb.len(), 0);
         assert_eq!(rb.last(), None);
-        assert_eq!(rb.oldest(), None);
+        assert_eq!(rb.front(), None);
         assert_eq!(rb.iter().count(), 0);
     }
 
     #[test]
     fn test_single_push() {
         let mut rb = RingBuffer::<u32, 4>::default();
-        rb.push(10);
+        rb.push_back(10);
         assert_eq!(rb.len(), 1);
         assert!(!rb.is_empty());
         assert_eq!(rb.last(), Some(10));
-        assert_eq!(rb.oldest(), Some(10));
+        assert_eq!(rb.front(), Some(10));
         assert_eq!(rb.iter().copied().collect::<Vec<_>>(), vec![10]);
     }
 
     #[test]
     fn test_partial_fill() {
         let mut rb = RingBuffer::<u32, 4>::default();
-        rb.push(1);
-        rb.push(2);
-        rb.push(3);
+        rb.push_back(1);
+        rb.push_back(2);
+        rb.push_back(3);
         assert_eq!(rb.len(), 3);
         assert_eq!(rb.last(), Some(3));
-        assert_eq!(rb.oldest(), Some(1));
+        assert_eq!(rb.front(), Some(1));
         assert_eq!(rb.iter().copied().collect::<Vec<_>>(), vec![1, 2, 3]);
     }
 
@@ -117,11 +128,11 @@ mod tests {
     fn test_exactly_full() {
         let mut rb = RingBuffer::<u32, 4>::default();
         for i in 1..=4 {
-            rb.push(i);
+            rb.push_back(i);
         }
         assert_eq!(rb.len(), 4);
         assert_eq!(rb.last(), Some(4));
-        assert_eq!(rb.oldest(), Some(1));
+        assert_eq!(rb.front(), Some(1));
         assert_eq!(rb.iter().copied().collect::<Vec<_>>(), vec![1, 2, 3, 4]);
     }
 
@@ -129,12 +140,11 @@ mod tests {
     fn test_wrap_around() {
         let mut rb = RingBuffer::<u32, 4>::default();
         for i in 1..=6 {
-            rb.push(i);
+            rb.push_back(i);
         }
-        // Buffer should contain [3, 4, 5, 6], oldest overwritten.
         assert_eq!(rb.len(), 4);
         assert_eq!(rb.last(), Some(6));
-        assert_eq!(rb.oldest(), Some(3));
+        assert_eq!(rb.front(), Some(3));
         assert_eq!(rb.iter().copied().collect::<Vec<_>>(), vec![3, 4, 5, 6]);
     }
 
@@ -142,19 +152,19 @@ mod tests {
     fn test_many_wraps() {
         let mut rb = RingBuffer::<u32, 3>::default();
         for i in 1..=100 {
-            rb.push(i);
+            rb.push_back(i);
         }
         assert_eq!(rb.len(), 3);
         assert_eq!(rb.last(), Some(100));
-        assert_eq!(rb.oldest(), Some(98));
+        assert_eq!(rb.front(), Some(98));
         assert_eq!(rb.iter().copied().collect::<Vec<_>>(), vec![98, 99, 100]);
     }
 
     #[test]
     fn test_debug() {
         let mut rb = RingBuffer::<u32, 3>::default();
-        rb.push(1);
-        rb.push(2);
+        rb.push_back(1);
+        rb.push_back(2);
         assert_eq!(format!("{:?}", rb), "[1, 2]");
     }
 }
--- a/quickwit/quickwit-common/src/shared_consts.rs
+++ b/quickwit/quickwit-common/src/shared_consts.rs
@@ -65,7 +65,7 @@ pub const SCROLL_BATCH_LEN: usize = 1_000;
 pub const INGESTER_PRIMARY_SHARDS_PREFIX: &str = "ingester.primary_shards:";
 
 /// Prefix used in chitchat to broadcast per-source ingester capacity scores and open shard counts.
-pub const INGESTER_CAPACITY_PREFIX: &str = "ingester.capacity:";
+pub const INGESTER_CAPACITY_SCORE_PREFIX: &str = "ingester.capacity_score:";
 
 /// File name for the encoded list of fields in the split
 pub const SPLIT_FIELDS_FILE_NAME: &str = "split_fields";
--- a/quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_capacity_score.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_capacity_score.rs
@@ -19,7 +19,7 @@ use bytesize::ByteSize;
 use quickwit_cluster::{Cluster, ListenerHandle};
 use quickwit_common::pubsub::{Event, EventBroker};
 use quickwit_common::ring_buffer::RingBuffer;
-use quickwit_common::shared_consts::INGESTER_CAPACITY_PREFIX;
+use quickwit_common::shared_consts::INGESTER_CAPACITY_SCORE_PREFIX;
 use quickwit_proto::ingest::ingester::IngesterStatus;
 use quickwit_proto::types::{IndexUid, NodeId, SourceId, SourceUid};
 use serde::{Deserialize, Serialize};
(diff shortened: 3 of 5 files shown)


quickwit/quickwit-common/src/shared_consts.rs
```diff
@@ -64,6 +64,9 @@ pub const SCROLL_BATCH_LEN: usize = 1_000;
 /// Prefix used in chitchat to broadcast the list of primary shards hosted by a leader.
 pub const INGESTER_PRIMARY_SHARDS_PREFIX: &str = "ingester.primary_shards:";
 
+/// Key used in chitchat to broadcast the ingester affinity score and open shard counts.
+pub const INGESTER_AFFINITY_PREFIX: &str = "ingester.affinity";
```
guilload: We already use the word `affinity` for searchers split affinity. I think we can find another ok name for this metric that we don't use already.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
```diff
@@ -0,0 +1,433 @@
...
+use serde::{Deserialize, Serialize};
+use tokio::task::JoinHandle;
+use tracing::{error, warn};
+
+use super::BROADCAST_INTERVAL_PERIOD;
+use crate::ingest_v2::state::WeakIngesterState;
+
+pub type OpenShardCounts = Vec<(IndexUid, SourceId, usize)>;
+
+const WAL_CAPACITY_LOOKBACK_WINDOW_LEN: usize = 6;
```
guilload: This could use a comment. I assume you had a duration in mind for that window and then divided by `BROADCAST_INTERVAL_PERIOD` to get to 6. What's that window duration?

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
```diff
@@ -0,0 +1,433 @@
...
+
+struct WalCapacityTimeSeries {
+    wal_capacity: ByteSize,
+    readings: VecDeque<ByteSize>,
```
guilload: There's a better implementation of a timeserie based on a rotating time window in `broadcast`. This is a common pattern. So, move the og implementation in `common`. Abstractify enough so it can be used for both uses cases, import and use it here.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
```diff
@@ -0,0 +1,433 @@
...
+}
+
+impl WalCapacityTimeSeries {
+    fn new(wal_capacity: ByteSize) -> Self {
```
guilload: mem or disk? the name should say it.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
```diff
@@ -0,0 +1,433 @@
...
+
+    /// How much remaining capacity changed between the oldest and newest readings.
+    /// Positive = improving, negative = draining.
+    /// At most `WAL_CAPACITY_LOOKBACK_WINDOW_LEN` readings are kept.
+    fn delta(&mut self) -> Option<f64> {
+        if self.readings.is_empty() {
+            return None;
+        }
+        let oldest = if self.readings.len() > WAL_CAPACITY_LOOKBACK_WINDOW_LEN {
+            self.readings.pop_back().unwrap()
```
guilload: Use `expect` and state the invariant/conditation that allow you to call `expect` safely:
`.expect("window should not be empty")`
`.expect("window should have more than 1 measurements")`

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
```diff
@@ -0,0 +1,433 @@
...
+        } else {
+            *self.readings.back().unwrap()
+        };
+        let current = *self.readings.front().unwrap();
+        Some(self.as_capacity_usage_pct(current) - self.as_capacity_usage_pct(oldest))
+    }
+}
+
+/// Computes an affinity score from 0 to 100 using a simple PI controller.
```
guilload: I'd rather have a score between 0-10. I don't think 11 vs 19 makes much difference and it will allow us to gossip less. That's why gossiped throughput are rounded for instance. We can keep the score between 0-100 here and perform the division at gossip time if you want. Comment as to why we divide.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
```diff
@@ -0,0 +1,433 @@
...
+    }
+    let p = 80.0 * remaining_capacity;
+    let drain = (-capacity_delta).clamp(0.0, 0.10);
+    let i = 20.0 * (1.0 - drain / 0.10);
+    (p + i).clamp(0.0, 100.0) as u32
+}
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct IngesterAffinity {
+    pub affinity_score: u32,
```
guilload: If we're playing this game then `u16`. `u8` also works but readers are going to interpret it as a raw byte not an int up to 255.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
```diff
@@ -0,0 +1,433 @@
...
+            *self.readings.back().unwrap()
+        };
+        let current = *self.readings.front().unwrap();
+        Some(self.as_capacity_usage_pct(current) - self.as_capacity_usage_pct(oldest))
+    }
+}
+
+/// Computes an affinity score from 0 to 100 using a simple PI controller.
+///
+/// The score has two components:
```
guilload: I think you actually implemented a PD Controller (Proportional-Derivative)

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
```diff
@@ -0,0 +1,433 @@
...
+        };
+        tokio::spawn(async move { broadcaster.run().await })
+    }
+
+    async fn snapshot(&self) -> Result<(ByteSize, OpenShardCounts)> {
+        let state = self
+            .weak_state
+            .upgrade()
+            .context("ingester state has been dropped")?;
+
```
guilload: Just lock the whole thing fully and make the code more readable.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
```diff
@@ -0,0 +1,433 @@
...
+        let mut interval = tokio::time::interval(BROADCAST_INTERVAL_PERIOD);
+
+        loop {
+            interval.tick().await;
+
+            let (wal_used, open_shard_counts) = match self.snapshot().await {
+                Ok(snapshot) => snapshot,
+                Err(error) => {
+                    error!("failed to snapshot ingester state: {error}");
+                    return;
```
guilload: The WAL can take multiple `BROADCAST_INTERVAL_PERIOD` intervals to load. The task should not not stop when we're loading the WAL, only if the state is dropped.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
```diff
@@ -0,0 +1,433 @@
...
+
+            self.broadcast_affinity(&affinity).await;
+        }
+    }
+
+    async fn broadcast_affinity(&self, affinity: &IngesterAffinity) {
+        let value = serde_json::to_string(affinity)
+            .expect("`IngesterAffinity` should be JSON serializable");
+        self.cluster
+            .set_self_key_value(INGESTER_AFFINITY_PREFIX, value)
```
guilload: You can't broadcast that over a single key because the open shard counts can be very long.
-> one key per index/source

quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -87,6 +88,17 @@ impl InnerIngesterState {
             .max_by_key(|(available_permits, _)| *available_permits)
             .map(|(_, shard)| shard)
     }
+
+    pub fn get_open_shard_counts(&self) -> Vec<(IndexUid, SourceId, usize)> {
+        self.shards
+            .values()
+            .filter(|shard| shard.is_open())
```
guilload:
```suggestion
            .filter(|shard| shard.is_advertisable && !shard.is_replica() && shard.is_open())
```

quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -87,6 +88,17 @@ impl InnerIngesterState {
             .max_by_key(|(available_permits, _)| *available_permits)
             .map(|(_, shard)| shard)
     }
+
+    pub fn get_open_shard_counts(&self) -> Vec<(IndexUid, SourceId, usize)> {
```
guilload, line 92: No unit test?

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
guilload: (The value length is an issue because chitchat uses UDP and every update must fit in a single datagram (MTU))

quickwit/quickwit-common/src/shared_consts.rs
nadav-govari: Yep, how's ingester `capacity`? As in, literally the capacity of the ingester to ingest new requests.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
nadav-govari: Adding. It was meant to be 30 seconds.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
nadav-govari: We can just make it usize.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
nadav-govari: Updated to the following cases:
1. State dropped: error, stop task
2. Ingester not initialized: no-op
3. Ingester ready: happy path

quickwit/quickwit-common/src/shared_consts.rs
nadav-govari: Renamed the task to BroadcastIngesterCapacity and all references from affinity to capacity.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
nadav-govari: LocalShardUpdate and BroadcastIngesterCapacity now both use this new RingBuffer, which is in quickwit-common.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
nadav-govari: Disk, modified.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
nadav-govari: Noted, though this isn't relevant any longer with the RingBuffer changes.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
nadav-govari: Added the comment. Spoke offline about 1-100 vs 1-10. Conclusion: chitchat will only advertise changes, and we can prevent a lot of re-broadcasts at the cost of an order of magnitude of granularity. We might revisit this in the future when trying to find places to extract more performance from at the cost of complexity.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
nadav-govari: You're right. I confused the terms but I think PD accomplishes what we're looking for here, imperfectly. My first order of priority was keeping this as simple as possible to start with. The second was that rather than trying to normalize to some desired state, we want to keep a balance across all nodes, and penalize nodes that get too hot too quickly. The good thing is we can iterate on this easily as we observe the change in performance.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
nadav-govari: Done.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_affinity.rs
nadav-govari: Made it similar to LocalShardsUpdate, one key per index/source.

quickwit/quickwit-ingest/src/ingest_v2/state.rs
nadav-govari: Took it.

quickwit/quickwit-ingest/src/ingest_v2/state.rs
nadav-govari: Added.

quickwit/quickwit-common/src/ring_buffer.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    pub fn last(&self) -> Option<T> {
+        if self.len == 0 {
+            return None;
+        }
+        self.buffer.last().copied()
+    }
+
+    pub fn oldest(&self) -> Option<T> {
```
guilload:
```suggestion
    pub fn front(&self) -> Option<T> {
```

quickwit/quickwit-common/src/ring_buffer.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+impl<T: Copy + Default + Debug, const N: usize> Debug for RingBuffer<T, N> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_list().entries(self.iter()).finish()
+    }
+}
+
+impl<T: Copy + Default, const N: usize> RingBuffer<T, N> {
+    pub fn push(&mut self, value: T) {
```
guilload:
```suggestion
    pub fn push_back(&mut self, value: T) {
```
Let's just copy (half of) the `VecDeque` API.

quickwit/quickwit-common/src/ring_buffer.rs
```diff
@@ -0,0 +1,160 @@
...
+// limitations under the License.
+
+use std::fmt::{Debug, Formatter};
+
+/// A fixed-capacity circular buffer that overwrites the oldest element when full.
+///
+/// Elements are stored in a flat array of size `N` and rotated on each push.
+/// The newest element is always at position `N - 1` (the last slot), and the
+/// oldest is at position `N - len`.
+pub struct RingBuffer<T: Copy + Default, const N: usize> {
```
guilload, line 31: Noice

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_capacity.rs
```diff
@@ -0,0 +1,457 @@
...
+/// with the one that is exactly `WAL_CAPACITY_LOOKBACK_WINDOW_LEN` steps ago. Otherwise, that
+/// reading would be discarded when the next reading is inserted.
+const WAL_CAPACITY_READINGS_LEN: usize = WAL_CAPACITY_LOOKBACK_WINDOW_LEN + 1;
+
+struct WalDiskCapacityTimeSeries {
+    wal_capacity: ByteSize,
+    readings: RingBuffer<ByteSize, WAL_CAPACITY_READINGS_LEN>,
+}
+
+impl WalDiskCapacityTimeSeries {
```
guilload: I thought we discussed using memory?

quickwit/quickwit-common/src/ring_buffer.rs
guilload, line 31: Claude can easily make `push` _O(1)_, right?

quickwit/quickwit-common/src/shared_consts.rs
```diff
@@ -64,6 +64,9 @@ pub const SCROLL_BATCH_LEN: usize = 1_000;
 /// Prefix used in chitchat to broadcast the list of primary shards hosted by a leader.
 pub const INGESTER_PRIMARY_SHARDS_PREFIX: &str = "ingester.primary_shards:";
 
+/// Prefix used in chitchat to broadcast per-source ingester capacity scores and open shard counts.
+pub const INGESTER_CAPACITY_PREFIX: &str = "ingester.capacity:";
```
guilload:
```suggestion
pub const INGESTER_CAPACITY_SCORE_PREFIX: &str = "ingester.capacity_score:";
```

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_capacity_score.rs
```diff
@@ -0,0 +1,457 @@
+// Copyright 2021-Present Datadog, Inc.
```
guilload, line 1: Let's use `capacity_score` everywhere.

quickwit/quickwit-ingest/src/ingest_v2/broadcast/local_shards.rs
```diff
@@ -152,7 +148,7 @@ impl LocalShardsSnapshot {
 
 /// Takes a snapshot of the primary shards hosted by the ingester at regular intervals and
 /// broadcasts it to other nodes via Chitchat.
-pub(super) struct BroadcastLocalShardsTask {
+pub struct BroadcastLocalShardsTask {
```
guilload, line 151:
```suggestion
pub(crate) struct BroadcastLocalShardsTask {
```

quickwit/quickwit-common/src/ring_buffer.rs
nadav-govari: Done

quickwit/quickwit-common/src/ring_buffer.rs
nadav-govari: Done

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_capacity.rs
nadav-govari: Yeah, now that I realize they're capped in the chart, I think they're functionally the same, but memory feels like a cleaner number to read. So I switched it to memory.

quickwit/quickwit-common/src/ring_buffer.rs
nadav-govari: Yes it can :)

quickwit/quickwit-common/src/shared_consts.rs
nadav-govari: Done

quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_capacity_score.rs
nadav-govari: Done

PR #6157, opened 2026-02-12
Offload leaf search work to AWS Lambda functions

fulmicoton-dd:
## Summary

The goal is to handle traffic spikes gracefully without provisioning additional searcher nodes: when the local search queue is saturated, overflow splits are transparently routed to Lambda for processing.

### How offloading works

The offloading decision happens **on the leaf side**, inside the `SearchPermitProvider`. The permit provider already manages a bounded queue of pending split search tasks (gated by memory budget and download slots). When a leaf search request arrives, the provider checks the current queue depth against a configurable `offload_threshold`. If granting permits for all requested splits would exceed this threshold, only enough splits to fill up to the threshold are processed locally — the rest are marked for offloading.

The offloaded splits are batched (up to `max_splits_per_invocation` splits per batch, balanced by document count) and sent to Lambda in parallel. Each Lambda invocation runs the same leaf search code path and **returns per-split results individually**. This is important: the per-split responses are fed back into the `IncrementalCollector` and populate the **partial result cache**, so subsequent queries hitting the same splits benefit from cached results regardless of whether the split was searched locally or on Lambda.

### Auto-deployment

Depending on the configuration, the Lambda function code can be **deployed automatically** at startup. The `quickwit-lambda-client` crate embeds a compressed Lambda binary at compile time. When `auto_deploy` is configured, Quickwit will:
1. Check if a published Lambda version matching the current binary already exists (identified by a description tag `quickwit:{version}-{hash}`)
2. Create or update the function and publish a new version if needed
3. Garbage-collect old versions (keeping the current one + 5 most recent)

This ensures the Lambda function always matches the running Quickwit version without any external deployment tooling. Manual deployment is also supported for users who prefer to manage Lambda functions through Terraform or other IaC tools.

### Configuration

Lambda offloading is opt-in. Add a `lambda` section under `searcher` in the node configuration:

```yaml
searcher:
  lambda:
    offload_threshold: 100     # queue depth before offloading kicks in (0 = always offload)
    max_splits_per_invocation: 10
    auto_deploy:
      execution_role_arn: arn:aws:iam::123456789012:role/quickwit-lambda-role
      memory_size: 5 GiB
      invocation_timeout_secs: 15
```

### New crates

- **`quickwit-lambda-client`**: Handles Lambda invocation (with metrics) and auto-deployment logic. Embeds the Lambda binary at build time.
- **`quickwit-lambda-server`**: The Lambda function handler itself — receives a `LeafSearchRequest`, runs `multi_index_leaf_search`, and returns per-split `LeafSearchResponse`s.

### Key changes in existing crates

- **`quickwit-search`**: New `LambdaLeafSearchInvoker` trait; `SearchPermitProvider` gains `get_permits_with_offload` to split work between local and offloaded; `leaf.rs` orchestrates local and Lambda tasks in parallel.
- **`quickwit-config`**: New `LambdaConfig` and `LambdaDeployConfig` structs under `SearcherConfig`.
- **`quickwit-serve`**: Initializes the Lambda invoker at startup when configured.
- **`quickwit-proto`**: New `LeafSearchResponses` wrapper message for batched per-split responses.

### DIFF
--- a/quickwit/quickwit-query/src/elastic_query_dsl/term_query.rs
+++ b/quickwit/quickwit-query/src/elastic_query_dsl/term_query.rs
@@ -75,6 +75,7 @@ pub struct TermQueryParams {
     case_insensitive: bool,
 }
 
+#[cfg(test)]
 pub fn term_query_from_field_value(field: impl ToString, value: impl ToString) -> TermQuery {
     TermQuery {
         field: field.to_string(),
--- a/quickwit/quickwit-query/src/elastic_query_dsl/terms_query.rs
+++ b/quickwit/quickwit-query/src/elastic_query_dsl/terms_query.rs
@@ -12,14 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::{BTreeSet, HashMap};
+
 use serde::Deserialize;
 
-use crate::elastic_query_dsl::bool_query::BoolQuery;
 use crate::elastic_query_dsl::one_field_map::OneFieldMap;
-use crate::elastic_query_dsl::term_query::term_query_from_field_value;
 use crate::elastic_query_dsl::{ConvertibleToQueryAst, ElasticQueryDslInner};
 use crate::not_nan_f32::NotNaNf32;
-use crate::query_ast::QueryAst;
+use crate::query_ast::{QueryAst, TermSetQuery};
 
 #[derive(PartialEq, Eq, Debug, Deserialize, Clone)]
 #[serde(try_from = "TermsQueryForSerialization")]
@@ -87,15 +87,14 @@ impl TryFrom<TermsQueryForSerialization> for TermsQuery {
 
 impl ConvertibleToQueryAst for TermsQuery {
     fn convert_to_query_ast(self) -> anyhow::Result<QueryAst> {
-        let term_queries: Vec<ElasticQueryDslInner> = self
-            .values
-            .into_iter()
-            .map(|value| term_query_from_field_value(self.field.clone(), value))
-            .map(ElasticQueryDslInner::from)
-            .collect();
-        let mut union = BoolQuery::union(term_queries);
-        union.boost = self.boost;
-        union.convert_to_query_ast()
+        let mut terms_per_field = HashMap::new();
+        let values_set: BTreeSet<String> = self.values.into_iter().collect();
+        terms_per_field.insert(self.field, values_set);
+
+        let term_set_query = TermSetQuery { terms_per_field };
+        let query_ast: QueryAst = term_set_query.into();
+
+        Ok(query_ast.boost(self.boost))
     }
 }
 
--- a/quickwit/quickwit-query/src/query_ast/term_set_query.rs
+++ b/quickwit/quickwit-query/src/query_ast/term_set_query.rs
@@ -19,7 +19,7 @@ use tantivy::Term;
 
 use crate::InvalidQuery;
 use crate::query_ast::{
-    BuildTantivyAst, BuildTantivyAstContext, QueryAst, TantivyQueryAst, TermQuery,
+    BoolQuery, BuildTantivyAst, BuildTantivyAstContext, QueryAst, TantivyQueryAst, TermQuery,
 };
 
 /// TermSetQuery matches the same document set as if it was a union of
@@ -32,6 +32,53 @@ pub struct TermSetQuery {
 }
 
 impl TermSetQuery {
+    fn has_fast_only_field(&self, context: &BuildTantivyAstContext) -> bool {
+        for full_path in self.terms_per_field.keys() {
+            if let Some((_, field_entry, _)) =
+                super::utils::find_field_or_hit_dynamic(full_path, context.schema)
+                && field_entry.is_fast()
+                && !field_entry.is_indexed()
+            {
+                return true;
+            }
+        }
+        false
+    }
+
+    fn build_bool_query(
+        &self,
+        context: &BuildTantivyAstContext,
+    ) -> Result<TantivyQueryAst, InvalidQuery> {
+        let should_clauses = self
+            .terms_per_field
+            .iter()
+            .flat_map(|(full_path, values)| {
+                values.iter().map(|value| {
+                    QueryAst::Term(TermQuery {
+                        field: full_path.to_string(),
+                        value: value.to_string(),
+                    })
+                })
+            })
+            .collect();
+
+        let bool_query = BoolQuery {
+            should: should_clauses,
+            ..Default::default()
+        };
+
+        bool_query.build_tantivy_ast_impl(context)
+    }
+
+    fn build_term_set_query(
+        &self,
+        context: &BuildTantivyAstContext,
+    ) -> Result<TantivyQueryAst, InvalidQuery> {
+        let terms_it = self.make_term_iterator(context)?;
+        let term_set_query = tantivy::query::TermSetQuery::new(terms_it);
+        Ok(term_set_query.into())
+    }
+
     fn make_term_iterator(
         &self,
         context: &BuildTantivyAstContext,
@@ -67,14 +114,80 @@ impl BuildTantivyAst for TermSetQuery {
         &self,
         context: &BuildTantivyAstContext,
     ) -> Result<TantivyQueryAst, InvalidQuery> {
-        let terms_it = self.make_term_iterator(context)?;
-        let term_set_query = tantivy::query::TermSetQuery::new(terms_it);
-        Ok(term_set_query.into())
+        if self.has_fast_only_field(context) {
+            self.build_bool_query(context)
+        } else {
+            self.build_term_set_query(context)
+        }
     }
 }
 
 impl From<TermSetQuery> for QueryAst {
     fn from(term_set_query: TermSetQuery) -> Self {
         QueryAst::TermSet(term_set_query)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::{BTreeSet, HashMap};
+
+    use tantivy::schema::{FAST, INDEXED, Schema};
+
+    use super::TermSetQuery;
+    use crate::query_ast::{BuildTantivyAst, BuildTantivyAstContext};
+
+    #[test]
+    fn test_term_set_query_with_fast_only_field_returns_bool_query() {
+        let mut schema_builder = Schema::builder();
+        schema_builder.add_u64_field("fast_field", FAST);
+        let schema = schema_builder.build();
+
+        let terms_per_field = HashMap::from([(
+            "fast_field".to_string(),
+            BTreeSet::from(["1".to_string(), "2".to_string()]),
+        )]);
+        let term_set_query = TermSetQuery { terms_per_field };
+
+        let tantivy_query_ast = term_set_query
+            .build_tantivy_ast_call(&BuildTantivyAstContext::for_test(&schema))
+            .unwrap();
+
+        let bool_query = tantivy_query_ast
+            .as_bool_query()
+            .expect("Expected BoolQuery for fast-only field, but got a different query type");
+        assert_eq!(bool_query.should.len(), 2);
+        assert_eq!(bool_query.must.len(), 0);
+        assert_eq!(bool_query.must_not.len(), 0);
+        assert_eq!(bool_query.filter.len(), 0);
+    }
+
+    #[test]
+    fn test_term_set_query_with_indexed_field_uses_term_set() {
+        let mut schema_builder = Schema::builder();
+        schema_builder.add_u64_field("indexed_field", FAST | INDEXED);
+        let schema = schema_builder.build();
+
+        let terms_per_field = HashMap::from([(
+            "indexed_field".to_string(),
+            BTreeSet::from(["1".to_string(), "2".to_string()]),
+        )]);
+        let term_set_query = TermSetQuery { terms_per_field };
+
+        let tantivy_query_ast = term_set_query
+            .build_tantivy_ast_call(&BuildTantivyAstContext::for_test(&schema))
+            .unwrap();
+
+        // Should return a leaf query (TermSetQuery wrapped in TantivyQueryAst)
+        let leaf = tantivy_query_ast
+            .as_leaf()
+            .expect("Expected a leaf query (TermSetQuery), but got a complex query");
+
+        // Verify it's a TermSetQuery by checking the debug representation
+        let debug_str = format!("{leaf:?}");
+        assert!(
+            debug_str.contains("TermSetQuery"),
+            "Expected TermSetQuery, got: {debug_str}"
+        );
     }
 }


docs/configuration/lambda-config.md
```diff
@@ -0,0 +1,217 @@
...
+        }
+      }
+    }
+  ]
+}
+```
+
+### Lambda execution role
+
+The Lambda function requires an execution role with S3 read access to your index data. CloudWatch logging permissions are not required.
```
trinity-1686a: why is CloudWatch mentioned?

quickwit/quickwit-lambda-client/src/invoker.rs
```diff
@@ -0,0 +1,185 @@
...
+        &self,
+        request: LeafSearchRequest,
+    ) -> Result<Vec<LeafSearchResponse>, SearchError> {
+        // Serialize request to protobuf bytes, then base64 encode
+        let request_bytes = request.encode_to_vec();
+        let payload = LeafSearchRequestPayload {
+            payload: BASE64_STANDARD.encode(&request_bytes),
+        };
+
+        let payload_json = serde_json::to_vec(&payload)
```
trinity-1686a, line 119: LeafSearchRequest should implement (De)Serialize, if we're forced to use json, i think direct json is probably cleaner than `json(bae64(protobuf(request)))`

quickwit/quickwit-lambda-client/build.rs
```diff
@@ -0,0 +1,107 @@
...
+//! and places it in OUT_DIR for embedding via include_bytes!
+//!
+//! The Lambda binary is built separately in CI and published as a GitHub release.
+
+use std::env;
+use std::path::PathBuf;
+
+/// URL to download the pre-built Lambda zip from GitHub releases.
+/// This should be updated when a new Lambda binary is released.
+const LAMBDA_ZIP_URL: &str = "https://github.com/quickwit-oss/quickwit/releases/download/lambda-ff6fdfa5/quickwit-aws-lambda--aarch64.zip";
```
trinity-1686a, line 29: as with all ressources downloaded from the internet, i think we should check the hash is the one expected (if someone's account get compromised, we don't want people to be able to bait&switch what lambda is going to eventually get executed)


(ftaod, i don't mean that in the "make sure the lambda client is compatible with the lambda-server and use a hash as a versioning mechanism", but "let's ensure that any ressource that is referenced in tree can be authenticated as being the same as the one intended by the person that added that reference in tree")

quickwit/quickwit-lambda-server/src/handler.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    let num_splits = all_splits.len();
+    info!(num_splits, "processing leaf search request (per-split)");
+
+    // Process each split in parallel. The SearchPermitProvider inside
+    // SearcherContext gates concurrency based on memory budget.
+    let mut split_search_futures: Vec<tokio::task::JoinHandle<_>> =
+        Vec::with_capacity(all_splits.len());
+    for (leaf_req_idx, split) in all_splits {
+        let leaf_request_ref = &leaf_search_request.leaf_requests[leaf_req_idx];
```
trinity-1686a: it took me multiple reading through the code to understand why `all_splits` existed and we needed to index `leaf_search_request.leaf_requests` here.
I think moving the creation of the LeafSearchRequest inside the flap_map that created all_splits would be a lot easier to understand

quickwit/quickwit-lambda-server/src/handler.rs
```diff
@@ -0,0 +1,131 @@
...
+            leaf_requests: vec![quickwit_proto::search::LeafRequestRef {
+                index_uri_ord: leaf_request_ref.index_uri_ord,
+                doc_mapper_ord: leaf_request_ref.doc_mapper_ord,
+                split_offsets: vec![split],
+            }],
+        };
+
+        let searcher_context = ctx.searcher_context.clone();
+        let storage_resolver = ctx.storage_resolver.clone();
+        split_search_futures.push(tokio::task::spawn(multi_index_leaf_search(
```
trinity-1686a: i'm not a fan of using tokio::spawn for this kind of tasks. imo this should be a JoinSet, with some logic to preserve ordering. wdyt?

quickwit/quickwit-search/src/list_terms.rs
```diff
@@ -326,14 +327,19 @@ pub async fn leaf_list_terms(
...
+            compute_initial_memory_allocation(
+                split,
+                searcher_context
+                    .searcher_config
+                    .warmup_single_split_initial_allocation,
+            )
+        })
+        .collect();
+    // allow offload to lambda
+    // https://github.com/quickwit-oss/quickwit/issues/6150
```
trinity-1686a: it's not clear that this is a todo

quickwit/quickwit-search/src/search_permit_provider.rs
```diff
@@ -40,9 +40,16 @@ pub struct SearchPermitProvider {
 }
 
 pub enum SearchPermitMessage {
-    Request {
-        permit_sender: oneshot::Sender<Vec<SearchPermitFuture>>,
+    RequestWithOffload {
         permit_sizes: Vec<u64>,
+        /// Maximum number of pending requests. If granting permits all
```
trinity-1686a:
```suggestion
        /// Maximum number of pending requests. If granting all
```

quickwit/quickwit-lambda-client/src/deploy.rs
```diff
@@ -0,0 +1,972 @@
...
+        let response = match request.send().await {
+            Ok(resp) => resp,
+            Err(SdkError::ServiceError(err)) if err.err().is_resource_not_found_exception() => {
+                info!(
+                    function_name = %function_name,
+                    "lambda function does not exist yet"
+                );
+                return Ok(None);
+            }
+            Err(e) => {
```
trinity-1686a, line 214: nit: there might be transient error we want to retry

quickwit/quickwit-lambda-client/src/deploy.rs
```diff
@@ -0,0 +1,972 @@
...
+    // Fast path: if the function does not exist, we can create and publish the function atomically.
+    if let Some(version) = try_create_function(client, function_name, deploy_config).await? {
+        return Ok(version);
+    }
+
+    // Function already exists — we need to update the code.
+    // This will create or update a version called "$LATEST" (that's the actual string)
+    //
+    // We cannot directly publish here, because updating the function code does not allow
+    // use to pass a different description.
```
trinity-1686a:
```suggestion
    // us to pass a different description.
```

quickwit/quickwit-lambda-client/src/deploy.rs
```diff
@@ -0,0 +1,972 @@
...
+async fn try_create_function(
+    client: &LambdaClient,
+    function_name: &str,
+    deploy_config: &LambdaDeployConfig,
+) -> anyhow::Result<Option<String>> {
+    let memory_size_mb = deploy_config
+        .memory_size
+        .as_u64()
+        .div_ceil(1024u64 * 1024u64) as i32;
+    let timeout_secs = deploy_config.invocation_timeout_secs as i32;
```
trinity-1686a, line 301: does that means that changing these parameters without issueing a new version has no effect? if so, it should be clearly documented. it's odd that quickwit would in a way "remember" its previous configuration when you update it

quickwit/quickwit-lambda-client/src/deploy.rs
```diff
@@ -0,0 +1,972 @@
...
+        );
+
+        if let Err(e) = client
+            .delete_function()
+            .function_name(function_name)
+            .qualifier(&version_str)
+            .send()
+            .await
+        {
+            warn!(
```
trinity-1686a: some errors shouldn't warn ("version doesn't exist" for instance: it's possible it got deleted concurrently by another quickwit)

quickwit/quickwit-lambda-client/src/deploy.rs
```diff
@@ -0,0 +1,972 @@
...
+                "{}-{}",
+                env!("CARGO_PKG_VERSION").replace('.', "_"),
+                env!("LAMBDA_BINARY_HASH")
+            )
+        })
+        .as_str()
+}
+
+/// Returns the version description for our qualifier.
+fn version_description() -> String {
```
trinity-1686a: i think i would prefer this function to only be called early in the codestack, and having its result passed around

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -48,12 +50,64 @@ use tokio::task::{JoinError, JoinSet};
...
+    // Ties are broken by count to help balance item counts when weights are equal.
+    for (weight, item) in weighted_items {
+        let lightest_batch_idx = batch_weights
+            .iter()
+            .zip(batches.iter())
+            .enumerate()
+            .filter(|(_, (_, batch))| batch.len() < max_items_per_batch)
+            .min_by_key(|(_, (batch_weight, batch))| (**batch_weight, batch.len()))
+            .map(|(idx, _)| idx)
+            .unwrap();
```
trinity-1686a: i'm pretty sure this would be easier to reason about if it was a `BinaryHeap<Reverse<(weight, batch)>>` that got `peek_mut`ed, and for which we pop out elements when we add the `max_items_per_batch`th element

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -685,7 +720,8 @@ fn visit_aggregation_mut(
     modified_something
 }
 
-// equivalent to Bound::map, which is unstable
+/// Maps a `Bound<T>` to a `Bound<U>` by applying a function to the contained value.
+/// Equivalent to `Bound::map`, which is currently unstable.
 pub fn map_bound<T, U>(bound: Bound<T>, f: impl FnOnce(T) -> U) -> Bound<U> {
```
trinity-1686a: we could get ride of that, [`Bound::map`](https://doc.rust-lang.org/std/ops/enum.Bound.html#method.map) is stable now

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1352,9 +1389,169 @@ fn disable_search_request_hits(search_request: &mut SearchRequest) {
...
+
+    let batches: Vec<Vec<SplitIdAndFooterOffsets>> = greedy_batch_split(
+        splits,
+        |split| split.num_docs,
+        lambda_config.max_splits_per_invocation,
+    );
+
+    let mut search_request_for_leaf = request.clone();
+    search_request_for_leaf.start_offset = 0;
+    search_request_for_leaf.max_hits += request.start_offset;
```
trinity-1686a: this is already done when LeafSearchRequests are created, this should be a no-op

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1352,9 +1389,169 @@ fn disable_search_request_hits(search_request: &mut SearchRequest) {
...
+
+    let mut lambda_tasks_joinset = JoinSet::new();
+    for batch in batches {
+        let batch_split_ids: Vec<String> =
+            batch.iter().map(|split| split.split_id.clone()).collect();
+        let leaf_request = LeafSearchRequest {
+            search_request: Some(search_request_for_leaf.clone()),
+            doc_mappers: vec![doc_mapper_str.clone()],
+            index_uris: vec![index_uri.as_str().to_string()], //< careful here. Calling to_string() directly would return a redacted uri.
```
trinity-1686a, line 1450: redacted how?!

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1352,9 +1389,169 @@ fn disable_search_request_hits(search_request: &mut SearchRequest) {
...
+    while let Some(join_res) = lambda_tasks_joinset.join_next().await {
+        let Ok((batch_split_ids, result)) = join_res else {
+            error!("lambda join error");
+            return Err(SearchError::Internal("lambda join error".to_string()));
+        };
+        let mut locked = incremental_merge_collector.lock().unwrap();
+        match result {
+            Ok(per_split_responses) => {
+                for response in per_split_responses {
+                    if let Err(err) = locked.add_result(response) {
```
trinity-1686a: i would have expected to see an insertion into the partial request cache somewhere right about here

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1352,9 +1389,169 @@ fn disable_search_request_hits(search_request: &mut SearchRequest) {
...
+                        error!(error = %err, "failed to add lambda result to collector");
```
trinity-1686a: on other code paths, this insert a failed_split instead of logging

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1363,85 +1560,139 @@ pub async fn single_doc_mapping_leaf_search(
...
 
-    let split_filter = CanSplitDoBetter::from_request(&request, doc_mapper.timestamp_field_name());
-    let split_with_req = split_filter.optimize(request.clone(), splits)?;
+    // We simplify the request as much as possible.
+    let split_filter: CanSplitDoBetter =
+        CanSplitDoBetter::from_request(&request, doc_mapper.timestamp_field_name());
+    let mut split_with_req: Vec<(SplitIdAndFooterOffsets, SearchRequest)> =
+        split_filter.optimize(&request, splits)?;
+    for (split, search_request) in &mut split_with_req {
+        rewrite_request(search_request, split, doc_mapper.timestamp_field_name());
```
trinity-1686a: i initially thought this was applying the restriction of all splits on a single search, i think this would make it clearer
```suggestion
    for (split, single_split_search_request) in &mut split_with_req {
        rewrite_request(single_split_search_request, split, doc_mapper.timestamp_field_name());
```

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -435,43 +489,26 @@ fn compute_index_size(hot_directory: &HotDirectory) -> ByteSize {
...
     }
 
-    if let Some(cached_answer) = ctx
-        .searcher_context
-        .leaf_search_cache
-        .get(split.clone(), search_request.clone())
-    {
-        leaf_search_state_guard.set_state(SplitSearchState::CacheHit);
-        return Ok(Some(cached_answer));
-    }
```
trinity-1686a, line 473: i think this could stay: it could happen that we get a cache miss at the top level, wait for a bit while acquiring a permit, and now the value is there

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -1363,85 +1560,139 @@ pub async fn single_doc_mapping_leaf_search(
...
-                aggregations_limits.clone(),
             )
             .in_current_span(),
         );
         task_id_to_split_id_map.insert(handle.id(), split_id);
     }
 
-    // TODO we could cancel running splits when !run_all_splits and the running split can no
-    // longer give better results after some other split answered.
+    // Step 5: Await all local tasks.
```
trinity-1686a: i didn't see step 1-4

trinity-1686a (commented):
i think we should somehow mark this feature as Experimental for the time being (warn on startup when it's enabled maybe?). i suspect there might be changes we'll want to do if/when we add support for other faas provider, in particular on the side of what config looks like.

docs/configuration/lambda-config.md
fulmicoton-dd:
```suggestion
The Lambda function requires an execution role with S3 read access to your index data.
```

fulmicoton-dd: That's out of sync. I'll stick to not talking about it to avoid noise, but if you add more permissions it is possible to get cloudwatch logging for the lambda code.

quickwit/quickwit-lambda-client/src/invoker.rs
fulmicoton-dd: I strongly disagree :).

The reason I serialize protobuf is to make sure we do not increase the API surface.
Adding JSON serialization is not impactless too:
As we discussed, we are not the best enforcer of this, but ideally we would like backward compatibility.
The lambda versioning is not saving us from that: we still receive Request objects that have been generated upstream.

This is something we do not want to enforce on two serialization layers.

docs/configuration/lambda-config.md
trinity-1686a: okay, then maybe there can be a short section saying that .-)

quickwit/quickwit-lambda-client/src/invoker.rs
trinity-1686a: fair.

trinity-1686a: are we forced to go with json? Is direct protobuf (without it being gRPC per say) not an option?

fulmicoton: good point. It has to be http, but I think protobuf directly is ok.

fulmicoton: digging into it, lambda could use any payload, but the serving loop provided by the SDK requires JSON serialization (lambda_runtime::run). We could implement that part, but I'd rather stick to the current impl for the moment.

quickwit/quickwit-lambda-server/src/handler.rs
fulmicoton: I prefer the current implem but I can change that if you prefer.

trinity-1686a:
we're not.

quickwit/quickwit-lambda-client/src/deploy.rs
fulmicoton: Good point!

fulmicoton: I added the config into the description.

quickwit/quickwit-lambda-client/src/deploy.rs
fulmicoton: I changed the code so it is now only called once.

quickwit/quickwit-lambda-client/build.rs
fulmicoton: ok ill do this and switch back to sha256. Second pre-image attack is not easy with md5 but well...

quickwit/quickwit-lambda-client/src/deploy.rs
fulmicoton: The pod will restart

quickwit/quickwit-search/src/leaf.rs
fulmicoton: added cache population. one tricky part is that the key needs to be the simplified req.

quickwit/quickwit-search/src/leaf.rs
fulmicoton: Individual split errors are now registered as such.

quickwit/quickwit-search/src/leaf.rs
fulmicoton: agreed

quickwit/quickwit-search/src/leaf.rs
fulmicoton: you are right. I removed it here and in range queries too.

quickwit/quickwit-search/src/leaf.rs
fulmicoton: Uri has some special mechanic to redact password when you call display. to_string() calls display.

It might seem silly for index uris, but people use custom storage so it is not that crazy to assume there are  credentials in there. (we also discussed putting encryption keys here)

quickwit/quickwit-search/src/leaf.rs
fulmicoton: hmmm. I think the max_hits was actually a but. Thanks for spotting.

quickwit/quickwit-search/src/leaf.rs
fulmicoton: ok I readded it with a comment explaining the redundancy

quickwit/quickwit-lambda-server/src/handler.rs
fulmicoton: ouch that's AI slop i let slip... and this is larger than I thought. I'll dig a little.

docs/configuration/lambda-config.md
```diff
@@ -0,0 +1,240 @@
...
+1. Creates the Lambda function if it doesn't exist
+2. Updates the function code if the embedded binary has changed
+3. Publishes a new version with a unique identifier
+4. Garbage collects old versions (keeps current + 5 most recent)
+
+This is the recommended approach as it ensures the Lambda function always matches the Quickwit binary version.
+
+### Manual deployment
+
+You can deploy the Lambda function manually without `auto_deploy`:
```
trinity-1686a, line 76: in the future, we can probably make this easier to setup (provide a subcommand that does the deploying, but doesn't start a quickwit server; provide a subcommand that gives the url of the expected release, and version description for the current quickwit binary)

docs/configuration/lambda-config.md
```diff
@@ -0,0 +1,240 @@
...
+1. Download the Lambda zip from [GitHub releases](https://github.com/quickwit-oss/quickwit/releases)
+2. Create or update the Lambda function using AWS CLI, Terraform, or the AWS Console
+3. Publish a version with description format `quickwit:{version}-{sha1}` (e.g., `quickwit:0_8_0-fa752891`)
```
trinity-1686a: is that still accurate now that description contains some configuration params?

quickwit/quickwit-lambda-client/src/deploy.rs
```diff
@@ -0,0 +1,1010 @@
...
+                env!("LAMBDA_BINARY_HASH")
+            )
+        })
+        .as_str()
+}
+
+/// Returns the version description for our qualifier.
+///
+/// We also pass the deploy config, as we want the function to be redeployed
+/// if the deployed is changed.
```
trinity-1686a:
```suggestion
/// We also pass the deploy config, as we want the function to be redeployed
/// if the config is changed.
```

docs/configuration/lambda-config.md
trinity-1686a: that's sha256 now i think

docs/configuration/lambda-config.md
```diff
@@ -0,0 +1,240 @@
...
+}
+```
+
+No additional configuration is needed on the Quickwit side.
+
+## Versioning
+
+Quickwit uses content-based versioning for Lambda:
+- A SHA1 hash of the Lambda binary is computed at build time
+- This hash is embedded in the Lambda function description as `quickwit:{version}-{sha1_short}`
```
trinity-1686a: that's sha256 now

quickwit/quickwit-search/src/search_permit_provider.rs
```diff
@@ -579,12 +615,85 @@ mod tests {
...
+    async fn test_get_permits_with_offload_threshold_max_returns_all() {
+        let permit_provider = SearchPermitProvider::new(100, ByteSize::mb(100));
+        let permits = permit_provider
+            .get_permits_with_offload(vec![ByteSize::mb(1); 8], usize::MAX)
+            .await;
+        assert_eq!(permits.len(), 8);
+    }
+
+    #[tokio::test]
+    async fn test_get_permits_with_offload_threshold_zero_returns_none() {
```
trinity-1686a, line 630: can this also check that nothing was added to the internal queue?

PR #6159, opened 2026-02-13
Implement node based routing table

nadav-govari:
### Description
This PR creates a node based routing table alongside the existing one. The new table isn't used yet. This creates the APIs for loading it with data, and for picking nodes, including the power of two choices implementation.

It's added to the router's state, which subscribes to the new IngesterCapacityUpdate broadcast and uses it to update the table entries.

There are additional changes required before this can be landed in production.
* Remove shard references from the Persist protos.
* Modify error handling: rate limiting, closed shards go away.
* Remove closed shard reporting in GetOrCreateOpenShards.
* Report capacity scores in GetOrCreateOpenShards responses.
* Use the new table in routing decisions.

### How was this PR tested?
New unit tests were added.

These changes are not yet ready for live testing.

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/state.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/state.rs
@@ -154,8 +154,10 @@ impl IngesterState {
     /// queues. Empty queues are deleted, while non-empty queues are recovered. However, the
     /// corresponding shards are closed and become read-only.
     pub async fn init(&self, wal_dir_path: &Path, rate_limiter_settings: RateLimiterSettings) {
-        let mut inner_guard = self.inner.lock().await;
+        // Acquire locks in the same order as `lock_fully` (mrecordlog first, then inner) to
+        // prevent ABBA deadlocks with the broadcast capacity task.
         let mut mrecordlog_guard = self.mrecordlog.write().await;
+        let mut inner_guard = self.inner.lock().await;
 
         let now = Instant::now();
 


quickwit/quickwit-ingest/src/ingest_v2/node_routing_table.rs
```diff
@@ -0,0 +1,420 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::{HashMap, HashSet};
+
+use indexmap::IndexMap;
```
guilload: I don't understand the reason for using an `IndexMap`. This could use a comment below. Tests also pass when using a regular `HashMap`.

quickwit/quickwit-ingest/src/ingest_v2/node_routing_table.rs
```diff
@@ -0,0 +1,420 @@
...
+    ) -> Option<&IngesterNode> {
+        let eligible: Vec<&IngesterNode> = self
+            .nodes
+            .values()
+            .filter(|node| {
+                node.capacity_score > 0
+                    && node.open_shard_count > 0
+                    && ingester_pool.contains_key(&node.node_id)
+                    && !unavailable_leaders.contains(&node.node_id)
+            })
```
guilload, line 74: You can look into `choose_multiple` from `rand` to simply the logic in `power_of_two_choices`.

quickwit/quickwit-ingest/src/ingest_v2/node_routing_table.rs
```diff
@@ -0,0 +1,420 @@
...
+    /// Score from 0-10. Higher means more available capacity.
+    pub capacity_score: usize,
+    /// Number of open shards on this node for this (index, source) pair. Tiebreaker for power of
+    /// two choices comparison - we favor a node with more open shards.
+    pub open_shard_count: usize,
+}
+
+#[derive(Debug)]
+pub(super) struct RoutingEntry {
+    pub nodes: IndexMap<NodeId, IngesterNode>,
```
guilload:
```suggestion
    nodes: IndexMap<NodeId, IngesterNode>,
```

quickwit/quickwit-ingest/src/ingest_v2/node_routing_table.rs
nadav-govari: I thought that the nondeterminism of using a hashmap would "clash" with the random picks and potentially skew the distribution, but that was wrong, apparently the distributions are both still uniformly random. So no need for index map, changing it to hash map.

quickwit/quickwit-ingest/src/ingest_v2/node_routing_table.rs
nadav-govari: Took it

PR #6163, opened 2026-02-19
Use new node based routing table for routing decisions

nadav-govari:
### Description

Hooks up the newly added node based routing table to be used for routing decisions instead of the existing shard based one.

Removes the subscriptions for shard position updates and local shard updates from the ingesters as they're no longer needed.

The old routing table and protos need to be cleaned up. That will be in a followup PR after this one.

### How was this PR tested?
Unit tests and tested on the dogfooding cluster. Works well, and routing works correctly. Cluster is stable until the point where WALs fill up.

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_capacity_score.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/broadcast/ingester_capacity_score.rs
@@ -48,6 +48,7 @@ struct WalDiskCapacityTimeSeries {
 
 impl WalDiskCapacityTimeSeries {
     fn new(memory_capacity: ByteSize) -> Self {
+        #[cfg(not(test))]
         assert!(memory_capacity.as_u64() > 0);
         Self {
             memory_capacity,


quickwit/quickwit-ingest/src/ingest_v2/debouncing.rs
```diff
@@ -69,7 +68,6 @@ impl GetOrCreateOpenShardsRequestDebouncer {
 #[derive(Default)]
 pub(super) struct DebouncedGetOrCreateOpenShardsRequest {
     subrequests: Vec<GetOrCreateOpenShardsSubrequest>,
-    pub closed_shards: Vec<ShardIds>,
```
guilload, line 72: We should still keep the ability for an ingester to tell us about its open and closed shards in the persist response.

quickwit/quickwit-ingest/src/ingest_v2/router.rs
```diff
@@ -102,12 +98,7 @@ pub struct IngestRouter {
 }
 
 struct RouterState {
-    // Debounces `GetOrCreateOpenShardsRequest` requests to the control plane.
```
guilload, line 105: Keep the comment?

quickwit/quickwit-ingest/src/ingest_v2/router.rs
```diff
@@ -171,22 +152,19 @@ impl IngestRouter {
         ingester_pool: &IngesterPool,
     ) -> DebouncedGetOrCreateOpenShardsRequest {
         let mut debounced_request = DebouncedGetOrCreateOpenShardsRequest::default();
-
-        // `closed_shards` and `unavailable_leaders` are populated by calls to `has_open_shards`
-        // as we're looking for open shards to route the subrequests to.
-        let unavailable_leaders: &mut HashSet<NodeId> = &mut workbench.unavailable_leaders;
+        let unavailable_leaders = &workbench.unavailable_leaders;
```
guilload, line 155: Let's keep the type annotation for readibility.

quickwit/quickwit-ingest/src/ingest_v2/router.rs
```diff
@@ -152,12 +139,6 @@ impl IngestRouter {
 
     pub fn subscribe(&self) {
         let weak_router_state = WeakRouterState(Arc::downgrade(&self.state));
-        self.event_broker
-            .subscribe::<LocalShardsUpdate>(weak_router_state.clone())
-            .forever();
-        self.event_broker
```
guilload, line 158: This is used as a belt and suspender mechanism for keep track of indexed positions and truncate via gossip notifications instead of direct RPCs. We must keep it.

quickwit/quickwit-ingest/src/ingest_v2/router.rs
nadav-govari: Okie doke

quickwit/quickwit-ingest/src/ingest_v2/router.rs
nadav-govari: 👍

quickwit/quickwit-ingest/src/ingest_v2/router.rs
nadav-govari: Discussed offline - this is just the router's subscription. The ingesters use a different one: https://github.com/quickwit-oss/quickwit/blob/a004654dc2a785ab192581d1a22bf6e2a9c8bef4/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs#L419

quickwit/quickwit-ingest/src/ingest_v2/debouncing.rs
nadav-govari: Addressing this in a follow up PR.

PR #6257, opened 2026-03-31
[metrics] Support DDSketch in the parquet pipeline

mattmkim:
### Description
This PR can be reviewed commit by commit. 

This PR updates the parquet pipeline to process DDSketches. See https://datadoghq.atlassian.net/wiki/spaces/QKHS/pages/6291357728/DDSketch+in+Parquet for more information about the DDSketch spec.

### How was this PR tested?


quickwit/quickwit-parquet-engine/src/ingest/arrow_sketches.rs
```diff
@@ -0,0 +1,430 @@
...
+            timestamp_secs_builder.append_value(dp.timestamp_secs);
+            count_builder.append_value(dp.count);
+            sum_builder.append_value(dp.sum);
+            min_builder.append_value(dp.min);
+            max_builder.append_value(dp.max);
+            flags_builder.append_value(dp.flags);
+
+            let keys_inner = keys_builder.values();
+            for &k in &dp.keys {
+                keys_inner.append_value(k);
```
ppknap, line 146: Where do we store the length of keys/counts? I.e. how do we recover the number of elements per `data_point` from `_inner` arrays?

mattmkim: they are stored implicitly by Arrow - it stores an array of offsets, so you can get the length of row i with offsets[i+1] - offsets[i]

quickwit/quickwit-parquet-engine/src/split/metadata.rs
```diff
@@ -189,8 +211,9 @@ pub struct MetricsSplitMetadata {
 /// names for JSON backward compatibility while the in-memory representation uses
 /// `Option<Range<i64>>`.
 #[derive(Serialize, Deserialize)]
-struct MetricsSplitMetadataSerde {
-    split_id: SplitId,
+struct ParquetSplitMetadataSerde {
+    kind: ParquetSplitKind,
```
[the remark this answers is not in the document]
mattmkim: can ignore, don't care about this right now

quickwit/quickwit-common/src/lib.rs
```diff
@@ -223,6 +223,19 @@ pub fn is_metrics_index(index_id: &str) -> bool {
...
+///
+/// Sketches indexes use the Parquet/DataFusion pipeline with sketch-specific
+/// processors and writers.
+pub fn is_sketches_index(index_id: &str) -> bool {
+    index_id.starts_with("sketches-")
+}
+
+/// Returns whether the given index ID uses the Parquet/DataFusion pipeline.
+pub fn is_parquet_pipeline_index(index_id: &str) -> bool {
+    is_metrics_index(index_id) || is_sketches_index(index_id)
```
rishabh, line 236: I think this is fine for now, but we'd be adding every parquet pipeline here every time we add one.

Also, every index that starts with sketches- doesn't necessarily have to be a Parquet pipeline.

The "fix" seems to be creating something like a ParquetMapping, and that's like a _whole thing_, so we can get back to it later.

rishabh (approved):

quickwit/quickwit-metastore/migrations/postgresql/28_create-sketch-splits.up.sql
```diff
@@ -0,0 +1,78 @@
...
+
+    -- Full metadata
+    split_metadata_json TEXT NOT NULL,
+
+    -- Timestamps
+    create_timestamp TIMESTAMP NOT NULL DEFAULT (CURRENT_TIMESTAMP AT TIME ZONE 'UTC'),
+    update_timestamp TIMESTAMP NOT NULL DEFAULT (CURRENT_TIMESTAMP AT TIME ZONE 'UTC'),
+    publish_timestamp TIMESTAMP,
+
+    FOREIGN KEY(index_uid) REFERENCES indexes(index_uid)
```
rishabh: Should we add a `ON DELETE CASCADE` here? If we ever need to delete this, we'd have to clean up the table ourselves first.

The metrics table also doesn't include this.

mattmkim: yeah, will add

mattmkim: will need another migration for metrics https://datadoghq.atlassian.net/browse/CLOUDPREM-556

PR #6262, opened 2026-04-03
Add OTel metrics dual export alongside Prometheus

shuheiktgw:
## Summary

- Adds OpenTelemetry metrics instrumentation alongside the existing Prometheus metrics, enabling dual export to both backends.
- Each metric type (`IntCounter`, `IntGauge`, `Gauge`, `Histogram`, and their `Vec` variants) now records to both Prometheus and an OTel instrument.
- OTel instrument creation is **lazy/on-demand**: instruments are not built at metric registration time. Instead, an `OtelState` holds a builder closure and a `OnceLock`. The instrument is created on first use after `install_otel_meter()` has been called. This avoids ordering issues at startup (metrics are registered as `LazyLock` statics, often before the OTel meter provider is configured).

## Known limitation: `IntGauge` relative operations

OTel gauges only support absolute `record(value)` — there is no `add`/`sub` API. For relative operations (`inc`, `dec`, `add`, `sub`), the implementation reads the Prometheus gauge value after mutation and records that to OTel. This is **not atomic**: under concurrent updates the OTel side may briefly observe a stale value. This is acceptable for now because gauges are inherently point-in-time approximations and the next update self-corrects. A TODO is left in the code to manage a single `AtomicI64` as the source of truth if strict correctness is needed later.


shuheiktgw:
I also pushed a branch, [refactor/otel-metrics-dual-export-resolved](https://github.com/quickwit-oss/quickwit/tree/refactor/otel-metrics-dual-export-resolved), where I experimented with solving the IntGauge issue described above by using a callback. The idea was to store the gauge value within the struct and let the exporter scrape it via a callback each time metrics are exported, avoiding race conditions.

This was still a very early-stage POC, and I’ve put it aside for now since it started to feel too complex—but sharing it for context in case we want to revisit this approach.

shuheiktgw:
## Overview

To support OTel Metric Exporters, I've split the former `IntGauge` type into two
semantically distinct types based on how they are used:

- **IntGauge** — for metrics that use only `set()` to record absolute values. Backed by OTel `Gauge<i64>`.
- **IntUpDownCounter** — for metrics that use relative operations (`inc()`, `dec()`, `add()`, `sub()`). Backed by OTel `UpDownCounter<i64>`.

Previously, relative operations on `IntGauge` had to read back the Prometheus value after
mutation and feed it into the OTel gauge via `record()`. This was non-atomic under
concurrent updates and semantically incorrect for OTel. The new split eliminates this by
using `UpDownCounter::add(delta)` directly.

Both types still use `prometheus::IntGauge` on the Prometheus side, so Prometheus metric
names and values are unchanged — except for the two cases noted below.

## Breaking changes

### Prometheus metric name changes

#### `leaf_search_single_split_tasks`

This was a single metric with a status label (`"pending"` / `"ongoing"`), but the two variants had incompatible semantics:

- `ongoing` is managed via `UpDownCounterGuard` (`inc/dec`)
- `pending` is set via `set(queue.len())`

To fix this, the metric was split into two independent metrics with the correct types:

- `quickwit_search_leaf_search_single_split_tasks{status="ongoing"}`
  → `quickwit_search_leaf_search_single_split_tasks_ongoing` (**IntUpDownCounter**)
- `quickwit_search_leaf_search_single_split_tasks{status="pending"}`
  → `quickwit_search_leaf_search_single_split_tasks_pending` (**IntGauge**)

#### `in_flight_data_bytes{component="wal"}`

This field was part of `InFlightDataGauges`, whose fields are all `IntUpDownCounter`.  
However, WAL memory usage is reported via `set(absolute_value)`, which requires `IntGauge`.

The WAL metric was moved out to a dedicated `IntGauge` in `INGEST_V2_METRICS`:

- `quickwit_memory_in_flight_data_bytes{component="wal"}`
  → `quickwit_ingest_wal_memory_allocated_bytes`

---

### OTel instrument type changes (Gauge → UpDownCounter)

The following metrics are now exported as OTel `UpDownCounter` instead of `Gauge`. This shouldn’t be a problem, since we’re adding the OTel metrics exporter in this PR and the metrics from the metrics endpoint remains unchanged.

Affected metrics:

- `quickwit_memory_in_flight_data_bytes`
- `quickwit_thread_pool_ongoing_tasks`
- `quickwit_thread_pool_pending_tasks`
- `quickwit_{subsystem}_grpc_requests_in_flight`
- `quickwit_actor_inboxes_count`
- `quickwit_indexing_indexing_pipelines`
- `quickwit_indexing_split_builders`
- `quickwit_ingest_wal_acquire_lock_requests_in_flight`
- `quickwit_search_searcher_local_kv_store_size_bytes`
- `quickwit_metastore_acquire_connections`
- `quickwit_ongoing_requests`
- `quickwit_pending_requests`
- `quickwit_storage_object_storage_get_slice_in_flight_count`
- `quickwit_storage_object_storage_get_slice_in_flight_num_bytes`
- `quickwit_cache_in_cache_count`
- `quickwit_cache_in_cache_num_bytes`
- `quickwit_cache_virtual_in_cache_count`
- `quickwit_cache_virtual_in_cache_num_bytes`

---

## Not yet addressed

### Prometheus backend removal and `get()` method

The original plan was to serve the `/metrics` endpoint via the OTel Prometheus Exporter,
which would allow:

- Dropping the `prometheus` crate dependency
- Removing `get()` methods used only for the Prometheus read-back pattern

However, the Rust `opentelemetry-prometheus` crate is already deprecated with no viable replacement.

For now:

- The Prometheus backend is retained
- OTel dual-write layer remains in place

https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-prometheus

> Warning: This crate is no longer recommended for use. Development of the Prometheus exporter has been discontinued. See the related https://github.com/open-telemetry/opentelemetry-rust/issues/2451. This crate depends on the unmaintained protobuf crate and has unresolved security vulnerabilities. Version 0.29 will be the final release.

### OtelMetric attributes optimization

Metric attributes were changed:

- From `Vec<KeyValue>` → `Arc<[KeyValue]>`
- Label names from `Vec<String>` → `Arc<[Key]>`

~~However, this may not be sufficient, and I would really appreciate it if you could guide me on how we could further optimize the attribution handling 🙇~~

Updated: I've introduced a cache https://github.com/quickwit-oss/quickwit/pull/6262#issuecomment-4250733358

shuheiktgw:
I applied the same caching strategy the `prometheus` crate uses, but at the Quickwit wrapper level — caching the fully composed metric. Each `*Vec` type now holds an `Arc<RwLock<HashMap<u64, T, BuildNoHashHasher>>>` cache. 

1. **Fast path (read lock)**  
Hash the label values with FNV, look up in the cache. On hit, return a clone (Arc refcount bumps only — zero allocations).

2. **Slow path (write lock)**  
Build the metric (Prometheus lookup + OTel attributes construction), re-check the cache to handle concurrent inserts, then store and return.

### Key implementation details

- **FNV hashing** for speed on short strings, matching the `prometheus` crate's choice.
- **Length-prefixed hashing** to prevent collisions when label boundaries shift (e.g. `["ab", "c"]` vs `["a", "bc"]`). 
- **NoHashHasher** to avoid redundant re-hashing inside the `HashMap` — the key is already a well-distributed FNV hash.
- **Arc<RwLock<...>>** so clones of a `*Vec` (e.g. when a `GrpcMetricsLayer` is cloned into per-connection services) share the same cache.

PR #6277, opened 2026-04-07
feat: case-insensitive regex for lowercasing tokenizers

congx4:
## Summary
- Regex queries now automatically prepend `(?i)` when the field's tokenizer lowercases indexed terms (e.g. `default`, `raw_lowercase`, `lowercase`), so patterns like `.*ECONNREFUSED.*` match correctly against lowercase-indexed data
- `to_field_and_regex` returns a `ResolvedRegex` struct (with tokenizer name), and both `build_tantivy_ast_impl` and warmup `visit_regex` use `TokenizerManager::tokenizer_does_lowercasing` to decide whether to add the flag
- Skips the prefix if the regex already contains `(?i)`

## Motivation

When Trino's Elasticsearch connector translates a SQL `LIKE` predicate on a text field, it sends a **regex queries** to the ES-compatible API. For example:

```sql
-- Trino SQL
SELECT * FROM logs WHERE extra_fts LIKE '%ECONNREFUSED%';
```

In our deployment, the `extra_fts` field uses the `datadog` tokenizer, which lowercases all indexed terms. This means the inverted index only contains lowercase tokens (e.g. `econnrefused`), even if the original log message contained `ECONNREFUSED`.

However, **regex queries** is used without `case_insensitive`:

```json
{"query": {"regexp": {"extra_fts": ".*ECONNREFUSED.*"}}}
```

The regex `.*ECONNREFUSED.*` is matched against the inverted index which only contains lowercase tokens (because the `datadog` tokenizer — or any other lowercasing tokenizer like `default`, `raw_lowercase`, `lowercase` — lowercases during indexing). The uppercase pattern never matches.

Unlike `WildcardQuery`, `RegexQuery` cannot normalize literal parts through the tokenizer because regex metacharacters are interleaved with literal text — you can't reliably separate them for normalization.

The fix: when the field's tokenizer lowercases its output, automatically prepend `(?i)` to make the regex match case-insensitively. This aligns regex behavior with how wildcard queries already work (just via a different mechanism).

### Why the warmup must also change

Quickwit pre-warms search data by walking the query AST and loading relevant term dictionary entries. The warmup builds a regex automaton to scan the term dictionary. If the warmup uses `.*ECONNREFUSED.*` (case-sensitive) but the actual query uses `(?i).*ECONNREFUSED.*` (case-insensitive), the warmup misses the matching terms and the query returns no results. Both must apply the same `(?i)` logic.

## Test plan
- [x] Unit test: lowercasing tokenizer causes `(?i)` prefix (`test_regex_case_insensitive_with_lowercasing_tokenizer`)
- [x] Unit test: already-`(?i)` regex is not doubled (`test_regex_already_case_insensitive_not_doubled`)
- [x] Unit test: `raw` tokenizer does NOT get `(?i)` (`test_regex_no_case_insensitive_with_raw_tokenizer`)
- [x] Unit test: `to_field_and_regex` returns correct tokenizer name for text and JSON fields
- [x] Unit test: `tokenizer_does_lowercasing` correctly identifies lowercasing vs non-lowercasing tokenizers


quickwit/quickwit-query/src/tokenizers/tokenizer_manager.rs
```diff
@@ -91,6 +91,16 @@ impl TokenizerManager {
         self.get_tokenizer(analyzer)
     }
 
+    /// Returns true if the given tokenizer lowercases its output.
+    pub fn tokenizer_does_lowercasing(&self, tokenizer_name: &str) -> bool {
```
trinity-1686a: can you use this function to simplify get_normalizer() ?
also it should probably return None when using an unknown tokenizer

quickwit/quickwit-query/src/query_ast/regex_query.rs
```diff
@@ -48,10 +56,11 @@ impl RegexQuery {
 }
 
 impl RegexQuery {
+    /// Resolves this regex query against the given schema.
     pub fn to_field_and_regex(
```
trinity-1686a:
```suggestion
    pub fn to_resolved(
```

quickwit/quickwit-query/src/query_ast/regex_query.rs
```diff
@@ -104,14 +126,30 @@ impl BuildTantivyAst for RegexQuery {
...
+        // (?i) flag, automatically make the match case-insensitive. Without
+        // this, an upstream regex like `.*ECONNREFUSED.*` would never match
+        // because the inverted index only contains lowercase tokens.
+        let does_lowercasing = match resolved.tokenizer_name.as_deref() {
+            Some(name) => context.tokenizer_manager.tokenizer_does_lowercasing(name),
+            None => false,
+        };
+        let regex = if !resolved.regex.starts_with("(?i)") && does_lowercasing {
+            format!("(?i){}", resolved.regex)
+        } else {
```
trinity-1686a: i wonder if this is always correct, or if there are other `(?X)` modifier that could be put and cause us to miss a case-Insensitive modifier

quickwit/quickwit-query/src/query_ast/regex_query.rs
```diff
@@ -280,20 +318,129 @@ mod tests {
...
+    #[test]
+    fn test_tokenizer_does_lowercasing() {
+        let tokenizer_manager = crate::tokenizers::create_default_quickwit_tokenizer_manager();
+
+        assert!(tokenizer_manager.tokenizer_does_lowercasing("raw_lowercase"));
+        assert!(tokenizer_manager.tokenizer_does_lowercasing("default"));
+        assert!(tokenizer_manager.tokenizer_does_lowercasing("lowercase"));
+        assert!(!tokenizer_manager.tokenizer_does_lowercasing("raw"));
+        assert!(!tokenizer_manager.tokenizer_does_lowercasing("nonexistent"));
+    }
```
trinity-1686a: wrong module, but this test should exists

quickwit/quickwit-query/src/query_ast/regex_query.rs
```diff
@@ -280,20 +318,129 @@ mod tests {
...
+        // Already has (?i), should not be doubled
+        let query = RegexQuery {
+            field: "field".to_string(),
+            regex: "(?i).*ECONNREFUSED.*".to_string(),
+        };
+
+        let result = query.build_tantivy_ast_impl(&context);
+        assert!(
+            result.is_ok(),
+            "regex query with existing (?i) should build successfully"
```
trinity-1686a: would a query with `(?i)(?i)something` fail with this test? (is it testing something, or would the test pass anyway without the branch to check for already case-insensitive query?)

quickwit/quickwit-query/src/query_ast/regex_query.rs
```diff
@@ -280,20 +318,129 @@ mod tests {
...
+        let schema = schema_builder.build();
+
+        let query = RegexQuery {
+            field: "raw_field".to_string(),
+            regex: "abc.*xyz".to_string(),
+        };
+        let resolved = query.to_field_and_regex(&schema).unwrap();
+        assert_eq!(resolved.tokenizer_name.as_deref(), Some("raw"));
+        // The regex should NOT have (?i) since raw doesn't lowercase
+        assert_eq!(resolved.regex, "abc.*xyz");
```
trinity-1686a: add `(?i)` is done in `build_tantivy_ast_impl`, not in `to_field_and_regex`, this test doesn't do what it claims

trinity-1686a (changes requested):
i think the logic for adding `(?i)` should exist only once, probably in `to_field_and_regex` (or whatever its name end up being)
having that logic in two places makes it more error-prone, and i think harder to test too (at least one test doesn't do what it claims because of that, and other tests could check more if the logic was moved)

quickwit/quickwit-query/src/tokenizers/tokenizer_manager.rs
congx4: `tokenizer_does_lowercasing` now returns `Option<bool>` (`None` for unknown). Refactored `get_normalizer` to use it.

quickwit/quickwit-query/src/query_ast/regex_query.rs
congx4: renamed to `to_resolved`.

quickwit/quickwit-query/src/query_ast/regex_query.rs
congx4: replaced `starts_with("(?i)")` with `regex_has_case_insensitive_flag()` that parses the leading `(?...)` group and checks for `i` among the flags. Handles `(?is)`, `(?mi)`, etc. Added dedicated tests.

quickwit/quickwit-query/src/query_ast/regex_query.rs
congx4: moved `test_tokenizer_does_lowercasing` to `tokenizer_manager.rs` tests.

quickwit/quickwit-query/src/query_ast/regex_query.rs
congx4: Test now asserts the actual regex string: `assert_eq!(resolved.regex, "(?i).*ECONNREFUSED.*")` for the prepend case and `assert_eq!(resolved.regex, "(?i).*ECONNREFUSED.*")` (unchanged) for the already-has-flag case. Also moved (?i) logic into `to_resolved` so tests can directly inspect the result.

quickwit/quickwit-query/src/query_ast/regex_query.rs
congx4: moved (?i) logic into `to_resolved` so all tests now check the regex string directly on the resolved result. No more testing through `build_tantivy_ast_impl`.

trinity-1686a (approved):

PR #6284, opened 2026-04-09
Spawn merge pipeline from new compactor

nadav-govari:
### Description

Refactors the MergeScratch to not use the tracking objects that the MergePlanner used to keep track of ongoing merges, since merges are now one shots and are tracked in short-lived objects directly.

I'm trying to keep the existing stuff from breaking so that we can still run CI on these PRs. As a result there will be a decent amount of dead code that I'll need to go back and change.

### How was this PR tested?
Unit tests. Nothing to really test significantly yet.


quickwit/quickwit-compaction/src/compaction_pipeline.rs
```diff
@@ -132,32 +180,153 @@ impl CompactionPipeline {
...
+            .spawn_actor()
+            .set_kill_switch(self.kill_switch.clone())
+            .spawn(merge_packager);
+
+        // MergeExecutor
+        let split_downloader_io_controls = IoControls::default()
+            .set_throughput_limiter_opt(self.io_throughput_limiter.clone())
+            .set_component("split_downloader_merge");
+        let merge_executor_io_controls =
+            split_downloader_io_controls.clone().set_component("merger");
```
nadav-govari, line 243: TODO: throughput throttling

PR #6286, opened 2026-04-09
feat: compute deterministic timeseries_id column at ingest

g-talbot:
## Summary

- Adds a `timeseries_id` column (Int64) to the metrics Arrow batch, computed as a deterministic SipHash-2-4 of the series identity columns
- Hash includes `metric_name`, `metric_type`, and all tags — excludes temporal columns (`timestamp_secs`, `start_timestamp_secs`, `timestamp`) and value columns (`value`, plus DDSketch components from #6257: `count`, `sum`, `min`, `max`, `flags`, `keys`, `counts`)
- Column is already declared in the metrics default sort schema (`metric_name|service|env|datacenter|region|host|timeseries_id|timestamp_secs/V2`), so the writer automatically sorts by it and places it in the correct physical position
- Adds `TimeseriesId` variant to `ParquetField` enum and updates `SORT_ORDER`

## Design reference

[Sorted Series Column for QW Parquet Pipeline](https://docs.google.com/document/d/1-qCxdJrRWgq9NIW67pOg-FMhGIEciEJQxkraQwu-_eI) — this PR implements the Timeseries ID component; the full Sorted Series composite key is a follow-up.

## Test plan

- [x] 8 unit tests for `compute_timeseries_id` (determinism, exclusions, order independence, key/value non-interchangeability)
- [x] All 200 existing tests in `quickwit-parquet-engine` and `quickwit-opentelemetry` pass with updated column counts
- [ ] Clippy clean (pre-existing warning in `reorder_columns` not introduced by this PR)

### DIFF
--- a/quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
+++ b/quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
@@ -2625,6 +2625,7 @@ impl MetastoreService for PostgresqlMetastore {
         if request.split_ids.is_empty() {
             return Ok(EmptyResponse {});
         }
+        let index_uid: IndexUid = request.index_uid().clone();
 
         info!(
             index_uid = %request.index_uid(),
@@ -2632,16 +2633,23 @@ impl MetastoreService for PostgresqlMetastore {
             "deleting metrics splits"
         );
 
-        // Only delete splits that are marked for deletion
-        // Match the non-metrics delete_splits pattern: distinguish
-        // "not found" (warn + succeed) from "not deletable" (FailedPrecondition).
+        // Match the non-metrics delete_splits pattern: CTE with FOR UPDATE
+        // to lock rows before reading state, avoiding stale-state races under
+        // concurrent mark_metrics_splits_for_deletion. Distinguishes "not found"
+        // (warn + succeed) from "not deletable" (FailedPrecondition).
         const DELETE_SPLITS_QUERY: &str = r#"
             WITH input_splits AS (
                 SELECT input_splits.split_id, metrics_splits.split_state
                 FROM UNNEST($2::text[]) AS input_splits(split_id)
-                LEFT JOIN metrics_splits
-                    ON metrics_splits.index_uid = $1
-                    AND metrics_splits.split_id = input_splits.split_id
+                LEFT JOIN (
+                    SELECT split_id, split_state
+                    FROM metrics_splits
+                    WHERE
+                        index_uid = $1
+                        AND split_id = ANY($2)
+                    FOR UPDATE
+                ) AS metrics_splits
+                USING (split_id)
             ),
             deleted AS (
                 DELETE FROM metrics_splits
@@ -2680,7 +2688,7 @@ impl MetastoreService for PostgresqlMetastore {
             .bind(&request.split_ids)
             .fetch_one(&self.connection_pool)
             .await
-            .map_err(|sqlx_error| convert_sqlx_err(&request.index_uid().index_id, sqlx_error))?;
+            .map_err(|sqlx_error| convert_sqlx_err(&index_uid.index_id, sqlx_error))?;
 
         if !not_deletable_ids.is_empty() {
             let message = format!(
--- a/quickwit/quickwit-opentelemetry/src/otlp/arrow_metrics.rs
+++ b/quickwit/quickwit-opentelemetry/src/otlp/arrow_metrics.rs
@@ -24,11 +24,13 @@ use std::io::Cursor;
 use std::sync::Arc;
 
 use arrow::array::{
-    ArrayRef, Float64Builder, RecordBatch, StringDictionaryBuilder, UInt8Builder, UInt64Builder,
+    ArrayRef, Float64Builder, Int64Builder, RecordBatch, StringDictionaryBuilder, UInt8Builder,
+    UInt64Builder,
 };
 use arrow::datatypes::{DataType, Field, Int32Type, Schema as ArrowSchema};
 use arrow::ipc::reader::StreamReader;
 use arrow::ipc::writer::StreamWriter;
+use quickwit_parquet_engine::timeseries_id::compute_timeseries_id;
 use quickwit_proto::bytes::Bytes;
 use quickwit_proto::ingest::{DocBatchV2, DocFormat};
 use quickwit_proto::types::DocUid;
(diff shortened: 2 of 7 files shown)


quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
```diff
@@ -2625,23 +2625,31 @@ impl MetastoreService for PostgresqlMetastore {
...
-                LEFT JOIN metrics_splits
-                    ON metrics_splits.index_uid = $1
-                    AND metrics_splits.split_id = input_splits.split_id
+                LEFT JOIN (
+                    SELECT split_id, split_state
+                    FROM metrics_splits
+                    WHERE
+                        index_uid = $1
+                        AND split_id = ANY($2)
+                    FOR UPDATE
```
alanfgates, line 2650: Adding `FOR UPDATE` means your locking these rows.  Is that what you want?  I'm not clear why this change requires the addition of this lock.

alanfgates (approved):
One question on metadata changes, other than that lgtm.

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
alanfgates: FWIW, Claude likes this addition, saying:
> The addition of FOR UPDATE in the CTE subquery correctly prevents a TOCTOU race where concurrent mark_metrics_splits_for_deletion could change split state between the CTE read and the DELETE. This matches the established pattern in the non-metrics delete_splits at line 1076-1083 of the same file. Good fix.

quickwit/quickwit-parquet-engine/src/timeseries_id.rs
```diff
@@ -0,0 +1,498 @@
...
+//! Rust versions, or process restarts. It is persisted in Parquet files and used
+//! as a sort/grouping key. Any change would silently corrupt compaction and
+//! query results.
+//!
+//! The implementation uses SipHash-2-4 with fixed keys `(0, 0)` from the
+//! `siphasher` crate, and feeds bytes via Rust's `Hash` trait. The `Hash`
+//! implementation for `str` writes `bytes ++ [0xFF]` and for `u8` writes
+//! `[byte]`; this has been stable since Rust 1.0. A pinned stability test
+//! (`test_hash_stability_pinned`) will catch any regression.
+
```
alanfgates, line 35: Claude comments:
> The stability contract relies on Rust's Hash trait implementation for str (writes bytes ++ [0xFF]) and u8 (writes [byte]). While this has been stable since Rust 1.0 and the pinned tests would catch a regression, this is not a guaranteed API contract — it's an implementation detail of the standard library. A Rust toolchain update could theoretically change it.
> Recommendation: Consider hashing raw bytes directly (e.g., hasher.write(metric_name.as_bytes()); hasher.write(&[0xFF]);) rather than going through the Hash trait. This would make the stability explicit in the code rather than relying on an undocumented trait implementation. The pinned tests are a good safety net but catch the problem only after a build — explicit byte feeding prevents it.

It's not clear to me if the added safety here is worth the effort, especially considering a change to the hash implementation for str would probably break a lot of things.

quickwit/quickwit-opentelemetry/src/otlp/arrow_metrics.rs
```diff
@@ -122,6 +127,11 @@ impl ArrowMetricsBatchBuilder {
             metric_type_builder.append_value(dp.metric_type as u8);
             timestamp_secs_builder.append_value(dp.timestamp_secs);
             value_builder.append_value(dp.value);
+            timeseries_id_builder.append_value(compute_timeseries_id(
```
alanfgates, line 135: Comment from Claude:
> compute_timeseries_id is called per-row and sorts the tags each time. For batches where most rows share the same tag key set, this is redundant work. For typical batch sizes this is unlikely to be a bottleneck, but worth being aware of for large batches.

This seems valid, especially since we're storing one value per row, as compaction proceeds we'll be getting identical tag sets next to each other.  Maybe it's too early to optimize this, but leaving a TODO noting we could do it later seems good.

mattmkim:
@codex review

quickwit/quickwit-opentelemetry/src/otlp/arrow_metrics.rs
```diff
@@ -83,6 +86,7 @@ impl ArrowMetricsBatchBuilder {
         fields.push(Field::new("metric_type", DataType::UInt8, false));
         fields.push(Field::new("timestamp_secs", DataType::UInt64, false));
         fields.push(Field::new("value", DataType::Float64, false));
+        fields.push(Field::new("timeseries_id", DataType::Int64, false));
 
         for &tag_key in &sorted_tag_keys {
             fields.push(Field::new(
```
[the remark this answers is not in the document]
mattmkim: not impossible that a user would want to submit a `timeseries_id` tag, we'll probably want to deal with this at some point. i also don't think we want to explicitly reserve the `timeseries_id` tag. adding a TODO

mattmkim:
commit history was messed up, i cherry-picked the 4 relevant commits on top of main and force pushed. just in case, i pushed the original history to the https://github.com/quickwit-oss/quickwit/tree/matthew.kim/gtt/sorted-series-column remote branch

quickwit/quickwit-opentelemetry/src/otlp/arrow_metrics.rs
mattmkim: will add a TODO

quickwit/quickwit-parquet-engine/src/timeseries_id.rs
mattmkim: im going to take Claude's comment here

quickwit/quickwit-parquet-engine/src/timeseries_id.rs
```diff
@@ -0,0 +1,517 @@
...
+) -> i64 {
+    let mut hasher = SipHasher::new_with_keys(0, 0);
+
+    hash_str(&mut hasher, metric_name);
+    hasher.write(&[metric_type]);
+
+    // Collect and sort tags for deterministic ordering.
+    let mut sorted_tags: Vec<(&str, &str)> = tags
+        .iter()
+        .filter(|(k, _)| !EXCLUDED_TAGS.contains(&k.as_str()))
```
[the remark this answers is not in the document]
mattmkim: we're not supporting sketches, yet.

PR #6293, opened 2026-04-11
Mature merger

rdettai-sk:
### Description

Many merge oportunities are missed by the regular pipelines. This is meant to be a tool to reduce split fragmentation on splits.

Main ideas:
- only one instance of this running, focuses on mature split to avoid conflicting with merge pipelines
- we want to start with the quick wins (many tiny splits within very short time spans), then progressevely look for more expensive opportunities.

### How was this PR tested?


rdettai-sk:
Sorry, this was not meant for upstream (yet 😉)

PR #6307, opened 2026-04-14
Upgrade some dependencies

### LINKED ISSUE #5649: Reenable AWS S3 checksumming

guilload:
### Description
Also upgrade S3 SDK and introduce `disable_checksums` storage config option to maintain backward compatibility for non-S3 backends. Fix #5649.

### How was this PR tested?
`make fix`

### DIFF
--- a/quickwit/quickwit-config/src/storage_config.rs
+++ b/quickwit/quickwit-config/src/storage_config.rs
@@ -329,6 +329,8 @@ pub struct S3StorageConfig {
     pub disable_multi_object_delete: bool,
     #[serde(default)]
     pub disable_multipart_upload: bool,
+    #[serde(default)]
+    pub disable_checksums: bool,
 }
 
 impl S3StorageConfig {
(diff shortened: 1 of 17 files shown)


quickwit/quickwit-config/src/storage_config.rs
```diff
@@ -329,6 +329,8 @@ pub struct S3StorageConfig {
     pub disable_multi_object_delete: bool,
     #[serde(default)]
     pub disable_multipart_upload: bool,
+    #[serde(default)]
+    pub disable_checksums: bool,
```
trinity-1686a, line 333: this should be documented so people not using flavours can still make things work with their s3 provider

trinity-1686a (approved):

quickwit/quickwit-config/src/storage_config.rs
guilload: You are absolutely right. Will push a follow up PR.

Issue #6288, opened 2026-04-10
ExistQuery may not be accurate when relying on fast-field


evance-br:
### Describe the bug

ExistQuery such as `field:*` can be wrong when they are satisfied from fast-field storage instead of `index_field_presence`. 
When fast field normalizer is set to `raw`, the fast-field path treats the whole string as one token and drops values longer than 255 bytes, so nothing is written to the fast column for those docs. 
Queries like `NOT field:*` can then exclude documents that do have `field` in the source JSON, while term queries on the same field (e.g. field:"github") still work because they use the inverted index.

### Suggestions:
Maybe instead of dropping the values we could truncate the values, this will break ordering on that fast column but at the moment with the logic of dropping, it's already broken.

### Quickwit version
 - Quickwit 0.8.2
 - Latest dockerhub image (quickwit/quickwit:edge-slim-bookworm)

PR #6309, opened 2026-04-15
Make fast-field normalizers truncate large token instead of dropping

evance-br:
### Description

This changes the implementation of the Quickwit fast-fields normalizers `raw` and `lowercase` to truncate large token to  `DEFAULT_TRUNCATE_TOKEN_LENGTH = 225` instead of dropping the token as it's currently done. This will fix the field presence when the field is fast.

### How was this PR tested?

- Added a unit test
- Ingested data that were failing index field presence query and verified it now work consistently after this change.

Related to: https://github.com/quickwit-oss/quickwit/issues/6288

### DIFF
--- a/quickwit/quickwit-query/src/tokenizers/mod.rs
+++ b/quickwit/quickwit-query/src/tokenizers/mod.rs
@@ -15,6 +15,7 @@
 mod chinese_compatible;
 mod code_tokenizer;
 mod tokenizer_manager;
+mod truncate_tokenizer;
 
 use std::sync::LazyLock;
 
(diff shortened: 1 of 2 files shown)


fulmicoton:
Can you make it a different/new normalizer?

evance-br:
@fulmicoton  Do you mean, with a name that can be used to configure? We will now have `lowercase`, `raw`, and this one.
Sure I can do that but what would you suggest as good name (key)? 
Also should we have a `lowercase` and `raw` version of this one?

PR #6315, opened 2026-04-18
truncate split lists in error logs using PrettySample

fmassot:
## Problem
When many splits fail during search, error logs dump every failed split into a single line, producing multi-KB unreadable logs at scale

## Solution
Use `PrettySample::new(&splits, 5)` to show only the first 5 items + a count, matching the existing pattern in `leaf.rs:1782`

**Before** (hundreds of splits failing):
```
failed_splits = [SplitSearchError { error: "timeout", split_id: "abc123", ... }, SplitSearchError { ... }, ... 200 more ...]
```

**After:**
```
num_failed_splits = 205, failed_splits = [SplitSearchError { ... }, ... (4 more) ... (200 others)]
```


quickwit/quickwit-search/src/fetch_docs.rs
```diff
@@ -80,14 +81,14 @@ async fn fetch_docs_to_map(
...
     .map_err(|error| {
-        let split_ids = splits
+        let split_ids: Vec<&str> = splits
             .iter()
-            .map(|split| split.split_id.clone())
-            .collect_vec();
-        error!(split_ids = ?split_ids, error = ?error, "error when fetching docs in splits");
+            .map(|split| split.split_id.as_str())
+            .collect();
+        error!(num_splits = split_ids.len(), split_ids = ?PrettySample::new(&split_ids, 5), error = ?error, "error when fetching docs in splits");
```
trinity-1686a, line 88: this log is strange, you'd think we failed at searching into these splits, but afaict, we only failed to search into one of these

trinity-1686a (approved):

quickwit/quickwit-search/src/fetch_docs.rs
fmassot: Agreed... I changed the logs, better?

quickwit/quickwit-search/src/fetch_docs.rs
```diff
@@ -65,32 +65,27 @@ async fn fetch_docs_to_map(
...
             .ok_or_else(|| anyhow::anyhow!("failed to find offset for split {}", split_id))?;
-        split_fetch_docs_futures.push(fetch_docs_in_split(
-            searcher_context.clone(),
-            global_doc_addrs,
-            index_storage.clone(),
-            split_and_offset,
-            doc_mapper.clone(),
-            snippet_request_opt,
-        ));
+        let split_id = split_id.to_string();
```
fmassot, line 68: We have one allocation here... but I think we can accept this.

trinity-1686a: we can afford this allocation

trinity-1686a (approved):

PR #6332, opened 2026-04-22
Remerge multi day splits

rdettai-sk:
### Description

Add --split-timestamp-days-range  to enable merges of splits spanning multiple days. During a given run, only a specific range can be executed (typically 0, 1 or 2 days). Specifying 0 days only considers split for which the start and end is on the same day, which is the current behavior.


rdettai-sk:
again wrong upstream :-(

PR #6338, opened 2026-04-24
fix: encode descending sort columns correctly in sorted_series key

g-talbot:
## Summary

- The sorted_series key uses storekey for order-preserving binary encoding, but storekey always encodes ascending
- For sort schemas with descending pre-timestamp columns (e.g. `-service`), the encoded bytes were not inverted, causing sorted_series bytes to not be monotonically ascending in the file
- This would break DataFusion streaming aggregation assumptions and merge correctness for any schema with descending pre-timestamp columns
- Fix: bitwise-NOT the storekey bytes for descending columns (the standard ordered-code technique, see Google's OrderedCode)

### Changes

- `KeyColumn` now carries `descending: bool` from the parsed sort schema
- `encode_row_key` inverts storekey bytes for descending columns (ordinal + value together)
- `invert_bytes` helper does the byte flipping
- `timeseries_id` is always ascending (it's a hash tiebreaker, not a semantic ordering)

## Test plan

- [x] New test: ascending schema gives `alpha < zebra` in key bytes; descending schema gives `zebra < alpha` (inverted)
- [x] All 273 existing tests pass (zero regressions)
- [x] `cargo clippy` clean
- [x] `rustfmt` clean


quickwit/quickwit-parquet-engine/src/sorted_series/tests.rs
```diff
@@ -837,3 +837,39 @@ proptest! {
         );
     }
 }
+
+#[test]
+fn test_descending_column_inverts_key_order() {
```
mattmkim, line 842: used AI to help, I think this implementation doesn't work if we have a row with a null value for a descending sort field? just gonna paste the example. but AI says we can not invert the ordinal bytes, and just invert the encoded values, which i think makes sense? 

```text
sort schema:
metric_name | -service | timeseries_id | timestamp_secs / V2

rows:
A: metric_name=cpu, service="beta",  timeseries_id=10
B: metric_name=cpu, service="alpha", timeseries_id=20
C: metric_name=cpu, service=null,    timeseries_id=30
```

The writer sorts that as:

```text
A, B, C
beta, alpha, null
```

So the keys start like:

```text
A ("beta"):  [ ... !service_ordinal, !encoded("beta"),  tsid_ordinal, ... ]
B ("alpha"): [ ... !service_ordinal, !encoded("alpha"), tsid_ordinal, ... ]
C (null):    [ ... tsid_ordinal, ... ]
```

And that is the problem:

```text
tsid_ordinal       is small, like 0x02
!service_ordinal   is large, like 0xFE
```

So bytewise comparison says:

```text
null row key < non-null row key
```

which means `sorted_series` wants:

```text
C, A, B
```

But the file is physically:

```text
A, B, C
```

So the key order decreases at the end. That violates the monotonicity invariant in [quickwit/quickwit-parquet-engine/src/sorted_series/tests.rs:684](/Users/matthew.kim/quickwit/quickwit/quickwit-parquet-engine/src/sorted_series/tests.rs:684).

If there are no nulls in descending pre-`timeseries_id` columns, this specific bug does not happen.

g-talbot: Good catch.  I didn't realize it was inverting the ordinal too. Fixed.

mattmkim (approved):

PR #6340, opened 2026-04-24
enable partitioning for metrics data

mattmkim:
### Description
Enables partitioning of metric (points and sketches) data. A `partition_id` will be added to split metadata (the `split_metadata_json` field).

Summary of functionality (basically mirrors logs): 
  - Evaluates the index `partition_key` routing expression against RecordBatch rows.
  - Splits incoming metrics/sketch Arrow batches into partition-local batches in ParquetDocProcessor.
  - Carries all partition-local batches from one RawDocBatch in a single ProcessedParquetBatch, with one checkpoint delta covering the whole source batch.
  - Updates ParquetIndexer to maintain per-partition accumulators and flush the whole workbench together on threshold, force commit, timeout, or shutdown. (potential downside of this is the potential of having small partition batches/small files, but this greatly simplifies the implementation, and is what logs does).
  - Enforces max_num_partitions for metrics partitions and routes overflow partitions to OTHER_PARTITION_ID.
  - Updates ParquetPackager to write one split per flushed partition batch and persist partition_id into split metadata.

 This PR keeps the new Arrow-backed routing and metrics pipeline changes behind the existing `metrics` feature. `quickwit-doc-mapper` now exposes its Arrow `RoutingExprContext` only when `quickwit-doc-mapper/metrics` is enabled, and `quickwit-indexing/metrics` wires that feature in alongside its existing optional `arrow` and `quickwit-parquet-engine` dependencies.

for @g-talbot : `partition_id` is not a column in postgres, it's in a JSON column, so we can't effectively filter/group by partition_id in postgres. but looking at how logs does compaction, they don't filter by `partition_id` from postgres, they just group file sin memory by `partition_id` after getting all the metadata back. 

### How was this PR tested?


g-talbot (approved):
LGTM'ing to keep you unblocked, but there's a codex comment and it's a real one that I had to deal with in the merger too.

PR #6350, opened 2026-04-27
Add compaction metrics

nadav-govari:
### Description

Adds metrics for the new compaction planner and compactors.
Emits successes, failures, and in progress, by source_uid and merge level. Merge level is approximated by taking the max of the splits being merged.

Also tracks queue length on the planner, which can eventually be used to set an autoscaling policy.


quickwit/quickwit-compaction/src/planner/compaction_planner.rs
```diff
@@ -153,16 +156,32 @@ impl CompactionPlanner {
         let splits = self
             .metastore
             .list_splits(request)
-            .await?
+            .await
```
nadav-govari, line 159: we want grpc metrics

PR #6354, opened 2026-04-29
feat: Phase 3d+3e — ParquetMergePipeline supervisor + publisher feedback

g-talbot:
## Summary

Stacked on #6358 (Phase 3c). Adds the supervisor and feedback loop that make the merge pipeline runnable.

- **ParquetMergePipeline supervisor**: spawns all merge actors (publisher → sequencer → uploader → executor → downloader → planner), periodic health-check supervision loop, respawn on failure with backoff, graceful shutdown via `FinishPendingMergesAndShutdownPipeline` that disconnects feedback and runs finalize policy for cold windows.
- **Publisher feedback**: adds `parquet_merge_planner_mailbox_opt` to `Publisher` (feature-gated `cfg(feature = "metrics")`). After successful `ParquetSplitsUpdate` publish of new ingested splits, sends `ParquetNewSplits` to the planner. Merge outputs are not fed back (guards against infinite loops).
- **DisconnectMergePlanner** extended to clear both Tantivy and Parquet planner mailboxes.

### Note for reviewers: spawn concurrency semaphore

The Parquet merge pipeline has its own `SPAWN_PIPELINE_SEMAPHORE` (limit 10), separate from the Tantivy merge pipeline's semaphore. This means up to 10 Parquet pipelines and 10 Tantivy pipelines could be spawning simultaneously (20 total), each hitting the metastore. This may be desirable for isolation (one pipeline type can't starve the other's spawns), but reviewers should decide if a shared semaphore would be better for metastore protection. Note that actual merge *execution* concurrency is already shared via `MergeSchedulerService`'s single semaphore.

## Test plan

- [x] 3 pipeline tests (spawn+supervise, shutdown drain, initial splits)
- [x] 3 existing publisher tests pass (no regression)
- [x] Compiles with and without `metrics` feature
- [x] `cargo clippy` clean, license headers OK

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/metrics_pipeline/parquet_merge_pipeline.rs
+++ b/quickwit/quickwit-indexing/src/actors/metrics_pipeline/parquet_merge_pipeline.rs
@@ -41,6 +41,7 @@ use quickwit_common::temp_dir::TempDirectory;
 use quickwit_parquet_engine::merge::policy::ParquetMergePolicy;
 use quickwit_parquet_engine::split::ParquetSplitMetadata;
 use quickwit_proto::metastore::MetastoreServiceClient;
+use quickwit_storage::Storage;
 use tokio::sync::Semaphore;
 use tracing::{debug, error, info, instrument};
 
@@ -51,14 +52,16 @@ use super::{METRICS_PUBLISHER_NAME, ParquetUploader};
 use crate::actors::pipeline_shared::wait_duration_before_retry;
 use crate::actors::publisher::DisconnectMergePlanner;
 use crate::actors::{MergeSchedulerService, Publisher, Sequencer, UploaderType};
-use quickwit_storage::Storage;
 
-/// Spawning a merge pipeline puts pressure on the metastore, so we limit
-/// concurrent spawns (shared with Tantivy merge pipelines).
+/// Limits concurrent Parquet merge pipeline spawns to avoid overwhelming the
+/// metastore. This is a separate semaphore from the Tantivy merge pipeline's.
 static SPAWN_PIPELINE_SEMAPHORE: Semaphore = Semaphore::const_new(10);
 
 pub const SUPERVISE_LOOP_INTERVAL: Duration = Duration::from_secs(1);
 
+/// Holds actor handles for health-checking and lifecycle management.
+/// When `None` in the parent pipeline, no actors are running (pre-spawn or
+/// post-terminate). The supervisor checks these on each `SuperviseLoop` tick.
 struct ParquetMergePipelineHandles {
     merge_planner: ActorHandle<ParquetMergePlanner>,
     merge_split_downloader: ActorHandle<ParquetMergeSplitDownloader>,
@@ -69,6 +72,9 @@ struct ParquetMergePipelineHandles {
 }
 
 impl ParquetMergePipelineHandles {
+    /// Rate-limits progress checks to once per HEARTBEAT interval.
+    /// Without this, every supervision tick would check progress, which
+    /// is wasteful — actors only need to demonstrate liveness periodically.
     fn should_check_for_progress(&mut self) -> bool {
         let now = Instant::now();
         let check_for_progress = now > self.next_check_for_progress;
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-indexing/src/actors/metrics_pipeline/parquet_merge_pipeline.rs
```diff
@@ -0,0 +1,621 @@
...
+    /// followed by SuperviseLoop (which starts periodic health checks).
+    async fn initialize(&mut self, ctx: &ActorContext<Self>) -> Result<(), ActorExitStatus> {
+        self.handle(Spawn::default(), ctx).await?;
+        self.handle(SuperviseLoop, ctx).await?;
+        Ok(())
+    }
+}
+
+impl ParquetMergePipeline {
+    pub fn new(
```
mattmkim, line 143: are we waiting to run compaction outside of the indexing pipeline? i don't see the `ParquetMergePipeline` being created anywhere.

until we are able to run compaction as its own service, should we wire everything up in the meantime?

quickwit/quickwit-indexing/src/actors/metrics_pipeline/parquet_merge_pipeline.rs
```diff
@@ -0,0 +1,621 @@
...
+        // 1. Merge publisher — publishes merged splits to the metastore and feeds back
+        //    ParquetNewSplits to the planner for further merging.
+        let merge_publisher = Publisher::new(
+            METRICS_PUBLISHER_NAME,
+            QueueCapacity::Unbounded,
+            self.params.metastore.clone(),
+            None, // No Tantivy planner
+            None, // No source
+        )
+        .set_parquet_merge_planner_mailbox(self.merge_planner_mailbox.clone());
```
mattmkim, line 262: related comment about wiring up compaction in the indexing pipeline - i believe we need to set the merge planner mailbox on the index publisher, as well.

1) adding a `pub merge_planner_mailbox: Mailbox<MergePlanner>,` to `MetricPipelineParams` https://github.com/quickwit-oss/quickwit/blob/4d6fd212d23a4794e5eb9b4a38da62e27897632e/quickwit/quickwit-indexing/src/actors/metrics_pipeline/pipeline.rs#L81-L99
2) Calling `set_parquet_merge_planner_mailbox` on the publisher https://github.com/quickwit-oss/quickwit/blob/4d6fd212d23a4794e5eb9b4a38da62e27897632e/quickwit/quickwit-indexing/src/actors/metrics_pipeline/pipeline.rs#L317-L323

quickwit/quickwit-indexing/src/actors/metrics_pipeline/publisher_impl.rs
```diff
@@ -87,6 +87,22 @@ impl Handler<ParquetSplitsUpdate> for Publisher {
         info!("publish-metrics-splits");
         suggest_truncate(ctx, &self.source_mailbox_opt, checkpoint_delta_opt).await;
 
+        // Feedback loop: notify the merge planner about newly ingested splits so
+        // it can plan compaction merges.
+        //
+        // Guard: only feed back splits from ingest (replaced_split_ids is empty).
```
mattmkim: this comment doesn't make sense to me? don't we definitely want to try and merge already merged splits?

there are guards in place that prevent infinite merges, right? (max # of merges a file can go through?)

mattmkim (approved):
LGTMing to unblock, but some comments

quickwit/quickwit-indexing/src/actors/metrics_pipeline/parquet_merge_pipeline.rs
g-talbot: https://github.com/quickwit-oss/quickwit/pull/6367

quickwit/quickwit-indexing/src/actors/metrics_pipeline/parquet_merge_pipeline.rs
g-talbot: https://github.com/quickwit-oss/quickwit/pull/6367

quickwit/quickwit-indexing/src/actors/metrics_pipeline/publisher_impl.rs
g-talbot: Fixed.  You're correct.  Now does what Tantivy does.

PR #6359, opened 2026-04-29
feat: compute per-output split metadata in merge engine

g-talbot:
## Summary

Stacked on #6351 (Phase 2). Should be merged **before** #6352 (Phase 3a).

The merge engine now extracts `metric_names`, `time_range`, and `low_cardinality_tags` (service) from each output file's actual rows during the merge write pass.

**Problem**: `MergeOutputFile` previously only had physical metadata (num_rows, size_bytes, row_keys, zonemaps). The downstream `metadata_aggregation` function inferred logical metadata by unioning all input splits. This is incorrect when `num_outputs > 1` — each output contains a different subset of the globally sorted rows and should have metadata reflecting only its own data.

**Fix**: Each `MergeOutputFile` now carries per-output logical metadata extracted from the `sorted_batch` before writing. Reuses `extract_metric_names`, `extract_service_names`, `extract_time_range` from `split_writer` (made `pub(crate)`).

## Test plan

- [x] New test `test_merge_per_output_metadata_from_actual_rows` — verifies 2-output merge has correct per-output metric_names and time_range
- [x] Updated `test_merge_multiple_outputs` with per-output metadata assertions
- [x] All 66 merge tests pass (including proptests)
- [x] `cargo clippy` clean

### DIFF
--- a/quickwit/quickwit-parquet-engine/src/merge/writer.rs
+++ b/quickwit/quickwit-parquet-engine/src/merge/writer.rs
@@ -39,6 +39,8 @@ use super::{InputMetadata, MergeConfig, MergeOutputFile};
 use crate::row_keys;
 use crate::sort_fields::parse_sort_fields;
 use crate::sorted_series::SORTED_SERIES_COLUMN;
+use crate::split::TAG_SERVICE;
+use crate::storage::split_writer::{extract_metric_names, extract_service_names, extract_time_range};
 use crate::storage::{
     PARQUET_META_NUM_MERGE_OPS, PARQUET_META_ROW_KEYS, PARQUET_META_ROW_KEYS_JSON,
     PARQUET_META_SORT_FIELDS, PARQUET_META_WINDOW_DURATION, PARQUET_META_WINDOW_START,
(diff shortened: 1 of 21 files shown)


quickwit/quickwit-parquet-engine/src/merge/writer.rs
```diff
@@ -114,6 +118,19 @@ pub fn write_merge_outputs(
         let output_filename = format!("merge_output_{}.parquet", Ulid::new());
         let output_path = output_dir.join(&output_filename);
 
+        // Extract per-output logical metadata from the actual rows.
+        let metric_names = extract_metric_names(&sorted_batch)
```
mattmkim, line 122: codex caught this - `extract_metric_names` assumes that metric_name is always dictionary encoded: https://github.com/quickwit-oss/quickwit/blob/9763a93815bad7ea824d9424074ff1ee606ffbc1/quickwit/quickwit-parquet-engine/src/storage/split_writer.rs#L205-L224

there's a chance that the `optimize_output_batch` does not dictionary encode the `metric_name` column though, if we don't hit the threshold. we might need to update this `extract_metric_names` helper.

g-talbot: Good catch.  Fixed.

mattmkim (approved):

PR #6361, opened 2026-04-29
Use automaton for multi term query warmups

rdettai-sk:
### Description

We noticed that term sets and multi term queries can be very slow. Things are especially bad for term sets targeting dynamic fields -> the whole `_dynamic` postings list is currently warmed up, which can be huuuuuge (in our case close to 10% of the split).

This PR uses the automaton warmup for each field that has more than N (currently N=2) non positional terms inside the query. 

This PR doesn't address the question of positional terms.

Unfortunately I didn't find a good way to test that the warm up is indeed smaller.

### DIFF
--- a/quickwit/quickwit-doc-mapper/src/doc_mapper/mod.rs
+++ b/quickwit/quickwit-doc-mapper/src/doc_mapper/mod.rs
@@ -26,6 +26,7 @@ use std::collections::{HashMap, HashSet};
 use std::fmt::Debug;
 use std::ops::Bound;
 
+use anyhow::bail;
 pub use doc_mapper_builder::DocMapperBuilder;
 pub use doc_mapper_impl::DocMapper;
 pub use field_mapping_entry::{
@@ -41,6 +42,7 @@ pub use field_mapping_type::FieldMappingType;
 use serde_json::Value as JsonValue;
 use tantivy::Term;
 use tantivy::schema::{Field, FieldType};
+use tantivy_fst::Automaton as TantivyFstAutomaton;
 pub(crate) use tokenizer_entry::{
     NgramTokenizerOption, RegexTokenizerOption, TokenFilterType, TokenizerType,
 };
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-doc-mapper/src/doc_mapper/mod.rs
```diff
@@ -76,10 +78,70 @@ pub struct TermRange {
...
+    /// An exact-match automaton for a TermSet query.
+    TermSet(ExactSetAutomaton),
+}
+
+/// A byte-level DFA that accepts exactly the strings in a sorted, deduplicated byte-sequence
+/// set. State = `(depth, lo, hi)` meaning all terms in `self.terms[lo..hi]` share the first
+/// `depth` bytes consumed so far. Transitions are computed via binary search, avoiding any
+/// upfront DFA materialisation.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ExactSetAutomaton {
```
trinity-1686a, line 93: can that struct be moved in a dedicated file, and can you add more testing to the automaton please (a proptest would be nice too)

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -219,8 +218,13 @@ pub(crate) fn build_query(
             .or_default() |= need_position;
     });
 
+    coalesce_multi_term_fields_into_automatons(
+        &mut terms_grouped_by_field,
+        &mut automatons_grouped_by_field,
+        2,
```
trinity-1686a, line 224: this should be a const declared early in the file. I don't know what a good value is, but 2 is too low. Running an automaton on the sstable index is much costlier than doing a couple of point queries in it. I don't know what a good value is, but i'd be more comfortable with this being at least 10-15 until we have measurements to know how to thune this. Otherwise I'm worried this might impact `simple queries like this` on the default field, and similar query patterns

quickwit/quickwit-doc-mapper/src/query_builder.rs
```diff
@@ -231,44 +235,54 @@ pub(crate) fn build_query(
...
+///
+/// A minimum of `term_threshold` terms is required because
+/// `warm_postings_automaton` has higher per-call overhead than a direct point
+/// lookup: spawning a CPU task and traversing the sstable twice. That overhead
+/// is only worth paying when there are enough terms to coalesce.
+///
+/// Terms that require positions are left in `terms_grouped_by_field` unchanged,
+/// as they must be fetched individually.
+///
+/// TODO: should positional terms also support some form of grouping?
```
trinity-1686a, line 254: this would need tantivy support to exist, but wouldn't be that hard to do: a TermInfo contains both offset to posting list and position list. I'd like to have evidence that this coalescing is useful on term queries before investing ressources into what might be a speedup for only the longest of phrase queries

trinity-1686a (commented):
sorry for the delay :-/

PR #6363, opened 2026-04-29
Zone Map Pruning for Metrics

alexanderbianchi:
## Summary

- Adds conservative scan-time metadata pruning for metrics splits after metastore split discovery.
- Uses exact split metadata first (`metric_name`, low-cardinality tags), then falls back to per-column `zonemap_regexes` when available.
- Supports string equality, `IN`, and simple prefix `LIKE` patterns such as `host LIKE 'ID-07%'`.
- Evaluates prefix `LIKE` with DFA prefix-language matching against the zonemap regex, so it checks whether any value matching the prefix could exist in the split.
- Caches compiled exact-match regexes and prefix DFAs with `mini-moka`.
- Keeps splits when metadata is missing, regexes are invalid, DFA compilation exceeds the per-regex limit, or an expression is outside the supported pruning subset.
- Allows metadata pruning for any declared table column that has split metadata, not only the default metrics tag columns.

## Notes

Zonemap metadata is a `column_name -> superset_regex` map generated by the parquet writer for string-valued sort-schema columns. This PR only prunes when the fetched split metadata has usable information for the queried column; DataFusion still applies the row-level filter afterward.

The Java/logs evaluator tests that apply to metrics string zonemaps were ported. Logs-specific integer min/max, hash rowset, ReaderQL function, and case-insensitive ReaderQL cases are out of scope for this metrics/DataFusion path.

## Testing

- `cargo fmt -p quickwit-datafusion`
- `cargo test -p quickwit-datafusion`


alexanderbianchi:
@codex review

quickwit/quickwit-datafusion/src/sources/metrics/table_provider.rs
```diff
@@ -56,6 +62,7 @@ const METRICS_SORT_ORDER: &[&str] = &[
     "timeseries_id",
     "timestamp_secs",
 ];
+const ZONEMAP_DFA_SIZE_LIMIT_BYTES: usize = 1_000_000;
```
g-talbot, line 68: What's this number?

quickwit/quickwit-datafusion/src/sources/metrics/table_provider.rs
```diff
@@ -277,6 +315,190 @@ fn column_name_from_expr(expr: &Expr) -> Option<String> {
...
+                "ignoring invalid split zonemap regex"
+            );
+            return true;
+        }
+    };
+    values.iter().any(|value| regex.is_match(value))
+}
+
+fn zonemap_may_match_any_prefix(superset_regex: &str, prefixes: &[String]) -> bool {
+    let dfa_config = dense::Config::new()
```
g-talbot: It will be very important for performance to cache the config and DFA.  It can be very expensive to build these.

quickwit/quickwit-datafusion/src/sources/metrics/table_provider.rs
```diff
@@ -476,4 +711,147 @@ mod tests {
             "default metrics sort metadata should enable the advertised ordering"
         );
     }
+
+    #[test]
```
g-talbot, line 847: Are we doing all the same tests that the Java code did?

g-talbot (changes requested):
Two concerns.  The first is that the compiled DFA for the regex should be cached.  Creating it can be a very expensive operation.  Second, let's make sure that all of the tests that the Java code ran against this matching are translated here.  Other than that, this LG and I'll LGTM once these are addressed.

quickwit/quickwit-datafusion/src/sources/metrics/table_provider.rs
alexanderbianchi: Addressed for the matching semantics that apply to metrics metadata. I added tests mirroring the Java evaluator string-zonemap cases: missing metadata, exact regex, superset regex, case-sensitive matching, multi-column AND, conservative OR, IN lists, invalid regex, newline/DOTALL behavior, and unsupported string forms staying conservative. The Java integer min/max, hash rowset, ReaderQL function, and case-insensitive ReaderQL equality cases are logs/Event Store-specific, so I called those out in the PR description instead of porting them as metrics tests.

quickwit/quickwit-datafusion/src/sources/metrics/table_provider.rs
alexanderbianchi: Addressed. The scan path now caches compiled `regex::Regex` values and compiled prefix DFAs by zonemap regex string, uses shared DFA/syntax config, and keeps the DFA cache bounded separately because each DFA can be much larger than a `regex::Regex`.

quickwit/quickwit-datafusion/src/sources/metrics/table_provider.rs
alexanderbianchi: Added a code comment here. The 1 MB limit caps DFA determinization/memory for pathological zonemap regexes; if compilation exceeds it, we fall back to conservative keep-split behavior.

alexanderbianchi:
@codex review

g-talbot (approved):

PR #6365, opened 2026-04-30
update tantivy

PSeitz-dd:
## Summary
- Bump tantivy to commit 4480cf0

### DIFF
--- a/quickwit/Cargo.lock
+++ b/quickwit/Cargo.lock
@@ -6915,7 +6915,7 @@ checksum = "1a80800c0488c3a21695ea981a54918fbb37abf04f4d0720c453632255e2ff0e"
 [[package]]
 name = "ownedbytes"
 version = "0.9.0"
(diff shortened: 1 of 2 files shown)


quickwit/Cargo.lock
```diff
@@ -209,7 +209,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
```
trinity-1686a: this version downgrade feels wrong. why is it happening?

PSeitz-dd: I was thinking this would be task claude could do unsupervised... apparently not.

I did the bump manually and there is no downgrade anymore

trinity-1686a (approved):

PR #6368, opened 2026-05-01
paginate metric splits during query execution

mattmkim:
### Description
We were hitting grpc message limits when trying to query splits from the metastore. 

1) Results are not compressed (we could turn it on, but this would apply to all metastore clients - was it not enabled for some reason?)
2) `ListMetricsSplits` is unary RPC (while `ListSplits` is streaming 🙃 ), and we were not paginating. this PR adds pagination. we should eventually move to streaming RPC. 

### How was this PR tested?


mattmkim:
@codex review

alexanderbianchi (approved):

PR #6383, opened 2026-05-05
Improve ingester lock instrumentation

guilload:
### Description
Improve ingester lock instrumentation:
- remove overkill macro
- instrument all places where we acquire lock
- track lock hold durations

Will move to static metrics (and will maybe have to reintroduce macro) when metrics PR is merged.

### How was this PR tested?
cargo test

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
@@ -66,7 +66,7 @@ use crate::ingest_v2::doc_mapper::get_or_try_build_doc_mapper;
 use crate::ingest_v2::metrics::report_wal_usage;
 use crate::ingest_v2::models::IngesterShardType;
 use crate::mrecordlog_async::MultiRecordLogAsync;
-use crate::{FollowerId, estimate_size, with_lock_metrics};
+use crate::{FollowerId, estimate_size};
 
 /// Minimum interval between two reset shards operations.
 const MIN_RESET_SHARDS_INTERVAL: Duration = if cfg!(any(test, feature = "testsuite")) {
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -318,9 +321,11 @@ impl Ingester {
 
         match advise_reset_shards_result {
             Ok(Ok(advise_reset_shards_response)) => {
-                let mut state_guard =
-                    with_lock_metrics!(self.state.lock_fully().await, "reset_shards", "write")
-                        .expect("ingester should be ready");
+                let mut state_guard = self
+                    .state
+                    .lock_fully("reset_shards")
```
nadav-govari: Can we differentiate places that acquire the lock multiple times in the same context (we lock above and then again here, and presumably also in other functions?)

quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -310,33 +310,49 @@ impl IngesterState {
...
-        // We assume that the mrecordlog lock is the most "expensive" one to acquire, so we acquire
-        // it first.
-        let mrecordlog_opt_guard = self.mrecordlog.write().await;
-        let inner_guard = self.inner.lock().await;
+        // We assume that the mrecordlog lock is the most "expensive" one to acquire, so we
+        // acquire it first.
+        let ((mrecordlog_opt_guard, inner_guard), acquired_at) =
+            track_acquire_lock(operation, "full", async {
+                let mrecordlog_opt_guard = self.mrecordlog.write().await;
+                let inner_guard = self.inner.lock().await;
```
nadav-govari, line 352: value in tracking these separately?

guilload: I don't know yet. So I'm going to wait until I know I want it for sure.

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
guilload: Yep, good idea. Done.

PR #6390, opened 2026-05-06
use current node load estimation when placing search jobs

trinity-1686a:
When placing jobs, first query all nodes for their current load, and bias placement toward less loaded nodes to even load on the cluster
this improves on a problem where some nodes might be overloaded while other are underloaded, causing queueing despite not all nodes being at max capacity
in testing, this was seen improving slightly p50+ under constant light load, and increased the max qps before latency explodes. Metrics also showed all searcher busy when some would be only part-time working before

future improvements:
- we could debounce calls to GetLoad
- fetch_docs could reuse the same searcher as used in the leaf_search phase, this guarantees a footer-cache hit and respect pre-existing load without the need for more GetLoad calls on the critical path


quickwit/quickwit-search/src/root.rs
```diff
@@ -1674,18 +1674,19 @@ async fn assign_client_fetch_docs_jobs(
...
         .await?;
 
     Ok(assigned_jobs)
 }
 
 // Measure the cost associated to searching in a given split metadata.
-fn compute_split_cost(split_metadata: &SplitMetadata) -> usize {
+pub(crate) fn compute_split_cost(num_docs: u64) -> usize {
     // TODO this formula could be tuned a lot more. The general idea is that there is a fixed
     // cost to searching a split, plus a somewhat-linear cost depending on the size of the split
```
PSeitz, line 1766: this should include if we have aggregations or not, or ideally a cost from the query

quickwit/quickwit-search/src/search_job_placer.rs
```diff
@@ -177,10 +201,50 @@ impl SearchJobPlacer {
...
-                load: 0,
+                load: None,
             })
             .collect();
 
+        if load_aware {
+            // Seed each candidate node with its current load so the placer avoids
+            // routing work to already-loaded nodes. If a node fails to report its
+            // load (error or timeout), `load` stays `None`: we still route work
+            // there if all other nodes are overloaded, but we prefer reachable
```
PSeitz, line 218: maybe we should error instead if all nodes are not reachable or overloaded

quickwit/quickwit-search/src/root.rs
trinity-1686a: i'm not sure how to factor that in exactly, we don't have a good cost model for aggregations.

A simple `count() group by X` on a low cardinality field isn't usually very expensive, but a `cardinality(Y) group by Z` with both being high cardinality is much more expensive, though i have no idea of by how-much. And this also depends on the selectivity of the query

quickwit/quickwit-search/src/service.rs
```diff
@@ -320,6 +324,10 @@ impl SearchService for SearchServiceImpl {
         let search_plan = search_plan(search_request, self.metastore.clone()).await?;
         Ok(search_plan)
     }
+
+    async fn get_load(&self) -> usize {
+        self.searcher_context.search_permit_provider.get_load()
```
Platane, line 328: would it be useful to add a metrics to track the load per node ?

quickwit/quickwit-search/src/root.rs
PSeitz-dd: We would probably have a very rough estimate on the query AST. Can you add the TODO to eventually track aggregation cost and estimate query selectivity?

quickwit/quickwit-search/src/search_job_placer.rs
```diff
@@ -177,10 +201,50 @@ impl SearchJobPlacer {
...
+            // nodes first.
+            //
+            // The timeout is intentionally short: a slow response is treated the
+            // same as no response so that one unresponsive node cannot delay the
+            // entire query.
+            const GET_LOAD_TIMEOUT: Duration = Duration::from_millis(200);
+            let load_futures = candidate_nodes.iter_mut().map(|node| {
+                let mut client = node.client.clone();
+                async move { tokio::time::timeout(GET_LOAD_TIMEOUT, client.get_load()).await }
```
PSeitz, line 227: I wonder if we should have a 1 sec cache for the load to reduce traffic

quickwit/quickwit-search/src/search_job_placer.rs
```diff
@@ -190,22 +254,55 @@ impl SearchJobPlacer {
 
         let total_load: usize = jobs.iter().map(|job| job.cost()).sum();
 
+        // Compute `target_load` using only reachable nodes (those with a known
```
PSeitz, line 263: Braindump:

I think there's a risk involved with regards to caching. 
Assuming node A has the highest affinity for split X
=> We can assume node A has caches filled for split X. 

Scenario: high load. 
node A = 110, node B = 100.
=> We end up putting queries for split X split on node B. This has a higher cost because node B needs to fill the caches for split X (2x query cost?). The formula we use always adds a constant 5 to fill caches. 
This may cause cache evictions on node B, which increases the real cost more.
=> higher total load on the cluster

The query will be longer on node B, which will report the cost for longer, so it has self-correcting properties.

We can expect for some splits to be more queried often, because they are newer, so we want more nodes on them (for some time) and this is in the right direction.

trinity-1686a: > I think there's a risk involved with regards to caching

yes i agree. The hope, and what tests up to know suggests, is that the more even load makes it up in both individual query latency and throughput (qps)

quickwit/quickwit-search/src/search_permit_provider.rs
```diff
@@ -332,7 +333,9 @@ impl SearchPermitActor {
                 // Increment total_job_cost for all tasks entering the queue (both pending and
                 // those that will be granted immediately by assign_available_permits below).
                 let added: usize = task_metadata.iter().map(|m| m.job_cost).sum();
-                self.total_job_cost.fetch_add(added, Ordering::Relaxed);
+                // fetch_add returns the previous value, so add `added` to get the new total.
+                let new_load = self.total_job_cost.fetch_add(added, Ordering::Relaxed) + added;
+                SEARCHER_NODE_LOAD.set(new_load as f64);
```
Mallets, line 338: nit: it seems that `SEARCHER_NODE_LOAD` stores the same value (_1-to-1 mapping_) of `total_job_cost`.  `SEARCHER_NODE_LOAD` provides a `get()` method to retrieve the value if needed.

trinity-1686a: i'm generally worried by global mutable state. imo it's okay to write to one for telemetry purpose, but i prefer not relying on it for any actual logic

PSeitz-dd (approved):

PR #6395, opened 2026-05-06
clean up docs

g-talbot:


quickwit/quickwit-cli/src/logger.rs
```diff
@@ -241,18 +241,9 @@ pub fn setup_logging_and_tracing(
...
-                "CP_DOGSTATSD_SERVER_HOST",
-                "127.0.0.1".to_string(),
-                false,
-            )
-        });
-    let port: u16 = quickwit_common::get_from_env_opt("CLOUDPREM_DOGSTATSD_SERVER_PORT", false)
-        .unwrap_or_else(|| quickwit_common::get_from_env("CP_DOGSTATSD_SERVER_PORT", 8125, false));
+    let host: String =
+        quickwit_common::get_from_env("QW_STATSD_SERVER_HOST", "127.0.0.1".to_string(), false);
+    let port: u16 = quickwit_common::get_from_env("QW_STATSD_SERVER_PORT", 8125, false);
```
fulmicoton-dd: ouch

fulmicoton-dd:
@g-talbot can you move your doc/internals documentation to a repo that belongs to byoc-metrics?

quickwit/quickwit-cli/src/logger.rs
```diff
@@ -238,58 +238,48 @@ pub fn setup_logging_and_tracing(
...
-        }
-    }
-    metrics_exporter_dogstatsd::DogStatsDBuilder::default()
-        .set_global_prefix("cloudprem")
-        .with_global_labels(global_labels)
-        .with_remote_address(addr)
-        .context("failed to parse DogStatsD server address")?
-        .install()
-        .context("failed to register DogStatsD exporter")?;
+pub fn install_invariant_recorder() {
```
fulmicoton-dd: why do we need this? More importantly, why do we need this in quickwit-cli? Could that be in quickwit-dst?

quickwit/quickwit-cli/src/logger.rs
```diff
@@ -238,58 +238,48 @@ pub fn setup_logging_and_tracing(
...
+    INVARIANT_CHECKED.with_label_values([name]).inc();
     if !passed {
-        metrics::counter!("pomsky.invariant.violated", "invariant" => name).increment(1);
+        INVARIANT_VIOLATED.with_label_values([name]).inc();
     }
 }
 
+#[cfg(not(test))]
+static INVARIANT_CHECKED: std::sync::LazyLock<quickwit_common::metrics::IntCounterVec<1>> =
+    std::sync::LazyLock::new(|| {
```
fulmicoton-dd: same as above. why is this in cli?

quickwit/quickwit-cli/src/main.rs
```diff
@@ -105,7 +105,7 @@ async fn main_impl() -> anyhow::Result<()> {
         setup_logging_and_tracing(command.default_log_level(), ansi_colors, build_info)?;
 
     #[cfg(not(test))]
-    quickwit_cli::logger::setup_metrics(build_info)?;
```
fulmicoton-dd, line 108: why do we remove setup_metrics ? This thing is how we track which build pomsky customers are using

quickwit/Cargo.lock
```diff
@@ -3893,12 +3893,6 @@ dependencies = [
  "cfg-if",
 ]
 
-[[package]]
```
fulmicoton-dd, line 3896: can you make sure this file is generated via cargo check? We have noticed CLAUDE.md tend to edit it on its own from time to time.

fulmicoton-dd:
@g-talbot see comments inline

mattmkim (approved):

quickwit/quickwit-cli/src/main.rs
g-talbot: I told the AI to revert this to what it was before 6246, and here's its response.

 ▎ Thanks — that's a fair pushback. Cleaner answer: I've now removed the recorder wiring from quickwit-cli entirely (commit b33782a5b), so this commit is just reverting main.rs and logger.rs to their pre-#6246   
  ▎ shapes with respect to quickwit-dst. The quickwit-dst crate stays (along with the check_invariant! calls inside quickwit-parquet-engine that are debug_assert-style invariant checks), but no production recorder
  ▎  is installed by the OSS binary anymore.                                                                                                                                                                         
  ▎                                                              
  ▎ Pomsky build-version tracking isn't affected: register_build_info_metric() in main.rs:81 still registers quickwit_build_info{version=…, commit_hash=…, build_date=…, commit_tags=…, target=…} to the Prometheus  
  ▎ registry, exposed on /metrics. Whatever scrapes that endpoint can still identify customer builds.
  ▎                                                                                                                                                                                                                  
  ▎ If pomsky additionally needs the DogStatsD wire format, that's a deployment-specific concern (CLOUDPREM_/CP_/QW_ env vars, hardcoded Kubernetes label conventions, fixed cloudprem global prefix) that's cleaner 
  ▎ to install in pomsky's downstream — quickwit_dst::invariants::set_invariant_recorder is still pluggable, so a 5-line wrapper in pomsky's binary or a downstream patch covers it.

quickwit/quickwit-cli/src/logger.rs
g-talbot: Removed.

quickwit/Cargo.lock
g-talbot: Just checked.  It was.  This is good.

.github/CODEOWNERS
```diff
@@ -1,8 +1,9 @@
 # CODEOWNERS — see https://docs.github.com/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 #
-# Last matching rule per file wins. Only the owners on the matching rule are
-# auto-requested for review; quickwit-core members retain repo write access
-# and can still approve any PR manually.
+# Last matching rule per file wins. Approval from any listed owner satisfies
+# the requirement. quickwit-dev is listed on every rule so it can always
```
fulmicoton-dd, line 4: what is this change about?

quickwit/quickwit-cli/src/main.rs
```diff
@@ -104,9 +104,6 @@ async fn main_impl() -> anyhow::Result<()> {
     let (env_filter_reload_fn, tracer_provider_opt) =
         setup_logging_and_tracing(command.default_log_level(), ansi_colors, build_info)?;
 
-    #[cfg(not(test))]
-    quickwit_cli::logger::setup_metrics(build_info)?;
```
fulmicoton-dd, line 108: ????

Mallets: I think this call should have been preserved?

PR #6399, opened 2026-05-06
fix: set maturity on parquet splits

mattmkim:
### Description
We were not setting maturity value on splits in the metastore, so they were all being set to `0`. Sets maturity value on the ingest/merge path on the uploader, similar to what logs does. 

Need this for compaction, so we can fetch immature splits, and correctly designate when splits are no longer candidates for compaction.

### How was this PR tested?


mattmkim:
@codex review

g-talbot (approved):

guilload (approved):

PR #6404, opened 2026-05-07
Improvements from live testing

nadav-govari:
### Description

There were a number of observations made when testing.

1. When deploying >2 compactors, the split cache becomes an impediment rather than an optimization. We end up with a bunch of splits on disk that we never end up using in a merge. Rather than fix the split cache, remove it (for now. I might come back and fix this down the road)

2. The merge policy `operations` only produced one merge per level per call, which resulted in huge split backlogs buildings. We now construct as many merge operations as we can on every tick/

3. The splits table wasnt sorted the way we needed, so our scans were expensive. Added an index to make this easier.

4. Got rid of the scan cursor. We can do this naturally with a scan limit and a maturity greater than. We'll always fetch the <= 5000 oldest immature splits. As splits get merged and marked for deletion, the head will move naturally.

### How was this PR tested?

Extensive testing on a test cluster that sees meaningful traffic, including a backfill of over 900k immature splits using the compactor architecture.

### DIFF
--- a/quickwit/quickwit-compaction/src/planner/compaction_planner.rs
+++ b/quickwit/quickwit-compaction/src/planner/compaction_planner.rs
@@ -36,19 +36,16 @@ use super::compaction_state::CompactionState;
 use super::index_config_metastore::{IndexConfigMetastore, IndexEntry};
 use crate::planner::metrics::COMPACTION_PLANNER_METRICS;
 
+/// Cap on splits fetched per tick. Every tick, the planner re-scans the immature published set,
+/// sorted by `maturity_timestamp` ASC  so the most-urgent splits are processed first when a backlog
+/// exists. Splits beyond this cap aren't lost -- they bubble into range as the front of the queue
+/// is merged off.
+const SCAN_PAGE_SIZE: usize = 5_000;
+#[derive(Debug)]
 pub struct CompactionPlanner {
     state: CompactionState,
     index_config_metastore: IndexConfigMetastore,
-    cursor: i64,
     metastore: MetastoreServiceClient,
-}
-
-impl Debug for CompactionPlanner {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("CompactionPlanner")
-            .field("cursor", &self.cursor)
-            .finish()
-    }
 }
 
 const SCAN_AND_PLAN_INTERVAL: Duration = Duration::from_secs(5);
@@ -112,15 +109,11 @@ impl Handler<ReportStatusRequest> for CompactionPlanner {
     }
 }
 
-const STARTUP_LOOKBACK: Duration = Duration::from_secs(24 * 60 * 60);
-
 impl CompactionPlanner {
     pub fn new(metastore: MetastoreServiceClient) -> Self {
-        let cursor = OffsetDateTime::now_utc().unix_timestamp() - STARTUP_LOOKBACK.as_secs() as i64;
         CompactionPlanner {
             state: CompactionState::new(),
             index_config_metastore: IndexConfigMetastore::new(metastore.clone()),
-            cursor,
             metastore,
         }
     }
@@ -141,8 +134,6 @@ impl CompactionPlanner {
             if index_entry.is_split_mature(&split.split_metadata) {
                 continue;
             }
-            self.cursor = self.cursor.max(split.update_timestamp);
-            info!(max_timestamp=%self.cursor, "[compaction planner] update metastore cursor min_timestamp cursor");
             self.state.track_split(split.split_metadata);
         }
     }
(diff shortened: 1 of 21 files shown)


quickwit/quickwit-compaction/src/planner/compaction_planner.rs
```diff
@@ -36,21 +36,18 @@ use super::compaction_state::CompactionState;
 use super::index_config_metastore::{IndexConfigMetastore, IndexEntry};
 use crate::planner::metrics::COMPACTION_PLANNER_METRICS;
 
+/// Cap on splits fetched per tick. Every tick, the planner re-scans the immature published set,
+/// sorted by `maturity_timestamp` ASC  so the most-urgent splits are processed first when a backlog
+/// exists. Splits beyond this cap aren't lost -- they bubble into range as the front of the queue
+/// is merged off.
+const SCAN_PAGE_SIZE: usize = 5_000;
+#[derive(Debug)]
```
guilload, line 44:
```suggestion
const SCAN_PAGE_SIZE: usize = 5_000;

#[derive(Debug)]
```

quickwit/quickwit-compaction/src/planner/compaction_state.rs
```diff
@@ -108,14 +110,22 @@ impl CompactionState {
...
-                    .insert(split.split_id().to_string());
+        // `MergePolicy::operations` emits at most one op per level per call, which under a backlog
+        // leaves the bulk of `splits` untouched per tick. Loop until no new operations are created.
+        loop {
+            let operations = merge_policy.operations(splits);
+            if operations.is_empty() {
+                break;
+            }
+            for operation in operations {
+                for split in operation.splits_as_slice() {
```
guilload, line 121:
```suggestion
                for split in &operation.splits {
```
Whoever wrote `splits_as_slice` is a terrorist.

quickwit/quickwit-metastore/migrations/postgresql/27_create-index-splits-published-maturity-timestamp.down.sql
```diff
@@ -0,0 +1 @@
+DROP INDEX IF EXISTS splits_published_maturity_timestamp_idx;
```
guilload: That newline

quickwit/quickwit-metastore/migrations/postgresql/27_create-index-splits-published-maturity-timestamp.up.sql
```diff
@@ -0,0 +1,19 @@
...
+-- whose merge timed out or failed.
+--
+-- The btree on (maturity_timestamp, split_id) lets postgres seek to the live
+-- "still immature" range in index order, satisfying both the filter and the
+-- ORDER BY without an extra sort. The split_id column is included as a
+-- tiebreaker so postgres returns deterministic pages under LIMIT.
+--
+-- The partial predicate is restricted to split_state = 'Published' because
+-- partial-index predicates must be IMMUTABLE; "now()" cannot appear here.
+CREATE INDEX IF NOT EXISTS splits_published_maturity_timestamp_idx
```
guilload: This will make diesel-guard and our users' DB not happy:
```
❌ Unsafe migration detected in quickwit-metastore/migrations/postgresql/27_create-index-splits-published-maturity-timestamp.up.sql

❌ ADD INDEX without CONCURRENTLY  (line 17)

Problem:
  Creating index 'splits_published_maturity_timestamp_idx' on table 'splits' without CONCURRENTLY acquires a SHARE lock, blocking writes (INSERT, UPDATE, DELETE). Duration depends on table size. Reads are still allowed.

Safe alternative:
  Use CONCURRENTLY to build the index without blocking writes:
     CREATE INDEX CONCURRENTLY splits_published_maturity_timestamp_idx ON splits;

  Note: CONCURRENTLY takes longer and uses more resources, but allows concurrent INSERT, UPDATE, and DELETE operations. The index build may fail if there are deadlocks or unique constraint violations.

  Considerations:
  - Requires more total work and takes longer to complete
  - If it fails, it leaves behind an "invalid" index that should be dropped

  Note: CONCURRENTLY cannot run inside a transaction block.
  Add `-- no-transaction` as the first line of the migration file.
```

guilload (approved):

quickwit/quickwit-compaction/src/planner/compaction_planner.rs
```diff
@@ -151,7 +142,8 @@ impl CompactionPlanner {
         let query = ListSplitsQuery::for_all_indexes()
             .with_split_state(SplitState::Published)
             .retain_immature(OffsetDateTime::now_utc())
-            .with_update_timestamp_gte(self.cursor);
+            .sort_by_maturity_timestamp()
```
nadav-govari, line 145: investigate sending NOT IN in flight merges

PR #6407, opened 2026-05-08
feat: page-bounded Arrow decoder per data page (PR-6a.2)

g-talbot:
## Summary
- Rebuilds the page-stream → Arrow decoder to be **page-bounded** — `StreamDecoder::decode_next_page()` returns one [`DecodedPage`] per call (rg_idx, col_idx, page_idx, row_start, ArrayRef) instead of materialising an entire row group at a time.
- Memory: ~one in-flight page (compressed + decompressed bytes) + one cached dictionary page per (rg, col) when dict-encoded. The decoder does NOT buffer a row group, column chunk, or any materialised array beyond the one currently being emitted.
- PR-6b's merge engine consumes [`DecodedPage`]s in storage order (row-group-major, column-major-within-rg, page-major-within-col), applies merge plan slicing per page, and streams output pages directly into the writer without column-chunk staging.

## How it works
1. Pull one [`Page`] from the underlying [`ColumnPageStream`]. Skip `INDEX_PAGE` (historical Thrift variant, not emitted by production writers).
2. Look up or initialise per-(rg, col) state: a `PageQueue` that feeds parquet-rs's [`ColumnReader`] one page at a time, plus a counter tracking rows decoded so far.
3. Convert the [`Page`] to parquet-rs's `column::page::Page` enum: decompress via [`parquet::compression::create_codec`] (requires the `experimental` feature), translate `format::Encoding` (Thrift wrapper) → `basic::Encoding` (Rust enum) via a manual i32 match (no public conversion in parquet-rs), drop optional statistics.
4. Push the converted page onto the queue. Dictionary/index pages absorb silently for use by subsequent data pages.
5. For a data page: ask the [`ColumnReader`] to decode exactly `header.num_values` records via `read_records(...)` calls in a loop, pulling values + def/rep levels into typed buffers.
6. Build an `ArrayRef` from `(values, def_levels, rep_levels)` per the column's parquet physical type. Emit [`DecodedPage`].

## Type coverage
Flat physical types: `Boolean`, `Int8/16/32` + `UInt8/16/32` (parquet `Int32` with logical annotation), `Int64`/`UInt64` (parquet `Int64`), `Float32`, `Float64`, `Utf8`/`LargeUtf8`/`Binary`/`LargeBinary` (parquet `ByteArray`). Dictionary-encoded pages are decoded via the cached dict page → values pipeline.

`List<T>` / `LargeList<T>` where outer + inner are non-nullable and inner is a flat primitive — covers DDSketch `keys` (`List<Int16>`) and `counts` (`List<UInt64>`). Dremel def/rep levels (max_def=1, max_rep=1) are decoded via the same `read_records` path; arrow offsets are computed via `list_offsets_from_levels`.

Other nested shapes (nullable list inner/outer, `Struct`, `Map`, `FixedSizeList`, multi-leaf nested) return an unsupported-type error rather than silently falling back to a different mechanism.

## Sync ⇄ async bridging
The page stream is async (S3 reads); `PageReader` (from parquet-rs) is sync. Bridged via `Arc<Mutex<VecDeque<ColumnPage>>>` per (rg, col): `decode_next_page` pulls from the stream (async), pushes onto the queue, then the sync `PageReader` impl pops from the queue when the `ColumnReader` asks for the next page. `peek_next_page` and `skip_next_page` are properly implemented to support the parquet-rs reader's state machine.

## Schema handling
`parquet_to_arrow_schema(parquet_schema, None)` bypasses the ARROW:schema hint that would otherwise force Dictionary types — input parquet files written from arrow declare Dictionary columns in ARROW:schema metadata, but their page-encoded values are plain values that decode to `StringArray`/etc. Decoding without the hint gives consistent flat-primitive output that the merge engine then interleaves.

## Tests
9 tests, all passing:
- `test_drain_single_rg_round_trip`, `test_drain_multi_rg_round_trip` — full round-trip via `decode_next_page` matches `ParquetRecordBatchReaderBuilder`.
- `test_decoded_page_row_indexing` — `row_start` correctly tracks per-(rg, col) row offsets.
- `test_eof_idempotent` — repeated calls after EOF stay `Ok(None)`.
- `test_nullable_column_round_trip` — def-level decoding for nullable cols.
- `test_compression_codecs` — snap, gzip, zstd round-trip.
- `test_page_bounded_queue_depth` — verifies the internal page queue depth stays ≤ 2 across a long stream (the page-bounded contract).
- `test_list_uint64_round_trip` — `List<UInt64>` (DDSketch shape) round-trip.
- `test_io_failure_surfaces_as_page_stream_error` — body GET failures propagate as `PageStream(Io)`, not masked as decode errors.

## Stack
Base: `gtt/column-page-stream-trait` (PR-5a #6406).

PR-6b (#6409) builds the streaming merge engine on top of this decoder.


quickwit/quickwit-parquet-engine/src/storage/page_decoder.rs
```diff
@@ -0,0 +1,1889 @@
...
+    }
+
+    fn peek_next_page(
+        &mut self,
+    ) -> parquet::errors::Result<Option<parquet::column::page::PageMetadata>> {
+        // Used by the rep-level decoder (`at_record_boundary`) to know
+        // whether the next page begins a new record. We build the
+        // metadata directly from the front of our queue.
+        let guard = self.queue.lock().expect("PageQueue mutex poisoned");
+        Ok(guard.front().map(page_metadata_from_column_page))
```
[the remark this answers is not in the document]
g-talbot: Fixed in 9e747923d, refined in 09c68870f.

## Initial fix (9e747923d)

`decode_next_page` now maintains a **one-page lookahead**: after locating a state with a queued data page to decode, it pulls one more page from the underlying stream and routes it to its (rg, col) queue *before* calling `read_records`.

- Lookahead routes to the **same** (rg, col) → that state's queue holds [current_data_page, lookahead_data_page], so `peek_next_page` returns the real next-page metadata when the column reader checks `at_record_boundary` for V1 record continuation. parquet-rs correctly extends record decoding across the page boundary.
- Lookahead routes to a **different** (rg, col) → the current state's `peek_next_page` correctly returns `None` because this column chunk's pages are now exhausted from the current state's perspective.

The loop is restructured to first check for any state with an unconsumed queued data page (covers both the freshly-pulled case and a previous call's lookahead landing in a different state's queue), and only pulls from the stream when no state has work to do.

Regression test `test_list_record_spanning_pages_preserved` writes a `List<UInt64>` column with `data_page_size_limit = 20` bytes (forces a 50-element list record to span several V1 pages) and verifies all 50 values are preserved intact after decoding.

## Refinement (09c68870f)

The lookahead is now **conditional on `max_rep_level > 0`** (i.e., it only activates for `List<T>` / `LargeList<T>` columns).

Reason: an unconditional pre-fetch advances the underlying stream past the current column chunk. That broke callers that drop a `StreamDecoder` mid-traversal and create a fresh one over the same `ColumnPageStream` — the merge engine's phase 0 (sort col drain) followed by a phase 3 fresh body-col decoder is exactly this pattern, and the pre-fetched page was being lost. Flat columns get one value per record, so V1 record-spanning across pages doesn't apply to them — restricting the lookahead to repeated columns gives both the correctness fix and the cross-decoder safety.

`test_list_record_spanning_pages_preserved` still exercises the lookahead path (the test column has `max_rep_level = 1`). The downstream merge engine tests on PR-6b.2 (#6409) and PR-6c.2 (#6410) now pass after the rebase.

Memory: still page-bounded; `test_page_bounded_queue_depth` continues to pass.

quickwit/quickwit-parquet-engine/src/storage/page_decoder.rs
```diff
@@ -0,0 +1,2055 @@
...
+    fn find_state_with_queued_data_page(&self) -> Option<(usize, usize)> {
+        for (&key, state) in self.columns.iter() {
+            let q = state.queue.lock().expect("PageQueue mutex poisoned");
+            if q.iter().any(|p| {
+                matches!(
+                    p,
+                    ColumnPage::DataPage { .. } | ColumnPage::DataPageV2 { .. }
+                )
+            }) {
+                return Some(key);
```
adamtobey: This pattern of lock, find a queue to pull from, unlock, then have someone try to pull from that queue feels suspicious to me. If this queue has only one consumer then I guess it's sound. Is there any case where decode_next_page returns None because of that?

quickwit/quickwit-parquet-engine/src/storage/page_decoder.rs
```diff
@@ -0,0 +1,2055 @@
...
+            }
+        }
+
+        // Read num_values from the front data page (skipping any
+        // dictionary pages that may precede it in the queue).
+        let state = self
+            .columns
+            .get(&key)
+            .expect("caller's key must have a state");
+        let num_values_in_page = {
```
adamtobey: Another place that makes me suspicious of TOCTOU

quickwit/quickwit-parquet-engine/src/storage/page_decoder.rs
```diff
@@ -0,0 +1,2055 @@
...
+            } else {
+                full.push(None);
+            }
+        }
+        BooleanArray::from(full)
+    };
+    Ok(Arc::new(arr))
+}
+
+fn build_int32_array(
```
adamtobey, line 853: Are there no existing functions for this kind of thing?

adamtobey (commented):
@g-talbot before I approve I'm curious for your thoughts on my TOCTOU comments. I'm guessing this all works in careful harmony, but I don't love those bits in isolation

quickwit/quickwit-parquet-engine/src/storage/page_decoder.rs
g-talbot: Good instinct — the pattern looked suspicious because the `Arc<Mutex<_>>` *suggests* concurrency that doesn't actually exist. Addressed in 5bd2a8074 two ways:

1. **Document the invariant.** Added a "Single-consumer invariant" section to `StreamDecoder`'s docstring: `decode_next_page` takes `&mut self`, so only one task ever observes `self.columns`. The `Arc<Mutex<_>>` is there because parquet-rs's `PageReader: Send` trait shape requires interior mutability + `Send` (the queue handle has to live both here, where we push, *and* inside the boxed `dyn PageReader` the `ColumnReader` pops from), not because of cross-thread concurrency. All pushes happen synchronously from `decode_next_page`; all pops happen synchronously from inside `decode_one_data_page_into_array` driven by the same call. No lock/unlock/lock sequence can race.

2. **Make the invariant visible at the call site.** Replaced `find_state_with_queued_data_page` (which returned just the key, forcing a second lock acquisition downstream to read `num_values`) with `next_decodable_head`, which returns `((rg, col), num_values)` together from one lock pass. The lock-find-unlock-relock-refind shape is gone — there's just one peek now.

To your direct question: no, `decode_next_page` never returned `None` because of this. The single-consumer guarantee held; the code just didn't *look* like it did at the call site.

quickwit/quickwit-parquet-engine/src/storage/page_decoder.rs
g-talbot: Same single-consumer invariant covers this site — addressed together in 5bd2a8074. The old "lock-find-num_values / drop-lock / `get_mut`-state / decode" sequence collapsed into `next_decodable_head` returning `(key, num_values)` together (one lock pass), then `decode_head(key, num_values_in_page)` does the work without re-locking. The new `StreamDecoder` docstring makes the "why this can't race" reasoning explicit; see reply on the line-249 comment for the full argument.

quickwit/quickwit-parquet-engine/src/storage/page_decoder.rs
g-talbot: Good catch — there are. `parquet::arrow::array_reader::PrimitiveArrayReader::consume_batch` plus the `coerce_i32` / `coerce_i64` widening helpers do exactly this; `make_byte_array_reader` and `ListArrayReader` cover the byte and list branches. The reason we roll our own:

1. **`array_reader` is gated behind parquet-rs's `experimental` Cargo feature**, which we don't enable. The module is `#[doc(hidden)] pub mod` when the feature is on, and the module doc says explicitly: *"This is an internal implementation detail of the Parquet reader, and is not intended for public use."* parquet-rs reserves the right to break it across versions.

2. **Granularity is fine either way.** `PrimitiveArrayReader::read_records` is page-iterator-driven and would compose with our `PageQueueReader`, so the swap would work mechanically. The cost is "depend on an unstable API," not "doesn't fit our model."

Tradeoff as I see it: ~250 lines of stable-API code vs. a smaller wrapper over an unstable one. Given how mechanical the conversion is (def_levels → null buffer → typed array + numeric widening), I lean toward keeping ours. If parquet-rs ever stabilizes `array_reader`, this section becomes a thin adapter.

Two changes in 5bd2a8074 to make the lineage explicit and cut the duplication:

- **Factored out `build_primitive::<P, T, F>(...)`** — a generic helper that mirrors parquet-rs's `coerce_i32` / `coerce_i64` table. Each arm of `build_int32_array` / `build_int64_array` is now one line; `build_float32_array` / `build_float64_array` also use it. Removes the repeated null-clone + cast + typed-array-constructor pattern.

- **Module comment above the `build_*` family** names the upstream equivalents (`PrimitiveArrayReader::consume_batch`, `coerce_i32` / `coerce_i64`, `make_byte_array_reader`, `ListArrayReader`) and explains why we re-implement. Future readers won't have to re-derive this.

adamtobey (approved):

PR #6409, opened 2026-05-08
feat: streaming column-major merge engine with page-bounded body cols (PR-6b.2)

g-talbot:
## Summary
- Streaming column-major merge engine on top of PR-6a.2's per-page Arrow decoder. **Body col memory is bounded by output page size**, not column-chunk size.
- Preserves caller-specified M:N output splitting at `sorted_series` boundaries — the existing `compute_output_boundaries` logic is unchanged; the merge engine just runs its column writes streamingly.
- Single-RG inputs only in this PR (real or PR-5-adapter-presented). Multi-RG metric-aligned input streaming lands in PR-6c.2 alongside the per-merge-region restructure.

## Architecture (Husky multi-input → multi-output sorted merge)

**Phase 0 (async)** — drain sort cols + `sorted_series` from each input via `StreamDecoder::decode_next_page()`. Husky column ordering puts sort cols first within each row group's body bytes, so the decoder stops after they're fully decoded; body col pages stay un-read in the stream for phase 3.

**Phase 1** — `compute_merge_order` over per-input sort-col `RecordBatch`es using the existing k-way `(sorted_series, timestamp_secs)` heap.

**Phase 2** — `compute_output_boundaries` with the caller's `num_outputs`, splitting at sorted_series transitions. Same algorithm as the non-streaming engine.

**Phase 3 (blocking + block_on bridges)** — streaming write. All M output writers are alive concurrently inside one `tokio::task::spawn_blocking`. For each column in Husky order, every output's col K is written in turn via `write_next_column_arrays`:

- **Sort col / sorted_series**: applied via `arrow::compute::interleave` from the already-buffered phase-0 data, split into `OUTPUT_PAGE_ROWS`-sized chunks → page-sized arrays.
- **Body col**: each output page is assembled via `arrow::compute::interleave` from input page slices. The sync iterator passed to `write_next_column_arrays` drives the async decoder via `tokio::runtime::Handle::block_on(decode_next_page())` from inside the spawn_blocking task. Pages flush to the writer's sink as `SerializedColumnWriter`'s page-size threshold trips — memory stays bounded by the in-flight output page + a few in-flight input pages per input.

After all M outputs' col K is done, every input decoder is at the start of col K+1 in its single row group. Move to col K+1.

## Why async/tokio
`ColumnPageStream` is async (S3 reads). The parquet writer is sync. Single `spawn_blocking` + `Handle::block_on` from sync iterators is cleaner than per-output tokio tasks: the writers + borrowed `RowGroupBuilder`s all live in one scope, lifetimes track properly, and merge work is mostly CPU-bound so parallelising outputs across threads wouldn't help much.

## Single-RG-input simplification
PR-6b.2 rejects multi-RG inputs with a clear error message. Rationale: with Husky ordering, each RG's body bytes look like `[sort_cols, body_cols]`. For multi-RG inputs the layout is `[sort_cols_RG0, body_cols_RG0, sort_cols_RG1, body_cols_RG1, ...]`. To collect sort cols across all RGs in phase 0 without buffering RG[i-1]'s body cols would require a second body GET (sparse range request for sort col bytes only), which violates the "two GETs per input" contract.

The fix lives in **PR-6c.2** as the per-merge-region restructure: drain sort cols of ONE input RG at a time (= one merge region), produce output for that region, then advance to the next RG. The sort-prefix alignment (`prefix_len ≥ 1`) guarantees that one merge region = at most one RG per input, so the per-region phase 0 never needs to span RGs.

Today's real inputs satisfy the single-RG assumption: post-PR-3 ingest splits will be single-RG by design, and legacy multi-RG splits route through **PR-5's `LegacyMultiRGAdapter`** (#6408) which buffers the whole file and presents it as a synthetic single-RG stream.

## M:N preserved
Caller's `num_outputs` config is honoured. The engine splits at `sorted_series` transitions (the existing `compute_output_boundaries` algorithm), which can occur mid-input-RG since they're at row-key boundaries. The "don't let the file get too big" file-size cap (which adds *additional* splits beyond the caller's N at any sort-key transition) lands in PR-6c.2.

## Tests
9 tests, all passing:
- `test_two_inputs_simple_merge` — 2 inputs → 1 output, sort order preserved.
- `test_output_is_single_row_group` — single-metric_name input + num_outputs=1 → single output RG.
- `test_total_rows_preserved` — N inputs → M outputs, MC-1 invariant.
- `test_sort_schema_mismatch_rejected` — incompatible sort schemas error.
- `test_kv_metadata_propagated_to_output` — `qh.*` KVs propagate, `num_merge_ops` increments.
- `test_all_empty_inputs_no_output` — zero-row inputs → empty output set.
- `test_output_drainable_by_stream_decoder` — output round-trips through PR-6a.2's decoder.
- `test_multi_rg_input_rejected` — multi-RG input rejected with clear error.
- `test_body_col_streams_many_pages_per_column_chunk` — **page-bounded contract sanity**: with `data_page_row_count_limit=1000` on an 8000-row merge, output value col spans ≥ 2 pages.

441/441 crate tests pass; clippy clean; doc, machete, license, fmt all clean.

## Helpers exposed
Existing helpers in `merge/writer.rs` are now `pub(super)` (`apply_merge_permutation`, `build_merge_kv_metadata`, `build_sorting_columns`, `resolve_sort_field_names`, `verify_sort_order`) so `streaming.rs` reuses the same MC-3 / KV / sorting-columns construction the non-streaming engine uses. PR-7 will fold the non-streaming engine away.

## Stack
Base: `gtt/parquet-page-decoder` (PR-6a.2 #6407).

PR-6c.2 (#6410) extends with: per-merge-region restructure → unlocks multi-RG metric-aligned input support + multi-RG output at metric_name boundaries + file-size cap with sort-key-boundary splits.

### DIFF
--- a/quickwit/quickwit-parquet-engine/src/merge/streaming.rs
+++ b/quickwit/quickwit-parquet-engine/src/merge/streaming.rs
@@ -1,0 +1,2235 @@
+// Copyright 2021-Present Datadog, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Streaming column-major merge engine with page-bounded body cols.
+//!
+//! Architecture (Husky multi-input → multi-output sorted merge):
+//!
+//! 1. **Phase 0 (async): drain sort cols** from each input. With Husky column ordering, sort cols +
+//!    `sorted_series` are the prefix of each row group's body bytes, so we can stop the decoder
+//!    after those are fully decoded. The remaining body col pages stay un-read in the input stream,
+//!    ready for phase 3.
+//! 2. **Phase 1: compute merge order** via the existing k-way merge on `(sorted_series,
+//!    timestamp_secs)` from the per-input sort col [`RecordBatch`]es. Produces a run-length-encoded
+//!    merge plan over input row positions.
+//! 3. **Phase 2: compute output boundaries** with the caller's `num_outputs`, splitting at
+//!    `sorted_series` transitions so each output file's key range is non-overlapping with adjacent
+//!    files.
+//! 4. **Phase 3 (blocking + block_on bridges): streaming write**. All output writers are alive for
+//!    the duration. For each column in Husky order, every output's col K is written in turn:
+//!    - Sort col / `sorted_series`: applied via `take` from the already-buffered phase 0 data.
+//!    - Body col: each output page is assembled via [`arrow::compute::interleave`] from input page
+//!      slices, with decoders advanced page-by-page via `Handle::block_on` from inside a sync
+//!      iterator. Pages flush to the writer's sink as [`SerializedColumnWriter`]'s page-size
+//!      threshold trips — memory stays bounded by the in-flight output page plus a small number of
+//!      in-flight input pages.
+//!
+//! After all M outputs' col K is done, every input decoder is at the
+//! start of col K+1 in its single row group. Move to col K+1.
+//!
+//! ## Single-RG inputs assumption
+//!
+//! PR-6b.2 only handles **single-row-group inputs**. With multi-RG
+//! inputs the body bytes interleave with successive RGs' sort cols
+//! (`sort_cols_RG0`, `body_cols_RG0`, `sort_cols_RG1`, ...), so
+//! draining sort cols from RG1 onwards requires either consuming +
+//! discarding body cols of RG0 from the stream or buffering them.
+//! Neither fits the page-bounded contract; multi-RG-input streaming
+//! lands in a follow-up. Today's real inputs are: (a) post-PR-3
+//! single-RG ingest splits, or (b) PR-5's legacy adapter that
+//! presents arbitrary multi-RG splits as one synthetic RG. Both
+//! satisfy the assumption.
+//!
+//! [`SerializedColumnWriter`]: parquet::file::writer::SerializedColumnWriter
+
+#![allow(dead_code)]
+
+use std::collections::{HashMap, HashSet};
+use std::ops::Range;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+use anyhow::{Context, Result, anyhow, bail};
+use arrow::array::{Array, ArrayRef, RecordBatch, new_null_array};
+use arrow::compute::interleave;
+use arrow::datatypes::{DataType, Field, Schema as ArrowSchema, SchemaRef};
+use parquet::file::metadata::ParquetMetaData;
+use tokio::runtime::Handle;
+use tracing::info;
+use ulid::Ulid;
+
+use super::merge_order::{MergeRun, compute_merge_order, compute_output_boundaries};
+use super::schema::{align_inputs_to_union_schema, optimize_output_batch};
+use super::writer::{
+    apply_merge_permutation, build_merge_kv_metadata, build_sorting_columns,
+    resolve_sort_field_names, verify_sort_order,
+};
+use super::{InputMetadata, MergeConfig, MergeOutputFile};
+use crate::row_keys;
+use crate::sort_fields::{
+    equivalent_schemas_for_compaction, is_timestamp_column_name, parse_sort_fields,
+};
+use crate::sorted_series::SORTED_SERIES_COLUMN;
+use crate::split::TAG_SERVICE;
+use crate::storage::page_decoder::{DecodedPage, StreamDecoder};
+use crate::storage::split_writer::{extract_metric_names, extract_time_range};
+use crate::storage::streaming_writer::StreamingParquetWriter;
+use crate::storage::{
+    ColumnPageStream, PARQUET_META_NUM_MERGE_OPS, PARQUET_META_RG_PARTITION_PREFIX_LEN,
+    PARQUET_META_SORT_FIELDS, PARQUET_META_WINDOW_DURATION, PARQUET_META_WINDOW_START,
+};
+use crate::zonemap::{self, ZonemapOptions};
+
+/// Output page size in rows for body-col assembly. Each call to the
+/// sync iterator passed to [`write_next_column_arrays`] yields one
+/// `ArrayRef` of up to this many rows; the parquet writer flushes
+/// physical pages independently as encoded bytes cross
+/// `data_page_size_limit`. 1024 keeps assembled arrays small enough
+/// to bound per-output memory but large enough to amortise per-page
+/// fixed costs.
+///
+/// [`write_next_column_arrays`]: crate::storage::streaming_writer::RowGroupBuilder::write_next_column_arrays
+const OUTPUT_PAGE_ROWS: usize = 1024;
+
+/// Streaming N-input → M-output column-major merge.
+///
+/// See module docs for the four phases. Returns one
+/// [`MergeOutputFile`] per output file produced (zero-row outputs are
+/// dropped). Caller's `config.num_outputs` is the upper bound on the
+/// number of files; fewer are returned when there are not enough
+/// `sorted_series` transitions to split at.
+pub async fn streaming_merge_sorted_parquet_files(
+    inputs: Vec<Box<dyn ColumnPageStream>>,
+    output_dir: &Path,
+    config: &MergeConfig,
+) -> Result<Vec<MergeOutputFile>> {
+    if inputs.is_empty() {
+        bail!("merge requires at least one input");
+    }
+    if config.num_outputs == 0 {
+        bail!("num_outputs must be at least 1");
+    }
+
+    // Validate that all inputs are single-RG (or zero-RG, which means
+    // the file has no data). PR-6b.2 simplification — see module docs.
+    for (idx, stream) in inputs.iter().enumerate() {
+        let num_rgs = stream.metadata().num_row_groups();
+        if num_rgs > 1 {
+            bail!(
+                "streaming merge requires single-row-group inputs in PR-6b.2 (input {idx} has \
+                 {num_rgs} row groups); multi-RG metric-aligned inputs land in a follow-up. \
+                 Legacy multi-RG (rg_partition_prefix_len=0) inputs must go through the PR-5 \
+                 adapter, which presents them as a single synthetic row group."
+            );
+        }
+    }
+
+    let input_meta = extract_and_validate_input_metadata(&inputs)?;
+
+    info!(
+        num_inputs = inputs.len(),
+        num_outputs = config.num_outputs,
+        sort_fields = %input_meta.sort_fields,
+        "starting streaming sorted parquet merge"
+    );
+
+    let output_dir = output_dir.to_path_buf();
+    let writer_config = config.writer_config.clone();
+    let num_outputs = config.num_outputs;
+
+    // Move everything onto a blocking task. Inside, the decoders need to
+    // make async I/O calls (page fetches over the network); we drive
+    // those via `handle.block_on(...)` from inside sync iterators that
+    // feed the parquet writer's column-write methods. The writer is
+    // sync; this single-task pattern avoids the lifetime complexity of
+    // moving borrowed `RowGroupBuilder`s across tokio tasks.
+    let result = tokio::task::spawn_blocking(move || -> Result<Vec<MergeOutputFile>> {
+        let handle = Handle::current();
+
+        let mut inputs = inputs;
+        let mut decoders_state = build_input_decoders_state(&mut inputs)?;
+
+        // Phase 0
+        let sort_col_batches =
+            drain_sort_cols_all_inputs(&handle, &mut decoders_state, &input_meta.sort_fields)?;
+
+        if sort_col_batches.iter().all(|b| b.num_rows() == 0) {
+            info!("all inputs empty, producing no output");
+            return Ok(Vec::new());
+        }
+
+        // Phase 1: align inputs to a union sort-col schema so the merge-order
+        // comparator sees uniformly-typed `sorted_series` + `timestamp_secs`.
+        let (sort_union_schema, aligned_sort_batches) =
+            align_inputs_to_union_schema(&sort_col_batches, &input_meta.sort_fields)?;
+        let merge_order = compute_merge_order(&aligned_sort_batches, &input_meta.sort_fields)?;
+
+        // Phase 2: split merge order into M outputs at sorted_series boundaries.
+        let boundaries =
+            compute_output_boundaries(&merge_order, &aligned_sort_batches, num_outputs)?;
+
+        let total_rows: usize = aligned_sort_batches.iter().map(|b| b.num_rows()).sum();
+        info!(
+            total_rows,
+            num_outputs = boundaries.len(),
+            "streaming merge order computed"
+        );
+
+        // Pre-compute per-input row → (output_idx, output_position) destination map.
+        // Used by every column write to slice take/interleave indices per page.
+        let destinations =
+            build_input_row_destinations(&aligned_sort_batches, &merge_order, &boundaries);
+
+        // Phase 3
+        let outputs = write_streaming_outputs(
+            &handle,
+            &mut decoders_state,
+            &aligned_sort_batches,
+            &sort_union_schema,
+            &merge_order,
+            &boundaries,
+            &destinations,
+            &input_meta,
+            &writer_config,
+            &output_dir,
+        )?;
+
+        // MC-1: total row count preserved.
+        let output_total: usize = outputs.iter().map(|o| o.num_rows).sum();
+        quickwit_dst::check_invariant!(
+            quickwit_dst::invariants::InvariantId::MC1,
+            output_total == total_rows,
+            ": streaming merge input rows={}, output rows={}",
+            total_rows,
+            output_total,
+        );
+
+        Ok(outputs)
+    })
+    .await
+    .context("streaming merge blocking task panicked")??;
+
+    Ok(result)
+}
+
+/// Per-input state held across phase 0 → phase 3 inside the blocking
+/// task. The decoder borrows mutably from its input stream; both live
+/// here so the borrow checker is happy with one struct per input.
+struct InputDecoderState {
+    /// Owned stream (so the decoder's `&mut dyn ColumnPageStream` borrow
+    /// is anchored to a stable address inside this struct).
+    stream: Box<dyn ColumnPageStream>,
+    metadata: Arc<ParquetMetaData>,
+    /// Arrow schema of this input (from parquet → arrow conversion).
+    arrow_schema: SchemaRef,
+}
+
+/// Build per-input state. The streams are moved in from the caller.
+fn build_input_decoders_state(
+    inputs: &mut Vec<Box<dyn ColumnPageStream>>,
+) -> Result<Vec<InputDecoderState>> {
+    let mut states = Vec::with_capacity(inputs.len());
+    for stream in inputs.drain(..) {
+        let metadata = Arc::clone(stream.metadata());
+        let parquet_schema = metadata.file_metadata().schema_descr();
+        let arrow_schema = parquet::arrow::parquet_to_arrow_schema(parquet_schema, None)
+            .context("converting parquet schema → arrow")?;
+        states.push(InputDecoderState {
+            stream,
+            metadata,
+            arrow_schema: Arc::new(arrow_schema),
+        });
+    }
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-parquet-engine/src/merge/streaming.rs
```diff
@@ -0,0 +1,2235 @@
...
+) -> Result<()> {
+    // If cache already covers target_row, nothing to do.
+    if let Some(last) = page_cache.last() {
+        let last_end = last.row_start + last.array.len();
+        if target_row < last_end {
+            return Ok(());
+        }
+    }
+
+    let mut decoder = StreamDecoder::new(&mut *state.stream);
```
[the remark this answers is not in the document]
g-talbot: Confirmed and fixed in 81f7414. You were right on both counts: the per-call `StreamDecoder::new(&mut *state.stream)` was resetting per-column `rows_decoded` so subsequent pages came back with `row_start = 0`, and the per-assembler `page_cache` was dropping pages whose row range straddled an output boundary even though the stream couldn't be rewound.

The decoder now lives on `InputDecoderState` via a new `StreamDecoder::from_owned(Box<dyn ColumnPageStream>)` constructor, so the same decoder threads phase 0 → phase 3 for an input and preserves all per-(rg, col) state. The body-col page cache + cursor moved onto `InputDecoderState` too, with a `reset_body_col_state()` call at the start of each body column (the only place where leftover pages would belong to a different column). Cross-output reuse just works because the state outlives each output's assembler.

Two regression tests now exercise both shapes:

- `test_body_col_multi_input_page_preserves_row_start` — 2500 rows × 50-row input pages, single output, so `OUTPUT_PAGE_ROWS=1024` forces multiple output pages per col → multiple `advance_decoder_to_row` calls. Asserts every value lands at the right row.
- `test_body_col_cache_persists_across_outputs` — two metric names → two outputs, 400 rows with 50-row input pages so some input pages straddle the output boundary. Asserts the full set of input values survives across both output files.

Follow-up commit on PR-6c.2 ports the same change to the per-merge-region path (with the per-(region, col) reset placed inside `process_region`).

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
```diff
@@ -0,0 +1,2235 @@
...
+
+        // Track service names while streaming the service col.
+        let track_service = col_name == "service";
+
+        // Drive the assembler via a sync iterator. We must `?`-propagate
+        // assembly errors out of the iterator; collect into a Vec and
+        // return on first error.
+        let pages: Vec<ArrayRef> = assembler
+            .into_iter()
+            .collect::<Result<Vec<_>>>()
```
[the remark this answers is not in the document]
g-talbot: Fixed in 81f7414. The `.collect::<Result<Vec<_>>>()` is gone — `write_next_column_arrays` now consumes a `StreamingBodyColIter` adapter that yields one `ArrayRef` per `next()` call and stashes the first assembly error in a side cell (`error_slot: &mut Option<anyhow::Error>`). After the writer returns, the caller checks the slot first and surfaces the assembly error with full context before the write-side error (which is typically a downstream symptom of the assembler bailing early).

The service-name collection moved into the same adapter via an optional `&mut HashSet<String>`, so it stays in lockstep with the streamed pages instead of running over the whole column afterwards. Memory per body col is now bounded by `OUTPUT_PAGE_ROWS` × per-input page count, not column-chunk size.

Same change applied to PR-6c.2's per-region path.

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
```diff
@@ -0,0 +1,2547 @@
...
+    // Find this col's per-input parquet leaf index (one per input).
+    // Inputs whose schema doesn't have this col contribute null rows
+    // and don't advance their decoder for this col.
+    let mut input_col_indices: Vec<Option<usize>> = Vec::with_capacity(decoders_state.len());
+    let mut input_target_rows: Vec<usize> = Vec::with_capacity(decoders_state.len());
+    for state in decoders_state.iter() {
+        match state.arrow_schema.index_of(col_name) {
+            Ok(idx) => {
+                input_col_indices.push(Some(idx));
+                let rg = state.metadata.row_group(0);
```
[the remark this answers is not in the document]
g-talbot: Confirmed and fixed in 0512794. Reproduced your scenario locally — merging a 0-row + 50-row pair panicked inside parquet-rs at `metadata/mod.rs:248` with `index out of bounds: the len is 0 but the index is 0`, exactly where you predicted.

The body-col loop now checks `state.metadata.num_row_groups() == 0` up front and pushes `None` into `input_col_indices` for empty inputs (same handling as inputs whose schema lacks the column). Also dropped the unused `input_target_rows` vec that was only there as a side effect of the panicking lookup.

Regression test `test_zero_row_input_mixed_with_non_empty` constructs a 0-row + 50-row pair and asserts the merge produces one output with all 50 populated values in input row order. Verified the test fails against the pre-fix code with the same parquet-rs panic before passing on the fixed code.

Same test ported to PR-6c.2 — its per-region engine only iterates `region.contributing` for body cols (so empty inputs are naturally skipped), but the explicit test guards against future iteration broadening.

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
```diff
@@ -0,0 +1,2589 @@
...
+) -> Result<SchemaRef> {
+    let sort_schema = sort_optimised.schema();
+    let mut fields: Vec<Arc<Field>> = Vec::with_capacity(full_union_schema.fields().len());
+    for field in full_union_schema.fields() {
+        // If sort_optimised dropped this field entirely, the field was
+        // all-null in this output's sort cols — drop from output too.
+        // (Body cols are always retained.)
+        let is_sort_field = sort_schema.index_of(field.name()).is_ok();
+        if is_sort_field || full_union_schema.index_of(field.name()).is_ok() {
+            fields.push(Arc::clone(field));
```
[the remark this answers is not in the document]
g-talbot: Fixed in 1b80d47. You diagnosed this precisely — `sort_optimised.has(name) || full_union.has(name)` was tautologically true for every iterated `field` because `field` came from `full_union_schema.fields()`, so the intended all-null-sort-field drop never fired.

The fix takes the sort union schema as an explicit parameter so we can tell sort fields apart from body fields:

```rust
fn derive_output_schema(
    full_union_schema: &SchemaRef,
    sort_union_schema: &SchemaRef,
    sort_optimised: &RecordBatch,
) -> Result<SchemaRef> {
    for field in full_union_schema.fields() {
        let is_sort_field = sort_union_schema.index_of(field.name()).is_ok();
        if is_sort_field {
            // Sort field: keep only if optimize_output_batch kept it.
            if sort_optimised.schema().index_of(field.name()).is_ok() {
                fields.push(Arc::clone(field));
            }
        } else {
            // Body field: always kept.
            fields.push(Arc::clone(field));
        }
    }
    ...
}
```

Body fields are kept unconditionally on purpose. Per-output body-col presence tracking would require pre-reading every body column for every output, which is the column-chunk-bounded buffering the streaming path exists to avoid. (Your point about non-streaming `optimize_output_batch` stripping all-null body cols too is correct as a comparison; we accept that regression for PR-6b.2 / PR-6c.2 as documented in the function comment.)

Regression test `test_derive_output_schema_drops_all_null_sort_field` calls the helper directly with a synthetic union + sort-optimised pair (full union has `metric_name`, `env`, `timestamp_secs`, `value`; sort_optimised dropped `env`; sort_union contains the three sort fields). Asserts the derived per-output schema is `[metric_name, timestamp_secs, value]` — `env` dropped, body field `value` preserved. Verified the test fails against the pre-fix logic with `[metric_name, env, timestamp_secs, value]` vs `[metric_name, timestamp_secs, value]` mismatch.

PR-6c.2 uses a different schema architecture (one global union schema per output file, no per-region optimisation), so this specific bug doesn't reproduce there; rebased it on top of the PR-6b.2 fix without code changes.

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
```diff
@@ -0,0 +1,3634 @@
...
+    // sync; this single-task pattern avoids the lifetime complexity of
+    // moving borrowed `RowGroupBuilder`s across tokio tasks.
+    let result = tokio::task::spawn_blocking(move || -> Result<Vec<MergeOutputFile>> {
+        let handle = Handle::current();
+
+        let mut inputs = inputs;
+        let mut decoders_state = build_input_decoders_state(&mut inputs)?;
+
+        // Phase 0
+        let sort_col_batches =
```
adamtobey, line 199: To understand: we do have a hard requirement on column ordering across all files? Sort cols must be first, then all other cols must be in consistent order?

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
```diff
@@ -0,0 +1,3634 @@
...
+        Some(s) => s,
+        None => bail!("at least one input is required"),
+    };
+
+    Ok(InputMetadata {
+        sort_fields,
+        window_start_secs: consensus_window_start.unwrap_or(None),
+        window_duration_secs: consensus_window_duration.unwrap_or(0),
+        num_merge_ops: max_merge_ops + 1,
+        rg_partition_prefix_len: consensus_prefix_len.unwrap_or(0),
```
adamtobey, line 469: Are we considering these truly optional? Or should we enforce that we do have them and do output them?

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
```diff
@@ -0,0 +1,3634 @@
...
+        };
+
+        if !sort_col_parquet_indices.contains_key(&page.col_idx) {
+            bail!(
+                "input {input_idx} returned a non-sort page (col {}) before all sort cols were \
+                 drained — sort-cols-first storage ordering violated",
+                page.col_idx,
+            );
+        }
+        if page.rg_idx != 0 {
```
adamtobey, line 585: I'm guessing this will be relaxed later if we can support metric-name rowgroups that have consistent ordering across all input files?

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
```diff
@@ -0,0 +1,3634 @@
...
+        .collect();
+    let mut rows_per_output: Vec<usize> = vec![0; boundaries.len()];
+
+    for (out_idx, boundary) in boundaries.iter().enumerate() {
+        let runs = &merge_order[boundary.clone()];
+        for run in runs {
+            for r in 0..run.row_count {
+                let input_row = run.start_row + r;
+                per_input[run.input_index][input_row] = Some((out_idx, rows_per_output[out_idx]));
+                rows_per_output[out_idx] += 1;
```
adamtobey: nit: rows_for_current_output could be a register then rows_per_output[out_idx] = rows_for_current_output after the loop?

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
```diff
@@ -0,0 +1,3634 @@
...
+
+/// Compute the per-output schema. For PR-6b.2 we use the
+/// (string-normalised) union schema as the output schema directly —
+/// fields stay Utf8/LargeUtf8 rather than being re-dict-encoded.
+/// Reason: streaming-decoded input arrays come out of the page
+/// decoder as plain `StringArray`/`BinaryArray` (not Dictionary), and
+/// dict re-encoding per output page would add a per-page CPU cost we
+/// don't want to take in the page-bounded path. Re-introducing
+/// dict-encoded output strings can be done later by tracking
+/// cardinality during the streaming pass — call site is here.
```
adamtobey, line 1028: Reasonable TODO. If we know things are dictionary encoded we can optimize for merging dictionaries and similar cases, better than un-encode -> interleave -> re-encode

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
```diff
@@ -0,0 +1,3634 @@
...
+        let col_parquet_idx = match s.input_col_indices[input_idx] {
+            Some(c) => c,
+            None => {
+                // This input lacks this col entirely — null contributions.
+                // We'll handle null-filling in the interleave step below.
+                continue;
+            }
+        };
+        let max_needed_row = *input_rows.iter().max().expect("non-empty");
+        advance_decoder_to_row(
```
adamtobey: If I understand this function correctly I think something like `fill_page_cache_to_row` would be a clearer name. My initial read on this was "why are we advancing the decoder without reading any result? Is there a case where we're skipping data?" but that's not what the function does as I understand it

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
```diff
@@ -0,0 +1,3634 @@
...
+                let pages = &s.decoders_state[input_idx].body_col_page_cache;
+                if pages.is_empty() {
+                    // No pages decoded for this input (no rows from this input go to this output).
+                    // Use a zero-row placeholder; we won't index into it.
+                    input_array_refs.push(new_null_array(s.out_field.data_type(), 0));
+                    input_cache_starts.push(0);
+                } else {
+                    let cache_start = pages[0].row_start;
+                    let arrays: Vec<&dyn Array> = pages.iter().map(|p| p.array.as_ref()).collect();
+                    let concatenated = arrow::compute::concat(&arrays).with_context(|| {
```
adamtobey, line 1758: It's a little hard to follow all of this, but is it correct to say that `pages` here is just double buffering the pages on the stream as it's read? It's not going to actually end up filling with all the pages in the column?

adamtobey: OK actually it seems like the page cache will hold all the input pages for this column that contribute to the current output, dropped at the end of this function, and filled from advance_decoder_to_row?

adamtobey (approved):
I think this all makes sense to me. It does get a bit confusing in ways that I suspect could be factored better, but I think it's reasonable to not worry about that now (or maybe I'm wrong about that). Left some comments, but didn't find anything I'd consider an issue

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
g-talbot: **Edit (later, after re-examining the page-cache memory model):** my original answer below is correct for *which array we read* — body cols are looked up by name from the union schema, so different inputs can declare body cols in different orders and we still get the right values. But it's wrong about the *memory-bound side*. The streaming engine's O(N × constant) body-col cache only holds if each input's body cols are in the same order as the union schema iterates them — alphabetical lexicographic after the sort cols ("Husky order"). 

Why: phase 3 asks each input's decoder to advance to the next union-schema col. Parquet emits column chunks in declared schema order, so the decoder reads pages in *that input's* storage order. If an input's body cols are in a different order, `fill_page_cache_to_row` has to drain every body col preceding the requested one on the wire, and those pages get cached under their own `col_idx` until *their* turn comes up in the union iteration. Cache size grows to a full column-chunk's worth per misaligned col — that's vertical caching, not horizontal, and defeats the two-GET streaming model.

I missed this on the original reply. Fixed in `9a7136d15` (correction force-pushed under the existing PR):

- New `assert_inputs_in_husky_body_col_order` runs right after `build_input_decoders_state` (before phase 0) and bails on any input whose body cols aren't lexicographic after the sort cols. Concrete error message names the offending pair.
- New regression test `test_assert_inputs_in_husky_body_col_order_rejects_misaligned_input` builds an input with body cols `[value, metric_type]` (alphabetical would be `[metric_type, value]`) and asserts the merge bails before phase 3 runs.
- Expanded the `body_col_page_cache` docstring with the memory-bound argument that depends on this precondition. The helper docstring on `assert_inputs_…` carries the "why this is load-bearing" explanation.

Streaming writer + legacy Husky writer both emit in Husky order today, so the assertion catches future producer drift, not current production traffic.

Original (incomplete) reply, kept for context:

---

Yes — sort cols first, then body cols, is a hard requirement for storage ordering *within each input*. The contract:

- **Sort cols come first.** The error at L545 enforces this — any body-col page arriving before all sort cols are drained triggers a bail. Producers (`streaming_writer` + legacy writer) both honour it; the streaming engine's phase 0 (`drain_sort_cols_all_inputs`) relies on it.
- ~~**Body cols are in parquet storage order (= schema order), per input.** No cross-input constraint on body-col order — the body-col loop in phase 3 drives from the *union* schema and looks each column up by name (`s.input_col_indices[input_idx]`), so different inputs can declare body cols in different orders or omit them entirely. SS-3 covers the "input lacks this body col" case with implicit-null fills.~~ See the correction above — the cross-input *ordering* (lexicographic after sort cols) is a hard contract for the memory bound, even though the *which-cols-are-present* freedom (SS-3 null fills) remains.

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
g-talbot: `rg_partition_prefix_len` is genuinely optional in the data model. Older splits (pre-prefix-aligned-RG era) and any split not written by the streaming engine simply lack the KV. The logic above this line already enforces that all inputs *that do declare a value* agree (else we'd have bailed earlier), so falling through to 0 means "none of the inputs claimed a prefix," not "we lost the value."

The flip side — mixing 0 with non-zero prefixes — is allowed only through the legacy-promotion executor added in PR-6423, which sets `mixed_prefix_ok = true` to bypass the equality check in `merge_parquet_split_metadata`. Pushed a comment block above the `unwrap_or(0)` capturing this so the next reader doesn't have to re-derive it.

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
g-talbot: Yes — PR-6c.2 / #6424 (stacked on top of this PR) is exactly that relaxation. Multi-RG inputs with prefix-aligned row groups go through `process_region` / composite-key encoding. Added a comment at the bail site pointing to the follow-up PR so future readers don't have to trace through the stack to find it.

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
g-talbot: Nice catch — applied. Local `rows_for_current_output` accumulates inside the inner loops, single store to `rows_per_output[out_idx]` at the end of each outer iteration.

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
g-talbot: Agreed. The TODO is on the Phase 4d roadmap (RLE / dict-merge optimisations). Today the merge path goes through `arrow::compute::interleave` over decoded plain arrays; once we track dictionary cardinality + key-space identity across inputs, we can route same-dict inputs through a key-renumbering merge that skips the un-encode / re-encode round-trip entirely. Same idea applies to RLE-encoded numeric columns.

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
g-talbot: Renamed to `fill_page_cache_to_row` (and updated callers / docstrings). Your read of the original name was the same as mine on a re-read — "advancing the decoder" implied a side-effect on the decoder cursor, when really the only side-effect is appending pages to the per-input cache. New name leads with the actual effect.

quickwit/quickwit-parquet-engine/src/merge/streaming.rs
g-talbot: Right idea, but it's bounded much more tightly than "all pages of the column." Per-input cache size is bounded by `ceil(OUTPUT_PAGE_ROWS / input_page_rows) + small_slack` — driven by `fill_page_cache_to_row` (formerly `advance_decoder_to_row`), which fetches only enough pages to cover the rows from this input that contribute to **one output page** (1024 rows, set by `OUTPUT_PAGE_ROWS`). Eviction at the bottom of `assemble_one_output_page` drops pages whose rows fall below the cursor after each output page emits.

Total memory across inputs: `N_inputs × ceil(OUTPUT_PAGE_ROWS / input_page_rows)` — horizontal across inputs, not vertical down a column. The invariant is asserted by `test_ms7_body_col_page_cache_bounded_regardless_of_input_size` (MS-7): going from a 3 000-row fixture to a 30 000-row fixture (10× more input pages) must not grow the resident page count by more than 2. The failure message reads *"body col write is no longer page-bounded; likely buffering whole column chunks"* — i.e., the vertical-caching regression you were sniffing at would trip the test.

See the (newly expanded) doc comment on `InputDecoderState::body_col_page_cache` for the on-record version of this argument.

g-talbot: Yes, with one important clarification: "the current output" here is one **output page** (1024 rows, capped by `OUTPUT_PAGE_ROWS`), not one output file. `assemble_one_output_page` is called in a loop, once per output page; pages are dropped at the end of each call (lines 1751–1762) as soon as the cursor passes their last row, not just at the end of an output file. That's what keeps the cache horizontal-per-input instead of vertical-across-the-column. See the longer answer threaded under your original comment.

PR #6411, opened 2026-05-08
Wire up shared indexing split cache between indexers and compactors

nadav-govari:
### Description
Pulls up the indexing cache so it can optionally be shared between compactors and indexers. By default it's on for backwards compatibility. But when running split compactors, you'll likely want it off.

### How was this PR tested?
Unit tests.

### DIFF
--- a/quickwit/quickwit-cli/src/tool.rs
+++ b/quickwit/quickwit-cli/src/tool.rs
@@ -17,6 +17,7 @@ use std::io::{IsTerminal, Stdout, Write, stdout};
 use std::num::NonZeroUsize;
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::sync::Arc;
 use std::time::{Duration, Instant};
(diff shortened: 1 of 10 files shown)


quickwit/quickwit-cli/src/tool.rs
```diff
@@ -417,6 +418,8 @@ pub async fn local_ingest_docs_cli(args: LocalIngestDocsArgs) -> anyhow::Result<
         &HashSet::from_iter([QuickwitService::Indexer]),
     )?;
     let universe = Universe::new();
+    let split_cache =
+        Arc::new(IndexingSplitCache::from_config(&indexer_config, &config.data_dir_path).await?);
```
nadav-govari, line 422: check enabled services to determine whether we need a cache at all rather than relying on config values being 0

nadav-govari, line 422: read from node config

PR #6416, opened 2026-05-11
feat: track resource stats across split, leaf and root search

fulmicoton:
### Description

Introduce a `resource_stats` payload on `LeafSearchResponse` and `SearchResponse` so callers can see how a query spent its budget across the split → leaf → root hierarchy.

#### Proto

Three new messages in `quickwit-proto/protos/quickwit/search.proto`:

- **`SplitResourceStats`** — per-split counters (`split_num_docs`, `matched_num_docs`), input working set (`input_memory_bytes`), storage download (`download_num_bytes`, `download_num_requests`), and per-phase microsecond timings (permit wait, warmup, cpu-pool wait, cpu search).
- **`LeafResourceStats`** — per-leaf counters: partial-result-cache hits, lambda dispatched / lambda-succeeded splits, the lambda-bottleneck signal, locally-executed splits; plus `split_resources_worst` (worst split by phase-sum), `split_resources_sum` (field-wise sum over local splits), and the leaf's `wall_time_microsecs`.
- **`RootResourceStats`** — per-search aggregates: `leaf_resources_worst`, `leaf_resources_sum` (field-wise sums of every leaf), leaf-call counts (excluding and including retries), and failed-split count.

#### Plumbing

- `CountingStorage` (new wrapper in `quickwit-storage`) records downloaded bytes / GET requests per request.
- `leaf_search_single_split` assembles `SplitResourceStats` from the counters and per-phase `Instant`s; it writes to either `localexec_*` or `lambda_success_*` depending on `quickwit_common::is_running_in_lambda` — a process-wide cached bool driven by the `AWS_LAMBDA_FUNCTION_NAME` env var the Lambda runtime sets.
- `run_offloaded_search_tasks` captures the lambda dispatch totals (`lambda_num_splits` / `lambda_num_docs`) and forwards them via a new `IncrementalCollector::add_lambda_totals`. Lambda failure counting stays in the client because transport-level failures cannot be observed by the lambda server.
- `LeafSearchCache::put` rewrites `resource_stats` to the cache-hit shape at write time, so cache reads need no per-read mutation.
- `multi_index_leaf_search` stamps `wall_time_microsecs` on the merged `LeafResourceStats`.
- `lambda_bottleneck` is determined inside `single_doc_mapping_leaf_search` via an `AtomicBool` race between the local and offload futures (last writer wins after `tokio::join!`).
- `ClusterClient::leaf_search` takes an `Arc<AtomicU64>` so the root can count leaf attempts including retries.
- `search_partial_hits_phase` now returns an `Option<RootResourceStats>` alongside the merged response; `compute_root_resource_stats` produces it from the per-leaf responses before they are merged. `root_search_aux` attaches it to `SearchResponse.resource_stats`.
- The gRPC `Scroll` handler surfaces `resource_stats` when a scroll page triggered an inner search (pure cache-hit pages stay `None`).
- `add_leaf_stats` is fully extensive: every numeric field is summed, including `wall_time_microsecs` and `lambda_bottleneck` (the latter becomes a count of leaves where lambda was the bottleneck once aggregated).

#### Lambda outcome metric

`LAMBDA_METRICS.leaf_search_requests_total` / `leaf_search_duration_seconds` now carry an `outcome` label whose value is `success`, `timeout` or `other`. Timeout is detected from SDK / EFS / SnapStart timeout errors.

### How was this PR tested?

- Unit tests on `add_split_stats`, `add_leaf_stats`, `compute_root_resource_stats`, and `CountingStorage`. The "sums every field" tests destructure on the proto types so a new field forces a compile error pointing at the test.
- End-to-end gRPC root test `test_root_search_populates_resource_stats` exercising the same free function the `RootSearch` handler invokes, with mocked metastore and leaf services.
- `cargo nextest run -p quickwit-search` passes locally.
- `cargo nextest run -p quickwit-storage --tests counting_storage` passes locally.

### DIFF
--- a/quickwit/quickwit-proto/protos/quickwit/search.proto
+++ b/quickwit/quickwit-proto/protos/quickwit/search.proto
@@ -392,14 +392,13 @@ message SplitResourceStats {
 // If the configuration allows it, leaf nodes can offload part of their computation to
 // lambdas.
 //
-// `split_resources_worsts` / `split_resources_sum` aggregations on
+// `split_resources_worst` / `split_resources_sum` aggregations on
 // `LeafResourceStats` are only computed for locally-executed splits.
 //
 // All numeric fields are extensive (summed when merging across leaves).
-// `split_resources_worsts` is the exception — it is the top-2 worst single-split
-// contributions selected by ranking (largest first), not summed.
-// `lambda_bottleneck` is 0 or 1 at the source and becomes a count of leaves
-// where lambda was the bottleneck once aggregated.
+// `split_resources_worst` is the exception — it is selected by ranking, not
+// summed. `lambda_bottleneck` is 0 or 1 at the source and becomes a count of
+// leaves where lambda was the bottleneck once aggregated.
 // `min_wait_for_search_permit_microsecs` and `min_wait_for_cpu_pool_microsecs`
 // are also exceptions — they are merged with `min` rather than `sum` (see
 // their per-field doc below).
@@ -413,12 +412,11 @@ message LeafResourceStats {
     uint64 localexec_num_splits = 8;
     // Sum of `split.num_docs` across locally-executed splits.
     uint64 localexec_num_docs = 9;
-    // Top-2 worst single-split contributions, ranked by
-    // `warmup + wait_for_cpu_pool + cpu_predicate + cpu_collection + cpu_harvest`
-    // (intentionally excludes `wait_for_search_permit`). Largest first: index
-    // 0 is the absolute worst, index 1 is the second worst (or absent if only
-    // one local split contributed).
-    repeated SplitResourceStats split_resources_worsts = 10;
+    // The worst single-split contribution, ranked by `warmup + cpu_search`
+    // (intentionally excludes the queueing phases `wait_for_search_permit`
+    // and `wait_for_cpu_pool` so the ranking reflects "how much work this
+    // split actually did").
+    SplitResourceStats split_resources_worst = 10;
     // Field-wise sum of `SplitResourceStats` across all locally-executed splits.
     // If you want to compute averages, divide by localexec_num_splits.
     SplitResourceStats split_resources_sum = 11;
@@ -455,12 +453,8 @@ message LeafResourceStats {
 
 // Resource statistics for a root search.
 message RootResourceStats {
-    // Top-2 leaves by `wall_time_microsecs`, largest first: index 0 is the
-    // absolute worst leaf, index 1 is the second worst (or absent if only
-    // one leaf contributed stats). Useful to tell whether a single
-    // straggler dominated a query or whether several leaves were
-    // comparably slow.
-    repeated LeafResourceStats leaf_resources_worsts = 1;
+    // The leaf with the largest `wall_time_microsecs`.
+    LeafResourceStats leaf_resources_worst = 1;
     // Field-wise sum of all leaf stats. `wall_time_microsecs` is summed even though
     // it is not strictly extensive — the sum still gives a useful view of total leaf time.
     // If you want to compute averages, divide by leaf_num_calls: failed leaf attempts
--- a/quickwit/quickwit-proto/protos/quickwit/search.proto
+++ b/quickwit/quickwit-proto/protos/quickwit/search.proto
@@ -409,17 +409,17 @@ message LeafResourceStats {
     uint64 partial_result_cache_num_docs = 2;
 
     // Number of splits executed locally (excluding cache hits and lambda).
-    uint64 localexec_num_splits = 8;
+    uint64 localexec_num_splits = 3;
     // Sum of `split.num_docs` across locally-executed splits.
-    uint64 localexec_num_docs = 9;
+    uint64 localexec_num_docs = 4;
     // The worst single-split contribution, ranked by `warmup + cpu_search`
     // (intentionally excludes the queueing phases `wait_for_search_permit`
     // and `wait_for_cpu_pool` so the ranking reflects "how much work this
     // split actually did").
-    SplitResourceStats split_resources_worst = 10;
+    SplitResourceStats split_resources_worst = 5;
     // Field-wise sum of `SplitResourceStats` across all locally-executed splits.
     // If you want to compute averages, divide by localexec_num_splits.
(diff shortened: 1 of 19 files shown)


quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -355,12 +358,98 @@ message LeafSearchRequest {
...
+    uint64 partial_result_cache_num_docs = 2;
+
+    // Number of splits executed locally (excluding cache hits and lambda).
+    uint64 localexec_num_splits = 8;
+    // Sum of `split.num_docs` across locally-executed splits.
+    uint64 localexec_num_docs = 9;
+    // The worst single-split contribution, ranked by
+    // `warmup + wait_for_cpu_pool + cpu_predicate + cpu_collection + cpu_harvest`
+    // (intentionally excludes `wait_for_search_permit`).
+    SplitResourceStats split_resources_worst = 10;
```
trinity-1686a: why include wait_for_cpu_pool but exclude wait_for_search_permit? Neither is the fault of the split, both are related to the whole node's load

quickwit/quickwit-storage/src/counting_storage.rs
```diff
@@ -0,0 +1,229 @@
...
+            dyn ::core::future::Future<Output = StorageResult<()>>
+                + ::core::marker::Send
+                + 'async_trait,
+        >,
+    >
+    where
+        'life0: 'async_trait,
+        'life1: 'async_trait,
+        'life2: 'async_trait,
+        Self: 'async_trait,
```
trinity-1686a: is there a particular reason to spell this out instead of letting async_trait generate it?

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -630,14 +636,46 @@ async fn leaf_search_single_split(
...
+                    wait_for_search_permit_microsecs: wait_for_search_permit.as_micros() as u64,
                     warmup_microsecs: warmup_duration.as_micros() as u64,
-                    cpu_thread_pool_wait_microsecs: cpu_thread_pool_wait_microsecs.as_micros()
-                        as u64,
-                });
+                    wait_for_cpu_pool_microsecs: cpu_thread_pool_wait_microsecs.as_micros() as u64,
+                    cpu_search_microsecs: cpu_start.elapsed().as_micros() as u64,
+                };
+                leaf_search_response.resource_stats =
+                    Some(if quickwit_common::is_running_in_lambda() {
```
trinity-1686a: i feel like this code is a bit weird in how it's written
i'd be in favour of 
```rust
leaf_search_response.resource_stats = if is_running_in_lambda () {
    // Running inside an AWS Lambda runtime [...]
	Some(..)
} else {
let split_stats = ..;
   Some(..)
}

File: quickwit/quickwit-search/src/leaf.rs [hunk 4]
```diff
@@ -1670,10 +1738,32 @@ pub async fn single_doc_mapping_leaf_search(
...
 
-    let (offloaded_res, _) =
-        tokio::join!(run_offloaded_search_tasks_fut, run_local_search_tasks_fut);
+    // Each path stores its own outcome on completion (offloaded → true,
+    // local → false). The load after `tokio::join!` therefore observes the
+    // value written by whichever future finished last: lambda was the
+    // bottleneck iff the offloaded path is the last one to write.
+    let offloaded_is_bottleneck = Arc::new(AtomicBool::new(false));
+    let flag_local_writer = offloaded_is_bottleneck.clone();
+    let flag_offloaded_writer = offloaded_is_bottleneck.clone();
```
trinity-1686a: nit: Arc not needed here, with structured concurrency we have clear lifetimes to a simple ref is enough (we do need the atomic for shared mutability though)
```suggestion
    let offloaded_is_bottleneck = AtomicBool::new(false);
    let  offloaded_is_bottleneck_ref = & offloaded_is_bottleneck;
```

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -355,12 +358,98 @@ message LeafSearchRequest {
...
+    // Bytes downloaded from storage during the request, as measured
+    // by a request-scoped storage wrapper.
+    uint64 download_num_bytes = 3;
+    // Number of storage GET requests issued during the request.
+    uint64 download_num_requests = 4;
+    // Number of documents matched by the query in this split (we always count).
+    uint64 matched_num_docs = 5;
+
+    // Time the split spent waiting to acquire its search permit.
+    uint64 wait_for_search_permit_microsecs = 6;
```
trinity-1686a, line 379: i think for wait_for_search_permit_microsecs and wait_for_cpu_pool_microsecs it would make sens to record leaf-level smallest value, to be able to tell appart "we spend a while acquiring semaphores because the node was loaded" (high min) from "we spend a while acquiring semaphores because this was a huge query" (low min)

quickwit/quickwit-search/src/lib.rs
```diff
@@ -346,125 +347,320 @@ pub fn searcher_pool_for_test(
...
+            matched_num_docs: 20_000,
+            wait_for_search_permit_microsecs: 200_000,
+            warmup_microsecs: 2_000_000,
+            wait_for_cpu_pool_microsecs: 20_000_000,
+            cpu_search_microsecs: 200_000_000,
+        };
+        // Destructure on the proto type itself so a newly-added field forces
+        // an update to this test (otherwise the assertion below would silently
+        // miss it).
+        let SplitResourceStats {
```
trinity-1686a, line 495: this would be catched when creating acc and other? this isn't go, unless ..Default::default() missing fields are hard error, not zero-initialized

trinity-1686a (approved):
i think nothing take into account the fetch_docs phase. It's not typically the most expensive part, but that's still a blindspot

quickwit/quickwit-storage/src/counting_storage.rs
fulmicoton: no. you are right!

quickwit/quickwit-proto/protos/quickwit/search.proto
fulmicoton: that's a great idea.

PR #6427, opened 2026-05-13
bugfix: dont schedule indexing tasks on retiring indexers

### LINKED ISSUE #6413: Indexers not gracefully terminating and getting force killed
**Describe the bug**
Occasionally, indexers do not terminate gracefully, and get force killed after timeout (5 mins). 

I believe this is what is happening:

1. On SIGTERM, indexer-1 drains its local shards
2. However, indexer-1 might be assigned shard A that is owned by indexer-0. Data will continue to be written to shard A because its owned by healthy indexer-0. indexer-1 will continue to fetch data from indexer-0 WAL, and indexer-1 will not get decomissioned.
3. The control plane doesn't know if an indexer is being decomissioned, so if an indexer was assigned a remote shard, it will continue to be assigned that remote shard. 

**Expected behavior**
The control plane should probably reassign shard A to an indexer that's not being decomissioned. https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs#L390

**Configuration:**

ncoiffier-celonis:
### Description

Attempt to fix https://github.com/quickwit-oss/quickwit/issues/6413

This PR tracks the ingester status in the control plane indexer pools, and avoid to schedule indexing tasks on retiring indexers.

Special case: for a single node indexer (for example most of the integration tests), we still need to schedule indexing tasks on retiring indexers, otherwise the shards are never "drained", and the decomission process times out without publishing the splits.

### Remark

I'm not sure how I can create an integration test that reproduce the bug without leaking the shard assignment details to the outside.

### How was this PR tested?

Unit tests only. Not tested locally, no integration tests.


guilload (approved):

ncoiffier-celonis:
@guilload thank you for the approval. Should I just merge this one?

PR #6429, opened 2026-05-13
Add IO metrics

guilload:
### Description
Keep track of bytes written in the WAL and add process level metrics.

### How was this PR tested?
Tested on benchmark machine

### DIFF
--- a/quickwit/quickwit-common/src/metrics.rs
+++ b/quickwit/quickwit-common/src/metrics.rs
@@ -15,7 +15,7 @@
 use std::sync::LazyLock;
 
 pub use prometheus::{exponential_buckets, linear_buckets};
-use quickwit_metrics::{Gauge, LazyGauge, gauge, lazy_gauge};
+use quickwit_metrics::{Gauge, LazyCounter, LazyGauge, gauge, lazy_counter, lazy_gauge};
 
 pub fn index_label(index_id: &str) -> &str {
     static PER_INDEX_METRICS_ENABLED: LazyLock<bool> =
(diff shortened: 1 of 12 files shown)


quickwit/quickwit-common/src/metrics.rs
```diff
@@ -451,3 +451,50 @@ pub fn index_label(index_id: &str) -> &str {
...
+    pub read_bytes_total: IntCounter,
+    pub write_bytes_total: IntCounter,
+    pub read_syscalls_total: IntCounter,
+    pub write_syscalls_total: IntCounter,
+}
+
+impl Default for IoMetrics {
+    fn default() -> Self {
+        Self {
+            read_bytes_total: new_counter(
```
nadav-govari: We're doing inc_by with deltas, but the names are `_total` - we should probably omit total.

quickwit/quickwit-ingest/src/mrecordlog_async.rs
```diff
@@ -131,7 +161,15 @@ impl MultiRecordLogAsync {
         let span = info_span!("mrecordlog.truncate", queue, position);
         let queue = queue.to_string();
         self.run_operation(span, move |mrecordlog| {
-            mrecordlog.truncate(&queue, position)
```
nadav-govari, line 134: why deos this change work? one passes a u64 and the other passes a RangeInclusive? Is there some magical promotion that takes place somewhere?

nadav-govari (approved):
You have some merge conflicts

quickwit/quickwit-ingest/src/mrecordlog_async.rs
guilload: The API of mrecordlog has changed.

PR #6442, opened 2026-05-18
Use CRC32C for checksums instead of MD5

nadav-govari:
### Description

Spans show that MD5 computations can take on the order of seconds, and were being computed on the non-blocking threads. Luckily we were only doing multipart uploads on massive splits so the vast majority of our uploads were unaffected.

MD5 isn't recommended anymore anyway, and on modern architecture, CRC32C has hardware acceleration. (Theres a software fallback if the architecture doesnt support it: https://docs.rs/crc32c/latest/crc32c/)

As for setting both checksum types on the completed part, this will only return the algorithm that the request was sent with: https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html

### How was this PR tested?
Local deployment uploads succeed.


quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -344,6 +332,7 @@ impl S3CompatibleObjectStorage {
             self.s3_client
                 .create_multipart_upload()
                 .bucket(self.bucket.clone())
+                .checksum_algorithm(ChecksumAlgorithm::Crc32C)
```
fulmicoton-dd: ah cool ! I did not know it was possible

fulmicoton-dd:
It is a great idea, but it will break compatibility on most non S3 object storage relying on this client.

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -94,6 +95,9 @@ pub struct S3CompatibleObjectStorage {
     retry_params: RetryParams,
     disable_multi_object_delete: bool,
     disable_multipart_upload: bool,
+    // `None` means native AWS S3. Non-S3 flavors (GCS, MinIO, Garage, DigitalOcean) do not
+    // reliably honor CRC32C, so we only request the SDK to compute & send it for native S3.
+    flavor: Option<StorageBackendFlavor>,
```
guilload: I'd rather have a proper `checksum_algorithm: ChecksumAlgorithm` on `S3StorageConfig` defaulting to CRC32C and otherwise MD5 for all other flavors. I'd also like to keep the ability to easily revert to MD5 if something goes wrong. It seems that GCP also supports CRC32C. We have ways to test that.

trinity-1686a:
Garage supported CRC32C before its 1.0.0, minio has it since 2015 afaict, i think it's fine to use this algorithm by default for all flavours, and to give a config escape door for any s3 provider that doesn't have this

nadav-govari:
@fulmicoton @trinity-1686a either of you want to take another look?

quickwit/quickwit-config/src/storage_config.rs
```diff
@@ -330,7 +344,7 @@ pub struct S3StorageConfig {
     #[serde(default)]
     pub disable_multipart_upload: bool,
     #[serde(default)]
-    pub disable_checksums: bool,
+    pub checksum_strategy: ChecksumStrategy,
```
trinity-1686a: can you also update docs/configuration/storage-config.md

quickwit/quickwit-config/src/storage_config.rs
```diff
@@ -343,22 +357,22 @@ impl S3StorageConfig {
...
                 self.disable_multi_object_delete = true;
                 self.disable_multipart_upload = true;
-                self.disable_checksums = true;
+                self.checksum_strategy = ChecksumStrategy::Disabled;
             }
             Some(StorageBackendFlavor::MinIO) => {
                 self.region = Some("minio".to_string());
                 self.force_path_style_access = true;
-                self.disable_checksums = true;
+                self.checksum_strategy = ChecksumStrategy::Disabled;
```
trinity-1686a: Garage and MinIO support crc32c (source: git history), Ceph documentation mention crc32c so i assume DO has it too, and guilload said Gcs supports it to, i think we can leave the checksum strategy as default for all flavours unless we end up with reports that it causes issues

fulmicoton-dd: guilload said Gcs supports it to > I am not sure this is correct, but I think we use a specific client for GS no?

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -420,6 +450,7 @@ impl S3CompatibleObjectStorage {
         Ok(delete_requests)
     }
 
+    #[tracing::instrument(skip_all, fields(part_number = part.part_number, num_bytes=part.len()))]
```
fulmicoton-dd: do we really care about tracing here? :-/

quickwit/quickwit-config/src/storage_config.rs
```diff
@@ -330,7 +344,7 @@ pub struct S3StorageConfig {
     #[serde(default)]
     pub disable_multipart_upload: bool,
     #[serde(default)]
-    pub disable_checksums: bool,
```
fulmicoton-dd, line 333: that does not look backward compatible. I think noone uses it, and they will get a clear error message, so it is probably ok if it is a calculated risk.

fulmicoton-dd (approved):

fulmicoton-dd:
Approved, but please update the doc as suggested by @trinity-1686a

quickwit/quickwit-config/src/storage_config.rs
```diff
@@ -38,6 +38,20 @@ pub enum StorageBackend {
     S3,
 }
 
+/// Strategy used to checksum object-storage uploads.
+#[derive(Debug, Clone, Copy, Default, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ChecksumStrategy {
```
guilload:
```suggestion
pub enum ChecksumAlgorithm {
```

quickwit/quickwit-config/src/storage_config.rs
guilload: Nadav, you can make this backward compatible with some serde trick. We can use an alias and then deserialize `false` to disabled and true to the new default.

quickwit/quickwit-config/src/storage_config.rs
nadav-govari: DO support: https://docs.digitalocean.com/reference/api/spaces/ (its supported)
To use CRC32C, you have to supply a GCP-specific header, which we won't be able to do from the S3 SDK. So I'll leave it disabled for GCS and enable it for the other two.

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
nadav-govari: This is actually one of the more useful spans I've used - it's how I found that MD5 was taking a lot of compute. S3 operations can sometimes get rate limited and have variable throughput rates, which becomes more important on a compactor (which is the only work they do)

guilload: Debug or trace level is also an option.

nadav-govari: Ill lower the level to debug.

docs/configuration/storage-config.md
```diff
@@ -46,17 +46,18 @@ This section contains one configuration subsection per storage provider. If a st
...
-| `disable_checksums` | Disables [checksums](https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html) on requests and responses. Required by S3-compatible providers that do not support the additional checksum algorithms enabled by default in recent versions of the AWS SDK (Digital Ocean, Garage, GCS, MinIO). | `false` |
+| Property | Description                                                                                                                                                                                                                                                                                                         | Default value             |
+| --- |---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------|
```
guilload: The new formatting of the table is gross and makes the diff harder to review.

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -94,6 +96,7 @@ pub struct S3CompatibleObjectStorage {
     retry_params: RetryParams,
     disable_multi_object_delete: bool,
     disable_multipart_upload: bool,
+    checksum_strategy: quickwit_config::ChecksumAlgorithm,
```
guilload:
```suggestion
    checksum_algorithm: quickwit_config::ChecksumAlgorithm,
```

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -154,7 +157,10 @@ pub async fn create_s3_client(s3_storage_config: &S3StorageConfig) -> S3Client {
     s3_config.set_stalled_stream_protection(Some(stalled_stream_protection));
     s3_config.set_timeout_config(aws_config.timeout_config().cloned());
 
-    if s3_storage_config.disable_checksums {
+    if matches!(
+        s3_storage_config.checksum_algorithm,
+        quickwit_config::ChecksumAlgorithm::Disabled
```
guilload: nit: I'd add `is_disables()` and `is_md5()` helpers directly on the `ChecksumAlgorithm` enum.

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -356,32 +383,42 @@ impl S3CompatibleObjectStorage {
...
-
+            .collect();
         let mut parts = Vec::with_capacity(multipart_ranges.len());
-
         for (multipart_id, multipart_range) in multipart_ranges.into_iter().enumerate() {
-            let read = payload
-                .range_byte_stream(multipart_range.clone())
-                .await?
-                .into_async_read();
-            let md5 = compute_md5(read).await?;
```
guilload, line 377: Let's move compute_md5 into a spawn blocking.

docs/configuration/storage-config.md
nadav-govari: Shockingly this isn't claude, it's intellij. Fixing

quickwit/quickwit-config/src/storage_config.rs
```diff
@@ -330,7 +354,10 @@ pub struct S3StorageConfig {
     #[serde(default)]
     pub disable_multipart_upload: bool,
     #[serde(default)]
-    pub disable_checksums: bool,
+    pub checksum_algorithm: ChecksumAlgorithm,
+    /// Deprecated: applies into `checksum_algorithm: disabled`.
+    #[serde(default, skip_serializing)]
```
nadav-govari, line 359: remove the option

PR #6439, opened 2026-05-17
Optimize list fields

guilload:
### Description
- split code in multiple files
- optimize de/serialization of list fields entries
- remove de/serialization on cache get/put
- move CPU-bound work out of tokio runtime
- cap number of concurrent fields metadata downloads
- pushdown filtering down to leaves
- fix naming, clean up code, etc.

### How was this PR tested?
Test locally: indexed on main, issued a bunch of list fields queries and then switched to branch and issued same queries with matching results.

PR #6443, opened 2026-05-18
feat(_mapping): timestamp pushdown + column-hints fast path (on top of #6439)

congx4:
## What's in this PR

Today `GET /_elastic/{index}/_mapping(s)` calls `list_fields` over every published split. On indexes with hundreds of thousands of dynamic fields this can take several seconds, and over a certain threshold the leaf hits `QW_FIELD_LIST_SIZE_LIMIT` (100k by default) and the request fails.

1. **Timestamp pushdown** — new `?start_timestamp=…&end_timestamp=…` URL params on `_mapping(s)`, forwarded into `ListFieldsRequest` verbatim. The metastore prunes the candidate split set by time window before any leaf fan-out. Unit is epoch seconds, half-open interval — matching the existing `ListFieldsRequest` proto contract.

2. **field_patterns** — new `?fields=…` URL param (comma-separated names). Push fields patterns to leaf_list_fields to reduce the workload(sort, merge).

## Test plan

- [x] `cargo build -p quickwit-serve` — clean
- [x] `cargo clippy -p quickwit-serve --tests` — clean
- [x] `cargo +nightly fmt --all -- --check` — clean
- [x] `cargo nextest run -p quickwit-serve --lib elasticsearch_api::` — 75 / 75 pass


quickwit/quickwit-serve/src/elasticsearch_api/model/index_mapping_query_params.rs
```diff
@@ -0,0 +1,102 @@
...
+/// see `parse_field_hints` in `rest_handler.rs` for the fast-path semantics.
+#[serde_with::skip_serializing_none]
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
+pub struct IndexMappingQueryParams {
+    #[serde(default)]
+    pub start_timestamp: Option<i64>,
+    #[serde(default)]
+    pub end_timestamp: Option<i64>,
+    #[serde(default)]
+    pub fields: Option<String>,
```
guilload: Are those field patterns?

quickwit/quickwit-serve/src/elasticsearch_api/model/index_mapping_query_params.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::IndexMappingQueryParams;
+
+    #[test]
+    fn empty_query_string_yields_none() {
+        let params: IndexMappingQueryParams = serde_urlencoded::from_str("").unwrap();
```
guilload: Can we use `serde_qs`? I believe it's already a dependency.

quickwit/quickwit-serve/src/elasticsearch_api/model/mappings.rs
```diff
@@ -66,11 +66,25 @@ impl ElasticsearchMappingsResponse {
...
+        list_fields_response: Option<&ListFieldsResponse>,
+        requested_fields: &[String],
+    ) -> Self {
+        let filter: HashSet<&str> = requested_fields.iter().map(String::as_str).collect();
         let indices = indexes_metadata
             .into_iter()
             .map(|index_metadata| {
                 let field_mappings = &index_metadata.index_config.doc_mapping.field_mappings;
                 let mut properties = build_properties(field_mappings);
+                if !filter.is_empty() {
```
guilload: Can we push down the filter to the leaves and single split list fields requests so we don't need to do this here?

quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
```diff
@@ -199,8 +199,37 @@ async fn get_index_metadata(
     Ok(index_metadata)
 }
 
+/// Splits `fields=` on commas, trims, drops empties. `None` if nothing usable remains.
+fn parse_field_hints(params: &IndexMappingQueryParams) -> Option<Vec<String>> {
+    let raw = params.fields.as_deref()?;
+    let tokens: Vec<String> = raw
+        .split(',')
+        .map(|s| s.trim().to_string())
```
guilload: nit: in general, it's better to allocate after filtering, but it does not matter here.

quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
```diff
@@ -199,8 +199,37 @@ async fn get_index_metadata(
...
+        .collect();
+    if tokens.is_empty() {
+        None
+    } else {
+        Some(tokens)
+    }
+}
+
+fn collect_declared_top_level_names(indexes_metadata: &[IndexMetadata]) -> HashSet<String> {
+    let mut names = HashSet::new();
```
guilload: nit: functional way might be nicer... depending on who you ask :)

guilload: Don't forget to use `with_capacity` if you know the final size in advance. It also signals to the reader that there will be no filtering or "explosion"

quickwit/quickwit-serve/src/elasticsearch_api/model/index_mapping_query_params.rs
congx4: yes, they are. I will rename it to field_pattern to align with your code.

quickwit/quickwit-serve/src/elasticsearch_api/model/mappings.rs
congx4: we have pushed down the filter to the leaves. https://github.com/quickwit-oss/quickwit/pull/6443/changes#diff-390322509891292c93be51dccd4522bb2b0d352e41f3be483d41e92cc4bfdc68R278

quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
```diff
@@ -199,8 +199,29 @@ async fn get_index_metadata(
     Ok(index_metadata)
 }
 
+/// Splits `field_patterns=` on commas, trims, drops empties. `None` if nothing
+/// usable remains.
+fn parse_field_patterns(params: &IndexMappingQueryParams) -> Option<Vec<String>> {
```
guilload: nit: This could be a method on `IndexMappingQueryParams`.

quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
```diff
@@ -199,8 +199,29 @@ async fn get_index_metadata(
...
+        .map(str::trim)
+        .filter(|s| !s.is_empty())
+        .map(str::to_string)
+        .collect();
+    if patterns.is_empty() {
+        None
+    } else {
+        Some(patterns)
+    }
+}
```
guilload:
```suggestion
        let field_patterns: Vec<String> = params
        .field_patterns.as_deref()?
        .split(',')
        .filter_map(|pattern| {
            let trimmed_pattern = pattern.trim();

            if trimmed_pattern.is_empty() {
                None
            } else {
                Some(trimmed_pattern.to_string())
            }
        })
        .collect();

    if field_patterns.is_empty() {
        None
    } else {
        Some(field_patterns)
    }
}
```

quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
```diff
@@ -214,22 +235,24 @@ pub(crate) async fn es_compat_index_mapping(
...
         .collect();
+
+    let field_patterns = parse_field_patterns(&params);
+
     let list_fields_request = quickwit_proto::search::ListFieldsRequest {
         index_id_patterns,
-        field_patterns: Vec::new(),
-        start_timestamp: None,
-        end_timestamp: None,
+        field_patterns: field_patterns.unwrap_or_default(),
```
guilload: It looks like returning an empty Vec in `parse_field_patterns` is fine if you're going to `unwrap_or_default` or default here.

quickwit/quickwit-serve/src/elasticsearch_api/rest_handler.rs
```diff
@@ -214,22 +235,24 @@ pub(crate) async fn es_compat_index_mapping(
...
+        start_timestamp: params.start_timestamp,
+        end_timestamp: params.end_timestamp,
         query_ast: None,
     };
     let list_fields_response = search_service
         .root_list_fields(list_fields_request)
         .await
         .ok();
-    let response = ElasticsearchMappingsResponse::from_doc_mapping(
```
guilload, line 228:
```rust
let response = ElasticsearchMappingsResponse::from_doc_mapping(...)
Ok(response)
```

guilload (approved):
LGTM. Left a bunch of nits.

PR #6446, opened 2026-05-19
Update the instruction set used to build our images

fulmicoton-dd:
both in Arm and x86.

This remains very conservative.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,10 +3,15 @@ rustflags = ["--cfg", "tokio_unstable"]
 rustdocflags = ["--cfg", "tokio_unstable"]
 
 [target.x86_64-unknown-linux-gnu]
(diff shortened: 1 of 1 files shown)


trinity-1686a (approved):
i'm not too worried about x86, v3 is pretty widespread now (except for Atoms)

i wonder if neoverse-n1 is as supported. On cloud hardware it seems oracle and gcp also have that capabilities, from what i can read, all apple-silicon (though that rarely means a linux-triple). It seems that consumer hardware (raspi 4) doesn't support that featureset `diff <(clang -mcpu=cortex-a72 -print-enabled-extensions) <(clang -mcpu=neoverse-n1 -print-enabled-extensions)`

i wonder if we should document somewhere the mcpu we require, and how to use a different one

PR #6452, opened 2026-05-19
Exclude known split ids; double merge concurrency

nadav-govari:
### Description

Better utilize resources - semaphore for merge executions, everything else is IO. It allows us to run interleave pipelines better and get more net merges out with less resources. Will make this configurable in a future PR.

Also added the list of tracked splits to exclude from the metastore scan. This is necessary because before, we'd always fetch the 5000 oldest splits that are not yet mature. At a certain point we end up retrieving all of the splits that make up pending L3+ merges, and since those are already tracked, it's keeping us from scanning newer splits that are better candidates to be merged.

### How was this PR tested?

### DIFF
--- a/quickwit/quickwit-compaction/src/compactor_supervisor.rs
+++ b/quickwit/quickwit-compaction/src/compactor_supervisor.rs
@@ -35,6 +35,7 @@ use quickwit_proto::indexing::MergePipelineId;
 use quickwit_proto::metastore::MetastoreServiceClient;
 use quickwit_proto::types::NodeId;
 use quickwit_storage::StorageResolver;
+use tokio::sync::Semaphore;
 use tracing::{error, info};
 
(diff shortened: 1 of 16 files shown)


quickwit/quickwit-compaction/src/compactor_supervisor.rs
```diff
@@ -81,11 +84,14 @@ impl CompactorSupervisor {
         event_broker: EventBroker,
         compaction_root_directory: TempDirectory,
     ) -> Self {
+        let num_pipeline_slots = max_concurrent_merge_executions * 2;
         let pipelines = (0..num_pipeline_slots).map(|_| None).collect();
```
nadav-govari, line 88: vec macro

nadav-govari: just fyi @guilload - cant do `vec![None; num_pipeline_slots]` because it clones the given value, which is of type Option<CompactionPipeline>, and CompactionPipeline isn't `Clone`

PR #6453, opened 2026-05-19
Revert to using existing merge flow when standalone compactors isnt explicitly enabled

nadav-govari:
### Description

It would be hard to maintain two parallel performance profiles of standalone compactors running on their own nodes, but then also on indexer nodes - performance profiles, balancing resource usage, etc. for two standalone services that need to share. The existing merge pipelines on indexing pipelines is proven and works, and additionally, all existing workflows are on this pattern; it'll also be easier to revert into this mode if there are problems with standalone compactors.

For that reason, standalone compactors will be explicitly on opt in to start with, as an experimental feature. A pod will only start if the compactor service is explicitly enabled (its not in default services), and if the flag is on. It will bail otherwise.

This restores the existing merge pipelines in indexing code. If the compactors flag is on, it doesnt spawn the merge components, assuming that work is fully delegated to compactors.

### How was this PR tested?

Local pipeline, unit tests, and restored old deleted tests.

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexing_service.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexing_service.rs
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::models::DetachMergePipeline;
+use quickwit_common::io;
+use futures::TryStreamExt;
 use std::collections::{HashMap, HashSet};
 use std::fmt::{Debug, Formatter};
 use std::path::PathBuf;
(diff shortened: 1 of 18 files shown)


quickwit/quickwit-serve/src/lib.rs
```diff
@@ -712,6 +734,7 @@ pub async fn serve_quickwit(
         }
     }
 
+    // this is the search SplitCache, not to be confused with the IndexingSplitCache.
     let split_cache_opt: Option<Arc<SplitCache>> =
```
nadav-govari, line 738: rename the struct

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -197,7 +234,14 @@ impl IndexingService {
             pipeline_uid,
         };
         let index_config = index_metadata.into_index_config();
-        self.spawn_pipeline_inner(ctx, pipeline_id.clone(), index_config, source_config, None)
+        self.spawn_pipeline_inner(
+            ctx,
+            pipeline_id.clone(),
+            index_config,
+            source_config,
+            None,
```
nadav-govari, line 242: make sure

PR #6454, opened 2026-05-20
Add UnicodeSegmenter tokenizer

fulmicoton-dd:
The hidden goal of this is to unify the lambda binary for quickwit and Datadog BYOC logs.

## Summary

- Add `UnicodeSegmenterTokenizer` based on Unicode word boundaries (UAX #29), using the `unicode-segmentation` crate.
- Registered as both `unicode_segmenter` and `datadog` tokenizers in the default tokenizer manager.

## Test plan

- [ ] `cargo nextest run -p quickwit-query` — unit tests for the new tokenizer
- [ ] `cargo clippy --workspace --all-features --tests`

### DIFF
--- a/.claude/skills/simple-pr/SKILL.md
+++ b/.claude/skills/simple-pr/SKILL.md
@@ -42,19 +42,42 @@ Create a short, descriptive branch name based on the changes (e.g., `fix-typo-in
 
 Create and checkout the branch: `git checkout -b {username}/{short-descriptive-name}`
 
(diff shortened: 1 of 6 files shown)


Platane (approved):
👍 

you have changes on on the claude pr skill. They looks good but I am not sure you want then in this PR.

fulmicoton-dd:
@Platane yes that was on purpose.

PR #6466, opened 2026-05-26
Refactoring of the code maintaining ClusterNodes

fulmicoton:
- we now rely on Arc<str> instead of String for NodeId. Chitchat now emits Arc<str>, making cloning cheap
- nesting ClusterMember within ClusterNode. Both struct had a confusing relationship. Now the relationship is clearer. ClusterNode is Member + a channel.
- as a legacy code, we used to maintain a watch channel with the list of ClusterMember. That channel was maintained by own task, generating copies of all members upon cluster modifications. These members were only used in unit test. This PR removes this code that was marked as "to deprecate"

### DIFF
--- a/quickwit/quickwit-control-plane/src/tests.rs
+++ b/quickwit/quickwit-control-plane/src/tests.rs
@@ -73,7 +73,7 @@ pub fn test_indexer_change_stream(
                 ClusterChange::Add(node)
                     if node.enabled_services().contains(&QuickwitService::Indexer) =>
                 {
-                    let node_id = node.node_id().to_owned();
+                    let node_id = node.node_id.clone();
                     let generation_id = node.chitchat_id().generation_id;
                     let indexing_tasks = node.indexing_tasks().to_vec();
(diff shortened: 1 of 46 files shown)


quickwit/quickwit-control-plane/src/tests.rs
```diff
@@ -70,10 +72,8 @@ pub fn test_indexer_change_stream(
         let indexing_clients = indexing_clients.clone();
         Box::pin(async move {
             match cluster_change {
-                ClusterChange::Add(node)
-                    if node.enabled_services().contains(&QuickwitService::Indexer) =>
-                {
-                    let node_id = node.node_id().to_owned();
+                ClusterChange::Add(node) if node.is_service_enabled(QuickwitService::Indexer) => {
```
guilload:
```suggestion
                ClusterChange::Add(node) if node.is_indexer() => {
```

guilload (approved):

quickwit/quickwit-control-plane/src/tests.rs
fulmicoton: yes

PR #6471, opened 2026-05-28
Make standalone compactors config top-level

nadav-govari:
### Description

Makes the standalone compactors config node-level, which makes more sense than in the indexing configs, since its used by three different services: janitor, indexer, compactor.

### How was this PR tested?

Unit tests and fixed other tests.


quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -861,6 +850,11 @@ pub struct NodeConfig {
     pub ingest_api_config: IngestApiConfig,
     pub jaeger_config: JaegerConfig,
     pub compactor_config: CompactorConfig,
+    /// Standalone compactors means that indexer nodes don't run merges.
+    /// This flag affects the behavior of the compactor service, the indexer service,
+    /// and the janitor service (the planner lives there). If this flag is enabled and the
+    /// compactor service is enabled on this node, it'll error. The janitor and indexer will
```
trinity-1686a: The janitor and indexer will?

quickwit/quickwit-integration-tests/src/tests/basic_tests.rs
```diff
@@ -162,47 +162,3 @@ async fn test_multi_nodes_cluster() {
 
     sandbox.shutdown().await.unwrap();
 }
-
-#[tokio::test]
-async fn test_indexer_implicitly_runs_compactor() {
```
trinity-1686a, line 167: why is that test being removed?

nadav-govari: Originally, we were going to force standalone compactors to be false (meaning indexers still had to do merges). To still get merges, we were going to have indexer nodes run both the indexer and compactor service. We decided since then to not go that route, and indexers should no longer implicitly run the compactor service as a result.

quickwit/quickwit-config/src/node_config/mod.rs
nadav-govari: I think im just removing this comment. No need to put whole documentation inline.

PR #6472, opened 2026-05-29
feat(vrl): add `drop_on_abort` option to VRL transform (#5664)

### LINKED ISSUE #5664: Add `drop_on_abort` option to VRL processing

siva-abstract-security:
## Summary

Closes #5664.

Adds a `drop_on_abort` boolean to `TransformConfig` (default `true`, matching Vector's [VRL semantics](https://vector.dev/highlights/2021-04-21-vrl-abort/)). When enabled, documents terminated by a VRL `abort` expression are dropped silently, with no warn log, and surfaced via a distinct counter so operators can distinguish intentional filtering from genuine VRL failures.

## Behavior

| Outcome | `drop_on_abort: true` (default) | `drop_on_abort: false` |
|---|---|---|
| `abort` expression (with or without message) | silent drop → `transform_aborts` → `DocProcessorError::TransformAbort` | warn log → `transform_errors` → `DocProcessorError::Transform` |
| `Terminate::Error` (e.g. `parse_json!` failure) | warn log → `transform_errors` | warn log → `transform_errors` |
| `Ok(doc)` | forwarded to indexer | forwarded to indexer |

`num_invalid_docs()` excludes aborted documents (intentional drops, not invalid); a new `num_dropped_docs()` accessor surfaces them. New Prometheus label value `docs_processed_status="transform_abort"` on the existing `quickwit_doc_processor_docs_total` metric — no dashboard breakage.

## Usage

```yaml
# Filter-style — aborted docs no longer spam warn logs
transform:
  script: |
    if .level == "DEBUG" { abort }
  drop_on_abort: true   # default; can be omitted
```

```yaml
# Strict mode: surface aborts as errors
transform:
  script: ".body = parse_json!(.raw)"
  drop_on_abort: false
```

## Design notes

- **Default `true` (Vector parity).** Anyone calling `abort` already accepts the doc being dropped; they want the noise gone. Strict-mode users can opt back into the old behavior.
- **Separate `TransformAbort` variant.** Aborts are intentional drops, not errors. Carrying that distinction through the error type lets `record_error` route cleanly and keeps `num_invalid_docs()` semantically honest.
- **`TransformConfig::new()` signature unchanged.** Defaults the new field; ~5 callers in cli, control-plane, metastore tests stay untouched.

## Tests — 29 assertions across edge cases, error conditions, error handling, validations

**`TransformConfig` serde (8 inner cases):**
- default applied when field absent (YAML + JSON)
- explicit `true` / `false` roundtrip
- wrong type (string `"true"`) rejected
- unknown field rejected by `deny_unknown_fields`
- `TransformConfig::new()` defaults to true

**`VrlProgram` (8 tests):**
- success passthrough (with and without `drop_on_abort`)
- bare `abort` × {drop_on_abort=true → TransformAbort, false → Transform}
- `abort "with reason"` parity with bare abort
- VRL runtime errors (`Terminate::Error`) always route to Transform regardless of flag
- conditional-abort filter use case
- `Display` impl surfaces the abort reason for log readability
- num_bytes preservation

**`DocProcessor` integration (6 tests through full actor pipeline):**
- single abort + drop_on_abort=true → transform_aborts counter, not forwarded
- single abort + drop_on_abort=false → transform_errors counter
- VRL runtime error × {true, false} → transform_errors (never transform_aborts)
- conditional batch (3 docs, 1 aborts) → counters split + correct downstream forwarding
- **mixed-category batch** (1 valid + 1 abort + 1 json-parse-error + 1 docmapper-error) → all 4 counters independent
- direct `DocProcessorCounters::record_error` routing for all variants

## Test plan

- [x] `cargo test -p quickwit-config --features vrl` — 18 passing (incl. 5 new test fns)
- [x] `cargo test -p quickwit-indexing --features vrl` — 25 passing (incl. 14 new test fns)
- [x] `cargo clippy -p quickwit-config -p quickwit-indexing --features vrl --tests` — clean
- [x] `cargo +nightly fmt --check` — clean
- [x] `cargo check` against quickwit-cli, control-plane, metastore — no breakage
- [x] **End-to-end through real binary**: 5 docs ingested via `quickwit tool local-ingest --transform-script`, REST query returned exactly 3 hits (`KEEP-1`, `KEEP-2`, `KEEP-3`) with the `drop` field stripped and body upcased — confirming the silent-drop behavior on the production path.

### DIFF
--- a/quickwit/quickwit-config/src/source_config/mod.rs
+++ b/quickwit/quickwit-config/src/source_config/mod.rs
@@ -172,6 +172,7 @@ impl crate::TestableForRegression for SourceConfig {
             transform_config: Some(TransformConfig {
                 vrl_script: ".message = downcase(string!(.message))".to_string(),
                 timezone: default_timezone(),
+                drop_on_abort: default_drop_on_abort(),
             }),
             input_format: SourceInputFormat::Json,
         }
@@ -633,20 +634,39 @@ pub struct TransformConfig {
     /// manipulations. Defaults to `UTC` if not timezone is specified.
     #[serde(default = "default_timezone")]
     timezone: String,
+
+    /// When `true` (the default, matching Vector semantics), documents that are explicitly
+    /// terminated via the VRL `abort` expression are dropped silently and routed to the
+    /// `transform_abort` counter. When `false`, an aborted document is treated identically
+    /// to an unexpected VRL runtime error: a warning is logged and the document is counted
+    /// as a transform error.
+    #[serde(default = "default_drop_on_abort")]
+    drop_on_abort: bool,
 }
 
 fn default_timezone() -> String {
     "UTC".to_string()
 }
 
+fn default_drop_on_abort() -> bool {
+    true
+}
+
 impl TransformConfig {
     /// Creates a new [`TransformConfig`] instance from the provided VRL script and optional
-    /// timezone.
+    /// timezone. `drop_on_abort` defaults to `true`.
     pub fn new(vrl_script: String, timezone_opt: Option<String>) -> Self {
         Self {
             vrl_script,
             timezone: timezone_opt.unwrap_or_else(default_timezone),
+            drop_on_abort: default_drop_on_abort(),
         }
+    }
+
+    /// Returns whether documents aborted via the VRL `abort` expression should be silently
+    /// dropped (no warn log, separate counter) instead of being treated as transform errors.
+    pub fn drop_on_abort(&self) -> bool {
+        self.drop_on_abort
     }
 
     #[cfg(feature = "vrl")]
@@ -709,7 +729,14 @@ impl TransformConfig {
         Self {
             vrl_script: vrl_script.to_string(),
             timezone: default_timezone(),
+            drop_on_abort: default_drop_on_abort(),
         }
+    }
+
+    #[cfg(any(test, feature = "testsuite"))]
+    pub fn with_drop_on_abort(mut self, drop_on_abort: bool) -> Self {
+        self.drop_on_abort = drop_on_abort;
+        self
     }
 }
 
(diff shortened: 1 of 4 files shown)


siva-abstract-security:
Thanks @chatgpt-codex-connector — both findings were real and are addressed in c59db5d64.

**P2 #1 — `drop_on_abort` aborts were still warning.** You're right: `VrlProgram::transform_doc` skipped the inner `warn!`, but `DocProcessor::process_raw_doc` then logged the returned `Err` via `rate_limited_warn!`, so the supposedly silent path was anything but. Added `DocProcessorError::should_warn()` (returns `false` only for `TransformAbort`) and gated the rate-limited log on it. New unit test `test_doc_processor_error_should_warn_excludes_transform_abort` locks this down across all variants.

**P2 #2 — aborts were inflating the indexed-doc total.** Confirmed via `IndexingStatistics::add_actor_counters` → `num_docs += num_processed_docs()` → CLI computes `indexed = num_docs - num_invalid_docs`. Since aborts were intentionally excluded from `num_invalid_docs`, they showed up as indexed. Removed `transform_aborts` from `num_processed_docs()` and added a separate `num_dropped_docs` field on `IndexingStatistics` so the abort count remains visible in API observations without polluting the indexed math. Extended the mixed-category batch test to pipe through `IndexingStatistics` and assert `num_docs - num_invalid_docs == 1` (the one doc actually indexed) with `num_dropped_docs == 1` surfacing the abort separately.

PR #6478, opened 2026-05-30
fix(storage): restore TLS for opendal GCS backend after 0.55→0.56 bump

### LINKED ISSUE #6477: GCS split uploads fail with invalid URL, scheme is not http since opendal 0.56 (reqwest built without TLS)pr
**Describe the bug**

Indexing to a **GCS (`gs://`) backend** fails: the indexer cannot upload splits and every
request to `https://storage.googleapis.com` errors with:

```
error sending request for url (https://storage.googleapis.com/<bucket>/...):
client error (Connect): invalid URL, scheme is not http
```

Root cause is a dependency/feature regression, not configuration. `quickwit/Cargo.toml` declares:

```toml
opendal = { version = "0.56", default-features = false }   # line 365
```

`opendal` 0.56's `default` feature set includes `reqwest-rustls-tls`
(`opendal-core/reqwest-rustls-tls` → `reqwest/rustls`), so `default-features = false` removes
opendal's only TLS backend. This used to be harmless: opendal 0.55 depended on `reqwest 0.12`, the
same major as the workspace `reqwest` (`quickwit/Cargo.toml:231`, built with `rustls-tls`), so Cargo
feature-unification gave opendal a TLS-capable client anyway. Since #6432 bumped opendal to **0.56**,
opendal depends on **`reqwest 0.13`** — a different major that does **not** unify with `reqwest 0.12`.
opendal's `reqwest 0.13` is therefore compiled with **no TLS backend** (only the `stream` feature),
and a TLS-less reqwest uses hyper-util's plain `HttpConnector`, which rejects any `https://` URL with
`invalid URL, scheme is not http` (before any network I/O).

Only the GCS backend is affected (`quickwit-storage/Cargo.toml:84`:
`gcs = ["dep:opendal", "opendal/services-gcs"]` enables no TLS feature). S3 uses the AWS SDK; Azure's
`azure_storage_blobs` sets `enable_reqwest_rustls`.

Regression introduced by **#6432 "Upgrade Rust and dependencies"** (opendal `0.55 → 0.56`).

1. Build quickwit from `main` (or run `quickwit/quickwit:edge`) — built with the default
   `opendal = { default-features = false }` and the `gcs` feature.
2. Create an index whose storage is GCS, i.e. `index_uri: gs://<bucket>/<index>` (or a node
   `default_index_root_uri` on `gs://`).
3. Ingest a few documents and wait for a commit so the indexer publishes a split.
4. The split upload fails; the indexer logs `... Connect: invalid URL, scheme is not http`.

Minimal standalone reproduction (no cluster, no credentials — pins the root cause to the
dependency graph). It mirrors quickwit's exact declarations and goes red on the buggy graph, green
once opendal's TLS feature is restored:

```toml
# Cargo.toml
[package]
name = "tls-repro"
version = "0.0.0"
edition = "2021"

[dependencies]
# Mirror quickwit's workspace declarations:
reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
opendal = { version = "0.56", default-features = false, features = ["services-http"] }
tokio   = { version = "1", features = ["macros", "rt-multi-thread"] }
```

```rust
// tests/repro.rs
use opendal::{services::Http, Operator};

#[tokio::test]
async fn opendal_can_make_https_requests() {
    let op = Operator::new(Http::default().endpoint("https://storage.googleapis.com"))
        .unwrap()
        .finish();
    if let Err(err) = op.read("robots.txt").await {
        let msg = format!("{err:?}");
        assert!(
            !msg.contains("invalid URL, scheme is not http"),
            "opendal's reqwest has no TLS backend: {msg}"
        );
    }
}
```

`cargo test` → fails with `client error (Connect): invalid URL, scheme is not http`.
`cargo tree -e features -i reqwest@0.13` shows opendal's reqwest carries only the `stream` feature —
no `rustls`/`native-tls`.

**Expected behavior**

GCS split uploads succeed over HTTPS (as they did before the opendal `0.55 → 0.56` bump).

Restoring opendal's TLS backend fixes it:

```toml
opendal = { version = "0.56", default-features = false, features = ["reqwest-rustls-tls"] }
```

After this, opendal's `reqwest 0.13` pulls in `hyper-rustls`/`rustls`/`tokio-rustls`, the standalone
test makes a real HTTPS round-trip (gets a normal `404 NoSuchBucket` from GCS instead of the scheme
error), and split uploads work again.

**Configuration:**

   ```
   Quickwit 0.8.0
   ```
   Reproduced from source on `main` and on the published `quickwit/quickwit:edge` image (built from
   every `main` merge), i.e. any build after #6432.

2. The index_config.yaml

   ```yaml
   version: 0.8
   index_id: example-gcs-index
   # The trigger is the gs:// (GCS) storage backend; the doc mapping is irrelevant.
   index_uri: gs://my-bucket/example-gcs-index
   doc_mapping:
     field_mappings:
       - name: timestamp
         type: datetime
         fast: true
       - name: message
         type: text
     timestamp_field: timestamp
   indexing_settings:
     commit_timeout_secs: 30
   search_settings:
     default_search_fields: [message]
   ```

fmassot:
## Summary

- `opendal 0.56` moved its internal HTTP client from `reqwest 0.12` to `reqwest 0.13` (different Cargo semver major)
- With `default-features = false`, opendal's `reqwest 0.13` was compiled without any TLS backend — only the `stream` feature
- Cargo no longer unifies features between the workspace `reqwest 0.12` (which has `rustls-tls`) and opendal's `reqwest 0.13`, so TLS was silently missing
- Every GCS request to `https://storage.googleapis.com` failed immediately with `invalid URL, scheme is not http`

Fix: explicitly enable `reqwest-rustls-tls` on the opendal workspace declaration. This feature was already in opendal's defaults but was stripped by `default-features = false`.

Only GCS is affected — S3 uses the AWS SDK, Azure's `azure_storage_blobs` sets `enable_reqwest_rustls` directly.

Regression introduced by #6432. Closes #6477.

## Test plan

- [x] `cargo tree -e features -i reqwest:0.13 -p quickwit-storage --features gcs` shows `reqwest feature "rustls"` in the output
- [x] `cargo build -p quickwit-storage --features gcs` passes
- [x] GCS index ingest + split upload succeeds end-to-end

### DIFF
--- a/LICENSE-3rdparty.csv
+++ b/LICENSE-3rdparty.csv
@@ -165,6 +165,7 @@ cobs,https://github.com/jamesmunns/cobs.rs,MIT OR Apache-2.0,"Allen Welkie <>, J
 codespan-reporting,https://github.com/brendanzab/codespan,Apache-2.0,Brendan Zabarauskas <bjzaba@yahoo.com.au>
 colorchoice,https://github.com/rust-cli/anstyle,MIT OR Apache-2.0,The colorchoice Authors
 colored,https://github.com/mackwic/colored,MPL-2.0,Thomas Wickham <mackwic@gmail.com>
+combine,https://github.com/Marwes/combine,MIT,Markus Westerlind <marwes91@gmail.com>
 comfy-table,https://github.com/nukesor/comfy-table,MIT,Arne Beer <contact@arne.beer>
 community-id,https://github.com/traceflight/rs-community-id,MIT OR Apache-2.0,Julian Wang <traceflight@outlook.com>
 compression-codecs,https://github.com/Nullus157/async-compression,MIT OR Apache-2.0,"Wim Looman <wim@nemo157.com>, Allen Bui <fairingrey@gmail.com>"
@@ -419,6 +420,10 @@ jiff,https://github.com/BurntSushi/jiff,Unlicense OR MIT,Andrew Gallant <jamslam
 jiff-static,https://github.com/BurntSushi/jiff,Unlicense OR MIT,Andrew Gallant <jamslam@gmail.com>
 jiff-tzdb,https://github.com/BurntSushi/jiff,Unlicense OR MIT,Andrew Gallant <jamslam@gmail.com>
 jiff-tzdb-platform,https://github.com/BurntSushi/jiff,Unlicense OR MIT,Andrew Gallant <jamslam@gmail.com>
+jni,https://github.com/jni-rs/jni-rs,MIT OR Apache-2.0,jni team
+jni-macros,https://github.com/jni-rs/jni-rs,MIT OR Apache-2.0,The jni-macros Authors
+jni-sys,https://github.com/jni-rs/jni-sys,MIT OR Apache-2.0,"Steven Fackler <sfackler@gmail.com>, Robert Bragg <robert@sixbynine.org>"
+jni-sys-macros,https://github.com/jni-rs/jni-sys,MIT OR Apache-2.0,Robert Bragg <robert@sixbynine.org>
 jobserver,https://github.com/rust-lang/jobserver-rs,MIT OR Apache-2.0,Alex Crichton <alex@alexcrichton.com>
 js-sys,https://github.com/wasm-bindgen/wasm-bindgen/tree/master/crates/js-sys,MIT OR Apache-2.0,The wasm-bindgen Developers
 json_comments,https://github.com/tmccombs/json-comments-rs,Apache-2.0,Thayne McCombs <astrothayne@gmail.com>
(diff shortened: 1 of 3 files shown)


palkakrzysiek (approved):
Hi @fmassot, yes, this PR fixes the issue. Thanks for opening it.

The worrisome aspect of the issue is that no test caught the problem with connection to GCS after dependency upgrade. Please consider modifying existing tests to use secure connection, I put an example in https://github.com/quickwit-oss/quickwit/pull/6479/changes, it targets this PR branch (`fix/gcs-opendal-tls`) so feel free to merge if it makes sense to you.

trinity-1686a (approved):

fmassot:
@palkakrzysiek @trinity-1686a just tested on a GCS bucket with a fresh service account. Merging now.

PR #6487, opened 2026-06-03
fix(config): set MD5 checksums for DigitalOcean and GCS flavors since they don't support CRC32C, and explicitly set CRC32C for MinIO

fulmicoton-dd:
## Summary

- DigitalOcean does not support CRC32C: explicitly set `Md5` now that `Crc32c` is the default
- GCS was previously `Disabled` (which predated the CRC32C default); switch to `Md5` since it also lacks CRC32C S3 SDK support
- MinIO: explicitly set `Crc32c` for clarity (previously fell through to the new default)
- Add checksum-related assertions to the per-flavor serde tests

Follows up on #6442 which made CRC32C the default checksum algorithm.

## Test plan
- [ ] `cargo nextest run -p quickwit-config` passes
- [ ] Clippy and fmt clean

### DIFF
--- a/quickwit/quickwit-config/src/storage_config.rs
+++ b/quickwit/quickwit-config/src/storage_config.rs
@@ -370,20 +370,26 @@ impl S3StorageConfig {
             Some(StorageBackendFlavor::DigitalOcean) => {
                 self.force_path_style_access = true;
                 self.disable_multi_object_delete = true;
+                // doesn't support CRC32C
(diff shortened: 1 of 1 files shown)


guilload:
I remember Nadav mentioning that DO supports crc32c:
https://github.com/quickwit-oss/quickwit/pull/6442#discussion_r3275672212

PR #6492, opened 2026-06-04
fix(storage): avoid tokio::io::copy memory growth in OpenDAL GCS put

fulmicoton-dd:
## Summary

- Observed strong memory drift during indexing with the OpenDAL GCS backend
- Root cause: `tokio::io::copy`'s internal buffering strategy causes unbounded memory growth in this code path
- Fix: replace `tokio::io::copy` with an explicit read/write loop that accumulates at least 8 KB before flushing
- Same approach as https://github.com/apache/opendal/pull/7217/
- Added unit tests for the loop covering: batching of small reads, empty input, sub-threshold input, exact-threshold alignment, and large single reads

## Test plan

- [ ] `cargo nextest run -p quickwit-storage test_copy_read_write_loop` — all 5 unit tests pass
- [ ] Integration test with GCS backend to confirm memory profile is stable under sustained indexing

the unit test have been generated by claude.

### DIFF
--- a/quickwit/quickwit-storage/src/opendal_storage/base.rs
+++ b/quickwit/quickwit-storage/src/opendal_storage/base.rs
@@ -435,7 +435,7 @@ mod tests {
         copy_read_write_loop(&mut reader, &mut writer)
             .await
             .unwrap();
-        assert_eq!(writer.data, vec![42u8; 500]);
+        assert_eq!(writer.data, [42u8; 500]);
         assert_eq!(writer.writes, vec![500]);
     }
 
--- a/quickwit/quickwit-storage/src/opendal_storage/base.rs
+++ b/quickwit/quickwit-storage/src/opendal_storage/base.rs
@@ -12,16 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::fmt;
 use std::ops::Range;
 use std::path::Path;
+use std::{fmt, io};
 
 use async_trait::async_trait;
 use futures::AsyncWriteExt as FuturesAsyncWriteExt;
 use opendal::{DeleteInput, IntoDeleteInput, Operator};
 use quickwit_common::uri::Uri;
 use quickwit_metrics::HistogramTimer;
-use tokio::io::{AsyncRead, AsyncWriteExt as TokioAsyncWriteExt};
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-storage/src/opendal_storage/base.rs
```diff
@@ -92,6 +92,37 @@ impl OpendalStorage {
...
 
+/// We spotted a ever growing usage of RAM in the opendal GCS implementation.
+/// We are using the same fix as
+/// https://github.com/apache/opendal/pull/7217/
+async fn copy_read_write_loop<R, W>(input: &mut R, output: &mut W) -> io::Result<()>
+where
+    R: tokio::io::AsyncRead + Unpin,
+    W: tokio::io::AsyncWrite + Unpin,
+{
+    // Only flush to `output` once we have accumulated at least `WRITE_THRESHOLD`
```
guilload, line 103: I don't think it makes a difference in this specific context because I know for a fact that the reader returns 4 KiB on each read and the writer is an in-memory buffer, but at least we're safe if the implementation changes under our feet.

quickwit/quickwit-storage/src/opendal_storage/base.rs
```diff
@@ -270,3 +302,169 @@ impl From<opendal::Error> for StorageResolverError {
...
+    async fn test_copy_read_write_loop_below_threshold_produces_single_write() {
+        // 5 chunks of 100 bytes = 500 bytes, well below the 8 KB threshold.
+        // The single flush should happen on EOF.
+        let chunks = vec![vec![42u8; 100]; 5];
+        let mut reader = ChunkedReader::new(chunks);
+        let mut writer = RecordingWriter::default();
+        copy_read_write_loop(&mut reader, &mut writer)
+            .await
+            .unwrap();
+        assert_eq!(writer.data, vec![42u8; 500]);
```
guilload: nit: A `Vec` can be compared directly to an array.
```suggestion
        assert_eq!(writer.data, [42u8; 500]);
```

guilload (approved):

quickwit/quickwit-storage/src/opendal_storage/base.rs
fulmicoton-dd: Yes, exactly

PR #6493, opened 2026-06-04
make notifying merge planner best effort

mattmkim:
### Description
Indexing is coupled with merge planner pipeline availability - if we can't notify the merge planner, than indexing stops. For example, due to the pagination bug, indexing just stopped in one of the quickhouse org cluster.

For now, make the notification best effort. On restarts, we'll fetch all non merged files to compact. This kinda sucks, because restarts might not happen for a while, and depending on the merge policy, not sure if an old, uncompacted file will merge well. We're eventually moving towards a better/pull based model, with compaction as its own service. 

### How was this PR tested?


fulmicoton:
@mattmkim  Nadav has been working on externalizing merging to dedicated nodes.
The merge planning works by polling the metastore for new splits.

This problem should disappear once shipped.
(it is currrently in "test phase")

PR #6498, opened 2026-06-08
Update OpenTelemetry metrics dependencies

shuheiktgw:
## Summary

This PR aims to solve the same problem as https://github.com/quickwit-oss/quickwit/pull/6465, but with a different approach. Instead of reimplementing the libraries ourselves, we are opening PRs to fix the issues upstream and patching them on our side until those PRs are merged and released.

* Update the OpenTelemetry crates to 0.32 and patch them with the SDK fix from https://github.com/open-telemetry/opentelemetry-rust/pull/3545.
* Use metrics-exporter-otel 0.3.2 and patch it with the fix from https://github.com/palindrom615/metrics-exporter-otel/pull/2 so that we can use OpenTelemetry 0.32.

### DIFF
--- a/quickwit/Cargo.toml
+++ b/quickwit/Cargo.toml
@@ -170,7 +170,7 @@ lru = "0.18"
 matches = "0.1"
 md5 = "0.8"
 metrics = "0.24"
-metrics-exporter-otel = "0.3"
+metrics-exporter-otel = "0.3.2"
 metrics-exporter-prometheus = { version = "0.18", default-features = false }
 metrics-util = "0.20"
 mime_guess = "2.0"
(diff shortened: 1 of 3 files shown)


quickwit/Cargo.toml
```diff
@@ -416,8 +416,22 @@ tantivy-fst = "0.5"
...
+metrics-exporter-otel = { git = "https://github.com/shuheiktgw/metrics", rev = "238329060a300f653accf968cca831ac44a8d409" }
+opentelemetry = { git = "https://github.com/shuheiktgw/opentelemetry-rust", rev = "7cf7d697626a591394d75c1a7efddcb55198caa6" }
+opentelemetry-appender-tracing = { git = "https://github.com/shuheiktgw/opentelemetry-rust", rev = "7cf7d697626a591394d75c1a7efddcb55198caa6" }
+opentelemetry-http = { git = "https://github.com/shuheiktgw/opentelemetry-rust", rev = "7cf7d697626a591394d75c1a7efddcb55198caa6" }
+opentelemetry-otlp = { git = "https://github.com/shuheiktgw/opentelemetry-rust", rev = "7cf7d697626a591394d75c1a7efddcb55198caa6" }
+opentelemetry-proto = { git = "https://github.com/shuheiktgw/opentelemetry-rust", rev = "7cf7d697626a591394d75c1a7efddcb55198caa6" }
+opentelemetry_sdk = { git = "https://github.com/shuheiktgw/opentelemetry-rust", rev = "7cf7d697626a591394d75c1a7efddcb55198caa6" }
 sasl2-sys = { git = "https://github.com/quickwit-oss/rust-sasl/", rev = "085a4c7" }
 
+[patch."https://github.com/shuheiktgw/metrics"]
```
Mallets, line 431: I understand the patch on `opentelemetry` but why should your `metrics` fork be used here?

shuheiktgw: `palindrom615/metrics-exporter-otel` is a fork of `metrics-rs/metrics`, and I’ve already forked `metrics-rs/metrics`, so I can’t fork `palindrom615/metrics-exporter-otel` because GitHub doesn’t allow users to fork the same repo tree multiple times.

But I may just delete my `metrics-rs/metrics` fork and fork `palindrom615/metrics-exporter-otel` directly instead, which might simplify things here. Let me try 🙏

PR #6499, opened 2026-06-09
Update OpenTelemetry metrics dependencies

shuheiktgw:
## Summary

This PR aims to solve the same problem as https://github.com/quickwit-oss/quickwit/pull/6465, but with a different approach. Instead of reimplementing the libraries ourselves, we are opening PRs to fix the issues upstream and patching them on our side until those PRs are merged and released.

* Update the OpenTelemetry crates to 0.32 and patch them with the SDK fix from https://github.com/open-telemetry/opentelemetry-rust/pull/3545.
* Use metrics-exporter-otel 0.3.2 and patch it with the fix from https://github.com/palindrom615/metrics-exporter-otel/pull/3 so that we can use OpenTelemetry 0.32.

### DIFF
--- a/quickwit/Cargo.toml
+++ b/quickwit/Cargo.toml
@@ -170,7 +170,7 @@ lru = "0.18"
 matches = "0.1"
 md5 = "0.8"
 metrics = "0.24"
-metrics-exporter-otel = "0.3"
+metrics-exporter-otel = "0.3.2"
 metrics-exporter-prometheus = { version = "0.18", default-features = false }
 metrics-util = "0.20"
 mime_guess = "2.0"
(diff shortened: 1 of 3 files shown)


Mallets (approved):

quickwit/Cargo.toml
```diff
@@ -417,6 +417,13 @@ encoding_rs = "=0.8.35"
 
 [patch.crates-io]
 sasl2-sys = { git = "https://github.com/quickwit-oss/rust-sasl/", rev = "085a4c7" }
+metrics-exporter-otel = { git = "https://github.com/shuheiktgw/metrics-exporter-otel", rev = "b4032665307007ac03ac132fd60850d8c8b7af19" }
+opentelemetry = { git = "https://github.com/shuheiktgw/opentelemetry-rust", rev = "4279a0115915cdb0a18259d933f58a287b9188eb" }
```
PSeitz, line 421: this should have a comment why they are here and when they can be removed

PR #6502, opened 2026-06-10
retry S3 dispatch failures with TransientError kind

fmassot:
## Problem

On a large cluster, still observing 22 "failed to upload split" logs on transient s3 errors for which we should retry.

Pattern of logs:

```
failed uploading key *.split in bucket s3://indexes/indexes/datadog  Caused by:     0: storage error(kind=Internal, source=dispatch failure: other: client error (SendRequest): connection closed before message completed (DispatchFailure(DispatchFailure { source: ConnectorError { kind: Other(Some(TransientError)), source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(IncompleteMessage)), connection: Unknown } })))     1: dispatch failure: other: client error (SendRequest): connection closed before message completed (DispatchFailure(DispatchFailure { source: ConnectorError { kind: Other(Some(TransientError)), source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(IncompleteMessage)), connection: Unknown } }))

```

## What this PR does

`DispatchFailure` errors with `kind: Other(Some(TransientError))` are now classified as retryable

## Test plan

- [ ] Existing unit tests pass
- [ ] Integration tests with S3 pass

### DIFF
--- a/quickwit/quickwit-aws/src/error.rs
+++ b/quickwit/quickwit-aws/src/error.rs
@@ -39,7 +39,14 @@ where E: AwsRetryable
             SdkError::DispatchFailure(failure) => {
                 failure.is_io()
                     || failure.is_timeout()
-                    || matches!(failure.as_other(), Some(ErrorKind::TransientError))
+                    || matches!(
+                        failure.as_other(),
+                        Some(
+                            ErrorKind::TransientError
+                                | ErrorKind::ThrottlingError
+                                | ErrorKind::ServerError
+                        )
+                    )
             }
             SdkError::ResponseError(_) => true,
             SdkError::ServiceError(error) => error.err().is_retryable(),
--- a/quickwit/quickwit-aws/src/error.rs
+++ b/quickwit/quickwit-aws/src/error.rs
@@ -16,6 +16,7 @@
 
 use aws_runtime::retries::classifiers::{THROTTLING_ERRORS, TRANSIENT_ERRORS};
 use aws_sdk_s3::error::SdkError;
+use aws_smithy_types::retry::ErrorKind;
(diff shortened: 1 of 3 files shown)


guilload (approved):

quickwit/quickwit-aws/src/error.rs
```diff
@@ -35,7 +36,11 @@ where E: AwsRetryable
         match self {
             SdkError::ConstructionFailure(_) => false,
             SdkError::TimeoutError(_) => true,
-            SdkError::DispatchFailure(failure) => failure.is_io() || failure.is_timeout(),
+            SdkError::DispatchFailure(failure) => {
+                failure.is_io()
+                    || failure.is_timeout()
+                    || matches!(failure.as_other(), Some(ErrorKind::TransientError))
```
guilload: I think there are other variants in there that we should retry too:
```rust
 /// An error where the server explicitly told the client to back off, such as a 429 or 503 HTTP error.
ThrottlingError,

/// Server error that isn't explicitly throttling but is considered by the client
/// to be something that should be retried.
ServerError,
```

guilload: There's also `as_connector_error` on dispatch failure that has a bunch of retryable shit. Who knew retries were this hard.

Nvm: `is_io` and `is_timeout` is already using the underlying connector error.

rdettai-sk:
@fmassot @trinity-1686a I see you are struggling with failures in indexers when having connectivity issues. FYI we also increased the metastore retries for publish an stage calls (currently ~3s) because similarly to what you saw here, things can get messy when pipelines are restarted to quickly on transient metastore errors.

fmassot:
thanks for the heads up @rdettai-sk

PR #6504, opened 2026-06-11
Delete shards only when notified via gossip

guilload:
### Description
There are paths (deleting empty shards on init, truncation at eof) where an ingester can delete shards without the control-plane eventually knowing. It is actually easy to reproduce locally with a combination of ingest with `commit=force` and abrupt Ctrl+C.

This PR makes deleting shards via gossip the only path. There's still a very subtle path where this issue can happen. This will be the object of another PR.

### How was this PR tested?
Added unit tests

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/state.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/state.rs
@@ -39,7 +39,6 @@ use super::models::IngesterShard;
 use super::rate_meter::RateMeter;
 use super::replication::{ReplicationStreamTaskHandle, ReplicationTaskHandle};
 use super::wal_capacity_tracker::WalCapacityTracker;
-use crate::ingest_v2::mrecordlog_utils::queue_position_range;
 use crate::mrecordlog_async::MultiRecordLogAsync;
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -254,52 +253,60 @@ impl IngesterState {
...
+                        } else {
+                            Position::offset(*position_range.start() - 1)
+                        };
+                        (
+                            replication_position_inclusive,
+                            truncation_position_inclusive,
+                        )
+                    }
+                    // The queue is empty.
+                    None => (Position::Beginning, Position::Beginning),
```
nadav-govari: discussed offline: this isn't ideal. We'll explore if this is something we can get from the mrecordlog.

guilload: Fixed. Now relying on the queues summary API.

nadav-govari: Thanks for making the change.

nadav-govari (approved):

PR #6505, opened 2026-06-12
Change build_info metric to gauge

shuheiktgw:
### Description
Change the `build_info` metric registration from a counter to a gauge, while keeping the same metric name and labels. Otherwise, nodes report the counter as `1` when they start up and `0` after that, which is not what we expect from the `build_info` metric.

### DIFF
--- a/quickwit/quickwit-cli/src/metrics.rs
+++ b/quickwit/quickwit-cli/src/metrics.rs
@@ -13,10 +13,10 @@
 // limitations under the License.
 
 use quickwit_common::metrics::exponential_buckets;
-use quickwit_metrics::{LazyCounter, LazyHistogram, counter, labels, lazy_counter, lazy_histogram};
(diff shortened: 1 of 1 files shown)


Mallets:
I'm not sure to understand:

> Otherwise, nodes report the counter as 1 when they start up and 0 after that, which is not what we expect from the build_info metric.

A counter can only be incremented and never decremented, how it could be reported as `0` after it was incremented to `1`?

shuheiktgw:
@Mallets Sorry, I skipped too much of the explanation...

What we want from the build_info metric is for each node to keep reporting 1, just like the Prometheus build_info metric. That does not happen if we use a counter, because a counter represents the increase within a given interval. In this case, the interval is the time between metric exports. As a result, build_info goes back to 0 after the first export interval, which is not what we want and differs from the Prometheus build_info behavior.

Mallets (approved):

PR #6508, opened 2026-06-12
Hot-reload TLS cert periodically or on SIGHUP

guilload:
Hot-reload cert and key (but not the CA, still need a restart). We terminate TLS ourselves instead of using tonic's startup-only TLS config. In-flight connections keep their negotiated cert. Periodic poll, default to 5 minutes, or immediately on SIGHUP.

Add `quickwit-transport` crate in order to not duplicate a bunch of config structs (because `quickwit-transport` cannot depend on ``quickwit-config`)

Added unit tests and integration tests.


docs/configuration/node-config.md
```diff
@@ -87,6 +89,46 @@ We advise changing the default value of 20 MiB only if you encounter the followi
...
 
+## TLS configuration
+
+Both the REST API (`rest.tls`) and the internal gRPC services (`grpc.tls`) can be secured with TLS, optionally with mutual TLS (mTLS). The two sections share the same properties:
+
+| Property | Description | Default value |
+| --- | --- | --- |
+| `cert_path` | Path to the PEM-encoded X.509 certificate (or chain) presented by the server. Setting this enables TLS. | |
+| `key_path` | Path to the PEM-encoded private key matching `cert_path`. | |
+| `ca_path` | Path to the PEM-encoded CA certificate. Used by the server to validate client certificates when `verify_client_cert` is enabled, and by the gRPC client to validate peer certificates. | |
```
trinity-1686a: from the look of it, we support multiple ca per file, this could be mentioned, it's a useful feature when rotating ca (rolling restart everyone with a 2nd CA, then rolling restart without the expired one)

quickwit/quickwit-transport/src/channel.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+impl Default for ChannelFactory {
+    fn default() -> Self {
+        Self {
+            keep_alive_opt: None,
+            mode: TransportMode::Plaintext,
+        }
+    }
+}
```
trinity-1686a: this seem to be used only in tests and called explicitly, i'd prefer it not be the Default trait, that's a good way of making a plaintext transport somewhere because we derived Default on a bigger struct

quickwit/quickwit-transport/src/tls.rs
```diff
@@ -0,0 +1,463 @@
...
+    } else {
+        builder.with_no_client_auth()
+    };
+    client_config.alpn_protocols = vec![b"h2".to_vec()];
+    Ok(Arc::new(client_config))
+}
+
+/// Loads the CA certificate(s) at `ca_path` into a [`RootCertStore`].
+fn load_root_cert_store(ca_path: &str) -> anyhow::Result<RootCertStore> {
+    let ca_certs = load_certs(ca_path)?;
```
trinity-1686a: can we better bubble up this error? right now if tls is configured but not CA is set on gRPC, i think we only error out `failed to open : file not found` (no path, the default is empty), without mentioning what that file is supposed to refer too

quickwit/quickwit-transport/src/accept.rs
```diff
@@ -0,0 +1,80 @@
...
+                match tokio::time::timeout(TLS_HANDSHAKE_TIMEOUT, tls_acceptor.accept(tcp_stream))
+                    .await
+                {
+                    Ok(Ok(tls_stream)) => Some(Ok::<_, io::Error>(tls_stream)),
+                    Ok(Err(error)) => {
+                        error!("failed to perform TLS handshake: {error:#}");
+                        None
+                    }
+                    Err(_elapsed) => {
+                        error!("timed out while performing TLS handshake");
```
trinity-1686a: this should probably be warn only

trinity-1686a (approved):

quickwit/quickwit-transport/src/channel.rs
guilload: Agreed.

docs/configuration/node-config.md
guilload: We do support multiple ca per file. Will document.

PR #6509, opened 2026-06-12
Fix shard ownership bug when decommissioning indexer

nadav-govari:
### Description
Indexing plan broadcast bug when an indexer retires causes crash loop of multiple indexers trying to acquire and index one shard, which slows down indexing and decommissioning dramatically and crashes indexing pipelines (but doesnt corrupt data or cause any data loss).

### In detail
When an ingester signals that it’s entering the retiring state, it’s IngesterStatus is no longer Ready, excluding it from being included in the next indexing plan: https://github.com/quickwit-oss/quickwit/blob/f3880d52dfeff1f4c3795e0e046be1dc1f89b433/quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs#L390 

The indexers used to build the plan are the ones then iterated over to send the new indexing plan to: https://github.com/quickwit-oss/quickwit/blob/f3880d52dfeff1f4c3795e0e046be1dc1f89b433/quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs#L417 

The problem with this is that the decommissioning indexers are never notified of the new plan (since they’re not in the new plan). The new indexing plan reassigns the shards from this indexer to a different one, which calls AcquireShard with its own publish token, which updates the metastore. It then fetches and indexes the shard and publishes a split.

Since the decommissioning indexer doesn’t know that it’s supposed to stop indexing the shard, it continues indexing the shard as well, and tries to publish a split, but fails because the publish token is different. This triggers the killswitch and the pipeline restarts. This is where the bug is: since it never receives the new indexing plan, the IndexingService still has this shard in its assigned shards - this causes it to call AcquireShard itself, with a new publish token. So now it’s the owner of the shard. The new indexer tries to publish its next split, and now it fails with the same publish token problem, and it restarts and reacquires the shard for itself.

This ping pong happens until either the shard is completely indexed and reaches EOF, or until the decommissioning timeout is hit and the decommissioning indexer is force killed, at which point the new indexer takes over and continues indexing the shard as intended. 

Luckily the publish token works completely as intended and prevents us from having contention or corrupted data. It just means decommissioning takes forever and indexing is much slower.

Fix: Send the new indexing plan to all nodes in the ingester pool, not just the nodes in the plan

### Tests
Added unit test, added integration test.

There’s two additional future improvements that could aid here:

AcquireShards should be based on the most recent indexing plan

Currently, all of a retiring indexer’s shards are indexed on other indexers, and the retiring indexer sits idle. The indexing plan could be modified so that a decommissioning indexer is allowed to index its (and only its) local shards.

### DIFF
--- a/quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
+++ b/quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
@@ -431,13 +431,25 @@ impl IndexingScheduler {
                 .indexer(indexer.node_id.as_str())
                 .unwrap_or(&[])
                 .to_vec();
-            // We don't want to block on a slow indexer so we apply this change asynchronously
+            // We don't want to block on a slow indexer so we apply this change asynchronously.
+            // Bound the apply only for retiring/decommissioning indexers, so a slow or unreachable
+            // draining node can't hold the change-notification guard; ready indexers get no
+            // timeout.
+            let apply_deadline = matches!(
+                indexer.ingester_status,
+                IngesterStatus::Retiring | IngesterStatus::Decommissioning
+            )
+            .then_some(APPLY_INDEXING_PLAN_TIMEOUT);
             let notify_on_drop = notify_on_drop.clone();
             tokio::spawn(async move {
                 let client = indexer.client.clone();
                 let apply_plan_fut =
                     client.apply_indexing_plan(ApplyIndexingPlanRequest { indexing_tasks });
-                match tokio::time::timeout(APPLY_INDEXING_PLAN_TIMEOUT, apply_plan_fut).await {
+                let apply_result = match apply_deadline {
+                    Some(timeout) => tokio::time::timeout(timeout, apply_plan_fut).await,
+                    None => Ok(apply_plan_fut.await),
+                };
+                match apply_result {
                     Ok(Ok(_)) => {}
                     Ok(Err(error)) => {
                         warn!(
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -408,40 +408,38 @@ impl IndexingScheduler {
...
-                            generation_id=indexer.generation_id,
-                            "failed to apply indexing plan to indexer"
-                        );
-                    }
-                    drop(notify_on_drop);
+            tokio::spawn(async move {
+                if let Err(error) = indexer
+                    .client
+                    .clone()
+                    .apply_indexing_plan(ApplyIndexingPlanRequest { indexing_tasks })
```
guilload: Yes, let's make sure there's a timeout layer on top of the client, or we can add one here.

quickwit/quickwit-ingest/src/ingest_v2/fetch.rs
```diff
@@ -204,7 +204,7 @@ impl FetchStreamTask {
                         index_uid=%self.index_uid,
                         source_id=%self.source_id,
                         shard_id=%self.shard_id,
-                        to_position_inclusive=%self.from_position_inclusive - 1,
+                        to_position_inclusive=%to_position_inclusive,
```
guilload, line 207:
```suggestion
                        %to_position_inclusive,
```

guilload (approved):

PR #6511, opened 2026-06-15
feat(search): abort split warmup when a required term is missing

PSeitz-dd:
Abort split warmup when a required term is missing to not fully download fast fields.

A leaf search warms up a split by downloading term dictionaries, posting lists, fast fields and field norms before running the query. When the query has a term that *must* match (a single-term clause reachable through `must`/`filter` only) and that term's posting list is empty in the split, the query provably matches nothing there — so the rest of the warmup is wasted work. Moreover it downloads fastfields and may put unnecessary pressure on the cache.

This extracts the set of required terms from the query AST and, during warmup, cancels the remaining downloads (fast fields, field norms, other postings) as soon as a required term is found missing, returning an empty result for the split. `warm_postings` already reports term existence, which is used here.

- quickwit-query: `required_terms` extraction + `build_tantivy_query_and_required_terms`.
- quickwit-doc-mapper: `WarmupInfo::required_terms`.
- quickwit-search: race warmup against a `CancellationToken` fired by `warm_up_terms`; abort returns a counted empty response that still reports its (aborted) download volume; new `pruned_empty_term` metric.

# Benchmarks
Comparing: main with ff cache (baseline), main, abort

`list_7d_low_hits` is a single selective term.

We can see the downloaded data is reduced, the number of get requests is unchanged (requests still run in parallel). CPU time is reduced due to reduced IO.

Screenshot 2026-06-15 at 11 46 26

Screenshot 2026-06-15 at 11 45 14

Screenshot 2026-06-15 at 11 48 12

### DIFF
--- a/quickwit/quickwit-search/src/leaf.rs
+++ b/quickwit/quickwit-search/src/leaf.rs
@@ -52,6 +52,7 @@ use tantivy::fastfield::FastFieldReaders;
 use tantivy::schema::Field;
 use tantivy::{DateTime, Index, ReloadPolicy, Searcher, TantivyError, Term};
 use tokio::task::{JoinError, JoinSet};
+use tokio_util::sync::CancellationToken;
 use tracing::*;
 
 use crate::collector::{IncrementalCollector, make_collector_for_split, make_merge_collector};
@@ -254,6 +255,33 @@ pub(crate) async fn open_index_with_caches(
     Ok((index, hot_directory))
 }
 
+/// Outcome of [`warmup`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum WarmupOutcome {
+    /// The warmup completed; the split must be searched.
+    Completed,
+    /// A required term has an empty posting list, so the query provably matches
+    /// nothing in this split. The remaining warmup downloads were cancelled.
+    ProvablyEmpty,
+}
+
+/// Runs `fut`, racing it against `cancel`. If cancellation fires first, the
+/// (possibly in-flight) future is dropped — aborting its downloads — and
+/// `Ok(())` is returned. With no token, `fut` simply runs to completion.
+async fn run_cancellable(
+    cancel: Option<&CancellationToken>,
+    fut: impl std::future::Future<Output = anyhow::Result<()>>,
+) -> anyhow::Result<()> {
+    let Some(cancel) = cancel else {
+        return fut.await;
+    };
+    tokio::select! {
+        biased;
+        _ = cancel.cancelled() => Ok(()),
+        result = fut => result,
+    }
+}
+
 /// Tantivy search does not make it possible to fetch data asynchronously during
 /// search.
 ///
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -377,20 +450,39 @@ async fn warm_up_fastfields(
...
             let inv_idx = segment_reader.inverted_index(*field)?;
             for (term, position_needed) in terms.iter() {
                 let inv_idx_clone = inv_idx.clone();
-                warm_up_futures
-                    .push(async move { inv_idx_clone.warm_postings(term, *position_needed).await });
+                // Only a required term can prove the query empty. When such a
+                // term turns out to have an empty posting list, fire the token so
+                // the rest of the warmup is cancelled.
+                let cancel_on_empty = match abort_token {
+                    Some(abort_token) if required_terms.contains(term) => Some(abort_token),
```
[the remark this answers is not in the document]
PSeitz-dd: Can be covered in a future optimization

PSeitz-dd:
@codex review

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -264,32 +292,69 @@ pub(crate) async fn open_index_with_caches(
...
+        warm_up_postings(searcher, &warmup_info.term_dict_fields),
+    )
+    .instrument(debug_span!("warm_up_postings"));
+    let warm_up_automatons_future = run_cancellable(
+        abort_token.as_ref(),
+        warm_up_automatons(searcher, &warmup_info.automatons_grouped_by_field),
+    )
+    .instrument(debug_span!("warm_up_automatons"));
 
     tokio::try_join!(
```
trinity-1686a, line 359: as a follow-up, i think it'd be interesting to record how many splits where cancelled vs not, and if that ratio goes over something small (10% maybe?), run warm_up_terms_future first, and then the rest (if warm_up_terms_future didn't cancel everything already)

trinity-1686a (approved):

trinity-1686a:
how is Object Storage Download measured? If it's self reported, i think it misrepresent the result: we record s3 download at the end of the call if it succeeded. if it died partway, i think we report 0 and not whatever was actually downloaded

PSeitz-dd:
> how is Object Storage Download measured? If it's self reported, i think it misrepresent the result: we record s3 download at the end of the call if it succeeded. if it died partway, i think we report 0 and not whatever was actually downloaded

It's still counted here: https://github.com/quickwit-oss/quickwit/pull/6511/changes#diff-6edcd5ac88fdfc3dc9ffa0758d3dc980e8aacb482ad035afa84690bab47dd703R729
You can also see it in the screenshot, we download 300MB of the 4GB fastfields before aborting.

quickwit/quickwit-search/src/leaf.rs
PSeitz-dd: Yes, I agree. I've been also exploring a bit on having a global low cardinality terms dictionary. This would allow us to estimate the number of hits relatively well and cheap.

trinity-1686a:
yes, but the thing doing the counting (a CountingStorage) doesn't report anything if the call to `get_slice()` is aborted midway (nor does it really have a way to know how much was transmitted, all it know was how much was requested and if the io finished)

PR #6512, opened 2026-06-16
feat(search): prioritize result merging over split searches in thread pool

fulmicoton-dd:
- Adds `Priority::High` / `Priority::Normal` variants to `ThreadPool` and a backing priority queue so high-priority tasks run before queued normal-priority ones
- Small tasks (e.g. small block decompression) keep running on the previous "simple" implementation.
- Marks merge finalization tasks in `leaf.rs` and `root.rs` as `Priority::High` so the final response isn't delayed by still-queued split searches
- Adds unit tests covering priority ordering, cancellation skipping, and panic recovery

### DIFF
--- a/quickwit/quickwit-search/src/fetch_docs.rs
+++ b/quickwit/quickwit-search/src/fetch_docs.rs
@@ -175,9 +175,7 @@ async fn fetch_docs_in_split(
     .context("open-index-for-split")?;
     // we add an executor here, we could add it in open_index_with_caches, though we should verify
     // the side-effect before
-    let tantivy_executor = crate::search_thread_pool()
-        .get_underlying_rayon_thread_pool()
-        .into();
+    let tantivy_executor = quickwit_common::thread_pool::get_underlying_rayon_thread_pool().into();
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-search/src/fetch_docs.rs
```diff
@@ -175,9 +175,7 @@ async fn fetch_docs_in_split(
     .context("open-index-for-split")?;
     // we add an executor here, we could add it in open_index_with_caches, though we should verify
     // the side-effect before
-    let tantivy_executor = crate::search_thread_pool()
-        .get_underlying_rayon_thread_pool()
-        .into();
+    let tantivy_executor = quickwit_common::thread_pool::get_underlying_rayon_thread_pool().into();
```
trinity-1686a, line 178: that pool is a lot smaller than the one before. Can you try if fetching a large number of docs at once is impacted?

trinity-1686a (approved):
overall looks good, i have a concern with fetch_docs going on a much smaller pool. This could degrade performance especially when fetching many documents.

PR #6514, opened 2026-06-16
feat(search): add gauge for warmup memory bytes currently in flight

fulmicoton-dd:
## Summary
- Adds `leaf_search_warmup_ongoing_num_bytes` gauge to `quickwit-search` metrics
- Tracks total bytes currently held in warmup caches across all in-flight split searches
- Uses a `WarmupBytesGuard` RAII struct to ensure the gauge is decremented on all exit paths (normal return, early return for `ProvablyEmpty`, and error paths via `?`)

## Motivation
The existing `leaf_search_single_split_warmup_num_bytes` histogram records per-split warmup sizes after the fact. This gauge provides a real-time view of memory pressure from ongoing warmups, useful for dashboards and alerting.

## Test plan
- [ ] Compile check: `cargo check -p quickwit-search`
- [ ] Clippy: `cargo clippy --workspace --all-features --tests`
- [ ] Verify metric appears in `/api/v1/metrics` after running a search


loutPhilipps (approved):
Looks good to me - with very little knowledge on the codebase

PR #6516, opened 2026-06-16
Bump tantivy to c096b2a

PSeitz-dd:
Updates tantivy dependency to the latest commit on main.


fmassot (approved):
can you do only the bump here and remove the skill changes?

PSeitz-dd:
> can you do only the bump here and remove the skill changes?

done

PR #6518, opened 2026-06-16
Update indexer pool on all node updates; Fix is_indexed for empty shards

nadav-govari:
### Description

Fixes two bugs:

1. https://github.com/quickwit-oss/quickwit/pull/6185 made it so that cluster updates only fire when ingester status changes. This made it so that the reported indexing plan coming from indexers wasn't being updated. Specifically, this appeared when a node was decommissioned, its shards got reassigned, and the new indexer the shards got reassigned finished indexing the shard, and deleted it; in these cases, the control plane running plan loop would see that it wasn't up to date and fire a bunch of reapply plan requests. 
    * Fixed by updating on all indexer changes, as before

2. `is_indexed` is broken for empty shards as of https://github.com/quickwit-oss/quickwit/pull/6504 - `self.truncation_position_inclusive.is_eof()` can't hit because it's necessarily `Position::Beginning`.
    * Fixed by checking the truncation and replication position in is_indexed instead.
    
### How was this PR tested?
Unit tested and integration tested.

I ran a test on a local cluster with all services on one node, and then one service per node with multiple indexers; the test decommissioned one node at a time the way a deployment would, and eventually closed the last one too. In these tests, the number of searchable docs was the same as the number acked on ingest. I also set min shards to be higher than the number of indexers and sent less docs than the number of shards to confirm that empty shards truncate and delete gracefully; everything worked as intended in the test.

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
@@ -3755,7 +3755,10 @@ mod tests {
                 .await;
         }
 
-        assert_eq!(*status_rx.borrow_and_update(), IngesterStatus::Decommissioning);
+        assert_eq!(
+            *status_rx.borrow_and_update(),
+            IngesterStatus::Decommissioning
+        );
 
         event_broker.publish(ShardPositionsUpdate {
             source_uid: SourceUid {
--- a/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
@@ -3666,7 +3666,13 @@ mod tests {
         let index_uid = IndexUid::for_test("test-index", 0);
         let source_id = SourceId::from("test-source");
 
-        let shard = IngesterShard::new_solo(index_uid, source_id, ShardId::from(1)).build();
+        // Give the shard un-indexed data (replication position ahead of truncation) so that, once
+        // closed, it is not immediately `is_indexed`. This keeps the ingester in `Decommissioning`
+        // long enough for the test to observe that state. An empty shard would now fast-path
+        // straight to `Decommissioned` (nothing to index).
+        let shard = IngesterShard::new_solo(index_uid, source_id, ShardId::from(1))
+            .with_replication_position_inclusive(Position::offset(12u64))
+            .build();
         let queue_id = shard.queue_id();
 
         state_guard.shards.insert(queue_id.clone(), shard);
--- a/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
@@ -1279,8 +1279,6 @@ impl EventSubscriber<ShardPositionsUpdate> for WeakIngesterState {
                     .await;
             }
         }
-        // Gossip-driven deletion is the only way an empty shard is retired, so re-check here or a
-        // decommissioning ingester whose last shard is empty would never reach `Decommissioned`.
         state_guard.check_decommissioning_status().await;
     }
 }
@@ -3320,7 +3318,6 @@ mod tests {
             )
             .await
             .unwrap();
-        // An empty shard only reaches EOF once closed, and `is_indexed` requires a closed shard.
         state_guard.shards.get_mut(&queue_id).unwrap().close();
         drop(state_guard);
 
@@ -3341,7 +3338,6 @@ mod tests {
         let state_guard = ingester.state.lock_fully("test").await.unwrap();
         let shard = state_guard.shards.get(&queue_id).unwrap();
         shard.assert_truncation_position(Position::Beginning.as_eof());
-        assert!(shard.is_indexed());
         assert!(state_guard.mrecordlog.queue_exists(&queue_id));
         state_guard.mrecordlog.assert_records_eq(&queue_id, .., &[]);
     }
@@ -3714,8 +3710,6 @@ mod tests {
         let index_uid = IndexUid::for_test("test-index", 0);
         let source_id = SourceId::from("test-source");
 
-        // Give the shard un-indexed data (replication position ahead of truncation) so that, once
-        // closed, it is not immediately `is_indexed`.
         let shard = IngesterShard::new_solo(index_uid, source_id, ShardId::from(1))
             .with_replication_position_inclusive(Position::offset(12u64))
             .build();
@@ -3781,13 +3775,16 @@ mod tests {
 
         let shard = state_guard.shards.get_mut(&queue_id).unwrap();
         shard.truncation_position_inclusive = Position::Beginning.as_eof();
+        state_guard.check_decommissioning_status().await;
+        assert_eq!(state_guard.status(), IngesterStatus::Decommissioning);
 
+        state_guard.shards.remove(&queue_id);
         state_guard.check_decommissioning_status().await;
         assert_eq!(state_guard.status(), IngesterStatus::Decommissioned);
     }
 
     #[tokio::test]
-    async fn test_decommission_completes_when_empty_shard_retired_via_gossip() {
+    async fn test_decommission_completes_when_empty_shard_deleted_via_gossip() {
         let (_ingester_ctx, ingester) = IngesterForTest::default().build().await;
 
         let event_broker = EventBroker::default();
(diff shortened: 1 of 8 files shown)


quickwit/quickwit-ingest/src/ingest_v2/models.rs
```diff
@@ -268,7 +268,11 @@ impl IngesterShard {
     }
 
     pub fn is_indexed(&self) -> bool {
```
guilload: This needs a small UT: the beginning/beginning case.

guilload (approved):

quickwit/quickwit-ingest/src/ingest_v2/models.rs
```diff
@@ -268,7 +268,14 @@ impl IngesterShard {
...
-        self.shard_state.is_closed() && self.truncation_position_inclusive.is_eof()
+        // Closed and either truncated up to EOF, or empty (no record was ever written, so the
+        // replication position is still at the beginning). The empty case matters for empty
+        // replacement shards opened during a rebalance: they never reach an EOF truncation and
+        // would otherwise block the ingester's decommission forever. A non-empty shard must still
+        // reach an EOF truncation, so a node does not quit before committing the shard's EOF.
```
guilload:
```suggestion
```

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -1291,6 +1279,9 @@ impl EventSubscriber<ShardPositionsUpdate> for WeakIngesterState {
                     .await;
             }
         }
+        // Gossip-driven deletion is the only way an empty shard is retired, so re-check here or a
```
guilload:
```suggestion
        // Gossip-driven deletion is the only way an empty shard is deleted, so re-check here or a
```

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -3298,6 +3289,63 @@ mod tests {
             .assert_records_eq(&queue_id_02, .., &[]);
     }
 
+    #[tokio::test]
+    async fn test_ingester_truncate_empty_shard_to_eof() {
```
guilload, line 3291: 👍

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -3728,20 +3776,67 @@ mod tests {
...
 
-        ingester
-            .check_decommissioning_status(&mut state_guard)
-            .await;
+        state_guard.check_decommissioning_status().await;
         assert_eq!(state_guard.status(), IngesterStatus::Decommissioned);
     }
 
+    #[tokio::test]
+    async fn test_decommission_completes_when_empty_shard_retired_via_gossip() {
```
guilload:
```suggestion
    async fn test_decommission_completes_when_empty_shard_deleted_via_gossip() {
```

quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
```diff
@@ -3728,20 +3776,67 @@ mod tests {
...
+            updated_shard_positions: vec![(shard_id, Position::Beginning.as_eof())],
+        });
+
+        tokio::time::timeout(Duration::from_secs(5), async {
+            while *status_rx.borrow_and_update() != IngesterStatus::Decommissioned {
+                status_rx.changed().await.unwrap();
+            }
+        })
+        .await
+        .expect("ingester should reach Decommissioned once its empty shard is retired via gossip");
```
guilload:
```suggestion
        .expect("ingester should reach Decommissioned once its empty shard is deleted via gossip");
```

quickwit/quickwit-ingest/src/ingest_v2/models.rs
```diff
@@ -443,4 +443,48 @@ mod tests {
...
+        let new_solo_shard = || {
+            IngesterShard::new_solo(
+                IndexUid::for_test("test-index", 0),
+                SourceId::from("test-source"),
+                ShardId::from(1),
+            )
+        };
+
+        // An empty shard (replication and truncation both at `Beginning`) has no un-indexed data,
+        // but we have to wait for EOF to be gossipped or the shard will be orphaned on decommission
```
guilload:
```suggestion
```
stfu Claude, I don't wanna hear about gossip in this file.

quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -82,6 +82,16 @@ impl InnerIngesterState {
             .await;
     }
 
+    /// Checks whether the ingester is fully decommissioned and updates its status accordingly.
+    pub async fn check_decommissioning_status(&mut self) {
+        if self.status() != IngesterStatus::Decommissioning {
+            return;
+        }
+        if self.shards.values().all(|shard| shard.is_indexed()) {
```
guilload: I wonder if we should change the condition to `self.shards.is_empty()`. If a shard is truncated at eof, `is_indexed` is true, the ingester can be considered fully decommissioned and shutdown without deleting the shard. In the meatime, the control plane will delete the shard from its model. Later the node comes back up and the empty shard is never deleted.

quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -566,32 +576,40 @@ impl FullyLockedIngesterState<'_> {
...
-            && let Some(shard) = self.inner.shards.get_mut(queue_id)
-            && shard.truncation_position_inclusive < truncate_up_to_position_inclusive
-        {
+        let Some(shard) = self.inner.shards.get_mut(queue_id) else {
+            return;
+        };
+        if shard.truncation_position_inclusive >= truncate_up_to_position_inclusive {
+            return;
+        }
+        if let Some(truncate_up_to_offset_inclusive) = truncate_up_to_position_inclusive.as_u64() {
```
guilload, line 585: `let Some() else return;` removes one level of indention and is easier to read, maybe?

quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -566,32 +576,40 @@ impl FullyLockedIngesterState<'_> {
...
         }
+        info!(
+            "truncated shard `{queue_id}` at {truncate_up_to_position_inclusive} initiated via \
+             `{initiator}`"
+        );
+        self.inner
+            .shards
+            .get_mut(queue_id)
+            .expect("shard was present above and is only removed on the missing-queue path")
+            .truncation_position_inclusive = truncate_up_to_position_inclusive;
```
guilload: meh: Can we try to merge the two branches below so the borrow checker is happy and we don't have to get the shard back from the hashmap again:
```
let Some(shard) = self.inner.shards.get_mut(queue_id) else {
            return;
        };
        if shard.truncation_position_inclusive >= truncate_up_to_position_inclusive {
            return;
        }
```

quickwit/quickwit-integration-tests/src/tests/ingest_v2_tests.rs
```diff
@@ -1143,8 +1143,8 @@ async fn test_retiring_indexer_receives_empty_plan() {
                 Err(_) => false,
             }
         },
-        Duration::from_secs(1),
```
guilload, line 1146: Those useless tweaks are going to end up in flaky tests t_t No need to time out so aggressively 1-5s is fine. No need to schedule retries so aggressively, 50-100ms is enough. Revert.

guilload (approved):
Approved but address comments.

quickwit/quickwit-ingest/src/ingest_v2/state.rs
nadav-govari: We can't unfortunately, since we need to bail on the error cases, and there's another block after this one

quickwit/quickwit-ingest/src/ingest_v2/state.rs
nadav-govari: Yep

quickwit/quickwit-ingest/src/ingest_v2/state.rs
nadav-govari: This ended up simplifying a lot of things - no more is_indexed, which allows the internal state to resolve itself much more cleanly.

PR #6521, opened 2026-06-16
Serialize mrecords into WAL using Protobuf

guilload:
### Description
Serialize mrecords into WAL using Protobuf so we have a backward and forward-compatible way to evolve the mrecords schema. This PR adds the code to write and read the new format but mrecords are still written to the WAL using the legacy format.

### How was this PR tested?
Added unit tests

### Bench
```
mrecord_append_batch/v0 time:   [1.0982 ms 1.1502 ms 1.2107 ms]
                        thrpt:  [606.54 MiB/s 638.41 MiB/s 668.69 MiB/s]

mrecord_append_batch/v1 time:   [1.1707 ms 1.3087 ms 1.4137 ms]
                        thrpt:  [519.45 MiB/s 561.09 MiB/s 627.27 MiB/s]
```

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/mrecord.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/mrecord.rs
@@ -12,19 +12,31 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use bytes::{Buf, Bytes};
-use quickwit_proto::ingest::MRecordBatch;
-use tracing::warn;
+use bytes::{Buf, BufMut, Bytes, BytesMut};
+use prost::Message as _;
+use quickwit_proto::ingest::{CommitMRecord, DocMRecord, MRecordBatch, MRecordV1, m_record_v1};
 
-/// The first byte of a [`MRecord`] is the version of the record header.
-#[derive(Debug)]
+use super::metrics::{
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-ingest/src/ingest_v2/mrecord.rs
```diff
@@ -39,38 +51,110 @@ pub enum MRecord {
...
         }
     }
 
-    pub fn decode(mut buf: impl Buf) -> Option<Self> {
-        if buf.remaining() < 2 {
-            return None;
+    /// Encodes the record using the extensible `HeaderVersion::V1` format (a protobuf
+    /// `MRecordV1` payload). Not yet used by the write path; see [`Self::encode`].
+    pub fn encode_v1(&self) -> Bytes {
+        let mrecord = match self {
```
fulmicoton, line 90: is it used anywhere? tests only?

fulmicoton: my bad, we still write using the old format

fulmicoton:
I don't see any reason why the perf should be different (it does somethign very similar) but can you run the bench with enabling proto and add the result to the PR?

fulmicoton (approved):

guilload:
So v1 allocates more cause v0 is using the `buffer.chain(buffer)` trick and proto decoding/encoding is a bit heavier. That shows in the encode/decode microbenchmark but the more realistic benchmark for `append_record_batch` is in the same ballpark.

PR #6522, opened 2026-06-17
Introduce default retry policy to otlp exporters

DoumanAsh:
### Description

Unfortunately current SDK doesn't retry automatically as it is experimental feature 
This PR just introduces feature of retry mechanism to make sure OTLP export will at least retry before giving up
I recently discovered myself that retry is not builtin with default feature set
So JFYI.

Reference:
https://docs.rs/opentelemetry-otlp/0.32.0/opentelemetry_otlp/#grpc-retry-policy

### How was this PR tested?

It hasn't been tested

### DIFF
--- a/quickwit/Cargo.lock
+++ b/quickwit/Cargo.lock
@@ -7037,6 +7037,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9966929966d17620d7c316c643ba62631826e10021409357772d5eea84f62c35"
 dependencies = [
  "http 1.4.0",
+ "httpdate",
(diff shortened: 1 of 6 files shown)


shuheiktgw (approved):
Thanks!

shuheiktgw:
@DoumanAsh Could you sign your commit so that we can merge this PR? 
https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

DoumanAsh:
@shuheiktgw No problem, force pushed commit with signature

shuheiktgw:
Thank you!

PR #6523, opened 2026-06-17
feat(storage): add named S3-compatible backends via s3 profiles

papaharry:
### Description

- Adds `storage.s3.named.` config map so a single Quickwit cluster can address multiple S3-compatible backends (separate creds, region, endpoint, checksum settings, flavor flags per entry).
- URIs of the form `s3+://bucket/path` route to the named backend; plain `s3://` continues to use the primary endpoint.
- Preserves the `s3+` qualifier through `Uri::parse_str` normalization so the storage resolver can recover the backend name on deserialization. Without this, named URIs were silently rewritten to plain `s3://` and resolved to the primary endpoint (silently wrong routing).

- `quickwit-config/src/storage_config.rs`: `NamedS3StorageConfig` struct + `storage.s3.named` BTreeMap, projection back to `S3StorageConfig` via `as_s3_config()`, redact + Debug + checksum_algorithm passthrough.
- `quickwit-storage/src/object_storage/s3_compatible_storage_resolver.rs`: `parse_named_key`, per-name cached S3Client, dispatch in `resolve()`.
- `docs/configuration/storage-config.md`: list `s3+://` in the URI protocols + Named S3 backends section with YAML example.

### How was this PR tested?

- `uri::tests::test_uri_s3_named_preserved` — parse + serde roundtrip preserves the qualifier
- `storage_config::tests::test_storage_s3_named_backends_serde` — YAML deser, `as_s3_config` projection
- `storage_config::tests::test_storage_s3_named_backends_redact`
- `s3_compatible_storage_resolver::tests::test_parse_named_key`

This fork has been running on my staging environment for a couple of weeks and everything is working as expected. The indexes are created and the source config also works as intended

image


guilload:
I'm not fond of embedding the "profile" in the storage which eventually gets persisted in the metastore. Can we consider an alternative design with per bucket profiles:
```yaml
storage:
  s3:
    region: us-east-1            # base defaults for any unlisted bucket
    profiles:
      logs-bucket-eu:            # the bucket name IS the key
        endpoint: https://s3.eu-west-3.amazonaws.com
        region: eu-west-3
        access_key_id: ${SECONDARY_S3_ACCESS_KEY_ID}
        secret_access_key: ${SECONDARY_S3_SECRET_ACCESS_KEY}
      seaweed-logs:
        endpoint: http://seaweedfs-s3:8333
        force_path_style_access: true
```

papaharry:
Thanks for that. Yeah, makes sense and will make it clearer too. I'll update it (and test on my environment) and will tag you in for another review when I think it's ready again if that's ok (will be a few days before I update it and test it)

guilload:
Sounds good. Thanks for the PR, it's a useful feature.

papaharry:
Hi @guilload, the change has been done (example)
I've been running it locally and works as expected. Is this what you were suggesting?
image

PR #6525, opened 2026-06-17
Add range check to `parse_timestamp_float`

guilload:
### Description
Add range check to `parse_timestamp_float` to reject out-of-range values instead of silently wrapping when casting.

### How was this PR tested?
Added unit tests.

### DIFF
--- a/quickwit/quickwit-datetime/src/date_time_parsing.rs
+++ b/quickwit/quickwit-datetime/src/date_time_parsing.rs
@@ -27,6 +27,14 @@ const MIN_TIMESTAMP_SECONDS: i64 = 72_057_595;
 // Maximum supported timestamp value in seconds (16 Mar 2242 12:56:31 GMT).
 const MAX_TIMESTAMP_SECONDS: i64 = 8_589_934_591;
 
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-datetime/src/date_time_parsing.rs
```diff
@@ -76,6 +76,15 @@ pub fn parse_timestamp_float(
                 .join("`, `")
         ));
     }
+    // The float is interpreted as a number of (fractional) seconds since the Unix epoch. We apply
```
guilload: @trinity-1686a I find those comments very verbose, I'm considering removing them. Thoughts?

trinity-1686a: the "The float is interpreted as a number of (fractional) seconds since the Unix epoch" can be removed.
i'd probably add `when casting the ***u128*** nanoseconds to i64 below`, because it's not obvious that we get a u128 there, and moving to unsigned for timestamps wouldn't save us from bound check

trinity-1686a (approved):

PR #6528, opened 2026-06-18
Expose health check API on dedicated HTTP server

guilload:
### Description
Expose an optional health check API on dedicated HTTP server. Useful when the REST API is behind mTLS.

### How was this PR tested?
Added unit and integration tests


docs/configuration/node-config.md
```diff
@@ -89,10 +89,37 @@ We advise changing the default value of 20 MiB only if you encounter the followi
 `Error, message length too large: found 24732228 bytes, the limit is: 20971520 bytes.` In that case, increase `max_message_size` by increments of 10 MiB until the issue disappears. This is a temporary fix: the next version of Quickwit will rely exclusively on gRPC streaming endpoints and handle messages of any length.
 :::
 
+## Health check configuration
+
+This section configures an optional, **plaintext (no TLS)** HTTP server that exposes only the health endpoints `/health/livez` (liveness) and `/health/readyz` (readiness). Its purpose is to let liveness/readiness probes (for example from Kubernetes or a load balancer) reach the node even when the main REST API is put behind [TLS or mTLS](#tls-configuration), which a simple HTTP probe cannot negotiate.
+
+The health server is **disabled by default**. It starts only when `listen_port` is set (or the `QW_HEALTH_LISTEN_PORT` environment variable is provided). The same `/health/*` endpoints always remain available on the main REST API as well.
```
trinity-1686a, line 96: we should probably document the differences between `/health/*` and the health endpoint, namely that the later starts responding way sooner, while the later starts responding only when the node is ready (including /livez)

trinity-1686a (approved):

PR #6529, opened 2026-06-19
Add split_key_prefix_len to index config to shard S3 object keys

fulmicoton-dd:
## Summary

- Adds `split_key_prefix_len: u8` to `IndexConfig` (and `IndexTemplate`) to configure S3 key sharding per index
- Adds `compute_split_key_prefix(split_id, prefix_len)` to `quickwit-common` — extracts N characters from the ULID random portion (positions 10–25) as a prefix; logs a rate-limited warning and falls back to the flat scheme if the split ID is too short
- Adds `split_storage_path(split_id, prefix)` to `quickwit-common` — builds the storage path from a precomputed prefix string; empty prefix = legacy flat scheme
- Validates that `split_key_prefix_len <= 16` (ULID random portion length) at config load time
- `SplitMetadata` will gain a `prefix: String` field in a follow-up PR to wire the full pipeline (uploader, leaf search, merge, GC)

**Backward compatibility**: the field defaults to `0` (serde default), so all existing indexes and splits are unaffected. New splits on indexes with `split_key_prefix_len: 2` will land at `ND/01ARZ3.../01ARZ3....split` paths, distributing across 1024 S3 partitions instead of one.

## Test plan

- [x] `cargo nextest run -p quickwit-common -p quickwit-config --all-features` — 221 tests pass
- [x] `cargo clippy --workspace --all-features --tests` — no warnings
- [x] `cargo +nightly fmt --all -- --check` — no issues
- [x] Set `split_key_prefix_len: 2` in an index config, verify it round-trips through serde correctly
- [x] Set `split_key_prefix_len: 17`, verify it is rejected with a clear error message


quickwit/quickwit-indexing/src/models/split_prefix.rs
```diff
@@ -0,0 +1,115 @@
+// Copyright 2021-Present Datadog, Inc.
```
guilload, line 1: Having this file located under `actors` is surprising because this is not an actor.

quickwit/quickwit-indexing/src/models/split_prefix.rs
```diff
@@ -0,0 +1,115 @@
...
+/// directory.
+///
+/// Controlled by the `QW_SPLIT_KEY_PREFIX_LEN` environment variable (default: 0, i.e. the legacy
+/// flat scheme). Setting this to `2` creates 32^2 = 1024 buckets, which is sufficient for most
+/// workloads. Values above [`MAX_SPLIT_KEY_PREFIX_LEN`] are clamped.
+///
+/// The value is read from the environment once and cached for the lifetime of the process.
+fn split_key_prefix_len() -> u8 {
+    static SPLIT_KEY_PREFIX_LEN: LazyLock<u8> = LazyLock::new(|| {
+        let prefix_len: u8 = quickwit_common::get_from_env("QW_SPLIT_KEY_PREFIX_LEN", 0u8, false);
```
guilload, line 36: I wonder if we should harcode the len to a reasonable default and forget about it reducing the amount of boilerplate in the config. We used to used 4 chars at Airbnb: we would hash the original path and use the first four char of the hash.

quickwit/quickwit-common/src/lib.rs
```diff
@@ -112,6 +112,43 @@ pub fn split_file(split_id: impl Display) -> String {
...
+/// - Empty `prefix`: legacy flat scheme `{split_id}.split`
+/// - Non-empty `prefix`: `{prefix}/{split_id}.split` — the `/` separator is inserted here, so the
+///   `prefix` itself must NOT contain a trailing `/`
+///
+/// The prefix is computed once at split creation time and stored in `SplitMetadata`.
+pub fn split_storage_path(split_id: &str, prefix: &str) -> String {
+    if prefix.is_empty() {
+        return format!("{split_id}.split");
+    }
+    format!("{prefix}/{split_id}.split")
```
guilload, line 126: I think the `/` as a separator is not great because the AWS UI and S3 CLI is going to show them as a directories.

fulmicoton: ok changing that for - then?

fulmicoton: actually I cannot change this without breaking yahoo.

fmassot: we can delete the splits if needed

quickwit/quickwit-indexing/src/models/split_prefix.rs
fulmicoton: ok moved it to models.

quickwit/quickwit-indexing/src/models/split_prefix.rs
fulmicoton: split ulid are base32. 5 bits per chars. a prefix of 2 already gives us 1024 buckets too (and then the regular ulid).

the reason why I didn't do more was, on the off change that we need to find all splits within a range it can be done in 1024 requests. If we don't care then ok for 4. Let me know what you think.

guilload: I was more trying to make a point about hardcoding, not about the length specifically. Any length that works for you works for me.

PR #6537, opened 2026-06-24
Remove native-tls from transitive dependencies

fulmicoton-dd:
## Summary

Switch all crates to rustls-only TLS backends, eliminating `native-tls`, `hyper-tls`, `tokio-native-tls`, and related packages from the dependency tree.

Three packages were pulling in `native-tls` via their default features:
- `azure_identity`: default enables `enable_reqwest` → `reqwest/default-tls`; switched to `default-features = false, features = ["enable_reqwest_rustls"]`
- `google-cloud-auth`: default enables `default-tls` → `reqwest/default-tls`; switched to `default-features = false, features = ["rustls-tls"]`
- `pulsar`: `tokio-runtime` feature hardcodes `native-tls`; switched to `tokio-rustls-runtime`

`google-cloud-pubsub` is also tightened to `default-features = false, features = ["auth", "rustls-tls"]` for consistency.

Net result: 75 fewer entries in `Cargo.lock`, no native-tls/OpenSSL dependency at all when building with `--all-features`.

Also fixes an incomplete comment in the generated `quickwit.search.rs` proto code.

## Test plan

- [x] `cargo check --all-features` passes
- [x] `cargo tree --all-features -i native-tls` returns no matches
- [ ] `cargo nextest run --all-features` (CI)


guilload:
Nice!

PR #6542, opened 2026-06-24
Ensure all indexers broadcast standalone compactors before starting m…

nadav-govari:
### Description

When initially turning on `enableStandaloneCompactors`, the janitor running the merge planner and the compactors come online basically immediately and start performing merges. Indexers take a long time (minutes) to decommission, and deploy one by one. That means that we can end up with a split brain situation where there's merge pipelines running on the old indexers, and compactors from the new deploy are up and running. They then end up with a bunch of publish failures.

This addresses that by adding a standalone compactors value to ClusterMember, so that the merge planner can await for all indexers to show that they have standalone compactors enabled. This necessarily means that no merge pipelines are running on those indexers, so that starting standalone compaction is safe.

### How was this PR tested?

Unit tests, local cluster.

### DIFF
--- a/quickwit/quickwit-compaction/src/planner/compaction_planner.rs
+++ b/quickwit/quickwit-compaction/src/planner/compaction_planner.rs
@@ -19,6 +19,8 @@ use anyhow::Result;
 use async_trait::async_trait;
 use itertools::Itertools;
 use quickwit_actors::{Actor, ActorContext, ActorExitStatus, Handler};
+use quickwit_cluster::Cluster;
+use quickwit_common::rate_limited_tracing::rate_limited_info;
 use quickwit_metastore::{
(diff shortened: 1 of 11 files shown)


quickwit/quickwit-compaction/src/planner/compaction_planner.rs
```diff
@@ -119,14 +152,24 @@ impl Handler<ReportStatusRequest> for CompactionPlanner {
...
             state: CompactionState::new(),
             index_config_metastore: IndexConfigMetastore::new(metastore.clone()),
             metastore,
+            cluster,
         }
     }
 
+    async fn all_indexers_migrated(&self) -> bool {
+        self.cluster
+            .live_nodes()
```
nadav-govari, line 166: explain why live nodes but not ready nodes

guilload (approved):

PR #6548, opened 2026-06-25
Add a metastore read replica role for read-only routing

shuheiktgw:
### Description

This PR adds support for deploying metastore read replicas as a dedicated `metastore_read_replica` service and lets searchers opt into using them through `searcher.use_metastore_read_replica`.

### How it works

1. Users deploy standalone `metastore_read_replica` nodes with `metastore_read_replica_uri` pointing to a PostgreSQL read replica. If `searcher.use_metastore_read_replica` is true, searchers use `metastore_read_replica` instead of the primary metastore.
2. Searchers use a dedicated read-only metastore client, `MetastoreReadServiceClient`, which exposes only the read-only subset of metastore RPCs to prevent write methods from being used.

### Limitations / Future Improvements
- `metastore_read_replica` needs to run as a standalone service to keep the implementation simple. It shares the same gRPC service as `metastore`, so it cannot run alongside it, and running it separately also simplifies the metadata service connection handling.
- Currently, searchers configured to use metastore read replicas still need access to the primary metastore. Quickwit treats every node as an API server, and some API requests need to write to the metastore. However, the readiness check for the read-only metastore only verifies whether the read replicas are accessible. This means that if the primary database is down, the read-only metastore may be unable to access it and could drop write requests. We decided that this is the right trade-off: it separates the search path from the write path and keeps searchers alive during primary DB degradation.
- There are many more places where we could use the read replica metastore, but the primary focus of this PR is the searcher. We can expand support further later.


config/quickwit.yaml
```diff
@@ -168,6 +175,11 @@ indexer:
 # https://quickwit.io/docs/configuration/node-config#searcher-configuration
 #
 # searcher:
+#   # If true, routes read-only metastore requests from searchers, including
+#   # DataFusion when enabled, to nodes running the `metastore_read_replica`
+#   # service. Searchers require at least one `metastore_read_replica` node at
+#   # startup and do not fall back to the primary metastore.
+#   use_metastore_read_replica: false
```
shuheiktgw, line 194: The reason we need this flag in addition to `metastore_read_replica_uri` is that users can set `metastore_read_replica_uri` via the `QW_METASTORE_READ_REPLICA_URI` environment variable. Without this flag, searchers would need access to that same environment variable even though they don’t need the secret itself.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -498,52 +538,93 @@ pub async fn serve_quickwit(
...
-                    !connections.is_empty()
-                })
+    // Instantiate a read-only metastore replica server if the `metastore_read_replica` role is
+    // enabled on the node.
+    let read_only_metastore_server_opt: Option<MetastoreServiceClient> =
+        if node_config.is_service_enabled(QuickwitService::MetastoreReadReplica) {
+            let read_replica_uri = node_config.metastore_read_replica_uri.as_ref().expect(
+                "`metastore_read_replica_uri` should be set when the `metastore_read_replica` \
+                 role is enabled (validated at config load)",
+            );
```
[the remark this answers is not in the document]
shuheiktgw: This is going to be addressed in a different PR!
https://github.com/shuheiktgw/quickwit/pull/2

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -778,8 +865,11 @@ pub async fn serve_quickwit(
     let quickwit_services: Arc<QuickwitServices> = Arc::new(QuickwitServices {
         node_config: Arc::new(node_config),
         cluster: cluster.clone(),
-        metastore_server_opt,
-        metastore_client: metastore_through_control_plane.clone(),
+        metastore_server_opt: primary_metastore_server_opt
+            .as_ref()
+            .or(read_only_metastore_server_opt.as_ref())
+            .cloned(),
+        metastore_client: primary_metastore_through_control_plane.clone(),
```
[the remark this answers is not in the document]
shuheiktgw: Can be future improvement I suppose

quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -896,6 +907,26 @@ mod tests {
     use super::*;
     use crate::IndexerConfig;
 
+    #[test]
+    fn test_node_config_redact_metastore_uris() {
```
guilload, line 1013: 👍

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -357,10 +367,38 @@ fn validate(node_config: &NodeConfig) -> anyhow::Result<()> {
...
+    }
+    match &node_config.metastore_read_replica_uri {
+        Some(_) => Ok(()),
+        None => {
+            bail!(
+                "the `metastore_read_replica` service requires `metastore_read_replica_uri` to be \
+                 set"
+            )
+        }
+    }
```
guilload:
```suggestion
 if node_config.metastore_read_replica_uri.is_none() {
        bail!(
            "the `metastore_read_replica` service requires `metastore_read_replica_uri` to be set"
        );
    }
    Ok(())
}
```

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -1010,6 +1053,107 @@ mod tests {
             .unwrap();
     }
 
+    #[tokio::test]
+    async fn test_metastore_read_replica_role_accepts_postgres_uri() {
```
guilload: I think this test is redundant.

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -1010,6 +1053,107 @@ mod tests {
...
+        .await
+        .unwrap();
+        assert_eq!(
+            config.metastore_read_replica_uri.unwrap(),
+            "postgresql://user:pass@host:5432/db"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_metastore_read_replica_uri_can_be_set_from_env() {
```
guilload: I think this one is overkill. This is basically `ConfigValue` which is already tested.

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -1010,6 +1053,107 @@ mod tests {
...
+        assert!(
+            error
+                .to_string()
+                .contains("requires `metastore_read_replica_uri`")
+        );
+    }
+
+    #[tokio::test]
+    async fn test_metastore_read_replica_role_must_run_standalone() {
+        for service in [
```
guilload: I think the for-loop is overkill, especially since the branch that tests this is `!= 1`

quickwit/quickwit-config/src/service.rs
```diff
@@ -46,12 +50,25 @@ impl QuickwitService {
             QuickwitService::Searcher => "searcher",
             QuickwitService::Janitor => "janitor",
             QuickwitService::Metastore => "metastore",
+            QuickwitService::MetastoreReadReplica => "metastore_read_replica",
```
guilload, line 54:
```suggestion
            QuickwitService::MetastoreReadReplica => "metastore-read-replica",
```
`control_plane` was a mistake :) I'd rather follow the k8s convention. I agree this is super nit and arbitrary/debatable. I'm basically abusing my co-BDFL title.

quickwit/quickwit-config/src/service.rs
```diff
@@ -46,12 +50,25 @@ impl QuickwitService {
...
     pub fn supported_services() -> HashSet<QuickwitService> {
         all::<QuickwitService>().collect()
     }
+
+    /// Returns the services enabled on a node when none are explicitly configured.
+    ///
+    /// This is every supported service except [`QuickwitService::MetastoreReadReplica`], which is
+    /// opt-in: it requires a dedicated read replica URI and is meant to be deployed as its own set
+    /// of nodes, so it must never be enabled implicitly by an all-in-one node.
+    pub fn default_services() -> HashSet<QuickwitService> {
```
guilload, line 71: @nadav-govari this is going to impact your branch.

quickwit/quickwit-config/src/service.rs
```diff
@@ -70,6 +87,9 @@ impl FromStr for QuickwitService {
             "searcher" => Ok(QuickwitService::Searcher),
             "janitor" => Ok(QuickwitService::Janitor),
             "metastore" => Ok(QuickwitService::Metastore),
+            "metastore-read-replica" | "metastore_read_replica" => {
```
guilload, line 99: Sweet

quickwit/quickwit-metastore/src/metastore/read_service.rs
```diff
@@ -0,0 +1,107 @@
...
+    MetastoreResult, MetastoreService, MetastoreServiceClient, MetastoreServiceStream,
+};
+
+/// Read-only subset of the metastore RPC surface that is safe to route to a read replica.
+///
+/// Callers that only need stale-tolerant reads (e.g. searchers, the DataFusion analytics path)
+/// should depend on this trait rather than on the full [`MetastoreService`], so that writes are
+/// excluded at the type level. Add a method here only for RPCs that tolerate replication lag.
+#[async_trait]
+pub trait MetastoreReadService: fmt::Debug + Send + Sync + 'static {
```
guilload: Even if we have the `MetastoreReadService` trait, I'd like us to properly catch the Posgres error "read only" (code 25006) and return it properly instead of a generic ` MetastoreError::Db` error.

quickwit/quickwit-config/src/service.rs
fulmicoton-dd: I'd stick to _. This is not worth the effort and debatable.

One interesting argument pro "_" : - is often considered something you tokenize on, while _ does not.
So for instance, if you double click on a word to select it, or do maj arrow thingy to select stuff, or do b in vim, it will stop at the - but not at the _. 

snake case vs kebab case

quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -343,6 +343,9 @@ pub struct SearcherConfig {
     #[serde(default)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub storage_timeout_policy: Option<StorageTimeoutPolicy>,
+    /// Routes read-only metastore requests from searchers, including DataFusion when enabled, to
+    /// nodes running the `metastore_read_replica` service.
+    pub use_metastore_read_replica: bool,
```
guilload, line 425:
```suggestion
    /// nodes running the `metastore_read_replica` service.
    #[serde(default)]
    pub use_metastore_read_replica: bool,
```
Should default to `false` for backward compatibility and should be test. There is a test that tests default values somewhere.

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -357,10 +367,35 @@ fn validate(node_config: &NodeConfig) -> anyhow::Result<()> {
...
+///
+/// The [`QuickwitService::MetastoreReadReplica`] role serves the same gRPC service as a read-only
+/// metastore, so it must run standalone and requires `metastore_read_replica_uri` to connect to.
+fn validate_metastore_read_replica(node_config: &NodeConfig) -> anyhow::Result<()> {
+    let read_replica_enabled =
+        node_config.is_service_enabled(QuickwitService::MetastoreReadReplica);
+    if !read_replica_enabled {
+        return Ok(());
+    }
+    if node_config.enabled_services.len() != 1 {
```
guilload: Use `anyhow::ensure!`

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -357,10 +367,35 @@ fn validate(node_config: &NodeConfig) -> anyhow::Result<()> {
...
+        bail!(
+            "the `metastore_read_replica` service must run standalone and cannot be combined with \
+             any other service"
+        );
+    }
+    if node_config.metastore_read_replica_uri.is_none() {
```
guilload: Same here.

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -663,6 +700,10 @@ mod tests {
             config.metastore_uri,
             "postgresql://username:password@host:port/db"
         );
+        assert_eq!(
+            config.metastore_read_replica_uri.as_ref().unwrap().as_str(),
```
guilload: Is the `as_str()` needed?

quickwit/quickwit-metastore/src/metastore/postgres/error.rs
```diff
@@ -63,3 +67,92 @@ pub(super) fn convert_sqlx_err(index_id: &str, sqlx_error: sqlx::Error) -> Metas
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
guilload: I don't think this test is very useful. We should test the code against a real posgres, which should be doable because we already have the infrastructure in place.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -230,6 +231,48 @@ impl QuickwitServices {
...
+        100
+    }
+}
+
+async fn build_metastore_client(
+    cluster: &Cluster,
+    service: QuickwitService,
+    max_message_size: ByteSize,
+) -> anyhow::Result<MetastoreServiceClient> {
+    info!(%service, "connecting to metastore service");
```
guilload:
```suggestion
    info!(%service, "connecting to {service} service");
```

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -499,52 +538,93 @@ pub async fn serve_quickwit(
...
-        control_plane_client.clone(),
-        metastore_client,
-    ));
+    // Set up the "control plane proxy" for the primary metastore.
+    let primary_metastore_through_control_plane =
+        MetastoreServiceClient::new(ControlPlaneMetastore::new(
+            control_plane_client.clone(),
+            primary_metastore_client.clone(),
+        ));
+    let read_only_metastore_through_control_plane_opt = read_only_metastore_client_opt
```
guilload: We don't need a read-only metastore through control plane.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -1438,10 +1534,32 @@ fn with_arg<T: Clone + Send>(arg: T) -> impl Filter<Extract = (T,), Error = Infa
     warp::any().map(move || arg.clone())
 }
 
+async fn metastores_are_available(metastores: &[MetastoreServiceClient]) -> bool {
```
guilload: The primary goal of this work is isolation: searchers should read from the replica so they're not impacted by issues on the primary (and, conversely, so read activity doesn't load the write path.) That separation is the whole point of introducing the replica.

Being able to serve the CRUD / index-management API on a searcher node is a nice-to-have, but it's strictly secondary to searchers being able to search. So if the primary is down, a searcher whose reads are served by a healthy replica should stay ready for search traffic and degrading the CRUD API is an acceptable trade, taking the searcher out of rotation is not.

The tension here is that node readiness is a single boolean and can't express "ready for search, not ready for writes." The fix isn't to keep requiring the primary via .all(); it's to decouple the two: search-readiness follows the replica, write/management-readiness follows the primary. Gating searcher availability on the primary reintroduces exactly the coupling we set out to remove.

quickwit/quickwit-config/src/node_config/mod.rs
shuheiktgw: I've added '#[serde(default)]' and confirmed the test is defined [here](https://github.com/quickwit-oss/quickwit/blob/d1b69ddaeebb27b8097bf73e515ffc7aefa580bc/quickwit/quickwit-config/src/node_config/serialize.rs#L1109).

quickwit/quickwit-config/src/node_config/serialize.rs
shuheiktgw: Simplified it with `config.metastore_read_replica_uri.unwrap()`

quickwit/quickwit-metastore/src/metastore/postgres/error.rs
shuheiktgw: I've extended test_metastore_resolver_resolve_read_only_postgres and test the actual error.

- https://github.com/quickwit-oss/quickwit/pull/6548/changes/305759acbf0b9fc1a24c6743f32d576152f9a768

```
QW_TEST_DATABASE_URL=postgres://quickwit-dev:quickwit-dev@localhost:5432/quickwit-metastore-dev \
cargo test -p quickwit-metastore --features postgres test_metastore_resolver_resolve_read_only_postgres -- --nocapture

test result: ok. 1 passed; 0 failed
```

quickwit/quickwit-serve/src/lib.rs
shuheiktgw: Remove the control plane dependency 🙏 
- https://github.com/quickwit-oss/quickwit/pull/6548/changes/ab0720f9e709ac76d952f6f1f3855797081aab0a

quickwit/quickwit-serve/src/lib.rs
shuheiktgw: Updated it to use only the read replica from the read-only metastore for the readiness check.
- https://github.com/quickwit-oss/quickwit/pull/6548/changes/74c9efeeafac45153a4f12663070d410776defa4

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -614,52 +653,85 @@ pub async fn serve_quickwit(
...
+
+    let primary_metastore_client: MetastoreServiceClient =
+        if let Some(metastore_server) = &primary_metastore_server_opt {
+            metastore_server.clone()
+        } else {
+            build_metastore_client(
+                &cluster,
+                QuickwitService::Metastore,
+                grpc_config.max_message_size,
+            )
```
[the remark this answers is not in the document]
shuheiktgw: That's true, but a metastore read-replica node still exposes the API server, and some API endpoints need access to the primary metastore for writes. So we currently build the primary metastore client eagerly at startup and wait until at least one primary metastore node is discoverable. Readiness itself is still based on the read-replica metastore once the node has started.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -1048,12 +1127,23 @@ pub async fn serve_quickwit(
             )
         };
 
+    let metastore_readiness_client =
+        if let Some(read_only_metastore_client) = &read_only_metastore_client_opt {
+            // When a read replica metastore is configured, node readiness follows the replica only.
+            // This keeps search/read traffic available if the primary metastore is down.
+            // Write-capable index-management APIs are still routed to the primary metastore and
+            // may fail while this node remains ready.
+            read_only_metastore_client.clone()
```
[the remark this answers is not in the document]
shuheiktgw: This is intentional.

shuheiktgw:
I refactored the metastore initialization logic by introducing a `LocalMetastoreServer` enum and `start_metastore_service_if_needed`, with the goal of keeping all the related logic in one place.
- https://github.com/quickwit-oss/quickwit/pull/6548/commits/1ebe477999c7bf790cbfff24f4c5c6ccb1b5493d

shuheiktgw:
We discussed yesterday that the startup of the read-only metastore is unnecessarily blocked by the primary metastore. And yes, the read-only metastore does wait up to 300s for a metastore node before mounting the read-only metastore service.

But I took a closer look and realized this is actually fine, because a metastore read-replica node still exposes the API server, and some API endpoints need access to the primary metastore for writes. Readiness itself is still based on the read-replica metastore once the node has started.

WDYT?

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -244,6 +245,189 @@ impl QuickwitServices {
...
+                Ok(Some(metastore_server.clone()))
+            }
+            LocalMetastoreServer::Primary(_) | LocalMetastoreServer::NotServed
+                if should_use_remote_read_replica =>
+            {
+                let read_replica_client = Self::build_metastore_client(
+                    cluster,
+                    QuickwitService::MetastoreReadReplica,
+                    node_config.grpc_config.max_message_size,
+                )
```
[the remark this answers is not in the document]
shuheiktgw: Fixed it
- https://github.com/quickwit-oss/quickwit/pull/6548/changes/e8d51a982a3db6865424544bc532a7b73d2b874a

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -383,6 +392,37 @@ fn validate(node_config: &NodeConfig) -> anyhow::Result<()> {
...
+            node_config.metastore_read_replica_uri.is_some(),
+            "the `metastore_read_replica` service requires `metastore_read_replica_uri` to be set"
+        );
+    }
+    let searcher_uses_read_replica = node_config.is_service_enabled(QuickwitService::Searcher)
+        && node_config.searcher_config.use_metastore_read_replica;
+    // Avoid a deadlock where the searcher waits for a READY read replica, while the read
+    // replica waits for this primary metastore to become READY.
+    ensure!(
+        !(searcher_uses_read_replica && node_config.is_service_enabled(QuickwitService::Metastore)),
```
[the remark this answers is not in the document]
shuheiktgw: This is going to be addressed in a different PR!
https://github.com/shuheiktgw/quickwit/pull/2

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
```diff
@@ -3494,9 +3560,14 @@ mod tests {
         let query = ListSplitsQuery::for_all_indexes()
             .with_excluded_split_ids(HashSet::from([SplitId::from("s1"), SplitId::from("s2")]));
         append_query_filters_and_order_by(sql, query);
-        assert_eq!(
-            sql.to_string(PostgresQueryBuilder),
-            r#"SELECT * FROM "splits" WHERE split_id <> ALL(ARRAY ['s1','s2']::text[])"#
+        let sql = sql.to_string(PostgresQueryBuilder);
+        assert!(
```
guilload: The "rendering" of the query should be made deterministic.

quickwit/quickwit-metastore/src/metastore_resolver.rs
```diff
@@ -67,13 +85,20 @@ impl MetastoreResolver {
                 ));
             }
         };
+        if options.read_only && backend != MetastoreBackend::PostgreSQL {
+            return Err(MetastoreResolverError::UnsupportedBackend(
+                "read-only metastore connections are only supported for PostgreSQL".to_string(),
```
guilload:
```suggestion
                "read-only option is only supported for the PostgreSQL backend".to_string(),
```

quickwit/quickwit-metastore/src/metastore/postgres/error.rs
```diff
@@ -13,40 +13,44 @@
...
-                (pg_error_codes::UNIQUE_VIOLATION, _) => {
+                (Some(pg_error_codes::UNIQUE_VIOLATION), _) => {
                     error!(error=?boxed_db_error, "postgresql-error");
                     MetastoreError::Internal {
                         message: "unique key violation".to_string(),
                         cause: format!("DB error {boxed_db_error:?}"),
                     }
                 }
+                (Some(pg_error_codes::READ_ONLY_SQL_TRANSACTION), _) => MetastoreError::Forbidden {
+                    message: format!("postgresql metastore is read-only: {boxed_db_error}"),
```
guilload:
```suggestion
                    message: format!(""cannot issue write request on a read-only replica": {boxed_db_error}"),
```

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -244,6 +245,189 @@ impl QuickwitServices {
     }
 }
 
+/// The metastore gRPC server this node serves locally — or `NotServed` if it serves none.
```
guilload: We need to stop the unbounded growth of this file. Move the new metastore-related logic to a new file.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -1894,6 +2062,40 @@ mod tests {
         assert_eq!(response.status(), ServingStatus::NotServing.into());
     }
 
+    #[tokio::test]
+    async fn test_readiness_uses_read_replica_without_requiring_primary() {
```
guilload, line 1902: This doesn't test anything new, is it?

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -815,12 +948,20 @@ pub async fn serve_quickwit(
         ))
     };
 
+    let read_only_metastore_client_opt = local_metastore_server
+        .resolve_read_only_client(&cluster, &node_config)
+        .await?;
+
+    let metastore_read_client = read_only_metastore_client_opt
```
guilload: The naming is bad. This is not always a read-only client and as result later when we pass it to the readiness tasks, it's hard to know against which metastore the readiness task is running.

guilload (commented):
I think this works but unfortunately, the code was not written in a way that optimizes for "proof readability". The logic in `quickwit/quickwit-serve/src/lib.rs` needs more logging.

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
```diff
@@ -101,11 +104,69 @@ impl fmt::Debug for PostgresqlMetastore {
...
+    skip_migrations: bool,
+    skip_locking: bool,
+    skip_deferred: bool,
+}
+
+impl PostgresqlMetastoreOptions {
+    /// Default options, with the read-only and migration flags read from the environment.
+    fn from_env() -> Self {
+        Self {
+            read_only: get_bool_from_env(QW_POSTGRES_READ_ONLY_ENV_KEY, false),
```
PSeitz-dd: These should be cached

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
PSeitz-dd: done

quickwit/quickwit-serve/src/lib.rs
PSeitz-dd: moved to [metastore.rs](https://github.com/quickwit-oss/quickwit/pull/6548/changes#diff-a2d363e3264d82c354bad5e7b9452916011f49042bf405d9732cd397b9c8e3f6)

quickwit/quickwit-serve/src/lib.rs
PSeitz-dd: changed it so it tests the node is ready with only a replica

quickwit/quickwit-serve/src/lib.rs
PSeitz-dd: fixed the naming

PSeitz-dd:
> I think this works but unfortunately, the code was not written in a way that optimizes for "proof readability". The logic in `quickwit/quickwit-serve/src/lib.rs` needs more logging.

I added some more metastore specific logging

quickwit/quickwit-metastore/src/metastore_resolver.rs
```diff
@@ -200,4 +240,48 @@ mod tests {
             metastore_resolver.resolve(&postgres_uri).await.unwrap();
         }
     }
+
+    #[cfg(feature = "postgres")]
+    #[tokio::test]
+    async fn test_metastore_resolver_resolve_read_only_postgres() {
```
guilload, line 246: 👍

guilload (approved):
Approved but also need a label for metastore metrics to distinguish primary from replica.

shuheiktgw:
I’ve added the `metastore_kind` label to metrics in the metastore subsystem, consistent with the logs. Note that I didn’t add the label to the `deferred_migrations_apply_total` metric because migrations are never run against read replicas, so the label isn’t relevant there.

https://github.com/quickwit-oss/quickwit/pull/6548/commits/2aaaca2508a0c39c7dfbfcbdb7e767e919032a8f

Issue #6524, opened 2026-06-17
Unit test failure


fulmicoton:
Our scheduler proptest failed.

```
    thread 'indexing_scheduler::scheduling::scheduling_logic::tests::test_proptest_post_conditions' (8548) panicked at quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs:792:13:
    solution unstable!!!
    Solution 1: SchedulingSolution { indexer_assignments: [IndexerAssignment { indexer_ord: 0, num_shards_per_source: {2: 1, 6: 1} }, IndexerAssignment { indexer_ord: 1, num_shards_per_source: {2: 1, 3: 1} }], capacity_scaling_iterations: 1 }
    Solution 2: SchedulingSolution { indexer_assignments: [IndexerAssignment { indexer_ord: 0, num_shards_per_source: {2: 2, 6: 1} }, IndexerAssignment { indexer_ord: 1, num_shards_per_source: {3: 1} }], capacity_scaling_iterations: 1 }
    Solution 3: SchedulingSolution { indexer_assignments: [IndexerAssignment { indexer_ord: 0, num_shards_per_source: {2: 2} }, IndexerAssignment { indexer_ord: 1, num_shards_per_source: {3: 1, 6: 1} }], capacity_scaling_iterations: 0 }
    stack backtrace:
       0: __rustc::rust_begin_unwind
       1: core::panicking::panic_fmt
       2: quickwit_control_plane::indexing_scheduler::scheduling::scheduling_logic::tests::test_proptest_post_conditions::{{closure}}
       3: core::ops::function::impls::<impl core::ops::function::Fn<A> for &F>::call
       4: proptest::test_runner::runner::call_test::{{closure}}::{{closure}}
       5: <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
       6: std::panicking::catch_unwind::do_call
       7: __rust_try
       8: std::panic::catch_unwind
       9: proptest::test_runner::runner::call_test::{{closure}}
      10: proptest::test_runner::scoped_panic_hook::internal::with_hook
      11: proptest::test_runner::runner::call_test
      12: proptest::test_runner::runner::TestRunner::shrink
      13: proptest::test_runner::runner::TestRunner::run_one_with_replay
      14: proptest::test_runner::runner::TestRunner::gen_and_run_case
      15: proptest::test_runner::runner::TestRunner::run_in_process_with_replay
[...]

PR #6551, opened 2026-06-25
fix(control-plane): make scheduling solve() idempotent (#6524)

fulmicoton-dd:
## Problem

`solve()` in the control-plane scheduler was not idempotent (#6524). Capacity
inflation lived inside `place_unassigned_shards_ignoring_affinity`, which scaled
node capacities by ×1.2 in a local loop until every shard fit. The next run started has a step that removes
non fitting shard from the solution. 

If the shard removed were not added in `place_unassigned_shards_ignoring_affinity` there was no guarantee
that the resulting solution would be the same.

The problem was spotted by our proptest (yeaaaaah).
The regression test `test_reproduce_6524` captures the minimal reproducer: source 2
(250 mcpu, 1 shard) bounces between indexer 0 and indexer 1 on each call, so
`solution_2 != solution_3`.

## Fix

We move the capacity inflation logic so that it wraps the entire pipeline:

- `place_unassigned_shards_ignoring_affinity` and the extracted `attempt_solve` are
  now fallible (`Result<_, NotEnoughCapacity>`) and contain no inflation loop.
- `solve()` **ascends** to find the minimal inflation level at which the full
  pipeline succeeds (each attempt from the original `previous_solution`), then
  **descends** — re-feeding the candidate solution at lower levels — to find the
  true minimal feasible level for *that* solution.
- A single `problem_at_inflation_level` helper computes `base × 1.2^n` with one
  ceiling, guaranteeing identical capacities for a given level regardless of path.

The returned solution is feasible at `best_attempt` but not at `best_attempt - 1`,
so a subsequent `solve()` re-climbs to the same level where the pipeline is a no-op.
Hence `solve()` is idempotent.

## Disclaimer

Building the plan will take more CPU than before.

## Testing

- `test_reproduce_6524` now passes.
- `test_proptest_post_conditions` (idempotency proptest, `solution_2 == solution_3`)
  passes, including at 4096 cases.
- Full `quickwit-control-plane` suite green; clippy clean; formatted.

### DIFF
--- a/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
+++ b/quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
@@ -304,7 +304,7 @@ fn place_unassigned_shards_ignoring_affinity(
     // logical bug.
     for attempt_number in 0..30 {
         match attempt_place_unassigned_shards(&unassigned_shards[..], &problem, partial_solution) {
-            Ok(mut solution) => {
+            Ok(solution) => {
                 // the higher the attempt number, the more unbalanced the solution
                 if attempt_number > 0 {
                     tracing::warn!(
@@ -312,7 +312,6 @@ fn place_unassigned_shards_ignoring_affinity(
                         "capacity re-scaled, scheduling solution likely unbalanced"
                     );
                 }
-                solution.capacity_scaling_iterations = attempt_number;
                 return solution;
             }
             Err(NotEnoughCapacity) => {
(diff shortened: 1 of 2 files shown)


quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -782,29 +837,63 @@ mod tests {
...
             let solution_2 = solve(problem.clone(), solution_1.clone());
-            // TODO: This assert actually fails for some scenarii. We say it is fine
-            // for now as long as the solution does not change again during the
-            // next resolution:
-            // let has_solution_changed_once = solution_1.indexer_assignments != solution_2.indexer_assignments;
-            // assert!(!has_solution_changed_once, "Solution changed for same problem\nSolution 1:{solution_1:?}\nSolution 2: {solution_2:?}");
-            let solution_3 = solve(problem, solution_2.clone());
-            let has_solution_changed_again = solution_2.indexer_assignments != solution_3.indexer_assignments;
-            assert!(!has_solution_changed_again, "solution unstable!!!\nSolution 1: {solution_1:?}\nSolution 2: {solution_2:?}\nSolution 3: {solution_3:?}");
+            let has_solution_changed_again = solution_1.indexer_assignments != solution_2.indexer_assignments;
```
fulmicoton-dd, line 840: we now have a stricter definition of stability.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -782,29 +837,63 @@ mod tests {
...
+        // which is less than source 2's 1500 mcpu. Placement therefore requires
+        // capacity scaling, causing at least one node to exceed its original capacity.
         let mut problem =
             SchedulingProblem::with_indexer_cpu_capacities(vec![mcpu(3000), mcpu(3000)]);
         problem.add_source(1, NonZeroU32::new(2500).unwrap()); // Source 0
         problem.add_source(1, NonZeroU32::new(2500).unwrap()); // Source 1
         problem.add_source(1, NonZeroU32::new(1500).unwrap()); // Source 2
         let previous_solution = problem.new_solution();
-        let solution = solve(problem, previous_solution);
+        let solution = solve(problem.clone(), previous_solution);
```
fulmicoton-dd, line 885: I removed capacity_scaling_iterations from the Solution. It was only used in this unit test.

Instead I check that we did end up with an inflated solution

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -35,37 +35,110 @@ fn check_contract_conditions(problem: &SchedulingProblem, solution: &SchedulingS
...
-    // to their original capacity.
+    // to their original capacity. This corresponds to inflation level 0.
     inflate_node_capacities_if_necessary(&mut problem);
-    // As a heuristic, to offer stability, we work iteratively
-    // from the previous solution.
+    // We run a few asserts to ensure that the problem is correct.
+    check_contract_conditions(&problem, &previous_solution);
+    let base_problem = problem;
+
+    // Due to the inherent of the bin-packing, it is possible that the first inflation
```
nadav-govari: inherent nature?

nadav-govari (approved):
Approving. To summarize my understand of these changes: Before this change, we could guarantee that the previous solution was the best solution at the given inflation size, but we didn't guarantee that the new solution was optimally packed. Now we can. 

I don't see anything here that is a real regression risk. Though admittedly this code is extremely dense and it is not hard to miss something.

PR #6556, opened 2026-06-25
feat(search): cache never-match queries via the predicate cache

PSeitz:
Repeated super-selective or non-existent term queries over many splits re-ran warmup on every request.

When warmup proves a split empty (a required term has an empty posting list), record a fake empty entry in the existing predicate cache, and consult it before warmup so later requests with the same predicate short-circuit with no storage reads.

### DIFF
--- a/quickwit/quickwit-search/src/leaf.rs
+++ b/quickwit/quickwit-search/src/leaf.rs
@@ -38,7 +38,8 @@ use quickwit_proto::search::{
     SortOrder, SortValue, SplitIdAndFooterOffsets, SplitResourceStats, SplitSearchError,
 };
 use quickwit_query::query_ast::{
-    BoolQuery, CacheNode, QueryAst, QueryAstTransformer, RangeQuery, TermQuery,
+    BoolQuery, CacheNode, HitSet, PredicateCache, QueryAst, QueryAstTransformer, RangeQuery,
+    TermQuery,
 };
 use quickwit_query::tokenizers::TokenizerManager;
 use quickwit_storage::{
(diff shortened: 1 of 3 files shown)


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -621,6 +622,23 @@ fn compute_index_size(hot_directory: &HotDirectory) -> ByteSize {
...
+/// folded into the query AST by `rewrite_request`, so this key distinguishes time
+/// windows; an entry is only ever reused for an identical predicate over the same
+/// window on the same split.
+fn negative_cache_key(query_ast: &QueryAst) -> Option<String> {
+    let inner = match query_ast {
+        QueryAst::Cache(cache_node) => cache_node.inner.as_ref(),
+        other => other,
+    };
+    serde_json::to_string(inner).ok()
+}
```
trinity-1686a: i think it would be interesting for this to also extract Must/Filter arms in BoolQuery (and possibly lonely Should), if any "must" part of the query doesn't match, the whole query necessarily doesn't match

PSeitz: In it's current form it doesn't work as I want it yet. It doesn't match non-existent terms if they are part of a longer chain, e.g. "fielda:doesnotexist" and then "fielda:doesnotexist fieldb:asdf"

PSeitz: Updated to cache and check terms instead of query asts

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -2764,33 +2834,70 @@ mod tests {
...
+        let key = term_absence_cache_key(&Term::from_field_text(body, "hello"));
+        // Stable for the same term, and disjoint from the JSON query keys the `CacheNode`
+        // positive cache stores (which start with '{', never a hex field id).
+        assert!(!key.starts_with('{'));
         assert_eq!(
-            warmup(&searcher, &warmup_info).await.unwrap(),
-            WarmupOutcome::Completed
+            key,
+            term_absence_cache_key(&Term::from_field_text(body, "hello"))
+        );
```
trinity-1686a: this is just testing if `term_absence_cache_key` is pure

trinity-1686a (approved):

quickwit/quickwit-search/src/leaf.rs
PSeitz: yes that test is nonsense, removed it

PR #6558, opened 2026-06-25
Fix gRPC telemetry exporters

guilload:
Default gRPC processors are spawned in a dedicated OS thread without handle on the tokio runtime. Move to experimental async processors. Tested locally.

### DIFF
--- a/quickwit/quickwit-telemetry-exporters/Cargo.toml
+++ b/quickwit/quickwit-telemetry-exporters/Cargo.toml
@@ -13,15 +13,20 @@ license.workspace = true
 [dependencies]
 anyhow = { workspace = true }
 metrics = { workspace = true }
-metrics-opentelemetry = { workspace = true }
 metrics-exporter-prometheus = { workspace = true }
+metrics-opentelemetry = { workspace = true }
 metrics-util = { workspace = true }
 opentelemetry = { workspace = true }
 opentelemetry-appender-tracing = { workspace = true }
 opentelemetry-otlp = { workspace = true, features = ["experimental-grpc-retry", "experimental-http-retry"] }
-opentelemetry_sdk = { workspace = true }
+opentelemetry_sdk = { workspace = true, features = [
+    "experimental_logs_batch_log_processor_with_async_runtime",
+    "experimental_metrics_periodicreader_with_async_runtime",
+    "experimental_trace_batch_span_processor_with_async_runtime",
+] }
 serde_json = { workspace = true }
 time = { workspace = true, features = ["parsing"] }
+tokio = { workspace = true }
 tracing = { workspace = true }
 tracing-opentelemetry = { workspace = true }
 tracing-subscriber = { workspace = true }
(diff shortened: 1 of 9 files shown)


quickwit/quickwit-telemetry-exporters/src/otlp/metrics.rs
```diff
@@ -44,15 +46,17 @@ impl OtlpProtocol {
...
 }
 
+/// Builds the OTLP metrics recorder and its meter provider.
 pub(crate) fn build_recorder(
     service_version: &str,
     otlp_config: &OtlpExporterConfig,
 ) -> anyhow::Result<(OpenTelemetryRecorder, SdkMeterProvider)> {
     let metrics_protocol = otlp_config.metrics_protocol()?;
     let metric_exporter = metrics_protocol.metric_exporter()?;
+    let metric_reader = PeriodicReader::builder(metric_exporter, runtime::Tokio).build();
```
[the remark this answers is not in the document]
fmassot: @trinity-1686a thoughts on this one?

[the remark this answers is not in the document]
trinity-1686a: this seems to only be an issue when using an http exporter (i think we default to grpc, which uses tonic&async)
we can enable the reqwest feature and everything should be alright

quickwit/quickwit-telemetry-exporters/Cargo.toml
```diff
@@ -13,13 +13,17 @@ license.workspace = true
 [dependencies]
 anyhow = { workspace = true }
 metrics = { workspace = true }
-metrics-opentelemetry = { workspace = true }
 metrics-exporter-prometheus = { workspace = true }
+metrics-opentelemetry = { workspace = true }
 metrics-util = { workspace = true }
 opentelemetry = { workspace = true }
 opentelemetry-appender-tracing = { workspace = true }
 opentelemetry-otlp = { workspace = true, features = ["experimental-grpc-retry", "experimental-http-retry"] }
```
trinity-1686a, line 21: per codex, we need `reqwest-client` here, and this should be good to go

trinity-1686a (approved):

quickwit/quickwit-telemetry-exporters/Cargo.toml
guilload: I don't know if we can trust Codex on that one, given the other Codex-based PR that François pushed. There are two ways to fix this:
- use current exporters with blocking http client, don't know if `reqwest-client` does that
- use new experimental async exporters, this PR

guilload: I'll try the other fix, blocking client, probably safer for now.

guilload: Oh, you meant for that Codex P2, not fixing the whole thing just by doing that. Got you t_t My bad... Yeah, this one makes sense.

PR #6564, opened 2026-06-30
Add S3 prefix to reduce rate limiting

fulmicoton-dd:
# Description

We observed rate limiting from s3 on high workloads. Limits are set per prefix so that the fact that we are using ULID is counterproductive and create hotspots.

This PR introduces a new env variable called QW_RANDOM_SPLIT_PREFIX. When enabled, we will name 
splits by moving 4 chars from back of the random section to the front of the split id. We also separate this prefix from the rest of the string with a _.

This PR contains two commits. One tries to encapsulate further the notion of  Ulid.
The indexing split store was relying on ULID to get a notion of creation date that influence eviction.

The split store is meant to be removed once we have compactors. As an interim solution, we added local logic to identify the type of the split id and correctly extract the timestamp.

# Deploy

This change is backward compatible but not forward compatible.
To deploy, one must deploy on all nodes and then flag flip the environment variable.

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexer.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexer.rs
@@ -134,7 +134,7 @@ impl IndexerState {
             io_controls,
         )?;
         info!(
-            split_id=%indexed_split.split_id(),
+            split_id=indexed_split.split_id_str(),
             partition_id=%partition_id,
             "new-split"
         );
@@ -665,7 +665,7 @@ impl Indexer {
             return Ok(());
         }
         let num_splits = splits.len() as u64;
-        let split_ids = splits.iter().map(|split| split.split_id()).join(",");
+        let split_ids = splits.iter().map(|split| split.split_id_str()).join(",");
         info!(
             index=%self.indexer_state.pipeline_id.index_uid,
             source=self.indexer_state.pipeline_id.source_id.as_str(),
--- a/quickwit/quickwit-indexing/src/actors/indexer.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexer.rs
@@ -134,7 +134,7 @@ impl IndexerState {
             io_controls,
         )?;
         info!(
-            split_id=indexed_split.split_id_str(),
+            split_id=%indexed_split.split_id(),
             partition_id=%partition_id,
             "new-split"
         );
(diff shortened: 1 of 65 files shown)


quickwit/quickwit-indexing/src/actors/indexer.rs
```diff
@@ -134,7 +134,7 @@ impl IndexerState {
             io_controls,
         )?;
         info!(
-            split_id=%indexed_split.split_id(),
```
guilload, line 137: `SplitId` doesn't have `Display`?

quickwit/quickwit-indexing/src/models/indexed_split.rs
```diff
@@ -45,8 +44,8 @@ pub struct IndexedSplit {
 }
 
 impl IndexedSplit {
-    pub fn split_id(&self) -> &str {
-        &self.split_attrs.split_id
+    pub fn split_id_str(&self) -> &str {
```
guilload: nit: this is mostly used for debug/info statements `pub fn split_id(&self) -> &SplitId` works just as well. Either way is fine.

quickwit/quickwit-proto/src/types/split_id.rs
```diff
@@ -0,0 +1,213 @@
...
+        }
+    }
+
+    /// Returns the id as a string slice.
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+}
+
+impl AsRef<Path> for SplitId {
```
guilload: Do we need this? It seems dangerous because when we usually need SplitId as `Path`, it generally means we also want the `.split` suffix.

guilload (approved):

quickwit/quickwit-indexing/src/actors/indexer.rs
fulmicoton: it does and Itertools::join works with Display, so removing this. good catch

quickwit/quickwit-indexing/src/models/indexed_split.rs
fulmicoton: you are right.

quickwit/quickwit-proto/src/types/split_id.rs
fulmicoton: yes.

PR #6568, opened 2026-06-30
Reduce log volume

PSeitz-dd:
`rate_limited_tracing` macro (`rate_limited_info` etc.): the suppressed count is now folded into the emitted line as a `num_suppressed=N` field, instead of a separate preceding line. 
**In combination with #6549**, this lets us aggregate on the field and recover the true pre-suppression log count.

3 Types of Log Changes:
1. `INFO` → `DEBUG` (routine per-operation chatter): node pool-add ×4 (the ~907M dominant source), send-to-index-serializer, spawning pipeline, merge schedule/download, leaf split-finished/offsets, offload-to-lambda, resetting pipeline, adding-shards-assignment.

2. `INFO` → rate-limited `INFO` (1/min) (worth a heartbeat): new-split, actor-exit (success), assigning shards, env-var defaults, truncated-shard.

3. `ERROR` → rate-limited `ERROR` (1/min): the recurring lambda invocation failure (kept visible, not flooding).

`publish-new-splits` now carries `num_splits`, `num_docs`, and total on-disk `split_size`.


quickwit/quickwit-common/src/lib.rs
```diff
@@ -151,7 +151,7 @@ pub fn get_bool_from_env(key: &str, default_value: bool) -> bool {
     if let Some(flag_value) = get_bool_from_env_opt(key) {
         flag_value
     } else {
-        info!(default_value=%default_value, "using environment variable `{key}` default value");
+        crate::rate_limited_info!(limit_per_min = 1, default_value=%default_value, "using environment variable `{key}` default value");
```
fulmicoton: ???

quickwit/quickwit-actors/src/spawn_builder.rs
```diff
@@ -406,7 +406,7 @@ async fn actor_loop<A: Actor>(
         | ActorExitStatus::Quit
         | ActorExitStatus::DownstreamClosed
         | ActorExitStatus::Killed => {
-            info!(actor_id, phase = ?exit_phase, exit_status = ?after_process_exit_status, "actor-exit");
+            quickwit_common::rate_limited_info!(limit_per_min = 1, actor_id, phase = ?exit_phase, exit_status = ?after_process_exit_status, "actor-exit");
```
fulmicoton: I don't think we want to rate limit this one

fulmicoton:
I am all for reducing logging but this is a terrible PR!
Why do you want to rate limit the thing that displays the parsed cli options?

quickwit/quickwit-common/src/lib.rs
PSeitz-dd: we have  ~2M logs of that in the last week on the hooray cluster. I'll try to find the call sites and cache them instead of rate-limiting

quickwit/quickwit-actors/src/spawn_builder.rs
PSeitz-dd: I can remove it, but we had ~4M log lines of that in the last week

PSeitz-dd:
> I am all for reducing logging but this is a terrible PR! Why do you want to rate limit the thing that displays the parsed cli options?

It's not parsed CLI options, it's reading environment variables, which we logged ~2M times in the last week on the hooray cluster.
I changed it to cache the env vars instead of rate limiting (since this covered only default values)

quickwit/quickwit-common/src/lib.rs
PSeitz-dd: Added caching to the callsites (with some macro to avoid misuse in the future)

quickwit/quickwit-actors/src/spawn_builder.rs
PSeitz-dd: Removed the rate limiting

fulmicoton:
>I changed it to cache the env vars instead of rate limiting (since this covered only default values)

Yes sorry I meant env var.
And yes that's what I meant. If you see we log a lot of these, then the right move is to check why, and address it the way all of the codebase does.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -189,15 +189,13 @@ fn compute_load_per_shard(shard_entries: &[&ShardEntry]) -> NonZeroU32 {
...
-        NonZeroU32::new(default_load_per_shard).unwrap()
-    });
-    *DEFAULT_LOAD_PER_SHARD
+    let default_load_per_shard = quickwit_common::get_from_env_cached!(
+        u32,
+        "QW_DEFAULT_LOAD_PER_SHARD",
+        PIPELINE_FULL_CAPACITY.cpu_millis() / 4,
+        false
+    );
+    NonZeroU32::new(default_load_per_shard).unwrap()
```
fulmicoton, line 200: nitpick: I don't see the benefit of a macro here. Are we trying to same one LOC?

Less nitpick: if introducing a macro is something we want to do, it should be a separate PR.

quickwit/quickwit-ingest/src/ingest_v2/state.rs
```diff
@@ -598,7 +598,8 @@ impl FullyLockedIngesterState<'_> {
                 }
             }
         }
-        info!(
+        quickwit_common::rate_limited_info!(
```
fulmicoton: rate limited cannot be the right move here!

Generally speaking, rate limiting is rarely a good move for info statement.
If we don't need it (because it is useless or redundant) let's remove it.

Those info statement are good to investigate, when something goes wrong, what chain of events led to the problem.
If you rate limit, we cannot draw any conclusions about the absence of the log statement. We cannot no that indexing did not reach a given step for a given split for instance.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -1121,7 +1121,8 @@ fn build_ingester_insert_change(
     grpc_compression_encoding_opt: Option<CompressionEncoding>,
 ) -> Change<NodeId, IngesterPoolEntry> {
     let chitchat_id = node.chitchat_id();
-    info!(
+    quickwit_common::rate_limited_info!(
```
fulmicoton: No rate limiting on that no. If we have a lot of churn we need to feel its pain and not ignore it.

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -1209,7 +1210,8 @@ async fn setup_searcher(
             match cluster_change {
                 ClusterChange::Add(node) if node.is_searcher() => {
                     let chitchat_id = node.chitchat_id();
-                    info!(
+                    quickwit_common::rate_limited_info!(
```
fulmicoton: same as above.

quickwit/quickwit-common/src/lib.rs
```diff
@@ -156,6 +156,49 @@ pub fn get_bool_from_env(key: &str, default_value: bool) -> bool {
     }
 }
 
+/// Reads and parses an environment variable exactly once, caching the result for the lifetime of
```
fulmicoton: At this point, we should have a env.rs file in quickwit-common to gather all of these helpers.

Also not by implementing your macro only for a two calls, you creating a fracture in the way we deal with env variable.

Some stuff use a macro, some not.

quickwit/quickwit-serve/src/lib.rs
PSeitz: This is by far the biggest block with 700M logs in the last week (90% of all logs), so it may cause some cost for the customer. We did some experiments, so I'm not sure if this may happen on the customer side.

We could also increase the limit_per_min, to have some safeguard against it and still get a better signal. What do you think?

BTW. In combination with https://github.com/quickwit-oss/quickwit/pull/6549 we can easily recover the total sum of these logs.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
PSeitz: The main benefit is to avoid this issue in the future by adding a central way that includes caching. 
An different approach would be to have a central env hashmap in a oncelock, with all env variables in the repo. Which one do you think is better?

I can put it in a different PR

quickwit/quickwit-indexing/src/actors/indexer.rs
```diff
@@ -133,7 +133,7 @@ impl IndexerState {
             index_builder,
             io_controls,
         )?;
-        quickwit_common::rate_limited_info!(
+        debug!(
             limit_per_min = 1,
```
fulmicoton-dd: T_T limit_per_min is still here.

quickwit/quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -500,6 +500,7 @@ impl Handler<AssignShards> for IndexingPipeline {
         // If it is not, it will be respawned soon, and the shards will be assigned afterward.
         if let Some(handles) = &self.handles_opt {
             info!(
+                limit_per_min = 1,
```
fulmicoton-dd:
```suggestion
```

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
fulmicoton-dd: I liked what we did before because this is macro, but this nitpick/ personal taste.
I don't like the hashmap solution.

Macro is ok, but it should be in a separate file, and cover ALL places where we get env. This is NOT the case today.

quickwit/quickwit-indexing/src/source/ingest/mod.rs
```diff
@@ -531,7 +531,7 @@ impl Source for IngestSource {
             .collect();
 
         assert!(!added_shard_ids.is_empty());
-        info!(added_shards=?added_shard_ids, "adding shards assignment");
+        debug!(added_shards=?added_shard_ids, "adding shards assignment");
```
fulmicoton-dd: I don't think this should be debug.

PSeitz: removed the debug

quickwit/quickwit-indexing/src/actors/indexing_pipeline.rs
```diff
@@ -288,7 +288,7 @@ impl IndexingPipeline {
         let index_id = &self.params.pipeline_id.index_uid.index_id;
         let source_id = &self.params.pipeline_id.source_id;
 
-        info!(
+        debug!(
```
fulmicoton-dd: it should stay info. This happens once per creation of spawning indexing pipeline.

quickwit/quickwit-indexing/src/source/ingest/mod.rs
```diff
@@ -430,7 +430,7 @@ impl IngestSource {
             });
             return Ok(());
         }
-        info!(
+        debug!(
```
fulmicoton-dd: this can stay info I think

quickwit/quickwit-search/src/list_fields/mod.rs
```diff
@@ -37,8 +36,9 @@ pub use crate::list_fields::root::root_list_fields;
 /// a JSON type with random field names. This leads to huge memory consumption
 /// when building the response. This is a workaround until a way is found to
 /// prune the long tail of rare fields.
-static FIELD_LIST_SIZE_LIMIT: LazyLock<usize> =
-    LazyLock::new(|| quickwit_common::get_from_env("QW_FIELD_LIST_SIZE_LIMIT", 100_000, false));
+fn field_list_size_limit() -> usize {
+    quickwit_common::get_from_env_cached!(usize, "QW_FIELD_LIST_SIZE_LIMIT", 100_000, false)
```
fulmicoton-dd, line 40: I really fail to see what's improved in this.

fulmicoton-dd (approved):
see comments

quickwit/quickwit-common/src/lib.rs
PSeitz-dd: moved to env.rs
I'll make another PR to move stuff to the macro

PR #6573, opened 2026-07-02
Refactor Postgres migrator

nadav-govari:
## Description

Advisory locks reading: https://flaviodelgrosso.com/blog/postgresql-advisory-locks

### Issue 1
Today, the Postgres migrator begins a transaction, and runs all migrations inside their own transaction block, such that if any one fails, the whole thing gets unwound.

This is undesirable for two reasons:
1. The migrations build on each other; migration 3 failing does not mean theres a problem with migration 2
2. Concurrent DDL cannot be run inside of a transaction block

This refactors the migrator to run each migration directly, without a wrapping block. sqlx already applies migrations transactionally for us.

### Issue 2
Migrations must be successfully applied on startup for a metapod to mark itself ready on a deployment. Sqlx takes out a Postgres advisory lock, which is a Postgres feature that allows for utilizing Postgres' atomicity to lock on custom resources, which you define with a key of `(int, int)`. It then runs the migrations until completion.

This is a problem for any migration that takes more than the readiness timeout, which is currently 5 minutes. The readiness probe will kill the metastore pod mid applying migrations, and will start crashlooping. Additionally, other metastore pods will queue up to acquire this same lock so that they can too apply migrations.

The fix for this is as follows:
1. Migrations that are absolutely required for functionality stay where they are, with diesel guard inspecting all migrations in CI and preventing migrations that acquire exclusive locks on a table from being merged.
2. Migrations that can be applied concurrently, such as index creation, are moved to a second migrations directory introduced in this PR, `postgres_deferred`, and applied async post-deployment. To prevent the same racing behavior described above, Postgres advisory locks are again reused. A pod will attempt to acquire the advisory lock; if it succeeds, it is the migrations leader and will attempt to apply the migrations. If it doesn't, then it will assume another pod has the lock, and continue on with its day. At all points, live traffic is still served.

Finally, ignore_missing on migrations is marked true. This will allow us forwards and backwards compatibility with migrations and binaries, as of this change. It will be up to us to ensure that the code also respects this. 

### How was this PR tested?
Postgres integ tests, local cluster tests incoming.


quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
```diff
@@ -14,135 +14,207 @@
...
-            });
+/// Runs a migrator's pending up-migrations on `conn`. Shared by the required and deferred tracks.
+///
+/// `ignore_missing` tolerates applied rows absent from this migrator's set: rows from the other
+/// track, or migrations from a newer image after a rollback.
+async fn do_migrations(mut migrator: Migrator, conn: &mut PgConnection) -> MetastoreResult<()> {
+    migrator.set_ignore_missing(true);
+    // run_direct takes a Migrate connection directly, sidestepping run()'s Acquire bound that our
+    // pooled connection type doesn't satisfy.
+    migrator.run_direct(conn).await.map_err(|migrate_error| {
```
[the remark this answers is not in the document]
nadav-govari: sqlx only writes into the migrations table on success, and it writes success=true, so this isn't possible, bad false alarm.

quickwit/quickwit-metastore/migrations/postgresql_deferred/README.md
```diff
@@ -0,0 +1,7 @@
+# Deferred PostgreSQL migrations
```
guilload, line 1: Thanks for the README.

guilload: Let's make sure the diesel-guard CI check also applies to that folder.

quickwit/quickwit-metastore/migrations/postgresql_deferred/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+Long, degrade-gracefully-only migrations (e.g. `CREATE INDEX CONCURRENTLY`) run in a background task after readiness. Both tracks share the `_sqlx_migrations` table.
+
+- Version must be globally unique across both dirs (continue the single sequence).
+- Must be idempotent: `-- no-transaction` first line, then `DROP INDEX CONCURRENTLY IF EXISTS ...; CREATE INDEX CONCURRENTLY IF NOT EXISTS ...`.
```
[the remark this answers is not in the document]
guilload: I didn't interpret it that way but maybe because I know the context. Maybe write:
- Must be idempotent: `-- no-transaction` first line, then `<DDL statement>`. and then give two examples, one up, one down.?

quickwit/quickwit-metastore/src/metastore/postgres/metrics.rs
```diff
@@ -12,7 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use quickwit_metrics::{LazyGauge, lazy_gauge};
+use quickwit_metrics::{LazyCounter, LazyGauge, lazy_counter, lazy_gauge};
+
+pub(super) static DEFERRED_MIGRATIONS_APPLY_ERRORS: LazyCounter = lazy_counter!(
+        name: "deferred_migrations_apply_errors_total",
```
guilload:
```suggestion
        name: "deferred_migrations_apply_total",
```
Use a single metric, and track both successes and failures with a label?

quickwit/quickwit-metastore/src/metastore/postgres/metastore.rs
```diff
@@ -137,7 +132,7 @@ impl PostgresqlMetastore {
         )
         .await?;
 
-        run_migrations(&connection_pool, skip_migrations, skip_locking).await?;
+        Migrations::new(connection_pool.clone()).run().await?;
```
guilload: :))))

quickwit/Makefile
```diff
@@ -28,6 +28,8 @@ migrations-lint:
 	@echo "Linting Postgres migrations with diesel-guard"
 	@(command -v diesel-guard >/dev/null || (echo "diesel-guard is not installed. Please install using 'cargo install diesel-guard'." && exit 1))
 	@diesel-guard check quickwit-metastore/migrations/postgresql/
```
guilload, line 30: Would be nice to contribute to diesel-guard to add globing

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
```diff
@@ -14,135 +14,212 @@
...
+
+// The deferred migrations should be attempted by only one metastore pod. We do that by having
+// metastore pods try to acquire a lock to run the deferred migrations. Pods that don't get the lock
+// ignore deferred migrations. The locks are implemented using Postgres advisory locks.
+//
+// Advisory locks are Postgres locks on custom defined resources, represented by a key of two ints.
+// These two ints are arbitrary magic (but stable) numbers that make up the key for the deferred
+// migrations lock.
+const DEFERRED_MIGRATIONS_LOCK_KEY_1: i32 = 424242;
+const DEFERRED_MIGRATIONS_LOCK_KEY_2: i32 = 1;
```
guilload: where's the fun in "1"? 1337? 1776, American revolution? 1789, French revolution? Go crazy :)

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
```diff
@@ -14,135 +14,212 @@
...
-) -> MetastoreResult<()> {
-    let mut tx = pool.begin().await?;
-    let conn = tx.acquire().await?;
+    skip_deferred: bool,
+}
 
-    let mut migrator = get_migrations();
+impl Migrations {
+    pub(super) fn new(connection_pool: TrackedPool<Postgres>) -> Self {
+        Migrations {
```
guilload:
```suggestion
        Self {
```

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
```diff
@@ -14,135 +14,212 @@
...
     skip_locking: bool,
-) -> MetastoreResult<()> {
-    let mut tx = pool.begin().await?;
-    let conn = tx.acquire().await?;
+    skip_deferred: bool,
+}
 
-    let mut migrator = get_migrations();
+impl Migrations {
+    pub(super) fn new(connection_pool: TrackedPool<Postgres>) -> Self {
```
guilload:
```suggestion
    pub(super) fn from_env(connection_pool: TrackedPool<Postgres>) -> Self {
```

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
```diff
@@ -14,135 +14,207 @@
...
-}
 
-async fn check_migrations(migrator: Migrator, conn: &mut PgConnection) -> MetastoreResult<()> {
-    let dirty = match conn.dirty_version().await {
-        Ok(dirty) => dirty,
-        Err(migrate_error) => {
-            error!(error=%migrate_error, "failed to validate PostgreSQL migrations");
+    async fn run_deferred(pool: &TrackedPool<Postgres>, mut migrator: Migrator) -> DeferredOutcome {
+        let mut conn: PgConnection = match pool.acquire().await {
+            Ok(connection) => connection.detach(),
```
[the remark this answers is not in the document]
guilload: Do we need to detach the connection? It doesn't seem to buy us much.

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
```diff
@@ -14,135 +14,212 @@
...
-/// Initializes the database and runs the SQL migrations stored in the
-/// `quickwit-metastore/migrations` directory.
-#[instrument(skip_all)]
-pub(super) async fn run_migrations(
-    pool: &TrackedPool<Postgres>,
+fn get_deferred_migrations() -> Migrator {
+    sqlx::migrate!("migrations/postgresql_deferred")
+}
+
+enum DeferredOutcome {
```
guilload, line 49: I don't care so much about the naming in the code but if these outcomes end up as a metric labels, I'd rather use label values that we already use in the current metrics, such as `success` and `failure`. I always struggle to find the right label names and label values when I query metrics and autocompletion is slow.

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
```diff
@@ -14,135 +14,212 @@
...
-            return Err(MetastoreError::Internal {
-                message: "failed to validate PostgreSQL migrations".to_string(),
-                cause: migrate_error.to_string(),
-            });
-        }
-    };
+/// Runs a migrator's pending up-migrations on `conn`.
+/// `ignore_missing` is how we are able to ensure forwards/backwards compatibility - we don't error
+/// in the case a migration in the db is not present in our migration set.
+async fn do_migrations(mut migrator: Migrator, conn: &mut PgConnection) -> MetastoreResult<()> {
```
guilload:
```suggestion
async fn run_migrations(mut migrator: Migrator, conn: &mut PgConnection) -> MetastoreResult<()> {
```

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
```diff
@@ -14,135 +14,212 @@
...
+        }
     }
 }
 
-async fn check_migrations(migrator: Migrator, conn: &mut PgConnection) -> MetastoreResult<()> {
-    let dirty = match conn.dirty_version().await {
-        Ok(dirty) => dirty,
-        Err(migrate_error) => {
-            error!(error=%migrate_error, "failed to validate PostgreSQL migrations");
+fn internal_error(message: &'static str, cause: impl std::fmt::Display) -> MetastoreError {
```
guilload:
```suggestion
fn internal_error(message: &'static str, cause: String) -> MetastoreError {
```
nit: compile time monomorphization isn't free.

guilload (approved):

quickwit/quickwit-metastore/src/metastore/postgres/metrics.rs
nadav-govari: Applied

quickwit/Makefile
nadav-govari: I think Claude was overeager with this, it seems to also work correctly on a dir with just a Readme.

quickwit/quickwit-metastore/migrations/postgresql_deferred/README.md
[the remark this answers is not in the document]
nadav-govari: Updated, and added comments further clarifying

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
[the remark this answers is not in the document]
nadav-govari: Postgres will only automatically drop an advisory lock for us on the end of the session, or if the connection is dropped. So by detaching this connection, we get Postgres-ensured unlocking, including if we panic. Otherwise, without detaching the connection, we'd be responsible for manually unlocking at the end of our session, which is another query. So it's worth it. I'll add a comment.

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
nadav-govari: lmao. Given the nature of the project, it has to be 1789.

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
nadav-govari: Applied to the metrics - success/failure.

quickwit/quickwit-metastore/migrations/postgresql_deferred/README.md
nadav-govari: Yep, it uses a globstar: https://github.com/quickwit-oss/quickwit/blob/main/.github/workflows/ci.yml#L146

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
```diff
@@ -153,78 +236,267 @@ mod tests {
...
+    #[tokio::test]
+    #[serial_test::file_serial]
+    async fn test_run_required_apply_fails_when_read_only() {
+        let _ = tracing_subscriber::fmt::try_init();
+        let pool = test_pool(true).await;
+        // writing migrations fails in read-only mode
+        migrations(&pool, false)
+            .run_required(test_migrator().await)
+            .await
+            .unwrap_err();
```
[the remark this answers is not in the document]
nadav-govari: Disproven locally

quickwit/quickwit-metastore/migrations/postgresql_deferred/README.md
```diff
@@ -0,0 +1,22 @@
...
+Long, degrade-gracefully-only migrations (e.g. `CREATE INDEX CONCURRENTLY`) run in a background task after readiness. Both tracks share the `_sqlx_migrations` table.
+
+- Version must be globally unique across both dirs (continue the single sequence).
+- Must be idempotent. A statement that can't run in a transaction (e.g. `CREATE INDEX CONCURRENTLY`) needs `-- no-transaction` as the first line, then the DDL. Each statement auto-commits, so a killed migration must be safe to re-run (`CREATE INDEX CONCURRENTLY` can leave an invalid index, so drop it first). For example:
+
+29_add_foo_index.up.sql
+  ```sql
+  -- no-transaction
+  DROP INDEX CONCURRENTLY IF EXISTS foo_idx;
+  CREATE INDEX CONCURRENTLY IF NOT EXISTS foo_idx ON foo (bar);
```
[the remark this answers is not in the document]
nadav-govari: its a readme

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
```diff
@@ -14,135 +14,218 @@
...
+                return DeferredOutcome::Failure;
+            }
+        };
 
-        let Err(migrate_error) = migrate_result else {
-            tx.commit().await?;
-            return Ok(());
+        // This query returns either true or false: we got the lock, or we didn't. It's dropped when
+        // the connection is dropped.
+        let acquired: bool = match sqlx::query_scalar("SELECT pg_try_advisory_lock($1, $2)")
```
trinity-1686a, line 118: why use our own advisory lock if sqlx already has its own?

trinity-1686a (approved):

quickwit/quickwit-metastore/migrations/postgresql_deferred/README.md
guilload: Thank you.

quickwit/quickwit-metastore/migrations/postgresql_deferred/README.md
```diff
@@ -0,0 +1,22 @@
...
+  ```
+
+29_add_foo_index.down.sql
+  ```sql
+  -- no-transaction
+  DROP INDEX CONCURRENTLY IF EXISTS foo_idx;
+  ```
+
+  The `DROP` before the `CREATE` looks like it would drop the index on every run, but it does not: sqlx applies each migration at most once (tracked by version in `_sqlx_migrations`) and never re-runs one that already succeeded. The drop only matters on a retry after a partial failure. If a `CREATE INDEX CONCURRENTLY` is killed midway, Postgres leaves an *invalid* index and sqlx records nothing (the bookkeeping row is written only on success), so the next attempt re-runs the migration. A bare `CREATE INDEX CONCURRENTLY IF NOT EXISTS` would then see that invalid index and skip it forever, so we drop it first to force a clean rebuild.
```
guilload, line 21: Great comment. I don't think we've followed this pattern before so need to be careful when reviewing future changes including migrations.

quickwit/quickwit-metastore/src/metastore/postgres/migrator.rs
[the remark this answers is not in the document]
guilload: Thanks for the comment.

PR #6574, opened 2026-07-03
Run single-token phrase prefix as an uncapped prefix query

fulmicoton-dd:
A phrase prefix query expands its last term over the term dictionary but caps that expansion at `max_expansions` (default 50), making it an approximation of the true set of matching terms. When a phrase prefix tokenizes to a single term, however, it is really just a prefix query, and the cap is an unnecessary source of imprecision — this case is common on untokenized (e.g. raw) fields. We therefore lift the cap for single-term phrase prefixes (`max_expansions = u32::MAX`) and warm up the full term range accordingly.

Closes CLOUDPREM-775

### DIFF
--- a/quickwit/quickwit-query/src/query_ast/phrase_prefix_query.rs
+++ b/quickwit/quickwit-query/src/query_ast/phrase_prefix_query.rs
@@ -158,7 +158,7 @@ mod tests {
     use tantivy::collector::Count;
     use tantivy::query::Query as TantivyQuery;
     use tantivy::schema::{
-        JsonObjectOptions, Schema as TantivySchema, TextFieldIndexing, TextOptions,
+        JsonObjectOptions, Schema as TantivySchema, TEXT, TextFieldIndexing, TextOptions,
     };
     use tantivy::{Index, doc};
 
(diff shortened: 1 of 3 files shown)


trinity-1686a (approved):

quickwit/quickwit-query/src/query_ast/phrase_prefix_query.rs
```diff
@@ -114,24 +131,181 @@ impl BuildTantivyAst for PhrasePrefixQuery {
...
+        let mut schema_builder = TantivySchema::builder();
+        schema_builder.add_text_field("body", tantivy::schema::TEXT);
+        let schema = schema_builder.build();
+
+        // The `default` tokenizer drops punctuation-only input, yielding zero tokens.
+        let query = phrase_prefix_query("body", "!!!", 50);
+        let context = BuildTantivyAstContext::for_test(&schema);
+        let ast = query.build_tantivy_ast_impl(&context).unwrap();
+        assert_eq!(ast, TantivyQueryAst::match_none());
+    }
```
trinity-1686a, line 354: unless i missed it, there is not test to verify we don't increase the expansion when the phrase tokenize in >1 term, we should probably have one

fulmicoton-dd: adding one

PR #6583, opened 2026-07-08
Dont drop and create index in migration 29

nadav-govari:
### Description

Turns out you can't have two statements in one query without Postgres automatically putting it in a transaction, which is a problem when working with CONCURRENTLY.

### How was this PR tested?

Test cluster.


trinity-1686a (approved):
😭

PR #6584, opened 2026-07-08
aws: use caching DNS resolver for S3 HTTP client to reduce lookup churn

fulmicoton-dd:
## Summary
- The AWS SDK's default HTTP client resolves DNS via hyper's uncached `GaiResolver`, causing redundant blocking `getaddrinfo` lookups under high connection churn (e.g. many concurrent S3 range requests).
- Adds a `HickoryDnsResolver` in `quickwit-aws` backed by a caching Hickory resolver, with a configurable minimum TTL floor (`QW_S3_DNS_CACHE_MIN_TTL_SECS`, default 60s) so short-TTL S3 records are still cached long enough to absorb bursts.
- Falls back to the SDK's default resolver if the system DNS config can't be read.

## Test plan
- [x] `bash scripts/check_license_headers.sh`
- [x] `cargo clippy --workspace --all-features --tests`
- [x] `cargo +nightly fmt --all -- --check`
- [x] Unit test resolving `localhost` via the new resolver

### DIFF
--- a/quickwit/quickwit-aws/src/dns.rs
+++ b/quickwit/quickwit-aws/src/dns.rs
@@ -27,6 +27,7 @@ use std::time::Duration;
 use anyhow::Context as _;
 use aws_smithy_runtime_api::client::dns::{DnsFuture, ResolveDns, ResolveDnsError};
 use hickory_resolver::TokioResolver;
+use quickwit_common::rate_limited_warn;
 
 /// Minimum TTL enforced on positive DNS responses.
 ///
@@ -49,7 +50,7 @@ pub struct HickoryDnsResolver {
 impl HickoryDnsResolver {
     /// Builds a resolver from the host's DNS configuration (`/etc/resolv.conf`
     /// on Unix), overriding the cache options so short-TTL records are still
-    /// cached for at least `DNS_CACHE_MIN_TTL` seconds.
+    /// cached for at least `DNS_CACHE_MIN_TTL`.
     pub fn from_system_conf() -> anyhow::Result<Self> {
         let mut builder =
             TokioResolver::builder_tokio().context("failed to read system DNS configuration")?;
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-aws/src/dns.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use anyhow::Context as _;
+use aws_smithy_runtime_api::client::dns::{DnsFuture, ResolveDns, ResolveDnsError};
+use hickory_resolver::TokioResolver;
+
+/// Minimum TTL enforced on positive DNS responses, in seconds.
+///
+/// S3 endpoints advertise TTLs of only a few seconds. Without a floor, the
+/// Hickory cache would expire almost immediately and provide little benefit.
+const DNS_CACHE_MIN_TTL: Duration = Duration::from_secs(60);
```
guilload:
```suggestion
const DNS_CACHE_MIN_TTL: Duration = Duration::from_mins(1);
```

quickwit/quickwit-aws/src/dns.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use std::net::IpAddr;
+use std::sync::Arc;
+use std::time::Duration;
+
+use anyhow::Context as _;
+use aws_smithy_runtime_api::client::dns::{DnsFuture, ResolveDns, ResolveDnsError};
+use hickory_resolver::TokioResolver;
+
+/// Minimum TTL enforced on positive DNS responses, in seconds.
```
guilload:
```suggestion
/// Minimum TTL enforced on positive DNS responses.
```
It's a `Duration`.

quickwit/quickwit-aws/src/lib.rs
```diff
@@ -16,20 +16,48 @@ use aws_config::retry::RetryConfig;
...
+/// default `getaddrinfo`-based DNS resolver for a caching Hickory resolver, see
+/// [`HickoryDnsResolver`]. If the system DNS configuration cannot be read we log
+/// a warning and fall back to the SDK's default resolver rather than failing.
+fn build_http_client() -> SharedHttpClient {
+    let builder = HttpClientBuilder::new().tls_provider(tls::Provider::Rustls(CryptoMode::AwsLc));
+    match HickoryDnsResolver::from_system_conf() {
+        Ok(resolver) => builder.build_with_resolver(resolver),
+        Err(error) => {
+            warn!(
+                error=%error,
```
guilload:
```suggestion
                %error,
```

guilload (approved):
Noice

guilload:
If we fix the connection churn issue, do we actually need DNS caching?

fulmicoton:
> If we fix the connection churn issue, do we actually need DNS caching?

Not really, but yes.
I would really prefer to keep it in case some other problem cause connection churn. It guards searcher from impacting other components. Also it is rather light.

quickwit/quickwit-aws/src/dns.rs
fulmicoton: Ouch. I modified AI generated code.

quickwit/quickwit-aws/src/lib.rs
```diff
@@ -16,20 +16,48 @@ use aws_config::retry::RetryConfig;
...
+}
+
 /// Initialises and returns the AWS config.
 pub async fn get_aws_config() -> &'static aws_config::SdkConfig {
     static SDK_CONFIG: OnceCell<aws_config::SdkConfig> = OnceCell::const_new();
 
     SDK_CONFIG
         .get_or_init(|| async {
             aws_config::defaults(aws_behavior_version())
+                .http_client(build_http_client())
```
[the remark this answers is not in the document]
fulmicoton: i might need to look into that.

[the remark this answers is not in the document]
Platane: codex seems to think there is a way to inject the proxy config using `HttpClientBuilder::new().build_with_connector_fn`

in place of `builder.build_https`

quickwit/quickwit-aws/src/lib.rs
```diff
@@ -16,20 +16,47 @@ use aws_config::retry::RetryConfig;
...
 
+/// Builds the HTTP client used by all AWS SDK clients.
+///
+/// This mirrors `aws-smithy-runtime`'s `default_https_client()` (Rustls +
+/// aws-lc-rs crypto, proxy settings read from the environment,
+/// but wraps the default `getaddrinfo`-based DNS resolver with a cache, see [`CachingDnsResolver`].
+fn build_http_client() -> SharedHttpClient {
+    HttpClientBuilder::new().build_with_connector_fn(move |settings, runtime_components| {
+        let mut conn_builder =
+            Connector::builder().tls_provider(tls::Provider::Rustls(CryptoMode::AwsLc));
```
rdettai-sk, line 41: @fulmicoton FYI `Connector::builder()` enables TCP_NODELAY

```rust
    pub fn builder() -> ConnectorBuilder {
        ConnectorBuilder {
            enable_tcp_nodelay: true,
            ..Default::default()
        }
    }
```

As far as I can tell, the default http client was not. This is a side effect that should be OK (maybe even positive), but worth flagging.

PR #6585, opened 2026-07-09
control-plane: warn when the scheduler is stuck reapplying a non-converging plan

aanufriev-celonis:
## Summary

Make the indexing scheduler's "reapply last plan" control loop observable so a non-converging cluster can be detected and alerted on, instead of failing silently for an unbounded amount of time.

Part of #6601.

## Problem

The control plane runs a periodic control loop (`control_running_plan`) that compares the running plan reported by indexers against the last applied plan. When the tasks differ (same nodes), the loop re-sends the identical plan.

If the last applied plan can never converge, for example, an indexer that never picks up its assigned tasks, or a stale/un-runnable plan referencing a shard that no longer exists, the loop reapplies the same plan every cycle indefinitely. Ingestion for the affected shards stalls, and currently this is invisible: the loop logs the same `info` line every ~30s with no counter, so a transient one-cycle propagation lag is indistinguishable from a cluster that has been stuck for hours, and there is nothing to alert on.

## Change

- Add `num_consecutive_reapplies` to `IndexingSchedulerState`. It increments on each reapply and resets to 0 on convergence or when a new plan is scheduled.

- Once it reaches `REAPPLY_LOOP_WARN_THRESHOLD` (5 cycles, ~2.5 min at the 30s production interval), the per-cycle log escalates from `info` to `warn`, including the plan diff. The missing-vs-unplanned breakdown in the diff distinguishes an un-runnable task from a stale pipeline.

This is purely observability, no scheduling behavior changes.

## Testing

- New unit test `test_control_running_plan_counts_consecutive_reapplies` drives the reapply branch repeatedly, asserts the counter grows, then converges the indexer and asserts it resets to 0.

- `cargo test -p quickwit-control-plane` passes.

- `cargo +nightly fmt` and `cargo clippy -p quickwit-control-plane --all-features --tests` are clean.

### DIFF
--- a/quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
+++ b/quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
@@ -397,8 +397,8 @@ impl IndexingScheduler {
             last_applied_plan.indexing_tasks_per_indexer(),
         );
         if !indexing_plans_diff.has_same_nodes() {
-            // Scheduling a new plan supersedes any pending reapply loop, so reset the counter.
-            self.state.num_consecutive_reapplies = 0;
+            // `rebuild_plan` installs a new plan, which resets the reapply counter in
+            // `apply_physical_indexing_plan`.
             info!(plans_diff=?indexing_plans_diff, "running plan and last applied plan node IDs differ: schedule an indexing plan");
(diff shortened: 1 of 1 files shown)


aanufriev-celonis:
@codex review

PR #6586, opened 2026-07-09
Race condition bugfix in the dnscache.

fulmicoton:
I ended up not using hickory for the cache
in order to support all of the libc subtleties (resolv.conf)

Unfortunately, upon expiration, and until the dns entry is updated, all new connection create triggers a DNS update.

Because our DNS is slightly slow, this was causing a significantly inflated (and confusing) number of DNS calls.

The hickory approach on this was to block all callers.

I am not fond of this solution either: it would synchronize GET for no reason at all, and our TTL is pretty arbitrary. There is no need to pay for DNS latency at all.

This PR chooses to run DNS refresh asynchronously when the TTL is expired and the entry is stale.

### DIFF
--- a/quickwit/quickwit-aws/src/dns.rs
+++ b/quickwit/quickwit-aws/src/dns.rs
@@ -136,7 +136,7 @@ impl ResolveDns for CachingDnsResolver {
         let dns_entry_opt: Option<Arc<DnsEntry>> = cache.get(&host);
         // First insertion CAN trigger several DNS call, but this is not a problem.
         let dns_entry: Arc<DnsEntry> = if let Some(dns_entry) = dns_entry_opt {
-            dns_entry.clone()
+            dns_entry
         } else {
             let dns_entry = Arc::new(DnsEntry::new());
             cache.insert(host.clone(), dns_entry.clone());
--- a/quickwit/quickwit-aws/src/dns.rs
+++ b/quickwit/quickwit-aws/src/dns.rs
@@ -23,53 +23,147 @@
 //! `/etc/hosts`, mDNS, LDAP, etc), but keeps an in-memory cache of the
 //! results so repeated lookups for the same host don't each pay for a
 //! blocking `getaddrinfo` call.
+//!
+//! Cached entries never expire outright: once a host has been resolved at
+//! least once, lookups always return the cached (possibly stale) addresses
+//! immediately. Entries older than [`DNS_REFRESH_COOLDOWN`] instead trigger a
+//! background refresh that updates the cache in place once it completes, so
+//! callers are never blocked waiting on a fresh `getaddrinfo` call for a
+//! host they already have an answer for.
 
 use std::net::IpAddr;
-use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-aws/src/dns.rs
```diff
@@ -23,53 +23,147 @@
...
 
-/// TTL applied to cached lookups.
-const DNS_CACHE_TTL: Duration = Duration::from_mins(1);
+// On the first call, we do not have any answer to give back,
+// so we wait up to DNS_TIMEOUT seconds for an entry to have been populated
+// by the background task.
+const DNS_TIMEOUT: Duration = Duration::from_secs(1);
 
 /// Maximum number of cached hostnames.
+/// In practise this cache should only a few entries to connect to s3's endpoints.
```
trinity-1686a: something is missing in that sentence

quickwit/quickwit-aws/src/dns.rs
```diff
@@ -23,53 +23,147 @@
...
+            //
+            // Relaxed suffices on both arms: this CAS only elects a single winner.
+            self.next_resolve_attempt
+                .compare_exchange(
+                    next_resolve_attempt,
+                    now + DNS_REFRESH_COOLDOWN.as_millis() as u64,
+                    Ordering::Relaxed,
+                    Ordering::Relaxed,
+                )
+                .is_ok()
```
trinity-1686a, line 96: in general i'd say we only want to update entry ttl on success, but in this situation where we will be continuously re-asking for the same domain anyway, i don't think it actually matter; except if we currently have no IP address, and we just doomed ourselves for the next 5s (initial resolution)

quickwit/quickwit-aws/src/dns.rs
```diff
@@ -23,53 +23,147 @@
...
 impl ResolveDns for CachingDnsResolver {
-    fn resolve_dns<'a>(&'a self, name: &'a str) -> DnsFuture<'a> {
+    fn resolve_dns<'a>(&'a self, host: &'a str) -> DnsFuture<'a> {
         let cache = self.cache.clone();
-        let host = name.to_string();
+        let host = host.to_string();
+        let dns_entry_opt: Option<Arc<DnsEntry>> = cache.get(&host);
+        // First insertion CAN trigger several DNS call, but this is not a problem.
+        let dns_entry: Arc<DnsEntry> = if let Some(dns_entry) = dns_entry_opt {
+            dns_entry.clone()
```
trinity-1686a: i think this clone is useless

quickwit/quickwit-aws/src/dns.rs
fulmicoton: the charm of non AI code.

trinity-1686a (approved):

quickwit/quickwit-aws/src/dns.rs
fulmicoton: Yes I thought this was ok and simple here.

PR #6587, opened 2026-07-10
search: fix trace fragmentation and restructure spans

loutPhilipps:
This PR fixes a few issues with how traces are emitted to Datadog:

- nest per-split spans under `multi_index_leaf_search`
- remove duplicate grpc adapter spans
- add more granularity: `plan_splits_for_root_search`, `list_relevant_splits`, `fetch_docs_in_split`, and a per-split wrapper (`leaf_search_single_split_wrapper`) so each split's wait + warmup + tantivy form one subtree
- split the CPU-pool queue wait out of `tantivy_search` into `waiting_for_cpu_pool`.
- surface `search_partial_hits_phase` to have all leaves as children

image

### DIFF
--- a/quickwit/quickwit-common/src/tracing_utils.rs
+++ b/quickwit/quickwit-common/src/tracing_utils.rs
@@ -80,16 +80,6 @@ pub fn extract_context(metadata: &MetadataMap) -> Context {
     global::get_text_map_propagator(|propagator| propagator.extract(&extractor))
 }
 
-/// Extracts a W3C trace context from incoming gRPC request metadata and
-/// installs it as the parent of the currently-active tracing span. Use this
-/// at the entry of a gRPC handler that is itself wrapped in a
-/// `#[tracing::instrument]` so the handler's span is stitched into the
-/// caller's trace.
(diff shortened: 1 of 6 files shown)


quickwit/quickwit-common/src/tracing_utils.rs
```diff
@@ -80,16 +80,6 @@ pub fn extract_context(metadata: &MetadataMap) -> Context {
     global::get_text_map_propagator(|propagator| propagator.extract(&extractor))
 }
 
-/// Extracts a W3C trace context from incoming gRPC request metadata and
-/// installs it as the parent of the currently-active tracing span. Use this
-/// at the entry of a gRPC handler that is itself wrapped in a
-/// `#[tracing::instrument]` so the handler's span is stitched into the
-/// caller's trace.
-pub fn set_current_span_parent_from_metadata(metadata: &MetadataMap) {
```
guilload, line 88: Why is this no longer needed?

loutPhilipps: This is actually still needed here, good catch

PR #6588, opened 2026-07-10
search: add query_ast, counts, and warmup byte span attributes

loutPhilipps:
- add query AST, split and doc count, time range to the `root_search` span attributes
- add `warmup_mb` and `downloaded_mb` as attributes on the warmup span

image

image

### DIFF
--- a/quickwit/quickwit-search/src/leaf.rs
+++ b/quickwit/quickwit-search/src/leaf.rs
@@ -300,7 +300,6 @@ async fn run_cancellable(
 ///
 /// Returns whether the query is provably empty in this split (i.e. `on_absent` fired and
 /// warmup was short-circuited).
-#[instrument(skip_all)]
 pub(crate) async fn warmup(
     searcher: &Searcher,
     warmup_info: &WarmupInfo,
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-search/src/leaf.rs
```diff
@@ -779,7 +783,35 @@ async fn leaf_search_single_split(
...
+            downloaded_mb = tracing::field::Empty
+        );
+        let provably_empty = warmup(&searcher, &warmup_info, &record_absence)
+            .instrument(warmup_span.clone())
+            .await?;
+        warmup_span.record(
+            "warmup_mb",
+            byte_range_cache
+                .get_num_bytes()
+                .saturating_sub(cache_bytes_before_warmup) as f64
```
fulmicoton-dd: I don't think this is correct. Some of that data could have been in the fastfield cache for instance.

It kind on depends what you are interested in. If it is how much bytes are required for that task (regardless of whether it is downloaded or fetch from cache, then you just need to change the name of your field).

If you want the downloaded bytes, a metric exists at the Storage level.

quickwit/quickwit-common/src/tracing_utils.rs
```diff
@@ -90,6 +90,15 @@ pub fn set_current_span_parent_from_metadata(metadata: &MetadataMap) {
     let _ = Span::current().set_parent(parent_context);
 }
 
+/// Records an attribute on the currently-active span's OpenTelemetry span.
+///
+/// Unlike a `tracing` field (`#[instrument(fields(...))]` or `Span::record`), this does
+/// not propagate to log events, so it can carry verbose values (e.g. a query AST)
+/// without bloating logs. No-op when no OpenTelemetry layer is installed.
+pub fn record_current_span_attribute(key: &'static str, value: impl Into<opentelemetry::Value>) {
+    Span::current().set_attribute(key, value);
```
fulmicoton-dd: not he most useful helper. I would rather stick to inlining that code.

fulmicoton-dd (approved):

PR #6594, opened 2026-07-12
fix(config): apply --service override before validation

hushy:
A searcher node started with `--service searcher` could log a misleading "data dir volume too small" warning at startup, comparing its volume against the indexer split-store and ingest queue disk budgets. A searcher does not use those budgets, so the warning was wrong and misleading.

There were two related design issues:

1. the CLI `--service` override was applied after `load_node_config()` had already run service-dependent validation;
2. the disk-capacity check lived in config validation even though it inspects runtime host state.

### What changed

- thread the typed service override from the CLI through node-config loading;
- apply it to the resolved service set before validation, keeping the effective services in `NodeConfig`;
- keep CLI precedence over both `QW_ENABLED_SERVICES` and the config file, matching the previous post-load behavior;
- move the disk-capacity warning from `quickwit-config` to `quickwit-serve` startup;
- measure the data directory's backing disk once and compare it with the aggregate budgets of the effective enabled services;
- include `data_dir`, backing `device`, canonical `mount_point`, and `volume_size` in the warning;
- leave config loading unchanged when no CLI override is supplied;
- cover both CLI-to-config forwarding and CLI precedence over an invalid
  `QW_ENABLED_SERVICES` value with dedicated regression tests.

The budgets remain aggregated because all enabled services share the data-dir volume: budgets that fit individually can still exceed the disk together. A searcher without a configured split cache has zero expected disk usage and does not warn.

### Integration-test follow-up

Applying overrides before validation exposed an invalid setup in the CLI integration-test helper. It passed `QuickwitService::supported_services()`, which includes the standalone `compactor`, while its config left `enable_standalone_compactors` disabled. That only appeared to work previously because the override was applied after validation.

The helper now uses the normal default all-in-one service set, which intentionally excludes the standalone compactor. A code comment documents why `supported_services()` must not be used there.

### How was this PR tested?

- `cargo test -p quickwit-common fs`
- `cargo test -p quickwit-config` (104 passed)
- `cargo test -p quickwit-cli test_run_command_forwards_service_override_before_validation`
- `cargo test -p quickwit-serve disk_usage` (2 passed)
- `cargo test -p quickwit-cli --test cli --features=metrics -- --test-threads=1` (18 passed, 1 ignored)
- `cargo clippy -p quickwit-common -p quickwit-config -p quickwit-serve -p quickwit-cli --all-targets -- -D warnings`
- `bash scripts/check_log_format.sh`
- `git diff --check`

### DIFF
--- a/quickwit/quickwit-cli/src/lib.rs
+++ b/quickwit/quickwit-cli/src/lib.rs
@@ -230,14 +230,25 @@ pub fn start_actor_runtimes(
 }
 
 /// Loads a node config located at `config_uri` with the default storage configuration.
-async fn load_node_config(config_uri: &Uri) -> anyhow::Result<NodeConfig> {
+///
+/// When `service_override` is provided, it is applied before config validation so that
+/// service-dependent validation (e.g. the disk-usage check) reflects the actual runtime service
+/// set (e.g. `--service searcher`).
+async fn load_node_config(
+    config_uri: &Uri,
+    service_override: Option<&HashSet<QuickwitService>>,
+) -> anyhow::Result<NodeConfig> {
     let config_content = load_file(&StorageResolver::unconfigured(), config_uri)
         .await
         .context("failed to load node config")?;
     let config_format = ConfigFormat::sniff_from_uri(config_uri)?;
-    let config = NodeConfig::load(config_format, config_content.as_slice())
-        .await
-        .with_context(|| format!("failed to parse node config `{config_uri}`"))?;
+    let config = NodeConfig::load_with_service_override(
+        config_format,
+        config_content.as_slice(),
+        service_override,
+    )
+    .await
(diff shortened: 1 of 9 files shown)


guilload:
I think a better approach would be to move these disk size checks in the init of each individual services:
1. measure disk size once on start up in `quickwit_serve` 
2. each service (compactor, indexer, searcher) does check in its routine and warn if necessary

The warning should also be improved to specify on which device the size estimation is based.

PSeitz-dd:
I ran into a similar issue for the readonly metastore searcher config flag. config validation fails because it ignores the service override (fix is pretty much the same)

https://github.com/quickwit-oss/quickwit/pull/6548/changes/2027c9e3d4f3c87a462b1def73556c1dea862625

PR #6596, opened 2026-07-12
Replay indexing plans after indexer restarts

fmassot:
## Summary

- record the indexer generation targeted by each indexing-plan dispatch attempt
- replay an unchanged plan when a stable node ID is backed by a new process generation
- include retiring and decommissioning indexers so restarted draining nodes receive an empty plan
- bound plan-application waits for every eligible indexer

This is the second focused extraction from #6592 and is independent of #6595.

## Root cause

The scheduler compared stable node IDs and indexing tasks, but not process generations. An indexer could restart with the same node ID and appear unchanged, causing both periodic reconciliation and a plan rebuild to skip dispatching the current plan.

Ready-indexer apply futures also had no independent hard deadline. A local or otherwise unlayered client that never completed could retain the rebuild-completion guard indefinitely. Draining indexers already had a two-second deadline.

## Behavior

- Ready indexers use a 35-second deadline, above the ordinary 30-second gRPC client timeout.
- Retiring and decommissioning indexers retain the two-second deadline.
- The generation map uses `NodeId` and covers every dispatch-eligible node, including nodes omitted from the physical plan that must receive an empty cleanup plan.
- The map records dispatch attempts, not acknowledgements. A timeout does not prove that the indexer applied the plan.
- This PR does not add plan epochs or solve the pre-existing possibility of out-of-order concurrent plan RPCs.

## Validation

- generation-aware replay through both `control_running_plan` and `rebuild_plan`
- empty-plan replay to a restarted retiring indexer
- bounded waits for Ready, Retiring, and Decommissioning clients
- successful completion of a valid slow Ready-indexer apply
- `cargo test -p quickwit-control-plane --all-features --lib` (123 passed)
- `cargo clippy -p quickwit-control-plane --all-features --tests -- -D warnings`
- `cargo doc -p quickwit-control-plane --all-features --no-deps`
- `bash scripts/check_license_headers.sh`
- `bash scripts/check_log_format.sh`
- `git diff --check`

The regressions exercise the scheduler through `IndexingServiceClient` test clients; this draft does not add a full multi-node process-restart integration test.

### DIFF
--- a/quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
+++ b/quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
@@ -52,17 +52,31 @@ pub(crate) const MIN_DURATION_BETWEEN_SCHEDULING: Duration =
         Duration::from_secs(30)
     };
 
-pub(crate) const APPLY_INDEXING_PLAN_TIMEOUT: Duration = if cfg!(any(test, feature = "testsuite")) {
-    Duration::from_millis(10)
-} else {
-    Duration::from_secs(2)
-};
+// The indexing gRPC client has a 30-second timeout. Keep the Ready deadline above it so ordinary
+// client errors win the race while still bounding clients without a timeout layer.
+pub(crate) const READY_INDEXER_APPLY_PLAN_TIMEOUT: Duration =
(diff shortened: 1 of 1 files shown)


fulmicoton:
I don't understand this PR. @fmassot any idea what this is about?

PR #6599, opened 2026-07-13
ingest: batch WAL truncation in the shard-positions gossip handler

### LINKED ISSUE #6598: ingest v2: WAL truncation blocks persist/init_shards, causing ingest timeouts under load
**Describe the bug**

In ingest v2, the ingester's shard-positions gossip handler holds the ingester's global lock across an entire WAL-truncation pass, which blocks ingestion on that node while the pass runs.

`EventSubscriber<ShardPositionsUpdate>::handle_event` acquires the full ingester lock once (`lock_fully`) and holds it while it truncates/deletes every shard in the update.

`persist` and `init_shards` take that same lock. So the truncation pass is effectively a stop-the-world for that ingester, each truncate/delete performs WAL disk I/O, so on a node with many shards and/or a WAL on slower storage it can hold the lock for tens of seconds.

During that window:

- incoming `persist` requests queue behind the lock and (our case) exceed client ingest timeouts;

- a concurrent `init_shards`, auto-scale-up triggered by the same traffic surge, also blocks and times out, so the source cannot add capacity exactly when it needs to.

This is a contention/timing issue, so it reproduces under load. Conditions that trigger it:

1. Run an ingester (ingest v2) that leads many shards, ideally with the WAL on slower storage.
2. Drive a sustained ingestion surge into a source so the control plane attempts to scale the source up while shard positions are actively gossiped (frequent `ShardPositionsUpdate` truncations).
3. Observe on that ingester: WAL lock-acquire warnings, `ShardPositionsUpdate` handler-budget warnings, `failed to init shards ... request timed out`.

**Expected behavior**

Background WAL cleanup driven by gossiped shard positions must not block ingestion for the duration of the whole pass. Ingest latency (and `init_shards`) should not spike to seconds/tens-of-seconds (up to 50 in our case) because a truncation pass is in progress. The cleanup should yield the ingester lock frequently enough that queued `persist`/`init_shards` calls can interleave, performing small units of work rather than the entire update.

**Configuration:**

quickwit:d33dd237f8250899c223bb2cd7997f005ed616fe

aanufriev-celonis:
Fixes #6598

## Problem

See the original description of https://github.com/quickwit-oss/quickwit/issues/6598.

## Fix

Process the update in small batches (`GOSSIP_TRUNCATE_BATCH_SIZE = 8`), acquiring and releasing the lock **per batch** instead of once for the whole update. Between batches the lock is free, so queued `persist` / `init_shards` calls interleave.

This is turn-processing on the shared WAL lock, not added parallelism: the WAL is a single shared log, and its operations remain serialized. The change only bounds how long the cleanup task holds the lock before yielding it.

### Compatibility

The per-shard helpers already re-validate state on each call, so releasing and re-acquiring the lock mid-update cannot corrupt state:
- `truncate_shard` re-looks up the shard and only advances the truncation position;
- `delete_shard` treats an already-missing WAL queue as success.

## Testing

- Added `test_ingester_truncate_on_shard_positions_update_spans_multiple_batches`, which drives more shards than a single batch (mixed truncate + EOF-delete) and asserts every shard is processed correctly across the multiple lock acquisitions.

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
@@ -1258,27 +1258,56 @@ impl IngesterService for Ingester {
 impl EventSubscriber<ShardPositionsUpdate> for WeakIngesterState {
     #[instrument(name = "ingester.shard_positions_gossip", skip_all)]
     async fn handle_event(&mut self, shard_positions_update: ShardPositionsUpdate) {
+        // Truncating/deleting shards holds the full ingester lock, which `persist` and
+        // `init_shards` also need. Doing the whole update under a single lock can stall ingestion
+        // on this node for the entire pass (many shards, or a slow WAL). Process it in small
+        // batches, releasing the lock between them so writers can interleave. Safe to re-acquire
+        // mid-update: `truncate_shard` re-looks up the shard and only advances the truncation
+        // position, and `delete_shard` treats a missing WAL queue as already deleted.
+        const GOSSIP_TRUNCATE_BATCH_SIZE: usize = 8;
+
         let Some(state) = self.upgrade() else {
             warn!("ingester state update failed");
             return;
         };
+
+        let index_uid = shard_positions_update.source_uid.index_uid;
+        let source_id = shard_positions_update.source_uid.source_id;
+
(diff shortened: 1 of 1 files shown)


aanufriev-celonis:
@codex review

PR #6600, opened 2026-07-13
fix(control-plane): Avoid spawning a new control plan loop on every single actor restart.

fulmicoton-dd:
### Description

Bugfix: before a failure of the control plane actor (due to connectivity with the metastore for instance) would trigger a restart of the control plane actor. A chitchat watching task is monitoring for node membership changes and triggering rebalance shards on this change. That task was spawned in initialized, on a recycled mailbox. As a result, every single restart would start a new watching loop.

This PR starts a loop once and for all.

Also, instead of looking for changes in the cluster membership, it looks for change in the ingester pool itself.

Each pool owns a monotonic generation sender that is updated after mutations are applied. The control plane subscribes once for the lifetime of its supervised mailbox, and the ingest controller tracks the last successfully rebalanced generation so bulk or duplicate notifications are coalesced. Periodic control-loop rebalances remain unconditional.

This also removes the obsolete cluster change stream factory and test stream abstractions from the control-plane API.

### How was this PR tested?

- `cargo test -p quickwit-common tower::pool --lib`
- `cargo test -p quickwit-control-plane --lib`
- `cargo check -p quickwit-serve --lib`

### DIFF
--- a/quickwit/quickwit-cluster/src/change.rs
+++ b/quickwit/quickwit-cluster/src/change.rs
@@ -65,11 +65,6 @@ impl Stream for ClusterChangeStream {
     }
 }
 
-/// A factory for creating cluster change streams.
-pub trait ClusterChangeStreamFactory: Clone + Send + 'static {
-    fn create(&self) -> ClusterChangeStream;
-}
-
 /// Compares the digests of the previous and new set of lives nodes, identifies the changes that
 /// occurred in the cluster, and emits the corresponding events, focusing on ready nodes only.
 pub(crate) async fn compute_cluster_change_events(
@@ -332,8 +327,6 @@ async fn try_new_node(
 
 #[cfg(any(test, feature = "testsuite"))]
 pub mod for_test {
-    use std::sync::{Arc, Mutex};
-
     use quickwit_config::GrpcConfig;
     use tokio::sync::mpsc;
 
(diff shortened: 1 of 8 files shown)


fulmicoton-dd:
does not seem to good to me though

quickwit/quickwit-cluster/src/change.rs
```diff
@@ -65,11 +65,6 @@ impl Stream for ClusterChangeStream {
     }
 }
 
-/// A factory for creating cluster change streams.
-pub trait ClusterChangeStreamFactory: Clone + Send + 'static {
```
guilload, line 69: This deletion makes me happy.

quickwit/quickwit-common/src/tower/pool.rs
```diff
@@ -21,13 +21,18 @@ use std::hash::Hash;
...
+use tokio::sync::watch;
 
 use super::Change;
 
 /// A pool of `V` values identified by `K` keys. The pool can be updated manually by calling the
 /// `add/remove` methods or by listening to a stream of changes.
 pub struct Pool<K, V> {
     pool: Arc<RwLock<HashMap<K, V>>>,
+    /// Publishes a notification after each mutation.
+    pub changes_tx: watch::Sender<()>,
```
guilload: Shouldn't this be private? Only the pool should sent notifications, no?

quickwit/quickwit-control-plane/src/control_plane.rs
```diff
@@ -1110,65 +1100,20 @@ impl Handler<RebalanceShardsCallback> for ControlPlane {
...
-fn spawn_watch_indexers_task(
+fn spawn_watch_ingester_pool(
     weak_mailbox: WeakMailbox<ControlPlane>,
-    cluster_change_stream: ClusterChangeStream,
+    mut ingester_pool_changes_rx: watch::Receiver<()>,
 ) {
-    tokio::spawn(watcher_indexers(weak_mailbox, cluster_change_stream));
-}
-
-async fn watcher_indexers(
```
guilload, line 1130: I'm glad we're getting rid of this complex function, however:
- we no longer log why we rebalance
- we rebalance on every pool change whereas before we would ignore some changes, for instance an ingester joining in initializing status. Rebalance shards is debounced but it's hard to predict whether this change will have undesirable side-effects.

guilload (approved):

quickwit/quickwit-common/src/tower/pool.rs
fulmicoton: yes

quickwit/quickwit-cluster/src/change.rs
fulmicoton: Yep. And claude totally missed that.

I am really scared by how coding agent fail at spotting opportunities to make the code shorter / simpler.

quickwit/quickwit-control-plane/src/control_plane.rs
fulmicoton: One problem that I can foresee, is that our debounce logic is not  retrying if one call happened within the span where the lock was held.

So increasing events could actually remove reaction to valuable events.

The problem was present before.

The rebalance logic is triggered periodically anyway. This watch is here for reactivity. Do you think I should follow up to make sure the debouncer never leaves an event ignored. (meaning if we an event was debounced, we make sure to rerun debouncing at the end of the debouncing period.)

guilload: - So I think for 1. we can improve the logging directly in the pool changes and then have generic logging for rebalancing shards, maybe logging the trigger.

- No, I thought we rebalanced every minute or so but I checked the code and it's actually 5s!

Issue #6601, opened 2026-07-13
control-plane: recover from a non-converging indexing scheduler (stuck "reapply last plan" loop)


aanufriev-celonis:
### Summary

The control plane can get stuck indefinitely in a non-converging indexing-scheduler loop, stalling ingestion for the affected shards until manual restart of indexers.

This issue tracks the staged work to make the loop observable and then self-recovering.

### Background

`control_running_plan` compares what indexers report running with the last applied plan:

- Different nodes -> recompute the plan from the model (rebuild_plan).
- Same nodes, different tasks -> re-send the same plan, assuming an indexer just has not applied it yet.

That assumption breaks when the plan cannot be applied. It assigns a shard that no longer exists (our case) in the indexer's WAL (a phantom `Open` shard after a crash/restart onto a fresh volume). The indexer can never run that task, so `has_same_tasks()` stays `false` forever and the loop re-sends the same plan every ~30s. Reapplying cannot fix a plan that is wrong, it only re-sends it over and over again.

Today this is also invisible: the loop logs the same `info` line each cycle, so a one-cycle propagation lag is indistinguishable from a cluster stuck for hours, and there is nothing to alert on.

### Fix Plan

Stage 1. Observability - #6585

Stage 2. Recovery - #6602 (includes #6585 but can be merged separately, on top)

fulmicoton:
@aanufriev-celonis is this fixing a problem you observed in production?

When you talk about a non-converging indexing-scheduler loop, do you mean:
- the scheduled plan keeps changing despite a successful application of the plan (there was a bug in the scheduler logic that has been fixed... this is not what this is about is it?)
- the impossibility to apply a scheduling plan

Can you detail if this is something you actually observed, what happened?

aanufriev-celonis:
hi @fulmicoton 

This one is a real, recurring production problem. We hit it across many clusters, and it is currently our largest source of ingest `no_shards` errors.

In short, it is your second case: **the impossibility to apply a scheduling plan**.

The node set is stable, so we are on the `has_same_tasks() == false` branch (same nodes, different tasks). The last applied plan assigns a shard that the target indexer cannot run, so the loop re-sends the *identical* plan forever. The plan never changes and it just can never converge.

---

What we observe:

- Indexers log `no open shard found on ingester` continuously for the affected indexes (they log a lot, 500K logs per hour).
[...]

PR #6602, opened 2026-07-13
control-plane: rebuild the plan when the scheduler is stuck reapplying a non-converging plan

aanufriev-celonis:
## Summary

Make the indexing scheduler recover on its own from a persistently non-converging `control_running_plan` loop, instead of re-sending the same un-runnable plan forever.

Part of #6601.

> **Stacked PR.** This PR is built on top of #6585, which is not yet merged, so it currently also contains that PR's commits. Once #6585 is merged, I will rebase the current PR.

## Problem

See the original description of #6601.

## Change

- Add `REAPPLY_LOOP_RESCHEDULE_THRESHOLD` (10 cycles, ~5 min at the 30s production interval), above the existing `REAPPLY_LOOP_WARN_THRESHOLD` so the non-convergence warning still fires first.

- Once `num_consecutive_reapplies` crosses it, the loop calls `rebuild_plan(model)` instead of re-sending the stale plan. If the model has since been corrected, the rebuilt plan is runnable, and the cluster converges (installing a new plan resets the counter via the existing reset-on-new-plan logic).

- Rebuilds are throttled to once every `REAPPLY_LOOP_RESCHEDULE_THRESHOLD` cycles to avoid rebuild storms when the model still describes an unreachable state.

## Testing

- New unit test `test_control_running_plan_rebuilds_after_persistent_reapplies`: drives the reapply branch below the threshold (asserting the un-runnable task stays planned and the counter grows), then asserts the threshold cycle rebuilds from the model, drops the task, and resets the counter.

### DIFF
--- a/quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
+++ b/quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
@@ -329,6 +329,23 @@ impl IndexingScheduler {
     // Prefer not calling this method directly, and instead call
     // `ControlPlane::rebuild_indexing_plan_debounced`.
     pub(crate) fn rebuild_plan(&mut self, model: &ControlPlaneModel) {
+        // The regular rebuild seeds from the last applied plan to keep placement stable.
+        self.rebuild_plan_impl(model, None);
+    }
+
+    /// Rebuilds the physical plan from the current model.
+    ///
+    /// `seed_override` controls the plan the placement solver starts from:
+    /// - `None` seeds from `last_applied_physical_plan` (the default, keeps placement stable).
+    /// - `Some(seed)` seeds from the provided plan instead. The escalation path in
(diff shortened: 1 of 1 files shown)


aanufriev-celonis:
I had a couple more ideas in mind on how to tackle this problem, but decided to go with the least intrusive one. Below are some thoughts about alternative paths:

- teach indexers to communicate back to CP the plan apply results, but that is quite some protocol work, would be more changes in the end;
- try something around TTL, but I guess that idea didn't go far, I have found this commit from the past and I guess the idea was left: https://github.com/quickwit-oss/quickwit/commit/4b6a2402a26d7aad5cbc5702248c028b7864a89e;
- fix the model: currently, indexers after restart report reachable shared, but do not report lost ones, so CP does delete/trucate only those that are reported, and keeps ones that were not reported -> inconsistency; still, it can delete/close what was not reported, but this is a destructive change, so I decided to not go this way.

aanufriev-celonis:
@codex review

PR #6609, opened 2026-07-16
improve s3 retry log

PSeitz:


quickwit/quickwit-common/src/retry.rs
```diff
@@ -186,17 +190,22 @@ where
...
+    request_name: &'static str,
+    retry_params: &RetryParams,
+    f: impl Fn() -> Fut,
+) -> Result<U, E>
 where
     Fut: Future<Output = Result<U, E>>,
     E: Retryable + Debug + 'static,
 {
-    retry_with_mockable_sleep(retry_params, f, TokioSleep).await
+    retry_with_mockable_sleep(retry_params, request_name, f, TokioSleep).await
```
fulmicoton, line 202: are there cases where we don't know which method had a problem?

I thought it was mostly:
searcher -> GET
indexer -> PUT

PSeitz: for search it should be GET, for indexing is should be PUT, for merge it could be either. But would be good to be sure from the logs

PSeitz: The log currently just logs `request failed`, without more context you don't know it's S3

fulmicoton:
The extract context is useless. The caller is logging an error already.
View in context has the information

PR #6634, opened 2026-07-27
Add fingerprint-based log clustering

Mallets:
## Summary

This PR adds experimental fingerprint-based log clustering to improve split compression.

- Compute a configurable fingerprint for each document.
- Cluster documents with similar structures and field values.
- Order larger clusters first while preserving insertion order within each cluster.
- Pass the resulting split-local document ID mapping to Tantivy through `finalize_with_doc_id_mapping`.
- Apply clustering only when finalizing fresh splits. Later merges preserve the ordering of their input splits without clustering again.
- Add clustering and published-split metrics for measuring effectiveness and operational cost.

## Motivation

Logs normally enter Tantivy segments in arrival order, which is driven by batching, load balancing, and event timing rather than message structure. A segment can therefore interleave unrelated templates:

```text
api:    "server started at 8080"
worker: "job 123 finished in 42ms"
api:    "server started at 9090"
db:     "connection from 1.2.3.4"
worker: "job 456 finished in 13ms"
```

Stored log bodies and repeated fields compress better when similar documents are adjacent. Clustering documents before segment finalization gives the doc-store compressor longer repeated runs without changing the logical query model.

## Design

A `fingerprinter` computes two independent hashes:

1. A schema hash from the sorted set of leaf JSON paths, excluding configured paths.
2. A fields hash from configured raw string values and tokenized signatures.

Tokenized fields hash token types rather than literal values, up to 50 tokens. This groups volatile values with the same shape:

```text
"server started at 8080" -> Word Gap Word Gap Word Gap Number
"server started at 9090" -> Word Gap Word Gap Word Gap Number
```

```text
Raw document
    |
    v
DocProcessor
    |  computes Fingerprint { schema, grouping }
    v
ProcessedDoc { doc, fingerprint_opt, ... }
    |
    v
Indexer
    |  records each split-local doc ID in DocIdClusterer
    v
IndexedSplitBuilder::finalize
    |  builds a DocIdMapping: largest schema groups first,
    |  then largest grouping-hash groups within each schema
    v
Tantivy segment with similar documents stored together
```

The fingerprint exists only in the in-memory indexing pipeline (it's not stored in the split). Merge operations use Tantivy’s native merge path and do not compute fingerprints or reorder documents again.

The clustering flow is:

1. `DocProcessor` computes a fingerprint from the processed JSON document.
2. `Indexer` push a `(fingerprint_opt, doc_id)` into a `DocIdClusterer`.
3. At split finalization, `DocIdClusterer` creates the Tantivy `DocIdMapping`:
  - Structure groups are ordered by document count.
  - Fields-hash clusters within each structure group are ordered by document count.
  - Documents within each cluster retain insertion order.
  - Documents without fingerprints are appended in insertion order.
  - The resulting permutation is passed to Tantivy as a `DocIdMapping`.

## Benefits

- Recurring schemas, message templates, and stable dimensions gain better compression locality.
- Search semantics, checkpoints, split metadata, and user-visible schemas remain unchanged.
- Dynamic log schemas do not require a fixed sort schema.

## Trade-offs and risks

- Manual doc ID mapping writes a temporary uncompressed doc store and then rewrites it in permuted, compressed order, increasing local disk traffic.
- Fingerprinting adds ingest CPU, particularly for tokenized fields.
- Fingerprints provide best-effort locality, not a total order or query-pruning structure.
- Policies that include volatile data can under-group, while broad exclusions can over-group. Compression gains therefore depend on workload entropy and should be measured on a representative corpus together with CPU, memory, and temporary disk writes.
  
## Benchmark results

I ran this implementation across controlled and live workloads. Document clustering reduced stored bytes by approximately `11–26%` and improved compression ratios by `12–35%`, depending on the dataset. The benefit persisted across merge levels, confirming that write-time ordering survives later compaction.

Indexing throughput remained near baseline, and document reordering accounted for approximately `1%` of indexer CPU in live profiling. The main cost was local disk I/O: the manual document ID mapping path increased temporary writes by approximately `40%`. Process memory was neutral in controlled write-side profiling.

## Configuration

Document clustering is disabled by default and enabled when the node configuration contains `docs_clustering`:
```yaml
docs_clustering:
  - fingerprint:
      - kind: structure
        exclude: [custom]
  - fingerprint:
      - kind: raw
        path: status
      - kind: raw
        path: service
      - kind: tokenized
        path: message
```

Field kinds:
- `structure` hashes the document structure after removing excluded paths.
- `raw` hashes the exact string value at the configured path.
- `tokenized` hashes the token types of the first `50` tokens instead of their literal values, grouping messages with the same shape.

Clustering can be disabled process-wide, even when configured, with:

```sh
QW_DISABLE_DOCS_CLUSTERING=true
```

Changing the policy or environment override requires restarting the process.

The current implementation requires exactly two fingerprint policies:

- The first policy contains exactly one structure field. Its optional exclude list removes paths from the structure fingerprint.
- The second policy contains one or more raw or tokenized fields.

The support of an arbitrary number of fingerprint policies is left as a follow-up of this PR.

## Metrics

The PR adds metrics for:

- document cluster size;
- published split count;
- documents in published splits;
- stored bytes in published splits;
- uncompressed document bytes in published splits;
- published split-size distribution.
 
Published-split metrics share the `index`, `source`, and `merge_ops` labels, allowing stored bytes per document and effective compression ratios to be compared across merge levels.


docs/configuration/node-config.md
```diff
@@ -34,6 +35,33 @@ A commented example is available here: [quickwit.yaml](https://github.com/quickw
 | `default_index_root_uri` | Default index root URI that defines the location where index data (splits) is stored. The index URI is built following the scheme: `{default_index_root_uri}/{index-id}` | `QW_DEFAULT_INDEX_ROOT_URI` | `{data_dir}/indexes` |
 | environment variable only | Log level of Quickwit. Can be a direct log level, or a comma separated list of `module_name=level` | `RUST_LOG` | `info` |
 
+## Docs sorting configuration
+
+This section configures document sorting, which groups similar documents during indexing. It is optional and is disabled by default. Defining a `docs_sorting` policy enables document sorting unless `QW_ENABLE_DOCS_SORTING=false` is set. Setting `QW_ENABLE_DOCS_SORTING=true` does not enable document sorting by itself: a `docs_sorting` policy must still be present in the node configuration.
+
+The `docs_sorting.grouping` section defines how each document fingerprint is computed.
```
fulmicoton-dd: then maybe it should be called fingerprint?

docs/configuration/node-config.md
```diff
@@ -34,6 +35,33 @@ A commented example is available here: [quickwit.yaml](https://github.com/quickw
...
+        kind: structure
+        exclude: [tags]
+    grouping:
+      fields:
+        - path: status
+          kind: raw
+        - path: message
+          kind: tokenized
+```
+
```
fulmicoton-dd: This config definition is incredibly complicated? Can we do any simpler?

docs/configuration/node-config.md
Mallets: I think the change should be in the description and not on the section name `...defines how documents are grouped for sorting`.
I would instead change `fields` to `fingerprint`.

```yaml
docs_sorting:
  grouping:
    fingerprint:
      ...
```

It makes it explicit that we group by fingerprint.

Done in https://github.com/quickwit-oss/quickwit/pull/6634/changes/c611552a672798f275c50b39b35e5f26bd9d8004. Happy to revert if you don't agree with it.

docs/configuration/node-config.md
Mallets: Could you elaborate on what do you find incredibly complicated?

quickwit/quickwit-indexing/src/docs_clustering/fingerprinter.rs
```diff
@@ -0,0 +1,429 @@
...
+            hasher.write_u8(PATH_SEPARATOR);
+        }
+    }
+
+    fn hash_grouping_fields(&self, json_value: &JsonValue, hasher: &mut FnvHasher) {
+        for (path, kind) in self.grouping_fields.iter() {
+            // Grouping policies operate on strings. Missing fields and values of other JSON types
+            // are intentionally omitted from the grouping fingerprint.
+            let Some(value) = get_leaf_string(json_value, path) else {
+                continue;
```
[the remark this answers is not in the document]
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6634/changes/1ba0b59fc7be0f7e2dd0af78dc1fdfa9ce52d857

quickwit/quickwit-indexing/src/docs_clustering/fingerprinter.rs
```diff
@@ -0,0 +1,429 @@
...
+        walk(self, value, &mut current, &mut paths);
+        paths.sort_unstable();
+
+        for path in paths {
+            let mut first = true;
+            for component in path {
+                if !first {
+                    hasher.write(b".");
+                }
+                hasher.write(component.as_bytes());
```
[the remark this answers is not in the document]
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6634/changes/ea9a0eb72da26207b2ee0234bd571f09fad78ca8

docs/configuration/node-config.md
```diff
@@ -34,6 +35,33 @@ A commented example is available here: [quickwit.yaml](https://github.com/quickw
 | `default_index_root_uri` | Default index root URI that defines the location where index data (splits) is stored. The index URI is built following the scheme: `{default_index_root_uri}/{index-id}` | `QW_DEFAULT_INDEX_ROOT_URI` | `{data_dir}/indexes` |
 | environment variable only | Log level of Quickwit. Can be a direct log level, or a comma separated list of `module_name=level` | `RUST_LOG` | `info` |
 
+## Docs sorting configuration
+
+This section configures document sorting, which groups similar documents during indexing. It is optional and is disabled by default. Defining a `docs_sorting` policy enables document sorting unless `QW_ENABLE_DOCS_SORTING=false` is set. Setting `QW_ENABLE_DOCS_SORTING=true` does not enable document sorting by itself: a `docs_sorting` policy must still be present in the node configuration.
```
fulmicoton: let's remove this doc, this feature is too complex and too experimental to be a public feature.

If there are info we want somewhere, you can put it in docs/internal (these are not published).

Just prepend some disclaimer saying this is unstable/Experimental

quickwit/quickwit-config/src/node_config/serialize.rs
```diff
@@ -1294,6 +1303,123 @@ mod tests {
...
+        "#;
+        let node_config = load_node_config_with_env(
+            ConfigFormat::Yaml,
+            config_yaml.as_bytes(),
+            &HashMap::new(),
+            None,
+        )
+        .await
+        .unwrap();
+        let config = node_config.docs_sorting_config.unwrap();
```
fulmicoton: so by default it is disabled?

nitpick: I think today we express defautls like this by using the opposite flag QW_DISABLE_
I have personally no opinion on whether that's a good idea or not, but that's what the existing codebase does (I believe?)

Mallets: Yes, by default is disabled unless enabled in the config. Usually I always use the positive formulation of `enabled` instead of `disabled`. But I have nothing against it to have it `QW_DISABLE_` :)

quickwit/quickwit-indexing/src/actors/doc_processor.rs
```diff
@@ -492,13 +496,22 @@ impl DocProcessor {
...
+            .as_ref()
+            .map(|fingerprinter| fingerprinter.fingerprint(&json_value));
+        let JsonValue::Object(json_obj) = &mut json_value else {
+            unreachable!("document JSON value is always an object")
+        };
 
         let (partition, doc) = self
             .doc_mapper
-            .doc_from_json_obj(json_doc.json_obj, json_doc.num_bytes as u64)?;
+            .doc_from_json_obj(std::mem::take(json_obj), json_doc.num_bytes as u64)?;
```
fulmicoton: the use of std::mem:take is weird here.

```suggestion
  let JsonValue::Object(json_obj) = json_value else {
            unreachable!("document JSON value is always an object")
        };
        let (partition, doc) = self.doc_mapper.doc_from_json_obj(json_obj, num_bytes as u64)?;
```

quickwit/quickwit-indexing/src/actors/indexer.rs
```diff
@@ -131,6 +133,9 @@ impl IndexerState {
             self.indexing_directory.clone(),
             index_builder,
             io_controls,
+            self.fingerprinter_opt
+                .is_some()
+                .then(DocIdClusterer::default),
```
fulmicoton: Please, let's not use fluent stuff on bool. This just makes the code unreadable.

if self.fingerprinter_opt.is_some() {
    Some(DocIdClusterer::default())
} else {
    None
}

Last kid on the block : https://doc.rust-lang.org/beta/std/primitive.bool.html#method.then_some

quickwit/quickwit-indexing/src/models/indexed_split.rs
```diff
@@ -131,9 +139,33 @@ impl IndexedSplitBuilder {
         )
     )]
     pub fn finalize(self) -> anyhow::Result<IndexedSplit> {
-        let index = self.index_writer.finalize()?;
+        let split_attrs = self.split_attrs;
+        let index = match self.doc_id_clusterer_opt {
```
fulmicoton: nitpick: if let Some() is a tiny bit nicer in my opinion

fulmicoton (approved):

fulmicoton:
@Mallets approved. I had a bunch of nitpicks. the use of mem::take at least should be fixed I believe. I leave it to you to fix the others.

quickwit/quickwit-indexing/src/actors/doc_processor.rs
Mallets: Fixed in https://github.com/quickwit-oss/quickwit/pull/6634/commits/143e287c4cb6e6b29ea9c3e19e5e808e8e2450b7

quickwit/quickwit-indexing/src/actors/indexer.rs
Mallets: Fixed in https://github.com/quickwit-oss/quickwit/pull/6634/commits/143e287c4cb6e6b29ea9c3e19e5e808e8e2450b7

quickwit/quickwit-indexing/src/models/indexed_split.rs
Mallets: Fixed in https://github.com/quickwit-oss/quickwit/pull/6634/commits/143e287c4cb6e6b29ea9c3e19e5e808e8e2450b7

fulmicoton (approved):

fulmicoton-dd (approved):

quickwit/quickwit-indexing/src/actors/indexer.rs
guilload: that was a `map` in the first place, but you don't like that either :D

PR #6637, opened 2026-07-27
restrict bounds on cardinality test

### LINKED ISSUE #6294: Cardinality aggregation on small count seems inaccurate.
Tantivy uses a hll sketch for cardinality. This sketch is supposed to adapt itself internally for low cardinality.
First it lists coupons, then it upgrade to a set, then it upgrade to an actual hll sketch.

We observe this behavior in tantivy, and counts seem correct after union.
In Quickwit however, the results are wrong as highlighted by this test.

https://github.com/quickwit-oss/quickwit/blob/main/quickwit/rest-api-tests/scenarii/aggregations/0001-aggregations.yaml#L339-L366

(assert should be passing with an equality test)

We do not know yet the cause. (serialization code promoting to HLL? union code not doing what it is supposed to do?)

trinity-1686a:
the underlying issue has been fixed when upgrading to datasketch 0.3

fix https://github.com/quickwit-oss/quickwit/issues/6294

### DIFF
--- a/quickwit/rest-api-tests/scenarii/aggregations/0001-aggregations.yaml
+++ b/quickwit/rest-api-tests/scenarii/aggregations/0001-aggregations.yaml
@@ -356,14 +356,11 @@ json:
 expected:
   aggregations:
     unique_names:
-      value:
-        $expect: 'abs(val - 8) <= 2'
+      value: 8.0
     unique_response:
-      value:
-        $expect: 'abs(val - 6) <= 2' # TODO: Check. The correct number is 6
+      value: 5.0
     unique_dates:
-      value:
-        $expect: 'abs(val - 6) <= 3'
+      value: 6.0
 ---
 # Test extended stats aggregation
 method: [GET]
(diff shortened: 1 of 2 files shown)


quickwit/rest-api-tests/scenarii/aggregations/0001-aggregations.yaml
```diff
@@ -356,14 +356,11 @@ json:
 expected:
   aggregations:
     unique_names:
-      value:
-        $expect: 'abs(val - 8) <= 2'
+      value: 8.0
     unique_response:
-      value:
-        $expect: 'abs(val - 6) <= 2' # TODO: Check. The correct number is 6
```
trinity-1686a, line 363: the correct number was not 6, the values we have are
- 20
- 30
- 100
- 120
- 130

so 5

PSeitz (approved):

quickwit/rest-api-tests/scenarii/aggregations/0001-aggregations.yaml
fulmicoton: yeaaaah

fulmicoton (approved):

PSeitz-dd (approved):

PR #6638, opened 2026-07-27
Handle non-advertisable shards in idle task and during decommissioning

guilload:
### Description
A non-advertisable shard is a shard that was created by an ingester but is unknown to the control plane. This can happen when the control plane performs a successful init shards RPC on ingester but somehow the RPC response did not make it to the control plane (network issue?). For a shard to become advertisable, it needs to receive a persist or fetch RPC, which only happens if a router or an indexer knows about it, which means the control plane knows about.

~~If a shard has been non-advertisable, idle for a while, and empty, it is safe to delete. That's commit 6be93316e2c8b67f841041355c3d79a3450d07c6~~

If a shard is non-advertisable and empty, it should not prevent ingesters from decommissioning. That's commit 6b38a79f4284038eae7f0f57a21dd0c4ecb2698d

### How was this PR tested?
Added / updated unit tests

### DIFF
--- a/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
+++ b/quickwit/quickwit-ingest/src/ingest_v2/ingester.rs
@@ -3872,6 +3872,35 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_check_decommissioning_status_with_empty_orphan_shard() {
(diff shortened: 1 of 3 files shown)


nadav-govari (approved):

guilload:
Removed idle commit. It's not useful.

PR #4877, opened 2024-04-17
Add code to spread and maintain the phase of the different indexing p…

fulmicoton:
…ipeline.

Closes #4876

PR #6655, opened 2026-08-03
Add an option to spread indexing pipelines over time

loutPhilipps:
> [!TIP]
> This PR is better reviewed commit-by-commit.

## Problem

An indexer node's pipelines are started at the same time with the same commit timeout, leading to them committing at the same time. This is not an efficient usage of resources when there are many pipelines on a single indexer node.

This problem was addressed in https://github.com/quickwit-oss/quickwit/pull/4877 with the cooperative indexing feature by introducing both a time spread of pipelines, as well as a semaphore to prevent running too many pipelines concurrently.

When having a large index though, we don't really want to have a semaphore, but do want to spread indexing pipelines.

## What this PR does

Add an option `enable_spread_indexing_pipelines` on the indexer's config. For backwards compatibility reasons, enabling cooperative indexing also enables this feature, with the addition of a semaphore.

The implementation of `CooperativeIndexingCycle` has been mostly left intact, with the `indexing_permits` becoming optional.


quickwit/quickwit-indexing/src/actors/indexer.rs
```diff
@@ -548,12 +543,12 @@ impl Indexer {
...
-                CooperativeIndexingCycle::new(
+        let indexing_cycle_opt: Option<IndexingCycle> =
+            (spread_indexing_pipelines || cooperative_indexing_permits_opt.is_some()).then(|| {
+                IndexingCycle::new(
                     &pipeline_id,
                     indexing_settings.commit_timeout(),
-                    cooperative_indexing_permits,
+                    cooperative_indexing_permits_opt,
                 )
             });
```
loutPhilipps, line 554: This is where we enable the `IndexingCycle`, either if spreading pipelines or if cooperative indexing is enabled

quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -420,6 +423,7 @@ impl IndexingService {
             split_store,
             max_concurrent_split_uploads_index,
             cooperative_indexing_permits: self.cooperative_indexing_permits.clone(),
+            spread_indexing_pipelines: self.spread_indexing_pipelines,
```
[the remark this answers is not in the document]
loutPhilipps: This change will not modify the unused metrics indexing pipelines.

quickwit/quickwit-indexing/src/actors/indexer.rs
```diff
@@ -548,12 +543,12 @@ impl Indexer {
             // A configured fingerprinter supplies the mapping when the split is finalized.
             manual_doc_id_mapping: fingerprinter_opt.is_some(),
         };
-        let cooperative_indexing_opt: Option<CooperativeIndexingCycle> =
-            cooperative_indexing_permits_opt.map(|cooperative_indexing_permits| {
-                CooperativeIndexingCycle::new(
+        let indexing_cycle_opt: Option<IndexingCycle> =
+            (spread_indexing_pipelines || cooperative_indexing_permits_opt.is_some()).then(|| {
+                IndexingCycle::new(
```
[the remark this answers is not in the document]
loutPhilipps: This was already the case with cooperative indexing

quickwit/quickwit-indexing/src/actors/indexing_cycle.rs
```diff
@@ -329,27 +328,21 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_cooperative_indexing_nudge_to_phase() {
+    async fn test_indexing_cycle_nudge_to_phase() {
         tokio::time::pause();
-        let num_threads = 10;
         let num_pipelines = 100;
         let num_steps = 15;
-        let semaphore = Arc::new(Semaphore::new(num_threads));
```
loutPhilipps, line 337: My two cents is that having the semaphore in the reconcile test doesn't change anything to the test so I decided to test the general case instead. I can add another test for nudging to phase with cooperative indexing if needed.

fulmicoton-dd:
@loutPhilipps Your current approach only adds the phase at the start of the pipeline.
It is a little bit complicated, but the indexer has two different usage.

- cooperative indexing: we do not use that. It commits most often when the mailbox is drained and then, goes to sleep for a while. The idea here is that when you have 1000 pipeline to run, you want to keep the code organized like before but you want them to take turns.
- no cooperative indexing: that's what we do. we commit on committimeout. RIght now your PR does the rephasing on pipeline startup, but you need to maintain this phase by nudging the 
```rust
ctx.schedule_self_msg(
   self.indexing_settings.commit_timeout(),
   commit_timeout_message,
);
```
as well.

Also we cannot merge something like that without testing!
If we just merge PR too lightly, we will end up with software that is unmaintainable.

docs/configuration/node-config.md
```diff
@@ -244,6 +244,7 @@ This section contains the configuration options for an indexer. The split store
...
 | `cpu_capacity` | Advisory parameter used by the control plane. The value can expressed be in threads (e.g. `2`) or in term of millicpus (`2000m`). The control plane will attempt to schedule indexing pipelines on the different nodes proportionally to the cpu capacity advertised by the indexer. It is NOT used as a limit. All pipelines will be scheduled regardless of whether the cluster has sufficient capacity or not. The control plane does not attempt to spread the work equally when the load is well below the `cpu_capacity`. Users who need a balanced load on all of their indexer nodes can set the `cpu_capacity` to an arbitrarily low value as long as they keep it proportional to the number of threads available. | `num threads available` |
 | `enable_cooperative_indexing` | Enable sharing resources more efficiently when the number of indexes actively written to is significantly higher than the number of cores but might decrease the overall indexing throughput. | `false` |
+| `enable_spread_indexing_pipelines` | Enable spreading indexing pipelines in time to make sure they are not in sync. | `false` |
```
fulmicoton-dd, line 247: can we make that an environment flag? (the goal is just to make it as light as possible).

PR #6658, opened 2026-08-05
upgrade tantivy

trinity-1686a:


PSeitz-dd (approved):
I think the codex comment makes sense, otherwise LGTM

PR #6662, opened 2026-08-06
split search failures in outcome metrics

PSeitz:
Record storage, warmup, Tantivy execution failures, and panics as `category="error"` in `quickwit_search_split_search_outcome` metric

### DIFF
--- a/docs/reference/metrics.md
+++ b/docs/reference/metrics.md
@@ -79,7 +79,7 @@ PostgreSQL-backed metastores also expose connection pool gauges:
 
 | Namespace | Metric Name | Description | Type |
 | --------- | ----------- | ----------- | ---- |
-| `quickwit_search` | `leaf_searches_splits_total` | Number of leaf searches (count of splits) started | `counter` |
+| `quickwit_search` | `split_search_outcome` | Number of local leaf split search outcomes by `category` (`success`, operational `error`, cache/pruning, or cancellation phase). Retries are counted separately | `counter` |
 | `quickwit_search` | `leaf_search_split_duration_secs` | Number of seconds required to run a leaf search over a single split. The timer starts after the semaphore is obtained | `histogram` |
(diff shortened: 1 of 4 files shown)


docs/reference/metrics.md
```diff
@@ -79,7 +79,7 @@ PostgreSQL-backed metastores also expose connection pool gauges:
 
 | Namespace | Metric Name | Description | Type |
 | --------- | ----------- | ----------- | ---- |
-| `quickwit_search` | `leaf_searches_splits_total` | Number of leaf searches (count of splits) started | `counter` |
+| `quickwit_search` | `split_search_outcome` | Number of local leaf split search outcomes by `category` (`success`, operational `error`, cache/pruning, or cancellation phase). Retries are counted separately | `counter` |
```
Mallets, line 82: Double checking whether this change is about updating stale docs.
I don't see any metrics renaming in this PR (which is ok).

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -2170,6 +2176,8 @@ enum SplitSearchState {
     PrunedAfterWarmup,
     CpuQueue,
     Cpu,
+    // Reserved for infrastructure and search-execution failures, not invalid user requests.
+    Error,
```
Mallets, line 2180: Should we add an additional `error` tag to better categorise the error on the emitted metric?

Mallets (approved):

docs/reference/metrics.md
PSeitz-dd: Yes, there is no `leaf_searches_splits_total` anymore

quickwit/quickwit-search/src/leaf.rs
PSeitz-dd: I added a tag for some categories

PSeitz-dd: forgot to push: https://github.com/quickwit-oss/quickwit/pull/6663

PR #6674, opened 2026-08-11
prune intermediate results at leaf

PSeitz:
prune intermediate aggregation results before sending back from leaf to root node.

Otherwise a term aggregation on a high cardinality field over many splits will exceed the grpc payload.

### DIFF
--- a/quickwit/quickwit-search/src/collector.rs
+++ b/quickwit/quickwit-search/src/collector.rs
@@ -26,7 +26,7 @@ use quickwit_proto::search::{
 use quickwit_proto::types::SplitId;
 use serde::Deserialize;
 use tantivy::aggregation::agg_req::{Aggregations, get_fast_field_names};
-use tantivy::aggregation::intermediate_agg_result::IntermediateAggregationResults;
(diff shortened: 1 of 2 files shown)


fulmicoton (approved):

quickwit/quickwit-search/src/collector.rs
```diff
@@ -900,8 +900,11 @@ fn merge_intermediate_aggregation_result<'a>(
                         }
                     },
                 )?;
-            let serialized =
-                postcard::to_allocvec(&merged_opt.unwrap_or_default()).map_err(map_error)?;
+            let mut merged = merged_opt.unwrap_or_default();
+            // Leaf results can be merged again at the root or by a federated query. Keep the
```
trinity-1686a, line 904: for root it means we increase our doc_count_error_upper_bound bound in a situation where we shouldn't

trinity-1686a (approved):

quickwit/quickwit-search/src/collector.rs
PSeitz: https://github.com/quickwit-oss/quickwit/issues/6676

PR #6675, opened 2026-08-11
remove component:shortlived metric cache counter

PSeitz:


docs/reference/metrics.md
```diff
@@ -7,7 +7,7 @@ Quickwit exposes key metrics in the [Prometheus](https://prometheus.io/) format
 
 ## Cache Metrics
 
-Currently Quickwit exposes metrics for three caches: `fastfields`, `shortlived`, `splitfooter`. These metrics share the same structure.
+Currently Quickwit exposes metrics for two caches: `fastfields` and `splitfooter`. These metrics share the same structure.
```
trinity-1686a: we expose a bit more than 2 caches (at least partial request and predicate)

trinity-1686a (approved):

docs/reference/metrics.md
PSeitz: updated the docs

PR #6656, opened 2026-08-04
build(deps): upgrade opendal from 0.57 to 0.58

Xuanwo:
### Description

Upgrade Apache OpenDAL from 0.57 to 0.58.1 so Quickwit stays on the latest storage layer release (including GCS path encoding and HTTP transport fixes).

OpenDAL 0.58 has a few API changes that this PR adapts to:

- `Operator::new(...)` now returns `Result<Operator>` directly; the old `.finish()` step is gone.
- Custom HTTP clients no longer use `HttpClientLayer` / `raw::HttpClient`. The HTTPS TLS regression test now injects a reqwest client through `OperationContext` + `HttpTransporter` / `ReqwestTransport`.

### How was this PR tested?

- `cargo check -p quickwit-storage --features gcs`
- `cargo test -p quickwit-storage --features gcs --lib` (71 tests, including `test_gcs_storage_get_slice_over_https_with_verified_tls`)
- `dd-rust-license-tool --config quickwit/license-tool.toml --manifest-path quickwit/Cargo.toml check`

PR #6678, opened 2026-08-11
fix(storage): initialize OpenDAL's default HTTP transport

### LINKED ISSUE #6677: GCS requests fail after OpenDAL 0.58 upgrade
After #6656 upgraded OpenDAL to 0.58, GCS requests fail with `ConfigInvalid: default HTTP transport is not installed` because
Quickwit disables OpenDAL's default features. OpenDAL's default HTTP transport must be initialized explicitly.

dayaffe:
### Description

OpenDAL 0.58 moved HTTP handling behind a process-wide transport. Quickwit disables OpenDAL's default features, including `auto-register-services`, so the migration in #6656 left normal GCS operators with a lazy default transport but no installed implementation. The first GCS request therefore failed with `ConfigInvalid: default HTTP transport is not installed`.

This PR:

- Calls `opendal::install_default()` at the shared GCS constructor so the compiled Reqwest transport is registered as OpenDAL's process-wide default before the first GCS request
- Replaces the deprecated `reqwest-rustls-tls` alias with the canonical `http-transport-reqwest` feature. Both select the same Reqwest/Rustls transport in OpenDAL 0.58 but `reqwest-rustls-tls` will be deleted in 0.59

### How was this PR tested?

This issue was discovered while following the local development instructions in `CONTRIBUTING.md` and running `make test-all`. Before the fix, the GCS integration test consistently failed with `ConfigInvalid: default HTTP transport is not installed`.

After the fix:

- The GCS integration test passed both in isolation and as part of the
full suite.
- `make test-all` passed 3,030 tests with 17 skipped; the 10 failpoint
tests also passed.
- The Clippy, nightly rustfmt, and license-header checks passed.

Closes #6677

### DIFF
--- a/quickwit/Cargo.toml
+++ b/quickwit/Cargo.toml
@@ -384,7 +384,7 @@ azure_storage_blobs = { version = "0.21", default-features = false, features = [
   "enable_reqwest_rustls",
 ] }
 
-opendal = { version = "0.58", default-features = false, features = ["reqwest-rustls-tls"] }
+opendal = { version = "0.58", default-features = false, features = ["http-transport-reqwest"] }
 reqsign = { version = "0.18", default-features = false, features = ["google", "default-context"] }
 
 quickwit-actors = { path = "quickwit-actors" }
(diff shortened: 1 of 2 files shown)


guilload (approved):

guilload:
@dayaffe, we gotta fix CI so PRs such as https://github.com/quickwit-oss/quickwit/pull/6656 can't be merged anymore. The "coverage" workflow runs on main on merge but:
- it is currently broken
- when it wasn't broken, nobody cared and acted on the job failing.

Can you look into a solution?

PR #6680, opened 2026-08-11
Remove departed ingesters from routing table

nadav-govari:
### Description

Ingesters that leave the cluster aren't removed from the routing table. We were relying on the ingester pool to not select them. If these ingesters later rejoin the cluster, the old entry is still there- representing stale data. It's only cleared when the control plane inits shards on those ingesters, or they broadcast their first capacity update (both can take seconds). This clears the entry on departure to prevent that edge case.

### How was this PR tested?

Unit tests.

### DIFF
--- a/quickwit/quickwit-indexing/src/actors/indexing_service.rs
+++ b/quickwit/quickwit-indexing/src/actors/indexing_service.rs
@@ -777,7 +777,6 @@ impl IndexingService {
         Ok(Some(merge_planner_mailbox))
     }
 
-    //noinspection ALL
     /// For all Ingest V2 pipelines, assigns the set of shards they should be working on.
     /// This is done regardless of whether there has been a change in their shard list
     /// or not.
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-indexing/src/actors/indexing_service.rs
```diff
@@ -777,6 +777,7 @@ impl IndexingService {
         Ok(Some(merge_planner_mailbox))
     }
 
+    //noinspection ALL
```
guilload: :)

quickwit/quickwit-ingest/src/ingest_v2/mod.rs
```diff
@@ -96,6 +97,14 @@ impl IngesterPoolEntry {
 
 pub type IngesterPool = Pool<NodeId, IngesterPoolEntry>;
 
+/// Published when an ingester leaves the cluster.
+#[derive(Debug, Clone)]
+pub struct IngesterDeparture {
```
guilload:
```suggestion
pub struct IngesterLeft {
```

guilload (approved):

nadav-govari:
On hold as we discuss a more thorough, more complicated fix

nadav-govari:
Superseded by https://github.com/quickwit-oss/quickwit/pull/6680

PR #6681, opened 2026-08-12
Log generation ID consistently as string

guilload:
### Description
When logged this way, `generation_id = chitchat_id.generation_id`, log sites serialize as u64, and those nanosecond timestamps exceed JSON’s exact integer limit of `2^53 - 1`... Other log sites use `Display` so we end up with confusing logs, for instance `1786461916535218400` vs `1786461916535218310`.


nadav-govari (approved):

guilload:
Superseded by #6682.

PR #6683, opened 2026-08-12
Faster byte range cache

PSeitz:
Optimize byte range cache lookups.

In regex queries we have a lot of `FileHandle::read_bytes` lookups, which are passed to `NeedMutByteRangeCache::get_slice`. The path comparison was very expensive in these cases, although a lookup will never cross path boundaries.

This PR removes the repeated path lookups by moving it to one layer above to the `FileHandle`

Screenshot 2026-08-12 at 20 47 40


quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -12,270 +12,219 @@
...
-            return None;
-        };
-
-        let start = byte_range.start - k.range_start;
-        let end = byte_range.end - k.range_start;
-        Some(v.bytes.slice(start..end))
+        if Self::get_block(&self.blocks, byte_range.start, byte_range.end).is_none() {
+            Self::merge_ranges(&mut self.blocks, byte_range.start, byte_range.end)?;
+        }
+        let (block_start, value) = Self::get_block(&self.blocks, byte_range.start, byte_range.end)?;
```
trinity-1686a: if get_block was Some the first time, we call it again immediately rather than keeping the result ?!

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
```diff
@@ -12,270 +12,219 @@
...
         self.num_bytes += (final_range.end - final_range.start) as u64;
     }
 
-    // Return a block that contain everything between query.range_start and range_end
-    fn get_block<'a>(
-        &self,
-        query: &CacheKey<'a, T>,
+    /// Returns a block containing everything between `range_start` and `range_end`.
+    fn get_block(
+        blocks: &BTreeMap<usize, CacheValue>,
```
trinity-1686a: why does this not take &self? borrow checker issues?

trinity-1686a (approved):

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
PSeitz-dd: changed it to a single lookup

quickwit/quickwit-storage/src/cache/byte_range_cache.rs
PSeitz-dd: no reason really, changed it to self

not relevant here, but fun fact: passing function parameters proves to the compiler these function parameter don't alias and produces better assembly

trinity-1686a: so passing 3-4 args instead of a single self that contains those 3-4 args can generate better assembly because there are more NoAliases in the llvm IR?
that's... surprising

PSeitz-dd: Yes, more info here: https://github.com/rust-lang/rust/issues/149670
I encountered it with a 30% regression on term aggregations

They could prove they don't alias, but it's too hard apparently

PR #6693, opened 2026-08-14
Migrate Azure blob storage to azure_storage_blob 1.0.0

### LINKED ISSUE #6672: Azure: indexer fails with storage error(kind=Unauthorized) exactly 24h after startup when using Workload Identity
**Describe the bug**

When Quickwit authenticates to Azure Blob Storage via Azure Workload Identity (no `access_key` configured), the indexer loses access to object storage **exactly 24 hours after process start**, and never recovers on its own.

```
WARN uploader: quickwit_indexing::actors::uploader: Failed to upload split. Killing!
     cause=failed uploading key 01K….split in bucket azure://<container>/quickwit-indexes/<index>
    0: storage error(kind=Unauthorized, source=Azure error wrapper(inner=non-io error occurred which will not be retried))
```

Every split upload fails from that moment on. Because the error is classified non-retryable, the indexing pipeline is killed and never recovers — the only remedy is restarting the pod, after which it works for exactly another 24 hours.

The root cause is that `azure_identity` 0.21 reads `AZURE_FEDERATED_TOKEN_FILE` **once, at credential construction**, and reuses that assertion for the lifetime of the credential.

**Steps to reproduce**

1. Run Quickwit on AKS with Azure Workload Identity enabled (`azure.workload.identity/use: "true"`, service account annotated with a client ID, webhook injecting `AZURE_CLIENT_ID` / `AZURE_TENANT_ID` / `AZURE_FEDERATED_TOKEN_FILE`).
2. Configure Azure storage **without** an access key, so the token-credential path is taken:
   ```yaml
   default_index_root_uri: azure://<container>/quickwit-indexes
   storage:
     azure:
       account: <account>
   ```
3. Run an indexer with a continuously ingesting source and wait 24 hours.

**Expected behavior**

The credential is refreshed transparently and indexing continues indefinitely, as it does with S3 / IRSA.

**Actual behavior**

At T+24h every object storage write fails with `kind=Unauthorized` until the process is restarted.

**Configuration**

1. `quickwit --version`:
   ```
   quickwit version: 0.9.0 (x86_64-unknown-linux-gnu 2026-07-28T09:58:34Z 4b2775e)
   ```
   Image: `quickwit/quickwit:edge@sha256:687cea395588fb328a413b3ef5ef58889597a164eae43f4006f69e89d505ab6a`

2. Relevant `node.yaml`:
   ```yaml
   default_index_root_uri: azure://<container>/quickwit-indexes
   storage:
     azure:
       account: <account>   # no access_key -> token credential path
   ```
   Loaded config confirms `access_key: None`:
   ```
   storage_configs: StorageConfigs([Azure(AzureStorageConfig { account_name: Some("<account>"), access_key: None })])
   ```

---

**Root cause analysis**

`AzureBlobStorage` delegates credential selection to `azure_identity` when no access key is set:

https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-storage/src/object_storage/azure_blob_storage.rs#L178-L184

With the three workload-identity env vars present, the resolution chain is
`create_credential()` → `DefaultAzureCredential` → `EnvironmentCredential` → **`WorkloadIdentityCredential`**.

In `azure_identity` 0.21.0, that credential reads the token file exactly once, inside `create()`, and stores it:

```rust
// azure_identity-0.21.0/src/token_credentials/workload_identity_credentials.rs
let token = std::fs::read_to_string(token_file.clone())?;   // read ONCE
return Ok(WorkloadIdentityCredential::new(..., token));
...
async fn get_token(&self, scopes: &[&str]) -> Result<AccessToken> {
    federated_credentials_flow::perform(..., self.token.secret(), ...)  // frozen copy, forever
}
```

Two lifetimes then collide. Both measured on a live cluster:

| Token | Lifetime | Measurement |
|---|---|---|
| Projected service-account token (the client assertion) | **1 hour** | `kubectl create token --audience api://AzureADTokenExchange` → `exp - iat = 3600`; pod spec shows `expirationSeconds: 3600` |
| Entra ID access token for `https://storage.azure.com/.default` | **24 hours** | live token exchange → `expires_in: 86399` |

Sequence:

1. **T+0** — the token file is read once, exchanged, and a 24h access token is cached in `TokenCache`.
2. **T+0 → T+24h** — kubelet rotates the token file hourly, but Quickwit never re-reads it. No refresh is *attempted*, so the growing staleness stays invisible and everything works normally.
3. **T+24h** — `TokenCache::is_expired()` fires and the exchange is retried with an assertion that expired 23 hours earlier. Entra rejects it (`AADSTS700024: Client assertion is not within its valid time range`).
4. `DefaultAzureCredential` falls through to App Service / IMDS / Azure CLI credentials, none of which are available in the pod, so the whole chain fails and surfaces as `kind=Unauthorized`.
5. No recovery, because the frozen assertion lives in a credential owned by a long-lived pipeline. Only a process restart re-reads the file.

Observed timing matches to the second — pod started `2026-08-08T09:12:10Z`, first `Unauthorized` at `2026-08-09T09:12:19Z`: **24h 00m 09s**. Across a week of restarts the interval was consistently 24.2–25.4h (the drift is human reaction time).

**Only the indexer is affected.** `StorageResolver::resolve()` constructs a fresh `Storage` — and therefore a fresh credential and a fresh file read — on every call. The janitor and searcher re-resolve per operation and stayed healthy for 67h+ uptime on the same service account; the indexer resolves once when the indexing pipeline spawns and holds that credential for the process lifetime.

**Upstream status**

This is a known, fixed bug in the Azure SDK:

- Issue: https://github.com/Azure/azure-sdk-for-rust/issues/1739 — "Azure Workload Identity - Expired token", same `AADSTS700024` signature
- Fix: https://github.com/Azure/azure-sdk-for-rust/pull/1997 — "Periodically read workload identity token from file" (re-reads every 10 minutes)

The fix shipped in `azure_identity` 0.22.0.

**Why simply bumping the dependency does not work**

`azure_storage` and `azure_storage_blobs` were never published beyond **0.21.0** — the legacy Rust SDK storage crates are end-of-life. `azure_storage` 0.21 requires `azure_core ^0.21`, and `StorageCredentials::token_credential()` takes an `Arc<dyn TokenCredential>` from azure_core 0.21, whereas `azure_identity` ≥ 0.22 implements azure_core ≥ 0.22's trait — a different, incompatible trait. The new-SDK replacement `azure_storage_blob` is currently `1.1.0-beta.1`.

So the fix cannot arrive via https://github.com/quickwit-oss/quickwit/blob/main/quickwit/Cargo.toml#L378-L385 without migrating off the storage crates.

**Proposed fix**

A small, self-contained wrapper inside `quickwit-storage` that implements azure_core 0.21's `TokenCredential` and re-reads `AZURE_FEDERATED_TOKEN_FILE` (mirroring what upstream #1997 does), used in place of `azure_identity::create_credential()` at `azure_blob_storage.rs:182`. No dependency changes, no API break, and it can be scoped to the workload-identity case with a fallback to the existing behaviour otherwise.

siva-abstract-security:
migrates azure blob storage to the rewritten sdk - azure_core 1.1, azure_identity 1.0, azure_storage_blob 1.0. `azure_storage` is gone entirely and nothing replaces it. closes #6672.

the upgrade fixes the bug on its own - `azure_identity` 1.0 re-reads the federated token file once the cached copy is >600s old, so the assertion at T+24h is fresh and the indexer stops dying. no workaround needed.

two things 1.0 dropped that i had to rebuild:

1 - shared key signing. 1.0 does entra tokens only and MS says it's not coming back (Azure/azure-sdk-for-rust#2975). we document `access_key` and azurite speaks shared key only, so `azure_shared_key` does what `azure_storage` 0.21 used to.

2 - credential selection. `create_credential()` and `DefaultAzureCredential` are both gone, so `azure_credentials` picks workload vs managed identity from the env explicitly.

one thing worth your attention - the generated ops disagree about `Content-Length`. `stage_block` sets the header, `commit_block_list` leaves it to the transport, so signing covered an empty length while the wire carried a real one and shared key rejected it as `AuthorizationFailure` with nothing pointing at why. only azurite caught this - unit tests over the string construction didn't.

azurite also needs `--skipApiVersionCheck` now. 3.24.0 and even 3.36.0 (newest released) both predate the api version 1.0 sends. docker-compose carries that plus the image bump.

green: 80 unit tests, the full azurite integration suite, clippy, and `quickwit-cli` under `release-feature-set`.

what i don't know:

1 - haven't run this against a real azure account, only azurite. so managed identity and workload identity never actually executed, including the 24h refresh this closes.

2 - single part upload sets `blob_content_md5` instead of a transactional checksum, since the partitioned upload path doesn't expose one - stored with the blob rather than checked per request. multipart still checks per block via `stage_block`.

3 - whether `--skipApiVersionCheck` is hiding a real incompatibility. azurite took every op the suite runs, but that flag stops it telling us what it doesn't implement.

written with claude opus 5.


quickwit/quickwit-storage/src/object_storage/azure_blob_storage.rs
```diff
@@ -142,11 +155,16 @@ impl AzureBlobStorage {
     }
 
     /// Creates an emulated storage for testing.
+    ///
+    /// The 1.0 SDK has no `emulator()` helper, and could not offer one: the emulator
+    /// authenticates with a shared key, which the SDK no longer signs. The well-known
+    /// account and key are documented at
+    /// <https://learn.microsoft.com/azure/storage/common/storage-use-azurite>.
```
guilload:
```suggestion
    /// The well-known account and key are documented at:
    /// <https://learn.microsoft.com/azure/storage/common/storage-use-azurite>.
```

quickwit/quickwit-storage/src/object_storage/azure_blob_storage.rs
```diff
@@ -256,12 +328,38 @@ impl AzureBlobStorage {
         let _timer = HistogramTimer::new(&crate::metrics::OBJECT_STORAGE_PUT_OBJECT_DURATION);
         retry(&self.retry_params, || async {
             let data = Bytes::from(payload.read_all().await?.to_vec());
-            let hash = azure_storage_blobs::prelude::Hash::from(md5::compute(&data[..]).0);
-            self.container_client
-                .blob_client(name)
-                .put_block_blob(data)
-                .hash(hash)
-                .into_future()
+            let digest = md5::compute(&data[..]);
```
guilload: Any chance we can do CRC64-NVME?

quickwit/quickwit-storage/src/object_storage/azure_credentials.rs
```diff
@@ -0,0 +1,465 @@
...
+    }
+}
+
+/// Decides which credential the environment describes.
+///
+/// Precedence follows the chain `azure_identity::create_credential()` used to walk, where an
+/// environment credential came before managed identity. Dropping that ordering silently
+/// breaks every deployment that authenticates with a service principal, because the client
+/// secret is ignored and IMDS is contacted instead.
+fn select_token_credential_kind(var: impl Fn(&str) -> Option<String>) -> TokenCredentialKind {
```
guilload:
```suggestion
fn select_token_credential_kind(var_fn: impl Fn(&str) -> Option<String>) -> TokenCredentialKind {
```

quickwit/quickwit-storage/src/object_storage/azure_credentials.rs
```diff
@@ -0,0 +1,465 @@
...
+    let non_empty = |name: &str| match var(name) {
+        Some(value) if !value.trim().is_empty() => Some(value),
```
guilload: Don't we want to use the trimmed value?

quickwit/quickwit-storage/src/object_storage/azure_credentials.rs
```diff
@@ -0,0 +1,219 @@
...
+        // An empty or unrecognized value falls through to detection. The workload identity
+        // variables are injected by a mutating webhook rather than by an operator, so their
+        // presence is the signal that the pod runs under workload identity.
+        _ => {
+            if workload_identity_env_is_complete() {
+                info!("using azure workload identity credential");
+                return Ok(WorkloadIdentityCredential::new(None)?);
+            }
+            info!("using azure managed identity credential");
+            Ok(ManagedIdentityCredential::new(None)?)
```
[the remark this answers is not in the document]
dayaffe: Addressed in abb1429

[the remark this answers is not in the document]
dayaffe: Addressed in abb1429

dayaffe (commented):
Thanks for this contribution! 

I tested it against a real Azure account with 800k documents using both access-key and service-principal auth, and everything worked well.

I also confirmed the fix for #6672 directly. I ran one credential for the process lifetime, sped up the clock so the 24h token expiry hit in minutes, and verified that the refresh re-read the token file and carried on. The same setup on the old SDK fails with `AADSTS700024`, matching the issue.

I found one problem: `test_exists` fails against real Azure because `Put Block` rejects empty blocks with `400 InvalidHeaderValue`. Azurite accepts them, so the suite still passes locally. This fixed it for me in `put_single_part`:
```rust
if data.is_empty() {
    let block_list_content = RequestContent::try_from(BlockLookupList::default())
        .map_err(AzureErrorWrapper::from)?;
    block_blob_client
        .commit_block_list(block_list_content, None)
        .await?;
    return Result::<(), AzureErrorWrapper>::Ok(());
}
```
Everything else looked good from my testing.

quickwit/quickwit-storage/src/object_storage/azure_blob_storage.rs
```diff
@@ -256,12 +328,38 @@ impl AzureBlobStorage {
...
+            let stage_block_options = BlockBlobClientStageBlockOptions {
+                transactional_content_md5: Some(digest.0.to_vec()),
+                ..Default::default()
+            };
+            // `RequestContent::from` is an inherent function over `Vec<u8>`, which shadows
+            // the `From<Bytes>` impl, so go through `Into` to keep the `Bytes` as is.
+            let content: RequestContent<Bytes, NoFormat> = data.into();
+            block_blob_client
+                .stage_block(
+                    SINGLE_PART_BLOCK_ID,
```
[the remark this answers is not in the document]
dayaffe: This is likely not an issue for quickwit since split names are unique and retries are sequential but it's a one line fix that ensures the last writer wins so I think its worth adopting anyway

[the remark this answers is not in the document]
siva-abstract-security: fixed in cfee6c7 - block ids are `{upload ULID}:{part:05}` now, minted before the retry loop, and multipart got the same treatment since it had the same shape and would have ended up a different id length otherwise.

siva-abstract-security:
fixed in 288d8b9 with your snippet - an empty payload commits an empty block list instead of staging a zero length block, and cfee6c7 covers the codex block id race.

quickwit/quickwit-storage/src/object_storage/azure_blob_storage.rs
siva-abstract-security: taken as is in b8aa579

quickwit/quickwit-storage/src/object_storage/azure_credentials.rs
siva-abstract-security: done in 29cce17

quickwit/quickwit-storage/src/object_storage/azure_credentials.rs
siva-abstract-security: yeah good catch - we checked `trim().is_empty()` but stored the untrimmed value, so padding would go to Entra as part of the tenant/client id. fixed in 29cce17 plus a test. also unified the explicit `managedidentity` branch which wasn't trimming at all.

quickwit/quickwit-storage/src/object_storage/azure_blob_storage.rs
siva-abstract-security: yep - done in 008f624, both `stage_block` sites. no new crate either, `crc-fast` was already in the tree via `aws-smithy-checksums`. shared key signing needed nothing since `canonicalized_headers` already picks up `x-ms-*` generically.

two things worth flagging - the 1.0 sdk's own block blob tests pin the right vector (`V0JSBnCFdzM=` for `hello`) but label it ECMA-182, which is a different polynomial. it's actually CRC-64/NVME little endian, so i pinned that in a unit test.

and azurite doesn't validate `x-ms-content-crc64` at all - i swapped the checksum for 8 zero bytes and the suite still passed, where the old md5 path got rejected with `Provided contentMD5 doesn't match`. so the integration suite no longer covers this end to end, only the unit vector does. still on the "never run against a real azure account" pile.

PR #6694, opened 2026-08-14
AZ and decommissioning-aware indexing planning, plus optimizations

nadav-govari:
### Description

This change adds support for availability zone aware indexing plans, and for allowing decommissioning ingesters to index their own shards to help speed up decommissioning.

The change sits behind a flag, default off, that falls back to the existing flow if it's off. The existing unit tests were not touched (other than struct stuff for compilation). New tests were added, including static tests at huge scale, to ensure that the plans produced are reasonable.

### How was this PR tested?

Extensive unit testing and cluster testing. At one index, it achieves just about 100% same-az locality. At multiple indexes (I tested up to 9), it generally hovers above 90%, usually above 95%. Density is usually in the high 80s. There's additional optimizations coming in a follow-up PR that get us there.

### High level summary of changes

First, we strip foreign shards off retiring indexers, which are only eligible to index their own shards. This part is relatively straightforward. 

To assign shards back to retiring indexers, we reclaim foreign hosted shards from peers. This is done with a general accounting of self-hosted vs foreign shards; from there, we're able to figure out what should go where in the physical step.

Finally, to ensure locality and density, we perform the original density assignment step in each AZ; then, we assign the leftover shards the same as we did before.


quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -320,48 +411,97 @@ impl IndexingScheduler {
...
-            .iter()
-            .filter_map(|indexer| {
-                if indexer.indexing_capacity.cpu_millis() > 0 {
-                    Some((indexer.node_id.to_string(), indexer.indexing_capacity))
-                } else {
-                    None
-                }
-            })
-            .collect();
+        let locality_aware = enable_az_aware_scheduling();
```
fmassot, line 414: sounds like we are enable az aware scheduling but we just return a boolean
then what`locality_aware` means?

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
```diff
@@ -453,6 +548,87 @@ fn convert_scheduling_solution_to_physical_plan(
...
+fn shard_availability_zone<'a>(
+    shard_id: &ShardId,
+    shard_locations: &ShardLocations,
+    indexer_infos: &'a FnvHashMap<String, IndexerInfo>,
+) -> Option<&'a str> {
+    let hosting_node_id = shard_locations.get_shard_locations(shard_id).first()?;
+    indexer_availability_zone(hosting_node_id.as_str(), indexer_infos)
+}
+
+pub(crate) fn is_shard_nearby(
```
fmassot, line 606: `is_shard_in_indexer_az`?

quickwit/quickwit-control-plane/src/metrics.rs
```diff
@@ -16,13 +16,29 @@ use quickwit_metrics::{LabelNames, LazyCounter, LazyGauge, label_names, lazy_cou
 
 #[derive(Debug, Clone, Copy)]
 pub struct ShardLocalityMetrics {
```
fmassot, line 18: those definitions are a bit confusing

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -290,6 +313,74 @@ fn get_sources_to_schedule(
     }
     sources
 }
+/// In the case where there are only draining indexers left, they are eligible to index any shards,
+/// including shards hosted on other indexers. Otherwise, they can only index their own shards.
+fn determine_draining_indexer_eligibility(indexers: &[IndexerNodeInfo]) -> Eligibility {
```
fulmicoton-dd, line 318: I don't understand the logic here. Can you explain why we need this? What is the thing we are trying to avoid?

Assuming all indexer are decomissioning why is this. not ok for them to only be able to consume their own shards?

quickwit/quickwit-control-plane/src/metrics.rs
```diff
@@ -47,13 +63,16 @@ pub(crate) const INDEX_ID_LABEL_NAMES: LabelNames<1> = label_names!("index_id");
 
 static INDEXED_SHARDS: LazyGauge = lazy_gauge!(
         name: "indexed_shards",
-        description: "Number of (remote/local) shards in the indexing plan",
+        description: "Number of (remote/nearby/local) shards in the indexing plan",
```
fulmicoton-dd, line 66: nitpick: i like explicit stuff for that kind of thing.

nearby -> same_az

fulmicoton-dd: yeah well ignore that

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -290,6 +313,74 @@ fn get_sources_to_schedule(
...
+        eligibility,
+    }
+}
+
+fn build_indexer_infos(
+    indexers: &[IndexerNodeInfo],
+    locality_aware: bool,
+) -> FnvHashMap<String, IndexerInfo> {
+    let draining_eligibility = determine_draining_indexer_eligibility(indexers);
+    let mut indexer_infos: FnvHashMap<String, IndexerInfo> = FnvHashMap::default();
```
fulmicoton-dd, line 360: Wouldn't keeping NodeId be strictly better here?
Clone is Free. The type says what it is,etc.
```suggestion
    let mut indexer_infos: FnvHashMap<NodeId, IndexerInfo> = FnvHashMap::default();
```

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -290,6 +313,74 @@ fn get_sources_to_schedule(
...
+        .iter()
+        .any(|indexer| indexer.ingester_status == IngesterStatus::Ready);
+    if has_ready_indexer {
+        return Eligibility::SelfHostedOnly;
+    }
+    warn!("no ready indexer available, letting draining indexers index any shard");
+    Eligibility::Any
+}
+
+fn build_indexer_info(
```
fulmicoton-dd, line 329: I am not fond of the confusion induced by the existience of

IndexerNodeInfo (preexisting) and IndexerInfo (introduced).
One is just a shorter version of the other + some precomputed eligibility thing.

Can we think of a better name? Or should we just have IndexerNodeInfo and an eligibility accessor to it?

There is already a IndexerLocality object that is an even shorter version used in the scheduling problem

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
fulmicoton-dd: +1 let's keep to the same vocabulary.

I prefer az_ware everywhere personally

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -320,48 +411,97 @@ impl IndexingScheduler {
 
         let indexers: Vec<IndexerNodeInfo> = self.select_available_indexers_for_scheduling();
 
-        let indexer_id_to_cpu_capacities: FnvHashMap<String, CpuCapacity> = indexers
```
fulmicoton-dd, line 323: ah ok so i am the culprit for using String instead of NodeId :)

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -290,6 +313,74 @@ fn get_sources_to_schedule(
...
+        availability_zone: indexer.availability_zone.clone(),
+        eligibility,
+    }
+}
+
+fn build_indexer_infos(
+    indexers: &[IndexerNodeInfo],
+    locality_aware: bool,
+) -> FnvHashMap<String, IndexerInfo> {
+    let draining_eligibility = determine_draining_indexer_eligibility(indexers);
```
fulmicoton-dd, line 359: nitpick
```suggestion
    let draining_eligibility: Eligibility = determine_draining_indexer_eligibility(indexers);
```

People (me?) would expect Vec<Eligibility>. Explicit types are sometime useful for the compiler, but really it helps humans too. (singular form helps too of course)

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -290,6 +313,74 @@ fn get_sources_to_schedule(
...
+    IndexerInfo {
+        cpu_capacity: indexer.indexing_capacity,
+        availability_zone: indexer.availability_zone.clone(),
+        eligibility,
+    }
+}
+
+fn build_indexer_infos(
+    indexers: &[IndexerNodeInfo],
+    locality_aware: bool,
```
fulmicoton-dd, line 357: ah so locality_aware is az_awareness + decomissioning indexer does not index elsewhere?

I understand the benefit of the shortcut and the fact that you might want different words for it, but you called the flag az aware.
Also it should be documented somewhere . (on enable_az_aware_scheduling() maybe)?

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -320,48 +411,97 @@ impl IndexingScheduler {
...
+        if indexer_infos.is_empty() {
             if !sources.is_empty() {
                 warn!("no indexing capacity available, cannot schedule an indexing plan");
             }
             return;
         };
 
         let shard_locations = model.shard_locations();
-        let new_physical_plan = build_physical_indexing_plan(
+        let plan_from_previous = build_physical_indexing_plan(
```
fulmicoton-dd, line 427: nitpick

plan_from_previous confused me. I thought it meant the physical plan built from the previous solution.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -320,48 +411,97 @@ impl IndexingScheduler {
...
+            &indexer_infos,
+            locality_aware,
             self.state.last_applied_physical_plan.as_ref(),
             &shard_locations,
         );
-        let shard_locality_metrics =
-            get_shard_locality_metrics(&new_physical_plan, &shard_locations);
+        let locality_metrics_from_previous =
+            get_shard_locality_metrics(&plan_from_previous, &shard_locations, &indexer_infos);
+        let (new_physical_plan, shard_locality_metrics) = self.maybe_build_plan_from_scratch(
```
fulmicoton-dd, line 436: here you called it new_physical_plan.

I'd prefer if we had symmetric names.

e.g.
new_physical_plan_incremental
new_physical_plan_from_scratch

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -290,6 +313,74 @@ fn get_sources_to_schedule(
...
+        if indexer.indexing_capacity.cpu_millis() == 0 {
+            continue;
+        }
+        let indexer_info = build_indexer_info(indexer, draining_eligibility, locality_aware);
+        indexer_infos.insert(indexer.node_id.to_string(), indexer_info);
+    }
+    indexer_infos
+}
+
+fn build_indexer_statuses(indexers: &[IndexerNodeInfo]) -> FnvHashMap<String, IngesterStatus> {
```
fulmicoton-dd, line 371: Could have been NodeId. No string copy. Semantic heaven. Less koalas dying.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -290,6 +313,74 @@ fn get_sources_to_schedule(
...
+    indexers
+        .iter()
+        .map(|indexer| (indexer.node_id.to_string(), indexer.ingester_status))
+        .collect()
+}
+
+fn build_indexer_tasks(indexers: &[IndexerNodeInfo]) -> FnvHashMap<String, Vec<IndexingTask>> {
```
fulmicoton-dd, line 378: same as above.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -320,48 +411,97 @@ impl IndexingScheduler {
...
+        );
+        let locality_metrics_from_scratch =
+            get_shard_locality_metrics(&plan_from_scratch, shard_locations, indexer_infos);
+        if locality_metrics_from_scratch.locality_percent() <= locality_metrics.locality_percent() {
+            return (plan, locality_metrics);
+        }
+        info!(
+            locality_percent = locality_metrics.locality_percent(),
+            locality_percent_from_scratch = locality_metrics_from_scratch.locality_percent(),
+            "rebuilding the indexing plan from scratch to restore shard locality"
```
fulmicoton-dd, line 499:
```suggestion
            "rebuilt the indexing plan from scratch to restore shard locality"
```

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -320,48 +411,97 @@ impl IndexingScheduler {
...
+            sources,
+            indexer_infos,
+            locality_aware,
+            None,
+            shard_locations,
+        );
+        let locality_metrics_from_scratch =
+            get_shard_locality_metrics(&plan_from_scratch, shard_locations, indexer_infos);
+        if locality_metrics_from_scratch.locality_percent() <= locality_metrics.locality_percent() {
+            return (plan, locality_metrics);
```
fulmicoton-dd, line 494: maybe we should log a warn?

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -320,48 +411,97 @@ impl IndexingScheduler {
...
                 return;
             }
         }
+        self.state.last_applied_indexer_statuses = indexer_statuses;
         self.apply_physical_indexing_plan(new_physical_plan, Some(notify_on_drop));
         self.state.num_schedule_indexing_plan += 1;
     }
 
+    /// A plan built from the previous one can lose locality over time but never regain it. Below
+    /// the threshold we try one built from scratch (equivalent to restarting the control plane).
```
fulmicoton-dd, line 465: It might be worth commenting about the condition.

The cool down
The previous plan being below a threshold.
The new plan being an improvement.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -320,48 +411,97 @@ impl IndexingScheduler {
...
             &sources,
-            &indexer_id_to_cpu_capacities,
+            &indexer_infos,
+            locality_aware,
             self.state.last_applied_physical_plan.as_ref(),
             &shard_locations,
         );
-        let shard_locality_metrics =
-            get_shard_locality_metrics(&new_physical_plan, &shard_locations);
+        let locality_metrics_from_previous =
```
fulmicoton-dd, line 434: I don't think it is useful to compute it outside of maybe_build_plan_from_scratch and pass it as an arg.

fulmicoton-dd:
I don't see in the PR description the fact that now you can rebuild from scratch periodically.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -50,52 +50,34 @@ pub fn solve(
...
-
-    if best_attempt > 0 {
+    let inflation_attempt = minimal_feasible_inflation_attempt(&base_problem);
+    if inflation_attempt > 0 {
         // the higher the attempt number, the more unbalanced the solution
         tracing::warn!(
-            attempt_number = best_attempt,
+            attempt_number = inflation_attempt,
             "capacity re-scaled, scheduling solution likely unbalanced"
         );
```
fulmicoton-dd, line 59: I am not fond of the way this is cut.

If you want to isolate code in a function, it is important for its contract to be clear.
For instance, here, returning the number of inflation is terrible, because it means the definition of the inflation level is part of the contract.

Instead I recommend having a function that represent a step maybe?

let scaled_problem = inflate_capacities_if_necessary(base_problem); 

and do the logging in there.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -281,6 +273,69 @@ fn assert_enforce_nodes_cpu_capacity_post_condition(
...
+    solution: &mut SchedulingSolution,
+) {
+    for indexer_assignment in &mut solution.indexer_assignments {
+        let indexer_ord = indexer_assignment.indexer_ord;
+        if problem.is_eligible_for_foreign_shards(indexer_ord) {
+            continue;
+        }
+        let mut num_foreign_shards_per_source: Vec<(SourceOrd, u32)> = Vec::new();
+        for (&source_ord, &num_shards) in &indexer_assignment.num_shards_per_source {
+            let num_self_hosted_shards = problem.source_affinity(source_ord, indexer_ord);
```
fulmicoton-dd, line 289: should we use change num_foreign_shards_on_indexer to have it take SourceOrd instead of source and use it in strip_self_hosted_only_indexers too

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -281,6 +273,69 @@ fn assert_enforce_nodes_cpu_capacity_post_condition(
...
+    source: &Source,
+    num_shards_to_release: u32,
+    problem: &SchedulingProblem,
+    solution: &mut SchedulingSolution,
+) -> u32 {
+    let mut peer_ords: Vec<IndexerOrd> = (0..problem.num_indexers())
+        .filter(|&peer_ord| problem.is_eligible_for_foreign_shards(peer_ord))
+        .collect();
+    peer_ords.sort_by_key(|&peer_ord| {
+        solution.indexer_assignments[peer_ord].indexer_available_capacity(problem)
```
fulmicoton-dd, line 323: Don't we want to inverse sort by this? We want to remove shards from the indexer with the most work first no?

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -281,6 +273,69 @@ fn assert_enforce_nodes_cpu_capacity_post_condition(
...
+}
+
+fn release_foreign_shards_from_peers(
+    source: &Source,
+    num_shards_to_release: u32,
+    problem: &SchedulingProblem,
+    solution: &mut SchedulingSolution,
+) -> u32 {
+    let mut peer_ords: Vec<IndexerOrd> = (0..problem.num_indexers())
+        .filter(|&peer_ord| problem.is_eligible_for_foreign_shards(peer_ord))
```
fulmicoton-dd, line 320: Do we care at all about this prefiltering?

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -356,6 +412,177 @@ fn place_unassigned_shards_with_affinity(
...
+    solution: &mut SchedulingSolution,
+) -> u32 {
+    let available_capacity = available_cpu_capacity(indexer_ord, problem, solution);
+    let num_placable_shards = available_capacity.cpu_millis() / source.load_per_shard;
+    let num_shards_wanted = num_self_hosted_shards.min(num_placable_shards);
+    let num_shards_short = num_shards_wanted.saturating_sub(num_unaccounted_shards);
+    if num_shards_short == 0 {
+        return 0;
+    }
+    release_foreign_shards_from_peers(source, num_shards_short, problem, solution)
```
fulmicoton-dd, line 466: I expected this call to try to release shards from indexer_ord currently indexed on other indexers.

release_foreign_shards_from_peers seem to release any foreign shards.

Is this the wanted behavior?

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -356,6 +412,177 @@ fn place_unassigned_shards_with_affinity(
...
+}
+
+fn place_self_hosted_shards(
+    problem: &SchedulingProblem,
+    solution: &mut SchedulingSolution,
+) -> Vec<LeftoverShards> {
+    let num_locality_groups = problem.num_locality_groups();
+    let mut leftover_shards_per_source: Vec<LeftoverShards> = (0..problem.num_sources())
```
fulmicoton-dd, line 474: super nitpick.
you can ignore really

vec![somethign ; n] would have worked if you made LeftoverShards Cloneable (no need for copy)

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -356,6 +412,177 @@ fn place_unassigned_shards_with_affinity(
...
+        let source_ord = unassigned_source.source_ord as usize;
+        let leftover_shards = &mut leftover_shards_per_source[source_ord];
+        let mut num_unaccounted_shards = unassigned_source.num_shards;
+        // A draining indexer claims its own shards before any other host of the same source.
+        let (draining_indexer_ords, ready_indexer_ords): (Vec<IndexerOrd>, Vec<IndexerOrd>) =
+            unassigned_source
+                .affinities
+                .keys()
+                .copied()
+                .partition(|&indexer_ord| !problem.is_eligible_for_foreign_shards(indexer_ord));
```
fulmicoton-dd, line 488: nitpick I would remove the ! and swap the result pair

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -356,6 +412,177 @@ fn place_unassigned_shards_with_affinity(
...
+        for indexer_ord in draining_indexer_ords.into_iter().chain(ready_indexer_ords) {
+            let num_self_hosted_shards = unassigned_source.affinities[&indexer_ord];
+            if !problem.is_eligible_for_foreign_shards(indexer_ord) {
```
fulmicoton-dd, line 491: that's slop isn't it? This is a draining indexer ord so we know it is not eligible.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
```diff
@@ -356,6 +412,177 @@ fn place_unassigned_shards_with_affinity(
...
+                let num_shards_reclaimed = reclaim_self_hosted_shards_from_peers(
```
fulmicoton-dd, line 492: I am surprised. As explained in reclaim_self_hosted_shards_from_peers, i don't see that function reclaiming necessarily the right shards.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
fulmicoton-dd: I think the intent is sort_by_key more than partition + chain

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: Youre absolutely right, this was redundant in keeping logic that is no longer relevant.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: No longer relevant- this code has been removed

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: You were right, it was the plan from the previous solution. I'll rework this whole section to make it more clear.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: Refactored

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: Spoke with Paul offline - we prefer to keep `locality_aware` to not leak business logic into the math code.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
nadav-govari: See above on keeping the `locality_aware` naming, but I'll try to find a more fitting name for it.

quickwit/quickwit-control-plane/src/metrics.rs
nadav-govari: I agree. There is a relic in here - in the existing code, local shards are shards that are hosted by an indexer. Ideally for us, local means in the same AZ. But I fear that changing this now would be significantly more confusing.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: Ill change all of the maps to use NodeId. Including the ordinal vecs.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: IndexerLocality in the scheduling problem fits how we think about things currently, so I think that one should stay.

I agree in that we don't want another data structure if we can avoid it. However, this data structure allows us to gate the flag check one time, when we build this data structure, because we skip the AZ and Eligibility fields if its off. It then allows us to operate without any flag knowledge in the scheduling problem, which at the end of the day is worth the new struct. (And we'd otherwise be passing the whole IndexerNodeInfo struct into the scheduling problem, and its the one that carries the indexing client on it).

So I'll rename it.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: In an effort to save the planet, the entire `indexing_scheduler` module is now NodeId throughout.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: This Original Sin has been fixed and it's all NodeId everywhere now.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: I renamed IndexerNodeInfo to IndexerPoolEntry, which matches the IngesterPoolEntry we already had. That allows us to keep IndexerInfo, which is better suited for the indexing scheduling logic.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: I renamed the flag to enable_locality_aware_scheduling, which encompasses both the AZ awareness and decommissioning awareness. I added a comment on top of the flag function.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: Applied.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: We've found that warns for normal things that go wrong in the normal lifecycle distributed systems usually worries customers more than it provides utility, so either info or error is preferable depending on the case. In this case I can emit an info.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: Addressed.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: I agree, I've fixed it.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
nadav-govari: The reason for this change is that I had Claude run the proptest thousands and thousands of times and it found a reproducible idempotency violation bug when variable shard load is enabled. It had to do with having lots of small throughput sources on the same node as much larger sources. 

When variable shard load is disabled, and every shard has a load rate of 1000mcpu, the algorithm is still verifiably correct.

Also, the implementation here (silently) returns a brand new plan in the case this bug is run into in production. This is a disaster and truthfully I just missed it. I think that I'm going to ignore the bug at the moment (it exists today already) and return later and decide whether or not it's worth addressing.

nadav-govari: I ran the churn tests that are part of this PR and the total additional churn that this bug introduced was less than a percent over hundreds of incremental plan rebuilds, and it settles (that is, it eventually re-reaches a deterministic state). So I'm of the opinion that this is worth leaving. I'll describe the bug in detail in a minute.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
nadav-govari: Sure

nadav-govari: This is much cleaner.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
nadav-govari: Yes, sort by key returns asc, and indexer available capacity returns headroom (the lower the number, the less room)

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
nadav-govari: This is a leftover from something else that is a no-op now, you're right.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
fulmicoton-dd: my comment was about the code organization, not its semantics though. 

Am I missing some subtle thing that prevents organizing the code with a inflate_capacities_if_necessary function?

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
fulmicoton-dd: (that's the one remaining important comment)

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -48,6 +50,13 @@ use crate::{IndexerNodeInfo, IndexerPool};
 
 const DEFAULT_ENABLE_VARIABLE_SHARD_LOAD: bool = false;
 
+const DEFAULT_ENABLE_AZ_AWARE_SCHEDULING: bool = false;
+
+const DEFAULT_MIN_SHARD_LOCALITY_PERCENT: u32 = 30;
```
fulmicoton-dd, line 55: We expect something much higher, no?
You plan to set it to something higher via env variable?

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -160,6 +168,20 @@ fn enable_variable_shard_load() -> bool {
     *IS_SHARD_LOAD_CP_ENABLED
 }
 
+/// Locality aware scheduling is additional constraints on top of the regular indexing scheduler.
+/// With this enabled, availability zone will be used (if provided) to preferentially index shards
+/// in the same-AZ. It also allows decommissioning indexers to index their own shards to speed up
+/// the decommissioning process.
+fn enable_locality_aware_scheduling() -> bool {
```
fulmicoton-dd: nitpick: prefer naming that makes sure this is a accessorish and not an action:
enable_locality_aware_scheduling -> is_locality_aware_scheduling_enabled or locality_aware_scheduling_enabled

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
```diff
@@ -150,6 +171,13 @@ fn convert_physical_plan_to_solution(
     }
 }
 
+#[derive(Debug)]
+pub struct IndexerInfo {
+    pub cpu_capacity: CpuCapacity,
+    pub availability_zone: Option<String>,
```
fulmicoton-dd: are we switching that to Arc<str> in the other PR?

fulmicoton-dd (approved):

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
nadav-govari: I ended up reverting to what we had before.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
nadav-govari: This is no longer exactly how it's structured, but the answer is that we need to know how many shards a decommissioning indexer has for accounting purposes (to ensure that we put the right number on it), but reconciling the actual shard IDs doesn't matter here as this is a property of the physical part of the plan.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
nadav-govari: This is now outdated.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/scheduling_logic.rs
nadav-govari: Correect, and I don't think this is still in the code.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/mod.rs
nadav-govari: I got the other pool but missed this one, Im updating it now. This is a PR stack, Im going to merge them all into this one so eventually it'll all land.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: 👍

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
nadav-govari: Removed this altogether, we just do a better job of constructing from scratch now.

PR #6695, opened 2026-08-14
schedule leaf searches by (priority, remaining_splits) lexicographically

Abdul-Andha:
### Description

- Add `priority` to `SearchRequest`
- `LeafPermitRequest` was ordered by remaining_splits. Now it is ordered by (priority, remaining_splits)
  - This lets a query with more splits and a higher priority (smaller int is higher priority) skip the line against a query with less splits and a lower priority

### How was this PR tested?

- Unit tests
- `test_search_permit_priority_precedes_remaining_splits`


quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -271,6 +271,12 @@ message SearchRequest {
   // When true, skip finalization of aggregation results and return
   // the raw IntermediateAggregationResults bytes instead.
   bool skip_aggregation_finalization = 19;
+
+  reserved 20;
+
+  // Scheduling priority for leaf search execution. Negative values are allowed,
+  // and lower values have higher priority. Callers that omit it get priority 0.
+  optional int32 priority = 21;
```
guilload:
```suggestion
  int32 priority = 21;
```
No need for optional if we want to default to 0. Unless we want to semantically differentiate `None` from `Some(0)`

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -271,6 +271,12 @@ message SearchRequest {
   // When true, skip finalization of aggregation results and return
   // the raw IntermediateAggregationResults bytes instead.
   bool skip_aggregation_finalization = 19;
+
+  reserved 20;
```
guilload, line 278: Why reserve 20?

Abdul-Andha: 20 is used in pomsky. it will be easier to sync if we don't use 20, i think

guilload: Then add a comment.

quickwit/quickwit-search/src/search_permit_provider.rs
```diff
@@ -255,6 +261,7 @@ impl LeafPermitRequest {
         task_metadata: Vec<SplitSearchTaskMetadata>,
     ) -> (Self, Vec<SearchPermitFuture>) {
         assert!(!task_metadata.is_empty(), "task_metadata must not be empty");
+        let priority = task_metadata[0].priority;
```
trinity-1686a: it's a bit strange that all tasks must have the same priority, but there's still a per-task priority field

Abdul-Andha: i agree its a bit strange. i think the alternative is introducing a new struct for per-leafsearch metadata. what do you think?
```
struct LeafSearchTaskMetadata {
    priority: i32,
    splits: Vec<SplitSearchTaskMetadata>,
}
```

trinity-1686a: i think it looks better

trinity-1686a (approved):

PR #6700, opened 2026-08-17
feat(search): pin Foyer and add split-range disk cache config

congx4:
This is the first slice of a Foyer disk cache for exact split footer and body ranges. Follow-up PRs will add the cache key codec, Foyer device, and searcher wiring.

## Summary
- Pin Foyer `0.22.3` and mixtrics `0.2.5`, declare them on `quickwit-storage`, and regenerate `LICENSE-3rdparty.csv`.
- Add validated `searcher.split_range_disk_cache` config (disabled by default). `write_policy` defaults to `write-on-eviction` when omitted.

## Test plan
- [x] `cargo check -p quickwit-storage`
- [x] `cargo test -p quickwit-config split_range_disk_cache`
- [x] `cargo clippy -p quickwit-config -p quickwit-storage --all-features --tests -- -D warnings`
- [x] `cargo +nightly fmt --all -- --check`
- [ ] CI unit tests and license check

Made with [Cursor](https://cursor.com)

PR #6701, opened 2026-08-17
feat(storage): add split range cache key codec

congx4:
## Summary
- Add a portable Foyer `Code` impl for `SplitRangeCacheKey` (`object_uri` + exact half-open byte range).
- Persist keys as little-endian `u64` length/start/end so recovered entries stay valid across architectures, without enabling Foyer's serde/bincode feature.
- Drop `foyer` from the cargo-machete ignore list now that the crate is used.

## Stack
- Built on #6700 (Foyer pin + split-range disk-cache config). Please merge #6700 first; this PR's unique commit is the cache key codec.

## Test plan
- [x] `cargo test -p quickwit-storage --lib split_range_cache_key_codec`
- [x] `cargo clippy -p quickwit-storage --all-features --tests -- -D warnings`
- [x] `cargo machete`
- [ ] CI

Made with [Cursor](https://cursor.com)

### DIFF
--- a/quickwit/quickwit-storage/src/split_range_cache/key.rs
+++ b/quickwit/quickwit-storage/src/split_range_cache/key.rs
@@ -1,0 +1,211 @@
+// Copyright 2021-Present Datadog, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#![allow(dead_code)]
+
+use std::io::{Read, Write};
+use std::ops::Range;
+
+/// Exact cache key for a split object URI and a half-open byte range.
+///
+/// The on-disk layout is portable little-endian `u64` fields so recovered
+/// entries stay valid across 32-bit and 64-bit processes:
+/// `[uri_len][uri_bytes][range_start][range_end]`.
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct SplitRangeCacheKey {
+    pub object_uri: String,
+    pub byte_range: Range<usize>,
+}
+
+impl SplitRangeCacheKey {
+    pub(crate) fn new(object_uri: String, byte_range: Range<usize>) -> Self {
+        debug_assert!(
+            byte_range.start <= byte_range.end,
+            "split range cache keys use half-open ranges"
+        );
+        Self {
+            object_uri,
+            byte_range,
+        }
+    }
+}
+
+const U64_LEN: usize = 8;
+/// Object URIs come from `Storage::uri().join(path)`. S3 object keys max out
+/// at 1024 bytes; 8 KiB leaves room for scheme, bucket, and prefix.
+const MAX_OBJECT_URI_LEN: usize = 8 * 1024;
+
+fn ensure_uri_len_in_range(uri_len: usize) -> foyer::Result<()> {
+    if uri_len > MAX_OBJECT_URI_LEN {
+        return Err(foyer::Error::new(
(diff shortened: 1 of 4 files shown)


quickwit/quickwit-storage/src/split_range_cache/key.rs
```diff
@@ -0,0 +1,211 @@
...
+/// Manual `Code` impl: keep Foyer's serde feature off so bincode is not pulled
+/// in for keys or values. Foyer 0.22.3 already implements `Code` for `Bytes`,
+/// so only this key needs a codec. Enabling serde would encode each value byte
+/// as an integer; for a 15 MiB payload that is millions of serializer visits
+/// instead of one `write_all`.
+impl foyer::Code for SplitRangeCacheKey {
+    fn encode(&self, writer: &mut impl Write) -> foyer::Result<()> {
+        let uri_bytes = self.object_uri.as_bytes();
+        ensure_uri_len_in_range(uri_bytes.len())?;
+        let uri_len = u64_from_usize(uri_bytes.len(), "object URI is too long")?;
```
fulmicoton-dd: The two checks are redundant. The first one passing implies the second one will pass.
The second is REALLY unnecessary. Is the point to guard us in case we compile quickwit on a 128 bit platform?

quickwit/quickwit-storage/src/split_range_cache/key.rs
```diff
@@ -0,0 +1,211 @@
...
+        write_u64(writer, uri_len)?;
+        writer
+            .write_all(uri_bytes)
+            .map_err(foyer::Error::io_error)?;
+        let start = u64_from_usize(self.byte_range.start, "range start is too large")?;
```
fulmicoton-dd: same as above. When does u64_from_usize fail?

quickwit/quickwit-storage/src/split_range_cache/key.rs
```diff
@@ -0,0 +1,211 @@
...
+        let mut uri_bytes = vec![0; uri_len];
+        reader
+            .read_exact(&mut uri_bytes)
+            .map_err(foyer::Error::io_error)?;
+        let object_uri = String::from_utf8(uri_bytes).map_err(|error| {
+            foyer::Error::new(foyer::ErrorKind::Parse, "object URI is not UTF-8").with_source(error)
+        })?;
+        let start = usize_from_u64(read_u64(reader)?, "range start does not fit usize")?;
+        let end = usize_from_u64(read_u64(reader)?, "range end does not fit usize")?;
+        if start > end {
```
fulmicoton-dd: And we have a debug_assert in Self::new too.

quickwit/quickwit-storage/src/split_range_cache/key.rs
```diff
@@ -0,0 +1,211 @@
...
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct SplitRangeCacheKey {
+    pub object_uri: String,
+    pub byte_range: Range<usize>,
+}
+
+impl SplitRangeCacheKey {
+    pub(crate) fn new(object_uri: String, byte_range: Range<usize>) -> Self {
+        debug_assert!(
+            byte_range.start <= byte_range.end,
```
fulmicoton-dd: I would drop the Range assert and consider that such a Range is invalid to begin with (responsability of the caller to not build it, the intent of the API is clear and enforced by the contract.

I would also remove the constructor entirely, considering the struct building is actually clearer.

quickwit/quickwit-storage/src/split_range_cache/key.rs
congx4: removed this.

quickwit/quickwit-storage/src/split_range_cache/key.rs
congx4: removed this.

quickwit/quickwit-storage/src/split_range_cache/key.rs
congx4: removed this.

quickwit/quickwit-storage/src/split_range_cache/key.rs
congx4: removed this.

congx4:
Superseded by #6781, which preserves this work in a clean five-commit history rebased onto main.

PR #6705, opened 2026-08-18
Add object listing to storage trait

Mallets:
## Summary

This PR extends the `Storage` trait by adding a `list` method that allows to receive a stream of objects metadata present in a path.

The method returns `ListObjectsStream`, a boxed stream of `StorageResult<Vec<ListObjectMetadata>>`, where each item is one batch of listed objects. 

Streaming means callers can start processing results before the whole listing completes, which matters for prefixes holding a large number of objects. Each `ListObjectMetadata` carries:
- the object path (relative to the storage root, like every other `Storage` operation), 
- the size in bytes, 
- the last modification time.

Errors are reported through the stream rather than aborting it upfront, so a failure on the third page surfaces after the first two pages have already been delivered to the caller.

## S3 storage

The S3 implementation uses the `ListObjectsV2` API, which returns results in paginated responses. It automatically follows continuation tokens and emits each page as a batch in the returned stream, keeping pagination details hidden from the caller. 

Page-level failures are surfaced through the stream rather than aborting the listing upfront. A few implementation notes:
- Pages are fetched lazily: no request is issued until the stream is polled, and each page is requested only when the previous one has been consumed.
- Listing requests go through the shared request semaphore, respecting the same concurrency limit as the other object store operations.
- `ListObjectsV2Error` implements `AwsRetryable`, so listing gets the same retry classification as the other S3 calls.

## How was this PR tested?

- `cargo test -p quickwit-storage test_s3_compatible_storage_list_streams_all_pages --lib` — covers multi-page listing against a mocked S3 client, asserting that both pages are streamed, that the continuation token is sent on the second request, and that metadata is converted correctly.
- `cargo check -p quickwit-storage --all-features`
- `cargo clippy -p quickwit-storage --all-features --tests -- -D warnings`
- `cargo +nightly fmt -p quickwit-storage -- --check`


quickwit/quickwit-storage/src/storage.rs
```diff
@@ -127,6 +144,18 @@ pub trait Storage: fmt::Debug + Send + Sync + 'static {
     /// successfully deleted while others are not.
     async fn bulk_delete<'a>(&self, paths: &[&'a Path]) -> Result<(), BulkDeleteError>;
 
+    /// Lists object metadata for objects whose paths start with `prefix`.
+    ///
+    /// Returned paths are relative to this storage root, like every other [`Storage`] operation.
+    fn list(&self, _prefix: &Path) -> ListObjectsStream {
```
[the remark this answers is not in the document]
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/commits/93705848561c5b482932b75e75079c96b7483f9d

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -958,6 +959,158 @@ impl Storage for S3CompatibleObjectStorage {
...
+                            list_request = list_request.continuation_token(token);
+                        }
+                        list_request.send()
+                    })
+                    .await;
+                    match list_result {
+                        Ok(response) => response,
+                        Err(error) => {
+                            let storage_error = StorageErrorKind::Service
+                                .with_error(anyhow::anyhow!("failed to list objects: {error:?}"));
```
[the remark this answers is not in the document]
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/commits/93705848561c5b482932b75e75079c96b7483f9d

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -958,6 +959,158 @@ impl Storage for S3CompatibleObjectStorage {
...
+            bucket,
+            key_prefix,
+            storage_prefix,
+            retry_params,
+            list_position,
+            list_span,
+        ): ListState| {
+            let page_span = list_span.clone();
+            async move {
+                let continuation_token = match list_position {
```
guilload:
```suggestion
                let continuation_token_opt = match list_position {
```

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -958,6 +959,158 @@ impl Storage for S3CompatibleObjectStorage {
...
+                // Fetch one S3 page. The state owns everything needed by this future so the
+                // returned stream does not borrow the storage instance.
+                let response = {
+                    let _permit = REQUEST_SEMAPHORE.acquire().await;
+                    let list_result = aws_retry(&retry_params, || {
+                        let mut list_request = s3_client
+                            .list_objects_v2()
+                            .bucket(&bucket)
+                            .prefix(&key_prefix);
+                        if let Some(token) = continuation_token.as_deref() {
```
guilload:
```suggestion
                        if let Some(continuation_token) = continuation_token_opt.as_deref() {
```

guilload: nit: `&continuation_token` also works instead of `continuation_token.as_deref()`

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -958,6 +959,158 @@ impl Storage for S3CompatibleObjectStorage {
...
+                            let storage_error =
+                                StorageError::from(error).add_context("failed to list objects");
+                            return Err(storage_error);
+                        }
+                    }
+                };
+
+                // Convert S3 metadata to paths relative to the configured storage root.
+                let mut page_objects = Vec::with_capacity(response.contents().len());
+                for object in response.contents() {
```
guilload: the conversion logic from object to response could be encapsulated in one function. That might make the whole body of `list` easier to read.

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -958,6 +959,158 @@ impl Storage for S3CompatibleObjectStorage {
...
+                                StorageErrorKind::Internal.with_error(anyhow::anyhow!(
+                                    "listed object `{key}` is not under storage prefix `{}`: \
+                                     {error}",
+                                    storage_prefix.display()
+                                ));
+                            return Err(storage_error);
+                        }
+                    };
+                    let size_bytes = match object.size() {
+                        Some(size) => match u64::try_from(size) {
```
guilload: I think we can safely assume that `size` is always positive so `size as u64` is safe, or we can we can also `u64::try_from(size).expect("`size` should be positive");

Are you using Codex by any chance? I find that Codex is needlessly defensive and verbose for error handling.

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -958,6 +959,158 @@ impl Storage for S3CompatibleObjectStorage {
...
+        let fetch_next_page = |(
+            s3_client,
+            bucket,
+            key_prefix,
+            storage_prefix,
+            retry_params,
+            list_position,
+            list_span,
+        ): ListState| {
+            let page_span = list_span.clone();
```
guilload: I't be nicer to have a span for `list` and a child span for each call to `list_objects`

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -958,6 +959,158 @@ impl Storage for S3CompatibleObjectStorage {
         }
     }
 
+    #[instrument(name = "storage.s3.list", level = "debug", skip(self))]
```
guilload: Add prefix for the list call as span attribute?

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -958,6 +959,158 @@ impl Storage for S3CompatibleObjectStorage {
...
+                    let metadata = ListObjectMetadata {
+                        path: relative_path,
+                        size_bytes,
+                        last_modified,
+                    };
+                    page_objects.push(metadata);
+                }
+
+                let next_list_position = response
+                    .next_continuation_token()
```
guilload:
```suggestion
                    .next_continuation_token
```
You can take ownership of the continuation token avoid the copy below `token.to_string()`.

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -958,6 +959,158 @@ impl Storage for S3CompatibleObjectStorage {
...
+
+                // Fetch one S3 page. The state owns everything needed by this future so the
+                // returned stream does not borrow the storage instance.
+                let response = {
+                    let _permit = REQUEST_SEMAPHORE.acquire().await;
+                    let list_result = aws_retry(&retry_params, || {
+                        let mut list_request = s3_client
+                            .list_objects_v2()
+                            .bucket(&bucket)
+                            .prefix(&key_prefix);
```
guilload, line 1033: `into_paginator` exists on this builder, why not use that directly?

quickwit/quickwit-storage/src/storage.rs
```diff
@@ -26,6 +29,20 @@ use tracing::error;
 
 use crate::{BulkDeleteError, OwnedBytes, PutPayload, StorageErrorKind, StorageResult};
 
+/// Metadata for an object listed from storage.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ListObjectMetadata {
```
guilload:
```suggestion
pub struct ObjectMetadata {
```
nit: I think that not so tied to the list API that we can simply can this `ObjectMetadata`.

quickwit/quickwit-storage/src/storage.rs
```diff
@@ -26,6 +29,20 @@ use tracing::error;
...
+    /// Object path relative to this storage root.
+    pub path: PathBuf,
+    /// Object size in bytes.
+    pub size_bytes: u64,
```
guilload: Use `ByteSize`?

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
Mallets: In principle we can assume that size is always positive but the type returned by the sdk is `Option<i64>`. 

I prefer to stick to proper parsing by no-trust principle, i.e. never trust external input or assume that the current behavior is going to be also the future behavior.

I'm using Codex yes, but this is my doing :)

quickwit/quickwit-storage/src/storage.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/changes/9af790e463256af89911dda18b95ad9101347e39

quickwit/quickwit-storage/src/storage.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/changes/9af790e463256af89911dda18b95ad9101347e39

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/changes/9af790e463256af89911dda18b95ad9101347e39

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/changes/9af790e463256af89911dda18b95ad9101347e39

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/changes/9af790e463256af89911dda18b95ad9101347e39

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/changes/9af790e463256af89911dda18b95ad9101347e39

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/commits/471e761598025060ade376b165cdcbaa489b2b7f

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/commits/471e761598025060ade376b165cdcbaa489b2b7f

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
Mallets: Because `into_paginator` is not compatible with the usage of quickwit's `aws_retry(&retry_params, ...)`. It would definitely make the function simpler but we would loose control over the retry policy and alignment with all the other S3 operations (e.g., `get_to_bytes`, `get_slice_stream`, `copy_to`, `delete`, etc.).

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -958,6 +960,93 @@ impl Storage for S3CompatibleObjectStorage {
...
+                    if let Some(continuation_token) = &continuation_token_opt {
+                        list_request = list_request.continuation_token(continuation_token);
+                    }
+                    list_request.send()
+                })
+                .instrument(list_objects_span)
+                .await
+                .map_err(|error| StorageError::from(error).add_context("failed to list objects"))?;
+
+                let objects = convert_list_objects(response.contents(), &storage_prefix)?;
```
guilload: nit: because `convert_list_objects` can skips some object, `objects` can be empty. Would be nice to filter those out.

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
guilload: Fair. Dedicated spans is also nice.

guilload (approved):

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
Mallets: Nice catch. Done in https://github.com/quickwit-oss/quickwit/pull/6705/commits/65177c4f598e4fb2136d93cb287c22ee75951593.

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -1037,6 +1135,61 @@ impl Storage for S3CompatibleObjectStorage {
...
+    storage_prefix: &Path,
+) -> StorageResult<Vec<ObjectMetadata>> {
+    let mut object_metadata = Vec::with_capacity(objects.len());
+    for object in objects {
+        let Some(key) = object.key() else {
+            warn!("listed object has no key, skipping");
+            continue;
+        };
+        let relative_path = Path::new(key)
+            .strip_prefix(storage_prefix)
```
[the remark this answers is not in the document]
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/commits/3b0baa533f9f6d9818101455686778c6ec71dc95

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
```diff
@@ -958,6 +968,102 @@ impl Storage for S3CompatibleObjectStorage {
         }
     }
 
+    #[instrument(name = "storage.s3.list", level = "debug", skip(self), fields(prefix = %prefix.display()))]
+    fn list(&self, prefix: &Path) -> ListObjectsStream {
+        let s3_client = self.s3_client.clone();
+        let bucket = self.bucket.clone();
+        let key_prefix = self.key(prefix);
```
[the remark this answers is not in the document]
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6705/commits/b7e9e5d03eef4d50e9b804130d38760948309b0b

quickwit/quickwit-storage/src/object_storage/s3_compatible_storage.rs
guilload: Sounds good.

PR #6706, opened 2026-08-18
Add split ID inclusion filter

Mallets:
## Summary

This PR extends `ListSplitsQuery` with an optional inclusion filter that limits results to a caller-provided set of split IDs.

The filter is implemented consistently in both metastore backends:

* File-backed listings reject splits that are outside the inclusion set.
* PostgreSQL listings use a single typed array bind with `= ANY(...)`, avoiding PostgreSQL's parameter limit for large sets.
* The JSON field is optional when deserializing, preserving compatibility with requests sent by older callers during a rolling upgrade.

An empty inclusion set deliberately matches no splits; an absent filter leaves listings unchanged.

## How was this PR tested?

* `cargo test -p quickwit-metastore --all-features included_split_ids --lib` — validates the PostgreSQL SQL builder and backward-compatible query deserialization.
* `cargo test -p quickwit-metastore --all-features test_metastore_list_all_splits --lib` — the file-backed metastore integration test passed; the PostgreSQL variant requires `QW_TEST_DATABASE_URL`, which is not configured locally.


quickwit/quickwit-metastore/src/metastore/file_backed/file_backed_index/mod.rs
```diff
@@ -1291,6 +1291,12 @@ fn split_query_predicate(split: &&Split, query: &ListSplitsQuery) -> bool {
         }
     }
 
+    if let Some(included_split_ids) = &query.included_split_ids
+        && !included_split_ids.contains(&split.split_metadata.split_id)
```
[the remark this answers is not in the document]
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6706/commits/c3a5423dbf9f2e6a387c2683ce800f190f05a71a

quickwit/quickwit-metastore/src/metastore/file_backed/file_backed_index/mod.rs
```diff
@@ -1253,6 +1253,12 @@ fn split_query_predicate(split: &&Split, query: &ListSplitsQuery) -> bool {
         return false;
     }
 
+    if let Some(included_split_ids) = &query.included_split_ids
```
guilload: The inclusion and exclusion logic should live next to each other.

quickwit/quickwit-metastore/src/metastore/postgres/utils.rs
```diff
@@ -217,6 +217,28 @@ pub(super) fn append_query_filters_and_order_by(sql: &mut SelectStatement, query
         );
     }
 
+    if let Some(included_split_ids) = query.included_split_ids {
```
guilload: Same as above. Remove duplicated code between two blocks.

quickwit/quickwit-metastore/src/metastore/mod.rs
```diff
@@ -1077,6 +1077,15 @@ pub struct ListSplitsQuery {
     /// Only return splits whose (index_uid, split_id) are lexicographically after this split
     pub after_split: Option<(IndexUid, SplitId)>,
 
+    /// Include only splits whose `split_id` appears in this set. `None` means no inclusion filter;
+    /// an empty set matches no splits.
+    ///
+    /// `#[serde(default)]` keeps this field optional on the wire: the query crosses nodes as JSON,
+    /// so during a rolling upgrade an older caller serializes it without this key and an upgraded
+    /// metastore must still deserialize the query (an absent set means "no inclusion filter").
+    #[serde(default)]
```
guilload: skip serialization if is none to cover the other case (rollback)

quickwit/quickwit-metastore/src/metastore/file_backed/file_backed_index/mod.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6706/commits/b99cb9d1c2d61059c215db72d51875e26d28b593. 

I moved both cheap `included` and `excluded` checks at the beginning since [this comment](https://github.com/quickwit-oss/quickwit/pull/6706#discussion_r3806019193) applies to both.

quickwit/quickwit-metastore/src/metastore/postgres/utils.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6706/changes/e33cd4c1a3635926da3256bbfa5b743b74acdac8

quickwit/quickwit-metastore/src/metastore/mod.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6706/changes/e33cd4c1a3635926da3256bbfa5b743b74acdac8

guilload (approved):
nit: Do we want to align how inclusion and exclusion lists are defined?
```
inclusion: Option<HashSet>
exclusion: HashSet,
```

I know that `None` and `Some(HashSet::empty())` mean different things, but if we consider that the latter is not really useful, then we could get rid of the option.

Also if there's an inclusion list, does it make sense to evaluate the exclusion branch in the file backed metastore or add the exclusion predicate in Pg.

Finally, do we need to be defensive against inclusion exclusion lists with a non-null intersection?

Mallets:
I'm fine with aligning `inclusion` to `HashSet`, I simply tend to be explicit on _whose responsibility is_ via explicit types.

W.r.t. `if there's an inclusion list, does it make sense to evaluate the exclusion branch`, I'm usually on the side that if the caller explicitly set both, and the API allows both, we execute both. I'd rather treat it as an API design concern rather then every storage backend concern.

quickwit/quickwit-metastore/src/metastore/file_backed/file_backed_index/mod.rs
```diff
@@ -1409,6 +1418,11 @@ mod tests {
         assert!(!split_query_predicate(&&split_2, &query));
         assert!(split_query_predicate(&&split_3, &query));
 
+        let query = ListSplitsQuery::for_index(IndexUid::new_with_random_ulid("test-index"))
+            .retain_mature(time::OffsetDateTime::now_utc())
+            .with_included_split_ids(Default::default());
+        assert!(!split_query_predicate(&&split_1, &query));
```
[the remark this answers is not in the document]
Mallets: Fixed in https://github.com/quickwit-oss/quickwit/pull/6706/commits/830424f442c0d59085e2b2731acbca585341e157

PR #6711, opened 2026-08-19
feat(search): use stable node IDs for split affinity to keep disk cache warm when searcher pods restart

congx4:
This is independent of the split-range disk cache stack. It is useful for cache locality after a same-node restart, but it does not require that cache.

## Summary
- Search job placement (and related rendezvous uses: KV replication, `report_splits`) now hashes the cluster `node_id` / `QW_NODE_ID` instead of the gRPC `SocketAddr`.
- The searcher pool stays keyed by address, so RPC dialing and DataFusion worker URLs are unchanged. A restart that keeps the same node ID keeps the same split affinity even if the listen address changes.
- `node_id` docs note that searchers should keep this ID stable (for example a StatefulSet pod name) if the same splits should keep landing on the same node.

## What this does not fix
- **One-time remap:** deploying this changes every split's preferred searcher once, so any address-based warm cache goes cold for that rollout.
- **Unstable IDs:** if `QW_NODE_ID` is a random Deployment hostname, affinity is still unstable. The default remains the short hostname.
- **Generation is not hashed:** Chitchat's generation is a per-process incarnation (restart counter). Hashing it would move splits on every restart, which is the opposite of this PR. A higher generation keeps the same affinity as long as `node_id` is unchanged.
- **Scroll KV during mixed-version rollout:** `put_kv`/`get_kv` share this ranking. While old and new coordinators coexist, an in-flight scroll can miss a live replica (possible at 8+ searchers). Homogeneous fleet and replica restarts are unchanged from today; MiniKV is still in-memory.

## Test plan
- [x] `cargo test -p quickwit-search --all-features --lib`
- [x] `cargo test -p quickwit-serve --all-features --lib -- setup_searcher datafusion_worker`
- [x] `cargo test -p quickwit-janitor --all-features --lib`
- [x] `cargo test -p quickwit-datafusion --all-features`
- [x] `cargo clippy -p quickwit-search -p quickwit-serve -p quickwit-datafusion --all-features --tests -- -D warnings`
- [ ] Confirm a two-searcher restart with a stable `QW_NODE_ID` and a changed gRPC port keeps the same split assignment


quickwit/quickwit-search/src/search_job_placer.rs
```diff
@@ -115,13 +117,13 @@ impl SearchJobPlacer {
...
 struct SocketAddrAndClient {
-    socket_addr: SocketAddr,
+    node_id: NodeId,
     client: SearchServiceClient,
 }
 
 impl Hash for SocketAddrAndClient {
     fn hash<H: Hasher>(&self, hasher: &mut H) {
-        SocketAddrLegacyHash(&self.socket_addr).hash(hasher);
+        self.node_id.hash(hasher);
```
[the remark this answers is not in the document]
congx4: This is valid only during the rolling upgrades and it is recovered after the updates.

quickwit/quickwit-search/src/search_job_placer.rs
```diff
@@ -114,14 +116,14 @@ impl SearchJobPlacer {
...
-    socket_addr: SocketAddr,
+struct NodeIdAndClient {
+    node_id: NodeId,
     client: SearchServiceClient,
 }
 
-impl Hash for SocketAddrAndClient {
+impl Hash for NodeIdAndClient {
     fn hash<H: Hasher>(&self, hasher: &mut H) {
-        SocketAddrLegacyHash(&self.socket_addr).hash(hasher);
```
trinity-1686a, line 124: unless it's still used somewhere, can we remove `SocketAddrLegacyHash` from quickwit-common, it's probably dead code after this PR

trinity-1686a (approved):

quickwit/quickwit-search/src/search_job_placer.rs
congx4: yes, I removed it.

PR #6721, opened 2026-08-25
Add calculated predicate search plumbing

Mallets:
## Summary
- add calculated predicate expressions to the search protocol and lower them into Tantivy leaf queries
- warm predicate fast fields before synchronous search execution
- disable metadata-only count shortcuts when a calculated predicate must be evaluated
- update the Tantivy revision and adapt Quickwit to its custom field and segment component APIs

## Test plan
- [x] `cargo test -p quickwit-search calculated_predicate --lib`
- [x] `cargo test -p quickwit-search test_metadata_count_request_with_calculated_predicate --lib`
- [x] `cargo +nightly fmt --all`
- [ ] exercise a calculated predicate through the full remote-storage root-to-leaf search path


quickwit/quickwit-directories/src/hot_directory.rs
```diff
@@ -459,11 +460,24 @@ impl Directory for HotDirectory {
...
+        SegmentComponent::Postings,
+        SegmentComponent::Positions,
+        SegmentComponent::Terms,
+        SegmentComponent::Store,
+        SegmentComponent::FastFields,
+        SegmentComponent::FieldNorms,
+        SegmentComponent::Delete,
+    ];
+    let mut files = HashSet::new();
+    for segment_meta in index_meta.segments {
```
fulmicoton, line 473: I don't understadn what this is about? 

You were hit by the new tantivy plugin change that the list_files method does not exist anymore? I think there is another utility to do that computation no?

quickwit/quickwit-indexing/src/actors/packager.rs
```diff
@@ -187,15 +187,24 @@ fn list_split_files(
     scratch_directory: &TempDirectory,
 ) -> io::Result<Vec<PathBuf>> {
     let mut split_files = vec![scratch_directory.path().join("meta.json")];
+    let segment_components = [
+        SegmentComponent::Postings,
+        SegmentComponent::Positions,
+        SegmentComponent::Terms,
+        SegmentComponent::Store,
```
fulmicoton, line 194: same as above.

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -200,10 +200,85 @@ enum ListFieldsType {
     BYTES = 7;
     IP_ADDR = 8;
     JSON = 9;
+    CUSTOM = 10;
```
fulmicoton, line 203: do we need this in protobuf land? Shouldn't we just return an error or panic before?

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -280,6 +355,9 @@ message SearchRequest {
   // Scheduling priority for leaf search execution. Negative values are allowed,
   // and lower values have higher priority. Callers that omit it get priority 0.
   int32 priority = 21;
+
+  // Predicate expression evaluated by Tantivy against fast fields on each leaf.
+  optional CalculatedPredicate calculated_predicate = 22;
```
fulmicoton, line 360: I don't understand how this can make sense with the existing query_ast.

quickwit/Cargo.lock
```diff
@@ -14,7 +14,7 @@ version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
- "gimli",
+ "gimli 0.32.3",
```
fulmicoton, line 17: can you go through the added dependencies and double check if they are justfiied or not pulled at all?

quickwit/Cargo.lock
```diff
@@ -2434,6 +2443,186 @@ dependencies = [
...
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "16c273ba1bfc3a1cb27cb4f83df27134f3cb638a61f5ed79ab22cf86eb13da0d"
+dependencies = [
+ "cranelift-assembler-x64-meta",
+]
+
+[[package]]
+name = "cranelift-assembler-x64-meta"
+version = "0.134.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
```
fulmicoton, line 2469: that one is likely unnecessary for instance. This is not a change to do in quicwkit, but in tantivy but still very important to follow up.

quickwit/Cargo.lock
```diff
@@ -13301,6 +13557,28 @@ dependencies = [
  "web-sys",
 ]
 
+[[package]]
+name = "wasmtime-internal-core"
```
fulmicoton, line 13561: im surprised we need that. do we need to disable things in cranelift? (again, in tantivy)

quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -200,10 +200,85 @@ enum ListFieldsType {
...
+    int64 int_value = 1;
+    uint64 uint_value = 2;
+    // IEEE-754 bits for an f64 literal. This keeps SearchRequest Eq/Hash-safe.
+    fixed64 double_value_bits = 3;
+    string string_value = 4;
+    bool bool_value = 5;
+  }
+}
+
+message CalculatedPredicateFuncCall {
```
fulmicoton, line 235: The query ast is expressed using json today. I don't think it we need to change this?

quickwit/quickwit-directories/src/hot_directory.rs
fulmicoton: It seems like we DO need to modify tantivy.

PR #6722, opened 2026-08-25
Make compaction planner configurable

nadav-govari:
### Description

Makes the planner configurable. At high scale this could be desirable.


nadav-govari:
I dont think we need this yet. We can revive it if we do.

PR #6689, opened 2026-08-13
Store extra metadata in splits for disaster recovery

guilload:
### Description
Store extra metadata in splits so one can (somewhat) recover the metastore database from storage.

### How was this PR tested?
Add unit tests and integrations
Tested locally:
- wrote two splits locally with legacy format
- wrote two splits locally with new format
- performed searches against old splits, new splits, all splits

PR #6723, opened 2026-08-26
Optimize split recovery metadata reads from split bundles

Mallets:
## Summary

Disaster recovery only needs the recovery metadata stored in each split, but it may need to inspect a large number of splits. Separate requests to discover each split’s size, locate its footer, and fetch the recovery metadata would multiply object-storage GETs across the bucket.

Bucket listings already provide each split’s object length, so callers can pass it directly and avoid a separate metadata request. We then optimistically read the final `1 MiB` of the split. In the expected case, this contains the complete footer and recovery metadata, allowing retrieval with a single GET per split.

## Details
- Add `BundleStorage::fetch_file_from_split` for retrieving one bundled file when the split length is known.
- Parse the footer from the initial tail read and reuse those bytes when they contain the requested file.
- Widen the tail read or fetch the file separately only when an unusually large footer or legacy layout requires it.

## Test plan
- [x] `make fmt`
- [x] `cargo clippy -p quickwit-storage --all-features --tests`
- [x] `cargo nextest run -p quickwit-storage --all-features bundle_storage`

### DIFF
--- a/quickwit/quickwit-storage/src/bundle_storage.rs
+++ b/quickwit/quickwit-storage/src/bundle_storage.rs
@@ -239,7 +239,7 @@ fn locate_split_footer_range_from_metadata_len(
 
 fn locate_split_footer_range_in_tail(
     split_len: u64,
-    tail_bytes: &OwnedBytes,
+    tail_bytes: &[u8],
 ) -> anyhow::Result<FooterLocation> {
     // Legacy split layout:
     // [body][bundle metadata][metadata len][hotcache][hotcache len]
@@ -247,12 +247,11 @@ fn locate_split_footer_range_in_tail(
         tail_bytes.len() as u64 <= split_len,
         "split tail is longer than the split itself"
     );
-    let bytes = tail_bytes.as_slice();
-    let trailer_start = bytes
+    let trailer_start = tail_bytes
         .len()
         .checked_sub(SPLIT_FOOTER_TRAILER_NUM_BYTES)
         .context("split tail is too short to contain a footer trailer")?;
-    if let Some(footer_start) = deserialize_split_footer_trailer(&bytes[trailer_start..])? {
+    if let Some(footer_start) = deserialize_split_footer_trailer(&tail_bytes[trailer_start..])? {
         ensure!(
             footer_start <= split_len - SPLIT_FOOTER_TRAILER_NUM_BYTES as u64,
             "split footer starts after its trailer"
@@ -261,7 +260,8 @@ fn locate_split_footer_range_in_tail(
     }
 
     let hotcache_len =
-        u32::from_le_bytes(bytes[bytes.len() - HOTCACHE_LEN_NUM_BYTES..].try_into()?) as u64;
+        u32::from_le_bytes(tail_bytes[tail_bytes.len() - HOTCACHE_LEN_NUM_BYTES..].try_into()?)
+            as u64;
     let bundle_metadata_len_end = split_len
         .checked_sub(HOTCACHE_LEN_NUM_BYTES as u64)
         .and_then(|offset| offset.checked_sub(hotcache_len))
@@ -282,7 +282,7 @@ fn locate_split_footer_range_in_tail(
     let footer_range = locate_split_footer_range_from_metadata_len(
         split_len,
         bundle_metadata_len_start,
-        &bytes[relative_start..relative_end],
+        &tail_bytes[relative_start..relative_end],
     )?;
     Ok(FooterLocation::Located(footer_range))
 }
--- a/quickwit/quickwit-storage/src/bundle_storage.rs
+++ b/quickwit/quickwit-storage/src/bundle_storage.rs
@@ -93,6 +93,62 @@ impl BundleStorage {
     pub fn iter_files(&self) -> impl Iterator<Item = &PathBuf> {
         self.file_ranges.files.keys()
     }
+
+    /// Fetches a bundled file from a split.
+    ///
+    /// Use this only when retrieving a single file from the split. To retrieve multiple files,
+    /// prefer [`Self::open_from_storage`].
+    ///
+    /// The split length is provided by the caller (e.g. from object listing metadata) to avoid a
+    /// separate metadata request.
+    pub async fn fetch_file_from_split(
+        storage: Arc<dyn Storage>,
+        bundle_filepath: PathBuf,
+        split_path: &Path,
+        split_len: u64,
+    ) -> anyhow::Result<(OwnedBytes, Range<u64>)> {
+        let (split_bytes, footer_range) = fetch_split_tail(
+            storage.as_ref(),
+            split_path,
+            split_len,
+            DEFAULT_SPLIT_TAIL_WINDOW_NUM_BYTES,
+        )
+        .await?;
+
+        // Parse the bundle file ranges from the split bytes.
+        let tail_start = split_len - split_bytes.len() as u64;
+        let (file_ranges, _hotcache) =
+            BundleFileRanges::open_from_split_bytes(split_bytes.clone())?;
+        let file_range = file_ranges.get(&bundle_filepath).ok_or_else(|| {
+            anyhow::anyhow!(
+                "missing file `{}` in split bundle",
+                bundle_filepath.display()
+            )
+        })?;
+        ensure!(
+            file_range.start <= file_range.end,
+            "bundled file range starts after it ends"
+        );
+        ensure!(
+            file_range.end <= footer_range.start,
+            "bundled file range overlaps split footer"
+        );
+
+        // If the initial tail also contains the file, reuse it and complete in one GET (at least).
+        // Otherwise, fetch the file with an additional GET.
+        let file_bytes = if file_range.start >= tail_start {
+            let relative_start = usize::try_from(file_range.start - tail_start)?;
+            let relative_end = usize::try_from(file_range.end - tail_start)?;
+            split_bytes.slice(relative_start..relative_end)
+        } else {
+            let relative_start = usize::try_from(file_range.start)?;
+            let relative_end = usize::try_from(file_range.end)?;
+            storage
+                .get_slice(split_path, relative_start..relative_end)
+                .await?
+        };
+        Ok((file_bytes, footer_range))
+    }
 }
 
 const HOTCACHE_LEN_NUM_BYTES: usize = std::mem::size_of::<u32>();
@@ -104,6 +160,7 @@ const SPLIT_FOOTER_TRAILER_VERSION_NUM_BYTES: usize = std::mem::size_of::<u32>()
 pub(crate) const SPLIT_FOOTER_TRAILER_NUM_BYTES: usize = SPLIT_FOOTER_START_NUM_BYTES
     + SPLIT_FOOTER_TRAILER_VERSION_NUM_BYTES
     + SPLIT_FOOTER_TRAILER_MAGIC.len();
+const DEFAULT_SPLIT_TAIL_WINDOW_NUM_BYTES: u64 = 1024 * 1024;
 
 pub(crate) fn serialize_split_footer_trailer(
     footer_start_inclusive: u64,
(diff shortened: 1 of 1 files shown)


quickwit/quickwit-storage/src/bundle_storage.rs
```diff
@@ -93,6 +93,62 @@ impl BundleStorage {
...
+        );
+        ensure!(
+            file_range.end <= footer_range.start,
+            "bundled file range overlaps split footer"
+        );
+
+        // If the initial tail also contains the file, reuse it and complete in one GET (at least).
+        // Otherwise, fetch the file with an additional GET.
+        let file_bytes = if file_range.start >= tail_start {
+            let relative_start = usize::try_from(file_range.start - tail_start)?;
```
fulmicoton: when does this fail?

Mallets: Technically it will never fail there, it will fail first in the [fetch_split_tail](https://github.com/quickwit-oss/quickwit/pull/6723/changes/BASE..9214ad9da2dd0c4591659673228d2d6b828d6ef2#diff-df9309c1b9d3da58095cd2ca77c31a4f8bda8c7d245c6ad2ade0ba026cee8e43R324).

The failure case is when running on `32 bits` and the split is larger than `4GB`. However, I'm not sure either that the whole system can run on `32 bits`, in that case these faillable conversions can be replaced with a simple cast.

fulmicoton-dd: AI generates crazy defensive code like this. 
Last week Cong had a PR with u64::try_from(usize),
and codex keep generating thing like this for me too.

I don't think we need to accept those. It hurts readability IMHO

Mallets: Yeah, `u64::try_from(usize)` is non-sense. But for faillable conversions on supported platforms it makes sense to me. Said that, we can consider `32 bits` as a non-supported target.

Done in https://github.com/quickwit-oss/quickwit/pull/6723/commits/5947c79e29ecca4d53d618064aaa6f3ad050e457.

quickwit/quickwit-storage/src/bundle_storage.rs
```diff
@@ -145,44 +202,152 @@ pub async fn locate_split_footer_range(
...
+    initial_tail_window_num_bytes: u64,
+) -> anyhow::Result<(OwnedBytes, Range<u64>)> {
+    ensure!(
+        split_len >= SPLIT_FOOTER_TRAILER_NUM_BYTES as u64,
+        "split is too short to contain a footer"
+    );
+    ensure!(
+        initial_tail_window_num_bytes > 0,
+        "split tail window must be positive"
+    );
```
fulmicoton-dd: is this important?

quickwit/quickwit-storage/src/bundle_storage.rs
```diff
@@ -145,44 +202,152 @@ pub async fn locate_split_footer_range(
...
+
+    // The tail always covers the fixed trailer so that footer parsing has a lower bound to work
+    // with, even when the caller asks for a very small window.
+    let mut tail_window_num_bytes = initial_tail_window_num_bytes
+        .max(SPLIT_FOOTER_TRAILER_NUM_BYTES as u64)
+        .min(split_len);
+    // Read the tail in a loop until we find the footer.
+    let (tail_bytes, footer_range) = loop {
```
fulmicoton-dd: if possible avoid
- loop without clear exit condition. It makes proofreading hard and is error prone.
- break with value

fulmicoton-dd: here... I don't understand why this is a loop?

quickwit/quickwit-storage/src/bundle_storage.rs
Mallets: Not critical. Removed in https://github.com/quickwit-oss/quickwit/pull/6723/commits/904fcfb9388bf2089eb0f6734c31f8244851ac37.

quickwit/quickwit-storage/src/bundle_storage.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6723/commits/904fcfb9388bf2089eb0f6734c31f8244851ac37.

`loop` is a way to retry the exact operation twice. Fine to make it explicit here.

quickwit/quickwit-storage/src/bundle_storage.rs
```diff
@@ -145,44 +201,165 @@ pub async fn locate_split_footer_range(
...
+    let bundle_metadata_len = u32::from_le_bytes(bundle_metadata_len_bytes.try_into()?) as u64;
+    let footer_start = bundle_metadata_len_start
+        .checked_sub(bundle_metadata_len)
+        .context("split footer exceeds split length")?;
+    Ok(footer_start..split_len)
+}
+
+fn locate_split_footer_range_in_tail(
+    split_len: u64,
+    tail_bytes: &OwnedBytes,
```
fulmicoton:
```suggestion
    tail_bytes: &[u8],
```

Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6723/commits/5eb77e45224c2ef5992d3afb61deaab58117b14c

quickwit/quickwit-storage/src/bundle_storage.rs
```diff
@@ -145,44 +201,165 @@ pub async fn locate_split_footer_range(
...
+    initial_tail_window_num_bytes: u64,
+) -> anyhow::Result<(OwnedBytes, Range<u64>)> {
+    ensure!(
+        split_len >= SPLIT_FOOTER_TRAILER_NUM_BYTES as u64,
+        "split is too short to contain a footer"
+    );
+
+    // Start with the requested window, but always cover the fixed trailer and never read before
+    // the beginning of the split.
+    let initial_tail_num_bytes = initial_tail_window_num_bytes
```
fulmicoton: nitpick. I find it always max/min have some conterintuitive aspect.

We think: the MAXIMUM accepted value is split_len, so we counterintuitively need to compute
x.min(split_len).

initial_tail_window_num_bytes.clamp() helps with that.

Granted... this goes against my common rant about the proliferation of needless methods in rust's std.

quickwit/quickwit-storage/src/bundle_storage.rs
```diff
@@ -145,44 +201,165 @@ pub async fn locate_split_footer_range(
...
+    bundle_metadata_len_bytes: &[u8],
+) -> anyhow::Result<Range<u64>> {
+    let bundle_metadata_len = u32::from_le_bytes(bundle_metadata_len_bytes.try_into()?) as u64;
+    let footer_start = bundle_metadata_len_start
+        .checked_sub(bundle_metadata_len)
+        .context("split footer exceeds split length")?;
+    Ok(footer_start..split_len)
+}
+
+fn locate_split_footer_range_in_tail(
```
fulmicoton, line 243: why is this so complicated? Is it explained somewhere?
Has there been a change of format?

quickwit/quickwit-storage/src/bundle_storage.rs
```diff
@@ -145,44 +201,165 @@ pub async fn locate_split_footer_range(
...
+/// When the window falls short, the tail is re-read: once for a new split, whose trailer gives
+/// the footer start, and up to twice for a legacy split, whose footer length is derived from the
+/// trailing hotcache and bundle-metadata lengths.
+///
+/// Returns the tail, which may start before the footer, and the footer range within the split.
+async fn fetch_split_tail(
+    storage: &dyn Storage,
+    split_path: &Path,
+    split_len: u64,
+    initial_tail_window_num_bytes: u64,
```
fulmicoton, line 306: Is initial_tail_window_num_bytes useful as an argument? I only see a const being used?

quickwit/quickwit-storage/src/bundle_storage.rs
Mallets: Yes, the format has been changed in https://github.com/quickwit-oss/quickwit/pull/6689 and it is explained in: https://github.com/quickwit-oss/quickwit/blob/73c076edcdbbebd398c6952fed3839f92d52eed4/quickwit/quickwit-directories/src/bundle_directory.rs#L29

Note that this PR does not modify the format.

quickwit/quickwit-storage/src/bundle_storage.rs
Mallets: In the context of `fetch_split_tail` standalone function, I believe it's good to accept it as a parameter and it's the caller's responsibility to decide how big the `initial_tail_window` should be. The caller now uses a const, but any change in the logic of the caller shouldn't result in a change in the `fetch_split_tail` function itself.

quickwit/quickwit-storage/src/bundle_storage.rs
Mallets: I wasn't really aware of `clamp()`. Indeed it helps here. Done in https://github.com/quickwit-oss/quickwit/pull/6723/commits/575229abd605119d64dfe88afa6619548e7a38a6.

quickwit/quickwit-storage/src/bundle_storage.rs
```diff
@@ -93,6 +92,62 @@ impl BundleStorage {
...
+
+    /// Fetches a bundled file from a split.
+    ///
+    /// Use this only when retrieving a single file from the split. To retrieve multiple files,
+    /// prefer [`Self::open_from_storage`].
+    ///
+    /// The split length is provided by the caller (e.g. from object listing metadata) to avoid a
+    /// separate metadata request.
+    pub async fn fetch_file_from_split(
+        storage: Arc<dyn Storage>,
```
fulmicoton-dd, line 107: should it be &dyn?

quickwit/quickwit-storage/src/bundle_storage.rs
fulmicoton-dd: This is not a valid point, is it?
The function is private. The value will stay const forever and I don't see any reason to make it the caller responsibility to pick the value, apart from increasing complexity for no reason at all:
that function contract is needlessly complicated as a result.

quickwit/quickwit-storage/src/bundle_storage.rs
```diff
@@ -93,6 +92,62 @@ impl BundleStorage {
...
+        bundle_filepath: PathBuf,
```
fulmicoton-dd, line 108:
```suggestion
        bundle_filepath: &Path,
```

fulmicoton-dd, line 108: bundle filepath? So the manifest is just another file of the bundle I presume.

But the implementation is very specifically targeting that we append at the very end right?
Shouldn't we specialize it to only apply to manifest files? (i don't know how you called them)

fulmicoton-dd: Ah I guess this is not necessary. Worst case scenario we get it into fetch. Ok fair enough.

quickwit/quickwit-storage/src/bundle_storage.rs
```diff
@@ -93,6 +92,62 @@ impl BundleStorage {
     pub fn iter_files(&self) -> impl Iterator<Item = &PathBuf> {
         self.file_ranges.files.keys()
     }
+
+    /// Fetches a bundled file from a split.
+    ///
+    /// Use this only when retrieving a single file from the split. To retrieve multiple files,
+    /// prefer [`Self::open_from_storage`].
+    ///
+    /// The split length is provided by the caller (e.g. from object listing metadata) to avoid a
```
fulmicoton-dd, line 101: can you document the result semantics because this is not clear at all.

fulmicoton-dd (approved):

quickwit/quickwit-storage/src/bundle_storage.rs
Mallets: Done in https://github.com/quickwit-oss/quickwit/pull/6723/commits/579c0b4be44f626a04a212ae7443fb4bbc70fab0

quickwit/quickwit-storage/src/bundle_storage.rs
```diff
@@ -145,44 +204,164 @@ pub async fn locate_split_footer_range(
...
+        FooterLocation::Located(footer_range) => {
+            // The range is known, but the initial tail may not contain the complete footer.
+            footer_range
+        }
+        FooterLocation::ReadBundleMetadataLen(bundle_metadata_len_range) => {
+            // Extend the tail through the legacy bundle-metadata length field, then use that
+            // length to locate the beginning of the footer.
+            let required_tail_num_bytes = split_len - bundle_metadata_len_range.start;
+            let metadata_len_tail_bytes =
+                read_split_tail(storage, split_path, split_len, required_tail_num_bytes).await?;
```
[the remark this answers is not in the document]
Mallets: Legacy splits do not participate in disaster recovery because they do not contain the necessary recovery metadata.

PR #6724, opened 2026-08-26
fix(ci): Remove redundant LocalStack setup from full tests

dayaffe:
## Description

Removes redundant Python and LocalStack setup from the full test workflow. `make test-all` starts Docker Compose, which initializes the required S3 bucket.

Validated with a successful [full test suite run](https://github.com/quickwit-oss/quickwit/actions/runs/33000045227).


pront (approved):
👍 already putting `/ci-run-all-tests` to good use

guilload (approved):

Issue #6719, opened 2026-08-24
MemorySizedCache could be leaking memory.


ncoiffier-celonis:
**Describe the bug**

While looking at the code to try to understand why ~1.4GB of heap memory appears unaccounted for in my production setup, I discovered that `MemorySizedCache` doesn't seem to take into account the size of the key. This is particularly problematic for `LeafSearchCache` which uses the (large) `SearchRequest` as a cache key.

On my production environment, the `partial_request` cache is the default 64MB, but from the metrics, I observe ~1 million entries for 35MB (35B per entry, the cache is half full). By my estimate, the `SearchRequest`, which includes the query AST, should be ~1400B per entry (~1.4GB for all the keys), which matches the ~1.4GB of unaccounted heap memory in my production setup.

- The `LeafSearchCache` cache key is computed [here](https://github.com/quickwit-oss/quickwit/blob/v0.9.0/quickwit/quickwit-search/src/leaf_cache.rs#L78).
- `LeafSearchCache` uses a [MemorySizedCache](https://github.com/quickwit-oss/quickwit/blob/v0.9.0/quickwit/quickwit-search/src/leaf_cache.rs#L48) with the default [LRU](https://github.com/quickwit-oss/quickwit/blob/v0.9.0/quickwit/quickwit-config/src/node_config/mod.rs#L584) implementation
- The LRU cache implementation doesn't take the key into account when [putting an entry in the cache](https://github.com/quickwit-oss/quickwit/blob/v0.9.0/quickwit/quickwit-storage/src/cache/base_cache.rs#L199)

**Full disclaimer**

I haven't been able to identify with certitude the root cause of my problem, this bug report is mostly based on my analysis of the code.

**Steps to reproduce (if applicable)**

I haven't been able to clearly reproduce/find a metric that point to the exact problem, but a unit test similar to this seems to pass, confirming my suspicion:

```
    #[test]
    fn test_key_size_is_not_accounted_for() {
        let cache = MemorySizedCache::<String>::from_config(
            &ByteSize::b(16).into(),
            &CACHE_METRICS_FOR_TESTS,
        );

        let long_key = "k".repeat(1_000);
        // The entry is stored even though key + value is ~1000 bytes, way above the 16 bytes capacity
        cache.put(long_key.clone(), OwnedBytes::new(&b"0123456789"[..]));
    }
```

**Expected behavior**

A partial request cache of size 64MB shouldn't need ~2GB of additional heap to store the keys.

[...]

PR #6727, opened 2026-08-27
Use hash of the cache key to limit in-memory heap overhead for large search requests

### LINKED ISSUE #5878: `partial_request` cache count keeps increasing
### Description

We set `partial_request_cache_capacity` to 256MB. However, when checking the metrics, the cache count appears to keep increasing, and as a result, the Searcher’s memory usage also keeps growing.

Upon further inspection, other caches show a pattern of growing and shrinking over time, but the `partial_request` cache only increases continuously. Eventually, when the cache grows large enough, the process runs out of memory (OOM).

Looking at the `quickwit_cache_in_cache_num_bytes` metric, the size occupied by `partial_request` is below 250MB. Doesn’t this metric represent the total size of the `partial_request` cache? If I’m misunderstanding the meaning of `partial_request_cache_capacity`, is there another way to limit the number of `partial_request` cache entries?

As a workaround, we’ve confirmed that setting `partial_request_cache_capacity` to 0 prevents the cache count from increasing.

I’ve attached the Quickwit configuration and metric screenshots below for reference.

### Metrics 
* Number Of Caches (`partial_request`)
Image

* Number Of Caches (Others)

Image

* Size Of Caches (`partial_request`)

Image

* Size Of Caches (Others)

Image

* Searcher Memory

Image

### Quickwit Node Config
Version: quickwit:qw-airmail-20250522-hotfix
```
{
  "cluster_id": "quickwit-quickwit-searcher",
  "node_id": "quickwit-searcher-searcher-17",
  "enabled_services": [
    "searcher"
  ],
  "gossip_listen_addr": "0.0.0.0:7282",
  "grpc_listen_addr": "0.0.0.0:7281",
  "gossip_advertise_addr": "10.128.77.87:7282",
  "grpc_advertise_addr": "10.128.77.87:7281",
  "gossip_interval": {
    "secs": 1,
    "nanos": 0
  },
  "peer_seeds": [
    "quickwit-searcher-headless"
  ],
  "data_dir_path": "/quickwit/qwdata",
  "metastore_uri": "file:///quickwit/qwdata/indexes#polling_interval=30s",
  "default_index_root_uri": "s3://prod-kr-quickwit-index/indexes",
  "rest_config": {
    "listen_addr": "0.0.0.0:7280",
    "cors_allow_origins": [],
    "extra_headers": {},
    "tls": null
  },
  "grpc_config": {
    "max_message_size": "52.4 MB",
    "tls": null
  },
  "storage_configs": {
    "s3": {
      "access_key_id": null,
      "secret_access_key": null,
      "region": "ap-northeast-2",
      "endpoint": "https://s3.ap-northeast-2.amazonaws.com",
      "force_path_style_access": false,
      "disable_multi_object_delete": false,
      "disable_multipart_upload": false
    }
  },
  "metastore_configs": {},
  "indexer_config": {
    "split_store_max_num_bytes": "107.4 GB",
    "split_store_max_num_splits": 1000,
    "max_concurrent_split_uploads": 12,
    "max_merge_write_throughput": null,
    "merge_concurrency": 2,
    "enable_otlp_endpoint": true,
    "enable_cooperative_indexing": false,
    "cpu_capacity": "4000m"
  },
  "searcher_config": {
    "aggregation_memory_limit": "500.0 MB",
    "aggregation_bucket_limit": 65000,
    "fast_field_cache_capacity": "4.0 GB",
    "split_footer_cache_capacity": "1000.0 MB",
    "partial_request_cache_capacity": "256.0 MB",
    "max_num_concurrent_split_searches": 200,
    "max_num_concurrent_split_streams": 100,
    "request_timeout_secs": 30,
    "warmup_memory_budget": "100.0 GB",
    "warmup_single_split_initial_allocation": "1000.0 MB"
  },
  "ingest_api_config": {
    "max_queue_memory_usage": "2.1 GB",
    "max_queue_disk_usage": "4.3 GB",
    "replication_factor": 1,
    "content_length_limit": "10.5 MB",
    "shard_throughput_limit": "5.2 MB",
    "shard_burst_limit": "52.4 MB",
    "shard_scale_up_factor": 1.5
  },
  "jaeger_config": {
    "enable_endpoint": true,
    "lookback_period_hours": 72,
    "max_trace_duration_secs": 3600,
    "max_fetch_spans": 10000
  }
}
```

ncoiffier-celonis:
### Description

Attempt to fix https://github.com/quickwit-oss/quickwit/issues/5878 and https://github.com/quickwit-oss/quickwit/issues/6719.

Store the hash of the cache key for the `partial_request_cache`; this is particularly problematic for `LeafSearchCache` which uses the (large) `SearchRequest` as a cahce key, including the whole query and for queries that mostly return zero-hit splits.

Hash collision can happen (for a 64MB cache size, the probability of collision is 1e-27). We're limiting the impact of collisions by ensuring that splits belong to the correct search request. In the case of collision, at least data isolation at the index level is preserved (worse case scenario, we would be seeing incorrect document id from the same split or some missing data).

### How was this PR tested?

Unit tests only.

### DIFF
--- a/docs/configuration/node-config.md
+++ b/docs/configuration/node-config.md
@@ -307,7 +307,7 @@ This section contains the configuration options for a Searcher.
 | `aggregation_bucket_limit` | Determines the maximum number of buckets returned to the client. | `65000` |
 | `fast_field_cache_capacity` | Fast field in memory cache capacity on a Searcher. If your filter by dates, run aggregations, range queries, or even for tracing, it might worth increasing this parameter. The [metrics](../reference/metrics.md) starting by `quickwit_cache_fastfields_cache` can help you make an informed choice when setting this value. | `1G` |
 | `split_footer_cache_capacity` | Split footer in memory cache (it is essentially the hotcache) capacity on a Searcher.| `500M` |
-| `partial_request_cache_capacity` | Partial request in memory cache capacity on a Searcher. Cache intermediate state for a request, possibly making subsequent requests faster. Entries are keyed by a 128-bit hash of the request, with a negligeable probability of cache collision. It can be disabled by setting the size to `0`. | `64M` |
+| `partial_request_cache_capacity` | Partial request in memory cache capacity on a Searcher. Cache intermediate state for a request, possibly making subsequent requests faster. It can be disabled by setting the size to `0`. | `64M` |
 | `max_num_concurrent_split_searches` | Maximum number of concurrent split search requests running on a Searcher. | `100` |
 | `split_cache` | Searcher split cache configuration options defined in the section below. Cache disabled if unspecified. | |
 | `request_timeout_secs` | The time before a search request is cancelled. This should match the timeout of the stack calling into quickwit if there is one set.  | `30` |
--- a/docs/configuration/node-config.md
+++ b/docs/configuration/node-config.md
@@ -307,7 +307,7 @@ This section contains the configuration options for a Searcher.
 | `aggregation_bucket_limit` | Determines the maximum number of buckets returned to the client. | `65000` |
 | `fast_field_cache_capacity` | Fast field in memory cache capacity on a Searcher. If your filter by dates, run aggregations, range queries, or even for tracing, it might worth increasing this parameter. The [metrics](../reference/metrics.md) starting by `quickwit_cache_fastfields_cache` can help you make an informed choice when setting this value. | `1G` |
 | `split_footer_cache_capacity` | Split footer in memory cache (it is essentially the hotcache) capacity on a Searcher.| `500M` |
-| `partial_request_cache_capacity` | Partial request in memory cache capacity on a Searcher. Cache intermediate state for a request, possibly making subsequent requests faster. It can be disabled by setting the size to `0`. | `64M` |
+| `partial_request_cache_capacity` | Partial request in memory cache capacity on a Searcher. Cache intermediate state for a request, possibly making subsequent requests faster. Entries are keyed by a 128-bit hash of the request, with a negligeable probability of cache collision. It can be disabled by setting the size to `0`. | `64M` |
 | `max_num_concurrent_split_searches` | Maximum number of concurrent split search requests running on a Searcher. | `100` |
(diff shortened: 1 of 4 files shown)


Platane:
The check on split is smart
However it only works for list method, and list that have hits. This cache is mainly used to mark search that have 0 result. Because those allows to discard whole split at the cost of a few bits ( or so we thought 😬 ! )

I will argue to drop the check, and use a 256bit key if we want to be paranoid.

---

Also I think the 128 should count in the cache size, so we don't have memory going over the limit unexpectedly 

---

~Should we use sha256 instead ?~  I am convinced sip128 is fine

ncoiffier-celonis:
Thank you for the review @Platane !!

> The check on split is smart However it only works for list method, and list that have hits. This cache is mainly used to mark search that have 0 result. Because those allows to discard whole split at the cost of a few bits ( or so we thought 😬 ! )
> 
> I will argue to drop the check, and use a 256bit key if we want to be paranoid.

Done, see https://github.com/quickwit-oss/quickwit/pull/6727/commits/211fc3b84b8b5b042ed1956b3793dabb5b23c5c2

> 
> Also I think the 128 should count in the cache size, so we don't have memory going over the limit unexpectedly

For 2.4 millions entry, the heap overhead would be ~40MB for 128bits keys and ~80MB for 256bits keys. 

How would we count the 128 for the cache size? Do you mean something similar to https://github.com/quickwit-oss/quickwit/pull/6720?

> 
> ~Should we use sha256 instead ?~ I am convinced sip128 is fine

Sure. Worth noting the collision probability would go from 1e-27 with 128bits to 1e-65 with 256bits. Moreover, sha256 seems to be noticeably slower than spit128, after a bit of googling.

Platane:
just to be safe, let's generate cryptographically secure seeds. 

I committed on your branch + some clean up (some test no so relevant without the split id check)

I would like to merge that first, then https://github.com/quickwit-oss/quickwit/pull/6720

Platane (approved):

PR #6749, opened 2026-08-31
Lower default field list size limit to 10k

PSeitz-dd:
## Summary
- lower the default `QW_FIELD_LIST_SIZE_LIMIT` from 100,000 to 10,000, as responses containing 100,000 fields may exceed gRPC message size limits
- add a Quickwit-specific `limit` query parameter to the `_field_caps`, `_mapping`, and `_mappings` APIs
- propagate request-specific limits through root and leaf field merges; use `QW_FIELD_LIST_SIZE_LIMIT` when no request limit is provided
- document the new query parameter

## Tests
- `make fmt`
- `cargo nextest run -p quickwit-search list_fields`

### DIFF
--- a/quickwit/quickwit-proto/protos/quickwit/search.proto
+++ b/quickwit/quickwit-proto/protos/quickwit/search.proto
@@ -144,8 +144,7 @@ message ListFieldsRequest {
   // Maximum number of fields to return. Overrides QW_FIELD_LIST_SIZE_LIMIT.
   optional uint32 limit = 6;
 
-  // Control if the request will fail if split_ids contains a split that does not exist.
-  // optional bool fail_on_missing_index = 7;
+  reserved 7;
 }
 
 message LeafListFieldsRequest {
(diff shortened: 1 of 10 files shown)


quickwit/quickwit-proto/protos/quickwit/search.proto
```diff
@@ -141,8 +141,11 @@ message ListFieldsRequest {
   // When provided, only fields from documents matching this query are returned.
   optional string query_ast = 5;
 
+  // Maximum number of fields to return. Overrides QW_FIELD_LIST_SIZE_LIMIT.
+  optional uint32 limit = 6;
+
   // Control if the request will fail if split_ids contains a split that does not exist.
-  // optional bool fail_on_missing_index = 6;
+  // optional bool fail_on_missing_index = 7;
```
trinity-1686a: i'm not familiar with this field. Di you know if it should be considered reserved, or is it some documentation for a potential future improvement?

trinity-1686a (approved):

quickwit/quickwit-proto/protos/quickwit/search.proto
PSeitz-dd: I added it in the original PR, It was probably planned for a follow-up. 
https://github.com/quickwit-oss/quickwit/pull/4242/changes#diff-5eb9791a06ce038953b37caed18b46a0e2980c3276c70d9833c62336d8913f39R128

I'll change it to reserved, I don't think a commented entry makes sense.

PR #6762, opened 2026-09-03
Fix possible control-plane panic caused by race condition between reporting shard load and deleting index

ncoiffier-celonis:
### Description

On a test environment which uses Quickwit as a sidecar, and where I'm creating/deleting indices repeatedly, I think there is a possible race condition between reporting the shard load and deleting index, causing the control-plane actor to panic and to stay dead for ~30 seconds

This PR add 2 tests to show the problem, and fixes the `.expect("index should exist")` that causes the panic.

### How was this PR tested?

Unit tests only


fmassot:
LGTM. @nadav-govari can you take a look?

PR #6767, opened 2026-09-04
fix(datetime): reject a signed fractional part in a timestamp string

rawsun007:
`parse_timestamp_str` splits a timestamp string on `.` and parses the remainder with `i64::from_str`, which accepts a leading `+` or `-`. A sign is not a subsecond digit, so two malformed values parsed successfully into the wrong instant:

```
1700000000.-1          -> 1699999999990000000   (10ms *before* the whole second)
1700000000.+5          -> 1700000000050000000   (0.05s, not 0.5s)
1700000000.-999999999  -> 1699999999900000001
```

The `+5` case is wrong twice over: the sign occupies one of the nine digit slots, so `num_subsecond_digits` is 2 and the scale becomes `10^7`.

Neither is a rounding difference — an ingested timestamp field lands at a different time than the document says, with no error. `AGENTS.md` asks that unexpected states be treated as errors rather than silently succeeding, which is what this restores: the subsecond part must now be all ASCII digits, and anything else returns `None` so the remaining formats get their turn.

Unchanged: a trailing `.` (`123456789.`) still falls through to the existing empty-subseconds branch, digits past nine are still truncated, and `.abc` was already rejected.

Verification: `cargo test -p quickwit-datetime` — 34 passed. Removing just the new guard fails exactly the new test and nothing else. `cargo clippy -p quickwit-datetime --all-targets` is clean.

One gap to flag rather than paper over: I could not run `cargo +nightly fmt`, since this box has no nightly toolchain and `rustfmt.toml` is nightly-only. I kept the added lines under 100 columns and in the shape stable rustfmt agrees with, but please let CI's format check be the judge.

Generated with Claude Opus 5 (Claude Code), per the AI policy; the reproduction above was run locally.


quickwit/quickwit-datetime/src/date_time_parsing.rs
```diff
@@ -120,6 +120,13 @@ pub fn parse_timestamp_str(timestamp_str: &str) -> Option<TantivyDateTime> {
         if subsecond_digits_str.is_empty() {
             return parse_timestamp_str(timestamp_secs_str);
         }
+        // `i64::from_str` accepts a leading `+` or `-`, which is not a subsecond digit. Without
+        // this check, `1700000000.-1` parses as ten milliseconds *before* the whole second, and
+        // the sign also occupies one of the nine digit slots, so `1700000000.+5` yields 0.05s
+        // rather than 0.5s. Both are silently wrong instants rather than parse errors.
+        if !subsecond_digits_str.bytes().all(|byte| byte.is_ascii_digit()) {
```
guilload: Use `any`:
```rust
.any(|byte| !byte.is_ascii_digit())
```

quickwit/quickwit-datetime/src/date_time_parsing.rs
```diff
@@ -120,6 +120,13 @@ pub fn parse_timestamp_str(timestamp_str: &str) -> Option<TantivyDateTime> {
         if subsecond_digits_str.is_empty() {
             return parse_timestamp_str(timestamp_secs_str);
         }
+        // `i64::from_str` accepts a leading `+` or `-`, which is not a subsecond digit. Without
```
guilload: Remove the comment.

quickwit/quickwit-datetime/src/date_time_parsing.rs
rawsun007: Done, `any` with the negated predicate.

quickwit/quickwit-datetime/src/date_time_parsing.rs
rawsun007: Removed. `cargo test -p quickwit-datetime` still 34 pass, clippy clean.

guilload (approved):

PR #6781, opened 2026-09-08
feat(storage): add split range disk cache

congx4:
## Summary

- add a instance-level Foyer cache for split footer and body byte ranges including all kinds of data types(Fast fields, term dict, posting list, etc)
- configure cache capacity, admission, recovery, compression, reclaiming, and write throttling
- integrate the cache into searcher startup, shutdown, footer reads, and body reads
- export cache request, admission, and Foyer device metrics
- disable the long-lived fast-field RAM cache by default when Foyer is configured

## Test plan

- `make fmt`
- `cargo test -p quickwit-config --lib node_config::tests::test_split_range_disk_cache -- --nocapture`
- `cargo test -p quickwit-storage --all-features --lib split_range_cache -- --nocapture`
- `cargo test -p quickwit-search --all-features --lib split_range_cache_layer -- --nocapture`
- `cargo clippy -p quickwit-config -p quickwit-storage -p quickwit-search --all-features --tests -- -D warnings`

Made with [Cursor](https://cursor.com)


config/quickwit.yaml
```diff
@@ -210,6 +212,23 @@ indexer:
...
+#   # Omitted or null disables the cache. write_policy defaults to write-on-eviction.
+#   split_range_disk_cache:
+#     path: /quickwit/qwdata/split-range-v1
+#     disk_capacity: 1500G
+#     memory_capacity: 15G
+#     memory_eviction_policy: s3-fifo
+#     compression: lz4
+#     recover_mode: quiet
+#     block_size: 64M
+#     max_entry_size: 60M
```
PSeitz, line 225: a full float field will use 8byte per element

We target 10M docs, but we overshoot sometimes, so we may have 12M docs.
So the max size is closer to 100MB

what happens if the size is larger than `max_entry_size`?

docs/configuration/node-config.md
```diff
@@ -305,11 +305,12 @@ This section contains the configuration options for a Searcher.
...
 | `split_footer_cache_capacity` | Split footer in memory cache (it is essentially the hotcache) capacity on a Searcher.| `500M` |
 | `partial_request_cache_capacity` | Partial request in memory cache capacity on a Searcher. Cache intermediate state for a request, possibly making subsequent requests faster. It can be disabled by setting the size to `0`. | `64M` |
 | `max_num_concurrent_split_searches` | Maximum number of concurrent split search requests running on a Searcher. | `100` |
 | `split_cache` | Searcher split cache configuration options defined in the section below. Cache disabled if unspecified. | |
+| `split_range_disk_cache` | Process-wide on-disk cache for exact split footer and body byte ranges. Configuration options are defined in the section below. Cache disabled if unspecified. | |
```
PSeitz, line 313:
```suggestion
| `split_range_disk_cache` | Process-wide on-disk cache for split byte ranges. Configuration options are defined in the section below. Cache disabled if unspecified. | |
```

docs/configuration/node-config.md
```diff
@@ -323,6 +324,27 @@ This section contains the configuration options for the on-disk searcher split c
 | `max_num_splits` | Maximum number of splits allowed in the split cache.   | `10000` |
 | `num_concurrent_downloads` | Maximum number of concurrent download of splits. | `1` |
 
+### Searcher split range disk cache configuration
+
+This section contains the configuration options for the process-wide on-disk cache of exact split footer and body ranges. The cache is disabled when this section is omitted or set to `null`. If it is set and `fast_field_cache_capacity` is omitted, the long-lived fast field RAM cache is disabled; set a capacity explicitly to keep both.
```
PSeitz, line 329:
```suggestion
This section contains the configuration options for the process-wide on-disk cache of split byte ranges. The cache is disabled when this section is omitted or set to `null`. If it is set and `fast_field_cache_capacity` is omitted, the long-lived fast field RAM cache is disabled; set a capacity explicitly to keep both.
```

quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -403,17 +403,138 @@ impl SplitCacheLimits {
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum RecoverMode {
+    Quiet,
+}
+
+/// Disabled-by-default searcher disk cache for exact split byte ranges.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct SplitRangeDiskCacheConfig {
```
PSeitz, line 434: I would avoid putting to many foyer specific parameters in the config. We can have defaults and override them via env parameters instead.

quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -658,7 +790,7 @@ impl StorageTimeoutPolicy {
 impl Default for SearcherConfig {
     fn default() -> Self {
         SearcherConfig {
-            fast_field_cache: CacheConfig::default_with_capacity(ByteSize::gb(1)),
+            fast_field_cache: None,
```
PSeitz, line 793: why did you change the default here?

quickwit/quickwit-config/src/node_config/mod.rs
```diff
@@ -691,6 +824,20 @@ impl SearcherConfig {
     fn default_request_timeout_secs() -> NonZeroU64 {
         NonZeroU64::new(30).unwrap()
     }
+
+    /// Long-lived `.fast` RAM cache after applying defaults.
```
PSeitz, line 828: why long-lived? do we have a short-lived RAM cache?

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -159,16 +160,19 @@ async fn get_split_footer_from_cache_or_fetch(
...
-/// Returns hotcache_bytes and the split directory (`BundleStorage`) with cache layer:
-/// - A split footer cache given by `SearcherContext.split_footer_cache`.
+/// Returns hotcache_bytes and the split directory (`BundleStorage`).
 pub(crate) async fn open_split_bundle(
     searcher_context: &SearcherContext,
     index_storage: Arc<dyn Storage>,
     split_and_footer_offsets: &SplitIdAndFooterOffsets,
 ) -> anyhow::Result<(OwnedBytes, BundleStorage)> {
     let split_file = PathBuf::from(format!("{}.split", split_and_footer_offsets.split_id));
+    let foyer_storage: Arc<dyn Storage> = match &searcher_context.split_range_disk_cache_opt {
```
PSeitz, line 170: wrong variable name, the None case is not foyer

quickwit/quickwit-search/src/leaf.rs
```diff
@@ -159,16 +160,19 @@ async fn get_split_footer_from_cache_or_fetch(
...
+        Some(cache) => wrap_storage_with_split_range_cache(cache.clone(), index_storage.clone()),
```
PSeitz, line 171:
```suggestion
        Some(cache) => wrap_storage_with_disk_cache(cache.clone(), index_storage.clone()),
```

quickwit/quickwit-search/src/service.rs
```diff
@@ -418,6 +419,8 @@ pub struct SearcherContext {
     pub predicate_cache: Arc<PredicateCacheImpl>,
     /// Search split cache. `None` if no split cache is configured.
     pub split_cache_opt: Option<Arc<SearchSplitCache>>,
+    /// Process-wide split range disk cache. `None` if not configured.
```
PSeitz, line 422:
```suggestion
    /// Split range disk cache. `None` if not configured.
```
Process-wide  implies it's different, but it's not

quickwit/quickwit-search/src/service.rs
```diff
@@ -469,8 +475,12 @@ impl SearcherContext {
             searcher_config.max_num_concurrent_split_searches,
             searcher_config.warmup_memory_budget,
         );
-        let storage_long_term_cache =
-            Arc::new(QuickwitCache::new(&searcher_config.fast_field_cache));
+        let fast_field_cache = searcher_config.resolved_fast_field_cache();
+        let storage_long_term_cache = if fast_field_cache.capacity().as_u64() == 0 {
```
PSeitz, line 479: this check should be in QuickwitCache::new

quickwit/quickwit-serve/src/lib.rs
```diff
@@ -729,6 +729,20 @@ pub async fn serve_quickwit(
             None
         };
 
+    let split_range_disk_cache_opt = if node_config.is_service_enabled(QuickwitService::Searcher) {
+        match &node_config.searcher_config.split_range_disk_cache {
+            Some(config) => Some(Arc::new(
+                FoyerSplitRangeCache::open(config)
+                    .await
+                    .context("failed to open searcher split range disk cache")?,
```
PSeitz, line 737: when will this happen? data corruption?

we probably want a way to safeguard against corruption or format changes, and clear the cache instead of rendering it unusable

quickwit/quickwit-storage/src/split_range_cache/key.rs
```diff
@@ -0,0 +1,130 @@
...
+    reader
+        .read_exact(&mut buf)
+        .map_err(foyer::Error::io_error)?;
+    Ok(u64::from_le_bytes(buf))
+}
+
+/// Manual `Code` impl: keep Foyer's serde feature off so bincode is not pulled
+/// in for keys or values. Foyer 0.22.3 already implements `Code` for `Bytes`,
+/// so only this key needs a codec. Enabling serde would encode each value byte
+/// as an integer; for a 15 MiB payload that is millions of serializer visits
```
PSeitz, line 47: and that's a real performance problem?

quickwit/quickwit-storage/src/split_range_cache/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+use bytesize::ByteSize;
+use foyer::{Code, DeviceBuilder};
+pub(crate) use key::SplitRangeCacheKey;
+use quickwit_config::{
+    CachePolicy, DiskCompression, RecoverMode, SplitRangeCacheWritePolicy,
+    SplitRangeDiskCacheConfig,
+};
+pub use storage::{FoyerSplitRangeStorage, wrap_storage_with_split_range_cache};
+
+/// Process-wide Foyer hybrid cache for exact split byte-range payloads.
```
PSeitz, line 36:
```suggestion
/// Foyer hybrid cache for exact split byte-range payloads.
```

config/quickwit.yaml
congx4: it is just omitted. Foyer doesn't put any entries that is larger than the max_entry_size to disk. Do you think if we need to update to 128MB? From the bench tests around 1,2,4 QPS, there are total 5.8k entries that is rejected due to larger than the max_entry_size. https://dd.datad0g.com/dashboard/biv-7rp-k3i/byoc-logs-customer-telemetry-foyer?fromUser=true&fullscreen_end_ts=1789399115777&fullscreen_paused=true&fullscreen_refresh_mode=paused&fullscreen_section=overview&fullscreen_start_ts=1789396752165&fullscreen_widget=2364184038667261&graph-explorer__tile_def=N4IgbglgXiBcIBcIFsCmBnVAnCGQBoQtUBHAVwwXTgG1R0EBPAG1TlAAcBDVhBN%2BABMA9gHMA%2BgGNmXdOgiSCiRhwEh0w5hEFKA7toQALOCAB2wrMh4gAvoSaqTAIy5ZqhAGYXkZGdVh0IDwQss4qspjoAPTuIF6Wvlwm5NiMAIy2ALqEKTh4AaCCXAhc4hpkWJJqaAg4krGmXGjJFFjpSrmMJuhkyLDkCgDW%2BghlCBZcoqhlHFqjWFymU%2BKCEOiDUlyShtNcgshr8sKm4k7hchji4yXMwAAkFhLa%2BHfSZAzY4s935oLT2jYAARnQHAYiyY42AB0sikwjIpgQAAoAJRZQjEdAcY6YcTxKwIExINCYPLUGyZGxAA&refresh_mode=sliding&tpl_var_cluster_id%5B0%5D=yahoo-cong-foyer-small-18&from_ts=1789394453004&to_ts=1789408853004&live=true

quickwit/quickwit-config/src/node_config/mod.rs
congx4: sounds good, once I verify some settings, I will remove them from the config.

quickwit/quickwit-config/src/node_config/mod.rs
congx4: So the whole logic for changing the fast field cache is that we want to keep the existing fast field configuration when there is no disk cache defined. But when disk cache is defined, we want to remove fast field. 
so the fast field cache looks like below after the changes: 
omitted: disable it when Foyer is configured
explicitly configured: preserve the user’s value
omitted without Foyer: retain the effective 1 GB default

quickwit/quickwit-search/src/leaf.rs
congx4: I keep current name because SearchSplitCache is another disk cache from the code.

quickwit/quickwit-serve/src/lib.rs
congx4: Current disk cache config `recover_mode` is `quite` which skips recoverable corrupt blocks.  

This can happen when foyer can't open its device files or cache parameters are invalid.

quickwit/quickwit-storage/src/split_range_cache/metrics.rs
```diff
@@ -0,0 +1,445 @@
...
+);
+static _FOYER_STORAGE_DISK_IO_DURATION: LazyHistogram = lazy_histogram!(
+    name: "foyer_storage_disk_io_duration",
+    description: "foyer disk cache disk io duration",
+    system: "",
+    subsystem: "",
+    buckets: foyer_histogram_buckets("foyer_storage_disk_io_duration").unwrap(),
+);
+static _FOYER_STORAGE_ENTRY_SERDE_DURATION: LazyHistogram = lazy_histogram!(
+    name: "foyer_storage_entry_serde_duration",
```
fulmicoton-dd: why do we care about this?

quickwit/quickwit-storage/src/split_range_cache/metrics.rs
```diff
@@ -0,0 +1,445 @@
...
+    description: "Entries kept memory-only by admission checks",
+    subsystem: "storage",
+);
+static FAIL_OPEN_TOTAL: LazyCounter = lazy_counter!(
+    name: "split_range_disk_cache_fail_open_total",
+    description: "Foyer failures bypassed through lower storage",
+    subsystem: "storage",
+);
+
+fn foyer_histogram_buckets(name: &str) -> Option<Vec<f64>> {
```
fulmicoton-dd, line 41: I don't understand this at all. Why use a string as the key?

quickwit/quickwit-storage/src/split_range_cache/metrics.rs
```diff
@@ -0,0 +1,445 @@
...
+    description: "Split range disk cache requested bytes by result",
+    subsystem: "storage",
+);
+static ADMISSION_BYPASSES: LazyCounter = lazy_counter!(
+    name: "split_range_disk_cache_admission_bypasses_total",
+    description: "Entries kept memory-only by admission checks",
+    subsystem: "storage",
+);
+static FAIL_OPEN_TOTAL: LazyCounter = lazy_counter!(
+    name: "split_range_disk_cache_fail_open_total",
```
fulmicoton-dd: do we need this? can we just log an error?

quickwit/quickwit-storage/src/split_range_cache/metrics.rs
```diff
@@ -0,0 +1,445 @@
...
+
+fn foyer_histogram_buckets(name: &str) -> Option<Vec<f64>> {
+    match name {
+        "foyer_storage_op_duration" | "foyer_storage_disk_io_duration" => {
+            Some(Buckets::exponential(0.000_001, 2.0, 23))
+        }
+        "foyer_storage_inner_op_duration" => Some(Buckets::exponential(0.000_001, 2.0, 25)),
+        "foyer_storage_entry_serde_duration" => Some(Buckets::exponential(0.000_000_01, 2.0, 23)),
+        "foyer_storage_block_engine_buffer_efficiency" => Some(Buckets::linear(0.1, 0.1, 10)),
+        "foyer_storage_block_engine_recover_duration" => Some(Buckets::exponential(0.001, 2.0, 21)),
```
fulmicoton-dd: so many metrics we will never look at. Do we want all this?

quickwit/quickwit-storage/src/split_range_cache/storage.rs
```diff
@@ -0,0 +1,328 @@
...
+use super::{FoyerSplitRangeCache, SplitRangeCacheKey};
+use crate::stable_deref_bytes::into_owned_bytes;
+use crate::storage::SendableAsync;
+use crate::{
+    BulkDeleteError, ListObjectsStream, OwnedBytes, PutPayload, Storage, StorageError,
+    StorageErrorKind, StorageResult,
+};
+
+/// Foyer hybrid-cache entry header size in the 0.22.3 block engine.
+pub(crate) const FOYER_ENTRY_HEADER_SIZE: usize = 36;
```
fulmicoton-dd: what are these 36 bytes?

quickwit/quickwit-storage/src/split_range_cache/storage.rs
```diff
@@ -0,0 +1,328 @@
...
+        self.inner.file_num_bytes(path).await
+    }
+
+    fn uri(&self) -> &Uri {
+        self.inner.uri()
+    }
+}
+
+fn should_bypass_cache() -> bool {
+    fail_point!("split-range-cache-before-get", |_| true);
```
fulmicoton-dd: is this used?

fulmicoton-dd (approved):

quickwit/quickwit-storage/src/split_range_cache/storage.rs
congx4: this is the fixed header size from foyer. I have removed the admission-bypass metric, we don't need this any more.

quickwit/quickwit-storage/src/split_range_cache/metrics.rs
congx4: I removed all these metrics that didn't use from the dashboard here. https://dd.datad0g.com/dashboard/biv-7rp-k3i/byoc-logs-customer-telemetry-foyer?fromUser=true&refresh_mode=paused&tpl_var_cluster_id%5B0%5D=yahoo-cong-yahoo-parity&from_ts=1789586328245&to_ts=1789588128245&live=false&tile_focus=3574089649416361

quickwit/quickwit-storage/src/split_range_cache/metrics.rs
congx4: removed

quickwit/quickwit-storage/src/split_range_cache/metrics.rs
congx4: str here is just the metric name. Foyer’s RegistryOps API provides only: `name: Cow<'static, str>`. Therefore, we must inspect a string to identify the histogram. We could use some helper functions to replace this but we couldn't completely eliminate the string.

quickwit/quickwit-storage/src/cache/quickwit_cache.rs
```diff
@@ -40,6 +40,9 @@ impl From<Vec<(&'static str, Arc<dyn StorageCache>)>> for QuickwitCache {
 impl QuickwitCache {
     /// Creates a [`QuickwitCache`] with a cache on fast fields.
     pub fn new(cache_config: &CacheConfig) -> Self {
+        if cache_config.capacity().as_u64() == 0 {
+            return QuickwitCache::empty();
+        }
```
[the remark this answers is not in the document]
congx4: fixed in commit 81cf59d05. `QuickwitCache::new` now omits the `.fast` route only when the real capacity is zero **and** `virtual_caches` is empty, so zero-capacity policy experiments continue receiving accesses.

quickwit/quickwit-storage/src/split_range_cache/storage.rs
```diff
@@ -0,0 +1,215 @@
...
+        Err(unsupported_operation(&[path]))
+    }
+
+    async fn copy_to(&self, path: &Path, output: &mut dyn SendableAsync) -> StorageResult<()> {
+        self.inner.copy_to(path, output).await
+    }
+
+    async fn get_slice(&self, path: &Path, byte_range: Range<usize>) -> StorageResult<OwnedBytes> {
+        if byte_range.is_empty() {
+            return Ok(OwnedBytes::empty());
```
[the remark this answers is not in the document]
congx4: We are intentionally not adding decorator-specific handling for reversed ranges here. Callers are expected to provide valid half-open ranges.

config/quickwit.yaml
```diff
@@ -210,6 +212,23 @@ indexer:
...
 #      max_num_splits: 10000
 #      num_concurrent_downloads: 1
+#   # Foyer disk cache for exact split footer and body ranges.
+#   # Omitted or null disables the cache. write_policy defaults to write-on-eviction.
+#   split_range_disk_cache:
+#     path: /quickwit/qwdata/split-range-v1
+#     disk_capacity: 1500G
+#     memory_capacity: 15G
+#     memory_eviction_policy: s3-fifo
+#     compression: lz4
```
pabadrubio, line 222: Is this effective as default for this kind of data? I would assume we will mostly store data that has already some kind of compression and we will just waste CPU for very small size gains.

pabadrubio (approved):

config/quickwit.yaml
congx4: after testing, having a none compression performs better.

PR #6782, opened 2026-09-08
chore(CI): Run the full test suite automatically on every PR commit

dayaffe:
### Description

Runs make test-all on every commit pushed to any PR, from any author including forks. The only exception is a commit that touches nothing but markdown, docs/, or quickwit/quickwit-ui/, which is skipped. Everything else, including dependency bumps and any file type it doesn't recognize, runs the full suite.

This will be a required check.

### DIFF
--- a/.github/workflows/full-tests.yml
+++ b/.github/workflows/full-tests.yml
@@ -1,39 +1,28 @@
 name: Full test suite
 
-# Runs `make test-all` against a PR merge commit and reports status on its head.
+# Trusted dispatcher for the full test suite. `workflow_run` executes the
+# default-branch workflow definition, so PR authors cannot weaken the suite or
+# forge its result by editing it in the PR. It runs `make test-all` against the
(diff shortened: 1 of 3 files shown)


.github/workflows/full-tests.yml
```diff
@@ -59,82 +48,220 @@ jobs:
...
+    # (CodeQL `actions/untrusted-checkout` mitigation).
     permissions:
       contents: read
+      pull-requests: read
+    outputs:
+      pr_number: ${{ steps.resolve.outputs.pr_number }}
+      merge_sha: ${{ steps.resolve.outputs.merge_sha }}
+      affects_suite: ${{ steps.changes.outputs.affects_suite }}
     steps:
+      # Resolve the PR by head and a fresh merge SHA. Mergeability is computed
```
pront, line 61: @codex would it be cleaner to enable merge queue and make this a required check?

PR #6791, opened 2026-09-09
Use generation ID for routing table entries, and react to the ingester pool

nadav-govari:
### Description

Builds on https://github.com/quickwit-oss/quickwit/pull/6680. 

Currently, we rely on the ingester pool to gate ingester availability in the routing table. We never remove entries. As a result, we don't react when ingesters leave and rejoin the cluster. This mostly resolves itself on one-index clusters, because shards are more or less interchangeable. However, there's a logical problem with this on multi-index clusters: an ingester leaves the cluster, and rejoins, but might not get assigned any shards for a source it had previously, in which case it becomes a valid, inaccurate candidate for persist requests for that source.

In my testing with 9 source of varying size from "huge" to "very tiny", I observed a persistent 1% ingest error rate, that upon further investigation was due to this bug- requests were arriving at ingesters that simply didn't host any shards for that source, because they had done so in a previous incarnation.

The fix is twofold:

1. Ingesters leaving the ingester pool trigger a removal of their routing entries.
2. To ensure this is accurate, and that out of order events and other related race conditions don't occur, we now track generation ID on the ingester pool. This way, an ingester is only a candidate for routing if this exact incarnation is known to host shards for a source.

### How was this PR tested?

Unit testing. Live cluster testing incoming.


quickwit/quickwit-ingest/src/ingest_v2/routing_table.rs
```diff
@@ -310,12 +326,18 @@ impl RoutingTable {
             .sum();
 
         for (node_id, open_shard_count) in per_ingester_count {
+            let Some(generation_id) = ingester_pool.get(&node_id).map(|entry| entry.generation_id)
```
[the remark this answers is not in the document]
nadav-govari: This is the exact TODO that comes a line later. This is its own race condition. Unfortunately the fix for this is sending generation ID from the control plane, which becomes much more complicated, because it uses the shard proto which is also in the metastore model. I'll re-explore how invasive that is but when I looked into it it seemed like a mess.

quickwit/quickwit-ingest/src/ingest_v2/routing_table.rs
```diff
@@ -49,15 +51,14 @@ impl IngesterNode {
...
             return false;
         }
-        let is_ready = ingester_pool
+        let Some(ingester) = ingester_pool
             .get(&self.node_id)
-            .map(|ingester| ingester.status.is_ready())
-            .unwrap_or(false);
-
-        if !is_ready {
+            .filter(|ingester| ingester.status.is_ready())
```
fulmicoton-dd, line 56: super nitpick: I'd put that in a separate if statement 

let Some(ingester) = ... else { return false; }
ingester.generation_id != self.generation_id  
 && ingest.is_ready()

quickwit/quickwit-ingest/src/ingest_v2/routing_table.rs
```diff
@@ -264,8 +272,15 @@ impl RoutingTable {
             Ordering::Greater => return,
             Ordering::Equal => {}
         }
+        if let Some(existing) = entry.nodes.get(&node_id)
+            && existing.generation_id.as_u64() > generation_id.as_u64()
+        {
+            // drop a capacity update from an older incarantion of an ingester.
```
fulmicoton-dd, line 278:
```suggestion
            // drop a capacity update from an older incarnation of an ingester.
```

fulmicoton-dd (approved):

quickwit/quickwit-ingest/src/ingest_v2/routing_table.rs
nadav-govari: This changes the logic unfortunately- if it's not ready, we need to add it to unavailable ingesters.

PR #6793, opened 2026-09-10
add a timeout layer to our gcp connector

trinity-1686a:
currently on the query path, the timeout hitting is the leaf search one, at 30s, which causes a bunch of work to be loss because of one flaky connection


trinity-1686a:
replaced by #6802

PR #6795, opened 2026-09-14
Zonally aware ingest controller

nadav-govari:
### Description

Refactors the ingest controller to favor zonal balance by spreading shard assignments evenly among all known zones, if availability zone awareness is enabled. If AZ awareness is disabled, all of the zonal logic reduces to one global zonal group (keyed by `None`) and falls back to the logic as it exists today.

It does this by turning try_open_shards into a zonal function called for each zone. `try_open_shards` has three callsites: 
* get_or_create_open_shards
* scale_up_shards
* rebalance_shards

get_or_create_open_shards and scale_up_shards create net-new shards, which we ideally want to spread evenly across zones.

rebalance_shards, however, favors keeping shards within the same zone; otherwise, each zone will net-lose shards on a rebalance (since we have to close the original ones). It still favors global equilibrium first, but breaks ties within the same zone.

This allows us to then have an even, strongly local and dense scheduling plan, which is introduced in https://github.com/quickwit-oss/quickwit/pull/6694. 

### How was this PR tested?

Unit testing and extensive live cluster testing.


quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,54 +92,200 @@ fn fire_and_forget(
...
-    let position = rng.random_range(0..ingester_ids.len());
+) -> Option<&'a EligibleIngester> {
+    let min_load = eligible_ingesters
+        .iter()
+        .map(|ingester| ingester.num_open_shards.load(Ordering::Relaxed))
+        .min()?;
+    let minima: Vec<&EligibleIngester> = eligible_ingesters
+        .iter()
+        .filter(|ingester| ingester.num_open_shards.load(Ordering::Relaxed) == min_load)
+        .collect();
```
fulmicoton-dd, line 132: You are asking for trouble by using two loops here.
minima could be empty while the reader would assume it is impossible.

Can we do a single loop and update a Vec and min_load together?

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,54 +92,200 @@ fn fire_and_forget(
...
+    if eligible_ingesters.is_empty() {
+        return None;
     }
     let mut rng = rng();
     let mut ingester_ids = Vec::with_capacity(num_shards);
     for _ in 0..num_shards {
-        let ingester_id = pick_least_loaded_ingester(&mut ingester_ids_by_num_shards, &mut rng)?;
-        ingester_ids.push(ingester_id);
+        let picked = pick_least_loaded(eligible_ingesters, requested_zone.as_ref(), &mut rng)
+            .expect("eligible ingesters non-empty");
```
fulmicoton-dd: no guarantee that this is non empty. see comment above...

Or maybe the guarantee is given by a hidden contract, but we don't want that.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,54 +92,200 @@ fn fire_and_forget(
...
-/// Pick an ingester from `ingester_ids_by_num_shards`.
-/// We prioritize ingesters with the least number of shards and break ties randomly.
-///
-/// Once an ingester has been found, we update `ingester_ids_by_num_shards` to reflect the new
-/// state. In particular, the ingester node is moved from its previous num_shards level to its new
-/// num_shards level. In particular, a num_shards entry that is empty should be removed from the
-/// BTreeMap.
-fn pick_least_loaded_ingester<'a>(
-    ingester_ids_by_num_shards: &mut BTreeMap<usize, Vec<&'a NodeId>>,
+type Zone = String;
```
fulmicoton-dd: nitpick: I suspect `Arc<str>` would be better here, but it is hard to tell without going through the work of trying it out.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,54 +92,200 @@ fn fire_and_forget(
...
+fn total_shards(source_shard_counts: &SourceShardCount) -> usize {
+    source_shard_counts.values().sum()
+}
+
+/// In some cases, it could be advantageous to prefer to place shards in a specific zone (such as
+/// when rebalancing the shards of a decommissioned ingester). Otherwise, the default is to spread
+/// new shards evenly.
+enum ShardPlacement {
+    Balanced(SourceShardCount),
+    Zoned(HashMap<Option<Zone>, SourceShardCount>),
```
fulmicoton-dd, line 108: I don't fully understand the semantics here.

What's with the Option<Zone> key. Does it express absence of zones? In that case we are supposed to have a single key? In that case could we just use the Balanced Variant and remove the Option< in the key?

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,54 +92,200 @@ fn fire_and_forget(
...
+    let minima: Vec<&EligibleIngester> = eligible_ingesters
+        .iter()
+        .filter(|ingester| ingester.num_open_shards.load(Ordering::Relaxed) == min_load)
+        .collect();
+    let same_zone_minima: Vec<&EligibleIngester> = minima
+        .iter()
+        .copied()
+        .filter(|ingester| requested_zone.is_some() && ingester.zone.as_ref() == requested_zone)
+        .collect();
+    let candidates = if !same_zone_minima.is_empty() {
```
fulmicoton-dd: super nitpick: positive condition

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,54 +92,200 @@ fn fire_and_forget(
...
+    }
+    let mut shuffled: Vec<&Zone> = zones.iter().collect();
+    shuffled.shuffle(&mut rng());
+    shuffled
+        .iter()
+        .cycle()
+        .take(num_to_open)
+        .map(|zone| Some((*zone).clone()))
+        .counts()
+}
```
fulmicoton-dd, line 223: well, i guess it is readable and we don't care about efficiency. :)

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,54 +92,200 @@ fn fire_and_forget(
...
+    eligible_ingesters: &[EligibleIngester],
+) -> HashMap<Option<Zone>, SourceShardCount> {
+    let zones: HashSet<Zone> = eligible_ingesters
+        .iter()
+        .filter_map(|ingester| ingester.zone.clone())
+        .collect();
+    let mut num_shards_by_source_by_zone: HashMap<Option<Zone>, SourceShardCount> = HashMap::new();
+    for (source_uid, num_shards) in source_shard_counts {
+        // Number of shards to open for this source in each zone.
+        for (zone, count) in distribute_shards_across_zones(num_shards, &zones) {
```
fulmicoton-dd, line 237: hmmm. we have two "wanted" features getting in conflict here.

You want:
- indexer/ingester locality 
- spreading shards from a same source
- a minimal number of pipeline.

I'm not 100% sure I understand how the trade-off is taken between the 3, and whether the result is a function of the state path of the system or whether it is a deterministic thing. (e.g. today it is very path dependent and it is part of our pain)

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,54 +92,200 @@ fn fire_and_forget(
...
+        }
+    }
+    num_shards_by_source_by_zone
+}
+
+/// Matches successful replacement opens to shards that are still hosted by live ingesters.
+///
+/// The ingester pool can change while replacements are being opened. A shard whose ingester has
+/// disappeared from the pool can no longer be closed directly, so it is deliberately left for the
+/// control plane's self-healing mechanisms instead of turning normal cluster churn into a panic.
```
fulmicoton-dd, line 253: I have trouble understanding the nature of opened_by_zone and shards_to_rebalance here.

This is used when we move shards. Moving is open and close. We did the opening already and so we need to identify shards to close?

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,54 +92,200 @@ fn fire_and_forget(
...
+                num_matched += 1;
+            }
+            if num_matched < num_opened {
+                warn!(
+                    index_uid = %source_uid.index_uid,
+                    source_id = %source_uid.source_id,
+                    ?requested_zone,
+                    num_opened,
+                    num_matched,
+                    "could not match every replacement shard to a live predecessor"
```
fulmicoton-dd: Can this happen? We could have a comment or something explaining when this happens (an ingester went done in the middle of rebalance or something)

fulmicoton-dd: shouldn't the warn be before the break? and could num_matched be the loop argument?

`for num_matched in 0..num_opened`

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: I thought about this and forgot to come back and do this. I wanted to do it anyway. There's usually no more than 3 zones and we reference it a lot.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: Yes. Zones are not a requirement. If there are no zones, we end up with the None key as the one global zone and fall back to the existing behavior. So zones are expressed as Option everywhere as a result.

But I think I see what you're saying, is that if it's none, we can use balanced instead. Let me see if I can rework that.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: It's not quite so hidden - we check eligible_ingesters is_empty at line 191 above, and then pick_least_loaded picks the minimum, so getting back a None should be impossible

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: Addressed below - this shouldn't be able to be empty based on the call patterns, but I can bake it more explicitly into the contract if you prefer. And if you prefer the single loop/vec for readability I'm happy to make the change.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: An upcoming PR addresses this. Basically, without running an optimization solving state machine with exponential complexity, we get to choose one dimension that we actually maximize, and then do our best with the rest. Your original scheduling code did this with density.

I kept the same idea. Our priorities are:
* when creating new shards, attempt to spread them evenly in different zones
* then, create pipelines that are as dense as possible, minimizing the number of them
* then, within those pipelines, optimize them so that we prefer to be within the same zones.
  * This is coming in another PR in a little bit.
 
So locality is a nice-to-have, and a nice optimization we can do when it doesn't cost us anything in the other dimensions.
It degenerates a little bit when a source is small as you might not be able to avoid cross-az. But by nature we don't really care about this because the cost is minimal.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
fulmicoton-dd: Quite hidden!
There are different levels to advertise a contract.

Ideally you do that through the typesystem.
Otherwise you do that with doc + assert

The trouble with that kind of contract is that years later, someone could add a functionality that mutates these atomic in a concurrent manner and cause an issue.
After all, they are Atomics.

nadav-govari: Okie dokie, I'll fix :)

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: Rebalancing consists of first opening the new shards, then upon success, closing the existing ones. The condition that also exists today is - only close shards when a corresponding shard was actually successfully opened.

Doing zonality in rebalancing consists of: first, we want to try to keep the shards within the same zone if it doesn't hurt global equilibrium, but if it does, then spread it to other zones.

As a result, we want to close a shard from the original zone from which it came. The naming of "requested_zone" is awful, I'll fix it.

For example,

A: 7 shards
B: 5 shards
C: 6 shards

Rebalancing will want to move a shard from A to B. However, the locality preference will "request" to stay in A. This is how the invariant is enforced.

Instead, it's opened in B because this creates a better global balance.

But to match this with an equivalent shard to close, we need to know which zone it originally came from. We use the "requested" zone for this, A.

(For the invariant: by 'preferring' to stay in the same zone, but requiring global equilibrium first, we fill up the zones equally).

nadav-govari: This is written the way it is, where it tries to find a shard to open rather than just closes the vec completely, because opening shards is not guaranteed, and this gives that a lot of respect as something that can plausibly happen.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: Yes, it can, if some shards were not successfully opened. I'll add a comment. And I'll fix that too.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: Fixed with asserts and types.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: We need same_zone_minima first, so it's either !is_empty, or len > 0.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: And its cool :)

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: I went through everything to try to remove the Option in general. It's somewhat hard, because the code needs to support cases that are both zonal and not. The problem is that methods that care about zones (such as allocate_shards) legitimately need an option, or we end up needing to derive whether or not we need zonal placement at calltime, and I'd argue this is worse. I can make it abundantly clear through comments how to reason about this Option code.

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
nadav-govari: Fixed - upstreamed the emptiness check, assert that its not empty here

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,52 +92,198 @@ fn fire_and_forget(
...
+    assert!(!eligible_ingesters.is_empty());
+    let min_load = eligible_ingesters
+        .iter()
+        .map(|ingester| ingester.num_open_shards.load(Ordering::Relaxed))
+        .min()
+        .expect("There should be at least one eligible ingester");
+    let minima: Vec<&EligibleIngester> = eligible_ingesters
+        .iter()
+        .filter(|ingester| ingester.num_open_shards.load(Ordering::Relaxed) == min_load)
+        .collect();
```
fulmicoton-dd: the comment was not addressed. On the second loop, 
this function alone gives you no guaranteed that num_open_shards has not changed.

Can we do this thing in a single loop with a single call to .load Relaxed so that we can avoid  paranormal behavior in the future?

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
```diff
@@ -91,52 +92,203 @@ fn fire_and_forget(
...
-    let (&num_shards, ingester_ids) = ingester_ids_by_num_shards.iter_mut().next()?;
-    let position = rng.random_range(0..ingester_ids.len());
+) -> &'a EligibleIngester {
+    assert!(!eligible_ingesters.is_empty());
+    let mut min_load = usize::MAX;
+    let mut minima = Vec::new();
+    let mut same_zone_minima = Vec::new();
+
+    for ingester in eligible_ingesters {
+        let load = ingester.num_open_shards.load(Ordering::Relaxed);
```
fulmicoton-dd, line 127: thank you

quickwit/quickwit-control-plane/src/ingest/ingest_controller.rs
fulmicoton-dd: you are right, it actually makes more sense this way.

fulmicoton-dd (approved):

PR #6796, opened 2026-09-14
Add indexing scheduling plan density optimization

nadav-govari:
### Description

Introduces a hill climbing density optimization pass for the scheduling algorithm. Copied from the Readme:

The idea behind optimization is simple: if we're able to perform one single action that improves the plan, over time, we'll eventually
converge on a plan that's quite good. This is a hill climbing algorithm - we don't attempt to solve the optimization problem in one shot,
because 1. It's very complex, and the algorithm to do so is (dangerously) exponential, and 2. it's very disruptive - we'd likely reset a lot of pipelines at once.

Optimization is gated on three things: locality being enabled, the plan being stable, and a 30 second cooldown.

The stability requirement is because resetting a pipeline, which optimization does, is expensive, as
we throw away uncommitted work for the shards in that pipeline. So we try to perform optimization only on a stable cluster.
This works doubly well with deployments, because deployments, which are already resetting many pipelines in their own right, also leave the cluster
in a chaotic state. Optimization can kick in once the plan is stable (and the deployment finished).

Optimization itself works in a naive, greedy, simple way. For each source, find two pipelines that are running under capacity; rearrange the shards each is indexing
so that one pipeline fills up. 

If we do this enough times, eventually, we'll end up with the minimal number of pipelines, and optimal density.

### How was this PR tested?

Unit tests and extensive cluster tests.


quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
```diff
@@ -0,0 +1,297 @@
...
+    }
+    state.last_plan_repair_attempt_timestamp = Some(Instant::now());
+    true
+}
+
+pub(super) fn conditionally_optimize_plan(
+    physical_plan: &mut PhysicalIndexingPlan,
+    previous_plan: Option<&PhysicalIndexingPlan>,
+    sources: &[SourceToSchedule],
+    can_optimize_plan: bool,
```
fulmicoton-dd, line 51: If find passing `can_optimize_plan` very strange. Why isn't on the caller side?

if can_optimize_plan {
    optimize_plan(..)
}

fulmicoton-dd: ah I guess I would entirely inline that function on the single call site.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
```diff
@@ -0,0 +1,297 @@
...
+) {
+    if !can_optimize_plan || previous_plan != Some(physical_plan) {
+        return;
+    }
+    repair_physical_plan_density(physical_plan, sources);
```
fulmicoton-dd, line 56: Maybe this should be called:

    improve_physical_plan_density(physical_plan, sources);

I understand why you mean repair though. Your call.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
```diff
@@ -0,0 +1,297 @@
...
+    indexer_id: NodeId,
+    donor_task_ord: usize,
+    receiver_task_ord: usize,
+}
+
+fn repair_physical_plan_density(
+    physical_plan: &mut PhysicalIndexingPlan,
+    sources: &[SourceToSchedule],
+) -> bool {
+    for source in sources {
```
fulmicoton-dd, line 105: nitpick:

you are trying to express let's search for repair opportunities.
As soon as we find one, apply it and return true.

Maybe the code structure could express that logic better.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
```diff
@@ -0,0 +1,297 @@
...
+        if !matches!(source.source_type, SourceToScheduleType::Sharded { .. }) {
```
fulmicoton-dd, line 106: nitpick, we should probably add a is_sharded() accessor at one point.
I think I saw this pattern a lot.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
```diff
@@ -0,0 +1,297 @@
...
+            .enumerate()
+            .filter(|(_, task)| {
+                task.index_uid.as_ref() == Some(&source.source_uid.index_uid)
+                    && task.source_id == source.source_uid.source_id
+                    && task.params_fingerprint == source.params_fingerprint
+            })
+            .map(|(task_ord, _)| task_ord)
+            .collect();
+        task_ords.sort_by_key(|task_ord| (indexing_tasks[*task_ord].shard_ids.len(), *task_ord));
+        let [donor_task_ord, receiver_task_ord, ..] = task_ords.as_slice() else {
```
fulmicoton-dd, line 137: O_O I did not know we could do that. That's so cool!

fulmicoton-dd: so you only consider merging opportunities right?
if max_num_shards == 3, and you have 3 pipelines with 2 shards, density won't get improved?

fulmicoton-dd (approved):

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
nadav-govari: Can do

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
nadav-govari: Can do here as well

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
nadav-govari: Tweaked it, and added comments.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
nadav-govari: Good idea, added.

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
nadav-govari: Yes, I considered this. The approach is then to always collapse pipelines so that one is full. ie. 2 + 2 should leave 3 + 1, and then the next pass will fill the 1.

There were two drawbacks to this:
The first, it will take longer for the plan to settle, and it will reset more pipelines. It will also create partially full pipelines for some period of time, ie. to go from pipelines of 2/2/2 to 3/3, you first need to go to 3/1/2, then 3/3, which means you're also producing more small splits until that part is done. Because indexers are shuffled (important to not hammer one indexer/keep one shard from getting indexed for too long), it could be some number of minutes before it's re-addressed.

The second is that it's more code, and a little more complexity as a result.

We can do this. The density percentage (% of pipelines in our plan compared to the mathematical minimum) difference was about 10% . (ie. 240 pipelines vs 260 pipelines). The net result is ~two less compactions and write amplification. Perhaps it's worth it.

(thinking out loud).

PR #6799, opened 2026-09-14
Add locality optimization to indexing scheduler

nadav-govari:
### Description

Following the density hill climbing optimization, also add a locality hill climbing step that starts after density optimization is complete. From the Readme:

Then, once pipelines are sufficiently dense, for each source, find one pipeline in each zone at random that has cross-zonal shards.
Perform a simple swap of those shards to improve locality. Eventually, over time, we'll converge at a locality optimum given the current plan.

This optimization is nice and will improve a plan as much as is theoretically possible given enough time.

### How was this PR tested?

Unit tests and extensive cluster tests.


quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
```diff
@@ -163,20 +180,198 @@ fn apply_density_repair(physical_plan: &mut PhysicalIndexingPlan, repair: Densit
...
 }
 
+struct LocalityRepairPipeline {
+    indexer_id: NodeId,
+    task_ord: usize,
+    availability_zone: String,
+    shard_ids: Vec<ShardId>,
+}
+
+fn repair_physical_plan_locality(
```
fulmicoton-dd: Does it keep density equal? Can you comment that invariant if it is one? Would it be testable with asserts?

fulmicoton-dd: I think your intent was:
this leaves density constant
Iff the physical plan is mutated, we return true, and the locality is decreased in the new plan.

quickwit/quickwit-control-plane/src/indexing_scheduler/mod.rs
```diff
@@ -693,11 +699,42 @@ fn get_shard_locality_metrics(
     }
     ShardLocalityMetrics {
         num_remote_shards,
-        num_nearby_shards,
+        num_zonal_shards,
         num_local_shards,
     }
 }
 
+fn get_indexing_plan_density(
```
fulmicoton-dd, line 712:
```suggestion
/// Density is a (0..1] metric that measures whether we index shards on a minimal number 
/// of pipelines. The higher the better.
fn get_indexing_plan_density(
```

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/README.md
```diff
@@ -121,10 +121,15 @@ This works doubly well with deployments, because deployments, which are already
 in a chaotic state. Optimization can kick in once the plan is stable (and the deployment finished).
 
 Optimization itself works in a naive, greedy, simple way. For each source, find two pipelines that are running under capacity; rearrange the shards each is indexing
-so that one pipeline fills up. 
+so that one pipeline fills up.
 
 If we do this enough times, eventually, we'll end up with the minimal number of pipelines, and optimal density.
 
+Then, once pipelines are sufficiently dense, for each source, find one pipeline in each zone at random that has cross-zonal shards.
```
fulmicoton-dd:
```suggestion
Then, once we cannot more opportunities to increase density, for each source, find one pipeline in each zone at random that has cross-zonal shards.
```

fulmicoton-dd (approved):

quickwit/quickwit-control-plane/src/indexing_scheduler/scheduling/optimization.rs
nadav-govari: Yes, thats right. I can add that.

PR #6802, opened 2026-09-15
improvements to our gcs storage

trinity-1686a:
### Description

- add timeout and stalled-stream-protection like behaviour to individual calls
  - separate read and write operator, we send big chunks and that doesn't work well with the SSP-like timeout of opendal (it wants one chunk send by the deadline, but our chunks can be too big for that to happen, especially on compactors)
- share underlying operator between calls to storage factory, to share the underlying connection pool
- use proper caching of credential when doing service-account based authentication
  - opendal always cache authentication, but for gcs S-A auth, it only caches the service account keys (usually read from file) as they are enough to sign presigned requests. For non-presigned request, it perform au oauth flow per query *outside* the caching machinery, so there is effectively no caching of authentication
  - for other auth kinds, it seems to do the right thing
  - implement a ProvideCredential which calls the main credential chain, and if it gets a service account and no token, calls into a ServiceAccountTokenCredentialProvider (not included in any tagged release yet) to let the caching machinery work with the token

### How was this PR tested?

deployed on a test cluster, with good improvement in search reliability and latency

### DIFF
--- a/quickwit/quickwit-storage/src/opendal_storage/google_cloud_storage.rs
+++ b/quickwit/quickwit-storage/src/opendal_storage/google_cloud_storage.rs
@@ -20,11 +20,20 @@ use async_trait::async_trait;
 use quickwit_common::uri::Uri;
 use quickwit_config::{GoogleCloudStorageConfig, StorageBackend};
 use regex::Regex;
+use reqsign_core::{Context, ProvideCredential, ProvideCredentialChain};
+use reqsign_google::{
+    Credential, DefaultCredentialProvider, FileCredentialProvider,
+    ServiceAccountTokenCredentialProvider,
+};
 use tracing::info;
 
 use super::OpendalStorage;
 use crate::debouncer::DebouncedStorage;
 use crate::{Storage, StorageFactory, StorageResolverError};
+
+// Matches opendal's DEFAULT_GCS_SCOPE, which is more restrictive than `cloud-platform` used by
+// default in reqsign_google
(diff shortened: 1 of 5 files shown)


quickwit/quickwit-storage/src/opendal_storage/google_cloud_storage.rs
```diff
@@ -12,28 +12,44 @@
...
 use crate::{Storage, StorageFactory, StorageResolverError};
 
+// Matches opendal's DEFAULT_GCS_SCOPE, which is more restrictive than `cloud-platform` used by
+// default in reqsign_google
+const GCS_SCOPE: &str = "https://www.googleapis.com/auth/devstorage.read_write";
+
 /// Google cloud storage resolver.
 pub struct GoogleCloudStorageFactory {
     storage_config: GoogleCloudStorageConfig,
+    // this is technically an unbounded cache, the bound is the number of indexes that have existed
```
guilload: Not exactly true for Airmail-like use cases. We already have a simple LRU cache in the deps, right?

quickwit/quickwit-storage/src/opendal_storage/google_cloud_storage.rs
```diff
@@ -12,28 +12,44 @@
...
+    // since node start, which should be low enough this isn't catastrophic
+    storage_cache: Mutex<HashMap<Uri, Arc<DebouncedStorage<OpendalStorage>>>>,
```
nadav-govari, line 43: `>>>>`
(lol)

nadav-govari (approved):
Terrific (Trinity) work

PR #6804, opened 2026-09-16
fix(datetime): accept an eight-digit subsecond java date format token

rawsun007:
### Description

`SSSSSSSS` is missing from `JAVA_DATE_FORMAT_TOKENS`. The tokenizer walks that list in order and takes the first match, so eight `S` matched `SSSSSSS` and left one `S` over, producing two consecutive subsecond components. `build_fraction_of_second_item` uses `SubsecondDigits::OneOrMore`, so the first one consumes every digit and the second finds none:

```
the 'subsecond' component could not be parsed
```

Every width from one to nine `S` works except eight, for any input:

| `S` count | `2024-01-02 03:04:05.123456789` |
| --- | --- |
| 1..7 | parses |
| **8** | **fails** |
| 9 | parses |

That asymmetry is why this reads as an omission rather than a deliberate limit — the digit count is ignored everywhere else, since all widths map to the same `OneOrMore` item.

### How was this PR tested?

`test_parse_java_datetime_format_every_subsecond_width` loops one to nine and parses the same nanosecond input with each. On `main` it fails at eight; with the fix `cargo test -p quickwit-datetime` is 35 passing, clippy clean.

No nightly toolchain here and `rustfmt.toml` is nightly-only, so I kept the new match arm under 100 columns by hand rather than running stable `cargo fmt` over the file — please let the CI format check be the judge, as on #6767.

Written by Claude Opus 5 in Claude Code, running under my account.


rawsun007:
Pushed the formatting CI asked for, applied verbatim from the `cargo +nightly fmt` diff rather than guessed at:

- adding `"SSSSSSSS"` with no trailing comment changed the alignment block, so the comments on the two lines below lose a space each;
- the match arm becomes `| "SS" | "S" => build_fraction_of_second_item(token),` on one line rather than a braced block.

No nightly toolchain on this machine, which is why the first push had my hand-wrapping instead — the PR body flagged that and said CI should be the judge. It was.

`cargo test -p quickwit-datetime` is still 35 passing and clippy is clean on the reformatted file.

Written by Claude Opus 5 in Claude Code, running under my account.

trinity-1686a (approved):
thanks for contributing!

Sadly we cannot merge unsigned commits. Can you please sign yours please? Github provides some [instructions on how to do it](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification). I recommend using SSH signatures as it lets you manage a single key for signing and pushing (assuming you use SSH to push to Github)

rawsun007:
Thanks for the review, and understood on the signature.

Signing needs a key belonging to the account, which is not something the agent that wrote this patch can set up — @rawsun007 will generate an SSH signing key, register it, and re-sign the commit. Expect a force-push on this branch rather than a new PR.

The formatting fix from the earlier `cargo +nightly fmt` failure is already on the branch; CI has not re-run since, as the fork workflow approval does not survive a push.

Written by Claude Opus 5 in Claude Code, under @rawsun007's account.

rawsun007:
Signed — `1171a4e` now reads `verified: true`.

Nothing else changed on the branch: it still carries the `cargo +nightly fmt` fix from the earlier Lints failure. CI has not re-run since that push, as the fork workflow approval does not survive one.

Written by Claude Opus 5 in Claude Code, under @rawsun007's account.

PR #5125, opened 2024-06-14
5117 disable per index metrics

fulmicoton:

PR #6806, opened 2026-09-17
fix: indexing_pipelines metric doesn't honor the QW_DISABLE_PER_INDEX_METRICS flag

ncoiffier-celonis:
### Description

indexing_pipelines metric doesn't honor the QW_DISABLE_PER_INDEX_METRICS flag

Similar to what was done with https://github.com/quickwit-oss/quickwit/pull/5125. This new metric was added later with https://github.com/quickwit-oss/quickwit/pull/5880, that's probably why the flag was missed.

### How was this PR tested?

Regression unit tests only, no new tests added.


ncoiffier-celonis:
@fulmicoton-dd since you worked in the original PR, maybe you could have a look into this one?

PR #6808, opened 2026-09-17
Write split on source EOF

nadav-govari:
### Description

When decommissioning an indexer, all of a pipeline's shards may reach EOF well before the indexing pipeline timeout (30s). This adds a hint from the ingest source, passed through the doc processor, to the indexer that it can commit its workbench with the trigger `NoMoreDocs`.

### How was this PR tested?

Unit tests.


guilload:
Have you considered simply forcing the commit in the source when detecting that all the assigned shards are at EOF?
```rust
        if self
            .assigned_shards
            .values()
            .all(|assigned_shard| assigned_shard.current_position_inclusive.is_eof())
        {
            batch_builder.force_commit();
        }
```
It feels like this would be a more narrow change.

nadav-govari:
> Have you considered simply forcing the commit in the source when detecting that all the assigned shards are at EOF?
> 
> ```rust
>         if self
>             .assigned_shards
>             .values()
>             .all(|assigned_shard| assigned_shard.current_position_inclusive.is_eof())
>         {
>             batch_builder.force_commit();
>         }
> ```
> 
> It feels like this would be a more narrow change.

@guilload It's just much better overall, I thought that force commit here was on the batch itself for some reason. Simplified it

quickwit/quickwit-indexing/src/source/ingest/mod.rs
```diff
@@ -253,6 +253,15 @@ impl IngestSource {
             )
             .context("failed to record partition delta")?;
         assigned_shard.current_position_inclusive = to_position_inclusive;
+        if self
+            .assigned_shards
+            .values()
+            .all(|shard| matches!(shard.status, IndexingStatus::ReachedEof))
```
guilload, line 259: I think we want to match on `IndexingStatus::Complete | IndexingStatus::ReachedEof` or test for `current_position_inclusive.is_eof()`

guilload (approved):

